PR #2, opened 2017-12-01
Create specs for running distributed training with/without GPUs.

jlewi:
Some timing measurements

2 Worker CPU: 45.934 sec/step
2 Worker GPU: 4.659 sec/step 

@vishh  @foxish PTAL


vishh:
LGTM. `kubectl apply -R -f kubeflow` would run these examples as well. May be we should move all the manifests into a single directory before merging this?

vishh:
lgtm

PR #5, opened 2017-12-01
Move TF Job CRD to a more appropriately named directory

vishh:
cc @foxish @jlewi


foxish:
tf-operator?

vishh:
I don't mind the rename, I thought we referred to `Operators` as `Controllers`.
cc @kow3ns

vishh:
@foxish let's merge this as-is and rename again if necessary.

PR #8, opened 2017-12-04
Resolve TODO for RBAC on GKE 1.8+

foxish:
cc @vishh @jlewi 

jeremy - we also need this for the operator. Is that correct?


foxish:
As a follow-up, we might want to move the setup section to the top, before quick-start.

vishh:
lgtm

PR #28, opened 2017-12-08
Update Contributing.md, and add license headers

foxish:
@vishh @aronchick @jlewi PTAL

Using guidelines from https://github.com/google/new-project


vishh:
LGTM. Thanks for cleaning up @foxish

Issue #23, opened 2017-12-08
Tooling to manage configuration and deployment


foxish:
This is an issue to discuss how we manage the configuration in this repo. Options that have come up so far:

* The existing kubectl apply workflow
* helm
* ksonnet

Since we want a unified approach to deployment of the entire suite here, 
we should pick a direction that's best suitable for current and future components here.

@jlewi @kow3ns @yuvipanda

foxish:
Forking from the discussion in https://github.com/google/kubeflow/issues/19#issuecomment-350173280

yuvipanda:
Thanks for starting this, @foxish! 

I'll note that helm and ksonnet are not entirely mutually exclusive, with plugins like https://github.com/technosophos/helm-ksonnet. This is an evolving space, so lots of moving parts!

vishh:
Can we start with collecting pain points with the current approach and then evaluate other options?

geojaz:
I'm at the very very early stages of looking at this right now, and I'm new to both kubeflow and TF so please bear with me.

Some quick notes on pain points-

1. Ideally, this should be really portable. Currently we have slightly diverging paths for minikube/GKE/RBAC vs legacy auth. ie:
    * Do we need to create a clusteradmin role? 
    * How we actually connect to our notebook
2. We probably shouldn't force ML experts to learn to kubectl (unless they want to!) which means it would be wonderful to have a tool that manages the yaml for us.
3. Likewise, the TF config manifests can probably be managed in a better way. 

That said- in less than an hour, I have a little kubernetes cluster running ML experiments and that kicks ass. We can only go up from here. My 30 second inkling is that helm could be very helpful here and we wouldn't have to rewrite the wheel.

jlewi:
One paint point is specifying volume mounts to be attached to Juptyer notebooks. A common example would be attaching a volume corresponding to a shared filesystem. 

Example:
- A team needs a POSIX compliant filesystem because they want to use HDF5 libraries
- They want a shared location to store datasets 
- They want the filesystem to be auto-mounted into their Juptyer containers.

Right now you can do this by modifying [kubeflow/components/jupyterhub/manifests/config.yaml](https://github.com/google/kubeflow/blob/master/components/jupyterhub/manifests/config.yaml).

I think the problem with this approach is that it will cause problems if config.yaml changes and you want to pull in those changes while preserving whatever extra volumes you added.

See also:
jupyterhub/jupyterhub#1580

aronchick:
[...]

PR #36, opened 2017-12-14
Use ksonnet to configure/deploy Kubeflow

jlewi:
Use ksonnet to customize/deploy Kubeflow.

* Its intended to inform the discussion in #23 

* The current implementation supports JupyterHub, TfJob operator, TfServing

* The README.ksonnet.md provides instructions on how to use ksonnet.

* Provide a prototype for running the TfCnn benchmark.

* The directory organization is based on what ksonnet expects. Specifically the directory within the repo gets checked out into the repo. So we use the directory kubeflow so that pkgs will be installed in ".../vendor/kubeflow/<package>". 

* The core component has an option to specify a list of persistent disks to be exported and mounted via NFS.

/cc @aronchick @foxish


jlewi:
This is ready for review.

/cc @foxish  @vishh

README.ksonnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+Kubeflow is currently evaluating different strategies for allowing deployments to be customized ([google/kubeflow/issues/23](https://github.com/google/kubeflow/issues/23)).
+
+This directory contains a prototype for using ksonnet to help us evaluate ksonnet.
+
+If you are unfamiliar with ksonnet you may want to start by reading the [tutorial](https://ksonnet.io/docs/tutorial)
+
+## Build ks
+
+ksonnet doesn't support adding non default registries yet ((ksonnet/ksonnet/issues/38)[https://github.com/ksonnet/ksonnet/issues/38]) so we need to modify and [build from source](https://ksonnet.io/docs/build-install).
```
paulbaumgart: GitHub isn't rendering this link correctly. I believe it should be: `[ksonnet/ksonnet/issues/38](https://github.com/ksonnet/ksonnet/issues/38)`

README.ksonnet.md
```diff
@@ -0,0 +1,183 @@
...
+
+```
+# Clone the ksonnet repo into your GOPATH
+go get github.com/ksonnet/ksonnet
+```
+
+Checkout PR 228 which adds support for registry add
```
paulbaumgart: This has been merged: https://github.com/ksonnet/ksonnet/commit/29e3cd455beea7557f4f81430e9c6ef635444335

jlewi:
Thanks @paulbaumgart

jlewi:
@foxish @vishh Do you want to review this before I merge it in?

paulbaumgart:
@jlewi The README.ksonnet.md updates look good. Might be worth taking another pass at https://github.com/jlewi/kubeflow/blob/ksonnet_crd/kubeflow/README.md (perhaps just linking to the README.ksonnet.md section?)

The doc-gen link in kubeflow/README.md also seems to be broken.

jlewi:
@paulbaumgart  Thanks. I also auto-generated READMEs for each component.

README.ksonnet.md
```diff
@@ -0,0 +1,213 @@
...
+
+This directory contains a prototype for using ksonnet to help us evaluate ksonnet.
+
+If you are unfamiliar with ksonnet you may want to start by reading the [tutorial](https://ksonnet.io/docs/tutorial)
+
+## Build ks
+
+You need a version of ksonnet newer than the [0.7.0 release](https://github.com/ksonnet/ksonnet/releases).
+
+As of this writing, Heptio hasn't released any newer prebuilt binaries so you will need to [build from source](https://ksonnet.io/docs/build-install).
```
foxish: Will there be a new release soon? This seems like a bit of friction until then.

README.ksonnet.md
```diff
@@ -0,0 +1,213 @@
...
+ks registry add kubeflow github.com/google/kubeflow/tree/master/kubeflow
+ks pkg install kubeflow/core
+ks pkg install kubeflow/tf-serving
+ks pkg install kubeflow/tf-job
+```
+
+
+Create the Kubeflow core component. The core component includes 
+  * JupyterHub
+  * TensorFlow job controller
```
foxish: Rename `core` to `training`?

README.ksonnet.md
```diff
@@ -0,0 +1,213 @@
...
+To configure your job you need to set a bunch of parameters. To see a list of parameters run
+
+```
+ks prototype describe tf-job
+```
+
+Parameters can be set using `ks param` e.g. to set the Docker image used
+
+```
+ks param set ${JOB_NAME} image ${IMAGE}
```
foxish: Can we instead recommend the declarative method here? I think people would want to version their config as a reproducible artifact, and specifying it this way doesn't allow doing that.

README.ksonnet.md
```diff
@@ -0,0 +1,213 @@
...
+ks apply nocloud -c kubeflow-core
+```
+
+If the user is running on a Cloud they could create an environment for this.
+
+```
+ks env add cloud
+ks param set --cloud=gke
+```
+   * The cloud parameter triggers a set of curated cloud configs.
```
foxish: Describe what parameters this sets?

hack/convert_manifest_to_jsonnet.py
```diff
@@ -0,0 +1,29 @@
+"""A simple utility to convert a manifest to corresponding jsonnet."""
```
foxish: Parts of this script seem brittle, what's the purpose of it?

README.ksonnet.md
foxish: Naming something `core` indicates that it's a dependency for other subsequent steps, which is not strictly true.

kubeflow/core/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+The library files for core define a set of relevant *parts* (_e.g._, deployments, services, secrets, and so on) that can be combined to configure core for a wide variety of scenarios. For example, a database like Redis may need a secret to hold the user password, or it may have no password if it's acting as a cache.
+
+This library provides a set of pre-fabricated "flavors" (or "distributions") of core, each of which is configured for a different use case. These are captured as ksonnet *prototypes*, which allow users to interactively customize these distributions for their specific needs.
+
+These prototypes, as well as how to use them, are enumerated below.
+
+### io.ksonnet.pkg.kubeflow-core
+
+Kubeflow core components
```
foxish: Same comment as above about naming parts of it `core`

kubeflow/core/nfs.libsonnet
```diff
@@ -0,0 +1,271 @@
...
+		  },
+		}, // provisioner
+    },
+
+    serviceAccount: {
+      "apiVersion": "v1", 
+      "kind": "ServiceAccount", 
+      "metadata": {
+        "labels": {
+          "app": name + "nfs-provisioner"
```
foxish: Should the nfs-provisioner be its own component? Not sure if it is here already. I see it as  - if it launches a new (optional) component in the user's cluster, it should be its own component.

kubeflow/core/parts.yaml
```diff
@@ -0,0 +1,36 @@
...
+      "type": "git",
+      "url": "https://github.com/google/kubeflow"
+   },
+   "bugs": {
+      "url": "https://github.com/google/kubeflow/issues"
+   },
+   "keywords": [
+      "kubeflow",
+      "tensorflow",
+      "database"
```
foxish: database?

README.ksonnet.md
jlewi: JupyterHub isn't training specific and the CRD is needed for subsequent steps. The idea here is that we have certain core components that you probably only deploy once. Whereas for tf-serving and tf-job you probably have multiple ones.

I'm open to other names but training doesn't seem right.

README.ksonnet.md
jlewi: Right now nothing, but it should control things like the CRD config for different clusters.

README.ksonnet.md
jlewi: It is versioned. ks param set modifies params.jsonnet which is checked into code control.

hack/convert_manifest_to_jsonnet.py
jlewi: It was a hack hence its location in "hacks" for converting our existing yaml manifests to jsonnet.
In a subsequent PR when we clean up the old stuff and move to ksonnet we can delete it.

README.ksonnet.md
jlewi: Hopefully, heptio might push a release this week but no guarantees.

kubeflow/core/nfs.libsonnet
jlewi: Ideally and long term yes. I think the ideal way to do this would be to have distinct  packages for each thing; e.g a package for JupyterHub, a package for Spark, a package for the CRD etc...

We could then have one or more packages with one or more prototypes defining useful combinations of this components. Its not immediately clear what the organizing principle for those higher level packages should be.

In the short term though, its more convenient for users if they have fewer packages to install. So that was the motivation for putting them all in the same package.

Using more packages should be reasonable when ksonnet adds better support for managing package dependencies ; e.g. so that core can depend on some other packages and installing core automatically pulls in the other packages.

jlewi:
@foxish this should be ready for another look.

foxish:
Please squash before merging. Also - can we have an issue tracking the fixes that need to be done afterwards (like removing the hack/ directory, separating out NFS provisioner, etc)

jlewi:
I had to force the merge because the CLA check wouldn't run.

PR #40, opened 2017-12-19
Add a note about TF serving Deployment not working with Minikube's kvm driver

dimpavloff:
Hi,

First of all, great project, looking forward to it developing further!

I've noticed that the model-server pods are crashing in Minikube with:
```
Illegal instruction (core dumped)
```
(that's the full output). That error message got me to https://github.com/mozilla/DeepSpeech/issues/912 which suggested the CPU's virtualisation might be the problem. And indeed, changing my Minikube's driver from `kvm2` to `virtualbox` fixed it. 

It would be nice for this to work with KVM which in my experience has performed better than virtualbox when under load. In the meantime, here's a PR adding a note about it not working with KVM. It's possible the issue is just on my machine (Intel i7-6700 CPU) so feel free to reject the PR if it's actually working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,8 +27,12 @@ This documentation assumes you have a Kubernetes cluster already available. For 
 
 [Minikube](https://github.com/kubernetes/minikube) is a tool that makes it easy to run Kubernetes locally. Minikube runs a
 single-node Kubernetes cluster inside a VM on your laptop for users looking to try out Kubernetes or develop with it day-to-day.
+
 The below steps apply to a minikube cluster - the latest version as of writing this documentation is 0.23.0. You must also have
 kubectl configured to access minikube.
+
+Finally, the Virtualbox/VMware drivers for Minikube are recommended as there is a known
+issue between the KVM/KVM2 driver and the TensorFlow Serving deployment, tracked in [kubernetes/minikube#2377](https://github.com/kubernetes/minikube/issues/2377).
 
 ### Google Kubernetes Engine
 


foxish:
Thanks for reporting - can you file an issue for this under the minikube repo?
This seems like something that should work.

cc/ @kow3ns @r2d4

README.md
```diff
@@ -28,7 +28,8 @@ This documentation assumes you have a Kubernetes cluster already available. For
 [Minikube](https://github.com/kubernetes/minikube) is a tool that makes it easy to run Kubernetes locally. Minikube runs a
 single-node Kubernetes cluster inside a VM on your laptop for users looking to try out Kubernetes or develop with it day-to-day. 
 The below steps apply to a minikube cluster - the latest version as of writing this documentation is 0.23.0. You must also have 
-kubectl configured to access minikube.
+kubectl configured to access minikube. Please note that the Tensorflow Serving Kubernetes deployment does not currently work with
```
foxish: I'd just make this a separate point for now. "If you're using minikube, the Virtualbox/VMware drivers are recommended. There are known issues with the KVM/KVM2 driver and TensorFlow serving tracked in <issue>".

dimpavloff:
Sounds good, I'll try and gather some more info using gdb before opening the minikube issue. I'll do that in 10 days when I'm back from my holiday.

dimpavloff:
I've just created https://github.com/kubernetes/minikube/issues/2377 describing the issue and with some more info. If you don't mind, I'll wait to get a response from them before updating this PR appropriately.

jlewi:
My suggestion is to just update the README with a note about a possible issue with minikube and then link to the issue you just filed.

dimpavloff:
Fair enough, I've updated the Readme as ~suggested, PTAL :)

README.md
```diff
@@ -27,8 +27,10 @@ This documentation assumes you have a Kubernetes cluster already available. For
 
 [Minikube](https://github.com/kubernetes/minikube) is a tool that makes it easy to run Kubernetes locally. Minikube runs a
 single-node Kubernetes cluster inside a VM on your laptop for users looking to try out Kubernetes or develop with it day-to-day.
+
 The below steps apply to a minikube cluster - the latest version as of writing this documentation is 0.23.0. You must also have
-kubectl configured to access minikube.
+kubectl configured to access minikube. Finally, the Virtualbox/VMware drivers for Minikube are recommended as there is a known
```
jlewi: nit: Can you make "Finally, the Virtualbox...." a new paragraph?

dimpavloff: Done. I had it together with the minikube version and config because they all felt like requirements

jlewi:
Thanks!

PR #47, opened 2017-12-20
Add proposal link

ddysher:
Per discussion [here](https://github.com/google/kubeflow/issues/33), we are adding the proposal link to README to make it more discoverable.

- first commit remove useless trailing whitespaces
- second commit add the proposal link

@jlewi  @DjangoPeng  PTAL


DjangoPeng:
Good job. It's more convenient for contributors to discuss the design and scope of Kubeflow.

PR #49, opened 2017-12-20
Fix specifying of GPU resources.

### LINKED ISSUE #48: tf-cnn prototype doesn't add GPU resource requests
If we configure a tf-cnn component to use GPUs, we don't actually end up specifying GPUs in the resource request.

e.g.

```
ks param set --env=gke cnn num_gpus 1
ks param set --env=gke cnn num_workers 1

ks apply gke -c cnn
```
Produces pod spec

```
spec:
  containers:
  - args:
    - python
    - tf_cnn_benchmarks.py
    - --batch_size=32
    - --model=resnet50
    - --variable_update=parameter_server
    - --flush_stdout=true
    - --num_gpus=1
    env:
    - name: TF_CONFIG
      value: '{"cluster":{"master":["cnn-master-nvsh-0:2222"],"ps":["cnn-ps-nvsh-0:2222"],"worker":["cnn-worker-nvsh-0:2222"]},"task":{"type":"worker","index":0},"environment":"cloud"}'
    image: gcr.io/kubeflow/tf-benchmarks-gpu:v20171202-bdab599-dirty-284af3
    imagePullPolicy: IfNotPresent
    name: tensorflow
    resources:
      requests:
        cpu: 100m
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: default-token-mnnrt
      readOnly: true
    workingDir: /opt/tf-benchmarks/scripts/tf_cnn_benchmarks
  dnsPolicy: ClusterFirst
```
So the image is correctly set to use the GPU but we are missing GPUs in the resources; e.g. it should look like

```
resources:
   limits:
       nvidia.com/gpu: 1
```

jlewi:
* Resources need to be attached to the container; we were attaching them to the pod and as a result they were getting stripped out.

* Fixes #48


jlewi:
@foxish PTAL

kubeflow/tf-job/tf-job.libsonnet
```diff
@@ -3,29 +3,30 @@ local k = import 'k.libsonnet';
 {
   parts:: {    
     tfJobReplica(replicaType, number, args, image, numGpus=0):: 
+      local baseContainer = {
+                  "image": image, 
+                  "name": "tensorflow",                 
+                };
+      local containerArgs =  if std.length(args) >0 then
```
foxish: Style nit: `> 0`

foxish:
Just that one nit. LGTM.

PR #53, opened 2017-12-20
Update Jupyter docs to reflect the move to ksonnet

jlewi:
* Update the docs to provide instructions using ksonnet
* Remove the old YAML manifests since we now use ksonnet


foxish:
Do we want to remove the old YAML config and docs entirely? Is it because of the additional effort that might be needed to keep the ksonnet config and YAML files in sync?

foxish:
cc @kow3ns

jlewi:
I think we want to remove them because it creates confusion as well as a maintenance headache.

If people just want to see the manifests they can use `ks show` to generate them.

foxish:
Ok. Maybe we can add the `ks show` command to the docs then.

components/jupyterhub/README.md
```diff
@@ -42,7 +32,14 @@ The above will expose JupyterHub on http://localhost:8000. The pod name can be o
...
 
-Configuration for JupyterHub is shipped separately and contained within the configmap under `manifests/config.yaml`. It is a Python file that is consumed by JupyterHub on starting up. The supplied configuration has reasonable defaults for the requisite fields and **no authenticator** configured by default.
+Configuration for JupyterHub is shipped separately and contained within the configmap defined by the [core componenent](https://github.com/google/kubeflow/tree/master/kubeflow). It is a Python file that is consumed by JupyterHub on starting up. The supplied configuration has reasonable defaults for the requisite fields and **no authenticator** configured by default. Furthermore, we provide a number of parameters that can be used to configure
+the core component. To see a list of ksonnet parameters run
+
+```
+ks prototype describe kubeflow-core
+```
+
+If the provided parameters don't provide the flexibility you need, you can take advantage of ksonnet to customize the core component and use a config file fully specified by you.
```
foxish: Maybe specify how to get the manifest here?

jlewi: Filed #56

jlewi:
I added instructions on how to use ks show to the user guide so people can inspect the manifest.

Since this is a trivial changes I'm going to use my super powers to merge it.

PR #59, opened 2017-12-21
Update readme

jlewi:
* Add requirements section
* Add link to Katacode scenario
* Add back link to proposal which was accidentally removed.


jlewi:
@aronchick PTAL

jlewi:
@aronchick  LGTM'd via mail.

Issue #15, opened 2017-12-07
python model server


BrentDorsey:
@aronchick / @jlewi - I enjoyed the BoF: Machine Learning on Kubernetes talk this evening, thanks for organizing it!

It would be great if kubeflow could include a Python ML model container with support for:

- conda environment.yml requirements management
     - I use conda to build production ML containers that support multiple model types including
          - scikit-learn
          - theano
          - xgboost gradient boosted trees
- gRPC and REST Prediction API's with Swagger documentation
- horizontal pod autoscaling
- prometheus application and system metrics

yuvipanda:
For building user images, we (at Project Jupyter) have built http://github.com/jupyter/repo2docker. https://github.com/jupyterhub/binderhub (deployed at mybinder.org) is a JupyterHub service that lets you dynamically build & run images based on a git repo (heroku style!). Perhaps that'll be cool to have?

jlewi:
@BrentDorsey We definitely need containers suitable for serving a variety of models. I think a lot of companies have already developed this piece.  For example seldon has defined a [prediction API](http://docs.seldon.io/api-prediction.html) with implementations for various model types.

So hopefully we can figure out how to leverage one or more of the existing solutions rather than developing a new one.

@yuvipanda I'd love to understand binderhub more. One of the problems we have right now is if you're developing on cluster (e.g. via Juptyerlab thank you for that) and writing a program (e.g. TF) we need to build a container from within the cluster. Right now I don't have a good solution for building docker images from within the cluster. It sounds like binderhub might solve this problem.

/cc @cliveseldon

BrentDorsey:
@jlewi - I like [Pipeline.AI's prediction API](https://github.com/PipelineAI/pipeline), of course I'm biased as I'm a Pipeline.AI [committer](http://pipeline.ai/products/).

@yuvipanda - I'd also love to understand more about binderhub.

jlewi:
@BrentDorsey Do you have a pointer to AI's prediction server? I had trouble finding it in the repo.

BrentDorsey:
@Jeremy Lewi<mailto:jlewi@google.com> – Let me talk with Chris and get back to you, he’s moved some stuff around.

cliveseldon:
Hi,

 I work for Seldon, referenced above by @jlewi

 We have pushed our seldon-core product to github https://github.com/SeldonIO/seldon-core

[...]

PR #65, opened 2017-12-23
python prediction server

lluunn:
An implementation of Google Cloud Machine Learning Engine (CMLE) [Prediction API](https://cloud.google.com/ml-engine/docs/v1/predict-request).

The PR contains the Dockerfile that can run a python prediction server that provides the prediction service for a trained ML model. It supports tensorflow, scikit-learn, and xgboost. See [README](https://github.com/lluunn/kubeflow/blob/pred_server/components/prediction_server/README.md).


lluunn:
related to this [issue](https://github.com/google/kubeflow/issues/15)

jlewi:
Can you update the comment to explain what this PR is about?

What functionality is this Python Server provide? What does the API look like? Which frameworks does it support?

jlewi:
Also you might want to take a look at #15 and chime in about how this PR relates to that issue.

lluunn:
Updated the description, and added an OpenAPI spec. PTAL, thanks!

jlewi:
Thanks can you explain the architecture? 

For the model server? 
Does the server support HTTP and Rest? Is their common request handling infrastructure?
What is the interface between the model server and specific model types/framework?

What server infrastructure does the model use?

The Seldon API also provides a one off that allows sending data with more structure

```
message DefaultData {
  repeated string names = 1;
  oneof data_oneof {
    Tensor tensor = 2;
    google.protobuf.ListValue ndarray = 3;
  }
}

message Tensor {
  repeated int32 shape = 1 [packed=true];
  repeated double values = 2 [packed=true];
}
```

lluunn:
It's a python HTTP server (no gRPC support yet), using [paste](https://pypi.python.org/pypi/Paste)

For either TF, scikit, or xgboost, the request should have the same format as {"instances": [instance1, instance2, ...]} as described in [spec](https://gist.github.com/lluunn/22fe1f2eaf0859f7aee8c54a1aeb3275), and handled by the handler [here](https://github.com/lluunn/kubeflow/blob/pred_server/components/prediction_server/server/prediction_server_lib.py#L302).

In the case of TF, the TF ModelServer is running as a [subprocess](https://github.com/lluunn/kubeflow/blob/pred_server/components/prediction_server/server/frameworks/tf_prediction_server_lib.py#L151). We may want to make ModelServer to run as its own container later.

jlewi:
I don't think running the TF ModelServer as a subprocess is the right pattern for K8s. The more common pattern for K8s would be to have 2 containers in the pod one for the Python HTTP server and one for TF Serving.

So in the case of TF, is the primary purpose of this server to proxy HTTP to gRPC? Does it do anything else like transform the request to ExampleProto?

jlewi:
How does this compare to the Seldon [model server](https://github.com/SeldonIO/seldon-core/tree/master/wrappers/python)?

jlewi:
Would it make sense to turn this into a CMLE to TF Proxy? So a simple binary that converts the incoming request to a request that can be sent to TF serving? With the expectation that this would run as a sidecar in a pod running TF Serving.

yupbank:
hey... how is the status of this? i am interested in the feature.. 
if no one is picking up, i can help a bit..

lluunn:
Scope it down to a proxy from CMLE to TF serving sounds good. I will split the PR and start with that.

jlewi:
#183 was just submitted. Do we still need this PR?

lluunn:
I will close it for now. We will add more functionalities to the http server in another PR.

Issue #38, opened 2017-12-18
E2E Testing For Kubeflow.


jlewi:
We need to setup continuous E2E testing for google/kubeflow

- [ ] Create a basic an E2E test
     - [ ] Deploy and verify JupyterHub is working
     - [X] Deploy and verify TfJob is working
     - [ ] Verify TfServing is working
- [X] Setup prow for google/kubeflow
     - [X] Setup presubmit tests
     - [ ] Setup postsubmit tests

I think it makes sense to write the E2E test first and then once we have it we can setup prow integration.

foxish:
Related: https://github.com/kubernetes/test-infra/issues/6011

jlewi:
Track progress following these [instructions](https://github.com/tensorflow/k8s/tree/master/test-infra#notes-adding-tensorflowk8s-to-k8s-prow-instance) to setup prow.

- [ ] Step 1; PR to add google/kubeflow to prow configs; kubernetes/test-infra#6110 is out for review
- [x] Step 2; Setup GitHub prow webhook
- [ ]  Step 3; Add k8s robot to google/kubeflow
- [ ]  Step 4: Setup PR dashboard
- [ ] Step 5: Grant read access to Gubernator GCS bucket to kubeflow-testing service account

jlewi:
presubmits are passing
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-presubmit

postsubmits don't seem to be running
I think this is the same issue as tensorflow/k8s#370 i.e we aren't truncating the commit sha when constructing the Argo workflow name.

To fix this we should probably modify kubeflow/kubeflow to pull run_workflow_e2e.py from kubeflow/testing once kubeflow/testing#9 is submitted. That pr should fix this issue.

PR #71, opened 2017-12-28
A python script to test deploying Kubeflow

jlewi:
* test_deploy.py is a python script that 
    * Creates a ksonnet app
    * Deploys Kubeflow
    * Does some basic checks to see if the Kubeflow deployment is healthy

* test_deploy.py will be used to create continuous E2E tests for Kubeflow.

* Log output and test results are written to a local directory

* In a follow on PR we will add a binary to copy all the outputs from the
  local directory to Gubernator. This way most steps in the test don't
  need to deal with GCS.

* First piece of #38 (E2E test for Kubeflow)

* Add a pylintrc file


jlewi:
@foxish PTAL

.pylintrc
```diff
@@ -0,0 +1,399 @@
+[MASTER]
```
foxish: Is this a standard .pylintrc file?

testing/create_cluster.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# Create the GKE cluster used for testing Kubeflow deployments.
+# This should be a one-time thing because we reuse the same cluster for all tests.
+gcloud --project=mlkube-testing container clusters create \
+	--zone=us-east1-d  \
+	--machine-type=n1-standard-8 \
+	--cluster-version=1.8.4-gke.1 \
```
foxish: Pinning the cluster version? we should ideally parse out the `get-server-config` output and pick the latest that way. See https://github.com/kubernetes/test-infra/blob/master/kubetest/extract_k8s.go#L398-L409

testing/test_deploy.py
```diff
@@ -0,0 +1,218 @@
...
+    # Verify that the TfJob operator is actually deployed.
+    tf_job_deployment_name = "tf-job-operator"
+    logging.info("Verifying TfJob controller started.")
+    util.wait_for_deployment(api_client, namespace.metadata.name,
+                             tf_job_deployment_name)
+
+    # Verify that JupyterHub is actually deployed.
+    jupyter_name = "tf-hub"
+    logging.info("Verifying TfHub started.")
+    util.wait_for_statefulset(api_client, namespace.metadata.name, jupyter_name)
```
foxish: Do we want this to create the loadbalancer service as well (I think that was the old default behavior)? If that's still the case, it might be better to turn that off, or create a ClusterIP type of service.

testing/test_deploy.py
```diff
@@ -0,0 +1,218 @@
...
+    "--zone",
+    default="us-east1-d",
+    type=str,
+    help="The zone for the cluster.")
+
+  parser.add_argument(
+    "--github_token",
+    default=None,
+    type=str,
+    help=("The GitHub API token to use. This is needed since ksonnet uses the "
```
foxish: Link to https://github.com/ksonnet/ksonnet/blob/master/docs/troubleshooting.md? It was unclear at first why this script needed the github token.

.pylintrc
jlewi: This is the .pylintrc from tensorflow/k8s. I believe that was a standard .pylintrc that was then lightly modified for our specific use case.

testing/create_cluster.sh
jlewi: I removed this file. #72 replaced this script with more complete instructions in the README for setting up the test infrastructure.

The cluster setup is onetime manual. The only reason we pin to a specific version is because I don't know if the default is 1.8 yet. If it is then pinning is unnecessary. Either way, given its a one time manual step there's not much incentive to automating parsing of get-server-config.

testing/test_deploy.py
jlewi: I think the correct thing for the test to do is to deploy Kubeflow with the default options. I think the question about what are the right defaults for Jupyter is valid but orthogonal. I opened #80.

testing/test_deploy.py
jlewi: Done.

jlewi:
@foxish Thank you this is ready for another look.

foxish:
The code itself LGTM, but I'm having some trouble understanding how this fits in with the argo based workflow you detailed in the readme. Is the test_deploy script meant to grow to encompass more tests? or simply be a standalone simple test script that will be run as one stage within argo?
If the script is a starting point for more tests, would it make sense to use some python testing package to make it nicer to read as more tests are added?

jlewi:
So I expect eventually our E2E test will be a workflow that might look something like follows

Deploy Kubeflow Core (JupyterHub, TF Serving, TfJob CRD,...)
Run a bunch of tests in parallel
     - Verify CPU notebook works
     - Verify GPU notebook works
     - Verify CPU TfJOb works
     - Verify GPU job works
     - ....
TearDown

So the question is how to structure the individual tests. The pattern I've chosen is to make each test correspond to an entrypoint in one or more python programs. So the commands we might have for the above are

python -m testing.test_deploy deploy

python -m testing,test_jupyter 
...
python -m testing,test_tfjob run_job 
...
python -m testing.test_deploy teardown

There's a lot we could do to improve this.

I think the question you are asking is why not use a package like unittest to structure this a series of test classes with multiple test cases. Perhaps something like

```
class TfJobTest(unittest.TestCase):
   def testCpu(self):
       ....
    def testGpu(self):
       ....
``` 

I think the short answer was that `unittest` wasn't buying us much because it doesn't match prow. So I could imagine something like the following

```
class TfJobTest(prow.ProwTestCase):
    def testCpu(self):
         ...
    def testGpu(self):
        ...
```
where ProwTestCase takes care of outputting information like the junit files. Its not obvious to me that there's an existing library that would help here but I could be wrong. I'll file an issue.

PR #86, opened 2018-01-05
Add a section about who should consider using/contributing to Kubeflow.

jlewi:


jlewi:
@aronchick @nkashy1 PTAL

elmiko:
this is a nice addition, +1

README.md
```diff
@@ -19,6 +19,25 @@ Because ML practitioners use so many different types of tools, it is a key goal
 
 Ultimately, we want to have a set of simple manifests that give you an easy to use ML stack _anywhere_ Kubernetes is already running and can self configure based on the cluster it deploys into.
 
+
+## Who should consider using Kubeflow?
```
puneith: Shouldn't this be more generic that this? This sounds like its only restricted to these 3 cases but its not, correct?

README.md
```diff
@@ -19,6 +19,25 @@ Because ML practitioners use so many different types of tools, it is a key goal
...
+
+Based on the current functionality you should consider using Kubeflow if
+
+  * You want to train/serve TensorFlow models in different environments (e.g. local, on prem, and cloud)
+  * You want to use Jupyter notebooks to manage TensorFlow training jobs
+       * kubeflow is particularly helpful if you want to launch training jobs that use more resources (more nodes or more GPUs) then your notebook.
+  * You want to combine TensorFlow with other processes
+       * For example if you want to use [tensorflow/agents](https://github.com/tensorflow/agents) to run simulations to generate data for training
+         reinforcement learning models
+
```
aronchick: "We should note, the above list covers ONLY currently available functionality. We hope to expand 'Who should consider using Kubeflow' very soon! (See the next section)"

aronchick:
LGTM with one addition - +1 on great add.

README.md
elmiko: imo, the contributing section covers that nicely. but i agree it's always good to be clear about the intentions. i think the suggestion by @aronchick below is a nice nod in that direction.

README.md
jlewi: PTAL

README.md
jlewi: I wanted the list to reflect current capabilities where Kubeflow is a clear value add over competing solutions. Per David's comment I added a note below that this is based on current capabilities which we are working on expanding.

README.md
nkashy1: It would be good to have a list of functionality that we would like it to have. For example:

+ Run Kaggle kernels at scale

+ Deploy secure web services from Jupyter using Cloud Endpoints

+ etc.

This would serve as a pool of ideas that contributors could work on.

In general, though, this is a big step in the right direction.

nkashy1: "Functionality wish list"

jlewi: I think the wish list is GitHub issues.

nkashy1: Good point :)

PR #92, opened 2018-01-06
Client script for inception model server

nkashy1:
Added a script that allows users to run the hosted inception model on
images on their local filesystems or on Google Cloud Storage.

This is, with only very slight modifications for readability, the same
as the client provided by TensorFlow Serving -
https://github.com/tensorflow/serving/blob/master/tensorflow_serving/example/inception_client.py

As such, I am completely okay with us just linking to their script.

My initial intention was to make this as a notebook, but the problem is
that the tensorflow-serving-api Python package is only available for
Python 2 and the kubeflow-core environment only offers a Python 3
backend for the Jupyter notebook.

This is therefore a stopgap until I can introduce an appropriate image
in place of the one used by kubeflow-core.


nkashy1:
@jlewi : Please take a look. If you approve of this method of invoking the inception client, I will also add relevant content to the README and then we can merge.

Long-term, I want to make a notebook that users can deploy to the Jupyterhub pod that runs the same code.

jlewi:
Does our Docker image contain all the dependencies needed for this script?
Having a solution and notebook for generating predictions would be great. The code in this PR looks fine to me.

nkashy1:
We are missing the `tensorflow-serving-api` Python package in the TensorFlow Serving image (but it doesn't really belong there) and likely also in the Jupyterhub image.

nkashy1:
@jlewi : This is ready to go. Made updates to README and also added capability of running from Docker container. Please take a look.

components/k8s-model-server/README.md
```diff
@@ -162,3 +162,107 @@ You can learn more about [updating a Deployment](https://kubernetes.io/docs/conc
...
+cat sleeping on a comforter atop a sofa:
+
+![Cat on comforter on sofa](./inception-client/sleeping-pepper.jpg)
+
+You can also use to to label your own images.
+
+#### Running the script directly
+
+You can run the script directly in your local environment if Python2 is available to you. You will not be able to use the script with Python3
+as the [`tensorflow-serving-api` package](https://pypi.python.org/pypi/tensorflow-serving-api) is not yet Python3-capable.
```
jlewi: Can you link to a GitHub issue here for Python 3 support for the client model? If not please open one.

components/k8s-model-server/README.md
```diff
@@ -162,3 +162,107 @@ You can learn more about [updating a Deployment](https://kubernetes.io/docs/conc
...
+
+The [inception-client](./inception-client) directory contains a Python script you can use to make a call against the deployed model.
+
+This script is intended to be run externally to the kubernetes cluster as a demonstration that the inception model is correctly being served.
+You can run the script either directly from a Python2 environment or in a Docker container.
+
+#### Setup
+
+You will require the external IP for the inception service as well as the port it is being hosted on. The inception service should be
+listed under the value you used for the `MODEL_NAME` parameter in the ksonnet component. You can find this information using
```
jlewi: What are we telling people to do in terms of accessing the service from outside the cluster? Are we telling them to expose the service publicly?

If we're exposing the service publicly lets make this very explicit and lets open up follow on issues to think about how we can point users in the direction of securing it using appropriate credentials.

I think we should try to avoid telling people to expose services publicly. So I lean towards suggesting solutions that rely on `kubectl proxy` or `kubectl port-forward`. In an advanced section we can point people in the direction of making this more accessible using IAP or other methods to make the service accessible from outside the cluster.

Another way to side step the issue of external access to your K8s cluster is to have people send the predictions from with the cluster using `kubectl run` or by running from a Jupyter notebook running inside the cluster.

components/k8s-model-server/README.md
```diff
@@ -162,3 +162,107 @@ You can learn more about [updating a Deployment](https://kubernetes.io/docs/conc
...
+```commandline
+kubectl get services
+NAME         TYPE           CLUSTER-IP      EXTERNAL-IP      PORT(S)			 AGE
+$MODEL_NAME  LoadBalancer   <INTERNAL IP>   <SERVICE IP>     <CLUSTER PORT>:<NODE PORT>  <TIME SINCE DEPLOYMENT>
+```
+
+We will feed the `<SERVICE IP>` and `<CLUSTER PORT>` to the labelling script. We will use it to label the following image of a
```
jlewi: What is the cluster port? Where is set?

components/k8s-model-server/README.md
```diff
@@ -162,3 +162,107 @@ You can learn more about [updating a Deployment](https://kubernetes.io/docs/conc
...
+```commandline
+python label.py -s <SERVICE IP> -p <CLUSTER PORT> sleeping-pepper.jpg
+```
+
+#### Run in Docker container
+
+The [inception-client](./inception-client) directory also contains a [Dockerfile](./inception-client/Dockerfile) that will allow you to
+call out to the inception service from a container.
+
+From that directory, start by building the image:
```
jlewi: Lets open an issue to provide a prebuilt docker container.

components/k8s-model-server/README.md
nkashy1: Currently, it looks like the tf serving component already exposes the service - when I run `kubectl get svc`, I see an external IP and I don't require any additional authentication against it.

For the purposes of the component guide, I do want to show users how to make calls to a kubeflow service from an external client. I agree with keeping things as they are and add a follow-on issue for authentication, as you suggested.

components/k8s-model-server/README.md
nkashy1: Maybe I am using incorrect terminology - I meant the host port on the kubernetes cluster.

It is set here: https://github.com/google/kubeflow/blob/master/kubeflow/tf-serving/tf-serving.libsonnet#L27

components/k8s-model-server/README.md
nkashy1: Done

components/k8s-model-server/README.md
jlewi: I opened up #121 to change the default. I'd much prefer it if we didn't tell people to open up public IP's. I'd prefer it if our docs used kubectl and provided links or other instructions about how to add external IPs if desired.

jlewi:
Is this ready for another look?

nkashy1:
@jlewi : I think it's ready to merge - the only outstanding issue is that the service is exposed to the public on GKE but it makes sense to work on that issue separately from this PR.

components/k8s-model-server/inception-client/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM python:2.7.14
+
+RUN pip install grpcio tensorflow tensorflow-serving-api
+
+RUN mkdir -p /opt/label
+
+RUN mkdir -p /data
```
carmark: `VOLUME` should create the directory if the dir is not exist.  Or how about combining line 19 and 21 together?

components/k8s-model-server/inception-client/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM python:2.7.14
+
+RUN pip install grpcio tensorflow tensorflow-serving-api
```
carmark: add a option `--no-cache-dir` to reduce the image size?

nkashy1: Done

components/k8s-model-server/inception-client/Dockerfile
nkashy1: And done - got rid of the `VOLUME` bit

components/k8s-model-server/README.md
jlewi: I think the term would just be SERVICE port.

I would prefer the guide had an instruction like this
kubectl port-forward `kubectl get pods --selector=name=airflow -o jsonpath='{.items[*].metadata.name}'` 8180:8080

That showed people how to setup port forwarding and then people connect via the service.

Or better yet would be to run it from within cluster e.g via a Jupyter Notebook. From within the cluster they can the service name as a DNS name.

Or they could use 
kubectl run 
to start up a shell inside a Docker container running within the cluster in which case they can again use the ServiceName

components/k8s-model-server/README.md
nkashy1: My idea was to merge this first so there's some client code in the readme, and then fix the service issue. But if you prefer to fix the issue before merging, that's fine too.

Added checklist on https://github.com/google/kubeflow/issues/121, will work on it in the next couple of days.

components/k8s-model-server/README.md
nkashy1: Sounds good - this is related to https://github.com/google/kubeflow/issues/121 and I will update these instructions as part of that work.

components/k8s-model-server/README.md
jlewi: I don't think you need to fix #121 even with type LoadBalancer you can access the gRPC from inside the cluster just like you would if it was a ClusterIp type. So we can go ahead and just make the instructions describe access without using an external IP. This way we don't have to go back and revise the instructions in the future.

nkashy1: Good point, did as you said although getting images into kubernetes posed a bit of a dilemma - decided for a simple workflow in which they are provided with the image.

components/k8s-model-server/README.md
nkashy1: I went with `kuberentes run`

nkashy1: The reason I didn't do it from the jupyterhub is that it only offers a Python3 kernel and there isn't a clean way to add the tensorflow-serving-api package there. I could directly use the protos I suppose, but I decided to go with something that required less explanation.

nkashy1:
@jlewi : updated, please have a look

nkashy1:
@jlewi: Changed service type in `kubeflow/tf-serving/tf-serving.libsonnet` to "ClusterIP" and added instructions to `components/k8s-model-server/README.md` for changing service type (using `kubectl edit service`).

Out of curiosity, the `kubeflow/tf-serving/tf-serving.libsonnet` defines a `redis-data` volume - did that just sneak in through a `CTRL+V`?

components/k8s-model-server/README.md
```diff
@@ -163,3 +163,148 @@ You can learn more about [updating a Deployment](https://kubernetes.io/docs/conc
...
+
+The [inception-client](./inception-client) directory also contains a [Dockerfile](./inception-client/Dockerfile) that will allow you to
+call out to the inception service from a container. You can run this container on your local machine if you publicly exposed your
+`inception` service. If you would like to do this on GKE, simply run
+
+```commandline
+kubectl edit service inception
+```
+
+and change the service type to `NodePort` or `LoadBalancer`.
```
jlewi: I'd suggest removing NodePort option here.

If they use NodePort they'd have to make sure firewall rules don't block access to the nodes; they'd also have to get IP of their nodes.

PR #111, opened 2018-01-11
replace the model from tensorflow/models

yupbank:
addressing this https://github.com/google/kubeflow/issues/98

### DIFF
--- a/components/k8s-model-server/README.md
+++ b/components/k8s-model-server/README.md
@@ -74,7 +74,7 @@ gsutil mb gs://<bucket-name>
 ## Upload a Model
 You can train an inception model using the instructions from the 
 [TensorFlow documentation](https://www.tensorflow.org/tutorials/image_retraining), or you can download the pre-trained
-[example](inception.tar.gz).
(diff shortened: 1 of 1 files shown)


components/k8s-model-server/README.md
```diff
@@ -74,7 +74,7 @@ gsutil mb gs://<bucket-name>
 ## Upload a Model
 You can train an inception model using the instructions from the 
 [TensorFlow documentation](https://www.tensorflow.org/tutorials/image_retraining), or you can download the pre-trained
-[example](inception.tar.gz).
+[example](http://download.tensorflow.org/models/inception_v3_2016_08_28.tar.gz).
```
jlewi: I think gs://kubeflow-models/inception because you can serve directly from GCS which is what we showed in our blog post
http://blog.kubernetes.io/2017/12/introducing-kubeflow-composable.html

yupbank: i think i'll keep the download/upload example but add one more sentence explaining there is already a public gcs bucket available to use.

PR #115, opened 2018-01-11
update tf-cnn example use terminationPolicy to address not using master replica

yupbank:
* Related to #192 
* Now that terminationPolicy can be used to specify a non-master chief, we no longer need to add a master to run the tf-cnn we can just use worker 0 as the chief which allows the example to run out of the box.

### DIFF
--- a/tf-controller-examples/tf-cnn/launcher.py
+++ b/tf-controller-examples/tf-cnn/launcher.py
@@ -75,16 +75,6 @@ if __name__ == "__main__":
   command.append("--worker_hosts=" + worker_hosts)
   command.append("--task_index={0}".format(task_index))
 
-  # TODO(jlewi): This is a work around
-  # https://github.com/tensorflow/k8s/issues/192
-  # Since TfJob requires a master but the tf benchmark
-  # code doesn't know what to do with master we just run
-  # forever in the master.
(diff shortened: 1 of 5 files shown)


jlewi:
This is great; thank you so much for contributing?

/ok-to-test

Can you update our ksonnet prototype?
https://github.com/google/kubeflow/blob/master/kubeflow/tf-job/prototypes/tf-cnn-benchmarks.jsonnet

yupbank:
> Can you update our ksonnet prototype?

a second PR or just one PR for doing both?

jlewi:
@yupbank Whatever works for you. If you want to submit this first that works for me. Lets just leave the issue open until the ksonnet files are updated.

jlewi:
@yupbank If you want to submit this first please merge in latest changes and update the PR and let me know and I'll merge it.

yupbank:
let's merge this one first :)

thanks !

jlewi:
* Related to #192

* Now that terminationPolicy can be used to specify a non-master chief, we no longer need to add a master to run the tf-cnn we can just use worker 0 as the chief which allows the example to run out of the box.

jlewi:
I did not intend to make that last comment; that was supposed to be the commit message.

PR #119, opened 2018-01-12
change kjsonnet in tf-cnn to remove not used MASTER

yupbank:
* Related to https://github.com/google/kubeflow/issues/51
* Now that terminationPolicy can be used to specify a non-master chief, we no longer need to add a master to run the tf-cnn we can just use worker 0 as the chief which allows the example to run out of the box.
* Update the ksonnet to work with that


jlewi:
This is great thank you so much.

Issue #60, opened 2017-12-21
Support IAP on GKE


jlewi:
When Kubeflow is deployed on GKE it should be easy to configure it to setup IAP to secure remote access to Kubeflow services.
  
- [x] Ksonnet manifests for setting up ingress and other resources
- [x] Docs explaining how to enable IAP
- [x] Use Cloud Endpoints (or some other proxy) to validate JWT and reject traffic that bypassed IAP
    * Blocked by #104 - Can't connect to Jupyter Kernel using sidecar

Here are issues that need to be fixed to use Envoy as a proxy that handles JWT verification
- [x] istio/proxy#941 -Need support for ES256 algorithm which is what IAP uses
- [x] istio/proxy#939 - Need to support headers used by IAP
- [X] istio/proxy#929 - Envoy proxy needs to support rejecting all requests without valid JWT credentials
- [x] istio/proxy#930 - Need to support JWT validation without breaking GCP health checking
     - We can work around this just by having 2 envoy proxies in sequence 
     - both can be running in the same pod

jlewi:
So I have a prototype of enabling IAP https://github.com/jlewi/kubeflow/tree/iap. The one issue I'm hitting is that the proxy doing IAP validation doesn't seem to reject traffic that didn't go through IAP.

In this implementation a sidecar in the JupyterHub pod validates the incoming request and reject it if it didn't go through IAP or the JWT isn't valid.

One question I have is how can we configure JupyterHub to use the identity in the JWT but not verify it? It looks like there's a [JWTAuthenticator](https://github.com/mogthesprog/jwtauthenticator) but I don't actually want to authenticate it since the request was already authenticated in the side car.

Is there an easy way to tell JupyterHub to get identity from the JWT but not authenticate it?

/cc @yuvipanda @mogthesprog @foxish

mogthesprog:
I’m on mobile and out at the moment. But IIRC IAP sends some special headers X-Goog-User or something, so you could maybe use the Remote-User Authenticator for jupyter hub which takes the name of the user header as configuration? I can have another nose for the links when I get home. 

That should do what you need on then jupyterhub end I believe.

jlewi:
@mogthesprog Thanks for the quick response let me try that out.

jlewi:
That worked thank you very much. I just had to configure the JupyterHub config as follows

```
c.JupyterHub.authenticator_class ='jhub_remote_user_authenticator.remote_user_auth.RemoteUserAuthenticator'
[...]

Issue #11, opened 2017-12-05
Secure proxy 


jlewi:
Opening this issue to track addition of a secure proxy.

Kubeflow will consist of many applications running insecure web apps. Already the stack contains Juptyer notebooks and TensorBoard.

It would be nice if Kubeflow included a secure proxy so that a user could access Kubeflow components at URLs like 

https://my-kubeflow.my-domain.io/juptyerhub
https://my-kubeflow.my-domain.io/tf-jobs/my-job/tensorboard

Some issues to consider. 
  * Some of these apps are ephemeral. For example, 
           * TfJob spins up a TensorBoard instance whose lifetime is tied to the life time of the job. So we'd like to dynamically add routes.
  * We'd like to integrate nicely with different Clouds
           * For example, on GKE we'd like to use IAP 

I think we can use [Ambassador](https://www.getambassador.io/) to do this
   - Ambassador is a proxy based on Envoy
  -  Mappings can be dynamically added just by adding annotations to a K8s service

Here's a diagram illustrating what this would like

![kubeflow ingress](https://user-images.githubusercontent.com/777219/35186821-44228968-fdcf-11e7-99fe-f917ea3dd55c.png)

Using Ambassador everything sits behind a single ingress point. The ingress would be configured differently on different clouds but everything else should be the same.

On GKE we can secure the ingress using IAP
   * Hopefully Ambassador supports (or will support) Envoy's [JWT](https://github.com/istio/proxy/tree/master/src/envoy/auth) features so that we can use Ambassador to reject traffic that bypassed IAP

Some open questions
   * How do people securely connect to non-GKE clusters? 
          * Do people use VPN? Do we need an IAP equivalent?
   * Can we configure Ambassador to do JWT validation?
   * On non-GKE clusters I think we could use [oauth-proxy](https://github.com/bitly/oauth2_proxy) to replace IAP
          * This could authenticate the request using a variety of providers (Github, Google, Facebook) and provide a signed JWT. At which point the solution is largely the same as in the GKE case (i.e. use Envoy to validate JWT and reject traffic without a JWT)
          * i think we'd want to integrate oauth-proxy into our envoy proxy
                 * Ambassador already has support for external auth (but not oauth-proxy)
                 * Not sure about contour
     * How much can we get done for Kubecon May?

Alternatives
   * It looks like HeptIo has a similar project [contour](https://github.com/heptio/contour)
        * I think contour is designed as an ingress controller 

[...]

PR #120, opened 2018-01-12
Document flow improvements, expanded on Jupyter usage

elsonrodriguez:
Also added more info on the inception endpoint. Still need to provide an example with an inception client.

### DIFF
--- a/user_guide.md
+++ b/user_guide.md
@@ -1,21 +1,26 @@
-# Using Kubeflow 
+# Using Kubeflow
 
-If you are unfamiliar with ksonnet you may want to start by reading the [tutorial](https://ksonnet.io/docs/tutorial)
+This guide will walk you through the basics of deploying and interacting with Kubeflow. A basic understanding of Kubernetes, Tensorflow, and Ksonnet are useful in understanding the contents of this guide.
+
+* [Kubernetes](https://kubernetes.io/docs/tutorials/kubernetes-basics/)
+* [Tensorflow](https://www.tensorflow.org/get_started/)
+* [Ksonnet](https://ksonnet.io/docs/tutorial)
 
 ## Requirements
-
-  * ksonnet version [0.8.0](https://ksonnet.io/#get-started) or later.
-    * See [below](#why-kubeflow-uses-ksonnet) for an explanation of why we use ksonnet
-  * Kubernetes >= 1.8 [see here](https://github.com/tensorflow/k8s#requirements)
+ * Kubernetes >= 1.8 [see here](https://github.com/tensorflow/k8s#requirements)
+ * ksonnet version [0.8.0](https://ksonnet.io/#get-started) or later. (See [below](#why-kubeflow-uses-ksonnet) for an explanation of why we use ksonnet)
 
 ## Deploy Kubeflow
 
-Initialize a directory to contain your deployment
+We will be using Ksonnet to deploy kubeflow into your cluster.
+
+Initialize a directory to contain your ksonnet application.
+
 ```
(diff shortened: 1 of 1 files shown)


user_guide.md
```diff
@@ -95,21 +100,53 @@ http://xx.yy.zz.ww:31942
 
 For some cloud deployments, the LoadBalancer service may take up to five minutes display an external IP address. Re-executing `kubectl get svc` repeatedly will eventually show the external IP field populated.
 
-Once you have an external IP, you can proceed to visit that in your browser. The hub by default is configured to take any username/password combination. After entering the username and password, you can start a single-notebook server,
-request any resources (memory/CPU/GPU), and then proceed to perform single node training.
+Once you have an external IP, you can proceed to visit that in your browser. You should see a sign in prompt.
```
jlewi: Do we need to change this? I don't think we should be opening an external IP by default. So I think we should tell people how to connect via kubectl until we have a recipe for connecting securely via https (#11).

user_guide.md
```diff
@@ -95,21 +100,53 @@ http://xx.yy.zz.ww:31942
...
+  sess.run(train_step, feed_dict={x: batch_xs, y_: batch_ys})
+
+correct_prediction = tf.equal(tf.argmax(y,1), tf.argmax(y_,1))
+accuracy = tf.reduce_mean(tf.cast(correct_prediction, tf.float32))
+print(sess.run(accuracy, feed_dict={x: mnist.test.images, y_: mnist.test.labels}))
+```
+
+Paste the example into a new Python 3 Jupyter notebook and execute the code, this should result in a 0.9014 accuracy result against the test data.
+
+Please note that when running on most cloud providers, the public IP address will be exposed to the internet and is an
```
jlewi: As noted above can we change this and make the default deployment of JupyterHub use ClusterIP?

My hope is that various Cloud Providers will chime in and provide help making JupyterHub secure.

On GKE I think it makes sense to do this using IAP (#60).

One option to provide a Cloud Agnostic secure method for accessing JupyterHub would be to use JupyterHub's built in OAuth support. I think it has a GitHub authenticator. Should we open up an issue for that?

yuvipanda: Providing clean authentication is one of the goals of JupyterHub. There's a lot listed on https://github.com/jupyterhub/jupyterhub/wiki/Authenticators and more around the web.

user_guide.md
elsonrodriguez: The security culture for k8s examples is pretty loose, however the documentation already mentions that by default this is totally insecure.

I do feel this is a continuation of a bad precedent.

A good solution would involve getting kube-cert-manager and external-dns configured, along with adding whitelists to the Service objects via ksonnet based on the user's IP, however I'm focusing on just getting things working for now.

I can switch everything to use port-forward if needed.

elsonrodriguez:
Yeah, that inception client looks right! I'll give it a test within a day or two.

user_guide.md
jlewi: I'd prefer for us to use port-forward and not promote insecure methods.

elsonrodriguez: Mind if I tackle the port-forward in another PR? It'll take me a bit to mess with ksonnet and put a toggle in there for loadbalancer vs port-forward (for those who want to use LB still)

jlewi: I think that's fine; since the code is already using LoadBalancer by default and this PR is just documenting current behavior.

jlewi:
LGTM. Please sync and I'll merge this.

elsonrodriguez:
Rebased.

PR #152, opened 2018-01-26
Fix the instruction for setting up cluster to run test

lluunn:
1. Fix some text format in README
2. checkin the vendor for test-infra app.
3. set service_type for jupyterHubLoadBalancer in nfs-jupyter.jsonnet

### DIFF
--- a/testing/test-infra/app.yaml
+++ b/testing/test-infra/app.yaml
@@ -1,12 +1,6 @@
 apiVersion: 0.0.1
 kind: ksonnet.io/app
 libraries:
-  core:
-    gitVersion:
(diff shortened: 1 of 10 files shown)


testing/test-infra/app.yaml
```diff
@@ -1,12 +1,6 @@
 apiVersion: 0.0.1
 kind: ksonnet.io/app
 libraries:
-  core:
```
jlewi: Keep this and check in vendor.  We want to check in vendor so that the code in the repo is sufficient. Please update PR description accordingly.

lluunn: Done

Issue #154, opened 2018-01-26
[Discussion] Contour vs. Ambassador for reverse proxy at the edge


jlewi:
I think we want a reverse proxy at the edge to facilitate access from outside the cluster to all the resources inside the cluster (#11).

Some features we want 

- Ability to dynamically add routes (e.g. as we up/down TensorBoard or TFServing servers)
- Ability to do JWT verification

Since Envoy is becoming the workhorse for Kubernetes & Istio, an Envoy based solution seems ideal.

It looks like both [Contour](https://github.com/heptio/contour) and [Ambassador](https://www.getambassador.io/) do this.

Does anyone know how these solutions differ?

jlewi:
Here are some of the reasons why I initially chose Ambassador

1. I learned of Ambassador first and had a prototype working before learning about Contour
1. I don't understand how the contour model of using an ingress controller is meant to interact with cloud ingress controllers

Per the definition of [ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/#what-is-ingress) I think of ingress controllers as routing traffic into the cluster from outside.

I think there's a design question of whether all ingress should terminate on a single service (a bastion so to speak) and then be routed to the particular service or if the external loadbalancer should directly route traffic to individual services.

I favor a bastion type design as diagrammed in kubeflow/kubeflow#11

One reason for this is ingress is one of those K8s aspects that varies the most across cloud providers. I expect proper setup of ingress to be one of the biggest sources of user problems/questions. So I want to minimize the complexity of our ingress design.

With a bastion type design every Kubeflow cluster has a single ingress into the cluster that terminates on a single service Ambassador/Envoy in the cluster. From there the reverse proxy solution should be the same across K8s distributions. Furthermore, the ingress configuration doesn't change as you bring up services (e.g. TensorBoard instances, TFServing instances etc...). So with this design you basically setup ingress once as part of Kubeflow deployment and then it never changes. 

If we use ingress, then every time we change the configuration (e.g. adding a route for a new TensorBoard instance) we are potentially interacting with an external loadbalancer that is different across K8s distributions. That seems like a recipe for things to break in exciting ways.

Based on the docs, it was easy for me to see how Ambassador can be used in this case.

I imagine the same is doable with Contour; i.e. by having multiple ingress controllers and using the Controur ingress controller to route traffic within cluster. But that doesn't fit my mental model that Ingress is for north/south traffic.

[...]

PR #166, opened 2018-01-29
Use ambassador as a reverse proxy

jlewi:
* [Ambassdor](https://www.getambassador.io/) provides a reverse proxy that can be dynamically configured; we can mappings just by creating K8s services and anotating them with the prefix we want to serve the path on.

* Ambassador makes it really easy to configure our reverse proxy to direct traffic to an ephemeral service (e.g. a Tensorboard instance serving data for a particular job).

* This PR adds Ambassador and adds some default routes; e.g. we add a route to serve the K8s dashboard at /k8s/ui/

* In follow on PRs we can annotate other services (like the TFJobs UI) and TensorBoard deployments so that we can create mappings for those servers as well.

* Related to:
    #11 Secure Proxy 
    #60  IAP on GKE
   #154 Contour vs. Ambassador.


jlewi:
@lluunn  Ping?

lluunn:
will take a look later today or tomorrow morning.

ddysher:
I come across this PR while searching for ambassador usage in kubeflow; before creating an issue, I feel like ask here first:

- is there any documentation on how we use ambassador?  ambassador is installed via core package by default, but there's no docs on why we need it and how to use it in kubeflow
- How do we dynamically direct traffic, considering that ephemeral services come and go?

ddysher:
@jlewi 

/cc @gaocegege @codeflitting

jlewi:
@ddysher 

The architecture is described in kubeflow/kubeflow#11

Routes can be added to ambassador by adding an annotation to the K8s service. A good example is the annotation that we add to add routes to specific tensorboard instances
https://github.com/kubeflow/kubeflow/blob/a50c2fb0e46cb3d9812be1ea1623749e13f63191/kubeflow/examples/prototypes/tensorboard.jsonnet#L23

ddysher:
By we, you mean whoever deploy tensorboard, and via ksonnet?
> A good example is the annotation that we add to add routes to specific tensorboard instances

jlewi:
Ksonnet is orthogonal. I meant how Kubeflow is using it. All you need to do
is add the appropriate annotaion to your K8s service. Tensorboard is an
example of that.  You can use whatever templating solution you want or just
specify the spec via YAML

ddysher:
Thanks @jlewi.  Internally, we have components that manage this automatically (via ingress); i'll see if there is any integration point.

PR #171, opened 2018-01-30
Missing namespace scope for commands in userguide

barney-s:
missing namespace for kubectl commands in user guide


barney-s:
@jlewi minor fix to userguide. PTAL

Issue #21, opened 2017-12-08
Add Argo package to our ksonnet registry


jlewi:
https://applatix.com/open-source/argo/

This looks like it could be very useful for

- Decoratively describing pipelines/workflows
- Building containers from notebooks
  
We should create an Argo package in our Ksonnet registry to make it easy to deploy Argo with Kubeflow.

We should be able to use the libsonnet files in our test infrastructure as a starting point.

vishh:
We need automated builds as part of CI.
That is different from a user journey of moving from interactive to batch training. 
Can we have separate issues for both?

jessesuen:
Let us (argo) know we can be of help in any way with this effort. We took your feedback to heart and are implementing both DAG dependency based execution of tasks, as well as the ability to create/apply/patch/delete any k8s resource in steps (including CRDs).

jlewi:
Thats fantastic thank you.

jlewi:
Filed kubeflow/issues#87 to track overall CUJ per comment from @vishh.

Lets narrow the scope of this issue to adding Argo to our ksonnet registry.

jlewi:
Are test infra already uses ksonnet to deploy Argo
https://github.com/google/kubeflow/blob/master/testing/test-infra/components/argo.libsonnet
It should be straightforward to turn this into a component in our registry.

jlewi:
We need an E2E test for the component.

jlewi:
Opened up a new issue for the test. Closing this one.

PR #178, opened 2018-01-31
Argo ksonnet package

jlewi:
* An argo ksonnet package
* Makes it easy to deploy argo

* Related to #21

### DIFF
--- a/kubeflow/argo/argo.libsonnet
+++ b/kubeflow/argo/argo.libsonnet
@@ -7,17 +7,17 @@
 
   parts(namespace):: {
     all::[
-      argo.parts(namespace).crd,
-      argo.parts(namespace).config,
-      argo.parts(namespace).deploy,
-      argo.parts(namespace).deployUi,
-      argo.parts(namespace).uiService,
-      argo.parts(namespace).uiIngress,
-      argo.parts(namespace).serviceAccount,
-      argo.parts(namespace).roleBinding,
-      argo.parts(namespace).defaultRoleBinding,
+      $.parts(namespace).crd,
+      $.parts(namespace).config,
+      $.parts(namespace).deploy,
+      $.parts(namespace).deployUi,
+      $.parts(namespace).uiService,
+      $.parts(namespace).uiIngress,
+      $.parts(namespace).serviceAccount,
+      $.parts(namespace).roleBinding,
+      $.parts(namespace).defaultRoleBinding,
     ],
-    
+
     // CRD's are not namespace scoped; see
     // https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/
     crd: {
--- a/kubeflow/argo/argo.libsonnet
+++ b/kubeflow/argo/argo.libsonnet
@@ -263,7 +263,7 @@
           apiGroups: [""],
           resources: [
             "pods",
-            "pods-exec",
+            "pods/exec",
           ],
           verbs: [
             "create",
@@ -368,7 +368,7 @@
           apiGroups: [""],
           resources: [
             "pods",
-            "pods-exec",
+            "pods/exec",
             "pods/log",
           ],
           verbs: [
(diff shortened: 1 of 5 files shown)


kubeflow/argo/argo.libsonnet
```diff
@@ -0,0 +1,320 @@
...
+      apiVersion: "v1",
+      kind: "ServiceAccount",
+      metadata: {
+        name: "argo",
+        namespace: namespace,
+      },
+    },  // service account
+
+    // TODO(jlewi): Do we really need cluster admin privileges? Why?
+    // is this just because workflow controller is trying to create the CRD?
```
jessesuen: In v2.0.0-beta1, the controller no longer attempts to create the CRD. This is left to installer now.

cluster-admin would be needed if workflow-controller needs to manage workflows across namespaces. If the controller is configured to only operate in a single namespace (this can be configured in the controller's configmap), then I think admin should be sufficient. But I haven't tried this yet.

We identified the minimum privileges required by the controller here:
https://github.com/argoproj/argo/blob/master/cmd/argo/commands/const.go#L20

Note that UI and controller have different set of privileges.

In 2.0.0-beta1, the installer will create a cluster-admin using these privileges. But if you would like to hand roll the deployment/rbac manifests yourself, argo 2.0.0-beta1 now has `--dry-run` flag to print the manifests without installing, and you can tweak it from there.

jlewi: Very helpful thanks. I updated to the beta1 version and locked down the permissions. PTAL.

kubeflow/argo/argo.libsonnet
```diff
@@ -0,0 +1,339 @@
...
+                terminationMessagePath: "/dev/termination-log",
+                terminationMessagePolicy: "File",
+              },
+            ],
+            dnsPolicy: "ClusterFirst",
+            restartPolicy: "Always",
+            schedulerName: "default-scheduler",
+            securityContext: {},
+            serviceAccount: "argo",
+            serviceAccountName: "argo",
```
jessesuen: Oh I forgot to mention another change is that there are now two service accounts: one for controller, the other for UI with different permissions. (e.g. the UI needs access to logs whereas the controller does not). I notice in this change the UI is using the same service account (argo). I think you will want to create the argo-ui service account mapped to these permissions:

https://github.com/argoproj/argo/blob/master/cmd/argo/commands/const.go#L44

jlewi: I think you did mention that. PTAL.

kubeflow/argo/argo.libsonnet
```diff
@@ -0,0 +1,429 @@
...
+        },
+        name: "argo-ui",
+        namespace: namespace,
+      },
+      rules: [
+        {
+          apiGroups: [""],
+          resources: [
+            "pods",
+            "pods-exec",
```
jessesuen: Should this be pods/exec instead of pods-exec?

kubeflow/argo/argo.libsonnet
```diff
@@ -0,0 +1,429 @@
...
+        },
+        name: "argo",
+        namespace: namespace,
+      },
+      rules: [
+        {
+          apiGroups: [""],
+          resources: [
+            "pods",
+            "pods-exec",
```
jessesuen: Same here. I've only ever seen this pods/exec

jessesuen: Yes, the docs say to "use a slash to delimit the resource and subresource"
https://kubernetes.io/docs/admin/authorization/rbac/#referring-to-resources

jessesuen:
BTW, I will be working with ksonnet to generate an "official" argo workflow libsonnet for the workflow spec parts. This should be coming soon after I generate the OpenAPI specs for workflows.

kubeflow/argo/argo.libsonnet
jlewi: Thanks for catching this; it was a bad copy and paste error on my part.

jlewi:
Thanks.

Issue #174, opened 2018-01-30
Create a repository for examples: kubeflow/examples


jlewi:
I think we should create a repository where we can start hosting E2E examples.

Right now we have a couple E2E examples in various states of development

  *  https://github.com/cwbeitel/kubeflow-rl - Using TensorFlow Agents for RL
  * #157 GitHub Issue Summarization
  * #163 XGB example
  * Hot Dog Or Not - Kubeflow Kubecon 2017 demo

/cc @puneith @cwbeitel

puneith:
@jlewi Why do you think we need a separate repo and not Kubeflow be that place?

cwbeitel:
I have some additional examples in progress to add.

Would love to have them under e2e test. But for some examples you may want to do progressive testing to speed things up / cache previous runs. Such as with Bazel. That could be a mod to the way kubeflow runs tests or could be a reason to have a separate repo.

jlewi:
@puneith Because examples will be in various states of development and should probably be separate from the core code.

For the most part, this repo corresponds to our ksonnet registry.

jlewi:
Another reason for moving examples into their own repository is that it will probably be a different group of folks working on the examples then the folks working on our core ksonnet components.

puneith:
I agree with that model.

gaocegege:
Vote for separate repo 👍

elsonrodriguez:
I could go either way, but my doc fixes have included ksonnet changes. Managing two separate repos for things that are tied together may be painful.

jlewi:
@elsonrodriguez I think the fact that writing examples requires modifying our core ksonnet components is an indication of other problems

  * immaturity/limitations of the ksonnet templates
       * e.g. I think in some of your changes you had to add parameters to the TF serving component to allow more flexibility
  * version compatibility issues because we haven't started creating tagged releases yet
       * By March we should have a 0.1 branch for ksonnet registry and our examples should be written against that.

jlewi:
Lets give a separate repo a shot and then adapt as need be.

I think the main priority now is getting more examples that illustrate Kubeflow so velocity is the primary motivation. With a separate repo we can easily start giving a different set of folks write access and moving the ball forward. 

@elsonrodriguez If you want to take the lead on an inception example I'm happy to make you an approver.

jlewi:
Created the repo:
https://github.com/kubeflow/examples

Created a team examples-approvers and gave it write access to the repo.
[...]

PR #181, opened 2018-01-31
Model Server Component documentation triage.

elsonrodriguez:
- Made it so that the image created in the instructions is utilizied
- Removed downloaded model due to incompatibility with recent TF server versions
- Made tf-serving image configurable
- Minor flow/grammar tweaks

### DIFF
--- a/components/k8s-model-server/README.md
+++ b/components/k8s-model-server/README.md
@@ -189,7 +189,7 @@ $MODEL_NAME  LoadBalancer   <INTERNAL IP>   <SERVICE IP>     <SERVICE PORT>:<NOD
 We will feed the `<SERVICE IP>` and `<SERVICE PORT>` to the labelling script. We will use it to label the following image of a
 cat sleeping on a comforter atop a sofa:
 
-![Cat on comforter on sofa](./inception-client/images/sleeping-pepper.jpg =200x)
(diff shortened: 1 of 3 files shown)


elsonrodriguez:
Should I also add an example of using the inception client to the user_guide?

jlewi:
I thought @nkashy1 already added an example of using the client?

elsonrodriguez:
They did, but in the component documentation, not the user_guide. It's a bit of repetition, but might make things easier to understand.

If yes I can do in another PR.

jlewi:
How about creating an E2E example using inception? Some things that might be demonstrated

    * Generating predictions from outside the cluster (e.g. via secure reverse proxy)
    * A web app that can act as a FE [here's](https://github.com/vishh/container-engine-accelerators/tree/example-gpu/example/inception-golang/src) the code used in our demo at Kubecon
     * Clear docs for running the sample

Should we centralize all the docs and code specific to inception in one location and avoid duplication?

I think everything in components/  should be generic and not be specific (e.g. inception). We could create an examples/inception directory, although I think my preference would be to create a separate repository #174.

jlewi:
@elsonrodriguez since @lluunn LGTM'd this I'm happy to merge it once its sync'd and we can follow up in another issue PR regarding a more complete example for inception.

elsonrodriguez:
@jlewi synced

Issue #102, opened 2018-01-08
Define common API for prediction


jlewi:
We need a common API for inference. This is critical to being able to support a variety of different frameworks (e.g. TensorFlow, scikits, xgboost).

There's a lot of prior art here. A lot of the existing frameworks define inference APIs. We should start by collecting those and seeing if it makes sense to adopt a slightly modified version of one of them.

I think the aim should be to produce an [OpenAPI spec](https://en.wikipedia.org/wiki/OpenAPI_Specification) for the prediction API.

Some frameworks we should look at:
  * [Seldon IO](https://github.com/SeldonIO/seldon-core/blob/master/docs/reference/prediction.md)
  * [H20](https://github.com/h2oai)
  * [Pipeline.AI](https://github.com/pipelineai/pipeline)
  * [CMLE](https://cloud.google.com/ml-engine/docs/v1/predict-request)

 Does anyone have links to definitions for the H20 and Pipeline.AI APIs?

/cc @cfregly @seeldon @cliveseldon @lluunn

cliveseldon:
At Seldon, as the above link shows our API is described as a gRPC proto spec. For the external API we provide two endpoints, predict and feedback. Feedback allows "rewards" to be sent to allow users of the API to return the success or not of the predictions for use by online optimization algorithms, for example multi-armed bandit solvers.

The core message type SeldonMessage has various variants. The DefaultData has either a Tensor representation (most likely for use by gRPC users) and a NDArray representation (most likely for use by REST JSON users sending a list of list of features). There are also general string and binary representations for users whose use case doesn''t fit the DefaultMessage.

Added to this there is Meta and Status parts of the message. This allows the SeldonMessage to be quite generic and be used for both request and response parts of calls.

There seem to be a few questions:
  1. Does KubeFlow want to handle both REST and gRPC
  1. If both how will the gRPC spec be derived from the OpenAPI? We have an [issue](https://github.com/SeldonIO/seldon-core/issues/6)  that discusses our situation where we want to create the OpenAPI spec from gRPC spec. I think Kubernetes starts with the Go Structs which are commented and then the OpenAPI spec and proto files are generated from that.
 
There is also the question of how KubeFlow will integrate with the other projects mentioned and whether the API needs to be a separate thing that multiple projects can adhere to? 

@Maximophone from Seldon might have some comments as he worked on the API.

elmiko:
> I think the aim should be to produce an OpenAPI spec for the prediction API.

[...]

Issue #64, opened 2017-12-22
TfServing uber tracking bug


jlewi:
Opening an uber tracking bug to keep track of various features/improvements to ensure TfServing is well supported in Kubeflow

- [ ] Publish Docker images #50 
- [ ] Health/liveness checks #368
- [ ] Monitoring #369
- [ ] GPU support
    
/cc @rhaertel80 @kow3ns

nkashy1:
TensorFlow Serving client library support - right now TensorFlow Serving only publishes a package to PyPI for Python 2 ([link](https://pypi.python.org/pypi/tensorflow-serving-api)).

This means that users can't run this library from JupyterHub if they just follow our standard procedure.

yupbank:
Health/liveness checks can be done from http-proxy

jlewi:
@yupbank Can you explain what you mean by doing health and liveness checks from the http proxy? Would that be testing the correct thing?

yupbank:
oh... for that i was referring to  k8s liveness probe and readiness probe ...

jlewi:
@yupbank That part I understand but how does the http proxy help us determine the liveness/readiness of TFServing? The http proxy could be live/ready but the TFServing might not.

yupbank:
since now we have this https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/model_service.proto#L16

http proxy can be forced to only ready when TFServing is ready.

jlewi:
Closing this bug because we have more specific bugs for individual issues.

lluunn:
Should we have one for logging too?

zmoon111:
i use util percent of  cpu&gpu to check if it is in live status

PR #183, opened 2018-02-01
add http proxy for tf-serving

yupbank:
Since REST Api is still widely used and tf-serving the c++ binary don't support http request.

This pr use python tornado with async supported web server to do the proxy 

### relationship between this pr and #65 

differences: 
  - #65 providing a big application supports tf-serving, xgboost and sklearn.  while this pr is just a thin layer on top of tf-serving
  - #65 is using paste web server instead of tornado.
  - #65 is spawning a tf-serving process inside the docker, this pr use tf-serving as a grpc provider
  

similar:
  - both utilizing tensorflow/serving. 
  - both provide CMLE like REST api

since #65 is too big for one pr and the author intend to split it, this pr can potentially help the tf-serving part. as the code is only one file.

Related to:
  * Define common prediction API(#102)
  * TfServing Uber issue (#64)


components/k8s-model-server/http-proxy/test_server.py
```diff
@@ -0,0 +1,73 @@
+import json
```
jlewi: this file should be named server_test

yupbank:
> How does this compare to the http proxy functionality in #65?

we had a discussion, and i think @lluunn is good with this..


>Also how does it compare to the functionality in Seldon.io?

This is reusing https://github.com/tensorflow/serving while Seldon.io have their own serving backend and packaging standard.  It is also good for new ML server adopters. While for some people like us, we already doing tensorflow/savedModel based pipeline, we can really benefit from this

jlewi:
>>How does this compare to the http proxy functionality in #65?

>we had a discussion, and i think @lluunn is good with this..

I know but could you highlight any substantial differences between this pr and #65?

Is there any significant difference in the implementation? Do you both use Tornado?

Is there any difference in the REST API? Is the intent to maintain compatibility with the Google CMLE prediction API so that CMLE users could serve on Kubeflow?

yupbank:
@jlewi  hi, i have updated your questions in the first comment. 

>Is the intent to maintain compatibility 

for this... i guess for this pr, the intent is just borrowed the api from CMLE.

jlewi:
Thanks. I'll finish reviewing the PR as soon as I have a second.

I chatted with the folks here at Google and the prefer tornado over paste (better event loop) so that's an improvement.

components/k8s-model-server/http-proxy/README.md
```diff
@@ -0,0 +1,40 @@
+# REST API http proxy to tensorflow/serving
+----
+
+This project provides web server that proxy http rest api to Tensorflow/Serving grpc.
```
jlewi: Nit: "This component provides a"

components/k8s-model-server/http-proxy/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+The REST Api Spec is similar to Google Machine Learning Engine Model Serving.
```
jlewi: Google Cloud Machine Learning Engine Prediction API

components/k8s-model-server/http-proxy/app.yaml
```diff
@@ -0,0 +1,36 @@
+---
```
jlewi: What's the YAML for? Can we make it a ksonnet template? I think it would make sense to add a ksonnet template which has TF Serving and the proxy in a sideacar.

components/k8s-model-server/http-proxy/server.py
```diff
@@ -0,0 +1,142 @@
...
+import tensorflow as tf
+import tornado.web
+
+
+define("port", default=8888, help="run on the given port", type=int)
+define("rpc_timeout", default=1.0, help="seconds for time out rpc request", type=float)
+define("rpc_port", default=9000, help="tf serving on the given port", type=int)
+define("rpc_address", default='localhost', help="tf serving on the given address", type=str)
+define("instances_key", default='instances', help="requested instances json object key")
+define("cookie_secret", default='YOU_NEED_THIS', help="cookie secrets")
```
jlewi: What is the cookie secret for?

yupbank: just a web server thing. if the web server is using cookie. maybe i should remove it. it is not required

components/k8s-model-server/http-proxy/app.yaml
yupbank: i'll remove this and add following up ksonnet template

jlewi:
LGTM. Can you sync though, looks like files are out of date?

PR #206, opened 2018-02-05
Changing case for TfJob CRD.

elsonrodriguez:
Not sure if this is the right fix, but it works. I can do the proper thing if this is incorrect.


elsonrodriguez:
Looks like the proper fix is the exact opposite of this.

Working on it.

PR #208, opened 2018-02-06
Changing tensorflow.org to kubeflow.org. Fixing CRD deployment.

elsonrodriguez:
Sidenote: I've managed to pin most things to make sure the documentation is more robust. However I don't know how best to pin the kubeflow repo itself.

Ideally in the docs something like `ks registry add kubeflow github.com/kubeflow/kubeflow/tree/9a3e71f3/kubeflow` is done, however without a target tag I'd have to do a follow up PR to pin everything to the desired commit.

### DIFF
--- a/kubeflow/core/prototypes/all.jsonnet
+++ b/kubeflow/core/prototypes/all.jsonnet
@@ -6,7 +6,7 @@
 // @optionalParam namespace string default Namespace
 // @optionalParam disks string null Comma separated list of Google persistent disks to attach to jupyter environments.
 // @optionalParam cloud string null String identifying the cloud to customize the deployment for.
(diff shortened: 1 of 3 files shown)


jlewi:
@elsonrodriguez Sent you an invite to the Kubeflow org so you can trigger tests.

I think the way to pin things is to create a tagged branch and then tell people to use that branch rather than master.  If you wanted to go ahead and create an issue to do that and possible a PR, I'd be very supportive. I think that would be very helpful to figure out how we will actually manage releases.

You could pick a version tag like 0.1 (or use something less < 0.1 if you think we should reservere 0.1 for our March release).

jlewi:
Tests are broken because of #209. Do you want to review that PR?

jlewi:
I confirmed this manually fixes things.

jlewi:
@elsonrodriguez I merged #209 can you sync and rerun the tests?

PR #209, opened 2018-02-06
Update tests to use shared code in kubeflow/testing

### LINKED ISSUE #205: Use kubeflow/testing to run Argo workflows
* Now that kubeflow/testing is setup we should be using it to run our Argo workflows
* Here are the changes we should make

1. Use the external NFS volume as opposed to running NFS in cluster (this appears to be more performant)
1. Change our Dockerfile to be like the one for [tensorflow/k8s](https://github.com/tensorflow/k8s/tree/master/test/test-infra)
1. Delete the code in the [testing directory](https://github.com/kubeflow/kubeflow/tree/master/testing) that should be pulled from kubeflow/testing
   * bootstrap.sh
   * checkout.sh
   * argo_client.py
   * prow_artifacts.py

jlewi:
* Reuse code in kubeflow/testing for launching Argo jobs as E2E tests.
* Delete the code in the testing directory that has moved to kubeflow/testing
* Delete ksonnet configs for setting up the test infrastructure; these are all in kubeflow/testing now
* Fix #205


testing/Dockerfile
```diff
@@ -1,100 +1,8 @@
...
-RUN chmod a+x /usr/local/bin/bootstrap.sh
-
-COPY checkout.sh /usr/local/bin
-RUN chmod a+x /usr/local/bin/checkout.sh
-
-ENTRYPOINT ["/usr/local/bin/bootstrap.sh"]
\ No newline at end of file
+# Dockerfile used by out prow jobs.
+# The sole purpose of this image is to customize the command run.
+FROM gcr.io/mlkube-testing/test-worker:latest
```
elsonrodriguez: Does it make sense to pin this to a known working tag?

testing/workflows/components/workflows.libsonnet
```diff
@@ -142,9 +149,14 @@
...
                 args: [
-                  srcDir,
+                  srcRootDir,
                 ],
-                env: prow_env,
+                env: prow_env + [{
+                  "name": "EXTRA_REPOS",
+                  // TODO(jlewi): Once kubeflow/testing#12 is submitted pin
+                  // to kubeflow/testing HEAD
+                  "value": "tensorflow/k8s@HEAD;kubeflow/testing@HEAD:12",
```
lluunn: Do you need kubeflow/testing as EXTRA_REPOS? I think in https://github.com/kubeflow/testing/pull/12/files#diff-c385e0edb32af89cbebb9b1abbe2024cR91, it always checks out that repo?

testing/Dockerfile
jlewi: Maybe at some point in the future. Right now pinning would probably just get in the way of things.

testing/workflows/components/workflows.libsonnet
jlewi: Yes. Because we need a PR that isn't checked in yet. Also as noted in the TODO for kubeflow/testing#12 I think it makes sense to stop giving special treatement to kubeflow/testing.

testing/workflows/components/workflows.libsonnet
```diff
@@ -30,12 +30,19 @@
       // outputDir is the directory to sync to GCS to contain the output for this job.
       local outputDir = testDir + "/output";
       local artifactsDir = outputDir + "/artifacts";
-      local srcDir = testDir + "/src";
-      local image = "gcr.io/mlkube-testing/kubeflow-testing";
+      // Source directory where all repos should be checked out
+      local srcRootDir = testDir + "/src";
+      // The directory containing the kubeflow/kubeflow repo
+      local srcDir = srcRootDir + "/kubeflow/kubeflow";
+      local image = "gcr.io/mlkube-testing/test-worker:latest";
```
elsonrodriguez: Same as above.

jlewi: Same as above. I think eventually we will want to start pinning things but right now its not obvious that helps and that not pinning hurts. This is an area of flux in our test infra.

jlewi:
Merging to unblock #208 which should fix head.

Issue #194, opened 2018-02-01
Add GPU Support for k8s-model-server on Kubeflow


bkungfoo:
As a feature request, adding GPU support for TF-serving in Kubeflow can help speed up batch predictions that can take a long time to run on a CPU-only machine. Unfortunately, existing docker images, such as the one from the TF-serving repository, do not build correctly (or are not well-maintained): https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/Dockerfile.devel-gpu

It would be good to have kubeflow maintain a working GPU docker image that can be deployed with the k8s-model-server.

DjangoPeng:
Cool! It's very helpful to speed up the inferencing and prediction work.

flx42:
It's @flx42 :)

jlewi:
As part of this issue we should create an Argo workflow to build and test this Docker image as in #214.

jlewi:
Opened up:

#291 E2E test for TFServing with GPUs.
#292 Add GPU support to ksonnet packages.

PR #210, opened 2018-02-06
feat(model-server): add Dockerfile of model-server with gpu support

DjangoPeng:
Related to https://github.com/kubeflow/kubeflow/issues/194

### How to test

You could train a mnist model with https://github.com/tensorflow/serving/blob/master/tensorflow_serving/example/mnist_saved_model.py. 

Then mount the mnist saved model into tensorflow model server container and launch the gRPC server with mnist model.

Finally, you could launch a gRPC client with https://github.com/tensorflow/serving/blob/master/tensorflow_serving/example/mnist_client.py.

IMO, the saved model should not be added into the image. But we could put it into `model-server:1.4-example` image.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/210)


DjangoPeng:
I've built it with Nvidia 1080Ti on bare metal. Welcome for testing and feedback. 

@bkungfoo @jlewi PTAL

jlewi:
@lluunn Do you want to review this?

@vishh FYI looks like this is using TF 1.4 whereas your work was using TF 1.5.

DjangoPeng:
It's ready to be reviewed.

@vishh Do you use any new features released in TensorFlow 1.5? if so, I could build a `Dockerfile.gpu` for TensorFlow 1.5

@jlewi As we discussed before, Chinese folks can't pull any images from gcr.io without VPN. We'd better add and maintain another images repo of kubeflow in dockerhub. At first, we could add the `registry name` into `Makefile`.

components/k8s-model-server/docker/Dockerfile.gpu
```diff
@@ -0,0 +1,96 @@
...
+        zlib1g-dev \
+        && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# Set up grpc
+RUN pip install mock grpcio
+
+# Set up all envs
+ENV BAZELRC=/root/.bazelrc \
```
yupbank: https://github.com/tensorflow/tensorflow/blob/r1.4/tensorflow/tools/docker/Dockerfile.devel-gpu#L80-L83
certain env are already default  though...

components/k8s-model-server/docker/Dockerfile.gpu
```diff
@@ -0,0 +1,96 @@
...
+    cd serving && \
+    git checkout $TF_SERVING_VERSION && \
+    git submodule init && \
+    git submodule update --recursive
+
+# Configure Tensorflow to use the GPU
+RUN cd /root/serving/tensorflow && ./configure
+
+# Compile TF serving with CUDA support
+RUN cd /root/serving && \
```
yupbank: A general question, can we test this somewhere? since i had some experience with too many bazel jobs.. and my os shut it down..

components/k8s-model-server/docker/Dockerfile.gpu
DjangoPeng: What's the meaning of `already default`? There is no exported env in the Dockerfile.gpu

components/k8s-model-server/docker/Makefile
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 VERSION=1.0
+GPU_VERSION=1.4
```
yupbank: this is gpu version or tensorflow version or tensorflow-gpu version?

components/k8s-model-server/docker/Dockerfile.gpu
yupbank: oh... sorry, by that i mean, you are just setting the default value... which can be omitted following the tensorflow dockerfile example..

components/k8s-model-server/docker/Makefile
DjangoPeng: TensorFlow Version.

components/k8s-model-server/docker/Dockerfile.gpu
DjangoPeng: See first comment above.

jlewi:
@DjangoPeng ack; opened #211

@lluunn is working on an Argo workflow to build and release Docker images for TF Serving. It should be easy to modify that build process to publish to multiple locations; one being DockerHub.

For testing, rather than running mnist can we run a very simply model e.g. X * Y and verify that the ops ran on GPUs? (This can be a follow on PR). This is what we do to verify training works with GPUs.

components/k8s-model-server/docker/Makefile
```diff
@@ -21,7 +22,23 @@ all: build
 build:
 	docker build --pull -t ${PROJECT}/model-server:${VERSION} .
 
+build-gpu:
+	docker build --pull -t ${PROJECT}/model-server:${GPU_VERSION} -f Dockerfile.gpu .
```
lluunn: version=1.0, gpu_version=1.4, then we can't tell from the tag the the image is for gpu or not.
I think we should change it to xxx/model-server-gpu:$GPU_VERSION

jlewi:
@DjangoPeng Any update on this? Would be great if we could get this submitted soon and then work on combining it with @lluunn's work using Argo to build the images.

It would be great if by the end of this week we could be continuously building Docker images for TF serving for CPU & GPU.

DjangoPeng:
@jlewi @lluunn Thanks for your comment. I'll update today.

components/k8s-model-server/docker/Makefile
DjangoPeng: Gotcha, thanks

components/k8s-model-server/docker/Makefile
```diff
@@ -12,31 +12,72 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-VERSION=1.0
+TF_VERSION=1.0
```
DjangoPeng: Explicit is better than implicit.

components/k8s-model-server/docker/Dockerfile.gpu
DjangoPeng: I see. How about replacing `env` with `args` ?

DjangoPeng:
PTAL @jlewi @lluunn 

We'd better merge the PR to use Argo at first. Then we can pick up some envs and set them as args. WDYT @yupbank

components/k8s-model-server/docker/Dockerfile.gpu
yupbank: sounds good!

jlewi:
@DjangoPeng Whatever you @yupbank and @lluunn think best SGTM.

I'd even be fine committing this as is and iterating on it in a follow on PR to add continuous building with Argo.

bkungfoo:
@DjangoPeng Thanks for the tf-serving gpu Dockerfile! 
Do we have a working image already available on gcr.io/kubeflow? I'm currently using the image built by @vishh .

DjangoPeng:
> Do we have a working image already available on gcr.io/kubeflow? 

I'm not sure. 

If not, could you build and push it to gcr.io/kubeflow ? @jlewi

jlewi:
#291 is tracking adding a test and building the image continuously. We aren't doing it yet.

PR #216, opened 2018-02-07
Delete components/tf-controller

zmhassan:
The tf-controller directory should no longer be needed because we now use ksonnet.


ScorpioCPH:
@zmhassan Please squash your commits, thanks.

jlewi:
@ScorpioCPH I always squash the commits when I merge it into master. Is there any reason to before having the user do it? What does it due to code review history if the user squashes the commits?

Issue #211, opened 2018-02-06
Publish Docker images to DockerHub in addition to gcr.io


jlewi:
gcr.io is unfortunately blocked in China. So we should probably publish all our Docker images to DockerHub as well.

See #197 for getting ownership of kubeflow in Dockerhub.

s1113950:
@jlewi I'll take this one on

s1113950:
I'm in the process of getting my cla good to go (Intel, same team as @elsonrodriguez ), but should have the PR good to go by about EOD

jlewi:
What is your thinking in terms of how you will do this?

- Will our release pipelines just push to both locations?
- Will you run a sync job that will automatically sync all images from one registry to the other?

I think it would be nice if there was an easy way to mirror a docker registry.

s1113950:
My first idea was just to modify the release pipeline to push to both locations (I've been editing the various Makefiles to do this). Long-term, a sync job or a way to mirror the registries would be my preferred approach as well. 

It looks like we could do the sync relatively easily though? I found this https://cloud.google.com/container-registry/docs/using-dockerhub-mirroring. Do you think I should go this route from the beginning or stick with pushing to both places for now?

I also had a question regarding getting kubeflow ownership; I see that it's a public repo but I was unable to push to it. What's the process for getting that to be a community repo or getting permissions to push to it?

s1113950:
Take a look at https://github.com/kubeflow/kubeflow/pull/222 and let me know what you think @jlewi

jlewi:
If you're willing, I'd much prefer the mirroring route.

I think https://cloud.google.com/container-registry/docs/using-dockerhub-mirroring is mostly about a mirror of DockerHub in GCR to make fetching images from GCP more efficient. So one option would be to make DockerHub the source of truth and then rely on GCR mirror to make it efficient on GCP.

But I don't think that will work very well for our release/test infra since the mirror has a 30 minute delay. 

How difficult would it be to write a simple script that would list images using the Docker registry API and then push missing images from one repo to the other? It surprises me that no such tooling exists.

Regarding community ownership, we're in the process of setting up Gsuite for kubeflow.org. At which point we can create GCP projects in kubeflow.org for all our test/release infrastructure and easily share them with community members.

@lluunn is working on building images using Argo. I think the direction we want to go is to trigger Docker images as part of pre/post submits by using Prow to launch Argo.

s1113950:
[...]

Issue #197, opened 2018-02-02
Make https://hub.docker.com/r/kubeflow/jupyterhub/ a community resource


jlewi:
The repo is currently owned by:  @carmark has agreed to make it a community resource once we're ready.

Once we have gsuite setup for kubeflow.org we should create an appropriate group with appropriate members and ask xuelei to share the docker hub repo with it.

gaocegege:
FYI Lei Xue's GitHub ID is @carmark

carmark:
@gaocegege Thanks.
@jlewi Could u please use the github ID instead of email?

jlewi:
Sorry didn't know your GitHub id.

carmark:
NP, thanks.

jlewi:
@carmark How does DockerHub permissions work? I don't suppose it uses GoogleGroups?

jlewi:
Can we use a service account to automate pushing to DockerHub?

carmark:
@jlewi Yeah, we may create one service account to do that, and create some automate building on docker hub based on the Dockerfile.  I can setup it if you think is all right.

jlewi:
@carmark So my thinking is that we should use prow to trigger Argo workflows and use the Argo workflow to build and push Docker images. 

I think the advantage of using Argo over DockerHub or GCB is that its K8s native. Its also easier to add E2E testing of the image before pushing it. 

What do you think?

If we do it this way how do we authenticate to DockerHub? Do we use a service account with secret key?

We will use a K8s cluster in GCP project kubeflow-ci for our continuous builds. Its not fully setup yet; see kubeflow/testing#18.

In the meantime is there anything on the DockerHub side to be done?

carmark:
@jlewi It is better to test before pushing the docker hub, I can not agree any more.  If we do that as this way, we need a service account which has a username and password, and then I will set the permission on that account.  So the work on DockerHub is simple, I can do that anytime.

jlewi:
The Argo workflow for building the images is here
https://github.com/kubeflow/kubeflow/tree/master/releasing/releaser

But we probably don't want to add it directly to that Argo workflow because those images won't be tested yet.

I think the next steps would be

1. Create a DockerHub account "kubeflow-releaser"
2. Create a simple script to pull a list of our images and push them to DockerHub
 
@carmark Do you think you could work on this?

[...]

PR #222, opened 2018-02-08
Adds pushing images to dockerhub as well

s1113950:
Related to https://github.com/kubeflow/kubeflow/issues/211

This is my first contribution to kubeflow, and as such I didn't want to mess with how the Makefiles are laid out (they aren't standardized). Perhaps that's a future task?

I ran into this when trying to build `k8s-model-server/http-proxy` locally:

```
Step 5/8 : COPY requirements.txt /usr/src/app/
COPY failed: stat /var/lib/docker/tmp/docker-builder102266424/requirements.txt: no such file or directory
```
I didn't look into it in detail though. Given that I only made makefile targets that were the same as the previous commands, I don't feel like this is related but just want to make sure.


s1113950:
testing cla check

components/k8s-model-server/http-proxy/Makefile
```diff
@@ -1,16 +1,45 @@
+# Copyright 2017 Google Inc.
```
carmark: 2018?

components/tensorflow-notebook-image/Makefile
```diff
@@ -11,18 +11,34 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 
-all: container
-
 TAG?=$(shell git rev-parse HEAD)
-REGISTRY?=gcr.io/kubeflow
+GCR_REGISTRY?=gcr.io/kubeflow
+DOCKERHUB_REGISTRY?=hub.docker.com/r/kubeflow
```
carmark: How about using the `PROJECT_ID` macros?

testing/Makefile
```diff
@@ -12,20 +12,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-IMG = gcr.io/mlkube-testing/kubeflow-testing
-TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)-$(shell git diff | sha256sum | cut -c -6)
-DIR := ${CURDIR}
 
-all: build
+GCR_IMG = gcr.io/mlkube-testing/kubeflow-testing
+DOCKERHUB_IMG = hub.docker.com/r/kubeflow-testing
```
carmark: same as above.

tf-controller-examples/tf-cnn/Makefile
```diff
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-REGISTRY := gcr.io/kubeflow
-TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)-$(shell git diff | sha256sum | cut -c -6)
+GCR_REGISTRY := gcr.io/kubeflow
+DOCKERHUB_REGISTRY := hub.docker.com/r/kubeflow
```
carmark: same as above.

jlewi:
Is there a way to mirror one docker registry (gcr) to another (DockerHub) so we don't need to push to two locations everywhere?

What are the relevant permissions ownership of the DockerHub repo this is pushing to?

components/k8s-model-server/http-proxy/Makefile
s1113950: ahh, good catch. I noticed this file didn't have a header so I copied it over from another file without changing the year

components/tensorflow-notebook-image/Makefile
s1113950: I can do that. I noticed that each Makefile already was using different variables; I can standardize it to all use PROJECT_ID though

s1113950:
With the exception of `testing`, which I noticed was going to `mlkube-testing/kubeflow-testing`, all of them go to `hub.docker.com/r/kubeflow`. 
I don't have permission to push to this repo: https://hub.docker.com/r/kubeflow/ 
I saw on the original ticket @jlewi you had a comment here: `See #197 for getting ownership of kubeflow in Dockerhub.`; in testing this I pushed to our own Dockerhub instead by overriding the `DOCKERHUB_PROJECT` var during each make call. Presumably @carmark can change permissions to whatever is required for this to work? see https://github.com/kubeflow/kubeflow/issues/197 for more info

https://cloud.google.com/container-registry/docs/using-dockerhub-mirroring this is what I've found so far regarding mirroring docker registries. It looks like we'd follow the section under `Fetch images from Container Registry's mirror`. I'm a bit confused by it though. It looks like it's all related to a local docker daemon? So presumably we might not have to do anything on our end at all other than maybe a script that sets up that config for a user in their local docker daemon?

s1113950:
@carmark I completed the revisions requested

jlewi:
I updated the PR description because this is only a partial fix since we will actually have to rebuild the images and push them. There's also work to do setting up the DockerHub registry and we should probably move the GCR registry to a project that can be owned by the community.

carmark:
@s1113950 The registry mirror is used in docker daemon, so we can not use it to sync the image between docker hub and gcr.

PR #224, opened 2018-02-08
Update README.md to mention OpenShift SCC

pdmack:
Prescriptive fix for deploying the ambassador and jupyter-hub components in OpenShift


pdmack:
I signed it!

jlewi:
@pdmack sent you an invitation to the Kubeflow org so that you will be able to trigger tests.

PR #225, opened 2018-02-08
Made TFJob CRD requirement explicit

nkashy1:
(In user guide, just a documentation update)

Also made a small change to user guide link in resource section of main
README.


user_guide.md
```diff
@@ -180,6 +180,10 @@ In this example, you should be able to use the inception_client to hit ww.xx.yy.
 
 ### Submiting a TensorFlow training job
 
+**Note:** Before deploying a training job, you will have to make the `TFJob` custom resource available on your kubernetes cluster.
+The [tensorflow/k8s README](https://github.com/tensorflow/k8s) contains the `TFJob` CRD as well as deployment instructions.
```
jlewi: Isn't this confusing? The Kubeflow core deployment contains the TFJob CRD so we should probably point them at the deployment instructions for Kubeflow.

nkashy1: @jlewi : Ah, I didn't realize that there was an explicit dependency on `kubeflow-core` (which seems silly in hindsight). My impression was:

1. `kubeflow-core` -> notebook + programming environment

2. `tf-serving` -> inference with TensorFlow Serving

3. `tf-job` -> training + training environment

So when I read the user guide, I just skipped right to the training section and lazily avoided deploying `kubeflow-core`.

Have modified note to reflect my new understanding.

jlewi:
Please sync

PR #227, opened 2018-02-09
Run TFJob tests as part of Kubeflow tests

jlewi:
* We want to run the TFJob tests as part of the Kubeflow tests so that we can verify the operator is
  actually working.
* We modify the E2E workflow to checkout tensorflow/k8s and then invoke some of the tests.

### DIFF
--- a/testing/workflows/components/workflows.libsonnet
+++ b/testing/workflows/components/workflows.libsonnet
@@ -120,6 +120,8 @@
               },
             },
           ],  // volumes
+          // onExit specifies the template that should always run when the workflow completes.
+          onExit: "exit-handler",
           templates: [
             {
               name: "e2e",
@@ -130,8 +132,8 @@
                 }],
                 [
                   {
-                    name: "test-deploy",
-                    template: "test-deploy",
+                    name: "setup",
+                    template: "setup",
                   },
                   {
                     name: "create-pr-symlink",
@@ -142,6 +144,17 @@
                     name: "tfjob-test",
                     template: "tfjob-test",
                   },                
+                ],
+              ],
+            },
+            {
+              name: "exit-handler",
+              steps: [
+                [
+                  {
+                    name: "teardown",
+                    template: "teardown",
+                  },
                 ],
                 [{
                   name: "copy-artifacts",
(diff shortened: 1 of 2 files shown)


jlewi:
@gaocegege Please take a look. Once this is committed we can move tensorflow/k8s to kubeflow/tf-operator

jlewi:
@gaocegege PTAL. This is the other PR I'd like to get submitted before we move tensorflow/k8s to kubeflow/tf-operator.

This PR ensures we actually run some TFJob tests as part of the Kubeflow tests. So it will help us ensure head is in a good state.

testing/workflows/components/workflows.libsonnet
```diff
@@ -166,17 +194,30 @@
                 ],
               },
             },  // checkout
-            $.parts(namespace, name).e2e(prow_env, bucket).buildTemplate("test-deploy", [
+            $.parts(namespace, name).e2e(prow_env, bucket).buildTemplate("setup", [
               "python",
               "-m",
               "testing.test_deploy",
               "--project=mlkube-testing",
```
lluunn: use the local variable project and cluster?

testing/workflows/components/workflows.libsonnet
```diff
@@ -166,17 +194,30 @@
...
               "--test_dir=" + testDir,
               "--artifacts_dir=" + artifactsDir,
-            ]),  // test-deploy
+              "setup",
+            ]),  // setup
+            $.parts(namespace, name).e2e(prow_env, bucket).buildTemplate("teardown", [
+              "python",
+              "-m",
+              "testing.test_deploy",
+              "--project=mlkube-testing",
```
lluunn: same here

testing/workflows/components/workflows.libsonnet
```diff
@@ -166,17 +194,30 @@
                 ],
               },
             },  // checkout
-            $.parts(namespace, name).e2e(prow_env, bucket).buildTemplate("test-deploy", [
+            $.parts(namespace, name).e2e(prow_env, bucket).buildTemplate("setup", [
               "python",
               "-m",
               "testing.test_deploy",
```
lluunn: do you need to add "setup" for the subparser?

testing/workflows/components/workflows.libsonnet
jlewi: Done.

testing/workflows/components/workflows.libsonnet
jlewi: Done.

testing/workflows/components/workflows.libsonnet
jlewi: Its the last argument. The order of the arguments matters because the first set of parameters are not parameters of the subparser setup so the need to be specified before the subparser command.

PR #231, opened 2018-02-09
fix typo for argo-ui role binding

cliveseldon:
fixes #231


jlewi:
/ok-to-test

@cliveseldon Thanks. I sent you an invite to the org so that you will be able to trigger tests.

PR #242, opened 2018-02-14
Fix the docker file so we do not need wrap cmd every time

yupbank:
Description | Docker field name | Kubernetes field name
-- | -- | --
The command run by the container | Entrypoint | command
The arguments passed to the command | Cmd | args


yupbank:
^ `Failed to filter components; the following components don't exist: [ 'simple_tfjob' ]`

components/k8s-model-server/docker/Dockerfile
```diff
@@ -54,4 +54,4 @@ RUN set -x \
     && [ `id -u $MS_USER` -eq 1000 ] \
     && [ `id -g $MS_USER` -eq 1000 ]
 
-CMD ["/bin/bash"]
+ENTRYPOINT ["/bin/bash", "-c"]
```
jlewi: Why are you starting a shell?

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -72,11 +72,6 @@ local networkSpec = networkPolicy.mixin.spec;
                     name: name,
                     image: modelServerImage,
                     imagePullPolicy: defaults.imagePullPolicy,
-                    // TODO(jlewi): Talk to owensk to figure out why we wrap in a shell.
-                    command: [
-                      "/bin/sh",
-                      "-c",
-                    ],
                     args: [
                       "/usr/bin/tensorflow_model_server --port=9000 --model_name=" + name + " --model_base_path=" + modelPath,
```
jlewi: Shouldn't args be a list?

components/k8s-model-server/docker/Dockerfile
yupbank: true... i should have just delete it

PR #175, opened 2018-01-30
Use spartakus to report usage metrics

jlewi:
* Use Kubernetes reporting tool (spartakus) to report anonymous statistics about Kubeflow usage such as basic cluster stats.

* This is optional and we're making it opt in for now.

* One current limitation is that there's no easy way to give each kubeflow deployment a unique, random id, so it will be hard to distinguish different deployments.
    * users can manually assign a unique id.
    * We could potentially modify spartakus (or the Docker image) to generate a random id. The one downside of this is that the id would be regenerated if the pod restarts.

* Related to #55

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/175)

PR #246, opened 2018-02-15
Fix jsonnet issue in iap-envoy.jsonnet

ankushagarwal:
Error message: RUNTIME ERROR: Unexpected type string, expected boolean

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/246)


kubeflow/core/prototypes/iap-envoy.jsonnet
```diff
@@ -22,8 +22,6 @@ local envoyImage = import 'param://envoyImage';
 local audiencesParam = import 'param://audiences';
 local audiences = std.split(audiencesParam, ',');
 local disableJwtCheckingParam = import 'param://disableJwtChecking';
-local disableJwtChecking = if disableJwtCheckingParam || disableJwtCheckingParam == "true" then
-  true
-else false;
+local disableJwtChecking = disableJwtCheckingParam == "true";
```
jlewi: Can you patch core/util.libsonnet from here
https://github.com/kubeflow/kubeflow/pull/175/files#diff-fbd3d99678fdfe2d482cc3fc2152b615

and then use the function toBool to convert disableJwtCheckingParam?

I think that's better because we want to allow for disableJwtCheckingParam to be a string or a bool

jlewi:
Review status: 0 of 3 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

*[kubeflow/core/tests/util_test.jsonnet, line 3 at r2](https://reviewable.io:443/reviews/kubeflow/kubeflow/246#-L5RQBGi5L7wNpvhgG1z:-L5RQBGi5L7wNpvhgG2-:b-cko40x) ([raw file](https://github.com/kubeflow/kubeflow/blob/9d2b4a18d02bf3e69296593eeae79b9b9fb9f161/kubeflow/core/tests/util_test.jsonnet#L3)):*
> ```
> local util = import '../util.libsonnet';
> 
> std.assertEqual(util.upper("True"), "TRUE") &&
> ```

Nice!
Can you open up an issue to run the test as part of our E2E tests.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/246)*

PR #252, opened 2018-02-16
Update enable_iap.sh so that it updates the healthcheck path

ankushagarwal:
Update enable_iap.sh so that it updates the healthcheck path for envoy service to /healthz instead of the default /


jlewi:
nit: For future reference I really like PR descriptions that explain why we are making a change. In this explaining the fact that if we create the ingress before the backend the GCP health check won't be configured correctly.

docs/gke/enable_iap.sh
```diff
@@ -68,6 +68,15 @@ gcloud --project=${PROJECT} compute backend-services update ${BACKEND_SERVICE} \
       --global \
       --iap=enabled,oauth2-client-id=${CLIENT_ID},oauth2-client-secret=${CLIENT_SECRET}
 
+while [[ -z ${HEALTH_CHECK_URI} ]];
+  do HEALTH_CHECK_URI=$(gcloud compute --project=${PROJECT} health-checks list --filter=name~k8s-be-${NODE_PORT}- --uri);
+  echo "Waiting for the healthcheck resource PROJECT=${PROJECT} NODEPORT=${NODE_PORT} SERVICE=${SERVICE}...";
+  sleep 2;
+done
+
+# Update the healthcheck request path to "/healthz" since GCP load balancer defaults to "/"
```
jlewi: Please expand the comment to explain that this is to deal with the case in which ingress is created before the backing deployment the health check won't be set correctly.

ankushagarwal:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[docs/gke/enable_iap.sh, line 77 at r1](https://reviewable.io:443/reviews/kubeflow/kubeflow/252#-L5RARCiBIPrMy6TLBJw-r1-77:-L5U6nUeaSXj8e5KXhYO:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/4a3b401078ed79a0986dc0c6201eed4cc453902c/docs/gke/enable_iap.sh#L77)):*

Please expand the comment to explain that this is to deal with the case in which ingress is created before the backing deployment the health check won't be set correctly.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/252)*

PR #268, opened 2018-02-19
Add seldon ksonnet integration

### LINKED ISSUE #159: ksonnet component for Seldon.io deployment
We'd like to add a ksonnet component for Seldon.io's [CRD for model prediction](https://github.com/SeldonIO/seldon-core/blob/master/docs/crd/readme.md). 

This would enable Kubeflow users to serve all kinds of [models](https://github.com/SeldonIO/seldon-core).

The Deployment is way more powerful then that though since it allows prediction to be described as a graph of microservices.

To get started we need a componet like the one for [TFJob](https://github.com/google/kubeflow/blob/master/kubeflow/core/tf-job.libsonnet) to deploy the Seldon.io CRD.

An open question is whether we should also deploy the Seldon API front end which does oauth or we should rely on the generic reverse proxy solution we are building in Kubeflow (#11)

It would be great if after deploying Kubeflow, users could just load up the [Seldon notebooks](https://github.com/SeldonIO/seldon-core/blob/master/notebooks/kubectl_demo_gcp.ipynb) and run them.

@cliveseldon @ahousley

cliveseldon:
Fixes #159 

 * Called seldon rather than seldon-core not to overlap with existing core prototype
 * Raw JSON is keep in separate folder and then ksonnet Mixins used for construction.

Known issues

 * RBAC setting is clusterwide admin at present. Not done work to define a more limited defn.
 * Not sure if there is a style guide for jsonnet files but I'm using emacs and didn't find a nice mode to format them.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/268)

### DIFF
--- a/kubeflow/seldon/core.libsonnet
+++ b/kubeflow/seldon/core.libsonnet
@@ -1,0 +1,96 @@
+local k = import 'k.libsonnet';
+local deployment = k.extensions.v1beta1.deployment;
+local container = k.apps.v1beta1.deployment.mixin.spec.template.spec.containersType;
+local service = k.core.v1.service.mixin;
+local serviceAccountMixin = k.core.v1.serviceAccount.mixin;
+local clusterRoleBindingMixin = k.rbac.v1beta1.clusterRoleBinding.mixin;
+local clusterRoleBinding = k.rbac.v1beta1.clusterRoleBinding;
+local serviceAccount = k.core.v1.serviceAccount;
+local baseApife = import 'json/apife-deployment.json';
+local apifeService = import 'json/apife-service.json';
+local operatorDeployment = import 'json/operator-deployment.json';
+local redisDeployment = import 'json/redis-deployment.json';
+local redisService = import 'json/redis-service.json';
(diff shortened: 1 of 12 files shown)


jlewi:
LGTM. 

@cliveseldon Is there anyone else that should review this before I merge it?

Should we open up an issue to add E2E tests?

cliveseldon:
* People at Seldon have reviewed, not sure who outside seldon.
 * Yes e2e test issue would be good. Do you have any docs or existing examples? I assume we would need to set up the test infra locally. I couldn't find docs to prow?

Next steps after this is merged is to change the example-seldon docs to reference kubeflow setup and this. Plus to e2e tests.

jlewi:
Our current tests are here
https://github.com/kubeflow/kubeflow/tree/master/testing

Our E2E tests are written as Argo workflows where each step generally runs a python script.

So the first step would be to deploy Seldon core and verify that all the containers etc... start.
You could then try deploying an actual model and sending predictions to it.

You could either create a new workflow or add it to our existing E2E workflow.

PR #270, opened 2018-02-20
Add mount of pvc

### LINKED ISSUE #19: Fault tolerant storage for Jupterhub
Jupyter pods are storing data in the pod volume. So if the pod dies you would lose any notebook/file edits.

We should be using a fault tolerant volume so that if the pod dies we don't lose our data.

/cc @foxish

inc0:
While we were adding pvc for every jupyter instance, we didn't mount it
anywhere.
Let's mount it to work dir, as I assume this is dir where users will
likely put their notebooks. This will ensure that work will be retained
even if pod dies.

Fix #19
Fix #22 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/270)

### DIFF
--- a/kubeflow/core/jupyterhub_spawner.py
+++ b/kubeflow/core/jupyterhub_spawner.py
@@ -91,3 +91,18 @@ c.KubeSpawner.user_storage_pvc_ensure = True
 # How much disk space do we want?
 c.KubeSpawner.user_storage_capacity = '10Gi'
 c.KubeSpawner.pvc_name_template = 'claim-{username}{servername}'
+c.KubeSpawner.volumes = [
+  {
+    'name': 'volume-{username}{servername}',
+    'persistentVolumeClaim': {
+      'claimName': 'claim-{username}{servername}'
+    }
+  }
+]
+c.KubeSpawner.volume_mounts = [
+  {
+    'mountPath': '/home/jovyan/work',
+    'name': 'volume-{username}{servername}'
+  }
+]
+


kubeflow/core/jupyterhub_spawner.py
```diff
@@ -91,3 +91,18 @@ def extra_resource_limits(self):
...
+  {
+    'name': 'volume-{username}{servername}',
+    'persistentVolumeClaim': {
+      'claimName': 'claim-{username}{servername}'
+    }
+  }
+]
+c.KubeSpawner.volume_mounts = [
+  {
+    'mountPath': '/home/jovyan/work',
```
jlewi: Should everything in /home be on the PVC?

kubeflow/core/jupyterhub_spawner.py
```diff
@@ -91,3 +91,18 @@ def extra_resource_limits(self):
 # How much disk space do we want?
```
jlewi: Do we need to uncomment the line above that sets singleuser_fs_gid
https://github.com/kubeflow/kubeflow/pull/22#issuecomment-350500944

Did you test if this works?

kubeflow/core/jupyterhub_spawner.py
inc0: well we have tf examples there too, which I think is a nice touch, and that would disappear.

jlewi: What about about putting it under, /mnt ? I'd like to make it more obvious which directories correspond to PD and which don't.

inc0: jupyter tree lands in /home/jovyan, so if we put it in /mnt it won't be obvious how to use it by users.

jlewi: Ok. Please address other comments.

kubeflow/core/jupyterhub_spawner.py
inc0: yup, just tested more thoroughly:
1. created notebook in work dir
2. deleted jupyter pod
3. recreated pod
4. notebook was there

inc0: AD. singleuser_fs_gid - just checked - we run container as root. That makes it work without gid gimmicks but it's different issue on it's own:)

jlewi:
@inc0 please open an issue to add an E2E test to verify that pods actually mount the PVC.

It looks like there was a regression when we switched from YAML to ksonnet. #22 had added this feature to our manifests. So it would be good to have an E2E test to prevent further regressions.

PR #271, opened 2018-02-20
add prow_config.yaml

lluunn:
prow_config.yaml contains the workflows that we want to run by prow.

By using prow_config.yaml, we can use the same testing image for different kubeflow repos.

Related to #276 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/271)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -1,0 +1,10 @@
+# This file configures the workflows to trigger in our Prow jobs.
+# see kubeflow/testing/py/run_e2e_workflow.py
+workflows:
+  - app_dir: kubeflow/kubeflow/testing/workflows
+    component: workflows
+    name: kubeflow-e2e
+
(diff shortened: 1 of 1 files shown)


jlewi:
The test didn't succeed.
Also it looks like the test is only running a single workflow; should it be running two workflows?

jlewi:
You can ignore the test failure that should be fixed #274 if you sync.

jlewi:
Looks like a timeout. Maybe you need to increase it?

lluunn:
Something is wrong. It stuck at checkout for >11hrs.
http://testing-argo.kubeflow.io/timeline/kubeflow-test-infra/kubeflow-presubmit-tf-serving-image-271-152c3d9-294-a16f;tab=workflow

lluunn:
@jlewi The test passed with just one workflow.
I think the TF serving image workflow needs some fix to integrate with prow. I will do that in another PR.
PTAL, thanks

PR #279, opened 2018-02-22
Drop imagePullPolicy override

kflynn:
`imagePullPolicy: Always` isn't needed for Ambassador these days, and the default `ifNotPresent` is really much easier on everyone's network. We've taken this out of Datawire's default Ambassador YAML, and recommend doing the same for Kubeflow. Thanks!

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/279)


kflynn:
(I think that Datawire should have a company CLA already -- sing out if we need to do anything else!)

jlewi:
@kflynn Thank you for the fix. Would you mind following the instructions of the Google Bot and signing it individually. I think you need to do that so that the tooling can verify its signed.

kflynn:
Signed the CLA.

PR #281, opened 2018-02-22
Integrate tf serving image building with prow

lluunn:
pre/post-submit will also build / push / test TF serving image using the workflow in 
components/k8s-model-server/images/releaser

The image name is gcr.io/mlkube-testing/model-server, and tag is the workflow name.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/281)


lluunn:
The e2e currently failed to deploy TF serving component:

Error updating services kubeflow-presubmit-tf-serving-image-281-568560e-316-1ed6.inception: Service "inception" is invalid: [spec.ports[0].n
ame: Required value, spec.ports[1].name: Required value]

It should be fixed by #283

jlewi:
@lluunn #283 Should be fixed can you sync and rerun the tests?

lluunn:
It's now failing with: Timeout waiting for deployment inception.

I ran into this before using my own cluster, and the reason was not enough cpu. Not sure if it's the same here. TF serving component should need 4 cores.

lluunn:
Failed to pull image "gcr.io/kubeflow-ci/model-server:kubeflow-presubmit-tf-serving-image-281-855ce2a-325-d917": rpc error: code = Unknown desc = Error response from daemon: repository gcr.io/kubeflow-ci/model-server not found: does not exist or no pull access

lluunn:
Failure reason:
The workflow is running in mlkube-testing cluster, and the first step it built and pushed an TF serving image to gcr.io/kubeflow-ci (succeeded). 
The next step it deployed a tf-serving component (in the same mlkube-testing cluster) using the new image, and it failed because it couldn't pull the image from kubeflow-ci.

lluunn:
The test passed, PTAL @jlewi

jlewi:
It doesn't look like you are creating juni files 
http://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/281/kubeflow-presubmit/342/artifacts/

Prow relies on junit files to actually indicate that a test succeeded. 

The way this works is
1. Your test should use the following utility to create a junit xml file in the NFS share
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/test_util.py#L139

2. Your pipeline should have a copy artifacts step to copy the artifacts to GCS.

Its probably best to do this in a follow on PR as this already provides a lot of useful functionality.

testing/test_tf_serving.py
```diff
@@ -17,6 +17,7 @@
 from __future__ import print_function
 
 import logging
+import time
```
jlewi: So this file is the one that would probably create the junit file.
See what we do in test_runner.py
https://github.com/kubeflow/tf-operator/blob/master/py/test_runner.py#L140

lluunn: I think we can merge this one first if looks good, so we can have the test up and running now.
I will work on adding junit soon.

lluunn: Actually, I think it's small change. I will do it in this PR now.

lluunn:
Passing with 5 tests:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/281/kubeflow-presubmit/347/
PTAL @jlewi

testing/test_tf_serving.py
```diff
@@ -64,27 +66,55 @@ def main():
...
+    num_try = 1
+    result = None
+    while True:
+      try:
+        result = str(stub.Predict(request, 10.0))  # 10 secs timeout
+      except Exception as e:
+        num_try += 1
+        if num_try > 3:
+          raise
+        print('prediction failed: {}. Retrying...'.format(e))
```
jlewi: nit: Lets use logging not print.

lluunn:
Review status: 0 of 5 files reviewed at latest revision, 2 unresolved discussions.

---

*[testing/test_tf_serving.py, line 99 at r9](https://reviewable.io:443/reviews/kubeflow/kubeflow/281#-L63RhxzEHxKb9fYoRam-r9-99:-L68nH8L2hoWACeQtMA6:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/e0fae2d8f23ddb6b2c268cfcce62b3441e27527b/testing/test_tf_serving.py#L99)):*

nit: Lets use logging not print.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/281)*

jlewi:
:lgtm:

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/281#-:-L68vjRQLWpLi5A2zL5n:bnfp4nl)*

jlewi:
Reviewed 1 of 3 files at r1, 1 of 1 files at r2, 1 of 3 files at r8, 1 of 1 files at r9, 1 of 1 files at r10.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/281)*

PR #283, opened 2018-02-23
Add names to tfserving service ports

ankushagarwal:
Kubernetes requires port names for services with
more than one port

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/283)


ankushagarwal:
cc @yupbank

PR #287, opened 2018-02-23
Remove unused file build/check_errorf.sh

ankushagarwal:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/287)


ankushagarwal:
cc @vishh - since you added this file

jlewi:
Looks like some sort of scrip to checks formatting of errors in go. But this repository doesn't have any go code. So I don't think its all that relevant. For tf-operator we are using a go linter.

Issue #295, opened 2018-02-25
Add an option to set service type to TFServing deployment


jlewi:
There should be an option to easily set the service type for TFServing.

This way users can easily expose it externally using a load balancer, or use a type of node port.

fxue:
[PR 302](https://github.com/kubeflow/kubeflow/pull/302) was submitted for this issue.

jlewi:
Fixed by #302.

PR #302, opened 2018-02-27
add option service_type to set TFServing service type

fxue:
#295 
tf-serving : added optional parameter "service_type" default to clusterIP.
this parameter is then passed to modelService().

This can be tested as follows:
```BASH
ks param set ${MODEL_COMPONENT} service_type LoadBalancer
ks apply ${KF_ENV} -c ${MODEL_COMPONENT}
```
should see your serving component is now of type LoadBalancer instead of ClusterIP
```
NAME         TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)                         AGE
inception1   LoadBalancer   10.59.249.163   *.*.*.*   9000:31346/TCP,8000:32337/TCP   14m
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/302)


jlewi:
@fxue please sync

Issue #266, opened 2018-02-19
Better identity management in K8s


chrisheecho:
TBD

jlewi:
Can you provide more detail please? What are the current problems limitations? And why are these Kubeflow issues and not K8s issues?

jlewi:
I'm closing this in favor of existing issues that are more specific.
If there's something not covered by existing issues lets open up more specific ones.

#372 Pattern for creating/using secrets across multiple prototypes
#101 Lock down Jupyter pods.

kkasravi:
This may be a duplicate of #11 since auth is part of that issue (secure proxy)

jlewi:
@kkasravi Can we make this issue or a new issue more specific?

The current state of the world is we have 

- IAP which will provide a google identity to JupyterHub
- An http basic auth module (see #11) for Ambassador.

So was the intent to provide an auth module for ambassador that would use other oauth providers like GitHub and then supply that id to JupyterHub?

PR #305, opened 2018-02-27
[WIP] Issue 239 ambassador as proxy for jupyterhub

kkasravi:
This is a fix for #239 and is still a WIP - but I thought I would make the PR available for review

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/305)


jlewi:
@kkasravi Thanks for doing this. I've added WIP to the title just to make it easier to track. When its ready for review just update the title.

kkasravi:
I signed it!

kubeflow/core/ambassador.libsonnet
```diff
@@ -170,6 +170,10 @@
                       },
                     },
                   },
+                  {
+                    name: "AMBASSADOR_SINGLE_NAMESPACE",
```
jlewi: What's this do? Does this mean Ambassador only handles services in this namespace?
If that's so then we should probably delete the service we created to create a forwarding rule to the K8s dashboard.

kubeflow/core/jupyterhub.libsonnet
```diff
@@ -116,20 +116,30 @@ c.RemoteUserAuthenticator.header_name = 'x-goog-authenticated-user-email'",
...
+              "rewrite: /hub/",
+              "service: tf-hub-0." + namespace,
+            ]),
+        },  //annotations
       },
       spec: {
-        // We want a headless service so we set the ClusterIP to be None.
-        // This headless server is used by individual Jupyter pods to connect back to the Hub.
-        clusterIP: "None",
         ports: [
```
jlewi: Why are you making this a non-headless service?

kubeflow/registry.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: '0.1'
 kind: ksonnet.io/registry
 libraries:
   core:
-    version: master
+    version: issue_239_ambassador_as_proxy_for_jupyterhub
```
jlewi: FWIW this is what I do for debugging
https://github.com/kubeflow/kubeflow/blob/master/developer_guide.md

jlewi:
@kkasravi Could you provide an update on this PR?  Is it working? Is there anything blocking it?

kkasravi:
Update - We're finally going to test with our own cluster and going thru the iap instructions/setup. We didn't have permissions internally to do this before. I was still having some problems with jupyterhub using ambassador's /user/<uname> proxy as an external proxy that was created dynamically. I might close this PR since a number of changes have been moved in under other PRs. I'll open a new one with fewer changes. In any case I'll update by mid-week on how this is going.

jlewi:
Great thanks.
You could also just rebase this PR to avoid changes that have already been merged.

kkasravi:
@jlewi I created an IAP enabled cluster and integration does disable jupyterhub's default proxy (configurablehttpproxy) and moves the proxy to envoy.  Jupyterhub will use the authentication established by IAP and bring up a notebook w/o challenging the user.  So, I would recommend that both this PR and the related issue be closed. Most of the changes in this PR have been integrated in other PRs. General identity management (for example using github oauth2 instead of IAP) is an open issue #266 which I've been working on and will pickup there.

kkasravi:
Per my above comments (IAP does proxy requests which allow jupyterhub to use it's auth and bypass it's dummyauthenticator) I"m closing this PR.

Issue #300, opened 2018-02-26
Configure JupyterHub spawner to allow root operations by default


cwbeitel:
Regarding notebook server spawner options https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/jupyterhub_spawner.py#L79 currently c.KubeSpawner.args is being provided the arg"--allow-root" presumably because the original implementer meant for users to have root access in containers. But currently this is not available. 

By adding "GRANT_SUDO=yes" to the environment users do have sudo privs in spawned notebooks, e.g. can successfully run `sudo apt-get update`:

```python 
###################################################
# Spawner Options
###################################################
c.JupyterHub.spawner_class = KubeFormSpawner
c.KubeSpawner.singleuser_image_spec = 'gcr.io/kubeflow/tensorflow-notebook'
c.KubeSpawner.cmd = 'start-singleuser.sh'
c.KubeSpawner.args = ['--allow-root']
# First pulls can be really slow, so let's give it a big timeout
c.KubeSpawner.start_timeout = 60 * 10
c.KubeSpawner.environment = {
    "GRANT_SUDO": "yes"
}
```

If making this addition should determine if there is anything in c.KubeSpawner.environment to over-write and if so append the variable instead of setting all.

Related to https://github.com/kubeflow/kubeflow/issues/289

jlewi:
Seems useful as this would also allow people to install packages e.g. `via apt-get`.

Any info on what the security implications are?

cwbeitel:
I agree that seems worth considering and potentially problematic but I don't have any expert understanding in that area to contribute.

From my cursory reading it seems that running containers with root credentials makes it more likely that a kernel vulnerability could be exploited to gain access to the node running that container, given that with Docker the kernels are shared?

jlewi:
I think the use cases of 
- Mounting NFS volumes
- Installing additional packages
- Image building

Are convincing enough that I think its worth doing until we have more reason to believe its unsafe and should disable it.

aronchick:
cc @mayakacz

mayakacz:
Containers that don't need to run as root shouldn't run as root. See for example: https://docs.docker.com/engine/security/security/#linux-kernel-capabilities

My concern is the earlier comment, "its worth doing until we have more reason to believe its unsafe" - I strongly disagree. Making this kind of change later would be breaking, and it'll be painful to get users to migrate from an unsafe state to a safe state. If we can prevent this initially, that is preferred.

[...]

Issue #101, opened 2018-01-07
Lock down jupyter pods


jlewi:
What would it take to lock down individual Jupyter pods so that only the owner has access?

I think the use case would be you have multiple data scientists sharing a cluster. Each data scientist has access to slightly different datasets. So they want to use their own credentials (and not a shared service account). So we'd like the pods to be sufficiently locked down that they feel comfortable putting their credentials in the pod (e.g. `gcloud auth login`) and that except for a cluster admin, no one will be able to access the pod and grab their credentials. 

I think this might look like the following

- Use a sidecar in the Jupyter pod to check the identity of the request (e.g. by looking at the JWT) and only accept requests from  the owner of that server
- Use RBAC to lock down the pod.

Background on running jupyter notebook containers as root vs. non root
  - Initially Jupyter didn't run as root and this caused problems when attaching extra volumes see https://github.com/kubeflow/kubeflow/pull/22#issuecomment-350500944
  - @willb  pointed out that running containers as root doesn't work with OpenShift

yuvipanda:
The jupyter pods are already are authenticated with internal OAuth between
the Hub process and the user pods, so other users on the network shouldn't
be able to access the notebook pod without authentication.

We also default to disabling service account mounting in the spawned user
pod to prevent users from accessing the kubernetes API. I agree that we
should have easier ways to provide rolebindings for each user's pod if we
wanna allow access to the k8s API - there's ongoing work in
https://github.com/jupyterhub/kubespawner/pull/115 to make that possible.
Me and @foxish also discussed this in person at kubecon and believe we have
a reasonable way forward.

jlewi:
I think we will definitely need access to the K8s API because we want to allow users to create other K8s resources (e.g. TfJobs) from their notebooks. Although arguably they should be using user credentials for this and not a service account.

yuvipanda:
https://github.com/jupyterhub/kubespawner/issues/110 is partially for dealing with directly mounting OAuth credentials and what not from your login provider into the pod as a secret, which might help (if you are using google auth to log in too)

We don't mount any service accounts by default, but you can certainly enable mounting :)

We'll probably also add NetworkPolicy support to the helm chart at some point, locking down traffic to only be along the paths necessary by default. Specifically, only allow traffic between hub/proxy and pods, and disallow it between

jlewi:
[...]

PR #311, opened 2018-02-27
Fix #285 and make few tweaks to notebook Dockerfile

### LINKED ISSUE #285: How to add persistent volume in jupyter_spawner.py file ?
Hi,

I have followed these steps https://github.com/kubeflow/kubeflow/blob/master/user_guide.md to deploy kubeflow.
I'm able to see Jupyter notebook interface after logging into the jupyter hub.
now to add persistent volume if I follow the steps mentioned here https://github.com/kubeflow/kubeflow/pull/22/files/48ffe99d30eb31abc4d54c4553483b6c5b2034bc and make these changes in jupyter_spawner.py file.
tf-hub goes into CrashLoopBackOff error saying
IndentationError: unexpected indent

Can you suggest some solution to this to add persistent volume ?

inc0:
Besides adding chown /home/$USER also move non-install layers lower in
Dockerfile to make better use of docker cache during edits
Later we will probably switch to root user (issue #300), so it will make this whole section redundant.
Aslo touches #101 

Fixes #285

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/311)

### DIFF
--- a/components/tensorflow-notebook-image/start.sh
+++ b/components/tensorflow-notebook-image/start.sh
@@ -27,10 +27,12 @@ if [ $(id -u) == 0 ] ; then
     fi
 
     # Ensure pv directory belongs to user
-    if [ -d "/home/$NB_USER" ]; then
-        chown -R $NB_USER /home/$NB_USER
-    fi
-
+    for d in "$CONDA_DIR" "$JULIA_PKGDIR" "/home/$NB_USER"; do
+        if [[ ! -z "$d" && -d "$d" ]]; then
+            echo "Set ownership to uid $NB_UID: $d"
+            chown -R $NB_UID "$d"
+        fi
+    done
     # Change GID of NB_USER to NB_GID if NB_GID is passed as a parameter
     if [ "$NB_GID" ] ; then
         echo "Change GID to $NB_GID"
--- a/components/tensorflow-notebook-image/start.sh
+++ b/components/tensorflow-notebook-image/start.sh
@@ -24,14 +24,11 @@ if [ $(id -u) == 0 ] ; then
     if [ "$NB_UID" != $(id -u $NB_USER) ] ; then
         echo "Set user UID to: $NB_UID"
         usermod -u $NB_UID $NB_USER
-        # Careful: $HOME might resolve to /root depending on how the
-        # container is started. Use the $NB_USER home path explicitly.
-        for d in "$CONDA_DIR" "$JULIA_PKGDIR" "/home/$NB_USER"; do
-            if [[ ! -z "$d" && -d "$d" ]]; then
-                echo "Set ownership to uid $NB_UID: $d"
-                chown -R $NB_UID "$d"
-            fi
-        done
+    fi
+
+    # Ensure pv directory belongs to user
+    if [ -d "/home/$NB_USER" ]; then
+        chown -R $NB_USER /home/$NB_USER
     fi
 
     # Change GID of NB_USER to NB_GID if NB_GID is passed as a parameter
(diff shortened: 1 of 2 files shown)


inc0:
note @jlewi - we'll need to re-push images to hub after it merges

components/tensorflow-notebook-image/start.sh
```diff
@@ -24,14 +24,11 @@ if [ $(id -u) == 0 ] ; then
...
-        # container is started. Use the $NB_USER home path explicitly.
-        for d in "$CONDA_DIR" "$JULIA_PKGDIR" "/home/$NB_USER"; do
-            if [[ ! -z "$d" && -d "$d" ]]; then
-                echo "Set ownership to uid $NB_UID: $d"
-                chown -R $NB_UID "$d"
-            fi
-        done
+    fi
+
+    # Ensure pv directory belongs to user
```
jlewi: Why are you no longer changing the ownership of the CONDA_DIR and JULIA_PKG_DIR.

Can you explain what the new code does differently? It looks like the only difference is that instead of doing chown with the UID we use the username now.

Do you know why that makes a difference?

inc0: Difference is that, instead of changing only CONDA_DIR and JULIA_PKG_DIR (both of which lives in /home/$NB_USER) we chown whole /home/$NB_USER - most importantly /home/$NB_USER/work

Because we change user here in the script (line 48) and not in kubernetes manifest, PV isn't connected with correct user ownership, it assumes root (if you look at our notebook pod manifesto, it explicitly runs container as UID 0). I'm doing chown in startup script instead.

jlewi: Do you know why run the container as UID 0? Should we not be doing that?

The original code was 

```
for d in "$CONDA_DIR" "$JULIA_PKGDIR" "/home/$NB_USER"; do
```

So it chown'd the whole /home/$NB_USER dir.

It looks like the difference is that you are now doing this always as opposed to

```
if [ "$NB_UID" != $(id -u $NB_USER) ]
```
What I'm inferring from this is that NB_UID = id NB_USER therefore in the current startup script we are not executing the chown. 

However, for some reason the permsions on /home/${NB_USER} are not set correctly to allow that user to use that container.

If UID is 0, doesn't that mean we are running as root?

inc0: Ah you're right, I'll restore CONDA_DIR etc, issue was "if" above.

I think we shouldn't run it as root, maybe ensure perms for these dirs in Dockerfile. I'll open separate issue for that one tho as someone did that on purpose:)

Yeah we run pod as root, but start.sh does su $NB_USER <<command>> so we run jupyter itself as NB_USER. Not a typical way to deal with it tbh.

inc0:
If we run as root ( #300 ) this issue wouldn't happen too

inc0:
Done, I linked #101 which has slack link in discussion

jlewi:
/assign @willb 

Adding Will because apparently running as root causes problems for open shift.

jlewi:
@inc0 clued me in

- The pod is already running with `runAsUser:0`
    Here's a [gist](https://gist.github.com/jlewi/457e80e153725bf91f7f6d4eb60a65b2) for a pod showing it to be true

- We can't  figure out where this is being set; it doesn't appear to be in our config and we didn't see it in [KubeSpawner](https://github.com/jupyterhub/kubespawner/blob/f3a263791423ecbc84f802ec065183dd1efa23f4/kubespawner/objects.py#L79)

- Nothing in this PR is changing how we run as root so it shouldn't have any impact on OpenShift (its possible things are already broken on OpenShift)

- All this PR does is chown the /home/$NB_USER  but this is only done if the user is root (its guarded by an if) so it shouldn't have any impact in the case where we don't run as root.

- If we set the uid/gid of the pod correctly then we think K8s will automatically set the permissions on the PV.

- So this PR fixes a bug with PVC not working and shouldn't have any other impact. So I'm merging it.

Issue #244, opened 2018-02-14
tf.transform libraries in our notebooks


jlewi:
tf.Transform is a library for feature processing.
https://github.com/tensorflow/transform

We should add it to our Jupyter notebook images so that people can run it.

I think it requires Beam so this probably means installing the Apache Beam library as well.

katsiapis:
Transform already specifies it's requirements https://github.com/tensorflow/transform/blob/master/setup.py#L24 (except for Tensorflow for now, so that it plays nicely with both CPU and GPU).

kkasravi:
@jlewi i would like to pick this up. I have a good understanding of how ksonnet is used in kubeflow and know beam by way of integrating a beam runner

jlewi:
Fantastic.

jlewi:
@kkasravi I sent you an invite to the Kubeflow org. If you accept you should be able to trigger tests and I think I can assign the issue to you.

kkasravi:
@jlewi done

Question: will we be replacing tf-job with argo's equivalent (but without the build steps)? Because I have most of this in place and argo is already on master.

jlewi:
I don't understand the question. TFJob and Argo serve very different purposes in my mind. TFJob is about executing TF jobs particularly distributed jobs. Argo is about managing workflows. So the overlap isn't obvious to me.

kkasravi:
@jlewi i was referring to the example-seldon repo which defines 4 workflows: 2 for training and 2 for serving. The TF training one uses the argo Workflow to first build and push the code and then run the TFJob (https://github.com/kubeflow/example-seldon/blob/master/workflows/training-tf-mnist-workflow.yaml#L62) unless you're thinking that both serving and training should be done using seldon's CRD.

jlewi:
So a workflow might look as follows

1. Build a Docker image that contains the code you want + any dependencies (e.g. TF.Transform)
1. Run a preprocessing job using TF.Transform
1. Run a TFJob to train a model

TF.Transform uses Apache Beam so in step 2 we would be invoking a python program that submits the beam job and waits for it to finish. Depending on the Beam runner we are using that job will do a variety of things.
  - If we are using the Dataflow runner on GCP then the job just makes an RPC to the Dataflow service which will start and manage a bunch of Dataflow workers
  - For the non-GCP case we can run it using the DirectPipelineRunner in which case it will run non-distributed in the pod created by Argo.
[...]

PR #314, opened 2018-03-01
Add support for Python 2 kernels to the tensorflow notebook images.

ojarjur:
The Dockerfiles for these images previously tried to install the
Python 2 kernel, but omitted the `source activate` step prior
to the install step.

The result was that rather than installing the Python 2 kernel
at the user level, it reinstalled the Python 3 kernel which simply
duplicated the existing, globall-installed kernel.

With this change "Python 2" now shows up as an option in the Jupyter
UI when creating a new notebook, and in an existing notebook you
can switch from a Python 3 to a Python 2 kernel using the "kernel"
menu.

This change is a prerequisite to addressing #244, as the tf.Transform
library relys on Apache Beam, which currently only supports Python 2.

After this change lands, we should be able to add tf.Transform to
the `ipykernel_py2` Conda environment, and then it will be usable
in Python 2 kernels.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/314)


jlewi:
@ojarjur Thanks for the quick fix!
Do you mind syncing and I'll merge it.
Note we'll need to build and push new images.

ojarjur:
@jlewi Synced.

PR #323, opened 2018-03-01
Adding S3 support to model serving.

elsonrodriguez:
This allows the model serving image to download from any S3 compatible API.

The pattern may also find its way into other components such as tf-job in order to facilitate reading/writing to s3 during training.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/323)

### DIFF
--- a/kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
+++ b/kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
@@ -8,6 +8,12 @@
 // @optionalParam model_server_image string gcr.io/kubeflow-images-staging/tf-model-server:v20180227-master Container for TF model server
 // @optionalParam http_proxy_image string gcr.io/kubeflow/http-proxy:1.0 Container for http_proxy of TF model server
 // @optionalParam service_type string ClusterIP The service type for TFServing deployment.
+// @optionalParam aws_access_key_id string key_id S3 access key id
+// @optionalParam aws_secret_access_key string secret_key S3 secret access key
+// @optionalParam aws_region string us-west-1 S3 Region
+// @optionalParam s3_use_https string true Whether or not to use https for S3 connections
+// @optionalParam s3_verify_ssl string true Whether or not to verify https certificates for S3 connections
+// @optionalParam s3_endpoint string http://s3.us-west-1.amazonaws.com URL for your s3-compatible endpoint.
 
 // TODO(https://github.com/ksonnet/ksonnet/issues/222): We have to add namespace as an explicit parameter
 // because ksonnet doesn't support inheriting it from the environment yet.
@@ -21,8 +27,26 @@ local modelPath = import "param://model_path";
 local modelServerImage = import "param://model_server_image";
 local httpProxyImage = import "param://http_proxy_image";
 local serviceType = import "param://service_type";
+local awsAccessKeyID = import "param://aws_access_key_id";
+local awsSecretAccessKey = import "param://aws_secret_access_key";
+local awsRegion = import "param://aws_region";
+local s3UseHTTPS = import "param://s3_use_https";
+local s3VerifySSL = import "param://s3_verify_ssl";
+local s3Endpoint = import "param://s3_endpoint";
+
+
+local modelServerEnv = [
+  { name: "AWS_ACCESS_KEY_ID", valueFrom: { secretKeyRef: { name: name, key: "aws-access-key-id" }, }, },
+  { name: "AWS_SECRET_ACCESS_KEY", valueFrom: { secretKeyRef: { name: name, key: "aws-secret-access-key" }, }, },
+  { name: "AWS_REGION", value: awsRegion },
+  { name: "S3_REGION", value: awsRegion },
+  { name: "S3_USE_HTTPS", value: s3UseHTTPS },
+  { name: "S3_VERIFY_SSL", value: s3VerifySSL },
+  { name: "S3_ENDPOINT", value: s3Endpoint },
+];
 
 std.prune(k.core.v1.list.new([
-  tfServing.parts.deployment.modelServer(name, namespace, modelPath, modelServerImage, httpProxyImage),
+  tfServing.parts.deployment.modelSecret(name, namespace, awsAccessKeyID, awsSecretAccessKey),
+  tfServing.parts.deployment.modelServer(name, namespace, modelPath, modelServerImage, modelServerEnv, httpProxyImage),
   tfServing.parts.deployment.modelService(name, namespace, serviceType),
 ]))
--- a/kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
+++ b/kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
@@ -8,12 +8,9 @@
 // @optionalParam model_server_image string gcr.io/kubeflow-images-staging/tf-model-server:v20180227-master Container for TF model server
 // @optionalParam http_proxy_image string gcr.io/kubeflow/http-proxy:1.0 Container for http_proxy of TF model server
 // @optionalParam service_type string ClusterIP The service type for TFServing deployment.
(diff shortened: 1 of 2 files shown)


kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
```diff
@@ -8,6 +8,12 @@
 // @optionalParam model_server_image string gcr.io/kubeflow-images-staging/tf-model-server:v20180227-master Container for TF model server
 // @optionalParam http_proxy_image string gcr.io/kubeflow/http-proxy:1.0 Container for http_proxy of TF model server
 // @optionalParam service_type string ClusterIP The service type for TFServing deployment.
+// @optionalParam aws_access_key_id string key_id S3 access key id
+// @optionalParam aws_secret_access_key string secret_key S3 secret access key
```
jlewi: Is the access_key something you would want to check into source control?

elsonrodriguez: That's up to the user. However they store it, it can be passed in as a param to `ks generate`

jlewi: ksonnet apps are meant to be checked into source control. This means users should generally not store sensitive information in the ksonnet app e.g passwords.

So if aws_secret_access_key is a password then it shouldn't be a parameter because that would mean it gets checked into source control.

In this case a better approach would be to take as a parameter the name of a K8s secret which stores the password.

jlewi:
An alternative to adding a bunch of AWS specific parameters would be to create a new prototype that is AWS specific and possibly even AWSS3 specific.

I'm fine with the current PR but I don't think adding parameters for different use cases will ultimately scale. That will lead us to have 10's of disjoint parameters for different deployments (AWS, GCP, Canonical, Azure etc...). This defeats a lot of the value of having parameters because users can't use help to get a quick list of which parameters they might want to set.

Prototypes are one solution. Although even that approach could eventually leads to 10's of prototypes to choose from which could be equally problematic.

Another approach is to make it easy for users to lever jsonnet to customize the things they care about; but this would entail modifying jsonnnet files directly which probably isn't ideal for simple cases.

I don't think we should try to resolve this in this PR. I just want to start the discussion.

kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
inc0: well this ks will generate secret out of it too, not sure which is better - require secret pre-created or create from params here? In helm it would be the latter I guess.

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -105,7 +118,7 @@ local networkSpec = networkPolicy.mixin.spec;
                       "--model_name=" + name,
                       "--model_base_path=" + modelPath,
```
inc0: soo...how will modelPath work with S3? That was what I meant on Slack - do you provide model_path like idk https://path-to-my-object? How does that work?

elsonrodriguez: s3://s3-bucket/model-path

kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
elsonrodriguez: From a user perspective it never even occurred to me that you were supposed to check the output of ksonnet in somewhere. In my case I'm using ksonnet within a container in an argo workflow to deploy a model for serving. The ksonnet application just evaporates leaving the model running in k8s.

I also think it's better UX to be able to provide credentials via ks. This way you avoid having someone mess with kubectl directly in the middle of their workflow (whether automated or not)

As far as checking secrets in to git, I think it's fine as long as something like git-crypt is used. I can add a disclaimer advising such protections for those wanting to check in the output of ks.

elsonrodriguez:
Yeah it's a tough problem. I think we're just going to have to follow our gut on when it's time to split something apart. In this case I didn't feel it was good to make a new prototype because the S3 parameters are simply not used unless an s3 url is passed in. For that same reason I figured it was better for readability to not provide a conditional statement on whether or not to pass in the parameters.

If it were the case that the parameters interfered with any existing functionality then a strong case can be made for providing a conditional, and a conversation could be started about splitting out another prototype.

Something I can do here is group the parameters better and change the parameter descriptions so that the user understands that these are optional, and that they are for S3, and that they must all be set together.

elsonrodriguez:
I can't pull the image from the presubmit test down to try and replicate the error, anyone got any debugging tips?

kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
jlewi: Whether one is using ksonnet or raw yaml, I think one of the points of describing infrastructure decoratively is so that you can check it into version control. 

So I really don't think we should structure our configs in a way that would lead to secrets being stored in version control.

I think the long term solution here is that K8s will support external key management systems. So that you can tell K8s to create a secret from some key in your KMS and as part of deploying the infrastructure K8s will check whether you have permission to access the key and fetch it. T

The point is our ksonnet/yaml configs should never contain any data that is sensitive and should be encrypted; only unsensitive references (e.g. the name of a key, name of a file etc)...

So I don't think we should put secrets into our ksonnet configs.

I think the general pattern that we should encourage users to follow is you create a secret outside of ksonnet and then pass the name of the secret as a parameter.

An example is our test infrastructure
https://github.com/kubeflow/testing#create-a-gcp-service-account

jlewi:
@elsonrodriguez My suggestion would be to run the Argo workflow on your own cluster and comment out "onExit" so that you don't tear down the model. This will allow you to see what's going on and debug.

I can give you access to our new kubeflow-ci cluster if need be. We aren't using that as part of our continuous tests but its mostly setup. If you wanted to you could update the config to run on that cluster and then submit that as a PR.

kubeflow/tf-serving/tf-serving.libsonnet
inc0: How does that work? I mean model-path is a directory. Moreover it's directory where you're supposed to have multiple versions of models - as in it'll look for dirs 1/<< model files>> 2/<< model files >> etc. Your approach will mean you'll have to have all the versions of models in same object, which is...not ideal.

kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
```diff
@@ -8,6 +8,16 @@
 // @optionalParam model_server_image string gcr.io/kubeflow-images-staging/tf-model-server:v20180227-master Container for TF model server
 // @optionalParam http_proxy_image string gcr.io/kubeflow/http-proxy:1.0 Container for http_proxy of TF model server
 // @optionalParam service_type string ClusterIP The service type for TFServing deployment.
+// @optionalParam s3_create_secret string true Whether or not to create a secret containing aws-access-key-id and aws-secret-access-key. Otherwise will use provided secret.
```
jlewi: Can you open up an issue to figure out a better long term solution? e.g. making it its own prototype?

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -23,6 +23,20 @@ local networkSpec = networkPolicy.mixin.spec;
         },
       },
 
+      modelSecret(name, namespace, awsAccessKeyID, awsSecretAccessKey, labels={ app: name}): {
```
jlewi: I don't think we should create secrets in our ksonnet configs. The expectation of a secret is that its sensitive data.
So I'm assuming awsSecretAccessKey is a token/password that shouldn't be widely shared.

But the expectation is that deployment/ksonnet configs can be checked into source control.

If the data isn't sensitive why is it a secret and not a config map?

kubeflow/tf-serving/prototypes/tf-serving-all-features.jsonnet
elsonrodriguez: Opened #372

PR #334, opened 2018-03-02
Add signature support to http proxy

lluunn:
- Fix format (indent) for server.py
- Add signature support to server.py
- Add mnist test file

Verified mnist input works with http proxy and the model at gs://kubeflow-ci-test-models/mnist

We need the input dtype in signature for mnist because TF prefers int32 when casting: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/framework/tensor_util.py#L449

So there will be type mismatch if the model is expecting int64.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/334)


components/k8s-model-server/http-proxy/server.py
```diff
@@ -42,115 +44,179 @@
...
-            request.inputs[input_column].CopyFrom(tf.make_tensor_proto(values, shape=[len(values)]))
-
-        stub = self.settings['stub']
-        result = yield fwrap(stub.Predict.future(request, self.settings['rpc_timeout']))
-        output_keys = result.outputs.keys()
-        predictions = zip(*[tf.make_ndarray(result.outputs[output_key]).tolist() for output_key in output_keys])
-        predictions = [dict(zip(*t)) for t in zip(repeat(output_keys), predictions)]
-        self.write(dict(predictions=predictions))
+  @gen.coroutine
+  def post(self, model_name, version_name=None):
```
yupbank: tensorflow/serving support multi-model deployment. so we want something like this ?
```
if not self.settings['signature_map'].get(model_name):
    self.settings['signature_map'][mode_name] = get_signature_map(self.settings['stub'], model_name)
```

components/k8s-model-server/http-proxy/server.py
```diff
@@ -42,115 +44,179 @@
...
+      return {k: decode_b64_if_needed(v) for k, v in data.iteritems()}
+  else:
+    return data
+
+def get_signature_map(model_server_stub, model_name):
+  request = get_model_metadata_pb2.GetModelMetadataRequest()
+  request.model_spec.name = model_name
+  request.metadata_field.append("signature_def")
+  try:
+    response = model_server_stub.GetModelMetadata(request, 20)
```
yupbank: some constant names ?

components/k8s-model-server/http-proxy/server.py
lluunn: Done

components/k8s-model-server/http-proxy/server.py
lluunn: Done.

jlewi:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[components/k8s-model-server/http-proxy/server.py, line 76 at r2](https://reviewable.io:443/reviews/kubeflow/kubeflow/334#-L6dUaFYEpOWoPH9zQ_-:-L6dUaFYEpOWoPH9zQ_0:bv5p0ge) ([raw file](https://github.com/kubeflow/kubeflow/blob/29e313baa860e34a94f7803244cc426919e59445/components/k8s-model-server/http-proxy/server.py#L76)):*
> ```Python
> def decode_b64_if_needed(data):
>   if isinstance(data, list):
>     return [decode_b64_if_needed(val) for val in data]
> ```

Do we have unittests to cover these different use cases?

---

*[components/k8s-model-server/http-proxy/server.py, line 85 at r2](https://reviewable.io:443/reviews/kubeflow/kubeflow/334#-L6dUlPCCl14jQ5rIXni:-L6dUlPCCl14jQ5rIXnj:bm2z3ri) ([raw file](https://github.com/kubeflow/kubeflow/blob/29e313baa860e34a94f7803244cc426919e59445/components/k8s-model-server/http-proxy/server.py#L85)):*
> ```Python
>     return data
> 
> def get_signature_map(model_server_stub, model_name):
> ```

Doc string please. What does this method do?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/334)*

components/k8s-model-server/http-proxy/server.py
```diff
@@ -38,119 +40,184 @@
...
+
+  signature_def_map_proto = get_model_metadata_pb2.SignatureDefMap()
+  response.metadata["signature_def"].Unpack(signature_def_map_proto)
+  signature_def_map = signature_def_map_proto.signature_def
+  if not signature_def_map:
+    logging.error("Graph has no signatures.")
+
+  # Delete incomplete signatures without input dtypes.
+  invalid_signatures = []
+  for signature_name in signature_def_map:
```
yupbank: would this fail tensorflow/serving already ?

components/k8s-model-server/http-proxy/server.py
```diff
@@ -38,119 +40,184 @@
...
+    request_data = tornado.escape.json_decode(self.request.body)
+    instances = request_data.get(request_key)
+    if not instances:
+      self.send_error('Request json object have to use the key: %s' % request_key)
+    if len(instances) < 1 or not isinstance(instances, (list, tuple)):
+      self.send_error('Request instances object have to use be a list')
+    instances = decode_b64_if_needed(instances)
+    input_columns = instances[0].keys()
+
+    signature_name = request_data.get("signature_name", None)
```
yupbank: `signature_name = request_data.get("signature_name") ` default value of get is already None

components/k8s-model-server/http-proxy/server.py
```diff
@@ -38,119 +40,184 @@
...
+
+    request_key = self.settings['request_key']
+    request_data = tornado.escape.json_decode(self.request.body)
+    instances = request_data.get(request_key)
+    if not instances:
+      self.send_error('Request json object have to use the key: %s' % request_key)
+    if len(instances) < 1 or not isinstance(instances, (list, tuple)):
+      self.send_error('Request instances object have to use be a list')
+    instances = decode_b64_if_needed(instances)
+    input_columns = instances[0].keys()
```
yupbank: `input_columns = signature.inputs.keys()` ?

jlewi:
@lluunn Please address @yupbank's comments and then "/assign jlewi".

components/k8s-model-server/http-proxy/server.py
lluunn: Done.

components/k8s-model-server/http-proxy/server.py
lluunn: Done.

components/k8s-model-server/http-proxy/server.py
lluunn: Not sure, but here we will print out the explicit warning message, I think it will be more informative.

lluunn:
Review status: 0 of 3 files reviewed at latest revision, 7 unresolved discussions.

---

*[components/k8s-model-server/http-proxy/server.py, line 76 at r2](https://reviewable.io:443/reviews/kubeflow/kubeflow/334#-L6dUaFYEpOWoPH9zQ_-:-L6ro0M_ADjV-rmrosQE:b2nhswy) ([raw file](https://github.com/kubeflow/kubeflow/blob/29e313baa860e34a94f7803244cc426919e59445/components/k8s-model-server/http-proxy/server.py#L76)):*

Do we have unittests to cover these different use cases?
</blockquote></details>

Yes

---

*[components/k8s-model-server/http-proxy/server.py, line 85 at r2](https://reviewable.io:443/reviews/kubeflow/kubeflow/334#-L6dUlPCCl14jQ5rIXni:-L6ro-Nzx3XP34G0lDwX:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/29e313baa860e34a94f7803244cc426919e59445/components/k8s-model-server/http-proxy/server.py#L85)):*

Doc string please. What does this method do?
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/334)*

lluunn:
Comment addressed.
/assign jlewi

yupbank:
can you resolve the conflict?

lluunn:
Merged master.

jlewi:
Reviewed 2 of 3 files at r1, 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/334)*

jlewi:
@yupbank can you "/lgtm" if you are satisfied.

jlewi:
I just disabled the reviewable status check so as not to block auto-merge in the future (#356). 
Lets see if /skip works.
/skip

PR #339, opened 2018-03-04
Build the TF Serving GPU image as part of a release workflow.

### LINKED ISSUE #338: Build TFServing GPU Docker Image
We need to add the TFServing GPU image to the argo workflows we use to build our Docker images.

/assign jlewi

jlewi:
Create an initial release Argo workflow to build all our release artifacts.

* Build the TF GPU serving image as part of this workflow (as well as CPU version).
* This is the first step towards creating a separate Argo workflow to build all our release artifacts.
* I think it makes sense to have a separate workflow for our releases rather than repurposing our E2E tests.
    * I think the release workflow will be
            * Build all artifacts
            * Update ksonnet configs
            * run all E2E tests

* Immediate use of this workflow is to start automating builds of the GPU image for TF serving.

Fix #338

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/339)


jlewi:
The presubmit timed out waiting for the workflow to finish but it looks like it eventually succeeded. The problem is that building the GPU image is relatively slow since we have to recompile everything.

jlewi:
#347 should be submitted first but this is ready for review.

jlewi:
/hold 
until #347 submitted

PR #341, opened 2018-03-04
Add classify support for http-proxy

yupbank:
Right now, the default tf-model exported example is using classify api, this pr add classify support.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/341)

### DIFF
--- a/components/k8s-model-server/http-proxy/server.py
+++ b/components/k8s-model-server/http-proxy/server.py
@@ -124,6 +124,10 @@ def decode_b64_if_needed(value):
 
 
 class PredictHandler(tornado.web.RequestHandler):
+    """
+    Predict Hanlder proxy predict method, the input of tf savedModel is expected to be a `Map<strinbg, tf.Tensor>` protobuf
+    Defined here https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/prediction_service.proto#L23
+    """
     @gen.coroutine
     def post(self, model, version=None):
         request_key = self.settings['request_key']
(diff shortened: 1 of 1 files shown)


components/k8s-model-server/http-proxy/server.py
```diff
@@ -129,7 +188,9 @@ def get_application(**settings):
     return tornado.web.Application(
             [
             (r"/model/(.*):predict", PredictHandler),
+            (r"/model/(.*):classify", ClassifyHandler),
```
jlewi: Can you add docs explaining the difference between predict and classify?

yupbank: so with classify method, the default model we have `gs://kubeflow-models/inception`  would finally work with http-proxy.

components/k8s-model-server/http-proxy/server.py
```diff
@@ -86,7 +122,12 @@ def decode_b64_if_needed(value):
     else:
         return value
 
+
 class PredictHandler(tornado.web.RequestHandler):
+    """
+    Predict Hanlder proxy predict method, the input of tf savedModel is expected to be a `Map<strinbg, tf.Tensor>` protobuf
```
jlewi: What does protobuf mean in this context? i.e. does it mean the json representation of the protobuf?

components/k8s-model-server/http-proxy/server.py
jlewi: So the difference is ClassifyHandler expects as input a json dictionary with fields model_spec and input where as PredictHandler expects a map of names to tensor?

Can you update README.md to explain this?

yupbank:
> So the difference is ClassifyHandler expects as input a json dictionary with fields model_spec and input where as PredictHandler expects a map of names to tensor?

The different is actually implementation details, from the API we expose to user, it's the same...

Just that, when the user is exporting their model they have options to export as classify or predict.

https://github.com/tensorflow/serving/issues/794 
i have filed an issue upstream 

> Can you update README.md to explain this?

I will add one!

components/k8s-model-server/http-proxy/server.py
yupbank: the protobuf is referring to backend.. from the user perspective, the input schema is the same

jlewi:
/lgtm

More documentation would be useful but I don't think we need to block this PR for that.

@lluunn Could you please take a quick look since you are the prediction expert?

PR #342, opened 2018-03-04
Add docker file for mxnet serving

yupbank:
Config and lunch file are copied from https://github.com/awslabs/mxnet-model-server/tree/master/docker/mms_docker_cpu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/342)


jlewi:
What is the added value Kubeflow will provide?
Does AWS already publish Docker images?
If AWS already published docker images, why does Kubeflow need to maintain its own copies of the Dockerfiles and Docker images?
Could we just use the published Docker images and add a ksonnet prototype to make it easy to deploy the models?

yupbank:
from the repo.. 

>You will need to build the image yourself. The Docker image is not available on Docker Hub yet.

components/k8s-model-server/mxnet-serving/mms_docker_cpu/mms_app.conf
```diff
@@ -0,0 +1,47 @@
...
+4
+
+--worker-class
+gevent
+
+--limit-request-line
+0
+
+# Nginx configurations
+server {
```
jlewi: Is this running NGINX? Do we need it?

components/k8s-model-server/mxnet-serving/mms_docker_cpu/mms_app.conf
```diff
@@ -0,0 +1,47 @@
+# mxnet-model-server arguments
```
jlewi: What is this conf for?

jlewi:
@yupbank Any update?

components/k8s-model-server/mxnet-serving/mms_docker_cpu/mms_app.conf
yupbank: this is how the amazon version of mxnet seving works.. https://github.com/awslabs/mxnet-model-server
do we want to dig too deep into the implementation?

components/k8s-model-server/mxnet-serving/mms_docker_cpu/mms_app.conf
yupbank: this is how the amazon version of mxnet seving works.. https://github.com/awslabs/mxnet-model-server
do we want to dig too deep into the implementation?

jlewi: According to the README the image is available in DockerHub.
https://github.com/awslabs/mxnet-model-server/blob/master/docker/README.md

And I tried pulling 
 awsdeeplearningteam/mms_cpu:beta
and that seems to work.

Can you explain how mms works? 

The docs say if invoked directly it uses a standalone  Flask server but they recommend using GUnicorn for production.
https://github.com/awslabs/mxnet-model-server

It seems like NGINX is being used as a proxy for doing things like SSl validation. On K8s I don't think we would use NGINX for this (we would probably rely on Envoy/ISTIO).

However, here it mentions using NGINX to forward traffic to a unix socket because that's what gUNICORN binds.
https://github.com/awslabs/mxnet-model-server/blob/master/docker/README.md

Looking at the example:
https://github.com/awslabs/mxnet-model-server#serve-a-model

It seems like we would just need to create a K8s deployment in order to deploy an MXNet model. Would it be worth it to try to turn that into a Kubeflow example and see what features are needed to make that work better?

yupbank: nice! Then I’ll close this and make an example

jlewi: Great thanks!

yupbank:
closing  this and make a kubeflow/example instead

PR #354, opened 2018-03-05
Avoid code duplication between the two Dockerfiles

### LINKED ISSUE #321: Avoid code duplication in Dockerfiles for Jupyter notebook images
We build two versions of the Jupyter notebook images; one for CPU and one for GPU.

There's a lot of code duplication (e.g. list of python packages to install).

To avoid this duplication we should create a common shell script that can be invoked in the Dockerfile to perform the steps that should be the same.

This will help ensure there's less divergence over time.

ojarjur:
This change moves almost all of the commands run during the
building of our CPU and GPU Docker images into a single, shared
script called `install.sh`.

There are currently only two differences between the two image
types (those being the base image and the version of the
tensorflow build to install). The rest of the contents of the
Dockerfiles is meant to be identical.

However, maintaining two nearly-identical sets of instructions
is not practical. It requires that each change be duplicated
in both files, and we've seen that, in practice, this doesn't
happen reliably and the files drift over time.

By placing the bulk of the common code into a shared file, we
can reduce the risk of that happening in the future, while also
reducing the amount of work developers need to do when making
a change.

However, this change does have one downside; since all of the
install steps happen in a single Docker `RUN` command, modifying
any of them will result in the entire image having to be rebuilt.

This fixes #321

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/354)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile.cpu
+++ b/components/tensorflow-notebook-image/Dockerfile.cpu
@@ -7,55 +7,6 @@ USER root
 # features (e.g., download as all possible file formats)
 ENV DEBIAN_FRONTEND noninteractive
 
-RUN apt-get update && apt-get install -yq --no-install-recommends \
-    apt-transport-https \
-    build-essential \
-    bzip2 \
-    ca-certificates \
-    curl \
-    emacs \
-    fonts-liberation \
-    g++ \
-    git \
-    inkscape \
-    jed \
-    libav-tools \
-    libcupti-dev \
-    libsm6 \
-    libxext-dev \
-    libxrender1 \
-    lmodern \
-    locales \
-    lsb-release \
-    openssh-client \
-    pandoc \
-    pkg-config \
-    python \
-    python-dev \
-    sudo \
-    unzip \
-    vim \
-    wget \
-    zip \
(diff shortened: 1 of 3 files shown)


inc0:
I think better way would be to have base docker image and 2 images with FROM. Even better would be to use multi-stage builds to make images smaller.
Limiting layers might be useful if we add cleanups to them, this way we can provide smaller images, but that's different story.

ojarjur:
@inc0 I had not heard of multi-stage builds before. Thanks for the pointer; I will give that a shot.

inc0:
I'll push PR with example in a bit (solving different issue)

jlewi:
@inc0 @ojarjur Does multi-stage build really help here? Doesn't that only help if one stage is building an artifact for another stage.

If we want to preserve multiple layers for the image, can't we just have multiple shell scripts where we have 1 shell script per RUN step?

Another approach be to use a template which allows substituting in a different base image. I think Docker now has some support for parameterizing the base image.

inc0:
@jlewi right, I was thinking of something like having our main Dockerfile and then multi-stage parametrized build that will copy cpu or gpu artifacts respectively.

Issue #345, opened 2018-03-05
autoformat_jsonnet.sh  unknown predicate -E


jlewi:
When I run the autoformat_jsonnet command on my debian machine I get the following error

```
 ../scripts/autoformat_jsonnet.sh 
++ pwd
+ find -E /home/jlewi/git_kubeflow/releasing -iregex '.*\.(libsonnet|jsonnet)$' -exec jsonnet fmt '{}' -i --string-style d --comment-style s --indent 2 ';'
find: unknown predicate `-E'
Try 'find --help' for more information.
```

My guess is that "-E" isn't universally recognized by find.

PR #357, opened 2018-03-05
Remove -E from find command as it is not supported by GNU find

ankushagarwal:
BSD find BRE does not support alteration operator, so using
two find commands instead of one.

Tested on OSX and Ubuntu

Fixes https://github.com/kubeflow/kubeflow/issues/345

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/357)


scripts/autoformat_jsonnet.sh
```diff
@@ -18,10 +18,22 @@
 # Assumes jsonnet is on the path.
 set -ex
 
+script_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
jlewi: What does script_dir end up being set to? Why isn't using pwd sufficient?

I think we want to use the current working directory because we want to be able to use this script to format jsonnet files in other repositories.

ankushagarwal: script_dir is the directory in which `autoformat_jsonnet.sh` lives. 

pwd also works fine - the reason I made this change was to support calling `autoformat_jsonnet.sh` from any directory. The result would be same whether the script was executed from the root directory of the repo or the `scripts/` directory of the repo.

I can change it back to pwd if you prefer.

jlewi:
/lgtm

Looks like it got of sync. Once you resolve conflicts you can just "/approve" it yourself.

PR #359, opened 2018-03-06
Allow seperate naming of model name vs deployment name for TF Serving.

elsonrodriguez:
There are situations in which many models will be deployed with different names, however currently the model name uses the deployment name, which means changing the client to match each individual deployment.

This makes it so that you can set the model name independently, and reuse an unmodified client to query each deployed model.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/359)


elsonrodriguez:
I probably butchered the jsonnet syntax, if anyone has a better way let me know.

Issue #291, opened 2018-02-24
E2E Test for TFServing with GPUs


jlewi:
We need an E2E test for TF Serving with GPUs.

As part of this we should built it continuously with prow.

jlewi:
I have some cycles to work on this.
I'm going to start by adding a ksonnet component to our E2E test to deploy with GPUs.

jlewi:
Bump to P1 since we want to have GPU serving in our 0.1 release.

@lluunn How can we serve a model on GPUs and verify that GPUs were actually used?

jlewi:
https://stackoverflow.com/questions/42630762/how-to-verify-tensorflow-serving-is-using-gpus-on-a-gpu-instance
    Suggests looking at nvidia-smi output
    Some of these metrics should now be available via stackdriver

jlewi:
@lluunn This is blocked on #292 and the changes I requested in #383 to pass in a list of parameters to set on the ksonnet component. 

Once those are fixed do you want to pick this up? I think the next step would be adding appropriate steps to our E2E workflow to run the test using GPUs just like we do with CPUs.

lluunn:
I am changing the cluster to kubeflow-ci, which has GPU pool.
https://github.com/kubeflow/testing/issues/18

jlewi:
@lluunn Any update on the E2E test?

lluunn:
WIP #442

Issue #292, opened 2018-02-24
TFServing deployment should support GPUs


jlewi:
Our TFServing component should have options to support serving with GPUs.

jlewi:
/assign jlewi
/unassign @lluunn 

I've started work on this.

PR #362, opened 2018-03-06
Create a GPU model deployment to use for E2E testing of serving with GPUs

jlewi:
To support E2E testing of serving with GPUs add to our test app a TFServing deployment that uses GPUs.

* Fix the Docker file for the GPU image; we need to remove the symbolic links from /usr/local/nvidia to /usr/local/cuda
    * On GKE the device plugin will make drivers available at /usr/local/nvidia and we don't want this to
      override /usr/local/cuda
Related to #291

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/362)


components/k8s-model-server/images/releaser/environments/testing/params.libsonnet
```diff
@@ -0,0 +1,14 @@
+local params = import "../../components/params.libsonnet";
+params + {
+  components +: {
+    // Insert component parameter overrides here. Ex:
+    // guestbook +: {
+    //   name: "guestbook-dev",
+    //   replicas: params.global.replicas,
+    // },
```
ScorpioCPH: nit: remove commenting code.

components/k8s-model-server/images/releaser/environments/testing/main.jsonnet
```diff
@@ -0,0 +1,7 @@
+local base = import "../base.libsonnet";
+local k = import "k.libsonnet";
+
+base + {
+  // Insert user-specified overrides here. For example if a component is named "nginx-deployment", you might have something like:
+  //   "nginx-deployment"+: k.deployment.mixin.metadata.labels({foo: "bar"})
```
ScorpioCPH: Any thing take effect?

jlewi: No. This is just boilerplate generated by ks init.

jlewi:
/assign @lluunn 
@lluunn  PTAL.

components/k8s-model-server/images/releaser/components/gpu_model.jsonnet
```diff
@@ -0,0 +1,45 @@
...
+                     spec+: {
+                       containers: std.prune([tfServingContainer, httpProxyContainer]),
+                       // TODO(jlewi): For the CPU image we set the user and group to 1000 which are defined within the Docker container.
+                       // But we don't do the same for the GPU image.
+                       securityContext: null,
+                     },
+                   },
+                 },
+               };
+std.prune(k.core.v1.list.new([
```
lluunn: So this deploys a TF serving component with GPU image. Why is it part of releaser?

jlewi: The releaser is poorly named. This ksonnet app contains the E2E test workflow. The initial use for this is to deploy the image as part of our E2E tests and make sure it works.
https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml#L7

Its possible this will get replaced by #292; i.e. once our TFServing component supports GPUs we can just use that. 

The purpose of this PR is to provide a one off test that can be used as the basis for that component.

PR #366, opened 2018-03-06
Consolidate down to a single Dockerfile

### LINKED ISSUE #321: Avoid code duplication in Dockerfiles for Jupyter notebook images

ojarjur:
This change eliminates the duplication between the two Dockerfiles
(one for CPU-only instances and one for GPU-enabled instances) by
eliminating the second file.

The two files were meant to differ in two ways:

1. The base image used (ubuntu vs. nvidia/cuda)
2. The tensorflow package installed (tf-nightly vs. tf-nightly-gpu)

It turns out that we can just make both of those arguments (defined
by an `ARG` statement with a default value), and then the two
files can be merged into one.

The default values chosen for the args are the ones for the CPU
image, so running `docker build .` without any `--build-arg` flags
will produce the CPU image.

To build a GPU image, pass in the corresponding values for the
`BASE_IMAGE` and `TF_PACKAGE` arguments using `--build-arg` flags.

This fixes #321


ojarjur:
... Actually, hold off on this for now; the TF_PACKAGE isn't being resolved correctly for some reason.

I'll update this PR when it is ready for review.

components/tensorflow-notebook-image/Dockerfile
```diff
@@ -1,6 +1,9 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
-FROM ubuntu:latest
+
+ARG BASE_IMAGE=ubuntu:latest
+ARG TF_PACKAGE=tf-nightly
```
ojarjur: This ARG should be moved down to right before it is used.

ojarjur:
I got the TF_PACKAGE issue resolved; this PR should be ready for review now.

jlewi:
This is great.

Our release process actually uses this
script
https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/build_image.sh

which is invoked by argo here
https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/releaser/components/workflows.libsonnet#L136

Could you update these so that the release process will still work?

Issue #301, opened 2018-02-27
Test the TFServing configs with default image


jlewi:
We need a test to verify that the current TFServing component works with the default image.

The current E2E test builds a new TFServing image and then deploys and tests it.
This is great for testing a new Docker image, but it doesn't test the current default image works.

We should add options to the E2E workflow used for testing so that we can test TFServing without rebuilding the image.

jlewi:
I think its fine in the short term to change our E2E workflow to not build the image and instead run the test using the current ksonnet configs.

This will work well for our release process because the flow will be

- Run the Argo workflow to build all our containers
- Update the configs
- Create a PR and run E2E tests.

Longer term it would be good if we had an E2E test that also rebuilt the docker image and to selectively trigger the E2E test when certain files (like the Dockerfile) are changed so we don't always rerun the test.

jlewi:
It would probably make sense to make it a parameter of the current workflow to controller whether we rebuild the Docker image for TF serving and use that image or use the default image.

lluunn:
Fixed by #383

PR #383, opened 2018-03-07
[WIP] Add build-image option to TF serving workflow

lluunn:
If build_image = true, build the image and test it.
Else, use the default image (tag=1.0) to test.

build_image default to false.

Also fix the image name in Makefile

### DIFF
--- a/components/k8s-model-server/images/Makefile
+++ b/components/k8s-model-server/images/Makefile
@@ -19,7 +19,7 @@ GCR_PROJECT=gcr.io/${PROJECT_ID}
 DOCKERHUB_PROJECT=hub.docker.com/r/${PROJECT_ID}
 
 define build
-	docker build --pull -t $(1)/tf-model-server:${TF_VERSION} -f Dockerfile.cpu .
+	docker build --pull -t $(1)/tf-model-server-cpu:${TF_VERSION} -f Dockerfile.cpu .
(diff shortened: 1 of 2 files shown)


lluunn:
related: #301

jlewi:
:lgtm:

---

Review status: 0 of 3 files reviewed at latest revision, 3 unresolved discussions.

---

*[components/k8s-model-server/images/Makefile, line 19 at r4](https://reviewable.io:443/reviews/kubeflow/kubeflow/383#-L72h-JU-yjMdWSBzvSj:-L72h-JVZtlEFxYb0HVU:b877zjo) ([raw file](https://github.com/kubeflow/kubeflow/blob/2c121f3f8373e1fb9e3190f49726d71e565f33e1/components/k8s-model-server/images/Makefile#L19)):*
> ```Makefile
> PROJECT_ID=kubeflow
> GCR_PROJECT=gcr.io/${PROJECT_ID}
> DOCKERHUB_PROJECT=hub.docker.com/r/${PROJECT_ID}
> ```

We should maybe think about deleting this file rather than keep updating it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/383#-:-L72hNS5Ua2RJtPMc6gS:bnfp4nl)*

PR #386, opened 2018-03-07
Build http proxy image in release workflow

### LINKED ISSUE #353: Build http proxy for TF serving as part of our release workflow
#339 is adding an Argo workflow to build all our artifacts as part of our release.

We should add the HttpProxy image to this workflow as well.

lluunn:
Fixed  #353

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/386)


yupbank:
/lgtm

thanks!

Issue #376, opened 2018-03-07
[Discussion] Patterns for creating ksonnet prototypes?


jlewi:
For many prototypes like TFServing we might want to allow complex modifications

Example
   * Users might want to customize resource foot print of HttpProxy and/or TFServing
   * Users might want to attach secrets or volumes
   * Users might need to set multiple environment variables 
   * See #323 for example

Creating parameters for each flag won't scale. We need to structure our ksonnet prototypes to make it easier to override the parts people care about. Furthermore, this will probably require users modifying the component jsonnet files directly as opposed to just overriding paramters.

As an example if a user wants to override the resources for TFServing it would be nice if they could just supply a dictionary.

It would be useful to look at the ksonnet/parts registry for examples/motivation.

jlewi:
One issue is what to do about default parameter values for optional parameters.

As an example consider the default docker image to use for TFServing. The correct default image to use depends on whether the user is using CPU or GPU.

Option 1: Have two parameters "mode_server_cpu_image", "model_server_gpu_image"
Option 2: Have a single parameter "model_server_image" and set the value in GCL depending on GPUs.
Option 3: Don't define parameters in the prototype. Instead we can define a maps of defaultParams in the .libsonnet file and then override it with whatever parameters are set for the component/environment.

e.g. in tfserving.libsonnet
```
{
  parts:: {
      defaultParams:: {
          model_image:: ...
          ....
      }

     modelServer(overrides={}) :: {
         local params = defaultParams + overrides;
     }
```

I like option 2 much better but there are some drawbacks.

Until ksonnet/ksonnet#273 is fixed we need to set the default to some magic string like "null" and use that to detect when the value is default.

The other draw back is that the default parameter values won't show up when user does `ks param list`

I think the drawback of option 3 is that we don't have the benefit of the help strings encoded in the prototype.

jlewi:
Thinking about this more and based on
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/declarative-application-management.md

I think there are a class of components specifically

TfJob
Argo Workflows
TFServing images
Seldon models
TensorBoard jobs

Where I think it might make more sense to for an existing template/example and then let users modify it. I think this ultimately leads to more readable config than doing complicated things like overriding environment variables and requests limit.

We can support this by using ksonnet prototypes. When we generate a component from a prototype ksonnet copies that prototype into components directory.

[...]

PR #387, opened 2018-03-07
Refactor the TFServing component to better support GPUs and specific clouds

### LINKED ISSUE #292: TFServing deployment should support GPUs

jlewi:
* To support GPUs and specific clouds we refactor the component to make
  it easy to override the parts we care about (e.g. container environment
  variables, resources, etc...).

* We do this by moving the things we care about up to the root of tf-serving.libsonnet.

* We rely on jsonnet late binding (http://jsonnet.org/docs/tutorial.html).
Late binding allows us to devine dictionaries (e.g. params, tfServingContainer)
in tf-serving.libsonnet. We can then create manifests based on those
objects (e.g. tfDeployment). We can then override values (e.g. params) and
the derived objecs (e.g. tfDeployment) will use the overwritten values.

* We introduce a parameter "cloud" which allows us to control which "prototype" to use. We use this to use cloud specific customizations; like setting
  the environment variables on AWS to use S3.

* Late binding also makes it possible to select an appropriate default image
  based on whether GPUs are bing used or not while still allowing the
  user to override the images.

* We remove parameter definitions from the prototypes. The set of parameters
  ends up being conditional based on flags like cloud, GPUs so its
  not clear how scalable that was.

Related Issues:
   #376 patterns for ksonnet prototypes.
Fix #292

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/387)


lluunn:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[kubeflow/tf-serving/tf-serving.libsonnet, line 143 at r1](https://reviewable.io:443/reviews/kubeflow/kubeflow/387#-L75hbs_Zyo1ArNNRXhF:-L75hbs_Zyo1ArNNRXhG:b-l5qgqx) ([raw file](https://github.com/kubeflow/kubeflow/blob/ffc8fbd5e69d79a4cddc19802ee5b1d553d6e8ff/kubeflow/tf-serving/tf-serving.libsonnet#L143)):*
> ```
>           cpu: "1",
>         },
>         limits: {
> ```

Do you want to lower these numbers?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/387)*

lluunn:
There are conflicts. 
/lgtm

jlewi:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[kubeflow/tf-serving/tf-serving.libsonnet, line 143 at r1](https://reviewable.io:443/reviews/kubeflow/kubeflow/387#-L75hbs_Zyo1ArNNRXhF:-L769mLQM_IBaXCDygyW:byblvlm) ([raw file](https://github.com/kubeflow/kubeflow/blob/ffc8fbd5e69d79a4cddc19802ee5b1d553d6e8ff/kubeflow/tf-serving/tf-serving.libsonnet#L143)):*
<details><summary><i>Previously, lluunn (Lun-Kai Hsu) wrote…</i></summary><blockquote>

Do you want to lower these numbers?
</blockquote></details>

We probably should. If you have good numbers we can change it otherwise we'll leave it for a follow on PR. There's an existing issue related to http proxy resources.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/387)*

lluunn:
Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/kubeflow/kubeflow/387)*

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -1,152 +1,262 @@
...
+
+    // URL for your s3-compatible endpoint.
+    s3Endpoint: "http://s3.us-west-1.amazonaws.com,",
+  } + $.params,
+
+
+  components:: {
+
+    all::
+      if $.params.cloud == "aws" then
```
elsonrodriguez: This would be better as params.s3Enable. The S3 functionality isn't specific to AWS.

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -1,152 +1,262 @@
...
+        type: $.params.serviceType,
+      },
+    },  // tfService
+
+  },  // parts
+
+  // Parts specific to S3
+  s3parts:: $.parts {
+    s3Env:: [
+      { name: "AWS_ACCESS_KEY_ID", valueFrom: { secretKeyRef: { name: $.s3params.s3SecretName, key: $.s3params.s3SecretAcesskeyidKeyName } } },
```
elsonrodriguez: s/s3SecretAcesskeyidKeyName/s3SecretAccesskeyidKeyName

jlewi:
Looks like this got auto-merged because Lunkai is an approver (its a quirk of the automatic review system). I'll send a new PR.

PR #390, opened 2018-03-08
Update tfserving gpu image to create model-server user

### LINKED ISSUE #367: Set userid and group id in TF Serving GPU container
In the CPU container we create a user id and group id for user 1000 here
https://github.com/kubeflow/kubeflow/blob/master/components/k8s-model-server/images/Dockerfile.cpu#L53

We set that userid and groupid as part of the deployment here
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/tf-serving/tf-serving.libsonnet#L143

We should probably do something similar for the GPU container. Otherwise I think users might have to override the securityContext in the deployment so as to avoid setting user id and group id.

ankushagarwal:
Fixes #367

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/390)

### DIFF
--- a/components/k8s-model-server/images/Dockerfile.gpu
+++ b/components/k8s-model-server/images/Dockerfile.gpu
@@ -88,9 +88,16 @@ RUN cd /root/serving/tensorflow && ./configure
 RUN cd /root/serving && \
     bazel build -c opt --copt=-mavx --copt=-mavx2 --copt=-mfma --copt=-mfpmath=both --copt=-msse4.2 --config=cuda -k --verbose_failures --crosstool_top=@local_config_cuda//crosstool:toolchain tensorflow_serving/model_servers:tensorflow_model_server
 
-# Add some softlinks for tensorflow model server and CUDA
-RUN ln -s /root/serving/bazel-bin/tensorflow_serving/model_servers/tensorflow_model_server /usr/bin/tensorflow_model_server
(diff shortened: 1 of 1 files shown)


components/k8s-model-server/images/Dockerfile.gpu
```diff
@@ -88,9 +88,16 @@ RUN cd /root/serving/tensorflow && ./configure
 RUN cd /root/serving && \
     bazel build -c opt --copt=-mavx --copt=-mavx2 --copt=-mfma --copt=-mfpmath=both --copt=-msse4.2 --config=cuda -k --verbose_failures --crosstool_top=@local_config_cuda//crosstool:toolchain tensorflow_serving/model_servers:tensorflow_model_server
 
-# Add some softlinks for tensorflow model server and CUDA
-RUN ln -s /root/serving/bazel-bin/tensorflow_serving/model_servers/tensorflow_model_server /usr/bin/tensorflow_model_server
+# Make a copy to /usr/bin instead of making softlinks because /root has 700 permissions
+RUN cp /root/serving/bazel-bin/tensorflow_serving/model_servers/tensorflow_model_server /usr/bin/tensorflow_model_server
```
jlewi: FWIW I had a container running from the old image and I verified that if we copy it a new location it still runs; i.e. no linking issues.

PR #396, opened 2018-03-08
Fix k8s dashboard on Azure

wbuchwalter:
On azure k8s's dashboard is running on port `80` not `443`.
This PR parametrize the ambassador config for the dashboard based on the cloud name.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/396)


jlewi:
It looks like you need to sync. 

Overall the change looks good to me. But can you explain the overall ingress story on Azure? Are you planning on exposing individual microservices (TFJobs UI, Central UI, TB) on HTTPs for external access?

Why wouldn't you just expose Ambassador on HTTPs and then use that as a secure reverse proxy to all the individual services?

wbuchwalter:
@jlewi I do want to use Ambassador as single access point, and this PR makes the k8s dashboard work with ambassador on Azure.

The issue here is that the Ambassador configuration for the k8s dashboard expect to find the service exposed on tls internally: `"tls: true"`, but on Azure, the dashboard is exposed on port `80`, so querying:
`127.0.0.1:8001/api/v1/namespaces/kubeflow/services/ambassador:80/proxy/k8s/ui/` breaks.
This PR will set `"tls: false"` when on Azure so that the k8s dashboard works through ambassador.

jlewi:
@wbuchwalter The reason I've been slow to respond is because it wasn't showing up in my PR dashboard
You need to add me as a reviewer by doing /cc jlewi or assign it to me using /assign
For more info see
https://github.com/kubeflow/community/blob/master/CONTRIBUTING.md

PR #400, opened 2018-03-09
use tini to start model server

yupbank:
Even though model server itself can be the entry point itself but then run with PID 1 would make the server not responding to SIGINT or SIGTERM

which is problematic during k8s rolling up new version

Using [Tini](https://github.com/krallin/tini) has several benefits:

- It protects you from software that accidentally creates zombie processes,
  which can (over time!) starve your entire system for PIDs (and make it
  unusable).
- It ensures that the *default signal handlers* work for the software you run
  in your Docker image. For example, with Tini, `SIGTERM` properly terminates
  your process even if you didn't explicitly install a signal handler for it.
- It does so completely transparently! Docker images that work without Tini
  will work with Tini without any changes.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/400)

### DIFF
--- a/components/k8s-model-server/images/Dockerfile.cpu
+++ b/components/k8s-model-server/images/Dockerfile.cpu
@@ -53,3 +53,11 @@ RUN set -x \
     && useradd $MS_USER \
     && [ `id -u $MS_USER` -eq 1000 ] \
     && [ `id -g $MS_USER` -eq 1000 ]
+
+ENV TINI_VERSION v0.17.0
+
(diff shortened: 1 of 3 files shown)


lluunn:
Sorry I am not familiar with tini.
Could you update some description about what's tini and what does it do?
Also, is it common to use that in k8s?

Thanks

yupbank:
sure.. i have updated..

i think it's common, like we use it for `nginx ` when deploy with docker

lluunn:
Thanks! I can see we use tini in jupyter image as well.
/lgtm

Issue #394, opened 2018-03-08
Build Envoy Container with JWT validation


jlewi:
See #60 

We will need to build an envoy container that contains an updated binary with JWT validation support for IAP.

The team maintaining the JWT fork of envoy isn't building containers anymore but they are building a binary  which is available in GCS

```
curl -fLSs https://storage.googleapis.com/istio-build/proxy/envoy-alpha-${ISTIO_ENVOY_VERSION}.
tar.gz}
```

The fixes we need should be in this version.

```
export ISTIO_ENVOY_VERSION=0fb4886b463698702b6a08955045731903a18738
```

This was a Dockerfile I used in the past
https://github.com/jlewi/proxy/blob/iap/docker/Dockerfile.kubeflow

ankushagarwal:
Looks like we don't have the correct permissions

```
export ISTIO_ENVOY_VERSION=0fb4886b463698702b6a08955045731903a18738
curl -fLSs https://storage.googleapis.com/istio-build/proxy/envoy-alpha-${ISTIO_ENVOY_VERSION}.tar.gz

curl: (22) The requested URL returned error: 403 Forbidden
```

ankushagarwal:
This can be closed because https://github.com/kubeflow/kubeflow/pull/409 is merged

PR #401, opened 2018-03-09
Use the correct jwt validation config

ankushagarwal:
It is defined in https://github.com/istio/proxy/blob/master/src/envoy/http/jwt_auth/config.proto in the JWT message

Tested by deploying to dev.kubeflow.org and verifying that JWT validation works when IAP is turned off

Related https://github.com/kubeflow/kubeflow/issues/394
Related https://github.com/kubeflow/kubeflow/issues/60

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/401)

### DIFF
--- a/kubeflow/core/prototypes/iap-envoy.jsonnet
+++ b/kubeflow/core/prototypes/iap-envoy.jsonnet
@@ -4,7 +4,7 @@
 // @shortDescription Envoy proxies to handle ingress and IAP.
 // @param name string Name to give to each of the components
 // @optionalParam namespace string default Namespace
-// @optionalParam envoyImage string gcr.io/kubeflow/envoy:v20180124-0.4.0-50-g0d29aac-dirty-4d9e20 The image for envoy.
+// @optionalParam envoyImage string gcr.io/kubeflow-dev/envoy:0fb4886b463698702b6a08955045731903a18738 The image for envoy.
 // @optionalParam disableJwtChecking string false Disable JWT checking.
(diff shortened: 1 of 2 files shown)


kubeflow/core/prototypes/iap-envoy.jsonnet
```diff
@@ -4,7 +4,7 @@
 // @shortDescription Envoy proxies to handle ingress and IAP.
 // @param name string Name to give to each of the components
 // @optionalParam namespace string default Namespace
-// @optionalParam envoyImage string gcr.io/kubeflow/envoy:v20180124-0.4.0-50-g0d29aac-dirty-4d9e20 The image for envoy.
+// @optionalParam envoyImage string gcr.io/kubeflow-dev/envoy:0fb4886b463698702b6a08955045731903a18738 The image for envoy.
```
jlewi: nit: This is perfectly for now. But in general I suggest we follow a naming convention of something like the following for labels
vYYYYMMDD-GITHASH

Using the date is convenient for humans to ascertain at a glance which images are newer. The GITHASH does 2 thing
1. It ensures if we build a new image from new code we get a different tag 
2. It makes it easy to link an image to its source.

jlewi:
This makes me so happy.

kubeflow/core/prototypes/iap-envoy.jsonnet
ankushagarwal: ack

Issue #404, opened 2018-03-09
Replace two envoy containers with one in the envoy pod


ankushagarwal:
envoy now supports specifying paths on which we should use JWT validation. We can use this feature to do JWT validation on the non-healthcheck routes. This will allows us to replace the two envoy containers with a single one.

PR #407, opened 2018-03-09
Replace two envoy containers with a single one

ankushagarwal:
Use bypass_jwt field to bypass jwt validation for /healthz and
/noiap/whoami routes

Tested by disabling IAP - verified that /healthz and /noiap/whoami work
and other routes return an error - Required JWT token is missing

Fixes https://github.com/kubeflow/kubeflow/issues/404

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/407)

### DIFF
--- a/kubeflow/core/iap.libsonnet
+++ b/kubeflow/core/iap.libsonnet
@@ -35,8 +35,8 @@
         ports: [
           {
             name: "envoy",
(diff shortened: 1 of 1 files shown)


jlewi:
This is fantastic.

PR #411, opened 2018-03-12
Use kubeflow-ci as the test cluster

lluunn:
Related: https://github.com/kubeflow/testing/issues/18

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/411)

### DIFF
--- a/components/k8s-model-server/images/releaser/components/params.libsonnet
+++ b/components/k8s-model-server/images/releaser/components/params.libsonnet
@@ -7,7 +7,6 @@
     // Component-level parameters, defined initially from 'ks prototype use ...'
     // Each object below should correspond to a component in the components/ directory
     workflows: {
-      bucket: "mlkube-testing_temp",
       commit: "master",
       name: "new9",
       namespace: "kubeflow-test-infra",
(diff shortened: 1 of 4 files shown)


lluunn:
The test timeout waiting for checkout. But it actually succeeded later.
argo get -n=kubeflow-test-infra kubeflow-presubmit-kubeflow-e2e-411-d3b1875-609-112b

Guessing github-token not being used.

lluunn:
kubeflow/testing workflow still stuck at checkout.
TF serving workflow already passed.

lluunn:
I kubectl exec into the container.
git clone https://github.com/kubeflow/tf-operator.git tmp-aa
git clone https://github.com/kubeflow/tf-operator.git /mnt/test-data-volume/aaa is extremely slow

So the problem should be the mnt

lluunn:
I ssh into the NFS server, verified the files are there, so connection should be set correctly..

lluunn:
The NFS server was in a different zone from the k8s cluster. I deployed a new one in the same zone

lluunn:
Ankush, with this change the e2e tests will be run in kubeflow-ci cluster. Presubmit passed.

PR #413, opened 2018-03-12
Create a seldon-serve ksonnet prototype

ankushagarwal:
This prototype can be used to create components which contain easily
deployable "seldondeployment" objects. The use case for this is that
once a user wraps a seldon-core model into a docker image, the user can
create a "seldondeployment" using this component.

It contains the following params: namespace, name, image, replicas.

We can add more params later.

An example for using this would be

Create a component:
ks generate seldon-serve issue-summarization --name=issue-summarizatiion
--image=gcr.io/kubeflow-images-staging/issue-summarization-seldon:20180312
--replicas=3

Apply the component
ks apply env -c issue-summarization.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/413)

### DIFF
--- a/kubeflow/seldon/prototypes/serve-simple.jsonnet
+++ b/kubeflow/seldon/prototypes/serve-simple.jsonnet
@@ -1,11 +1,12 @@
 // @apiVersion 0.1
-// @name io.ksonnet.pkg.seldon-serve
-// @description A prototype to serve a seldon model
-// @shortDescription A prototype to serve a seldon model
+// @name io.ksonnet.pkg.seldon-serve-simple
+// @description A prototype to serve a single seldon model
+// @shortDescription A prototype to serve a single seldon model
 // @param name string Name to give this deployment
 // @param image string Docker image which contains this model
-// @optionalParam namespace string default Namespace
+// @optionalParam namespace string default Namespace in which this model will be deployed
 // @optionalParam replicas number 1 Number of replicas
+// @optionalParam endpoint string REST The endpoint type: REST or GRPC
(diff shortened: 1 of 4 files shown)


cliveseldon:
@ankushagarwal  Looks interesting. The main question is what is easiest for user to have to build different graph types. For a single model then what you have done is good but what about more complex graphs? I suppose we simply add further prototypes for A/B tests, ensembles, multi-armed bandits etc., to build up a whole library of different prototypes for normal tasks. 

Also:
  * The user might want to specify REST or GRPC for their model
  * it would be expected the user would then edit the PodTemplateSpec in "componentSpec" to add Volumes and other customizations needed.

Looks like a great start.

ankushagarwal:
Added support for specification of endpoint: REST or GRPC

Yes for things which cannot be easily parametrized such as volumes,  the user would be expected to edit the componentSpec directly.

Renamed the files to serve-simple.jsonnet and renamed the prototype to seldon-serve-simple

ankushagarwal:
My idea with this prototype was to be able to come up with an easy way to deploy a single model built using `seldonio/core-python-wrapper` without having to write the entire SeldonDeployment spec myself. 

I haven't given much thought to complex graph types. I guess it makes sense to have separate prototypes for the most common scenarios like: seldon-serve-ab-test, seldon-serve-ensemble, etc.

cliveseldon:
Looks good.

ankushagarwal:
Can you /lgtm so that I can go ahead and merge?

Issue #414, opened 2018-03-12
libcublas.so.9.0: cannot open shared object file: No such file or directory


dharmeshkakadia:
I am following the user guide (https://github.com/kubeflow/kubeflow/blob/master/user_guide.md) and got the following error. I am running k8s on azure with GPU instances. I am running master code. Is there an issue with the image?

---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
/opt/conda/lib/python3.6/site-packages/tensorflow/python/pywrap_tensorflow.py in <module>()
     57 
---> 58   from tensorflow.python.pywrap_tensorflow_internal import *
     59   from tensorflow.python.pywrap_tensorflow_internal import __version__

/opt/conda/lib/python3.6/site-packages/tensorflow/python/pywrap_tensorflow_internal.py in <module>()
     27             return _mod
---> 28     _pywrap_tensorflow_internal = swig_import_helper()
     29     del swig_import_helper

/opt/conda/lib/python3.6/site-packages/tensorflow/python/pywrap_tensorflow_internal.py in swig_import_helper()
     23             try:
---> 24                 _mod = imp.load_module('_pywrap_tensorflow_internal', fp, pathname, description)
     25             finally:

/opt/conda/lib/python3.6/imp.py in load_module(name, file, filename, details)
    242         else:
--> 243             return load_dynamic(name, filename, file)
    244     elif type_ == PKG_DIRECTORY:

/opt/conda/lib/python3.6/imp.py in load_dynamic(name, path, file)
    342             name=name, loader=loader, origin=path)
--> 343         return _load(spec)
    344 

ImportError: libcublas.so.9.0: cannot open shared object file: No such file or directory

During handling of the above exception, another exception occurred:

ImportError                               Traceback (most recent call last)
<ipython-input-1-b0e475b6b7de> in <module>()
----> 1 from tensorflow.examples.tutorials.mnist import input_data
      2 mnist = input_data.read_data_sets("MNIST_data/", one_hot=True)
      3 
      4 import tensorflow as tf
      5 

/opt/conda/lib/python3.6/site-packages/tensorflow/__init__.py in <module>()
     22 
     23 # pylint: disable=wildcard-import
---> 24 from tensorflow.python import *  # pylint: disable=redefined-builtin
     25 # pylint: enable=wildcard-import
     26 

/opt/conda/lib/python3.6/site-packages/tensorflow/python/__init__.py in <module>()
     47 import numpy as np
     48 
[...]

PR #415, opened 2018-03-13
Replace azure env by aks || acs-engine

wbuchwalter:
In azure there is two main ways to create a kubernetes cluster:
* acs-engine
* AKS

`acs-engine` is the open source project used behing the scene by AKS. acs-engine receives features first and then they make their way to AKS.
The NVIDIA driver installation was recently updated in acs-engine to make it simpler, and so the mount point are now different between the two platform.
acs-engine will also soon received the update to use NVIDIA device-plugin, so the two platform will use different configuration for a few months.

Consequently I replaced the `azure` cloud by `aks`or `acsengine` to avoid issues like this one: https://github.com/kubeflow/kubeflow/issues/414#issuecomment-372502012

Once the move to device-plugin is complete in AKS we will probably be able to entirely remove the cloud param since it won't be used anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/415)


jlewi:
/lgtm
/approve

Might be a good idea to open issues to add
1. ksonnet unittests for the Azure configs
2. E2E testing on Azure.

wbuchwalter:
@jlewi Can you re /lgtm? Had to pull master to fix issues with the presubmit hook. Thanks.

Issue #385, opened 2018-03-07
TFServing prototype for using GCS with service account key


jlewi:
#323 add a prototype for plumbing through credentials and config needed to serve models from S3.

I'm in the process of refactoring that to turn it into a separate prototype to make it more manageable.

We should create a similar prototype for GCS. For GCS we want to be able to use a secret containing a google service account and set GOOGLE_APPLICATION_CREDENTIALS.

jlewi:
@lluunn  This is blocked on #292 but you could pick it up once that's committed.

lluunn:
I don't understand the goal here.
We are able to serve model from GCS now right?
@jlewi

jlewi:
The question is how do you authenticate to GCP. 

We'd like to use a service account. But we don't want to rely on the VM's service account because multiple different pods could be running on a VM all with access to the VM service account. So in general we want the VM service account to have minimal permissions.

So the model is that users will 

- create a GCP service account
- they will then download the private key for that service account
- they will create a K8s secret from that key
- Their TF serving pods will mount the secret into the pod
-- the serving pod will set GOOGLE_APPLICATION_CREDENTIALS environment variable to tell Google client auth libraries to use the key.

So we want to update the TFServing component so that users 

- can set parameter cloud = "gcp"
- Setting cloud = "gcp" causes the TFServing pod to be modified as needed to use a service account
   - I.e. secret is mounted into the pod
   - GOOGLE_APPLICATION_CREDENTIALS environment variable is set on the pod.

lluunn:
#416 fix it.

PR #416, opened 2018-03-13
GCP credential support for TF serving

lluunn:
Related #385

Verified that serving with a model in my GCS bucket does not work unless passing the param:
cloud = gcp
gcpCredentialSecretName = secret_name

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/416)


yupbank:
do we need to update readme somewhere?

lluunn:
updated README. PTAL

Issue #43, opened 2017-12-20
Remove namespace as a package parameter


jlewi:
Most of our ksonnet packages have namespace as an explicit parameter.

This is a work around for ksonnet/ksonnet#222. Once that's fixed and components can inherit namespace from the environment we can remove namespace as an explicit parameter.

ankushagarwal:
We should be unblocked on this since https://github.com/ksonnet/ksonnet/issues/222 has closed

PR #426, opened 2018-03-14
Inherit namespace from ksonnet environment

ankushagarwal:
Now that ksonnet makes `env` available in it's components, we can
leverate that to use the environment's namespace instead of hardcoding a
default namespace in the component parameters.

So now the behaviour is that when components are generated from
prototypes, we default to using the environment's namespace. If the user
has overridden namespace as a parameter, that will get used for the
component.

Tested by deploying kubeflow-core and tf-serving to a new k8s cluster

Fixes https://github.com/kubeflow/kubeflow/issues/43

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/426)


lluunn:
With #411  test should pass

kubeflow/core/jupyterhub.libsonnet
```diff
@@ -1,7 +1,4 @@
 {
-  // TODO(https://github.com/ksonnet/ksonnet/issues/222): Taking namespace as an argument is a work around for the fact that ksonnet
-  // doesn't support automatically piping in the namespace from the environment to prototypes.
-  //
```
lluunn: Isn't this TODO for removing namespace param in parts(namespace)?

kubeflow/argo/prototypes/argo.jsonnet
```diff
@@ -3,12 +3,13 @@
...
-// TODO(https://github.com/ksonnet/ksonnet/issues/222): We have to add namespace as an explicit parameter
-// because ksonnet doesn't support inheriting it from the environment yet.
 
 local k = import "k.libsonnet";
 local argo = import "kubeflow/argo/argo.libsonnet";
 
-std.prune(k.core.v1.list.new(argo.parts(params.namespace).all))
+// updatedParams includes the namespace from env by default.
+// We can override namespace in params if needed
+local updatedParams = env + params;
```
lluunn: env is the params from environment?
Aren't we already setting per environment param to override default?

ankushagarwal: `env` is not `params`. `params` are defined per component in `component/params.libsonnet` and can be overridden per-environment in `environments/<env>/params.libsonnet`

`env` is the ksonnet environment definition. It is the value you get when you execute `ks env list`

For example
```
% ks env list
NAME    KUBERNETES-VERSION NAMESPACE SERVER
====    ================== ========= ======
default v1.7.0             kubeflow  https://35.184.102.122```
```

For this default environment, `env` will contain 
```
{
  "namespace": "kubeflow",
  "server": "https://35.184.102.122",
}
```

The main idea is that we should not have to define the `namespace` parameter per component, it will automatically be inherited by the environment in which we do a `ks apply <env>`

kubeflow/core/jupyterhub.libsonnet
ankushagarwal: The comment in the TODO mentions that `Taking namespace as an argument is a work around for the fact that ksonnet doesn't support automatically piping in the namespace from the environment to prototypes` - I don't think that is true. 

The workaround for ksonnet not piping the namespace from env to prototypes is we have an explicit parameter called `namespace` in our prototype - this PR addresses this issue by removing the `namespace` parameter from our prototypes.

lluunn:
For workflow 
I see

lluunn:
https://github.com/kubeflow/tf-operator/pull/464 should fix tf-operator test

jlewi:
This is great thanks!

ankushagarwal:
@jlewi Can you please take a look again. I pushed a few changes to fix tests.

Issue #375, opened 2018-03-07
Jupyter notebook image should pin TF version


ankushagarwal:
The latest image for jupyter notebook uses tensorflow 1.5-dev version and the TFServing image uses tensorflow 1.4 release.

I ran into an issue where I trained a model using the jupyter notebook, but had compatibility issues while serving that model with TFServing.

We should try to maintain the same tensorflow versions in the various kubeflow images.

jlewi:
Our notebooks are currently using the nightly build of TF.
https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/Dockerfile.cpu#L162

We should probably pin TF to a specific version. 

/cc @ojarjur

jlewi:
We should pin the version of TF in our notebook images.
We should also consider including the TF version in the image name so that we could potentially build images for different versions of TF.

aronchick:
Pinning is great, but i wonder how we make this an easily option - we've
seen lots of issues with various versions of frameworks not supporting
particular models, and i want people to be able to swap readily.

jlewi:
Agree. For notebook images we can build different notebook images for different TF versions.

Thanks to @ojarjur's changes to the Dockerfile, its already parameterized by the TF version. And thanks to work by @pdmack we can have a drop down box in the Jupyter UI from which users can select the appropriate image to use. So we could have images for different TF versions.

PR #427, opened 2018-03-14
update the cuda base docker image version to 9.0 to align with tf-nightly-gpu

yongwen:
this is to fix https://github.com/kubeflow/kubeflow/issues/414

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/427)


yongwen:
I signed it!

jlewi:
I'm not sure this is the right solution. 

I don't think we should be using the nightly build of TF (see #375). So we should pin TF in our notebook images to a specific version of TF.

Also the Makefile is no longer used to build our Docker images. We use the Argo workflow.

jlewi:
I'm closing this PR because I think the issue is fixed by the fix to #375

yongwen:
Thank you. Good to hear the better solution.

Issue #255, opened 2018-02-16
IAP component should use cert-manager to get a signed certificate


ankushagarwal:
Currently the dev.kubeflow.org app uses a self-signed certificate for https. 

We can use https://github.com/jetstack/cert-manager to provision a valid certificate.

jlewi:
There's an example of using the predecessor kube-lego in kubeflow/reporting (but not with IAP).

I think IAP creates some challenges because we need a way to not have IAP reject the well known ACME challenge.

ankushagarwal:
Reading up on the [ACME challenge](https://letsencrypt.org/how-it-works/)

```
For example, the CA might give the agent a choice of either:

1 Provisioning a DNS record under example.com, or
2 Provisioning an HTTP resource under a well-known URI on https://example.com/
```

If we go with #1, IAP should not be involved at all

jlewi:
@danisla pointed out that doing #2 is fairly straightforward.

We just create a separate ingress using the same IP and hostname to serve the ACME challenge. This will use the same loadbalancer but create a different GCP backend service. IAP is enabled per backend service. So we simply don't enable IAP on this backend service.

jlewi:
Here's a ksonnet component for using kube-lego
https://github.com/kubeflow/reporting/blob/master/collector-app/components/kube_lego.libsonnet

We can probably start with that but we should upgrade to cert-manager.

PR #429, opened 2018-03-14
Allow creating ingress with hostname

ankushagarwal:
This will help us create valid letsencrypt TLS certficates for ${FQDN}

Related to https://github.com/kubeflow/kubeflow/issues/255

### DIFF
--- a/kubeflow/core/iap.libsonnet
+++ b/kubeflow/core/iap.libsonnet
@@ -9,9 +9,9 @@
     // We do this because updating the ingress causes the backend service to change which disables IAP
     // and changes the backend service which is used for the JWT audience.
     // So we want to avoid updating the ingress when updating the Envoy pods or other backend services.
-    ingressParts(secretName, ipName):: std.prune(k.core.v1.list.new([
+    ingressParts(secretName, ipName, hostname):: std.prune(k.core.v1.list.new([
       $.parts(namespace).service,
-      $.parts(namespace).ingress(secretName, ipName),
+      $.parts(namespace).ingress(secretName, ipName, hostname),
     ])),
 
     envoy(envoyImage, audiences, disableJwt):: std.prune(k.core.v1.list.new([
@@ -486,7 +486,7 @@
       },
     },
 
-    ingress(secretName, ipName):: {
+    ingress(secretName, ipName, hostname):: {
       apiVersion: "extensions/v1beta1",
       kind: "Ingress",
       metadata: {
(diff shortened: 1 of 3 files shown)


kubeflow/core/iap.libsonnet
```diff
@@ -499,6 +499,7 @@
       spec: {
         rules: [
           {
+            [if hostname != "null" then "host"]: hostname,
```
lluunn: looks weird. guess there is a typo?

kubeflow/core/iap.libsonnet
```diff
@@ -499,6 +499,7 @@
       spec: {
         rules: [
```
lluunn: What's the effect of ingress having a host?

kubeflow/core/iap.libsonnet
ankushagarwal: No, that's correct. I wanted to use the parameter name `hostname` for `iap.jsonnet` component since it is more obvious.

k8s Ingress object uses the `host` key instead of `hostname` in it's spec.

kubeflow/core/iap.libsonnet
ankushagarwal: If we configure ingress with a hostname, we can leverage https://github.com/jetstack/cert-manager to automatically issue and renew valid SSL certificates for this ingress.

ankushagarwal: Updated the PR description as well.

kubeflow/core/iap.libsonnet
lluunn: I mean, if hostname == "null", this line will render
: null

Shouldn't it be 
[if hostname != "null" then "host": hostname]

kubeflow/core/iap.libsonnet
lluunn: thanks

kubeflow/core/iap.libsonnet
ankushagarwal: The syntax `[...]` means that it is a computed field in jsonnet. if hostname == "null", then the field will be omitted.

Look at Computed and Optional Fields section in http://jsonnet.org/docs/tutorial.html

lluunn: Ah I see, thanks

lluunn:
You need to merge master to fix the test.
/lgtm

ankushagarwal:
@lluunn Can you lgtm again?

PR #431, opened 2018-03-15
Libsonnet cleanup

### LINKED ISSUE #417: normalize libsonnets so their "all(params)" is only called from kubeflow/core/all.libsonnet
Per @jlewi 
```
          // TODO(jlewi): We should make `all` top level within each libsonnet file and
          // not a field within parts.
```

kkasravi:
/assign @ankushagarwal 

fixes #417

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/431)


kubeflow/core/jupyterhub.libsonnet
```diff
@@ -16,6 +14,16 @@
   ],
 
   parts(namespace):: {
+    jupyterHubConfigMap(jupyterHubAuthenticator, disks): {
+      local util = import "kubeflow/core/util.libsonnet",
+      local diskNames = util.toArray(disks),
+      local kubeSpawner = $.parts(namespace).kubeSpawner(jupyterHubAuthenticator, diskNames),
+      local cm = if std.length(diskNames) == 0 then
+        $.parts(namespace).jupyterHubConfigMapWithoutVolumes
```
ankushagarwal: jupyterHubConfigMapWithoutVolumes takes a parameter named spawner, but here you are not passing any parameter to jupyterHubConfigMapWithoutVolumes

kkasravi: both parts of the if logic return a function which is then called with spawner on line # 24. 
Original logic is similar.

ankushagarwal: The first part of the if logic returns a function which takes `spawner` as a parameter, but the second part `$.parts(namespace).jupyterHubConfigMapWithVolumes(diskNames)` returns an object, not a function.

kkasravi:
@ankushagarwal the logic related to jupyterHubConfigMapWithVolumes was never used or inserted. Instead params.disks is inserted via kubeSpawner and becomes part of jupyterhub_spawner.py.

PR #434, opened 2018-03-15
Run ks upgrade

lluunn:
We updated the ks version to 0.9, and we need this fix e2e tests

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/434)


ankushagarwal:
Since we're here, can you also upgrade these ksonnet apps:

* components/tensorflow-notebook-image/releaser/app.yaml
* releasing/releaser/app.yaml

lluunn:
It's working differently now.
We set the name of workflow, but it doesn't take effect:

 param set --env=kubeflow-presubmit-kubeflow-e2e-434-3834d8f-632-c7a9 workflows name kubeflow-presubmit-kubeflow-e2e-434-3834d8f-632-c7a9 

apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  name: jlewi-kubeflow-kubeflow-presubmit-test-227-643b
  namespace: kubeflow-test-infra

So the param are not set correctly.

lluunn:
Ah it's https://github.com/ksonnet/ksonnet/issues/346. Should use 0.9.1

PR #437, opened 2018-03-16
Add instructions to obtain valid certificates to iap.md

ankushagarwal:
Related to https://github.com/kubeflow/kubeflow/issues/255

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/437)

### DIFF
--- a/docs/gke/iap.md
+++ b/docs/gke/iap.md
@@ -162,7 +162,7 @@ gcloud projects add-iam-policy-binding $PROJECT \
   --member user:${USER_EMAIL}
 ```
 
-### Self signed certificates and browser security warnings
+### Obtaining a valid certificate from LetsEncrypt
 
 Since you are using a self signed certificate chrome and other browsers will give you a warning like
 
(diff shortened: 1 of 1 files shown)


docs/gke/iap.md
```diff
@@ -162,18 +162,74 @@ gcloud projects add-iam-policy-binding $PROJECT \
   --member user:${USER_EMAIL}
 ```
 
-### Self signed certificates and browser security warnings
+### Obtaining a valid certificate from LetsEncrypt
```
lluunn: Is there any requirement so that we can use this? Or anyone can just use it? Is it free?

docs/gke/iap.md
```diff
@@ -162,18 +162,74 @@ gcloud projects add-iam-policy-binding $PROJECT \
...
 
 ```
 Attackers might be trying to steal your information from ${ENDPOINT}(for example, passwords, messages, or credit cards). Learn more
 NET::ERR_CERT_AUTHORITY_INVALID
 ```
-  * You will need to ignore these warnings
-  * To avoid these warnings you will need to use a certificate signed by a signing authority
-  * [Lets Encrypt](https://letsencrypt.org/) and other sites provide free signed certificates.
 
+We can use [cert-manager](https://github.com/jetstack/cert-manager) to automatically obtain and refresh LetsEncrypt certificates.
```
lluunn: Could you add a link or explanation for LetsEncrypt

docs/gke/iap.md
ankushagarwal: It's free - anyone can use it.

docs/gke/iap.md
ankushagarwal: Done

ankushagarwal:
/hold

https://github.com/kubeflow/kubeflow/pull/441 seems to be a better solution to this

ankushagarwal:
Closing in favor of #441

PR #438, opened 2018-03-16
Use a CentOS 7 base image for the jupyterhub-k8s image

tmckayus:
There are some notes about motivation on this similar PR for tf-operator

https://github.com/kubeflow/tf-operator/pull/469

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/438)

### DIFF
--- a/components/jupyterhub/docker/Dockerfile
+++ b/components/jupyterhub/docker/Dockerfile
@@ -12,10 +12,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM python:3.6
+FROM centos:centos7
 
-RUN apt-get update && \
-    apt-get install -y npm nodejs-legacy
+RUN yum update -y && \
+    yum install -y epel-release && \
+    yum install -y python36 npm nodejs010 python34-pip && \
+    yum clean all -y
 
 RUN npm install -g configurable-http-proxy && \
     pip3 install --no-cache-dir \


tmckayus:
Tested manually but this can use some more testing / eyes

bhack:
I agree with https://github.com/kubeflow/tf-operator/pull/469#issuecomment-374841983

jlewi:
/unassign @gaocegege 
/assign jlewi

As discussed on  kubeflow/tf-operator#469
I think its fine to do (centos + runtime).

The explanation in that bug is we can always replace that in the future with a more slimmed down image; but as willb@ noted on another of these issues I think the centos images might be smaller. So unless we're going to switch to a truly minimal image (like alpine) it doesn't make much of a difference.

@tmckayus Can you fix the test and this should be good to go?

tmckayus:
@jlewi thanks, yes, I'll try to get to the test this week

bhack:
@jlewi I still suggest to handle this in `Dockerfile.<distro>` like in [other projects](https://github.com/NVIDIA/nvidia-docker?files=1).
This are [the sizes](https://hub.docker.com/r/library/python/tags/) of all the official python images.
I still belive that most of the users expect to have the default Dockerfile with apt commands.

tmckayus:
@jlewi  okay, looks like it just needed a rebase after all :)

jlewi:
@bhack Are you just suggesting that the file should be named `Dockerfile.centos` or are you suggesting that users expect ubuntu\apt as the default image?

bhack:
Both. But In the case I linked they are managed by build tools https://github.com/NVIDIA/nvidia-docker/blob/master/Makefile

jlewi:
@tmckayus Do you mind renaming the file to Dockerfile.centos per @bhack's suggestion.

inc0:
Alternative would be to use things like Alpine - lightweight barebone distro

tmckayus:
I would push back on this a little bit and say that if we have multiple Dockerfiles for different distros, then the community is committing to supporting them all. Any time someone makes a change to one, they are responsible for ensuring that all images are updated if necessary and that there is no breakage.  Additionally, we have to expand the test and release and make sure all of them are covered.

Is this something we want to do? If so, I'd still suggest merging the PR as is with a CentOS default for all of the reasons previously discussed in https://github.com/kubeflow/tf-operator/pull/469 and then follow up with separate PRs to exand test/release scripts to multiple Dockerfiles and include others at that point.

bhack:
Nvidia is doing it: https://hub.docker.com/r/nvidia/cuda/tags/. @tmckayus As you are a vendor I think that any vendor can support and be the maintainer/owner of his Dockerfile.<distro> file in this repo.

jlewi:
Actually why not use the official JupyterHub image as our base image and install any addons like the KubeSpawner into that?

Per #479 official JupyterHub images are here:
https://hub.docker.com/r/jupyterhub/jupyterhub

bhack:
@jlewi It is ok for me but I think that the `FROM` is still ubuntu in jupyterhub images. I think here the the topic is more if vendors want to maintain its own distro image and related build/push in this repository.

jlewi:
@bhack Understood; but I think the topic off what needs to happen so vendors can offer supported versions of Kubeflow goes well beyond the scope of this PR and picking base images.

I don't see a need for us in the short term to take on the burden of building JupyterHub images. I believe this should be handled by the JupyterHub project.

I think this is different from TFJobOperator because TFJobOperator is owned/maintained by Kubeflow. So its up to us to decide what the appropriate image should be.

For packages built and maintained outside Kubeflow (e.g. Pachyderm, Seldon, JupyterHub, Argo, etc...) I think the question of base image should be handled by those repositories.

bhack:
@jlewi You are right but the scope of this PR was to make different images that we use CentOS default. See also https://github.com/kubeflow/kubeflow/pull/438
So I think that we need a policy for vendors for PR like this one. IMHO the default upstream image is ok for me.

bhack:
P.s. And also https://github.com/kubeflow/tf-operator/pull/469 that we merged as CentOS default

tmckayus:
@jlewi, so if we had a jupyterhub base image supported outside of kubeflow that this image could be based on, would that satisfy?

Also, note, CentOS is not a vendor, it's a community.

elmiko:
> So I think that we need a policy for vendors for PR like this one. IMHO the default upstream image is ok for me.

@bhack CentOS is not a vendor but a community.

for me, centos is a much more solid base image to create cloud native tooling from. so, the default is not ok for me, especially when considering security and auditing for images i'd prefer a centos base.

bhack:
I know that centos is not a vendor but I was referring in general about vendors. If will come Canonical for pushing Ubuntu waht we will do? A distro flame? Ubuntu is a vendor? Is a community? Or both? Then I only affirmed that @tmckayus is a vendor but I don't know if he was acting as a Centos community member or Redhat employee.

elmiko:
i feel there are 2 issues here though, choosing a solid base image for our work, and "vendor" sponsored images.

imo, we should use centos as the base for cloud images as i feel that community is doing more to prepare the distro for secure cloud-based images.

if, or when, we have actualy vendors showing up saying "hey, kubeflow needs more ubuntu based images" then we could address that side of the argument. but for now, i think we are positioned better to have centos as the root.

just my two cents

jlewi:
@elmiko I think that is a reasonable position and for images that are produced and maintained by Kubeflow it makes sense to decide as a community what  base image to use. That's why for TFJob operator (kubeflow/tf-operator#469) it seemed reasonable to switch to a cent-os based image

But a lot of the packages included in Kubeflow today
- Argo
- Pachyderm
- JupyterHub
- Seldon

Are packages owned and maintained outside of Kubeflow. I don't think we have the resources at this point in time to take on the work of rebuilding images for all these packages in order to have a consistent  base image. We already have our hands full just building a release process for packages owned by Kubeflow.

I definitely want to understand how we can support vendors but that's probably a much bigger problem.

For JupyterHub, it seems like we can avoid building docker images altogether (at least for the time being) and just use the official JupyterHub images. That seems preferable IMO to taking on the added cost of building and maintaining another image.

elmiko:
> I don't think we have the resources at this point in time to take on the work of rebuilding images for all these packages in order to have a consistent base image. We already have our hands full just building a release process for packages owned by Kubeflow.

++, seems reasonable

> I definitely want to understand how we can support vendors but that's probably a much bigger problem.

agreed

> For JupyterHub, it seems like we can avoid building docker images altogether (at least for the time being) and just use the official JupyterHub images. That seems preferable IMO to taking on the added cost of building and maintaining another image.

this makes more sense to me, if we aren't going to commit resources to maintaining Dockerfiles and images from this repo then using the upstream is very sensible imo.

jlewi:
Closing this PR since its stale. There are proposals floating around in the community on how to address this problem.

PR #442, opened 2018-03-16
E2e test for TF serving with GPU

### LINKED ISSUE #291: E2E Test for TFServing with GPUs

lluunn:
Fix #291 
- refactor workflow to avoid code duplication
- Append cpu/gpu to the workflow name and junit name so that they can be distinguished
- Added result file for GPU. It's slightly different (0.0001) than CPU result (not sure why)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/442)

### DIFF
--- a/components/k8s-model-server/images/releaser/components/workflows.libsonnet
+++ b/components/k8s-model-server/images/releaser/components/workflows.libsonnet
@@ -357,39 +357,11 @@
             buildTemplate(
               "deploy-tf-serving",
               deploy_tf_serving_command,
-              [
-                {
-                  name: "DOCKER_HOST",
-                  value: "127.0.0.1",
-                },
-              ],
-              [{
-                name: "dind",
-                image: "docker:17.10-dind",
-                securityContext: {
-                  privileged: true,
-                },
-                mirrorVolumeMounts: true,
-              }],
-            ),  // deploy-tf-serving
+            ),
             buildTemplate(
               "deploy-tf-serving-gpu",
               deploy_tf_serving_gpu_command,
-              [
-                {
-                  name: "DOCKER_HOST",
-                  value: "127.0.0.1",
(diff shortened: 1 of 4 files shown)


lluunn:
presubmit shows succeeded, but the test failed...

lluunn:
/cc @jlewi 
PTAL

components/k8s-model-server/images/releaser/components/workflows.libsonnet
```diff
@@ -313,40 +372,30 @@
                 mirrorVolumeMounts: true,
               }],
             ),  // deploy-tf-serving
+            buildTemplate(
+              "deploy-tf-serving-gpu",
+              deploy_tf_serving_gpu_command,
+              [
+                {
```
jlewi: Are these environment variables correct? Why do you need to use docker in docker when you deploy TF serving?

testing/test_deploy.py
```diff
@@ -224,7 +223,7 @@ def teardown(args):
   core_api.delete_namespace(args.namespace, {})
 
 def determine_test_name(args):
-  return args.func.__name__
+  return args.func.__name__ + "-" + args.deploy_name if args.deploy_name else args.func.__name__
```
jlewi: Make this a multi-line if/else statement; its easier to read then squashing it on a single line.

components/k8s-model-server/images/releaser/components/workflows.libsonnet
lluunn: We needed because we were building the image. Removing now.

lluunn:
updated the PR description, and squashed.
PTAL, thanks

Issue #432, opened 2018-03-15
Proposal: We should have basic tests for every ksonnet prototype


ankushagarwal:
It's hard to review large PRs which involve jsonnet/libsonnet refactorings. Since jsonnet/libsonnet files are hard to write, it is easy to make small mistakes which can cause entire prototypes to be broken.

It will be good if we can have simple sanity tests for our ksonnet prototypes.

Our current presubmit tests only few ksonnet prototypes and they do an end-to-end testing.

It will good to have simple tests. The test for a prototype can be as simple as a shell script

```bash
#!/bin/bash
set -ex

# test-iap-envoy.sh

ks generate iap-envoy iap-envoy1 --namespace=ns --audiences=audiences
ks generate iap-envoy iap-envoy2 --namespace=ns --audiences=audiences --optionalParam=value
```

This is a simple test which tests that we are able to successfully generate components from the iap-envoy prototype and the parameters that we've declared in the prototype are valid.

ankushagarwal:
These tests should be simple enough that a user should be able to run them locally without relying on a k8s cluster.

ankushagarwal:
`ks validate` communicates the generated configs with the k8s server to validate configs. It requires a valid k8s cluster pointed to by the ks environment. 

My proposal is just to have 'compilation' tests - test if we are able to compile the *sonnet files successfully. A collaborator should be able to run them locally - the only dependency required should be `ks`

kkasravi:
@ankushagarwal What calls kubeflow/core/tests/util_test.jsonnet? I added one for jupyterhub - jupyterhub_test.jsonnet but i had to build jsonnet in order to run it. Not sure if I should check this in. Having jsonnet available would allow much better unit tests ...

ankushagarwal:
Right now it is manually run locally by running `jsonnet eval <test_file_name>`. We plan to run these as part of presubmit tests : https://github.com/kubeflow/kubeflow/issues/254

kkasravi:
OK - should I add a jupyterhub_test.libsonnet here that you could review? An option would be to have these for each library which I could do... Let me know your thoughts

ankushagarwal:
@kkasravi Yes, that would be great!

kkasravi:
[...]

PR #443, opened 2018-03-17
Run jsonnet tests as part of kubeflow-e2e workflow

ankushagarwal:
* Create test_jsonnet.py - it goes through all the jsonnet tests and executes it
* Add test_jsonnet.py to kubeflow-e2e workflow

Related to #432

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/443)


jlewi:
This is fantastic! Thank you so much for doing this.

/lgtm
/approve

PR #444, opened 2018-03-17
set termination policy to worker 0 by default when no master exists

raddaoui:
most tensorflow training examples use worker 0 as the chief worker instead of master. 
using ksonnet tf-job prototype to create training jobs dosen't allow that
and will make the tfjob fails if we set num_masters to 0.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/444)


jlewi:
Thanks!
/lgtm
/approve

Issue #428, opened 2018-03-14
Remove tensorboard link from Jupyter notebooks (hub?)


jlewi:
See #422 

There is an icon in Jupyter notebooks to launch tensorboard. We should remove that as using Jupyter to launch TensorBoard is no longer the recommended approach.

I think the initial idea was to use Jupyter/JupyterHub as a way to proxy connections to tensorboard. But we have a better reverse proxy solution with TensorBoard.

The plan moving forward is to add ksonnet components to make it easy to launch TensorBoard as a K8s deployment #297.

jlewi:
I think this is where we install tensorboard
https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/Dockerfile#L186

ernestmartinez:
Is there any procedure I can use in the interim to run Tensorboard?

jlewi:
@ernestmartinez My recommendation is to spawn TensorBoard manually as a K8s deployment per #297 
You can use this as an example
https://github.com/kubeflow/examples/pull/1/files#diff-b1c94ace60732530b2be5b0a67be2d40

kkasravi:
see #446

ernestmartinez:
Excuse my ignorance, but I'm not quite sure how to use the above link. I'm assuming it's a ks manifest, but I'm not quite sure if I should be installing a pkg, or applying it directly. 

Directions would be much appreciated.

jlewi:
@ernestmartinez The above link is not a package its a component. You could copy it into the components directory of your ksonnet app and then use it.

Issue #297, opened 2018-02-26
ksonnet prototypes for TensorBoard


jlewi:
We should provide some ksonnet components for tensorboard.

There are some examples in this PR
https://github.com/kubeflow/examples/pull/1/files#diff-b1c94ace60732530b2be5b0a67be2d40

The main thing is in addition to deploying TB, we'd like to add an ambassador route.

abkosar:
Hey @jlewi, I'm new to contributing to open source. I came across this issue and I would like to work on this. Can you please assign this to me?

jlewi:
Consider the issue assigned to you @abkosar 

I sent you an invite to the Kubeflow org; I think you need to accept in order for me to actually assign the issue to you.

abkosar:
Just accepted the invite @jlewi 
Thanks!

jlewi:
Welcome to the team!

wbuchwalter:
Hey @abkosar, any progress to share on this?

abkosar:
Hey @wbuchwalter, like we talked on Slack, still working on the issue, thanks for asking and help!

jlewi:
@abkosar Any update on this?

We have a .libsonnet and .jsonnet file checked in here:
https://github.com/kubeflow/examples/tree/master/agents/app/components

So  we just need to move the .libsonnet file to
kubeflow/kubeflow/core 

And the jsonnet file to 
kubeflow/kubeflow/core/prototypes
and refactor it a bit to make it a prototype

Any chance we could get a PR done this week?

abkosar:
@jlewi Just created the pull request following the directions. Tried to refactor the .jsonnet file  too but not sure if I refactored it correctly. 

But here is the link to the request: [#470](https://github.com/kubeflow/kubeflow/pull/470)

I'll appreciate any help and feedback. 

Thanks!

jlewi:
Fantastic thank you.

abkosar:
Hey @jlewi -- Updated the code with your directions and pushed it. Hopefully went a step forward, will appreciate any feedback.

Thanks!

PR #446, opened 2018-03-17
remove the pip install tensorboard

kkasravi:
/assign @jlewi

This is for the issue #428 "remove tensorboard link from jupyter" and #297 "add prototypes for TensorBoard". In general we're utilizing the ksonnet way to launch tensorboard rather than using a button in jupyterhub.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/446)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -183,7 +183,6 @@ SHELL ["/bin/bash", "-c"]
 RUN git clone https://github.com/tensorflow/models.git /home/$NB_USER/tensorflow-models && git clone https://github.com/tensorflow/benchmarks.git /home/$NB_USER/tensorflow-benchmarks
 # Import matplotlib the first time to build the font cache.
 ENV XDG_CACHE_HOME /home/$NB_USER/.cache/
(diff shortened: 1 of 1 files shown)


kkasravi:
screen shot 2018-03-17 at 10 16 57 am


this is a screen shot from a container i built for tensorflow-notebook-cpu and uploaded to hub.docker.com https://hub.docker.com/r/nervana/tensorflow-notebook-cpu/tags/

jlewi:
Thanks for this.

Could you update the PR description to explain why we ware making this change and link to the relevant issues? i.e. #428  to remove the link and #297 to add prototypes for TensorBoard.

jlewi:
@pdmack I think you helped a customer recently get this working. Any reason we should keep this rather than telling customers to directly launch TB and potentially providing a UI for that if we think one is needed?

kkasravi:
@jlewi updated the PR description

pdmack:
@jlewi Sorry, too slow. I think clear instructions for accessing TB will be fine. The user wasn't aware that the menu item was there, and therefore tried to launch it from the NB terminal. Also, they were using a spawned image which had TB launch issues (even from the menu).

PR #463, opened 2018-03-20
Update docs so that namespace is only set at ks env level

ankushagarwal:
Now that https://github.com/kubeflow/kubeflow/pull/426 is merged, we do
not need to set namespace at the component level. namespace will be
automatically inherited by the component from env.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/463)


jlewi:
Can you update the Requirements section to reflect the version of ksonnet required?
https://github.com/kubeflow/kubeflow#requirements

ankushagarwal:
Done.

PR #469, opened 2018-03-20
Add parameter to mount path for notebook PVC

### LINKED ISSUE #365: Option to disable use of PVC for Jupyter
See #336 on clusters without a default storage class; spawning of Jupyter pods will get stuck with pods in the container creating state waiting for a PV.

We should add options to disable the use of PVC for Jupyter to support these customers.

@ChanYiLin suggested following the method in
How about follow the method in ksonnet redis line 151 https://github.com/ksonnet/parts/blob/master/incubator/redis/redis.libsonnet

It looks like storage class defaults to "-"; does the "-" have some special meaning in K8s?

P1 because as #331 this is causing problems for some customers (although a workaround is noted in #336).

@inc0 can you take this?

inc0:
This change allows command ks param set kubeflow-core
jupyterNotebookPVCMount << mount path >>. If you specify empty string
("") there it will remove PVC creation at all, which is useful in envs
without proper StorageClass.

Fixes #365

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/469)

### DIFF
--- a/kubeflow/core/jupyterhub.libsonnet
+++ b/kubeflow/core/jupyterhub.libsonnet
@@ -182,7 +182,6 @@ c.RemoteUserAuthenticator.header_name = 'x-goog-authenticated-user-email'",
                   {
                     name: "NOTEBOOK_PVC_MOUNT",
                     value: notebookPVCMount
-
                   },
                 ]
               },  // jupyterHub container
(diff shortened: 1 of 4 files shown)


jlewi:
Looks good except for the failing test.

jlewi:
I think test flakes might have been fixed please try syncing.

jlewi:
Looks like its the jsonnet_test that's still failing.
#508 includes some updates to the test that should make it easier to keep the test up to date and might help in this PR.

jlewi:
Are we going to submit this PR or #503 

Can you try rebasing and squashing the commits to see if that fixes the CLA check?

/assign @kkasravi

inc0:
I'll close this one and let's work on #503

PR #476, opened 2018-03-21
Fix ambassador_test.jsonnet

ankushagarwal:
It was broken by #396

/cc @wbuchwalter

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/476)


ankushagarwal:
/cc @jlewi 
/cc @lluunn 
Adding more folks since this is blocking all PRs

PR #480, opened 2018-03-21
Update jsonnet tests so that it tests all files under kubeflow

ankushagarwal:
* This will catch any syntax errors in *sonnet files.
* Skip all ksonnet prototypes since they need additional deps

/cc @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/480)

### DIFF
--- a/testing/test_jsonnet.py
+++ b/testing/test_jsonnet.py
@@ -42,8 +42,8 @@ from kubeflow.testing import util
 # except ksonnet prototype definitions - they require additional
 # dependencies
 def should_test(f):
-  parts = os.path.splitext(f)
-  if parts[1] != '.jsonnet' and parts[1] != '.libsonnet':
+  _, ext = os.path.splitext(f)
+  if ext != '.jsonnet' and ext != '.libsonnet':
     return False
   parts = f.split('/')
   if len(parts) < 2:
--- a/testing/test_jsonnet.py
+++ b/testing/test_jsonnet.py
@@ -38,30 +38,32 @@ import argparse
 from kubeflow.testing import test_util
 from kubeflow.testing import util
 
-# is_ksonnet_prototype checks if the directory which
-# contains the file f is equal to 'prototypes'
-def is_ksonnet_prototype(f):
+# We should test all files which end in .jsonnet or .libsonnet
+# except ksonnet prototype definitions - they require additional
+# dependencies
+def should_test(f):
+  parts = f.split('.')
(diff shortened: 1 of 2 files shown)


testing/test_jsonnet.py
```diff
@@ -38,18 +38,32 @@
 from kubeflow.testing import test_util
 from kubeflow.testing import util
 
+# We should test all files which end in .jsonnet or .libsonnet
+# except ksonnet prototype definitions - they require additional
+# dependencies
+def should_test(f):
+  parts = f.split('.')
```
lluunn: os.path.splitext

testing/test_jsonnet.py
```diff
@@ -38,18 +38,32 @@
 from kubeflow.testing import test_util
 from kubeflow.testing import util
 
+# We should test all files which end in .jsonnet or .libsonnet
+# except ksonnet prototype definitions - they require additional
```
lluunn: Aren't the files in prototypes jsonnet, and otherwise libsonnet?

testing/test_jsonnet.py
ankushagarwal: Done

testing/test_jsonnet.py
ankushagarwal: We cannot test the prototypes directly since they are not valid jsonnet. For example: ksonnet injects the `params` and the `env` variable into the prototype when a component is generated from that prototype.

The idea is that we want to test everything(jsonnet and libsonnet file) under the kubeflow/ directory unless it is a ksonnet prototype.

lluunn: I mean, is the following true:
.jsonnet --> ksonnet prototype, cannot test
.libsonnet --> valid, can test

I manually checked, seems to be the case?

ankushagarwal: Yes. We shouldn't test any .jsonnet under .../prototypes/ since they are ksonnet prototypes.

testing/test_jsonnet.py
```diff
@@ -38,18 +38,32 @@
...
+# dependencies
+def should_test(f):
+  parts = os.path.splitext(f)
```
lluunn: _, ext = os.path.splitext(f)
Then use ext as parts[1] below.

I mean, shouldn't we test only libsonnet? So should_test will be return ext == '.libsonnet'

PR #488, opened 2018-03-23
Update buildTemplateImage to take workflow_name

ankushagarwal:
* Argo workflow names cannot have `.` character, so use `-` as a separator.

* jsonnet 0.9.5 does not have the `std.strReplace` function, so we need to
pass workflow_name as a separate parameter. With `std.strReplace`, we 
could have used `workflow_name = std.strReplace(tf_version, '.', '-')`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/488)


lluunn:
I disabled gpu test, sync should fix the test.
/lgtm

PR #471, opened 2018-03-20
Update tensorflow-notebook-image workflow to pin tf and cuda version

ankushagarwal:
* Build 1.4.1, 1.5.1, 1.6.0 versions for gpu and cpu
* Update ksonnet version
* Update bazel version
* Built all images locally to test it.

The images will have the following URI format
```
gcr.io/kubeflow-images-staging/tensorflow-1.4.1-notebook-gpu:20180320-afw9aow
gcr.io/kubeflow-images-staging/tensorflow-1.4.1-notebook-cpu:20180320-afw9aow
```

Fixes #375

Fixes #467 (Multiple Jupyter images for different TF Versions).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/471)

Issue #467, opened 2018-03-20
Build Jupyter Notebook images for supported versions of TF


jlewi:
We should start building/maintaining multiple Jupyter images corresponding to different versions of TF. 

Related #375

PR #491, opened 2018-03-23
Update default images for jupyterhub spawner

ankushagarwal:
Provide a dropdown with images for each tensorflow minor version.

We now pin down a tensorflow version for our jupyterhub image builds.

Related #375 #467 #471

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/491)


pdmack:
/lgtm

Any ideas on parameterizing these?

ankushagarwal:
I haven't given much thought to parametrizing these since it's only 6 images.

ankushagarwal:
After https://github.com/kubeflow/kubeflow/pull/471 we are no longer building tensorflow nightly builds.

Issue #484, opened 2018-03-22
TF serving GPU e2e test is flaky


lluunn:
Example: #482 the test failed in a doc PR

**In the step deploy-tf-serving-gpu, it verified deployment is up.**
INFO|2018-03-22T19:11:44|/mnt/test-data-volume/kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67/src/kubeflow/testing/py/kubeflow/testing/util.py|263| Deployment inception-gpu in namespace kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67 is ready

**However, in test-tf-serving-gpu step, it failed:**
INFO|2018-03-22T19:11:48|/mnt/test-data-volume/kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67/src/kubeflow/kubeflow/testing/test_tf_serving.py|99| prediction failed: AbortionError(code=StatusCode.UNAVAILABLE, details="Connect Failed"). Retrying...
INFO|2018-03-22T19:11:53|/mnt/test-data-volume/kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67/src/kubeflow/kubeflow/testing/test_tf_serving.py|99| prediction failed: AbortionError(code=StatusCode.UNAVAILABLE, details="Connect Failed"). Retrying...
INFO|2018-03-22T19:12:04|/mnt/test-data-volume/kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67/src/kubeflow/kubeflow/testing/test_tf_serving.py|116| Writing test results to /mnt/test-data-volume/kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67/output/artifacts/junit_kubeflow-tf-serving-image-inception-gpu.xml
INFO|2018-03-22T19:12:04|/mnt/test-data-volume/kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67/src/kubeflow/testing/py/kubeflow/testing/test_util.py|151| Creating /mnt/test-data-volume/kubeflow-presubmit-tf-serving-image-482-70f83cc-790-8b67/output/artifacts/junit_kubeflow-tf-serving-image-inception-gpu.xml
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
[...]

PR #497, opened 2018-03-23
Enable tf serving gpu test

lluunn:
The test is flaky: #484
However, it's hard to reproduce the error, so I am going to re-enable it and see.

/cc @jlewi 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/497)


lluunn:
It's really hard to repro the failure...
I am guessing that because we have limited GPUs in the cluster, and there were multiple presubmits running, some of the test timeout waiting for the resource? @jlewi 
Maybe we don't run this test for every presubmit, but only for postsubmit or periodic runs? WDYT?

Issue #506, opened 2018-03-26
Cut a 0.1 release


jlewi:
Tracking bug for cutting a 0.1 release

TODO

- [X] Update the TFJob image to pull in fixes to kubeflow/tf-operator#500
- [X] Create a tag/branch for our ksonnet configs
- [X] Update the Jupyter Images that our configs point to.
- [X] Update Docs to point to branch

Things to verify manually

- [ ] IAP works 
- [ ] TFJobs UI works
- [ ] Can spawn notebooks
- [ ] TFCNN runs

jlewi:
Branch has been cut
https://github.com/kubeflow/kubeflow/tree/v0.1-branch

jlewi:
Create a release candidate
https://github.com/kubeflow/kubeflow/releases/tag/v0.1.0-rc.0

jlewi:
We should cherry pick #548 which fixes the TFJob UI ksonnet configs into the component.

pdmack:
kubeflow/kubeflow#541

jlewi:
Its official v0.1.0
https://github.com/kubeflow/kubeflow/releases/tag/v0.1.0

Will send a PR updating the docs.

PR #508, opened 2018-03-26
Update images for our 0.1 release

jlewi:
* Update the Jupyter notebook images used to the most recently built
  Jupyter images.

* Cut a new release of TFJob operator and update the Kubeflow components to
  use it; The new release fixes a bunch of bugs in the operator.

* Update the instructions for releasing.

* Related to #506

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/508)

### DIFF
--- a/kubeflow/core/prototypes/all.jsonnet
+++ b/kubeflow/core/prototypes/all.jsonnet
@@ -7,7 +7,7 @@
 // @optionalParam disks string null Comma separated list of Google persistent disks to attach to jupyter environments.
 // @optionalParam cloud string null String identifying the cloud to customize the deployment for.
 // @optionalParam tfAmbassadorServiceType string ClusterIP The service type for the API Gateway.
-// @optionalParam tfJobImage string gcr.io/kubeflow-images-staging/tf_operator:v20180226-403 The image for the TfJob controller.
+// @optionalParam tfJobImage string gcr.io/kubeflow-images-staging/tf_operator:v20180326-6214e560 The image for the TfJob controller.
 // @optionalParam tfDefaultImage string null The default image to use for TensorFlow.
 // @optionalParam tfJobUiServiceType string ClusterIP The service type for the UI.
 // @optionalParam jupyterHubServiceType string ClusterIP The service type for Jupyterhub.
--- a/releasing.md
+++ b/releasing.md
@@ -23,19 +23,24 @@ Run the E2E test workflow using our release cluster
 [kubeflow/testing#42](https://github.com/kubeflow/testing/issues/42) will simplify this.
 
 ```
-PULL_BASE_SHA=<commit to build>
-DATE=`date +%Y%m%d`
-VERSION_TAG="v${DATE}-${PULL_BASE_SHA}"
 JOB_NAME="tf-operator-release"
-JOB_TYPE=postsubmit
+JOB_TYPE=tf-operator-release
 BUILD_NUMBER=$(uuidgen)
 BUILD_NUMBER=${BUILD_NUMBER:0:4}
 REPO_OWNER=kubeflow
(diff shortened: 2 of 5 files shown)


inc0:
Don't we want nightly as well?

releasing.md
```diff
@@ -120,6 +125,9 @@ ks param set --env=${ENV} workflows prow_env  \
 ks apply ${ENV} -c workflows
 ```
 
+Create a PR to update [jupyterhub_spawner.py](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/jupyterhub_spawner.py#L15) 
+to point to the newly built Jupyter notebook images.
```
lluunn: also update all.jsonnet default value of tfJobImage?

kubeflow/core/prototypes/all.jsonnet
```diff
@@ -7,7 +7,7 @@
 // @optionalParam disks string null Comma separated list of Google persistent disks to attach to jupyter environments.
 // @optionalParam cloud string null String identifying the cloud to customize the deployment for.
 // @optionalParam tfAmbassadorServiceType string ClusterIP The service type for the API Gateway.
-// @optionalParam tfJobImage string gcr.io/kubeflow-images-staging/tf_operator:v20180226-403 The image for the TfJob controller.
```
inc0: we need to find better way for that. Can we use latest rather than fixed revision? If we need to quickly patch something this will not be optimal.

releasing.md
```diff
@@ -23,19 +23,24 @@ Run the E2E test workflow using our release cluster
 [kubeflow/testing#42](https://github.com/kubeflow/testing/issues/42) will simplify this.
 
 ```
-PULL_BASE_SHA=<commit to build>
-DATE=`date +%Y%m%d`
-VERSION_TAG="v${DATE}-${PULL_BASE_SHA}"
 JOB_NAME="tf-operator-release"
-JOB_TYPE=postsubmit
+JOB_TYPE=tf-operator-release
```
pdmack: So is this also a predefined type like presubmit, postsubmit? The type value matches the job name so just trying to understand.

jlewi: The job type is defined by prow.
https://github.com/kubernetes/test-infra/tree/master/prow

The reason we set job type here is because the scripts invoked by our E2E tests e.g.
https://github.com/kubeflow/testing/blob/master/images/checkout.sh

Might be using JOB_TYPE. For example, checkout.sh might be using JOB_TYPE to determine whether its a post submit or presubmit job which affects how the script checks out the code (e.g. do we add refs to pull the PR.)

Our checkout.sh script doesn't use JOB_TYPE right now so technically we'd be fine not setting it. But we do anyway because the contract that checkout.sh is built around is that JOB_TYPE will be set.

kubeflow/core/prototypes/all.jsonnet
jlewi: I think we want to pin to a specific version. The whole point of cutting a release 0.1 is to provide users a fixed and reproducible environment. So if we were to push a new image and then users automatically started using that image because images were tagged to latest, then behavior would change for users in unexpected ways.

If users want to use "latest" and not pin then they can do that by overriding the image.

If we need to release a fix we would do this by cutting a new release; e.g. 0.1.1
So we would build a new TF operator image and publish it. We would then update the kubeflow components and then cut that as 0.1.1 and tell people to use the 0.1.1 tag and not the 0.1 tag to get the fix.

At least that's my current opinion about how it should work. The overall release process and compatibility guarantees are still being thought out, so its possible we will do something different.

releasing.md
jlewi: Done.

Issue #477, opened 2018-03-21
 ks env set default --namespace=kubeflow doesn't change the namespace


jlewi:
Run the following steps per our instructions

```
ks env set default --namespace=kubeflow
ks env list

NAME    KUBERNETES-VERSION NAMESPACE SERVER
====    ================== ========= ======
default v1.7.0             default   https://35.229.50.36
```
```
ks version
ksonnet version: 0.9.1
jsonnet version: v0.9.5
client-go version: 1.8
```

This appears to be an issue with ksonnet set. 

Work around is to create a new namepace.

```
ks env add gke --namespace=kubeflow

ks env list
NAME    KUBERNETES-VERSION NAMESPACE SERVER
====    ================== ========= ======
default v1.7.0             default   https://35.229.50.36
gke     v1.7.0             kubeflow  https://35.229.50.36
```

@ankushagarwal Have you run into this?

jlewi:
Here's the help for ks env

```
ks env
ERROR Command 'env' requires a subcommand

Usage:
  ks env [flags]
  ks env [command]

Available Commands:
  add         Add a new environment to a ksonnet application
  list        List all environments in a ksonnet application
  rm          Delete an environment from a ksonnet application
  set         Set environment-specific fields (name, namespace, server)
```

So I expect setting namespace should work.

jlewi:
Filed ksonnet/ksonnet#373

jlewi:
We should update our docs to describe the work around of adding a new environment.

/assign @ankushagarwal

ankushagarwal:
Looks like this was fixed here https://github.com/ksonnet/ksonnet/pull/371. I have requested a new release. If they don't create a new release, I will update the docs.

PR #515, opened 2018-03-27
Use ks version 0.9.2

ankushagarwal:
ks 0.9.1 had a bug in which ks env set default --namespace=kubeflow doesn't change the namespace 

This is fixed in ks 0.9.2 so we need to update to 0.9.2.

Fixes https://github.com/kubeflow/kubeflow/issues/477

Related to #506 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/515)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ If you want to use GPUs, be sure to follow the Kubernetes [instructions for enab
 
 ### Requirements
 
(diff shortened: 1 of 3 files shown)


jlewi:
nit; I updated the PR description. Its better to include the actual description in the PR description. This is convenient for reviewers and looking at the history because then we don't have to click through links to understand the issue.

/lgtm
/approve

PR #517, opened 2018-03-27
Http proxy default image

### LINKED ISSUE #511: Tf serving component should provide default HTTP image
Currently, if the user doesn't specify the HTTP proxy image, we don't deploy that.

I think we should have a flag, something like deploy_http_proxy.
If deploy_http_proxy = true, use the user specified image (if exist) or the default image.

lluunn:
- Change the flag for deploying http proxy to bool.
- Use a default image if user does not specify the image.

Fix #511

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/517)


lluunn:
The lgtm tag is not added, maybe /lgtm has to be on its own line?
@yupbank can you try again?
Thanks!

lluunn:
Thanks!
/approve

PR #518, opened 2018-03-27
Install jupyter widgets in tensorflow-notebook-image

### LINKED ISSUE #489: Javascripts widgets don't work in JupyterLab
They work in jupyter but not JupyterLab.

I think the [issue](https://github.com/jupyterlab/jupyterlab/issues/1548) is that we haven't enabled widgets for JupyterLab.

Here are the [instructions for installing widget support](https://ipywidgets.readthedocs.io/en/latest/user_install.html#installing-the-jupyterlab-extension).

I haven't got this working yet. I suspect the problem is that I'm using the Python 2.7 kernel which is not the default and I didn't install things correctly.

pdmack:
Fixes #489 

@jlewi Your instructions appear to be correct for py3. Not sure why it was failing in your env. I tested with [interact](http://ipywidgets.readthedocs.io/en/latest/examples/Using%20Interact.html) but maybe you have something else in mind.

screen shot 2018-03-27 at 5 03 38 pm

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/518)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -208,6 +208,13 @@ RUN conda_packages=$(conda list -e | cut -d '=' -f 1 | grep -v '#' | sort) && \
     pip install --no-cache-dir tensorflow-transform && \
     source deactivate
 
+# install notebook and lab widgets
+RUN 	conda install -n root -c conda-forge widgetsnbextension && \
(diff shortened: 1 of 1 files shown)


jlewi:
@willingc FYI we use prow to manage PRs so if you want to actually lgtm PRs you need to do "/lgtm"

Issue #215, opened 2018-02-07
Create a minimal release process for our ksonnet configs


jlewi:
Kubeflow was broken for quite a bit because the ksonnet configs got out of sync with the Docker images.

To avoid this we should start pinning/tagging known good releases.

Context:
    tensorflow/k8s#339
    Discussion in #208 

If we create a tag or branch then we can just use that tag with ksonnet

```
ks registry add kubeflow github.com/kubeflow/kubeflow/tree/0.1/kubeflow
```

At a minimum it would be good to start defining a process for creating releases to figure out what we will do for our 0.1 milestone.

Even better would be to think about how to automate the toil of releases e.g
    - Running tests to qualify the release
    - Publishing release notes

Are there existing tools we can use?

gaocegege:
>  Publishing release notes

https://github.com/skywinder/github-changelog-generator could generate changelog automatically.

jlewi:
@willb Can I assign this to you for now since you are working on a proposal for the release process?

jlewi:
We had another break #283 at head affecting serving.

I need to push a new Docker image for TFJob to fix #322

I'm thinking as part of this we should start documenting the steps it takes to do a release and then use that to inform a proposal.

I chatted briefly with @willb and it sounds like he's still researching various options.

In brief here's what I think this will look like

- Build test TFJob operator by launching an Argo workflow
- Create a PR updating the ksonnet configs to use the new image.
- Tag the kubeflow/kubeflow commit "STABLE"

I think we should start using the tag STABLE to refer to the most recent stable commit of our registry.

jlewi:
We now have an E2E test for TFServing.
But the E2E test builds a new Docker image for TFServing and then uses that.  This is great if we're building a new Docker image and want to verify it works but its not quite what we want in the case where we want to verify that the current ksonnet prototypes will work.

When we release our ksonnet configs the flow will be something like
  - Update the ksonnet components (e.g. update Docker images)
  - Run the tests to verify everything works
  - Submit a PR
  - Tag the commit stable

A simple thing to do would be to add another workflow which runs the E2E test for TFServing but without building a new image and without overriding the image used by the ksonnet component.

/cc @lluunn

jlewi:
To elaborate on the above I think we should have the following

- An Argo workflow that builds all the Docker images
   - e.g TFJob operator, TFServing, HTTP Proxy Jupyter Images, 
[...]

PR #519, opened 2018-03-28
Branching and tagging policy for releases

willb:
This commit

1.  adds a branching policy for release branches,
2.  adds a tagging policy for release candidates, and
3.  adds a tagging policy for releases.

Related to #215 and #506

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/519)


releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
 Create a PR to update [jupyterhub_spawner.py](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/jupyterhub_spawner.py#L15) 
 to point to the newly built Jupyter notebook images.
 
-## Update the ksonnet configs
+## Create a release branch (if necessary)
 
-Create a PR to update the ksonnet configs to use the latest images built in the previous steps.
+If you aren't already working on a release branch (of the form `v${VER}-branch`, where `${VER}` is a major-minor version number), then create one.  Release branches serve several purposes:
```
jlewi: I would suggest dropping 'v${VER}'
branch seems redundant
The branch will be used with ksonnet so that users can pull a stable version of the configs rather than the latest configs on master.

releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
...
 
-Presubmit tests should verify that the ksonnet configs work.
+1.  They allow a release wrangler or other developers to focus on a release without interrupting development on `master`,
+2.  they allow developers and sophisticated users to track the development of a release before a release candidate is declared, and
+3.  they simplify backporting critical bugfixes to a patchlevel release particular release stream (e.g., producing a `v0.1.1` from `v0.1-branch`), when appropriate.
```
jlewi: Probably worth mentioning the bit about providing stable ksonnet config users.

releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
 Create a PR to update [jupyterhub_spawner.py](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/jupyterhub_spawner.py#L15) 
 to point to the newly built Jupyter notebook images.
 
-## Update the ksonnet configs
+## Create a release branch (if necessary)
```
jlewi: I assume we just create a release branch via GitHub UI?

releasing.md
willb: Git's shared namespace for different kinds of refs can lead to user confusion.  I'm suggesting `-branch` to disambiguate between moving targets (e.g., the branch from which all v0.1-series releases will be tagged) and fixed targets (e.g., the tags v0.1, v0.1.1, etc).  If you're making the argument that `v0.1` should be a branch but `v0.1.0` should be a tag, I can see that, but I think it makes sense to choose names that make the distinction more explicit to users (especially since this will be the name they give to ksonnet).

releasing.md
willb: Yes, that's what I was driving at with "they allow developers and sophisticated users to track the development of a release before a release candidate is declared" but I agree that having a ksonnet registry URL that always points to a relatively stable set of images in a particular release stream is worth calling out separately.

releasing.md
willb: That's what I did!

releasing.md
jlewi: Makes sense.

releasing.md
jlewi: Can you add that to the doc?

releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
...
-Submit the PR.
+### Release branching policy
 
-Wait for a passing postsubmit of the submitted PR.
+A release branch should be substantially _feature complete_ with respect to the intended release.  Code that is committed to `master` may be merged or cherry-picked on to a release branch, but code that is directly committed to the release branch should be solely applicable to that release (and should not be committed back to master).
 
-Tag the commit corresponding to the passing postsubmit with the appropriate tags for the release.
+## Updating the release branch and tagging a release
+
+0. Ensure that the release notes are up to date.
```
jlewi: Where are the release notes? Maybe this should point to an issue to get release notes? :)

releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
...
+1. Update the ksonnet configs on the release branch to use the latest images built in the previous steps.
```
jlewi: Its not clear to me whether certain things e.g. pinning of the images would happen on the release branch or on the master. Maybe we should open an issue to resolve that? 

Maybe we should have a section that says after you cut the release branch perform any operations that only happen on the research branch and then explain what those might be.

releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
...
+2. Presubmit tests should verify that the ksonnet configs work.
+3. Submit a PR to the release branch.
+4. Wait for a passing postsubmit of the submitted PR.  
+5. Once it passes, update this document on the release branch with specific image hashes.
```
jlewi: Why would we update with the image hashes? Shouldn't their be a single commit and that commit should pin all the images to specific images?

releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
...
+6. If no known blocker issues remain, tag a release candidate. 
+
+### Tagging a release candidate
+
+A release candidate is a tag of the form `v${VER}.${PATCHLEVEL}-RC${N}`, where `N` is a small integer, and a release candidate tag always points to a commit on the corresponding minor release branch.  Push this tag to GitHub and announce a release vote on kubeflow-discuss.
```
jlewi: We should verify that postsubmits pass by finding the prow job
https://prow.k8s.io/?repo=kubeflow%2Fkubeflow&job=kubeflow-postsubmit

It might be more obvious to do
v${MAJOR}.${MINOR}.${PATCHLEVEL}-RC${N}

releasing.md
```diff
@@ -155,14 +155,46 @@ ks apply ${ENV} -c workflows
...
+
```
jlewi: How do we make a release show up here
https://github.com/kubeflow/kubeflow/releases

releasing.md
willb: Yes, let me explain:  when you put an explicit hash in `releasing.md`, I started to think it might make sense to have `releasing.md` on a release branch _actually contain_ the hashes used in the release, instead of placeholders.  This is of somewhat limited utility, but it does serve as accessible long-term documentation of what images were used in a particular release.

jlewi:
@willb Are you going to make some changes or should we submit as is?

releasing.md
jlewi: Lets do RC.${N} to match K8s
https://github.com/kubernetes/kubernetes/releases/tag/v1.10.0-rc.1
https://github.com/kubernetes/kubernetes/releases?after=v1.10.0-beta.2

And this allows the tag (RC/alpha/beta) to be easily split from the iteration.

willb: Sounds good.  I'll make the changes.

jlewi:
@willb Could you sync please and see if the test passes? It would be good to get this submitted.

willb:
@jlewi things are passing now; ptal?

PR #533, opened 2018-03-30
Update the TFJob image in preparation for a new RC.

jlewi:
* The updated TFJob includes the fix to delete pods when job finishes
  rather than leave them running.

* We need to pin the commit of the tf-operator in the extra repos so
  that we checkout the tests at the commit that matches the image.

Related to #506

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/533)


ankushagarwal:
@jlewi Does this need to be cherry-picked to the release branch?

PR #543, opened 2018-03-30
Add tf1.7 to tensorflow-notebook images

ankushagarwal:
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/543)


jlewi:
Note there is an issue with the CUDA version in our notebook images being too old for the newer versions of TF.
see #549

/lgtm
/approve

ankushagarwal:
We use cuda 8.0 for tf 1.4x.

We use cuda 9.0 for tf 1.5x and higher 

I confirmed with someone on the tensorflow team that we need cuda 9.0 for tf1.7

jlewi:
Thanks
/approve
/lgtm

PR #544, opened 2018-03-30
Add the "tensorflow-model-analysis" package to the notebook images

ojarjur:
More information about this package is [here](https://github.com/tensorflow/model-analysis).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/544)


components/tensorflow-notebook-image/Dockerfile
```diff
@@ -206,6 +206,9 @@ RUN conda_packages=$(conda list -e | cut -d '=' -f 1 | grep -v '#' | sort) && \
     python -m ipykernel install --user && \
     echo "${pip_only_packages}" | xargs -n 1 -I "{}" /bin/bash -c 'pip install --no-cache-dir {} || true' && \
     pip install --no-cache-dir tensorflow-transform && \
+    pip install --no-cache-dir tensorflow-model-analysis && \
+    jupyter nbextension install --py --symlink tensorflow_model_analysis --system && \
```
jlewi: Why --symlink?

ojarjur: The reason I included it was that I started with the instructions from the model-analysis README which uses `--symlink`, and that part doesn't cause issues in our build (the default install location did cause issues, but that is orthogonal).

From what I understand, passing in `--symlink` is considered a best-practice since it saves space on non-Windows systems. However, I don't have a strong preference one way or the other.

ojarjur:
Added a workaround to the issue where importing tensorflow during the build of the GPU image was failing.

PTAL

jlewi:
@ojarjur Could you add a comment in the Dockerfile explaining the work around?
Otherwise looks good.

ojarjur:
@jlewi Done.

PR #548, opened 2018-03-30
Add TfJob dashboard to ksonnet

### LINKED ISSUE #546: TFJob UI not included in ksonnet configs
TFJob UI not included in ksonnet configs.

We should fix that and then cherry pick into v0.1 #506 

/cc @wbuchwalter @ankushagarwal

wbuchwalter:
Fix #546 

I didn't had time to try this yet on a new deployment, so hold for now.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/548)


jlewi:
Can you sync so tests pass? Its failing because of GPU tests which were recently disabled.

I think we can submit without manual testing and then test the RC which we will do any way.

PR #557, opened 2018-04-02
Add PVC note to troubleshooting

inc0:
PVC causes a lot of issues with notebook spawn and it's not
straightforward to fix. Add troubleshooting note to help with that.


pdmack:
/lgtm 

Should this not be the default? 
/cc @jlewi 

@inc0 text suggestion:
```
By default JupyterHub spawns notebook instances with a persistent volume claim (PVC) mounted to `/home/jovyan/work`. That means your Kubernetes environment must support persistent volumes (PV) and have a default [StorageClass](https://kubernetes.io/docs/concepts/storage/storage-classes/). For more information regarding PV please refer to the [Kubernetes docs](https://kubernetes.io/docs/concepts/storage/persistent-volumes/).
￼If your setup doesn't have a valid default StorageClass, you can disable the PV mount:
```

jlewi:
@pdmack I'm on the fence about what the proper default should be. Using PD is much more useful in my opinion. Although the fact that the entire home directory isn't on PD strikes me as slightly confusing as well.

Do we have any sense of which K8s distros don't have PVC defined by default?

pdmack:
Not sure what you mean by "defined by default". OpenShift obviously [supports PV and PVC](https://docs.openshift.org/latest/architecture/additional_concepts/storage.html) but it is up to the project developer/user to set it up according to their needs. Do you mean [PVC protection](https://docs.openshift.org/latest/install_config/configuring_pvc_protection.html#install-config-configuring-pvc-protection)?

jlewi:
I meant don't have a default storage class defined.

After thinking about this more. I lean towards disabling PV by default in 0.1. My thinking is that users who want PV already need to do some digging to figure out that only /home/joyvan/work is backed up. So I'm not sure its a stretch to have them manually enable PV.

It might be nice to provide a warning in the spawner UI about persistence.

@inc0 Can you make this change so we can cherry pick it into 0.1?

inc0:
> @inc0 Can you make this change so we can cherry pick it into 0.1?

Sure, do we want to merge to master and cherry-pick as separate PR later? I'd say that's better approach - will protect us from having different commit in master than in 0.1.
We will need this process for all sorts of bugfixes later on so it might be good opportunity to try it;)

> After thinking about this more. I lean towards disabling PV by default in 0.1. My thinking is that users who want PV already need to do some digging to figure out that only /home/joyvan/work is backed up. So I'm not sure its a stretch to have them manually enable PV.

One scenario I'm afraid of is if people will start working on something, not knowing that their work is totally ephemeral, and when pod is recreated, all their work is wiped out. I think better alternative would be to use hosts disk if no storageclass is provided.

pdmack:
@jlewi OK, in that case "no". OpenShift (at least) [doesn't provide a default StorageClass](https://docs.openshift.org/latest/install_config/storage_examples/storage_classes_dynamic_provisioning.html#example1) atm.

jlewi:
> Sure, do we want to merge to master and cherry-pick as separate PR later? I'd say that's better approach - will protect us from having different commit in master than in 0.1.
We will need this process for all sorts of bugfixes later on so it might be good opportunity to try it;)

Yes that's the process we've been following.

> One scenario I'm afraid of is if people will start working on something, not knowing that their work is totally ephemeral, and when pod is recreated, all their work is wiped out. I think better alternative would be to use hosts disk if no storageclass is provided.

I agree with you that ephemeral is bad(P1). My current thinking is we need to find a least bad solution for a 0.1 release and then prioritize a better long term fix in keeping with the various open issues.

I think for 0.1. more people will be kicking the tires then doing actual work. So  I think being ephemeral by default is ok and preferable if it minimizes hard to debug errors.

inc0:
https://github.com/kubeflow/kubeflow/pull/577

PR #558, opened 2018-04-02
Enable TF serving gpu test

lluunn:
related #484

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/558)

### DIFF
--- a/components/k8s-model-server/images/releaser/components/workflows.libsonnet
+++ b/components/k8s-model-server/images/releaser/components/workflows.libsonnet
@@ -326,7 +326,7 @@
           ],  // volumes
 
           // onExit specifies the template that should always run when the workflow completes.
-          // onExit: "exit-handler",
+          onExit: "exit-handler",
 
           templates: [
             {
--- a/components/k8s-model-server/images/releaser/components/workflows.libsonnet
+++ b/components/k8s-model-server/images/releaser/components/workflows.libsonnet
@@ -232,17 +232,16 @@
           template: "test-tf-serving",
           dependencies: ["deploy-tf-serving"],
         },
-        // TODO(lunkai): investigate and re-enable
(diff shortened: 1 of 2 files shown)


lluunn:
From kubeflow-presubmit-tf-serving-image-558-fc89688-1030-e7d9

FailedScheduling 0/4 nodes are available: 1 Insufficient cpu, 3 Insufficient nvidia.com/gpu.

I need to add some cpu in the gpu pool

lluunn:
/cc @jlewi 

I want to re-enable this to capture the failure. Last time the failure was due to the cluster didn't have gpu.

jlewi:
Did you try repo'ing the failures using a bash script like [this](https://github.com/jlewi/kubeflow-dev/blob/master/run_tf_serving_release.sh).

It would be good to avoid reenabling it until we have some confidence its reliable?

Alternatively, you could modify our test infra so that in prow_config.yaml we could mark certain test failures as ignoreable. But you'd have to probably do it based on the name of the actual junit test case name.

lluunn:
I tested with 3 workflows running at same time, all succeeded. Can't reproduce the error now.

One observation: Stackdriver doesn't show logs from gpu deployment unless I disable the teardown. So I am guessing it's because the pod gets deleted before it can upload. I will try pause the exit handler.

jlewi:
/hold
/lgtm
/approve

I'll leave the decision to you. If you think we should reenable it as is then just cancel the hold.

Alternatively we could modify our test infrastructure to ignore the test failures so that we can bake the test.

lluunn:
It's not easy to tell whether a junit failure should be ignored as one workflow can have multiple junit results.
Since I could never reproduce the error, I will submit it now and see.
/cancel hold

PR #559, opened 2018-04-02
Rety upto 3 times while building tensorflow notebook images

ankushagarwal:
/cc @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/559)


components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
```diff
@@ -82,6 +82,9 @@
       // command: List to pass as the container command.
       local buildTemplate(step_name, command, env_vars=[], sidecars=[]) = {
         name: step_name,
+        retryStrategy: {
```
lluunn: Is it flaky? Can you add some comment on why we need retry

ankushagarwal: Done

PR #570, opened 2018-04-03
Update tf jupter notebook images to include tfma (from #544)

ankushagarwal:
tf1.4 gpu image has issues building with the latest tfma, so that's
not updated in this PR. Related  : https://github.com/kubeflow/kubeflow/issues/571

Also add tf1.7 images

/cc @jlewi
/cc @ojarjur

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/570)


jlewi:
Doesn't TFMA require newer versions of TF? Should we not install the package in images corresponding to unsupported versions of TFMA?

jlewi:
They only mention 1.6
https://github.com/tensorflow/model-analysis

So I suggest not installing it in images with TF versions older than 1.6

ankushagarwal:
Updated PR to only update tf1.6 and tf1.7

PR #572, opened 2018-04-03
Moved OAuth secret from param to named secret

### LINKED ISSUE #569: IAP should not store client secret in ksonnet
We currently store the client secret as a ksonnet param
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/iap.libsonnet#L178

This is not good since ksonnet configs get checked into source control.

We should instead use a K8s secret and our ksonnet config should get the secret from the secret.

We should cherry pick this into the 0.1 release (#506). Hence P0. 

@ankushagarwal can you take a look?

danisla:
Fixes #569 and updates the iap.md user guide.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/572)


ankushagarwal:
/lgtm
/approve
Thanks for the quick turnaround!

PR #576, opened 2018-04-03
Moved OAuth secret from param to named secret (#572)

jlewi:
Cherrypick the fix to use a K8s secret for the oauth client id rather than using a ksonnet parameter which would get checked into source control.

Related to #506

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/576)


ankushagarwal:
I already created https://github.com/kubeflow/kubeflow/pull/575

jlewi:
Thanks closing this one.

Duplicate of #575

PR #577, opened 2018-04-03
Disable PVC by default

inc0:
Per discussion, we think default should be as easy to setup as possible,
and dependency on StorageClass makes it less straightforward to start
working with kubeflow. Change this default to remove need for persistent
storage.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/577)

### DIFF
--- a/kubeflow/core/jupyterhub_spawner.py
+++ b/kubeflow/core/jupyterhub_spawner.py
@@ -100,7 +100,7 @@ c.KubeSpawner.start_timeout = 60 * 30
 # TODO(jlewi): Should we set c.KubeSpawner.singleuser_fs_gid = 1000
 # see https://github.com/kubeflow/kubeflow/pull/22#issuecomment-350500944
 pvc_mount = os.environ.get('NOTEBOOK_PVC_MOUNT')
-if pvc_mount:
+if pvc_mount and pvc_mount != 'null':
     c.KubeSpawner.user_storage_pvc_ensure = True
     # How much disk space do we want?
     c.KubeSpawner.user_storage_capacity = '10Gi'
--- a/user_guide.md
+++ b/user_guide.md
@@ -86,6 +86,11 @@ Now let's set `${KF_ENV}` to `cloud` or `nocloud` to reflect our environment for
 $ KF_ENV=cloud|nocloud
 ```
 
+By default Kubeflow does not persist any work that is done within the Jupyter notebook. That means if the container is destroyed or recreated, all of its contents, including users working notebooks and other files are going to be deleted. To enable the persistence of such files, the user will need to have a default StorageClass defined for [persistent volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/). If that is defined, persistence can be enabled by setting jupyterNotebookPVCMount to the available volume mount.
+```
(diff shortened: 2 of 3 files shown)


user_guide.md
```diff
@@ -86,6 +86,12 @@ Now let's set `${KF_ENV}` to `cloud` or `nocloud` to reflect our environment for
 $ KF_ENV=cloud|nocloud
 ```
 
+By default Kubeflow does not persist any work that is done within jupyter notebook. That means if container is destroyed or recreated, all it's contents, including users working notebooks etc. are going to be deleted. To enable persistence user will first need default will need to have default storage class for [persistent volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/). If that's the case, persistency can be enabled by setting jupyterNotebookPVCMount.
```
jlewi: "will first need default will need to have default storage class"
There is some duplication there.

user_guide.md
```diff
@@ -86,6 +86,12 @@ Now let's set `${KF_ENV}` to `cloud` or `nocloud` to reflect our environment for
 $ KF_ENV=cloud|nocloud
 ```
 
+By default Kubeflow does not persist any work that is done within jupyter notebook. That means if container is destroyed or recreated, all it's contents, including users working notebooks etc. are going to be deleted. To enable persistence user will need to have default storage class for [persistent volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/). If that's the case, persistency can be enabled by setting jupyterNotebookPVCMount.
```
pdmack: Minor word smithing:

```
By default Kubeflow does not persist any work that is done within the Jupyter notebook. That means if the container is destroyed or recreated, all of its contents, including users working notebooks and other files are going to be deleted. To enable the persistence of such files, the user will need to have a default StorageClass defined for [persistent volumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/). If that is defined, persistence can be enabled by setting jupyterNotebookPVCMount to the available volume mount.
```

kubeflow/core/jupyterhub_spawner.py
```diff
@@ -100,7 +100,7 @@ def extra_resource_limits(self):
 # TODO(jlewi): Should we set c.KubeSpawner.singleuser_fs_gid = 1000
 # see https://github.com/kubeflow/kubeflow/pull/22#issuecomment-350500944
 pvc_mount = os.environ.get('NOTEBOOK_PVC_MOUNT')
-if pvc_mount:
+if pvc_mount and pvc_mount != 'null':
```
jlewi: This is good for now but it would be better to do this in the jsonnet rather than propogating the hack of 'null' to.

inc0: yeah, but I can't find out how to assign empty string to default. As soon as it's fixed in ksonnet, we should do that

jlewi: I meant in ksonnet we should continue to use "null" and then in ksonnet we should have a utility function here
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/util.libsonnet#L14

Then can automatically handle converting "null" to ""

PR #591, opened 2018-04-04
Expanding argo permissions to enable creation of more objects

elsonrodriguez:
If argo's going to be used in ML worfklows, it may need extra permissions.

These are just to get the mnist example over the hump:

https://github.com/kubeflow/examples/pull/42#issuecomment-378322298

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/591)

### DIFF
--- a/kubeflow/argo/argo.libsonnet
+++ b/kubeflow/argo/argo.libsonnet
@@ -275,6 +275,8 @@
           apiGroups: [""],
           resources: [
             "configmaps",
+            "serviceaccounts",
+            "secrets",
           ],
           verbs: [
             "get",
@@ -295,6 +297,38 @@
           ],
         },
         {
+          apiGroups: [""],
+          resources: [
+            "services",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "list",
+            "watch",
+            "update",
+            "patch",
+          ],
+        },
+        {
+          apiGroups: [
+            "apps",
+            "extensions",
+          ],
+          resources: [
+            "deployments",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "list",
+            "watch",
+            "update",
+            "patch",
+            "delete",
+          ],
+        },
+        {
           apiGroups: [
             "argoproj.io",
           ],
@@ -307,6 +341,23 @@
             "watch",
             "update",
             "patch",
+          ],
+        },
+        {
+          apiGroups: [
+            "kubeflow.org",
+          ],
+          resources: [
+            "tfjobs",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "list",
+            "watch",
+            "update",
+            "patch",
+            "delete",
           ],
         },
       ],


kubeflow/argo/argo.libsonnet
```diff
@@ -294,6 +296,38 @@
             "delete",
           ],
         },
+        {
+          apiGroups: [""],
+          resources: [
+            "services",
+          ],
+          verbs: [
+            "create",
```
jlewi: Is this because you are using Argo to create and deploy services?

Do you know what service account the pod is executing as? 

My guess is the pod ends up using the Argo service account so that service account needs permissions for all ops executed by pods in the workflow.

I wonder if a better model is to give pods special service accounts and then create appropriate role bindings for those accounts.

That's how our E2E tests work. We use GCP service accounts not K8s service accounts. The pod then uses the RBAC permissions associated with that account.

Would you be willing to try that out with the mnist model? e.g. create a K8s service account with appropriate role bindings for your example. In your argo workflow attach that service account to the pods. I'm assuming thats an option in Argo but I could be wrong.

kubeflow/argo/argo.libsonnet
```diff
@@ -275,6 +275,8 @@
           apiGroups: [""],
           resources: [
             "configmaps",
+            "serviceaccounts",
```
jlewi: I assume this is to attach secrets and serviceaccounts to pods?

kubeflow/argo/argo.libsonnet
elsonrodriguez: You can specify any service account in argo, the default is `default`.

I opted just to expand the permissions of the existing argo service account, which had most of the permissions needed anyway.

I can also close this out and just put the object yamls for the service account, role, and binding in the mnist example.

kubeflow/argo/argo.libsonnet
elsonrodriguez: This is related to that ksonnet serviceaccount hack. The container needs to be able to get the service account in order to generate the kubeconfig.

kubeflow/argo/argo.libsonnet
jlewi: Sorry now that I understand this better I think it would be better if the e2e example created its own service account with appropriate roles and bindings and used that service account as needed with the argo workflow pods.

I think that's a better approach than expending the permissions of the argo controller itself.

elsonrodriguez:
Per slack convo, it looks like other people have a need for argo to have more permissions by default, especially for TFJob.

Let me know if any permissions should be stripped.

kubeflow/argo/argo.libsonnet
jlewi: Can you explain what the problem with this approach was? Is the problem just that it creates another hurdle users have to go through?

What if we
1. Defined a service account "kubeflow-workflow" to be used by Kubeflow workflows
1. Associated appropriate roles and bindings with this service account
1. Created Argo Workflow templates as needed that used this service account by default.

The reason I like this approach is because its easier to lock down; i.e. security concious users could just not to enable/create this service account.

But those users might still want to use Argo.

So I think its better if we try to run Argo with minimal permissions and then make it easy for users who don't want to be bothered to get stuff done.

jlewi:
@elsonrodriguez There's a question pending to you.

kubeflow/argo/argo.libsonnet
elsonrodriguez: That makes sense. I can close this PR, or just trim it it down to only allow access to TFJobs.

jlewi: I think my main question is whether the Argo service account as opposed to the service account used by the pods Argo creates needs expanded permissions?

I would think only the pod service accounts would need the ability to create TFJobs because I think Argo creates a pod to create a TFJob.

If that's the case I'd prefer not to  expand the permissions of the argo service account but instead create an appropriate service account for the pods.

jlewi:
What's the status of this PR?

@kkasravi Is working on a proposal regarding memberships and permission models.

PR #593, opened 2018-04-05
Fix envoy deployment

lluunn:
The envoy container didn't mount the config-volume

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/593)

### DIFF
--- a/kubeflow/core/iap.libsonnet
+++ b/kubeflow/core/iap.libsonnet
@@ -136,6 +136,10 @@
           mountPath: "/etc/envoy",
           name: "shared",
         },
+        {
+          mountPath: "/var/envoy-config/",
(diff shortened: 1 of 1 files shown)


danisla:
This volume is intentionally not mounted. The envoy container instead mounts the shared volume where the envoy config will eventually be created.

The envoy pod initially starts in a crash loop (because the config file is missing) while the IAP container finishes configuring IAP and writing out the envoy config to the shared volume. 

One way to prevent the crash loop is to add add the initContainer again that basically has the same functionality as the sidecar. Another way is to modify the envoy container command to sit and wait for the config file to exist before running envoy. The end result is the same.

lluunn:
I see, thanks!

Issue #105, opened 2018-01-09
Make it easy to get started with Kubeflow


jlewi:
We'd like to have a super simple getting started experience for people who want to kick the tires on Kubeflow.

In #90, ksonnet was highlighted as a potential barrier.

As a strawman I think a good getting started experience might be

```
kubectl run --image=gcr.io/kubeflow/bootstrap:v1
```

In which case the bootstrap program would automatically deploy a stock Kubeflow deployment on the cluster. This would prevent the user from having to know anything about ksonnet until they need to more complex things like customize the deployment.

We could enrich this by having bootstrap startup a webapp that could provide a guided walk through to make it easy for the user to set basic options.

elmiko:
i like the idea of a bootstrap container to do all the setup, i'm also a fan of enabling user growth by exposing the details. when this had been brought up in #23 , you had recommended the `ks show` command as a possible alternative for getting at the kubernetes manifests. i wonder if we could also demonstrate something like this in the quick start guide (eg "Here's how to get the kubernetes manifests"), or it could even be a quick export from the bootstrap webapp.

this way we could keep the ksonnet files as the primary source of truth for installation, give the users some quick commands to get at the kubernetes manifests and not need to maintain a separate set of files/instructions. i realize it may still require the user to have ksonnet (barring some sort of bootstrap webapp), but at least we could do a little "teaching to fish".

jlewi:
I like the idea of the webapp spitting out the ksonnet app either by giving you a tarball that you can download or pushing to a Git repo

foxish:
I think this might be a point to consider a kubeflow-operator as a concept.
It could be configurable and also allowing sufficient flexibility, while
abstracting away the underlying implementation, and offer high level knobs
dealing with the individual sub components.

elmiko:
@jlewi by "ksonnet app", which app were you referring to?

sub-mod:
Can we have the manifest file available along with ksonnet  code ?

erikerlandson:
I like the idea of a kubeflow command for dumping manifests.  That ought to eliminate the issues of developer burden raised in #90, since it becomes self serve

erikerlandson:
Also +1 for kubeflow operator, although I remain skeptical that there is just one uber-CRD for this

jlewi:
I'd like to keep this issue pretty narrowly scoped to providing a quick start that hides ksonnet. Lets open up other issues as needed to think about maintaining a Kubeflow deployment throughout its lifetime.

[...]

PR #595, opened 2018-04-06
Create a POC of an app to simplify deployment of Kubeflow

jlewi:
* The app is intended to optimize a Kubeflow app based on a user's
actual K8s setup.

* To illustrate this the app checks whether the user's cluster
has a default storage-class and uses that to determine whether to
enable PD for Jupyter notebooks.

* The intent is to publish this as a docker container which would
  also contain ksonnet.

* As described in the README a user should be able to start a shell
  in this container and walk through the steps of creating an App
  without having to install additional tools.

* This is currently using ksonnet at HEAD because they recently
  refactored the code and we want to take advantage of that; in particular
  they started organizing the code into actions.

* The code and generated app currently require a version of ks > 0.9.2; which means we currently need
 to build ks from source. We include a version of ks in the Docker image that is built from the appropriate
 commit.

Related to #105

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/595)


kkasravi:
@jlewi i may of hit a similar issue - I had to checkout ksonnet-lib under the kubeflow app. 

I've checked out your PR and am running it - I'll update here.

inc0:
Have you guys tried helm? To install helm you pull golang app "helm" and run helm init. That's it. Pretty cool ux

kunmingg:
@jlewi 
For the "bootstrap/vendor", if it won't be updated very often (not much version control needed), can we have it stored in separate repo or cloud storage accessible by endpoint?

jlewi:
@inc0 I've tried helm. I'm open to discussing helm but lets not do it in the context of this PR. I think to some extent the intent here is to do things that can't be solved by helm/ks. Take a look at my most recent comment on #105 to get an idea of where I want to go with it.

@kkasravi What issue are you refering to? ksonnet pushed a fix to ksonnet/ksonnet#420. I include a version of ks built at the appropriate commit in the docker image now.

@kunmingg checking in vendor is common practice for go.See kubeflow/tf-operator#222

bootstrap/developer_guide.md
```diff
@@ -0,0 +1,26 @@
+# Developer guide for bootstrap
+
+## Building the Operator
+
+Create a symbolic link inside your GOPATH to the location you checked out the code
+
+```sh
+mkdir -p ${GOPATH}/src/github.com/kubeflow
+ln -sf ${GIT_KUBEFLOW} ${GOPATH}/src/github.com/kubeflow/kubeflow kubeflow
```
jose5918: extra arg here

bootstrap/developer_guide.md
```diff
@@ -0,0 +1,26 @@
...
+Install dependencies, `-v` will ignore subpackage vendor
+
+```sh
+glide install -v
+```
+
+Build it
+
+```sh
+go install github.com/kubeflow/tf-operator/cmd/tf-operator
```
jose5918: go install github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap?

bootstrap/developer_guide.md
jlewi: Fixed

bootstrap/developer_guide.md
jlewi: Fixed

kunmingg:
@jlewi 
Thanks, and one more question:
```-v ${HOME}:/home/${USER}```
It might be aggressive to access ${HOME},
How about narrow it down to only sub-dirs it need to access?

jlewi:
@kunmingg Why is it aggressive to map in home directory? This is no different than user running a binary directly. The reason for mapping in the home directory is to

- Let the user store the app in their home directory
- Make use of config files for .kube and .gcloud

jlewi:
@inc0 @kunmingg Anymore comments? If not would you mind approving? Would be good to get this submitted so we can follow up with more improvements.

jlewi:
@kunmingg you need to "/lgtm"; only approvers can "/approve"

PR #605, opened 2018-04-06
import of cloud-endpoints component and support in iap-ingress component

### LINKED ISSUE #586: [GKE] Use Cloud Endpoints to provision domain
@danisla has pointed out that we can use Cloud Endpoints as a way to automatically generate a domain name that could be used with IAP.

We can do this by using Cloud Endpoints to register a service. This will give a DNS name and associated record for the IP address we specify. The service definition doesn't really matter since we won't actually use Cloud Endpoints for anything other than creating the DNS record.

I think we should do that so that we can eliminate another manual step (DNS creation/registration). We should still let users use a custom DNS name if they so choose.

My initial thought is we should create a small go program to

- Create the Cloud Endpoints Service
- Extract the DNS name
- Store it in a ConfigMap

Then we can just run it in a pod.

We'll need a GCP service account with suitable permissions.

I lean towards go because I think our efforts to simplify IAP and make it more robust (#550) are headed in the direction of creating a simple controller to continually monitor IAP/Ingress and make changes when needed.

/cc @danisla @ankushagarwal

danisla:
Fixes #586 

- Import of cloud-endpoints CRD and controller. Depends on kube-metacontroller which is bundled with the component.
- Updated iap-ingress component to create the `CloudEndpoint` resource if the FQDN matches the form of `NAME.endpoints.PROJECT.cloud.goog`
- Updated the IAP doc with steps for using Cloud Endpoints.

/cc @kunmingg
/cc @ankushagarwal 
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/605)

### DIFF
--- a/docs/gke/iap.md
+++ b/docs/gke/iap.md
@@ -10,15 +10,58 @@ when using GKE.
 
 If you aren't familiar with IAP you might want to start by looking at those docs.
 
+The instructions below reference the following environment variables which you will need to set for your deployment
+
+  * **CORE_NAME** The name assigned to the core Kubeflow ksonnet components (this is the name chosen when you ran `ks generate...`).
+  * **ENVIRONMENT** The name of the ksonnet environment where you want to deploy Kubeflow.
+  * **FQDN** The fully qualified domain name to use for your Kubeflow deployment.
+  * **IP_NAME** The name of the GCP static IP that you created above and will be associated with **DOMAIN**.
+  * **NAMESPACE** The namespace where Kubeflow is deployed.
+  * **ACCOUNT** The email address for your ACME account where certificate expiration notifications will be sent.
+
 ### Preliminaries
 
-##### Create an external static IP address
+#### Create an external static IP address
 
 ```
+PROJECT=$(gcloud config get-value project)
 gcloud compute --project=${PROJECT} addresses create kubeflow --global
 ```
 
-Use your DNS provider (e.g. Google Domains) create a type A custom resource record that associates the host you want e.g "kubeflow"
+#### Configure DNS record
+
+[Cloud Endpoints](https://cloud.google.com/endpoints/docs/) can be used to automatically provision a free DNS record for Kubeflow in the form of: `NAME.endpoints.PROJECT.cloud.goog`. 
+
+Enable the following APIs:
+- [Google Cloud Endpoints](https://console.cloud.google.com/apis/library/endpoints.googleapis.com/?q=Cloud%20Endpoints)
+- [Google Service Management](https://console.cloud.google.com/apis/library/servicemanagement.googleapis.com/?q=Service%20Management)
+
+Create a service account an IAM bindings for the cloud-endpoints-controller:
+
+```
+gcloud iam service-accounts create cloud-endpoints-controller \
+    --display-name cloud-endpoints-controller
+export SA_EMAIL=$(gcloud iam service-accounts list \
+    --filter="displayName:cloud-endpoints-controller" \
+    --format='value(email)')
+gcloud projects add-iam-policy-binding \
+      $PROJECT --role roles/servicemanagement.admin --member serviceAccount:$SA_EMAIL
+
+gcloud iam service-accounts keys create cloudep-sa.json --iam-account $SA_EMAIL
+
+kubectl create secret generic --namespace=${NAMESPACE} cloudep-sa --from-file=./cloudep-sa.json
+```
+
+Install the `cloud-endpoints` controller and set the `FQDN` environment variable used later:
+
+```
+ks generate cloud-endpoints cloud-endpoints --namespace=${NAMESPACE} --secretName=cloudep-sa
+ks apply ${ENVIRONMENT} -c cloud-endpoints
+
+export FQDN="kubeflow.endpoints.$(gcloud config get-value project).cloud.goog"
+```
+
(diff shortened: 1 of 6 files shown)


jlewi:
Awesome thanks for the quick fix.

jlewi:
/hold
/lgtm
/approve

/assign @ankushagarwal 

Ankush any comments?

ankushagarwal:
@danisla : When I tried this on a new k8s cluster, I get the following error from the cloud-endpoints-controller pod

Am I missing anything during the setup?
```
2018/04/06 23:36:49 [INFO] Fetching Project ID from Compute metadata API...
2018/04/06 23:36:49 [INFO] Fetching Numeric Project ID from Compute metadata API...
2018/04/06 23:36:49 [INFO] Instantiating GCE client...
2018/04/06 23:36:49 [INFO] Instantiating Google Cloud Service Management Client...
2018/04/06 23:36:49 [INFO] Initialized controller on port 80
2018/04/06 23:44:16 [DEBUG][kubeflow] Changed because parent sig different
2018/04/06 23:44:17 [INFO][kubeflow] Service does not yet exist, creating: kubeflow.endpoints.agwl-kubeflow.cloud.goog
2018/04/06 23:44:17 [ERROR] Could not sync state: [ERROR] Failed to creat Cloud Endpoints service: serviceName: kubeflow.endpoints.agwl-kubeflow.cloud.goog, err: googleapi: Error 403: Request had insufficient authentication scopes., forbidden
2018/04/06 23:44:18 [DEBUG][kubeflow] Changed because parent sig different
2018/04/06 23:44:18 [INFO][kubeflow] Service does not yet exist, creating: kubeflow.endpoints.agwl-kubeflow.cloud.goog
2018/04/06 23:44:18 [ERROR] Could not sync state: [ERROR] Failed to creat Cloud Endpoints service: serviceName: kubeflow.endpoints.agwl-kubeflow.cloud.goog, err: googleapi: Error 403: Request had insufficient authentication scopes., forbidden
```

ankushagarwal:
Looks like I had not enabled Cloud Endpoints API. Just enabled it. Trying again.

ankushagarwal:
Still getting the same error.

jlewi:
I don't think we should rely on the VM scope. That gives every pod cloud-platform scope.
Instead we should use a service account stored stored as a K8s secret.

[Here](https://github.com/kubeflow/testing#create-a-gcp-service-account) are some gcloud commands for creating the service account.

jlewi:
Is it possible for the CRD to check if cloud endpoints is enabled and if not enable it?

jlewi:
@danisla Ping any chance we could get support for secrets?

danisla:
Yes! I'm working on it now and should have the commit in today.

danisla:
@jlewi ok updated component and docs to use a service account and secret for credentials.

docs/gke/iap.md
```diff
@@ -10,15 +10,58 @@ when using GKE.
...
+Install the `cloud-endpoints` controller and set the `FQDN` environment variable used later:
+
+```
+ks generate cloud-endpoints cloud-endpoints --namespace=${NAMESPACE} --secretName=cloudep-sa
+ks apply ${ENVIRONMENT} -c cloud-endpoints
+
+export FQDN="kubeflow.endpoints.$(gcloud config get-value project).cloud.goog"
+```
+
+Alternatively if you already have a DNS provider (e.g. Google Domains) create a type A custom resource record that associates the host you want e.g "kubeflow"
```
jlewi: If we have our own domain then we don't need a cloud-endpoints component right?

kubeflow/core/prototypes/cloud-endpoints.jsonnet
```diff
@@ -0,0 +1,25 @@
+// @apiVersion 0.1
+// @name io.ksonnet.pkg.cloud-endpoints
+// @description Provides cloud-endpoints prototypes for creating Cloud Endpoints services and DNS records.
+// @shortDescription Cloud Endpoint domain creation.
+// @param name string Name for the component
+// @param secretName string Name of secret containing the json service account key.
+// @optionalParam secretKey string cloudep-sa.json Name of the key in the secret containing the JSON service account key.
+// @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
+
+// TODO(https://github.com/ksonnet/ksonnet/issues/222): We have to add namespace as an explicit parameter
```
jlewi: Delete this since the issue has been resolved you are inheriting namespace below.

docs/gke/iap.md
danisla: Correct, I can separate those sections for clarity.

danisla: The IP reservation gcloud step is also optional if you are using cloud-endpoints since the endpoint is coupled to any dynamically provisioned Ingress IP. We would just have to make the annotation on the Ingress conditional on the presence of an IP name.

jlewi: If both work lets document but if you have to reserve an IP I think that's fine. I think that's best practice anyway so that's what we'll probably recommend.

jlewi:
Test failure looks like  a flake. There was a problem talking to the GitHub API while initializing the app

```
Get https://api.github.com/repos/ksonnet/parts/commits/master: dial tcp: lookup api.github.com on 10.39.240.10:53: read udp 10.36.10.19:54866->10.39.240.10:53: i/o timeout
```

PR #512, opened 2018-03-27
Pin tf serving version to 1.4

lluunn:
latest doesn't work as in #510 
related #468

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/512)

Issue #468, opened 2018-03-20
Build TFServing images for different TF images


jlewi:
We should probably build different versions of TFServing images corresponding to different TF versions.

jlewi:
Is the HTTP Proxy TF Version specific?

lluunn:
No that should work for all TF versions.

pdmack:
Some strategy discussion in #608

ankushagarwal:
Should we use the same strategy as jupyter notebook images?

```
 gcr.io/kf-images/tensorflow-serving-1.4.1-cpu  
 gcr.io/kf-images/tensorflow-serving-1.4.1-gpu  
 gcr.io/kf-images/tensorflow-serving-1.5.1-cpu  
 gcr.io/kf-images/tensorflow-serving-1.5.1-gpu  
 gcr.io/kf-images/tensorflow-serving-1.6.0-cpu  
 gcr.io/kf-images/tensorflow-serving-1.6.0-gpu  
 gcr.io/kf-images/tensorflow-serving-1.7.0-cpu  
 gcr.io/kf-images/tensorflow-serving-1.7.0-gpu
```

lluunn:
We should have different tf version, cpu/gpu, and py2/py3 combination for TF serving image as above, e.g. gcr.io/kf-images/tensorflow-serving-1.6.0-py3-cpu
 
- jupyter notebook installs py2 and py3 together, but I think we have to separate that in serving case.
- Jupyter notebook has parameterized Docker [image](https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/Dockerfile). 
- The logic is embedded in the [workflow](https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/releaser/components/workflows.libsonnet#L130)
  

I think it's a little hard to read and maintain. And it's hard for users if they want to customize. 

How about listing the parameters in a file (yaml or json), and the workflow runs a script to read from the file and build the image?

cc @ankushagarwal

ankushagarwal:
> I think it's a little hard to read and maintain. And it's hard for users if they want to customize.

We should think of a simpler unified approach for building our images - jupyter notebook images, tf-operator images and tf-serving images - https://github.com/kubeflow/kubeflow/issues/666 would be a good place to track this.

jlewi:
@lluunn Is this item done? Can we close this issue?

PR #608, opened 2018-04-06
Upgrade tf-serving version to 1.6.0

pdmack:
General bump of TF deps. For my part, I'm specifically trying to get at the s3 support that came out with 1.5.0.

/cc @jlewi 
/cc @elsonrodriguez

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/608)


lluunn:
We pinned 1.4  in #512 because of https://github.com/tensorflow/serving/issues/819 

Have you verified the image works? Our current testing currently doesn't rebuild the image.

elsonrodriguez:
Pinning to 1.5 might be a nice workaround, gets you S3 without that bug.

pdmack:
@elsonrodriguez @lluunn yep. Got 1.6.0 CPU working but GPU has LD-related issues with CUDA. Will work on it a bit more but may just aim for 1.5.0 if progress isn't happening.

pdmack:
@lluunn @elsonrodriguez PTAL

```
# docker run --rm 1.6.0-tf-serving-test-cpu:mine /usr/bin/tensorflow_model_server --help | head -3
usage: /usr/bin/tensorflow_model_server
Flags:
	--port=8500                      	int32	port to listen on
# docker run --rm 1.6.0-tf-serving-test-gpu:mine /usr/bin/tensorflow_model_server --help | head -3
usage: /usr/bin/tensorflow_model_server
Flags:
	--port=8500                      	int32	port to listen on
```

@DjangoPeng, do these [flags](https://github.com/pdmack/kubeflow/blob/7c3ec41acdd7d697d141ecd9e1e26cb2a6a7dcb3/components/k8s-model-server/images/Dockerfile.gpu#L45) still look appropriate for 1.6.0? I enabled s3 flag.

jlewi:
FYI Per #468 we'd eventually like to produce multiple TF Serving images corresponding to different TF versions.

jlewi:
I think its fine to bump things to the latest version as a quick fix; as long as we don't overwrite the existing images people who need the older image tag can use that.

If we follow the approach we are taking for Jupyter notebooks, then we should be able to build multiple versions of TF Serving from master (i.e. not needing separate branches).

We could do this by having separate requirements.txt files and either separate Dockerfiles or parameterized Dockerfiles. Then in our Argo workflow we can just configure the build step as needed.

pdmack:
Do people strongly feel we need a proper progression through 1.4.1 and 1.5.0 before this change? Otherwise, I agree with @jlewi that we put it in and work out the general refactoring in the backlog.

PR #610, opened 2018-04-06
Initial ksonnet package for Pachyderm.

jlewi:
* Packages were created based on the helm chart
https://github.com/kubernetes/charts/tree/master/stable/pachyderm/templates

* Instructions for adding new packages is being added in #609.

* I've confirmed that the resulting components can successfully be deployed
  but I don't know if Pachyderm is working yet.

* We most likely need to expose more options e.g. to control things
  like GCS bucket.

* Related to #151 #611

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/610)


pdmack:
/lgtm

```
# oc get svc | grep pach
pachd              NodePort    10.103.96.5      <none>        650:30650/TCP,651:30651/TCP,652:30652/TCP   16m
# pachctl list-repo
NAME                CREATED             SIZE
```
I don't know that much about pachyderm but seems like it deployed OK.

BTW, this seems hella confusing with ksonnet:
```
 #  ks registry add kubeflow github.com/jlewi/kubeflow/tree/pachyderm/kubeflow --version=pachyderm
 #  ks pkg install kubeflow/pachyderm@pachyderm
```
Why am I over-specifying the GH ref for that branch?

pdmack:
/lgtm cancel
I think the dashboard should come out if its behind their paywall.

jlewi:
Agreed.

Issue #618, opened 2018-04-09
Tests are failing


jlewi:
Post submits show failures deploying Kubeflow on GKE
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-postsubmit

Presubmits having failures with minikube
https://github.com/kubeflow/kubeflow/pull/613

Documentation PR in which minikube failed.

jlewi:
```
gcloud --project=kubeflow-ci compute regions list
NAME                     CPUS    DISKS_GB     ADDRESSES  RESERVED_ADDRESSES  STATUS  TURNDOWN_DATE
asia-east1               0/24    0/4096       0/8        0/8                 UP
asia-northeast1          0/24    0/4096       0/8        0/8                 UP
asia-south1              0/24    0/4096       0/8        0/8                 UP
asia-southeast1          0/24    0/4096       0/8        0/8                 UP
australia-southeast1     0/24    0/4096       0/8        0/8                 UP
europe-west1             0/24    0/4096       0/8        0/8                 UP
europe-west2             0/24    0/4096       0/8        0/8                 UP
europe-west3             0/24    0/4096       0/8        0/8                 UP
europe-west4             0/24    0/4096       0/8        0/8                 UP
northamerica-northeast1  0/24    0/4096       0/8        0/8                 UP
southamerica-east1       0/24    0/4096       0/8        0/8                 UP
us-central1              0/320   0/4096       0/32       0/8                 UP
us-east1                 72/320  16300/16384  9/32       0/8                 UP
us-east4                 0/24    0/4096       0/8        0/8                 UP
us-west1                 0/24    0/4096       0/8        0/8                 UP
```

jlewi:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/613/kubeflow-presubmit/1072/
pod_id = "kubeflow-presubmit-kubeflow-e2e-minikube-613-82d0ce4-1072-42f9-745541553"
The test logs indicate we timed out waiting for the VM to be sshable

jlewi:
I looked at one of the gcloud insert operations

```
gcloud --project=kubeflow-ci compute operations describe --zone=us-east1-d operation-1523277349726-56969a5532731-18601d17-4611c4b6  
endTime: '2018-04-09T05:36:02.873-07:00'
error:
  errors:
  - code: QUOTA_EXCEEDED
[...]

PR #620, opened 2018-04-09
Fix setup/teardown of VM for minikube.

jlewi:
* deploy step should wait for VM to be created and report any errors
if the VM can't be created; e.g. because of insufficient quota.

* teardown step should ensure the VM's disk is deleted so that we don't
  leak disks.

* Fix a bug in wait for VM; there is no op_id since we are trying to
ssh into the VM and not executing a GCE operation.

Related to #618

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/620)

### DIFF
--- a/testing/test_deploy.py
+++ b/testing/test_deploy.py
@@ -419,21 +419,36 @@ def deploy_minikube(args):
     ],
   }
   request = instances.insert(project=args.project, zone=args.zone, body=body)
+  response = None
   try:
-    request.execute()
+    response = request.execute()
+    print("done")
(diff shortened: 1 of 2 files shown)


pdmack:
:-) No way for me to test this so...
/approve
/lgtm

jlewi:
@pdmack The code is tested by the presubmit which runs the minikube E2E test. Could you re "LGTM" since there were changes due to resolving the merge conflict.

jlewi:
Manually merged since CLA check will never pass.

PR #624, opened 2018-04-09
Add openmpi package

### LINKED ISSUE #535: TF MPI support
Filing on behalf of Kai Zhang from Alibaba @wsxiaozhang -
we are looking at running tensorflow in MPI way, like horovod. in that case, there won't be PS role. Is there any plan kubeflow to support MPI?

jiezhang:
Fix #535 

Add integration for Open MPI.
This package deploys a long-running MPI playground and the MPI command should be executed using `kubectl exec` after the cluster is up.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/624)


kubeflow/openmpi/assets.libsonnet
```diff
@@ -0,0 +1,128 @@
...
+    kind: "ConfigMap",
+    apiVersion: "v1",
+    metadata: {
+      name: "openmpi-assets",
+      labels: {
+        app: params.name,
+      },
+    },
+    data: {
+      "gen_hostfile.sh": |||
```
jlewi: I think its more readable to put the scripts in files e.g. gen_hostfile.sh    and then do
importstr "gen_hostfile",

kubeflow/openmpi/master.libsonnet
```diff
@@ -0,0 +1,113 @@
+{
+  all(params):: {
+    kind: "Pod",
```
jlewi: Why is the master just a pod? Isn't that brittle? What happens if the pod fails?

kubeflow/openmpi/prototypes/openmpi.jsonnet
```diff
@@ -0,0 +1,21 @@
...
+local k = import "k.libsonnet";
+local openmpi = import "kubeflow/openmpi/all.libsonnet";
+
+// updatedParams uses the environment namespace if
+// the namespace parameter is not explicitly set
+local updatedParams = params {
+  namespace: if params.namespace == "null" then env.namespace else params.namespace,
+};
+
+std.prune(k.core.v1.list.new(openmpi.all(updatedParams)))
```
jlewi: I suggest passing params and env both to all and then extracting the namespace their as in
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/new-package-stub/all.libsonnet#L22

kubeflow/openmpi/assets.libsonnet
```diff
@@ -0,0 +1,128 @@
...
+        set -exv
+
+        # Download kubectl
+        apk add --update ca-certificates
+        apk add --update -t deps curl
+        curl -L https://storage.googleapis.com/kubernetes-release/release/v1.9.6/bin/linux/amd64/kubectl -o /usr/local/bin/kubectl
```
jlewi: Why are you installing kubectl in an init script rather than creating a Docker image?
I think baking into the DockerImage will be more performant and more reliable than downloading it via the init container.

kubeflow/openmpi/assets.libsonnet
```diff
@@ -0,0 +1,128 @@
...
+        chmod +x /usr/local/bin/kubectl
+
+        # Look up hosts and prepare hostfile for openmpi
+        hostfile=/kubeflow/openmpi/workdir/hostfile
+        until [ "$(wc -l < ${hostfile})" -eq "%(workers)d" ]; do
+          rm -f ${hostfile}
+          pod_names=$(kubectl -n %(namespace)s get pod --selector=role=worker --field-selector=status.phase=Running -o=jsonpath='{.items[*].metadata.name}')
```
jlewi: This seems a bit brittle. The selector seems underspecified; e.g. what if you have more than 1 job in a namespace?

Why not follow @inc0's suggestion in #535 of using  a stateful set? Then each worker would have a well defined hostname and we don't need to contact the APIServer to get the pod names because they are deterministic?

Alternatively, 

Why not create a simple go program to create the hostfile for the job and store it in a configmap?

The advantage of making this a go program is
- better testing
- better error handling
- better logging

Why not follow @inc0's suggestion in #535 of using  a stateful set? Then each worker would have a well defined hostname.

kubeflow/openmpi/assets.libsonnet
jiezhang: I wanted to make the app generic so that we can deploy any docker image (with openmpi installed) using this app. For example, it seems strange to me to modify the horovod package to install kubectl. With this approach, I can deploy the horovod image with no change.

kubeflow/openmpi/assets.libsonnet
jiezhang: The workers are already stateful sets and have a well defined hostname (in the form of "<APP>-worker-<INDEX>"), but I don't want to hard code the INDEX here. The selector also makes sure all the workers are in running state before generating the hostfile.

This PR doesn't support multiple jobs in the same namespace. Given that this PR is already pretty large, I'm going to address that in a follow-up PR.

In terms of embedding go program or other initialization scripts in the docker image, I agree that it'll make the code cleaner and easier to unit test, but then we would need to modify the docker images to embed those files. 

By embedding the scripts in ConfigMap, I am able to decouple the docker image and this kubeflow app. The goal is to deploy any docker images (say horovod) without understanding how kubeflow works internally.

kubeflow/openmpi/master.libsonnet
jiezhang: You're right. This PR is far from production ready. It only deploys a set of long-running master+workers today, and then the actual mpi command is executed using `kubectl exec`.  Can I address this in a follow-up PR, given that this one is already pretty large?

kubeflow/openmpi/assets.libsonnet
jiezhang: The other advantage is that I can update the scripts easily without rebuilding the docker images.

kubeflow/openmpi/assets.libsonnet
jiezhang: Thanks for the suggestion. I have moved all scripts into the assets directory.

kubeflow/openmpi/prototypes/openmpi.jsonnet
jiezhang: Done.

kubeflow/openmpi/assets.libsonnet
jlewi: It looks like the sole purpose of this script is to generate the host file. If that's the case is there any reason to run it in the same container as the workers? You could run it in a separate pod and create a configmap that would then be mounted into the workers (you could also do it in an init container).

This way you wouldn't have to modify the actual worker image and you could have a separate docker image that could have the code for generating the host file.

jlewi:
Review status: 0 of 14 files reviewed at latest revision, 5 unresolved discussions.

---

*[kubeflow/openmpi/master.libsonnet, line 3 at r1](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9gUBcOQRPg6T0aS47b-r1-3:-L9onmno7ilMdOiJ53DL:b-n36e2g) ([raw file](https://github.com/kubeflow/kubeflow/blob/65530ac029d6358e643c9c3894cdebc5d892141e/kubeflow/openmpi/master.libsonnet#L3)):*
<details><summary><i>Previously, jiezhang (Jie Zhang) wrote…</i></summary><blockquote>

You're right. This PR is far from production ready. It only deploys a set of long-running master+workers today, and then the actual mpi command is executed using `kubectl exec`.  Can I address this in a follow-up PR, given that this one is already pretty large?
</blockquote></details>

If its just a matter of changing this to a StatefulSet or ReplicaSet then I think it should be done in this PR because  whether its a Pod or StatefulSet doesn't meaningfuly cut down on the size of the PR; its roughly the same lines of code.

If there's more to making it resilient then perhaps it can wait.

---

*[kubeflow/openmpi/master.libsonnet, line 7 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9onjr71OxGxqgzx4i_:-L9onjr71OxGxqgzx4ia:bkuah2q) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/master.libsonnet#L7)):*
> ```
>     metadata: {
>       name: "openmpi-master",
>       labels: {
> ```

Set namespace please

---

*[kubeflow/openmpi/secrets.libsonnet, line 2 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9okRG3-IjUBGpwWnbH:-L9okRG3-IjUBGpwWnbI:b-dyw9tw) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/secrets.libsonnet#L2)):*
> ```
> {
>   all(params):: {
> ```

Secrets should not be created via ksonnet because ksonnet is intended to be checked into source control and you wouldn't want to check private keys into source control.

The public key and authorized keys are probably ok to check in (please add a comment explaining that).

For the id_rsa the pattern would be to tell the user to manually create the secret e.g.
kubectl create secret generic id_rsa --from-file=is_rsa=/path/to/key

And then the component would take as a parameter the name of the secret.

---

*[kubeflow/openmpi/secrets.libsonnet, line 6 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9onPNwC6AY9lsi9CQR:-L9onPNwC6AY9lsi9CQS:b-nekfac) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/secrets.libsonnet#L6)):*
> ```
>     apiVersion: "v1",
>     metadata: {
>       name: "openmpi-secrets",
> ```

set namespace please.

---

*[kubeflow/openmpi/service.libsonnet, line 6 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9ol73W1tonNRnSmYXL:-L9ol73W1tonNRnSmYXM:btzuq2d) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/service.libsonnet#L6)):*
> ```
>     apiVersion: "v1",
>     metadata: {
>       name: params.name,
> ```

Here and for all components can you add a namespace parameter and set it like so
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/new-package-stub/all.libsonnet#L22

so that it can be automatically inherited from the ksonnet environment?

---

*[kubeflow/openmpi/worker.libsonnet, line 6 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9olSzrC5RQoCsf52sO:-L9olSzrC5RQoCsf52sP:b5meuyg) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/worker.libsonnet#L6)):*
> ```
>     apiVersion: "apps/v1",
>     metadata: {
>       name: "openmpi-worker",
> ```

set namespace

---

*[kubeflow/openmpi/assets/gen_hostfile.sh, line 14 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9om2ZCB_qIkQiCZNry:-L9olt-t-BgASNcBqM75:b-6uwwcr) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/assets/gen_hostfile.sh#L14)):*
> ```
> o
> ```
Per comments elsewhere why does the hostfile need to be generated at runtime? It looks like all the information is known ahead of time; e.g. we know the number of workers and each worker has a stable FQDN based on its name and namespace since its a statefulset.

So the hostnames will be 

${statefulset name}-0.${service_name}.${namespace}.svc.cluster.local
...
${statefulset name}-N.${service_name}.${namespace}.svc.cluster.local

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624)*

jiezhang:
Review status: 0 of 14 files reviewed at latest revision, 11 unresolved discussions.

---

*[kubeflow/openmpi/assets.libsonnet, line 18 at r1](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9gUBcOQRPg6T0aS47a-r1-18:-L9pjmZd065vPZZS7Icn:blkj911) ([raw file](https://github.com/kubeflow/kubeflow/blob/65530ac029d6358e643c9c3894cdebc5d892141e/kubeflow/openmpi/assets.libsonnet#L18)):*

It looks like the sole purpose of this script is to generate the host file. If that's the case is there any reason to run it in the same container as the workers? You could run it in an init container that would emit the hosts file to a volume on the pod and then mount that into the actual worker image.

This way you wouldn't have to modify the actual worker image and you could have a separate docker image that could have whatever tools are needed to generate the hostsfile.
</blockquote></details>

There are two scripts:
* gen_hostfile.sh runs in master's initContainer to generate the hostfile. It's currently using "alpine" image and therefore I have to install kubectl during initialization. If there is an official kubernetes docker image that contains kubectl, I can definitely switch over to that.
* init.sh runs in the container of both master and worker. Among other things, it sets up the sshd which is required to run openmpi. I don't think I can run sshd in the initContainer as it has to be long lived. 

---

*[kubeflow/openmpi/secrets.libsonnet, line 2 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9okRG3-IjUBGpwWnbH:-L9ppgdd44BiW4MQaUul:b-8grrve) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/secrets.libsonnet#L2)):*

Secrets should not be created via ksonnet because ksonnet is intended to be checked into source control and you wouldn't want to check private keys into source control.

The public key and authorized keys are probably ok to check in (please add a comment explaining that).

For the id_rsa the pattern would be to tell the user to manually create the secret e.g.
kubectl create secret generic id_rsa --from-file=is_rsa=/path/to/key

And then the component would take as a parameter the name of the secret.
</blockquote></details>

It's not created via ksonnet. The public and private key are both passed in as a parameter. See README.md for more details.
`ks generate openmpi ${COMPONENT} --image ${IMAGE} --pubkey=${PUBKEY} --prikey=${PRIKEY}`

---

*[kubeflow/openmpi/assets/gen_hostfile.sh, line 14 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9om2ZCB_qIkQiCZNry:-L9pqgmT3bbcz_OZu6zK:b-hbrcqd) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/assets/gen_hostfile.sh#L14)):*

> ```
> o
> ```
Per comments elsewhere why does the hostfile need to be generated at runtime? It looks like all the information is known ahead of time; e.g. we know the number of workers and each worker has a stable FQDN based on its name and namespace since its a statefulset.

So the hostnames will be 

${statefulset name}-0.${service_name}.${namespace}.svc.cluster.local
...
${statefulset name}-N.${service_name}.${namespace}.svc.cluster.local
</blockquote></details>

In addition to finding the statefulset name, the selector `--field-selector=status.phase=Running` ensures that all workers are in running status before we move on. The synchronization is needed if we want to kick off the training job right after init.sh.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624)*

jiezhang:
Review status: 0 of 15 files reviewed at latest revision, 11 unresolved discussions.

---

*[kubeflow/openmpi/assets.libsonnet, line 12 at r1](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9gUBcOQRPg6T0aS47a-r1-12:-L9q9MZdAq_IjcbEe_Dg:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/65530ac029d6358e643c9c3894cdebc5d892141e/kubeflow/openmpi/assets.libsonnet#L12)):*
<details><summary><i>Previously, jiezhang (Jie Zhang) wrote…</i></summary><blockquote>

Thanks for the suggestion. I have moved all scripts into the assets directory.
</blockquote></details>

Done.

---

*[kubeflow/openmpi/master.libsonnet, line 3 at r1](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9gUBcOQRPg6T0aS47b-r1-3:-L9q8aSn7rTNLQCCTF4N:b-pecm7n) ([raw file](https://github.com/kubeflow/kubeflow/blob/65530ac029d6358e643c9c3894cdebc5d892141e/kubeflow/openmpi/master.libsonnet#L3)):*

If its just a matter of changing this to a StatefulSet or ReplicaSet then I think it should be done in this PR because  whether its a Pod or StatefulSet doesn't meaningfuly cut down on the size of the PR; its roughly the same lines of code.

If there's more to making it resilient then perhaps it can wait.
</blockquote></details>

ReplicaSet won't work as it doesn't have a stable pod name. I can change this into a StatefulSet but then the master pod name would be something like "openmpi-master-0" which seems weird.  Can you elaborate the benefits of using StatefulSet vs Pod when replicas is set to 1? 

---

*[kubeflow/openmpi/master.libsonnet, line 7 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9onjr71OxGxqgzx4i_:-L9q9xgq5scx5bMPo36K:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/master.libsonnet#L7)):*

Set namespace please
</blockquote></details>

Done.

---

*[kubeflow/openmpi/secrets.libsonnet, line 6 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9onPNwC6AY9lsi9CQR:-L9qA-J172ZBFSc_8Vgm:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/secrets.libsonnet#L6)):*

set namespace please.
</blockquote></details>

Done.

---

*[kubeflow/openmpi/service.libsonnet, line 6 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9ol73W1tonNRnSmYXL:-L9qA0Tf-m8G0Q1mxw2Y:bbdnpna) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/service.libsonnet#L6)):*

Here and for all components can you add a namespace parameter and set it like so
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/new-package-stub/all.libsonnet#L22

so that it can be automatically inherited from the ksonnet environment?
</blockquote></details>

I am overriding namespace of params in all.libsonnet and use params.namespace in the rest of libsonnets (just like other packages). Does that look okay to you? Or you want to explicitly add a namespace parameter to all libsonnet?

---

*[kubeflow/openmpi/worker.libsonnet, line 6 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9olSzrC5RQoCsf52sO:-L9qCLD85Vi_pdh311ZK:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/worker.libsonnet#L6)):*

set namespace
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624)*

jlewi:
Review status: 0 of 15 files reviewed at latest revision, 9 unresolved discussions.

---

*[kubeflow/openmpi/master.libsonnet, line 3 at r1](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9gUBcOQRPg6T0aS47b-r1-3:-L9qiDV0D2S7KAfEVv89:brhby2x) ([raw file](https://github.com/kubeflow/kubeflow/blob/65530ac029d6358e643c9c3894cdebc5d892141e/kubeflow/openmpi/master.libsonnet#L3)):*
<details><summary><i>Previously, jiezhang (Jie Zhang) wrote…</i></summary><blockquote>

ReplicaSet won't work as it doesn't have a stable pod name. I can change this into a StatefulSet but then the master pod name would be something like "openmpi-master-0" which seems weird.  Can you elaborate the benefits of using StatefulSet vs Pod when replicas is set to 1? 
</blockquote></details>

Pods aren't durable. E.g. if the VM goes down the pod won't be recreated but if you use a stateful set the pod will be recreated. So I think statefulset is better.

---

*[kubeflow/openmpi/secrets.libsonnet, line 2 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9okRG3-IjUBGpwWnbH:-L9qiW3KDMFMufEpq7EI:brq4elg) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/secrets.libsonnet#L2)):*
<details><summary><i>Previously, jiezhang (Jie Zhang) wrote…</i></summary><blockquote>

It's not created via ksonnet. The public and private key are both passed in as a parameter. See README.md for more details.
`ks generate openmpi ${COMPONENT} --image ${IMAGE} --pubkey=${PUBKEY} --prikey=${PRIKEY}`
</blockquote></details>

But the effect of doing ks generate is that you create a params.json file in your ksonnet application that will store whatever values you have for the parameters. 

The expectation is that your ksonnet application will be checked into source control so that your configs can be versioned.

So the net result is if people follow standard operating procedure with ksonnet they would end up checking in their private key into source control which is not a good idea.

So the current recommended practice is to tell people to manually create K8s secrets by running kubectl
and then specify the name of the K8s secret in their configs.

---

*[kubeflow/openmpi/assets/gen_hostfile.sh, line 14 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9om2ZCB_qIkQiCZNry:-L9qjDN32CrNcjw4cQHM:bg7uuk7) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/assets/gen_hostfile.sh#L14)):*
<details><summary><i>Previously, jiezhang (Jie Zhang) wrote…</i></summary><blockquote>

In addition to finding the statefulset name, the selector `--field-selector=status.phase=Running` ensures that all workers are in running status before we move on. The synchronization is needed if we want to kick off the training job right after init.sh.
</blockquote></details>

Can we generate the hostfile statically and just use the init script to block until the jobs are ready?
This is a step in the right direction and we can figure out a better pattern for how to block until all workers have started.

I think a better way to block until all workers are started would be to run the kubectl check in a sidecar and then create a sentinel file on a pod volume that could be mounted into your main container.

Your main container could then just block until the sentinel file exists.

The advantage of this approach is that your side car could have its own dedicated Docker image which has kubectl baked into it so we don't have the variability that comes with downloading kubectl on startup.

We could do that in a follow on PR (but lets) file an issue.

However, I think in this PR we should generate the hostfile via ksonnet and store it in a configmap.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624)*

jiezhang:
Review status: 0 of 13 files reviewed at latest revision, 9 unresolved discussions.

---

*[kubeflow/openmpi/master.libsonnet, line 3 at r1](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9gUBcOQRPg6T0aS47b-r1-3:-L9rMYrQ2EmUxCOMBh-u:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/65530ac029d6358e643c9c3894cdebc5d892141e/kubeflow/openmpi/master.libsonnet#L3)):*

Pods aren't durable. E.g. if the VM goes down the pod won't be recreated but if you use a stateful set the pod will be recreated. So I think statefulset is better.
</blockquote></details>

Done.

---

*[kubeflow/openmpi/secrets.libsonnet, line 2 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9okRG3-IjUBGpwWnbH:-L9rMVS32SVac-lNOoCC:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/secrets.libsonnet#L2)):*

But the effect of doing ks generate is that you create a params.json file in your ksonnet application that will store whatever values you have for the parameters. 

The expectation is that your ksonnet application will be checked into source control so that your configs can be versioned.

So the net result is if people follow standard operating procedure with ksonnet they would end up checking in their private key into source control which is not a good idea.

So the current recommended practice is to tell people to manually create K8s secrets by running kubectl
and then specify the name of the K8s secret in their configs.
</blockquote></details>

Done.

---

*[kubeflow/openmpi/assets/gen_hostfile.sh, line 14 at r2](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624#-L9om2ZCB_qIkQiCZNry:-L9rM1Vs281eoO7_eiVG:b-w4p96e) ([raw file](https://github.com/kubeflow/kubeflow/blob/d130abb9ad1cfd2516699fd3388817435061840f/kubeflow/openmpi/assets/gen_hostfile.sh#L14)):*

Can we generate the hostfile statically and just use the init script to block until the jobs are ready?
This is a step in the right direction and we can figure out a better pattern for how to block until all workers have started.

I think a better way to block until all workers are started would be to run the kubectl check in a sidecar and then create a sentinel file on a pod volume that could be mounted into your main container.

Your main container could then just block until the sentinel file exists.

The advantage of this approach is that your side car could have its own dedicated Docker image which has kubectl baked into it so we don't have the variability that comes with downloading kubectl on startup.

We could do that in a follow on PR (but lets) file an issue.

However, I think in this PR we should generate the hostfile via ksonnet and store it in a configmap.

</blockquote></details>

Sounds good. I have moved hostfile into ConfigMap (assets.libsonnet).

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/kubeflow/kubeflow/624)*

ddysher:
@jlewi  yes, the gen_hostfile script; but my original question is not specific to this PR :)

I haven't really looked deeply in how we use ksonnet, do you have any pointers for why we use ksonnet?There's quite a few comments in this thread, so I put my previous comment here again:

- ksonnet should really be a syntax sugar to yaml file, and an easier way to deploy applications; but if we start adding custom scripts in it, things will eventually go south
- ksonnet makes it harder to get started with kubeflow IMO. I find it is listed as a prerequisite even to just deploy a single component in kubeflow.

/cc @gaocegege

gaocegege:
@ddysher @jlewi 

Yeah, I do not understand why we use ksonnet in Kubeflow. It is complex and increases the learning cost of Kubeflow. When setup a new repository, ksonnet implementation is a big overhead for the owners.

When I collaborate with @YujiOshima on hp-tuning, I find it is really hard to write the deployment using ksonnet. 

I tried to invited more people engaged into the community and contribute their work, but I found that ksonnet is a huge barrier.

I am glad to have a discussion about people in the community about the tool to deploy kubeflow. It scares off many potential contributors, IMO.

jlewi:
@jiezhang This is fantastic; thanks for agreeing to the changes.

@gaocegege I agree that a controller might be the right way to go. My initial thought that is that we should largely be able to repurpose the existing TFJob controller (maybe with a few generalizations).

@ddysher @gaocegege  I don't think we do or should use ksonnet for anything other than YAML templating. I agree there's friction around the UX but I think its solvable. In its most basic form ksonnet is just json; so anyone capable of writing a K8s manifest should be able to create a jsonnet component. Not sure if you noticed but I recently created a [guide](https://github.com/kubeflow/kubeflow/blob/master/docs_dev/ksonnet_packages.md#developer-guide-for-ksonnet-packages) to explain how to create a Kubeflow package; I also created a stub package that people could copy/paste to create a new package.

/lgtm
/approve

jiezhang:
@jlewi Thanks for reviewing the code. I just refactored the code and merged master.libsonnet and worker.libsonnet into workloads.libsonnet.

jiezhang:
@gaocegege @ddysher This package is not tightly coupled with horovod. The goal is to run any docker image with Open MPI installed. It overrides the entrypoint of the docker image to do all the k8s initialization. That's why I bundle some "assets" (scripts and configs) in the package. Without those assets, users of this package would need to understand k8s and modify their docker image to run in k8s.

jiezhang:
@jlewi Thanks. I've joined the Kubeflow org. I refactored the code in master & worker into workloads. Can you take another look?

jlewi:
nice

/lgtm
/approve

ddysher:
@jiezhang thanks for adding mpi support in kubeflow!

PR #633, opened 2018-04-10
Add apache beam package for python2

ankushagarwal:
/cc @jlewi


components/tensorflow-notebook-image/Dockerfile
```diff
@@ -227,6 +227,7 @@ RUN conda_packages=$(conda list -e | cut -d '=' -f 1 | grep -v '#' | sort) && \
       jupyter nbextension install --py --symlink tensorflow_model_analysis --system && \
       jupyter nbextension enable --py tensorflow_model_analysis --system; \
     fi && \
+    pip install --no-cache-dir 'apache-beam' 'apache-beam[gcp]' && \
```
jlewi: I checked in our notebook images and it seems apache_beam is already installed.
It looks like that's because tf-model-analysis pulls it in
https://github.com/tensorflow/model-analysis/blob/master/setup.py#L189

Explicitly installing it isn't a bad idea but I wonder if we should do it before tf.transform/tfma

pdmack: Probably good that developers see that it's explicitly added (wherever the right place should be).

PR #635, opened 2018-04-10
Update tensorflow serving image and tensorflow notebook image registries

ankushagarwal:
Point them to kubeflow-images instead of kubeflow-images-staging

Related to https://github.com/kubeflow/kubeflow/issues/534
/cc @jlewi


ankushagarwal:
@inc0 Can you review again?

PR #642, opened 2018-04-11
Change naming from jupyter to jupyterhub when referring to hub

willingc:
This PR should add more consistent naming (and less confusion down the road) by using jupyterhub when referencing the JupyterHub hub. Typically, the Jupyter team refers to the Jupyter notebook or the team itself when using Jupyter name. JupyterHub is used to reference all hub related components and activities.

@jlewi I think I caught all the non-doc places that this occurred.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/642)


jlewi:
Do you also want to rename https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/jupyterhub_spawner.py?

Since we aren't actually spawning JupyterHub?

willingc:
@jlewi Hmm... I didn't think of that one before. It may be best as kubeform_spawner. I guess it depends if you think there may be swapping of spawners over time. You make the call :-)

jlewi:
kubeform_spawner seems good to me.

Issue #106, opened 2018-01-09
Figure out ksonnet repo organization


jlewi:
We need to figure out how we want to organize our ksonnet registry in terms of [packages and prototypes](https://ksonnet.io/docs/concepts).

I think we have at least two different types of entities

- Individual components (e.g. JupyterHub, Argo, TensorFlow CRD, TF Serving, etc...)
- Stacks - collections of components

I think we will have multiple stacks which provide different collections of components; for example I could imagine a "Hadoop + TensorFlow" stack aimed at combining spark and TensorFlow components.

So my initial thought is that we should have individual ksonnet packages for each component and each stack.

We currently have 3 packages:

- Core
- tf-job
- tf-serving

However, these packages don't map to the entities mentioned above. If we were to follow the proposed model I think we would need to make the following changes

- Move TfJob CRD out of core and into tf-job
- Create a Package for JupyterHub
- Turn "Core" into a suitable stack maybe "Tensorflow" that reuses other packages.

I think the main issue with packages is that I don't think ksonnet provides a way to declare package dependencies. So users would have to manually install each package and its dependencies. This is fine as long as we have a small number of dependencies. This seems like a fixable issue though.

/cc @bryanl

ankushagarwal:
Closing this as duplicate of #376

PR #649, opened 2018-04-11
Pytorch ksonnet

jose5918:
- Adds pytorch-job pkg
  - Contains pytorch-operator prototype
  - Contains pytorch-job prototype

Motivation for this is to have a way to install the pytorch-operator from Kubeflow's registry.  This follows current Kubeflow patterns.

Aimed at addressing: https://github.com/kubeflow/pytorch-operator/issues/24

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/649)


kubeflow/core/prototypes/all.jsonnet
```diff
@@ -16,6 +16,8 @@
 // @optionalParam jupyterNotebookPVCMount string null Mount path for PVC. Set empty to disable PVC
 // @optionalParam reportUsage string false Whether or not to report Kubeflow usage to kubeflow.org.
 // @optionalParam usageId string unknown_cluster Optional id to use when reporting usage to kubeflow.org
+// @optionalParam pytorchJobImage string jgaguirr/pytorch-operator:latest The image for the PyTorchJob controller
```
jose5918: @jlewi Is there any place I can release the latest image to?

jlewi: kubeflow-images-staging

jlewi: gcr.io/kubeflow-images-staging

jose5918: @jlewi How do I push there?

jlewi: The Kubeflow release team has permission to push there.

We should probably automate this.

jlewi: Our release process is described here:
https://github.com/kubeflow/kubeflow/blob/master/docs_dev/releasing.md

So the steps would be something like the following

1. Create an Argo workflow as an E2E test builds the image and runs some tests to validate it.
1. Add a section to the release guide about how to build and release the operator image
1. Request someone in the kubeflow-release team does this

We should automate this so we can push images on each post submit. 

We could also potentially make our CI images public. Not sure what the best approach would be.

jlewi: We have gcr.io/kubeflow-images-public now. So you could use that as the URI

jlewi:
The current structure of our packages makes no sense; #106.
In particular I think the way we split TF across two packages.

My suggestion would be to put all the pytorch components in one package since they will most likely be used holistically.

I would suggest not putting them in their own package and not core.  This will make it easier to version them differently from core.

kubeflow/pytorch-job/prototypes/pytorch-job.jsonnet
```diff
@@ -0,0 +1,55 @@
...
+// @optionalParam image string null The docker image to use for the job.
+// @optionalParam image_gpu string null The docker image to use when using GPUs.
+// @optionalParam num_masters number 1 The number of masters to use
+// @optionalParam num_ps number 0 The number of ps to use
+// @optionalParam num_workers number 0 The number of workers to use
+// @optionalParam num_gpus number 0 The number of GPUs to attach to workers.
+
+// TODO(https://github.com/ksonnet/ksonnet/issues/235): ks param set args won't work if the arg starts with "--".
+
+local k = import "k.libsonnet";
```
jlewi: I suggest putting all all logic into the library and just passing to the library params and env.
Example
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/new-package-stub/prototypes/tf-prototype.jsonnet

This should be easier to maintain.

kubeflow/pytorch-job/prototypes/pytorch-operator.jsonnet
```diff
@@ -0,0 +1,21 @@
...
+local k = import "k.libsonnet";
+local operator = import "kubeflow/pytorch-job/pytorch-operator.libsonnet";
+
+// updatedParams uses the environment namespace if
+// the namespace parameter is not explicitly set
+local updatedParams = params {
+  namespace: if params.namespace == "null" then env.namespace else params.namespace,
+};
+
+std.prune(k.core.v1.list.new(operator.all(updatedParams)))
```
jlewi: Same comment as above; 
please follow
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/new-package-stub/prototypes/tf-prototype.jsonnet

kubeflow/pytorch-job/pytorch-job.libsonnet
```diff
@@ -0,0 +1,49 @@
+local k = import "k.libsonnet";
+
+{
+  parts:: {
```
jlewi: The list of components comprising the different prototypes should be listed in the libsonnet file as noted
here
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/new-package-stub/all.libsonnet#L5

This makes it easier to maintain/develop because we don't need to upgrade the prototype just the vendored library

kubeflow/pytorch-job/pytorch-operator.libsonnet
```diff
@@ -0,0 +1,281 @@
+{
+  all(params):: [
```
jlewi: Pass in env as well and get the namespace from env. We should no longer need to use params for the namespace

jose5918:
@jlewi Does that address the comments you had? Also is it possible to get someone to try releasing an image using the instructions in the pytorch repo?

jlewi:
I don't see any instructions on how to release an image in pytorch-operator.

kubeflow/pytorch-job/pytorch-job.libsonnet
```diff
@@ -0,0 +1,79 @@
...
+
+    local workerSpec = if numGpus > 0 then
+      util.pytorchJobReplica("WORKER", numWorkers, args, imageGpu, numGpus)
+    else
+      util.pytorchJobReplica("WORKER", numWorkers, args, image),
+    local masterSpec = util.pytorchJobReplica("MASTER", numMasters, args, image),
+    local replicas = [ masterSpec, workerSpec ],
+
+
+    PyTorchJob:: {
```
jlewi: nit: pyTorchJob seems more consistent

PR #650, opened 2018-04-12
[Azure] Update nvidia driver volumes for AKS

wbuchwalter:
Drivers installation was updated recently for AKS, this PR update our ks template to reflect that change.

I'm keeping the distinction between aks and acsengine even though they are identical today, because acs-engine is going to receive the update for device-plugin pretty soon.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/650)


jlewi:
Looks like a 502 calling GitHub API. Hopefully just a flake.

/lgtm
/approve

PR #655, opened 2018-04-13
[Azure] Jupyter spawner: driver volumes for azure

### LINKED ISSUE #435: [Jupyter/Azure] Drivers are not mounted when spawning a JupyterHub with GPU
When a user is running on Azure, and is asking the Jupyter spawner for a new JupyterHub with a GPU, the drivers are not mounted into the container, making the GPU unusable.
In the `tf-operator` we fixed this by having a `ConfigMap` with the necessary mount points for each different cloud.

This is the cause of https://github.com/kubeflow/kubeflow/issues/414 (at least for the original submitter).

cc @dharmeshkakadia

wbuchwalter:
Fix #435 
Currently Azure does not use device plugins, so the NVIDIA drivers need to be mounted for GPUs to work.
This is meant to be a temporary fix and will be removed as soon as AKS supports device-plugin.

cc @willingc

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/655)


wbuchwalter:
/assign @jlewi @ankushagarwal 
Could one of you take a loot ?

PR #654, opened 2018-04-13
Bootstraper polish:

kunmingg:
1. Use namespace pass in argument (--namespace=) or defalut namespace ('kubeflow'),
   create it if not exists.
2. Add unit test pkg (github.com/stretchr/testify) to gilde.
3. Edit Makefile for macOS compatibility

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/654)

PR #670, opened 2018-04-16
make Dockerfile & Makefile cross platform 

kunmingg:
by building bootstrap package within docker.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/670)

### DIFF
--- a/bootstrap/Dockerfile
+++ b/bootstrap/Dockerfile
@@ -1,6 +1,8 @@
 # TODO(jlewi): How can we make it work with golang:1.8.2-alpine
 FROM golang:1.8.2
 
+ARG KUBEFLOW_REPO=https://github.com/kubeflow/kubeflow.git
+
 # We need gcloud to get gke credentials.
 RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz && \
     tar xzf google-cloud-sdk.tar.gz -C / && \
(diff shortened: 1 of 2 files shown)


kunmingg:
@jlewi 
make build will not work till https://github.com/kubeflow/kubeflow/pull/654 (contains missing vendor dependencies) merged.
You can test it by ```make build REPO=https://github.com/kunmingg/kubeflow.git```

bootstrap/Dockerfile
```diff
@@ -1,6 +1,8 @@
 # TODO(jlewi): How can we make it work with golang:1.8.2-alpine
 FROM golang:1.8.2
 
+ARG KUBEFLOW_REPO=https://github.com/kubeflow/kubeflow.git
```
jlewi: For local builds can we access the files from a volume mount so that we don't have to clone the repo?

bootstrap/Dockerfile
```diff
@@ -37,7 +39,14 @@ RUN cd /opt/ksonnet && \
...
 RUN mkdir -p /opt/kubeflow
 
-COPY bootstraper /opt/kubeflow/
+ENV PATH /go/bin:/usr/local/go/bin:$PATH
+
+RUN mkdir -p $GOPATH/src/github.com/kubeflow && \
+    cd $GOPATH/src/github.com/kubeflow && \
+    git clone $KUBEFLOW_REPO
+
+RUN go build -i -o /opt/kubeflow/bootstraper ${GOPATH}/src/github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/main.go
```
jlewi: How about using multi-stage builds so the final image doesn't include the source tree?
https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

kunmingg: Good point, updated

bootstrap/Dockerfile
kunmingg: Now copy code from local repo during stage 1.

PR #671, opened 2018-04-17
Add batch support to openmpi package

jiezhang:
* Introduce "exec" parameter to run adhoc command

* Use redis for synchronization between master and worker

* Use pod instead of stateful set to run the workloads. This is needed to set restartPolicy to Never.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/671)


jiezhang:
cc @alsrgv

jlewi:
What's the fault tolerance model? If we're using pods and a pod gets preempted (e.g. node is under memory pressure or healthy); what happens to the job? Will it exit or recover?

@jiezhang and I chatted briefly in slack and IIUC there is an internal system that will be managing the job and handling these failures. So Kubeflow might need another solution (e.g. a CRD).

I don't think we need to resolve this but would be good to open an issue to track this.

jiezhang:
@jlewi In those cases, the job would fail. We probably need a job management service to monitor job status and reschedule the job as needed.

Note that even if the container is managed by StatefulSet, if one of the worker containers gets evicted while the job is running, the job will fail and cannot recover itself after StatefulSet provisions a new container. We still need an external service to monitor job status and provide fault tolerance.

Given MPI jobs involve multiple containers, I think it would be challenging to make each container fault tolerant. It's easier to re-deploy all the components if something goes wrong.

jiezhang:
@jlewi I have opened #677 to track it.

PR #680, opened 2018-04-19
create rolebinding within namespace to guarantee permission

kunmingg:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/680)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/server.go
+++ b/bootstrap/cmd/bootstrap/app/server.go
@@ -42,6 +42,8 @@ import (
 	k8sVersion "k8s.io/apimachinery/pkg/version"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
+	"os/exec"
+	"errors"
 )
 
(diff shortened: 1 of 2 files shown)


bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -201,6 +205,19 @@ func Run(opt *options.ServerOption) error {
 		return err
 	}
 
+	log.Infof("create rolebinding kubeflow-admin for role permission")
```
jlewi: Any particular reason we're shelling out to kubectl rather than using the goclient to make a request to the API?

bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -201,6 +205,19 @@ func Run(opt *options.ServerOption) error {
...
+	user, err := exec.Command("gcloud", "config", "get-value", "account").Output()
```
jlewi: We shouldn't be running gcloud commands unless we are using GKE. You can use the isGke function to check.

PR #687, opened 2018-04-19
Use a CentOS 7 base image for k8s-model-server

tmckayus:
Similar motivation as discussed in https://github.com/kubeflow/tf-operator/pull/469

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/687)


jlewi:
/unassign @gaocegege 
/assign @lluunn 

@lluunn is the primary owner of model server stuff. @lluunn it would be good to create an OWNERS file in the appropriate directory and add yourself to indicate that you are the primary owner.

components/k8s-model-server/images/Dockerfile.cpu
```diff
@@ -12,42 +12,75 @@
...
-RUN curl -O http://storage.googleapis.com/tensorflow-serving-apt/pool/tensorflow-model-server/t/tensorflow-model-server/tensorflow-model-server_1.6.0_all.deb
-RUN dpkg -i tensorflow-model-server_1.6.0_all.deb
+# Set up grpc
+ENV LD_LIBRARY_PATH=/opt/rh/python27/root/usr/lib64
+RUN pip install --upgrade pip
+RUN pip install mock grpcio numpy
 
+# Set up Bazel.
+
+ENV BAZELRC /root/.bazelrc
```
jlewi: @lluunn I'm guessing the switch to centos forces us to build TF serving ourself since they probably don't provide a YUM package.

I think we're already building it in the case of the GPU image. If that's the case I think being consistent is good. Does TF-serving publish a deb for GPU?

crobby:
In case it is preferable to have a leaner image that just installs the tensorflow_model_server package, I am working on such a version.  I currently have a tensorflow_model_server RPM package being built in copr: [here's my copr repo](https://copr.fedorainfracloud.org/coprs/croberts/tensorflow-model-server/).  I have an image built with it, but it is not tested yet.
I have also looked at making a similar RPM package for the gpu version, but it does not look like it will be possible due to the way the cuDNN bits are published.  If I'm wrong on that, please correct me and I can advance on that front.

jlewi:
@tmckayus Is this PR ready for review? If it is please remove WIP in from the title so that reviewers know its ready.

tmckayus:
@jlewi, thanks, removed ... conflicts will be fixed ASAP

marcoceppi:
I don't truly see an advantage to moving to CentOS which does not follow a reliable release schedule and provides a large image than the current one provided. The reasons in the previous thread truly aren't any different than what is provided in the current base image.

Unless there's a clear advantage for tensorflow (which has traditionally been based on Ubuntu).

For comparison. `ubuntu:16.04` image is 43MB `centos:6` is 70MB. As a follow up I'll be happy to demonstrate how packages provided in centos are older than that in 16.04.

Machine learning is moving fast, setting base images to larger, older software doesn't provide an advantage imo.

jlewi:
I think there are two paths forward

With Kubeflow we don't want to be in the business of picking between overlapping components; we can and do support a multiplicity of options (e.g. TensorFlow and PyTorch).

The same philosophy can be extended to base images. 

So from a project perspective we can support multiple flavors of an image (we are already building multiple variants of the images to support multiple versions of TF). 

For this particular image it doesn't look like supporting multiple base images would be all that difficult. I think the only difference is the package install commands and we could just have two different scripts.

That said there would be some work to update the build & release process. There would also be an ongoing maintenance cost.

If someone is willing to sign up to take on the extra work of supporting multiple base images then that's one option we can consider.

If no one is willing to take on the extra work, then we can as a community try to figure out the criterion by which we are going to pick base images.

elmiko:
@marcoceppi for me, i like having the security enhancements and audit trail that centos bakes in, i also feel the centos base is more suitable for production type environments.

that said, i definitely acknowledge that ubuntu has more recent packages included within. is there a way we can leverage the community maintenance angle to provide both sources?

i truly feel our users will understand the difference and ideally there will similar functionality provided by both.

willingc:
I agree with @jlewi's thoughts. It would be nice to see multiple base images (at least Ubuntu and Centos), but it does add some work to support and test multiple images.

marcoceppi:
We should pause this particular change then until that discussion settles

jlewi:
I'm closing this PR for now.
multi-distro support was brought up at the last community meeting and there is a proposal floating around
https://docs.google.com/document/d/1Wdxt1xedAj7qF_Rjmxy1R0NRdfv7UWs-r2PItewxHpE/edit#heading=h.x8csktfewyn8

Will wait for relevant folks to figure out next steps.

Issue #686, opened 2018-04-19
ks 0.10.0-alpha.2 does not work with install instructions


mattf:
$ ks version
```
ksonnet version: 0.10.0-alpha.2
jsonnet version: v0.10.0
client-go version: `1.8`
```
$ ks apply default -c kubeflow-core
```
ERROR find objects: RUNTIME ERROR: couldn't open import "kubeflow/core/all.libsonnet": no match locally or in the Jsonnet library paths
    <extvar:__ksonnet/components>:5:15-51    thunk <all> from <object <anonymous>>
    <extvar:__ksonnet/components>:10:32-35    thunk from <thunk from <object <anonymous>>>
    ~/my-kubeflow/lib/v1.9.1/k.libsonnet:96:24-29    thunk from <function <anonymous>>
    ~/my-kubeflow/lib/v1.9.1/k.libsonnet:97:36-41    thunk from <function <anonymous>>
    <builtin>    builtin function <type>
    <std>:1165:25-26    thunk from <thunk <ta> from <function <anonymous>>>
    <builtin>    builtin function <type>
    <std>:1167:29-31    thunk from <function <anonymous>>
    <builtin>    builtin function <primitiveEquals>
    <std>:1167:8-36    function <anonymous>
        
    ~/my-kubeflow/lib/v1.9.1/k.libsonnet:96:13-30    function <anonymous>
    ~/my-kubeflow/lib/v1.9.1/k.libsonnet:(92:22)-(96:30)    function <anonymous>
    <extvar:__ksonnet/components>:10:13-61    thunk from <object <anonymous>>
    <std>:1216:24-25    thunk from <thunk <t> from <function <anonymous>>>
    <builtin>    builtin function <type>
    <std>:1217:8-9    function <anonymous>
    <std>:1165:25-26    thunk from <thunk <ta> from <function <anonymous>>>
    <builtin>    builtin function <type>
    <std>:1167:29-31    thunk from <function <anonymous>>
    <builtin>    builtin function <primitiveEquals>
    <std>:1167:8-36    function <anonymous>
        
    <extvar:__ksonnet/components>:10:3-62    object <anonymous>
    During `manifestation`
```

ankushagarwal:
Hi @mattf does ks 0.9.2 work?

aronchick:
[...]

PR #694, opened 2018-04-20
Pin instructions to ks 0.9.2 (for now)

pdmack:
Seems like enough Kubeflow users have been bitten by some of the ks 0.10 stuff that we should try to help them first and ksonnet second.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/694)


pdmack:
#686

PR #695, opened 2018-04-20
Migrate images to kubeflow-images-public

ankushagarwal:
Related to https://github.com/kubeflow/kubeflow/issues/534

/cc @kunmingg 
/cc @pdmack

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/695)


pdmack:
Looks good. So I guess we'll sequence this before pointing to the newly minted notebook images in the hub dropdown? @ankushagarwal I guess those will also be pushed to "kubeflow-images-public"?
Or should we update that too here in this PR?

ankushagarwal:
Right now I am cloning all images from staging to public, once that's done and tests pass here, we can push this and then you can send a PR to point jupyter notebook the `v20180419-0ad94c4e` tag

PR #696, opened 2018-04-20
openmpi: master/workers sync mechanism is replaced with k8s api from redis

everpeace:
Currently, openmpi package uses redis to sync master/workers.  But I think it is overkilled a bit.  And we can watch master/workers status via k8s api.

In a pod, service account's token and kubernetes api server endpoint is embedded. So, we can see pods status just executing curl like this.

```
# I think --insecure is safe because this will be executed in a pod (inside of kubernetes cluster)
curl --insecure \
  --header "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
  https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT/api/v1/namespaces/$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)/pods
```

So, I replaced master/workers sync mechanism with k8s api from redis.

What do you think??  I'm very happy if I could have some feedbacks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/696)


everpeace:
cc/ @jiezhang @pdmack would you mind reviewing this?

everpeace:
@pdmack I'm so sorry. I force pushed my branch because I had typo...

pdmack:
It's fine. Test is running. Stay tuned.

pdmack:
/approve
/lgtm

Thanks @everpeace !

jiezhang:
@everpeace @pdmack The worker reaches "running" state once init.sh starts running, but at that point the worker is not ready yet - it still has a bunch of initialization to finish.

everpeace:
@jiezhang 

> - it still has a bunch of initialization to finish.

ah, you're write.  Then, master can wait until `ssh` to all workers succeeds.  What do you think? I'm very happy to fix it.

PR #698, opened 2018-04-20
openmpi: namespaced resource names should be prefixed with component name

everpeace:
openmpi package currently creates __fixed named__ pods which are `openmpi-master`, `openmpi-worker-n`, and `openmpi-assets`.  This prevents users from creating multiple openmpi components in one namespace.  

I think prefixing component name to namespaced resource names would be preferred.

```
$ COMPONENT=c1
$ ks generate openmpi ${COMPONENT} --image ${IMAGE} --secret ${SECRET} --workers ${WORKERS} --gpus ${GPUS} --exec "${EXEC}"
$ COMPONENT=c2
$ ks generate openmpi ${COMPONENT} --image ${IMAGE} --secret ${SECRET} --workers ${WORKERS} --gpus ${GPUS} --exec "${EXEC}"
$ ks apply default
...

$ k get po,svc,cm --show-all 
NAME             READY     STATUS    RESTARTS   AGE
po/c1-master     1/1       Running   0          12s
po/c1-worker-0   1/1       Running   0          11s
po/c1-worker-1   1/1       Running   0          11s
po/c2-master     1/1       Running   0          12s
po/c2-worker-0   1/1       Running   0          12s
po/c2-worker-1   1/1       Running   0          11s
NAME      TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)     AGE
svc/c1    ClusterIP   None         <none>        12345/TCP   13s
svc/c2    ClusterIP   None         <none>        12345/TCP   13s
NAME           DATA      AGE
cm/c1-assets   5         13s
cm/c2-assets   5         13s
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/698)

### DIFF
--- a/kubeflow/openmpi/assets.libsonnet
+++ b/kubeflow/openmpi/assets.libsonnet
@@ -3,11 +3,13 @@
     $.configMap(params),
   ],
 
+  name(params):: "%s-assets" % params.name,
+
   configMap(params):: {
     kind: "ConfigMap",
     apiVersion: "v1",
     metadata: {
-      name: "openmpi-assets",
+      name: $.name(params),
       namespace: params.namespace,
       labels: {
         app: params.name,
@@ -25,7 +27,7 @@
   genHostfile(params)::
     std.lines(
       std.map(
-        function(index) "openmpi-worker-%(index)d.%(name)s.%(namespace)s%(slots)s" % {
+        function(index) "%(name)s-worker-%(index)d.%(name)s.%(namespace)s%(slots)s" % {
           index: index,
           name: params.name,
           namespace: params.namespace,
--- a/kubeflow/openmpi/service.libsonnet
+++ b/kubeflow/openmpi/service.libsonnet
@@ -3,11 +3,13 @@
     $.service(params),
   ],
 
+  name(params):: params.name,
+
   service(params):: {
     kind: "Service",
     apiVersion: "v1",
     metadata: {
-      name: params.name,
+      name: $.name(params),
       namespace: params.namespace,
       labels: {
         app: params.name,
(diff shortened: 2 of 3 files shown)


pdmack:
@everpeace Going forward, I'll probably need to see sign-off from both you and @jiezhang. This is an area I have no particular insight into, but am happy to add "ok to test" at least.

jiezhang:
Why not simply use namespace to scope the resources?

everpeace:
It is because that there is a case which some users are not allowed to create namespace freely by cluster administrators.  as you might know that kubernetes has RBAC.  and Namespace is also [the unit of resource quota in kubernetes](https://kubernetes.io/docs/concepts/policy/resource-quotas/).  I think it's a natural way to restrict users creating namespaces (assign some namespace to them and set quotas) when multiple teams share a cluster with the fixed number of nodes.

everpeace:
@pdmack Thank you for the comment.  I understood.

jiezhang:
Per the documentation https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/ it sounds like namespace is the standard way to separate different users. And it supports resource quota. 

We plan to run this package in an external service so that end user doesn’t need to have any k8s setup or permission. So scoping each user/job in its own namespace is feasible for Uber’s use case. 

@jlewi What’s your recommendation?

jlewi:
In general we want to be able to run multiple jobs in a namespace. So we shouldn't need to create a new namespace for each job.

Can multiple jobs share the same openmpi resources? If not then we should probably bring up a distinct set of resources for each job by making the name unique.

I will send you invites to the org so you can run tests.

jlewi:
The approach in this PR. Of giving each resource a unique name based on the name of the job is consistent with what we do for other jobs (e.g. TFJobs).

So I think this is the correct think to do.

/lgtm
/approve

/hold @jiezhang 

/ok-to-test

@jiezhang When you are statisfied you can do "/hold cancel"

/assign @jiezhang

PR #701, opened 2018-04-20
openmpi: master should wait until worker pods are really ready.

everpeace:
As @jiezhang pointed out [#696](https://github.com/kubeflow/kubeflow/pull/696#issuecomment-383176969), once worker reaches "Running" state, bunch of initilization remains on workers and of cource sshd might not ready on workers.  So, user's command executed on master can fail. So, master should wait until it can 'ssh' to all workers.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/701)


everpeace:
ok, I'm closing this.

PR #702, opened 2018-04-20
Add jiezhang to OWNERS

jiezhang:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/702)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -9,6 +9,7 @@ approvers:
   - pdmack
   - ScorpioCPH
   - willb
+  - jiezhang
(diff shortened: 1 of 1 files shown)


jlewi:
IIUC you are mostly working on the openmpi/Horvod packages. So it would be better to create OWNERs files in those directories and yourself there.

As described [here](https://github.com/kubeflow/community/blob/master/CONTRIBUTING.md#maintaining-owners-files)

We want to grow the number of owners files. This helps us target code reviews to the right people.

PR #704, opened 2018-04-21
[openmpi] Introduce a sidecar container for inter-pod synchronization

jiezhang:
* openmpi-controller monitors the master pod's status and creates a semaphore file "term.sig" to signal openmpi-job to terminate
* openmpi-job is now decoupled from kubernetes
* openmpi-controller and openmpi-job shares a volume for inter-container communication
* openmpi-controller can be extended in the future to support data snapshot

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/704)

### DIFF
--- a/components/openmpi-controller/.gitignore
+++ b/components/openmpi-controller/.gitignore
@@ -1,0 +1,2 @@
+.openmpi-controller
+env
--- a/components/openmpi-controller/Dockerfile
+++ b/components/openmpi-controller/Dockerfile
@@ -1,0 +1,10 @@
+FROM python:3.6
+
+USER root
+
+ENV HOME /root
+
+ADD requirements.txt $HOME
+ADD controller $HOME/controller
+
+RUN pip3 install -r $HOME/requirements.txt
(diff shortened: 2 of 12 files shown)


everpeace:
@jiezhang Thank you for the solution 👍   And my apologies to have introduced insufficient syncs...  Don't get me wrong. No offence was meant 🙇 .    I just wanted to propose how to modify this package by opening PRs. 

`openmpi-controller` now depends on [python's kubernetes client version `6`](https://github.com/kubernetes-client/python).  Does this work on old versions of kubernetes??  I'm not sure on kubeflow's policy but, If it SHOULD, how could we test it ?

everpeace:
> I think it’s hard to maintain init.sh in the long term without making changes to the docker image

I agree with it.

>  I’m planning to add more features to it, e.g. backing up the logs and trained model to persistent storage. It’s much easier to implement more advanced features using python.

That's nice.  I would be very happy if I could contribute to it.  Is there any space for it?

> And it should be compatible with older versions of k8s according to the documentation: 

I misunderstood the meaning of '+' mark on the doc.  right, It should work.

I'm not one of a reviewer,  but, this PR looks good to me 😀

pdmack:
@jiezhang @everpeace thumbs up? We good?

Issue #709, opened 2018-04-23
cert-manager in "setting-up-iap-on-gke" not working


kunmingg:
Following instructions from
https://github.com/kubeflow/kubeflow/blob/master/docs/gke/iap.md.
Getting errors from cert-manager.

cert-manager pod log:
```
I0423 22:39:18.864010       1 server.go:68] Listening on http://0.0.0.0:9402
I0423 22:39:18.872922       1 leaderelection.go:174] attempting to acquire leader lease...
I0423 22:39:34.272045       1 leaderelection.go:184] successfully acquired lease kube-system/cert-manager-controller
I0423 22:39:34.380059       1 controller.go:136] issuers controller: syncing item 'kubeflow/letsencrypt-prod'
I0423 22:39:34.380148       1 controller.go:187] certificates controller: syncing item 'kubeflow/envoy-ingress-tls'
I0423 22:39:34.380465       1 sync.go:81] Issuer letsencrypt-prod not ready
E0423 22:39:34.380556       1 controller.go:196] certificates controller: Re-queuing item "kubeflow/envoy-ingress-tls" due to error processing: Issuer letsencrypt-prod not ready
I0423 22:39:34.385821       1 controller.go:187] certificates controller: syncing item 'kubeflow/envoy-ingress-tls'
I0423 22:39:34.386008       1 sync.go:81] Issuer letsencrypt-prod not ready
E0423 22:39:34.386085       1 controller.go:196] certificates controller: Re-queuing item "kubeflow/envoy-ingress-tls" due to error processing: Issuer letsencrypt-prod not ready
I0423 22:39:34.396352       1 controller.go:187] certificates controller: syncing item 'kubeflow/envoy-ingress-tls'
I0423 22:39:34.396473       1 sync.go:81] Issuer letsencrypt-prod not ready
E0423 22:39:34.396553       1 controller.go:196] certificates controller: Re-queuing item "kubeflow/envoy-ingress-tls" due to error processing: Issuer letsencrypt-prod not ready
I0423 22:39:34.423698       1 controller.go:187] certificates controller: syncing item 'kubeflow/envoy-ingress-tls'
I0423 22:39:34.423811       1 sync.go:81] Issuer letsencrypt-prod not ready
E0423 22:39:34.423853       1 controller.go:196] certificates controller: Re-queuing item "kubeflow/envoy-ingress-tls" due to error processing: Issuer letsencrypt-prod not ready
[...]

PR #714, opened 2018-04-24
controller bug fix

kunmingg:
fix https://github.com/kubeflow/kubeflow/issues/709

Original problem: 
1. Service ```cloud-endpoints-controller``` will always created with default namespace, which is not accessible by other resources when user specify their namespace.
2. When generate cert-manager, user should provide namespace as well, to make sure ```issuer``` is accessible by other resources.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/714)


ankushagarwal:
The changes look good, but can you add more description to the PR explaining what the problem was and how this patch fixes it.

kunmingg:
@ankushagarwal 
comment updated.

PR #743, opened 2018-04-30
Make TFJob operator a standalone prototype

### LINKED ISSUE #669: Create individual prototypes for just TFJob
We should consider creating a prototype for people who just want the TFJob controller.

This is trivial to do.

pdmack:
Closes #669 

PTAL
/cc @jlewi 
/cc @DjangoPeng 

Wasn't quite sure what to do with `testing/workflows` so left that more or less intact for now.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/743)


pdmack:
```shell
ks init my-tf-job
cd my-tf-job/
ks registry add kubeflow github.com/pdmack/kubeflow/tree/669-tfjob-prototype/kubeflow
ks pkg install kubeflow/tf-job@669-tfjob-prototype
ks generate tf-job-operator tf-job-operator
ks apply default -c tf-job-operator
```

kubeflow/core/all.libsonnet
```diff
@@ -3,13 +3,11 @@
...
     local jupyterhub = import "kubeflow/core/jupyterhub.libsonnet",
     local nfs = import "kubeflow/core/nfs.libsonnet",
-    local tfjob = import "kubeflow/core/tf-job.libsonnet",
     local spartakus = import "kubeflow/core/spartakus.libsonnet",
     local centraldashboard = import "kubeflow/core/centraldashboard.libsonnet",
     local version = import "kubeflow/core/version.libsonnet",
 
     all:: jupyterhub.all(params)
-          + tfjob.all(params)
           + ambassador.all(params)
```
jlewi: I think the all prototype should continue to install the TFJob operator. This way folks who want all the components can install them with a single package. But we can also have a prototype for just TFJob.

kubeflow/tf-job/prototypes/tf-job-operator.jsonnet
```diff
@@ -0,0 +1,26 @@
+// @apiVersion 0.1
```
jlewi: Do you think moving TFJob operator into the TFJob package makes sense? I'm uncertain what the proper organization of packages should be (see #106).

Its convenient for users not to install additional packages. Also, I think we'd like the all prototype to still install the TFJob operator. ksonnet doesn't handle dependencies between packages very well.

I'm not sure it makes sense to have a package just for TFJob. It might make sense to have a package for TensorFlow and put all the components for TensorFlow into that package.

What do you think?

pdmack: Ah, I see. Seems I was a little overzealous in breaking out the TFJob operator. ;-)
I do still think it's important to distinguish it from the "user" TFJob prototype.
So, I would prefer to name it one of:
* tf-job-operator
* tf-job-admin
* tf-job-service
* etc.

jlewi: tf-job-operator seems like a good name to me for the prototype.

jlewi:
How's this coming?

pdmack:
@jlewi next up after `/home/jovyan`...back in earnest probably tomorrow.

jlewi:
Thanks!

/lgtm
/approve

Issue #740, opened 2018-04-30
E2E test for Argo Package


jlewi:
Related to #21.

We need an E2E test. An E2E test should verify

- We can deploy the CRD
- We can submit a simple workflow that does hello world.

ankushagarwal:
Our current GKE e2e test uses the test cluster to test things out. I tried running the argo e2e test there but it broke things because it conflicted with the existing argo deployment on the test cluster. 

We should use a new cluster to run our e2e GKE tests - this will ensure that changes due to individual tests don't affect the test cluster.

PR #748, opened 2018-05-01
Add Argo Test

ankushagarwal:
Fixes https://github.com/kubeflow/kubeflow/issues/740

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/748)


ankushagarwal:
Looks like this PR is breaking CI since it deploys argo which conflicts with the existing argo

PR #764, opened 2018-05-07
[openmpi] make RBAC be optional

everpeace:
## What is the problem?
openmpi package _always_ creates a `ServiceAccount` and its `ClusterRoleBinding`.

However, end users who run this in batch mode sometimes are NOT allowed to create them by cluster administrator because the permission, granting `create` `ServcieAccount` and `ClusterRoleBinding`, is too strong.  Particularly, granting `create` for `ClusterRoleBinding` resources to some users nearly equal to give cluster admin rights to them. 

_NOTE: Actually openmpi prototype can work correctly if its service account can at least `get` for `pods` resources.  However I would like to handle this issue with another PR._

## How the PR fixes the problem

- introduce `serviceAccountName` paramter to openmpi prototype.  Users can specify service account name which would be provisioned by cluster administrators.  its default value is `"null"` (similar to other params).
- if it is not set ( `== "null"`), `ServiceAccount` and its `ClusterRoleBinding` will be create automatically.
- `serviceaccount.libsonnet` was introduced to define `ServiceAccount` and its `ClusterRoleBinding` because RBAC related entities is not related to `Service`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/764)

### DIFF
--- a/kubeflow/openmpi/all.libsonnet
+++ b/kubeflow/openmpi/all.libsonnet
@@ -1,5 +1,6 @@
 local assets = import "kubeflow/openmpi/assets.libsonnet";
 local service = import "kubeflow/openmpi/service.libsonnet";
+local serviceaccount = import "kubeflow/openmpi/serviceaccount.libsonnet";
 local workloads = import "kubeflow/openmpi/workloads.libsonnet";
 
 {
@@ -17,6 +18,7 @@ local workloads = import "kubeflow/openmpi/workloads.libsonnet";
     all::
       assets.all(updatedParams)
       + service.all(updatedParams)
+      + serviceaccount.all(updatedParams)
       + workloads.all(updatedParams),
   },
 }
--- a/kubeflow/openmpi/prototypes/openmpi.jsonnet
+++ b/kubeflow/openmpi/prototypes/openmpi.jsonnet
@@ -13,6 +13,7 @@
 // @optionalParam gpu number 0 Number of GPUs per worker.
 // @optionalParam cpu string null CPU limits per worker.
 // @optionalParam memory string null Memory limits per worker.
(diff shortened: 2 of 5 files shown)


everpeace:
Thank you reviewers!!

@jiezhang 

> What do you mean here? We're already using the "view" cluster role for the service account.

Thank you for responding my comment! Yes, we already create "view" clusterrolebinding.  I just wanted to mean "view" cluster role can be narrower.  current openmpi controller requires only to `get` `foo-master` pod.  So, I think below role and rolebinding would be minimal.   In that case we can use `RoleBinding` instead of `ClusterRoleBinding`.  I think `RoleBining` would be nice in security perspective because `RoleBinding` is namespaced resource (`ClusterRoleBinding` is cluster wide resource).  It would be easy to suppose that normal namespace admin(not cluster admin) might not be allowed to operate cluster wide resources.

```
{
  kind: "Role",
  apiVersion: "rbac.authorization.k8s.io/v1",
  metadata: {
    name: $.name(params),
    namespace: params.namespace,
  },
  // this role only allows to 'get' 'master' pod
  rules: [
    {
       apiGroups: [""],
       resources: ["pods"],
       resourceNames: [ workloads.masterName(params) ],
       verbs: ["get"],
    },
  ]
},
{
  // created service account can only 'get' 'master' pod
  kind: "RoleBinding",
  apiVersion: "rbac.authorization.k8s.io/v1",
  metadata: {
    name: $.name(params),
    namespace: params.namespace,
  },
  subjects:[
    {
      kind: "ServiceAccount"
      name: $.name(params),
      namespace: params.namespace,
    },
  ],
  roleRef: {
    kind: "Role",
    name: $.name(params),
  },
}
```

jiezhang:
@everpeace It's okay to make such a change for now, but I am concerned that this might be too restrictive and we might have to change it again when we make new API calls in the openmpi-controller.

everpeace:
@jiezhang Yeah, I understand what you mean.  That makes sense.  So, I think I won't make a PR for it.

PR #765, opened 2018-05-07
Support building different version of TF Serving images

lluunn:
- Dockerfiles are parameterized, and parameters are stored in `versions/` folder.
- can build different TF versions of TF serving image: 1.4, 1.6, 1.7, 1.6gpu
- 1.7gpu has problem needs investigatation
  `Cuda Configuration Error: cuDNN version detected from /usr/lib/x86_64-linux-gnu/include/cudnn.h (7.1.2) does not match TF_CUDNN_VERSION (7.1.2.21)`
- We can use the same pattern for other components (e.g. notebook images) so that they all share the 
  same python script to build images.
- We can then use this for releasing.

Previous discussion below:
I think we should have the parameters saving in a file for each version.
It might seem unnecessary for cpu images as they currently only have one param, but I think base image and other dependencies can be different.
Gpu images will have cuda version and possibly others.
Notebook images can also use this pattern..

related #468 
WDYT?
/cc @ankushagarwal @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/765)


ankushagarwal:
This is a good start

We want to meet the following requirements for building our images:

1. should be buildable locally
1. should be buildable via an argo workflow
1. should be parametrized
1. should share code like argo workflows, build script, etc
1. should share structure so that it's easier to add new images

Here is how we can structure our components directory to achieve that :

```
kubeflow.git/
  images/
    tensorflow-jupyter-notebook/
      Dockerfile
      params.yaml
    tf-operator/
      Dockerfile
      params.yaml
    tf-serving/
      Dockerfile
      params.yaml
  build.py
  releaser/
    components/
      workflows.jsonnet
      workflows.libsonnet
    environments/
```

params.yaml would contain all the parameters required to build different versions of the image

```
builds:
  "tensorflow-serving-1.4":
    package: "http://storage.googleapis.com/tensorflow-serving-apt/pool/tensorflow-model-server/t/tensorflow-model-server/tensorflow-model-server_1.4.0_all.deb"
  "tensorflow-serving-1.5":
    package: "http://storage.googleapis.com/tensorflow-serving-apt/pool/tensorflow-model-server/t/tensorflow-model-server/tensorflow-model-server_1.5.0_all.deb"
  "tensorflow-serving-1.7":
    package: "http://storage.googleapis.com/tensorflow-serving-apt/pool/tensorflow-model-server/t/tensorflow-model-server/tensorflow-model-server_1.7.0_all.deb"
```

The `releaser/` directory can contain an argo workflow for triggering releases.

lluunn:
Agree with the requirement. 
I think a folder per version might be better because at some point different versions may have different pip and apt requirement. Then we can just put the requirement.txt in the folder.
Also, it would be rare to update the config of previous versions, so should not be hard to maintain.

components/k8s-model-server/images/build_tf_serving_cpu.sh
```diff
@@ -0,0 +1,22 @@
+# Script to build TF Serving image.
```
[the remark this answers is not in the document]
lluunn: this will be removed

lluunn:
TF1.7 with GPU not quite works with currnet Dockerfile.gpu
```
INFO|2018-05-09T11:09:27|build_image.py|39| ERROR: /root/.cache/bazel/_bazel_root/01a289b7faaf5ec651fb0e4e35f862a1/external/org_tensorflow/tensorflow/tools/def_file_filter/def_file_filter_configure.bzl:22:1: file '@bazel_tools//tools/cpp:windows_cc_configure.bzl' does not contain symbol 'find_vc_path' (did you mean '_find_vc_path'?).
INFO|2018-05-09T11:09:27|build_image.py|39| ERROR: /root/.cache/bazel/_bazel_root/01a289b7faaf5ec651fb0e4e35f862a1/external/org_tensorflow/tensorflow/tools/def_file_filter/def_file_filter_configure.bzl:23:1: file '@bazel_tools//tools/cpp:windows_cc_configure.bzl' does not contain symbol 'find_msvc_tool' (did you mean '_find_msvc_tool'?).
INFO|2018-05-09T11:09:27|build_image.py|39| ERROR: Skipping 'tensorflow_serving/model_servers:tensorflow_model_server': error loading package '': Extension file 'tensorflow/tools/def_file_filter/def_file_filter_configure.bzl' has errors.
```
@pdmack any idea?

pdmack:
@lluunn Which version of bazel? Seen some issues building TF with 0.12 although the errors look different from this.

lluunn:
0.5.4 as in [here](https://github.com/kubeflow/kubeflow/blob/master/components/k8s-model-server/images/Dockerfile.gpu#L47)
I randomly tried different versions of bazel and got different errors...

lluunn:
Trying cuda9, cudnn7, now the error is 
```
Cuda Configuration Error: cuDNN version detected from /usr/lib/x86_64-linux-gnu/include/cudnn.h (7.1.2) does not match TF_CUDNN_VERSION (7.1.2.21)
```

https://github.com/tensorflow/tensorflow/issues/6446

pdmack:
@lluunn PTAL
https://gist.github.com/pdmack/807cb7494fdfde6aac10631fa483b55e#file-dockerfile-gpu

```
# docker run --rm hub.docker.com/r/kubeflow/tf-model-server-gpu:1.7 tensorflow_model_server --help | head -4
usage: tensorflow_model_server
Flags:
	--port=8500                      	int32	port to listen on
	--enable_batching=false          	bool	enable batching
```

components/build-image.md
```diff
@@ -0,0 +1,18 @@
+# Building images
```
ankushagarwal: Maybe just call this README.md so that it's rendered in github when browsing this directory?

components/build_image.py
```diff
@@ -0,0 +1,118 @@
+import argparse
```
ankushagarwal: Add a docstring with example invocation?

components/build_image.py
```diff
@@ -0,0 +1,118 @@
...
+    logging.info("Running: Environment:\n%s", "\n".join(lines))
+
+  process = subprocess.Popen(
+    command, cwd=cwd, env=env, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+
+  logging.info("Subprocess output:\n")
+  output = []
+  while process.poll() is None:
+    process.stdout.flush()
+    for line in iter(process.stdout.readline, ''):
```
ankushagarwal: `for line in iter(process.stdout.readline, b'')` for py3 compatiblity?

components/build_image.py
```diff
@@ -0,0 +1,118 @@
...
+import os
+import subprocess
+import sys
+import time
+import yaml
+
+
+def run(command, cwd=None, env=None, polling_interval=datetime.timedelta(seconds=1)):
+  """Run a subprocess.
+  Copied from kubeflow/test so it's easier to run locally.
```
ankushagarwal: Hopefully in the future we will be able to organize our codebase better so that we don't have to duplicate code like this.

components/build_image.py
```diff
@@ -0,0 +1,118 @@
...
+  if args.platform == "cpu":
+    command = [
+        "docker", "build", "--pull",
+        "--build-arg",  "TF_SERVING_PACKAGE={}".format(config['package']),
+        "-t", "{}/tensorflow-serving-{}:{}".format(args.registry, version, args.tag),
+        "-f", "Dockerfile.cpu", "."]
+  else:  # gpu
+    command = [
+        "docker", "build", "--pull",
+        "--build-arg", "BASE_IMAGE={}".format(config['base_image']),
```
ankushagarwal: Would it be possible to automatically build-args from `version-config.json` instead of having to manually specify them here?

For example if `version-config.json` contains 

```
{
  "base_image": "nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04",
  "tf_version": "1.7.0",
}
```

we can automatically add 2 `--build-args` for docker build:
```
--build-arg BASE_IMAGE=nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04
--build-arg TF_VERSION=1.7.0
```

lluunn:
@pdmack That works, thank you very much!
@ankushagarwal comments addressed, PTAL thanks!

pdmack:
No need/interest in 1.5 I take it?

lluunn:
It should be easy to add 1.5 later.

Issue #564, opened 2018-04-03
TFJob prototype should contain the full TFJob spec so that ks generate is mostly just copying the prototype


jlewi:
As with TFServing we should support using service account credentials to authenticate to the datastore.

One way to do this would be to modify the ksonnet components.

But it might also be interesting to look at using [admission controllers](https://kubernetes.io/docs/admin/extensible-admission-controllers/) to modify the manifest.

PR #768, opened 2018-05-07
Create an examples package in kubeflow

ankushagarwal:
examples package is intended to contain self-contained prototypes for
deploying tf-job, tf-serving, etc.

The idea is that users should be able to fully customize and extend
these prototypes by modifying the jsonnet

Fixes https://github.com/kubeflow/kubeflow/issues/564

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/768)

### DIFF
--- a/kubeflow/examples/prototypes/tf-serving-with-istio.jsonnet
+++ b/kubeflow/examples/prototypes/tf-serving-with-istio.jsonnet
@@ -102,6 +102,9 @@ local deployment = {
         labels: {
           app: appName,
         },
+        annotations: {
+          "sidecar.istio.io/inject": "true",
+        },
       },
       spec: {
         containers: [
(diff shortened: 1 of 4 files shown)


kubeflow/examples/prototypes/tf-serving-with-istio.jsonnet
```diff
@@ -0,0 +1,176 @@
...
+  metadata: {
+    labels: {
+      app: appName,
+    },
+    name: appName,
+    namespace: namespace,
+  },
+  spec: {
+    template: {
+      metadata: {
```
lluunn: Need this annotation
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/tf-serving/tf-serving.libsonnet#L172

ankushagarwal: Done

PR #777, opened 2018-05-08
Argo workflow e2e test

ankushagarwal:
Fixes https://github.com/kubeflow/kubeflow/issues/740

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/777)

### DIFF
--- a/testing/workflows/components/workflows.libsonnet
+++ b/testing/workflows/components/workflows.libsonnet
@@ -162,7 +162,7 @@
             {
               name: "e2e",
               dag: {
-                tasks: [
+                tasks: std.prune([
                   {
                     name: "checkout",
                     template: "checkout",
@@ -211,12 +211,23 @@
                           "",
                     ],
                   },
-
                   {
                     name: "pytorchjob-deploy",
                     template: "pytorchjob-deploy",
                     dependencies: ["deploy-kubeflow"],
                   },
+                  // Don't run argo test for gke since
+                  // it runs in the same cluster as the
+                  // test cluster. For minikube, we have
+                  // a separate cluster.
+                  if platform == "minikube" then
+                    {
+                      name: "test-argo-deploy",
+                      template: "test-argo-deploy",
+                      dependencies: ["deploy-kubeflow"],
+                    }
+                  else
+                    {},
                   {
                     name: "tfjob-test",
                     template: "tfjob-test",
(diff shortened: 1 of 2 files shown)


testing/workflows/components/workflows.libsonnet
```diff
@@ -162,7 +162,7 @@
             {
               name: "e2e",
               dag: {
-                tasks: [
+                tasks: std.prune([
```
lluunn: why do you need to prune here?

testing/workflows/components/workflows.libsonnet
```diff
@@ -211,12 +211,23 @@
...
                   },
-
                   {
                     name: "pytorchjob-deploy",
                     template: "pytorchjob-deploy",
                     dependencies: ["deploy-kubeflow"],
                   },
+                  // Don't run argo test for gke since
+                  // it runs in the same cluster as the
+                  // test cluster. For minikube, we have
```
lluunn: Isn't it deployed in its own namespace, so it's okay to be in the same gke cluster?

ankushagarwal: Having multiple controllers for argo in the same cluster causes race conditions for argo workflows - we had a downtime in the test infrastructure because of this last week.

testing/workflows/components/workflows.libsonnet
ankushagarwal: For gke, the argo workflow step will be `{}`. prune removes that entry.

lluunn: I thought we just need to prune here https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/workflows.jsonnet#L13?

testing/workflows/components/workflows.libsonnet
lluunn: I see

testing/workflows/components/workflows.libsonnet
ankushagarwal: I would prefer explicitly pruning the task list rather than relying on pruning at a higher level. I believe pruning at a higher level is not a good practice because sometimes there are parts of the manifest which require an explicit empty object, eg : https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/cert-manager.libsonnet#L7

PR #791, opened 2018-05-10
Python script to build TF notebook images

lluunn:
- share the same python script to build TF notebook and TF serving (and can extend to other components)
- add gnupg to Dockerfile, otherwise got `E: gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation`

Still verifying all versions work.
/hold

/cc @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/791)

### DIFF
--- a/components/README.md
+++ b/components/README.md
@@ -5,8 +5,8 @@ This allows us to use `build_image.py` to build all images.
 
 For example
 ```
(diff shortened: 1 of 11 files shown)


ankushagarwal:
/lgtm
/approve

Thanks!

PR #797, opened 2018-05-12
[openmpi] Download and upload data in the controller

### LINKED ISSUE #713: [openmpi] Upload trained models to persistent storage
We should take snapshots of the trained models generated by the MPI workers and upload them to persistent storage (say S3 or Google Cloud Storage).

Here're the changes needed:
1. `kubeflow-openmpi-data` is currently mounted as a shared volume between the `openmpi-job` container and `openmpi-controller` container. We need to make the size configurable.
1. When `openmpi-controller` detects that the master pod has terminated, it should upload the files in the shared mount point to persistent storage.
1. Make the persistent storage configurable.

jiezhang:
Fix #713

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/797)

### DIFF
--- a/components/openmpi-controller/controller/controller.py
+++ b/components/openmpi-controller/controller/controller.py
@@ -1,8 +1,9 @@
+import os
 from pathlib import Path
 
 from kubernetes import client, config
 from kubernetes.config.config_exception import ConfigException
-from util import log, api_retry, long_poll
+from util import log, api_retry, long_poll, s3_copy
 
 SIG_DIR = '.openmpi-controller'
 SIGCONT = f'{SIG_DIR}/SIGCONT'
@@ -19,11 +20,17 @@ class Controller:
     It communicates with the master pod using kubernetes API.
     """
 
-    def __init__(self, namespace, master, num_gpus, timeout_secs):
+    def __init__(self, namespace, master, num_gpus, timeout_secs,
+                 download_data_from, download_data_to, upload_data_from, upload_data_to):
         self.namespace = namespace
         self.master = master
         self.num_gpus = num_gpus
         self.timeout_secs = timeout_secs
+        self.download_data_from = download_data_from
+        self.download_data_to = download_data_to
+        self.upload_data_from = upload_data_from
+        self.upload_data_to = upload_data_to
+        self._validate_args()
         Path(SIG_DIR).mkdir()
 
     def __enter__(self):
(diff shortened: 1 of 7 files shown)


kubeflow/openmpi/prototypes/openmpi.jsonnet
```diff
@@ -17,9 +17,14 @@
...
 // @optionalParam schedulerName string default-scheduler scheduler name to use for the components.
-// @optionalParam controllerImage string jiez/openmpi-controller:0.0.1 Docker image of the openmpi-controller.
+// @optionalParam controllerImage string jiez/openmpi-controller:0.0.2 Docker image of the openmpi-controller.
 // @optionalParam initTimeout number 300 Timeout in seconds to abort the initialization.
 // @optionalParam nodeSelector string null Comma-delimited list of "key=value" pairs to select the worker nodes. e.g. "cloud.google.com/gke-accelerator=nvidia-tesla-k80"
+// @optionalParam s3Secret string null Name of secret containing s3 credentials (AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY).
+// @optionalParam downloadDataFrom string null URI where data are downloaded from. Only S3 bucket is supported at the moment.
+// @optionalParam downloadDataTo string null Local path where data are downloaded to.
+// @optionalParam uploadDataFrom string null Local path where data are uploaded from.
+// @optionalParam uploadDataTo string null URI where data are uploaded to. Only S3 bucket is supported at the moment.
```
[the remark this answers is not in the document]
jiezhang: Done.

components/openmpi-controller/controller/controller.py
```diff
@@ -43,10 +50,20 @@ def __exit__(self, exc_type, exc_val, exc_tb):
     def wait_ready(self):
         if self.num_gpus > 0:
             self._wait_nvidia_driver_present()
+        self._download_data()
         Path(SIGCONT).touch()
 
     def wait_done(self):
         self._wait_master_terminated()
+        self._upload_data()
```
alsrgv: I think we should make this method and `download_data()` public and call it explicitly in `main.py`.  Maybe rename to `upload_data_if_needed()`, `download_data_if_needed()`

jiezhang: Currently wait_ready creates SIGCONT file in wait_ready. If we have to call download_data_if_needed in main, then it must be called before wait_ready, which seems to expose too much information to to caller.

Later, we may want to do "s3 sync" periodically instead of uploading data at the end. I think it makes more sense to keep all the data sync logic within the controller.

alsrgv: OK, that sounds reasonable.

everpeace:
@jiezhang sorry that I couldn't review this pr in time.  LGTM!

PR #799, opened 2018-05-14
seldon load model files from PVC

ogre0403:
This PR is related to #733 . 

As for now, `seldon/seldon-serve-simple` prototype only read model files in docker image. 
Actually, seldon-serve can read env variables to find model file in any place. 
In this PR, add one more params: `pvcName`. If `pvcName` is not "null" or "", a PVC will be created by default storage class, and is mounted in `/mnt`. 

User can put their model file in PV, and specify the model location  in predictions file (eg. [IssueSummarization.py](https://github.com/kubeflow/examples/blob/master/github_issue_summarization/notebooks/IssueSummarization.py#L20))  or in env variable.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/799)

### DIFF
--- a/kubeflow/seldon/prototypes/serve-simple.jsonnet
+++ b/kubeflow/seldon/prototypes/serve-simple.jsonnet
@@ -24,13 +24,17 @@ local image = import "param://image";
 local namespace = updatedParams.namespace;
 local replicas = import "param://replicas";
 local endpoint = import "param://endpoint";
-local pvcname = import "param://pvcName";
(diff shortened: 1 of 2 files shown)


pdmack:
@ogre0403 could you run [this](https://github.com/kubeflow/kubeflow/blob/master/scripts/autoformat_jsonnet.sh) on your files please?

pdmack:
/approve
/lgtm

per @cliveseldon

PR #800, opened 2018-05-14
add meta api for serving

u2takey:
Add meta data api for model metadata, which is useful for checking model info.
Related rpc service is defined here: https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/prediction_service.proto#L29

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/800)

### DIFF
--- a/components/k8s-model-server/http-proxy/server.py
+++ b/components/k8s-model-server/http-proxy/server.py
@@ -19,6 +19,7 @@ from __future__ import print_function
 from itertools import repeat
 import base64
 import logging
+import grpc
 
 from google.protobuf.json_format import MessageToDict
 from grpc.beta import implementations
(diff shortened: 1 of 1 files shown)


components/k8s-model-server/http-proxy/server.py
```diff
@@ -192,6 +193,17 @@ def get_signature(signature_map, signature_name=None):
     raise KeyError("No signature found for signature key %s." % signature_name)
 
 
+class MetaHandler(tornado.web.RequestHandler):
+  """
+  Meta Hanlder proxy return Model metadata. Defined here https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/prediction_service.proto#L29
```
pdmack: "Handler"

pdmack:
Hmmm, tests are failing but your branch is _way_ out of date. As a sanity check, @u2takey can you rebase with upstream master please?

jlewi:
Please update the PR description as well and explain what a MetaAPI is for?

/assign @lluunn

components/k8s-model-server/http-proxy/server.py
```diff
@@ -192,9 +193,20 @@ def get_signature(signature_map, signature_name=None):
     raise KeyError("No signature found for signature key %s." % signature_name)
 
 
+class MetaHandler(tornado.web.RequestHandler):
+  """
+  Meta Handler proxy return Model metadata. Defined here https://github.com/tensorflow/serving/blob/master/tensorflow_serving/apis/prediction_service.proto#L29
```
lluunn: Currently it only supports signature map. We should add a comment here.

Issue #742, opened 2018-04-30
E2E test for bootstrapper


jlewi:
We should have an E2E test for the bootstrapper.

We could modify/fork our existing E2E test to use the bootstrapper.

/assign @kunmingg

kunmingg:
@jlewi 
Added https://github.com/kubeflow/kubeflow/issues/758 to current sprint.
How about use e2e to cover this case https://github.com/kubeflow/kubeflow/issues/758 as first step?

kunmingg:
@jlewi Will create PR today.

PR #803, opened 2018-05-14
add bootstrapper to e2e test

kunmingg:
Change e2e test on gke: now we use bootstrapper to setup test env & resources.
Each e2e test will build bootstrapper image from PR src so e2e test now takes more time (~5min) to run.
Good news is we cover bootstrapper logic in e2e test, and it should be more stable than env-setup scripts.

Cover https://github.com/kubeflow/kubeflow/issues/742.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/803)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/options/options.go
+++ b/bootstrap/cmd/bootstrap/app/options/options.go
@@ -30,6 +30,7 @@ type ServerOption struct {
 	Project       string
 	Email         string
 	IpName        string
+	RegistryUri	  string
 }
 
 // NewServerOption creates a new CMServer with a default config.
@@ -50,4 +51,5 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&s.Email, "email", "", "Your Email address for GCP account, if you are using GKE.")
 	fs.StringVar(&s.IpName, "ip-name", "kubeflow", "Name of the ip you reserved on GCP project")
 	fs.BoolVar(&s.InCluster, "in-cluster", false, "Whether bootstrapper is executed inside a pod")
+	fs.StringVar(&s.RegistryUri, "registry-uri", "/opt/kubeflow/kubeflow", "Location of kubeflow registry.")
 }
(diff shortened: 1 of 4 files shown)


bootstrap/cmd/bootstrap/app/options/options.go
```diff
@@ -50,4 +52,6 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 	fs.StringVar(&s.Email, "email", "", "Your Email address for GCP account, if you are using GKE.")
 	fs.StringVar(&s.IpName, "ip-name", "kubeflow", "Name of the ip you reserved on GCP project")
 	fs.BoolVar(&s.InCluster, "in-cluster", false, "Whether bootstrapper is executed inside a pod")
+	fs.StringVar(&s.RegistryUri, "registry-uri", "/opt/kubeflow/kubeflow", "Location of kubeflow registry.")
+	fs.BoolVar(&s.Testing, "testing", false, "Whether bootstrapper is executed for e2e test.")
```
jlewi: Why do we need an option in the bootstrapper about the E2E tests?

kunmingg: When testing = true, bootstrapper will exit after setting up resources.

kunmingg:
@jlewi 
I added new param ```registry-uri```.
When run e2e test, registry-uri = /path/to/PR/commit/src.
So bootstrapper will use PR commit code as registry source.

bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -480,7 +486,7 @@ func Run(opt *options.ServerOption) error {
 			log.Infof("Components applied: " + out.String())
 		}
 	}
-	if opt.InCluster {
+	if opt.InCluster && opt.KeepAlive {
```
ankushagarwal: Shouldn't we keepAlive when `opt.KeepAlive` is True regardless of what `opt.InCluster` is set to

kunmingg: When ```opt.InCluster=false```, bootstrapper is not executed inside a pod, then we don't need to prevent pod from exit. Likely that user is using bootstrapper in interactive mode. (Local docker)

ankushagarwal:
/lgtm
/approve

/hold

Feel free to run `/hold cancel` when ready to submit

Issue #807, opened 2018-05-14
Bootstrapper should deploy new Stackdriver agents


lluunn:
#806 
https://cloud.google.com/monitoring/kubernetes-engine/

lluunn:
We don't need this as deployment manager deploys the new agents as well

PR #814, opened 2018-05-15
deployment manager uses new stackdriver agents

### LINKED ISSUE #806: deployment manager should disable legacy Stackdriver agents
- For GKE 1.10 and above
- And then bootstrapper can deploy the new agents.

https://cloud.google.com/monitoring/kubernetes-engine/

lluunn:
Fix #806  #807 

https://cloud.google.com/deployment-manager/docs/configuration/supported-resource-types

/cc @jlewi @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/814)

### DIFF
--- a/docs/gke/configs/cluster.jinja
+++ b/docs/gke/configs/cluster.jinja
@@ -48,7 +48,6 @@ resources:
         - https://www.googleapis.com/auth/devstorage.read_only
         - https://www.googleapis.com/auth/logging.write
         - https://www.googleapis.com/auth/monitoring
-        
 
 # We manage the node pools as separate resources.
 # We do this so that if we want to make changes we can delete the existing resource and then recreate it.
--- a/docs/gke/configs/cluster.jinja
+++ b/docs/gke/configs/cluster.jinja
@@ -28,21 +28,27 @@ resources:
       name: {{ CLUSTER_NAME }}
       # 
       initialNodeCount: {{ properties['initialNodeCount'] }}
+      {% if properties['stackdriver-kubernetes'] %}
       # TODO: remove alpha when 10.2 is public.
+      # https://github.com/kubeflow/kubeflow/issues/821
       enableKubernetesAlpha: true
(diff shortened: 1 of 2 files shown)


docs/gke/configs/cluster.jinja
```diff
@@ -36,6 +36,11 @@ resources:
         - https://www.googleapis.com/auth/devstorage.read_only
         - https://www.googleapis.com/auth/logging.write
         - https://www.googleapis.com/auth/monitoring
+      # Logging and monitoring have default value [logging/monitoring].googleapis.com
+      # if not set. We set it to none and will deploy the new Stackdricer Kubernetes.
```
jlewi: Stackdriver Kubernetes agents

lluunn:
/hold
We might not need this.
We can disable Stackdriver agent after cluster created, so can be done by bootstrapper.
https://cloud.google.com/kubernetes-engine/docs/how-to/logging#disabling_logging

lluunn:
/hold cancel
@jlewi How about we just deploy the new agents using deployment manager?
I think it's the better way, comparing to disabling old agents here, and deploy new ones at bootstrapper.
Also it's much easier.

docs/gke/configs/cluster.jinja
```diff
@@ -28,14 +28,21 @@ resources:
       name: {{ CLUSTER_NAME }}
       # 
       initialNodeCount: {{ properties['initialNodeCount'] }}
-      # We need 1.9 to support GPUs without requiring Alpha.
-      initialClusterVersion: 1.9.6-gke.1
+      # TODO: remove alpha when 10.2 is public.
```
jlewi: Can you file an issue an reference the issue here? Could we add an option that can be enabled via the YAMl spec to enable Alpha features or not? And then selectively enable the logging and monitoring service below?

jlewi:
Using deployment manager is better to running gcloud commands.

lluunn:
@jlewi comment addressed, PTAL

PR #817, opened 2018-05-16
[WIP] Ambassador proxy

### LINKED ISSUE #239: Use Ambassador/Envoy as proxy for JupyterHub
Right now we deploy JupyterHub using the NodeJs proxy in JupyterHub. This means all connections to the notebook are proxied through JupyterHub.

Since we already have a reverse proxy in the form of Envoy we should be able to configure KubeSpawner to instead use Envoy as the proxy to notebooks and not rely on JupyterHub.

According to @yuvipanda JupyterHub also supports using Ingress to setup routes to the notebooks. As part of this the spawner creates Service endpoints to map to the notebooks. We might be able to modify this so that we don't create an ingress, but create services with the appropriate annotations to create reverse proxy rules through Ambassador.

kkasravi:
Fixes #239 
- [x] Replaces configurable-http-proxy with ambassador for /hub, /hub-lb, /user prefixes
- [x] A service is created dynamically with the /user prefix and includes the ambassador annotation when the jupyterhub is spawned
- [x] IAP should direct to ambassador (remove cluster_jupyterhub)
- [ ] jupyterhub spawn page should wait for service, pod and provide feedback to the user (WIP) if spawning is taking a long time (existing behavior)

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/817)


kkasravi:
status
- brings up login page
- brings up start server page
- spawn is failing - problem in kubeform_spawner.py - jupyter-<user> stalls in ContainerCreating. Currently debugging this problem.

jlewi:
nit: Please update the PR description to explain the purpose of the PR and (briefly) the changes needed.

jlewi:
The test failures look like a jsonnet formatting issue and some jsonnet tests. Do you need help with these?

Maybe the jupyterhub jsonnet test is just a matter of updating the expectation?

There's an autformatting tool that should fix the

kkasravi:
yes  i'll fix these tonight or very early in the morning. I've been working on the last item...

what remains is a 404 after clicking spawn. It should show the 'spawn_pending.html' page.
The 404 is due to the time delay it takes ambassador to notice a new service and pull in the annotations. I believe I now have the logic to wait for ambassador to recognize and add the new service annotation but the delay is adding extra time. Ambassador uses kubewatch and is single threaded due to python's limitations (GIL). I'm not sure what 3 ambassadors do to coordinate.

jlewi:
This is way more difficult than I expected; thank you so much for working through it.

kkasravi:
really close - I think the POST below (log of tf-hub-0) is redirected to the ambassador POD instead of the ambassador service which results in a 404 to the browser. But if I immediately hit refresh in the browser the hub page comes up. TL;DR I'm now waiting long enough for the annotation of the jupyter-<user> annotation to be propagated to ambassador service but I need to redirect to the ambassador service on the POST. Should be able to try this tonight. I'll keep you updated. 

```
[I 2018-05-26 00:43:39.260 JupyterHub jupyterhub_config:277] GET http://10.79.254.86:80/user/kam/ http response code is 200
[I 2018-05-26 00:43:39.265 JupyterHub log:122] 302 POST /hub/spawn → /user/kam/ (kam@10.76.1.47) 14918.55ms
```

kkasravi:
@kflynn 
Does a host_rewrite take care of 302 redirects where the location is within the kubernetes cluster?
Here is the flow
1) The tf-hub-0 service is annotated for ambassador. The prefix /hub/ routes to this service.
2) The pod under the tf-hub-0 service presents a form which will spawn a jupyter notebook
3) When the user clicks 'spawn' this results in a POST where a new service and pod is created. 
4) The new service is annotated for ambassador and we wait until the service and pod are created and ambassador recognizes the annotation for the new service which has a prefix of /user/
5) The POST in 3) above returns a 302 redirect and the location header points to the new service created.
6) The browser does the redirect but gets a 404

kflynn:
@kkasravi I'd expect that sequence to work, since you're waiting for Ambassador to have the route created. I'm not quite sure how you're looking for the the `host_rewrite` to interact with that? When your `POST` returns the `302`, it's complete, and the browser following the `302` is a completely new request that'll have to be matched in its own right.

kkasravi:
@jlewi yes - just hitting refresh in the browser brings up the jupyter notebook

@kflynn i didn't think I needed the http_rewrite but thought I might check. Prior to the 302 being returned the new service and pod pair are created. The new service has the ambassador annotation with the prefix /user/<name>/. The 302 is a redirect to that service. But it seems like the redirect should be a redirect to the ambassador service since otherwise the location header is an internal cluster IP.

jlewi:
Why is it redirecting to an internal IP? I would expect it appends /user/name/ to the prefix of /hub.

Put another way why would the redirect work in the case where we use JupyterHub as the proxy to the notebooks as opposed to using Ambassador? The URLs should always be the same; i.e
http://mydomain.com/hub
http://mydomain.com/user/name/

So I wouldn't expect kube_spawner.py to be dependent on how underlying infrastructure for routing URLs. I would expect kube_spawner.py to always be redirecting users to
http://mydomain.com/user/name/
?

kkasravi:
@jlewi i reviewed [writing a custom proxy](http://jupyterhub.readthedocs.io/en/latest/reference/proxy.html) and KubeProxy.add-route just needs to be modified from how i'm doing it now which is using `self.spawner.server.host`. I also found another fix where I needed to make sure each new pod and service had the full prefix /user/name/ in the ambassador annotation other ambassador would get confused. I'll get to this today hopefully.

kkasravi:
this is replaced by #915

PR #851, opened 2018-05-22
ksonnet changes to support deploying the v1alpha2 TFJob operator.

jlewi:
* K8s doesn't support installing multiple versions of a CRD in the same cluster.
* So we add an option to choose which version to support.
* We don't add an E2E test in this PR because its not straightforward.
  Currently the E2E test deploys Kubeflow in a different namespace but not
  a different cluster. This won't allow us to simulatenously test both
  versions.

* So to test multiple versions we will need to spin up a separate GKE cluster.

Related to: kubeflow/tf-operator#599

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/851)


jlewi:
Test failure looks like some random networking issue

```
Cloning into '/mnt/test-data-volume/kubeflow-presubmit-kubeflow-e2e-minikube-851-b67da5b-1642-e8ae/src/kubeflow/tf-operator'...
fatal: The remote end hung up unexpectedly
fatal: early EOF
fatal: index-pack failed
```

Issue #829, opened 2018-05-18
Bootstrapper should optionally use config map to specify ksonnet parameters


jlewi:
I propose the bootstrapper optionally take a YAML file that looks like the following

```
app:
  component:
     - name: (of component)
       prototype: Name of prototype to install
  packages:
     - name: (Of package to install)
  parameters:
     - component: name of component
       name: (name of parameter)
       value: (value of parameter)
```

The idea of passing parameters to the bootstrapper using a config map was first suggested by @inc0 . There were a couple questions about how to do this in particular

    1. Not creating yet another templating / config layer around ksonnet
    1. Exposing all parameters consistently and not selectively

It might be nice if this functionality was incorporated directly into `ks` so that one could do something like

```
ks init my-app --config=myconfig.yaml
```

**Motivation:**

We'd like the bootstrapper to be useable by Kubernetes deployment tools. Examples that are already integrating with Kubeflow

- Canonical Jujuj
- GCP Deployment Manager

These upstream tools can be responsible for

1. Provisioning the K8s cluster and non K8s resources
1. Creating K8s resources needed to run the bootstrapper (e.g. cluster role bindings)
1. Triggering the bootstrapper

The advantage of this approach is that all K8s manifests (even vendor specific ones) can be managed consistently as part of the Kubeflow ksonnet app. vendors can add packages, prototypes, and/ parameters specific to their distribution and then just enable them via the config map.

/priority p1
/cc @bryanl @inc0 @kunmingg @carmine @marcoceppi

PR #853, opened 2018-05-23
Use yaml config to manage k8s resources deployed by bootstrapper

kunmingg:
This provide us flexibility to control kubeflow resources when facing different code version / env. 
solve https://github.com/kubeflow/kubeflow/issues/829

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/853)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/options/options.go
+++ b/bootstrap/cmd/bootstrap/app/options/options.go
@@ -26,7 +26,7 @@ type ServerOption struct {
 	InCluster     bool
 	KeepAlive     bool
 	AppDir        string
-	ConfigDir	  string
+	Config        string
 	Email         string
 	NameSpace     string
 	RegistryUri   string
@@ -49,5 +49,5 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&s.InCluster, "in-cluster", false, "Whether bootstrapper is executed inside a pod")
 	fs.StringVar(&s.RegistryUri, "registry-uri", "/opt/kubeflow/kubeflow", "Location of kubeflow registry.")
 	fs.BoolVar(&s.KeepAlive, "keep-alive", true, "Whether bootstrapper will stay alive after setup resources.")
-	fs.StringVar(&s.ConfigDir, "config", "/opt/kubeflow/default.yaml", "Path to bootstrapper components config.")
+	fs.StringVar(&s.Config, "config", "/opt/kubeflow/default.yaml", "Path to bootstrapper components config.")
 }
--- a/bootstrap/cmd/bootstrap/app/options/options.go
+++ b/bootstrap/cmd/bootstrap/app/options/options.go
@@ -26,11 +26,9 @@ type ServerOption struct {
 	InCluster     bool
 	KeepAlive     bool
 	AppDir        string
-	KfVersion     string
-	NameSpace     string
-	Project       string
+	ConfigDir	  string
 	Email         string
-	IpName        string
+	NameSpace     string
 	RegistryUri   string
 }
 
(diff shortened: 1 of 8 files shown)


bootstrap/cmd/bootstrap/app/options/options.go
```diff
@@ -26,11 +26,9 @@ type ServerOption struct {
 	InCluster     bool
 	KeepAlive     bool
 	AppDir        string
-	KfVersion     string
-	NameSpace     string
-	Project       string
+	ConfigDir	  string
```
jlewi: Why a directory and not a single file?

bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -205,7 +249,7 @@ func setupNamespace(namespaces type_v1.NamespaceInterface, name_space string) er
 
 func createComponent(opt *options.ServerOption, kfApp *kApp.App, fs *afero.Fs, args []string) {
 	componentName := args[1]
-	componentPath := filepath.Join(opt.AppDir, "components", componentName+".jsonnet")
+	componentPath := filepath.Join(opt.AppDir, "Components", componentName+".jsonnet")
```
jlewi: Why is Components capitalized?

bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -221,6 +265,58 @@ func createComponent(opt *options.ServerOption, kfApp *kApp.App, fs *afero.Fs, a
...
 
+func appGenerate(opt *options.ServerOption, kfApp *kApp.App, fs *afero.Fs, appConfig *AppConfig) {
+	libs, err := (*kfApp).Libraries()
+
+	if err != nil {
+		log.Fatalf("Could not list libraries for app; error %v", err)
+	}
+	// Install packages.
+	for _, p := range appConfig.Packages {
+		//for _, p := range []string{"kubeflow/core", "kubeflow/tf-serving", "kubeflow/tf-job", "kubeflow/pytorch-job"} {
```
jlewi: Delete commented out code.

bootstrap/config/gcp_prototype.yaml
```diff
@@ -0,0 +1,44 @@
+# Sample config for kubeflow bootstrapper
+---
+# Apps that always apply
+DefaultApp:
```
jlewi: Why do the key names start with capital letters? This doesn't follow standard K8s or GCP deployment manager syntax. Can we make it lower case?

bootstrap/config/gcp_prototype.yaml
```diff
@@ -0,0 +1,44 @@
+# Sample config for kubeflow bootstrapper
+---
+# Apps that always apply
```
jlewi: Why are we creating multiple apps? Why wouldn't we just create a single app and use the ConfigMap to specify the app we want to create?

bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -45,19 +45,63 @@ import (
...
+}
+
+//
+type AppConfig struct {
+	Packages   []KsPackage
+	Components []KsComponent
+	Parameters []KsParameter
+}
+
+type BootConfig struct {
```
jlewi: Why are there multiple apps? And if you want to support multiple apps why wouldn't you use a container like a list or map.

My sense though is we should start with a single app unless we have a clear use case for multiple ones.

I think if a user wanted multiple apps they could run bootstrapper multiple times

bootstrap/cmd/bootstrap/app/server.go
kunmingg: goland's auto rename...

jlewi:
Awesome; thanks for the quick fix.

/lgtm
/approve

Issue #802, opened 2018-05-14
On GCP trigger bootstrapper with deployment manager


jlewi:
Background: #241 Recommended Kubeflow setup on GCP.

On GCP we'd like to recommend using deployment manager to provision GKE clusters and setup Kubeflow.

As part of that we'd like to trigger the bootstrapper from deployment manager.

jlewi:
I have a prototype of creating a K8s resource after the cluster here
https://github.com/jlewi/kubeflow/blob/gke_deploy_bootstrapper/docs/gke/configs/cluster.jinja#L156

It looks like deployment manager doesn't support  dependsOn to reference a template type. I believe a template type is something like cluster.jinja. But within a template type (cluster.jinja) we can depend on the type provider.

So I think the net net is in the short term we need to have a single cluster.jinja as opposed to splitting it into cluster.jinja and bootstrapper.jinja

jlewi:
Fixed.

PR #859, opened 2018-05-23
Use a config map to configure bootstrapper.

jlewi:
* Use deployment manager to create a config map containing a YAML
  spec that will be used to configure the ksonnet app created by the
  bootstrapper.

Related to #802 Trigger bootstrapper with deployment manager

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/859)

### DIFF
--- a/docs/gke/configs/cluster.jinja
+++ b/docs/gke/configs/cluster.jinja
@@ -378,6 +378,7 @@ e.g. creating namespaces, service accounts, stateful set to run the bootstrapper
               - /opt/kubeflow/bootstrapper
               - --in-cluster
               - --namespace=kubeflow
+              - --config=/etc/kubeflow/config.yaml
             env:
             - name: NAMESPACE
               value: "kubeflow"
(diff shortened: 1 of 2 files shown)


docs/gke/configs/cluster.jinja
```diff
@@ -352,8 +374,10 @@ e.g. creating namespaces, service accounts, stateful set to run the bootstrapper
           - name: kubeflow-bootstrapper
             image: {{ properties["bootstrapperImage"] }}
             workingDir: /opt/bootstrap
-            command: [ "/opt/kubeflow/bootstrapper"]
-            args: ["--in-cluster", "--namespace=kubeflow"]
+            command: 
+              - /opt/kubeflow/bootstrapper
+              - --in-cluster
+              - --namespace=kubeflow
```
kunmingg: Bootstrapper need config location: 
--config=/etc/kubeflow/config.yaml

jlewi:
Thanks @kunmingg PTAL

PR #868, opened 2018-05-24
Update to Ambassador 0.34.0

kflynn:
Update the Ambassador version to 0.34.0.


ankushagarwal:
/ok-to-test

@kflynn : Can you squash your changes into a single commit?

PR #872, opened 2018-05-25
deployment manager and prototype bug fix

kunmingg:
1. Add kubeflow-core to config, and fix typos.
2. Make bootstrapper image use master branch as registry, till we reelase 0.2.
3. Make bootstrapper use params from config when create components. (Some params are not optional.)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/872)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/server.go
+++ b/bootstrap/cmd/bootstrap/app/server.go
@@ -295,11 +295,24 @@ func appGenerate(opt *options.ServerOption, kfApp *kApp.App, fs *afero.Fs, appCo
 		}
 	}
 
-	// Create Components
-	for _, c := range appConfig.Components {
-		createComponent(opt, kfApp, fs, []string{c.Prototype, c.Name})
+	paramMapping := make(map[string][]string)
+	// Extract params for each components
+	for _, p := range appConfig.Parameters {
+		if val, ok := paramMapping[p.Component]; ok {
+			paramMapping[p.Component] = append(val, []string{"--" + p.Name, p.Value}...)
+		} else {
+			paramMapping[p.Component] = []string{"--" + p.Name, p.Value}
+		}
 	}
 
+	// Create Components
+	for _, c := range appConfig.Components {
+		params := []string{c.Prototype, c.Name}
+		if val, ok := paramMapping[c.Name]; ok {
+			params = append(params, val...)
+		}
+		createComponent(opt, kfApp, fs, params)
+	}
 	// Apply Params
 	for _, p := range appConfig.Parameters {
 		err = actions.RunParamSet(map[string]interface{}{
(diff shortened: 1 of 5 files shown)


docs/gke/configs/env-kubeflow.sh
```diff
@@ -11,14 +11,14 @@ export PROJECT=kubeflow
 
 # The name of the ip address as defined in cluster.jinja
 # We will reserve a GCP IP address to use for ingress to the cluster.
-export IP_NAME=static-ip
+# export IP_NAME=static-ip
```
jlewi: Just delete the commented out values.

jlewi:
The PR description could be a little more descriptive.

bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -482,7 +493,7 @@ func Run(opt *options.ServerOption) error {
 		// if use k8s client-go API, would be quite verbose if we create all resources one by one.
 		// TODO: use API to create ks Components
 		log.Infof("Apply kubeflow Components...")
-		rawCmd := "ks show default | kubectl apply -f -"
+		rawCmd := "ks show default | kubectl apply --validate=false -f -"
```
jlewi: Why are you setting validate to false? This seems dangerous. Where are we supplying invalid config?

docs/gke/configs/env-kubeflow.sh
```diff
@@ -9,25 +9,9 @@
...
-# We will reserve a GCP IP address to use for ingress to the cluster.
-export IP_NAME=static-ip
-
-# Set the namespace to the name of the namespace to deploy Kubeflow in.
-export NAMESPACE=kubeflow
-
 # Set config file to the YAML file defining your deployment manager configs.
 export CONFIG_FILE=cluster-${PROJECT}.yaml
-export PROJECT_CONFIG_FILE=project-${PROJECT}.yaml
+# export PROJECT_CONFIG_FILE=project-${PROJECT}.yaml
```
jlewi: Delete the commented out code.

bootstrap/cmd/bootstrap/app/server.go
kunmingg: Without validate=false, kubectl was complaining about not able to find CRD definitions. Skipping validate is a shortcut, let me see if there's other ways to apply.

jlewi: Setting validate to false is highly undesirable. Not sure why that would be necessary 
Also why can't we just do

ks apply? Why do we need to pipe it to kubectl apply?

kunmingg: Found the cause of validate bug:
https://github.com/kubeflow/kubeflow/pull/889

ksonnet doesn't support apply from pod but kubectl does; we can remove pipe to kubectl once ksonnet add this feature.

kunmingg:
@ankushagarwal rebased, thanks.

PR #873, opened 2018-05-25
Katib ksonnet package

lluunn:
Katib ksonnet components, following what are deploying in
https://github.com/kubeflow/katib/blob/master/scripts/deploy.sh

Fix https://github.com/kubeflow/katib/issues/32

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/873)

### DIFF
--- a/kubeflow/katib/modeldb.libsonnet
+++ b/kubeflow/katib/modeldb.libsonnet
@@ -26,226 +26,226 @@
         type: "ClusterIP",
         ports: [
           {
-	    port: 6543,
+            port: 6543,
             protocol: "TCP",
             name: "api",
-	  },
-	],
+          },
+        ],
         selector: {
           app: "modeldb",
           component: "backend",
-	}
-      }
+        },
+      },
     },  // backendService
     backendDeployment: {
-      "apiVersion": "extensions/v1beta1",
-      "kind": "Deployment",
-      "metadata": {
-        "labels": {
-          "app": "modeldb",
-          "component": "backend"
+      apiVersion: "extensions/v1beta1",
+      kind: "Deployment",
+      metadata: {
+        labels: {
+          app: "modeldb",
+          component: "backend",
         },
-        "name": "modeldb-backend",
-        "namespace": namespace
+        name: "modeldb-backend",
+        namespace: namespace,
       },
-      "spec": {
-        "replicas": 1,
-        "template": {
-          "metadata": {
-            "labels": {
-              "app": "modeldb",
-              "component": "backend"
+      spec: {
+        replicas: 1,
+        template: {
+          metadata: {
+            labels: {
+              app: "modeldb",
+              component: "backend",
             },
-            "name": "modeldb-backend"
+            name: "modeldb-backend",
           },
-          "spec": {
-            "containers": [
+          spec: {
+            containers: [
               {
-                "args": [
-                  "modeldb-db"
+                args: [
+                  "modeldb-db",
                 ],
-                "image": "mitdbg/modeldb-backend:latest",
-                "name": "modeldb-backend",
-                "ports": [
+                image: "mitdbg/modeldb-backend:latest",
+                name: "modeldb-backend",
+                ports: [
                   {
-                    "containerPort": 6543,
-                    "name": "api"
-                  }
-                ]
-              }
-            ]
-          }
-        }
-      }
+                    containerPort: 6543,
+                    name: "api",
+                  },
+                ],
+              },
+            ],
+          },
+        },
+      },
     },  // backendDeployment
 
     dbService: {
-      "apiVersion": "v1",
-      "kind": "Service",
-      "metadata": {
-        "labels": {
-          "app": "modeldb",
-          "component": "db"
+      apiVersion: "v1",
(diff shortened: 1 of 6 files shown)


lluunn:
@gaocegege @YujiOshima @jlewi PTAL

jlewi:
@lluunn Sorry for the delay. Just noticed ithis because it popped up in my dashboard finally. If you want me to review something /assign me. @ mentions will get buried in my inbox.

kubeflow/katib/modeldb.libsonnet
```diff
@@ -0,0 +1,252 @@
...
+                  },
+                ],
+              },
+            ],
+          },
+        },
+      },
+    },  // dbDeployment
+
+    frontendService: {
```
jlewi: Either in this PR or a follow on one we should add an Ambassador prefix to access the front end.

kubeflow/katib/modeldb.libsonnet
```diff
@@ -0,0 +1,252 @@
...
+                args: [
+                  "modeldb-backend",
+                ],
+                env: [
+                  {
+                    name: "ROOT_PATH",
+                    value: "/katib",
+                  },
+                ],
+                image: "katib/katib-frontend",
```
jlewi: Docker images should be exposed by parameters to make it easy to try new images.

kubeflow/katib/modeldb.libsonnet
```diff
@@ -0,0 +1,252 @@
...
+                imagePullPolicy: "Always",
```
jlewi: Default imagePullPolicy should be IfNotPresent. Always is not good practice; we want to use the local cache.

kubeflow/katib/modeldb.libsonnet
```diff
@@ -0,0 +1,252 @@
...
+                  },
+                ],
+              },
+            ],
+          },
+        },
+      },
+    },  // frontendDeployment
+
+    frontendIngress: {
```
jlewi: Lets get rid of the ingress. Our networking story is that services are exposed via the Ambassador reverse proxy. I don't want to start creating ingress per service until we better understand the motivation for doing so.

kubeflow/katib/prototypes/all.jsonnet
```diff
@@ -0,0 +1,23 @@
...
+
+local k = import "k.libsonnet";
+
+local vizier = import "kubeflow/katib/vizier.libsonnet";
+local modeldb = import "kubeflow/katib/modeldb.libsonnet";
+local suggestion = import "kubeflow/katib/suggestion.libsonnet";
+
+// updatedParams uses the environment namespace if
+// the namespace parameter is not explicitly set
+local updatedParams = params {
```
jlewi: You do not need this stanza. env should always contain environment because we are now require versions of ksonnet that ensure its true.

kubeflow/katib/modeldb.libsonnet
```diff
@@ -0,0 +1,252 @@
...
+            },
+            name: "modeldb-backend",
+          },
+          spec: {
+            containers: [
+              {
+                args: [
+                  "modeldb-db",
+                ],
+                image: "mitdbg/modeldb-backend:latest",
```
jlewi: parameterize the image please

kubeflow/katib/vizier.libsonnet
```diff
@@ -0,0 +1,265 @@
...
+          spec: {
+            containers: [
+              {
+                args: [
+                  "-w",
+                  "dlk",
+                  "-i",
+                  "k-cluster.example.net",
+                ],
+                image: "katib/vizier-core",
```
jlewi: parameterize the image please

kubeflow/katib/prototypes/all.jsonnet
lluunn: I don't get it. Do you mean we should always use namespace from env? 
I think the current logic is correct?

kubeflow/katib/modeldb.libsonnet
lluunn: I plan to do that in next PR. We have a different issue https://github.com/kubeflow/katib/issues/88

kubeflow/katib/prototypes/all.jsonnet
jlewi: Correct you can just get namespace from env.
The logic is correct but unnecessary. 

In earlier versions of ksonnet env wasn't available. So the way to get namespace was to make it available via a params.

A later version of ksonnet introduced env which contains namespace. But for backwards compatibility we still wanted to respect namespace in params. But we no longer need to do that.

lluunn: But then all components in one environment will have to have the same namespace?
Shouldn't we allow setting namespace per component?

jlewi: Perhaps in the future. There's an implicit assumption that everything is already in the same namespace. So I wouldn't expect setting different namespaces would just work.

lluunn: I see

PR #879, opened 2018-05-25
Add GKE Security Features to Deployment Manager config

ankushagarwal:
Add option for
* Private Clusters
* Separate Service Account
* masterIpv4CidrBlock
* Pod Security Policy
* Secure Node Metadata
* Authorized Networks for k8s Master
* Reduce oauth scopes for worker nodes and replace it with gke-default

/cc @jlewi @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/879)

### DIFF
--- a/docs/gke/configs/cluster-kubeflow.yaml
+++ b/docs/gke/configs/cluster-kubeflow.yaml
@@ -21,13 +21,15 @@ resources:
   # 1. Use a single template (.jinja file for all resources) or
   # 2. Create two separate deployments and launch the boot strapper
   # after the cluster is created.
-  # 
-  # Two separate deployments doesn't make much sense; we could just use 
+  #
+  # Two separate deployments doesn't make much sense; we could just use
   # kubectl at that point. So we put all resources in a single deployment.
 - name: kubeflow
   type: cluster.jinja
   properties:
     zone: us-east1-d
+    # Set this to v1beta1 to use beta features such as private clusters,
+    gkeApiVersion: v1
     # An arbitrary string appending to name of nodepools
     # bump this if you want to modify the node pools.
     # This will cause existing node pools to be deleted and new ones to be created.
@@ -35,14 +37,34 @@ resources:
     pool-version: v1
     # Two is small enough to fit within default quota.
     cpu-pool-initialNodeCount: 2
-    gpu-pool-initialNodeCount: 0    
+    gpu-pool-initialNodeCount: 0
     # Whether to deploy the new Stackdriver Kubernetes agents
     stackdriver-kubernetes: false
+    securityConfig:
+      # Whether to use a cluster with private IPs
+      # Use v1beta1 api
+      privatecluster: false
+      # masterIpv4CidrBlock for private clusters, if enabled
+      # Use v1beta1 api
+      masterIpv4CidrBlock: 172.16.0.16/28
+      # Protect worker node metadata from pods
+      # Use v1beta1 api
+      secureNodeMetadata: false
+      # Whether to enable Pod Security Policy Admission Controller
+      # Use v1beta1 api
+      podSecurityPolicy: false
+      masterAuthorizedNetworksConfig:
+          cidrBlocks:
+          - cidrBlock: 1.2.3.4/32
+          - cidrBlock: 5.6.7.8/32
+          enabled: false
     # Path for the bootstrapper image.
     bootstrapperImage: gcr.io/kubeflow-images-public/bootstrapper:v20180525-v0.1.1-71-ga9cdd5dd-e3b0c4
     # This is the name of the GCP static ip address to reserve for your domain.
     # This must be different for each Kubeflow deployment in your project.
     ipName: kubeflow-ip
+    # Name of the service account to use for k8s worker node pools
+    vmServiceAccountName: kubeflow-service-account
     # Provide the config for the bootstrapper. This should be a string
     # containing the YAML spec for the bootstrapper.
     #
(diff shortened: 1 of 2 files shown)


docs/gke/configs/cluster-kubeflow.yaml
```diff
@@ -21,33 +21,54 @@ resources:
...
-  # Two separate deployments doesn't make much sense; we could just use 
+  #
+  # Two separate deployments doesn't make much sense; we could just use
   # kubectl at that point. So we put all resources in a single deployment.
 - name: kubeflow
   type: cluster.jinja
   properties:
     zone: us-east1-d
+    # Set this to v1beta1 to use beta features such as private clusters,
+    apiVersion: v1
```
kunmingg: v1beta1 or v1?

ankushagarwal: I want to set a default value of `v1`. If the end-user wants to leverage beta features, they can set this to v1beta1

docs/gke/configs/cluster.jinja
```diff
@@ -72,59 +83,85 @@ resources:
...
+      ipAllocationPolicy:
+        createSubnetwork: true
+        useIpAliases: true
+      masterIpv4CidrBlock: {{ properties['masterIpv4CidrBlock'] }}
+      privateCluster: true
+      masterAuthorizedNetworksConfig: {{ properties['masterAuthorizedNetworksConfig'] }}
+      {% endif %}
       nodeConfig:
         machineType: n1-standard-1
+        serviceAccount: {{ properties['serviceAccountName'] }}@{{ env['project'] }}.iam.gserviceaccount.com
```
jlewi: Do we want to use the service account as the VM service account? I think a better practice is to move in the direction of creating secrets containing service account keys. This way we can provide finer grained permissions then what we get from node service accounts.

docs/gke/configs/cluster-kubeflow.yaml
jlewi: Is this the GKE API version? Should we call it gkeApiVersion?

docs/gke/configs/cluster.jinja
ankushagarwal: This service account will not be used by pods. This service account is intended to replace the default compute engine service account which has Project Editor Role. We will be using secrets and service account keys to give access to individual pods.

docs/gke/configs/cluster-kubeflow.yaml
ankushagarwal: Yes. Updated

docs/gke/configs/cluster.jinja
```diff
@@ -72,59 +83,85 @@ resources:
       monitoringService: monitoring.googleapis.com/kubernetes
       {% else %}
       initialClusterVersion: 1.9.6-gke.1
-      {% endif %}      
+      {% endif %}
+      podSecurityPolicyConfig:
```
jlewi: Should we make all the security parameters fields in a dictionary "securityFeatures" just to make the properties a little bit more human readable?

jlewi:
Looks like the test failed because you need the IAM API; you just need to add

```
- name: iam-api
  action: 'gcp-types/servicemanagement-v1:servicemanagement.services.enable'
  properties: 
    consumerId: {{ 'project:' + env['project'] }}
    serviceName: iam.googleapis.com
```

and then make sure the patch step depends on that step.

jlewi:
Same error

```
RROR: (gcloud.deployment-manager.deployments.create) Error in Operation [operation-1527296670560-56d117783ef01-9cc0d413-1104737f]: errors:
- code: RESOURCE_ERROR
 location: /deployments/z79-6e2d4ba-1689-1b9d/resources/kubeflow-service-account
 message: '{"ResourceType":"iam.v1.serviceAccount","ResourceErrorCode":"403","ResourceErrorMessage":{"code":403,"message":"Identity
   and Access Management (IAM) API has not been used in project 593963025935 before
   or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/iam.googleapis.com/overview?project=593963025935
   then retry. If you enabled this API recently, wait a few minutes for the action
   to propagate to our systems and retry.","status":"PERMISSION_DENIED","details":[{"@type":"type.googleapis.com/google.rpc.Help","links":[{"description":
"Google
   developers console API activation","url":"https://console.developers.google.com/apis/api/iam.googleapis.com/overview?project=593963025935"}]}],"statusM
essage":"Forbidden","requestPath":"https://iam.googleapis.com/v1/projects/kubeflow-ci/serviceAccounts","httpMethod":"POST"}}'
```

/test all

docs/gke/configs/cluster.jinja
```diff
@@ -108,7 +108,12 @@ resources:
           nodeMetadata: SECURE
         {% endif %}
         oauthScopes:
-        - gke-default
+        - https://www.googleapis.com/auth/devstorage.read_only
+        - https://www.googleapis.com/auth/logging.write
+        - https://www.googleapis.com/auth/monitoring
+        - https://www.googleapis.com/auth/service.management.readonly
+        - https://www.googleapis.com/auth/servicecontrol
```
jlewi: logging and monitoring make sense for the logging and monitoring agents but what about the others?

I assume devstorage is for GCR?

Why do we need service.management, servicecontrol, and trace?

ankushagarwal: I substituted `gke-default` [1] with the actual scopes. I will reduce the scopes.

[1] https://cloud.google.com/sdk/gcloud/reference/container/clusters/create

ankushagarwal:
@jlewi This is ready for review

PR #881, opened 2018-05-25
Update kubeform_spawner.py

pdmack:
Now that this is vendor-ed elsewhere (e.g., kubeflow/testing) we need to guard against pylint failures. A different issue should address pylint corrections.

### DIFF
--- a/kubeflow/core/kubeform_spawner.py
+++ b/kubeflow/core/kubeform_spawner.py
@@ -1,3 +1,4 @@
+#pylint: skip-file
 import json
 import os
 from kubespawner.spawner import KubeSpawner


jlewi:
Would it be better to configure our lint checks to ignore vendor directories?

This is the script we use for lint 
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/test_py_lint.py#L38

Could we just update that script so we don't recurse into vendor directories?

pdmack:
And yet...
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/test_py_lint.py#L41

In my tests, I also experienced difficulties trying to get the rc file to ignore the testing `vendor` dir.

pdmack:
Prefer option 2 I think. I'll have a look.

Issue #852, opened 2018-05-22
E2E test for TFJob v1alpha2 ksonnet package


jlewi:
Context: 
   * kubeflow/tf-operator#599 add ksonnet package for v1alpha2 of TFJob operator
   * #836 - Verify Kubeflow deployed correctly with deployment manager
   * #829 - Use a config map with bootstrapper to override packages installed

We need to update our E2E tests to test the TFJob operator v1alpha2 package.

The difficulty here is that we can't deploy two versions of a CRD in the same cluster. So we will need to modify our tests to spin up a new GKE cluster for each version of the operator.

My suggestion would be to switch to using deployment manager + bootstrapper. 

We have this [workflow](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/gke_deploy.libsonnet) which uses deployment manager to create a GKE cluster.

When #829 is fixed and the tests are updated to support this I think we can easily run the test for multiple versions of TFJob operator.

/priority p1

ankushagarwal:
Python script from #866 to deploy kubeflow clusters

ankushagarwal:
Looks like bootstrapper has a couple of bugs. Once https://github.com/kubeflow/kubeflow/pull/872/files gets merged I'll start work on this

PR #894, opened 2018-05-30
Use Deployment Manager to bring up clusters for running E2E tests

ankushagarwal:
* Use Deployment Manager to bring up clusters for running E2E tests
  * Running e2e tests in a separate cluster will provide better isolation
  * It will be easier to deploy multiple clusters for complex tests which involve multiple versions of CRDs
* Delete gke_deploy workflow as it is no longer needed
* Use deploy_kubeflow_gcp.py and teardown_kubeflow_gcp.py to manage the deployments
* Update workflows.libsonnet with the new workflow

Related to https://github.com/kubeflow/kubeflow/issues/852

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/894)


testing/dm_configs/cluster-kubeflow.yaml
```diff
@@ -23,15 +23,24 @@ resources:
     # Path for the bootstrapper image.
     bootstrapperImage: gcr.io/kubeflow-images-public/bootstrapper:latest
     bootstrapperConfig: |
+      registries:
+        # For e2e test target repo we skip registry repo and branch
+        # Instead we use PR code attached to container by volumn.
```
jlewi: volumn -> volume

testing/dm_configs/cluster-kubeflow.yaml
```diff
@@ -23,15 +23,24 @@ resources:
     # Path for the bootstrapper image.
     bootstrapperImage: gcr.io/kubeflow-images-public/bootstrapper:latest
     bootstrapperConfig: |
+      registries:
+        # For e2e test target repo we skip registry repo and branch
```
jlewi: Why can't we just specify the registry as file URL in the case where we want to test the code in a presubmit/postsubmit? And then for the file URL specify the location of the source code on the volume.

ankushagarwal: This comment is incorrect. I have updated this. In the workflow we build the bootstrap image and we bake the source code from the PR into the bootstrap image. The source code is obtained from the checked out git repo in the shared NFS.

The build_image step is [here](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/workflows.libsonnet#L444). The build_image.sh script is [here](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/build_image.sh)

testing/dm_configs/cluster-kubeflow.yaml
ankushagarwal: Updated

PR #904, opened 2018-06-01
Improvements and bug fixes in DM config.

### LINKED ISSUE #878: Deployment manager config should create service accounts and set IAM roles
Our deployment manager config should create service accounts and set IAM roles for those service accounts appropriately.

We should create two service accounts

1. For admin tasks like enabling endpoints 
1. For use with TFJobs etc...
   * Should have IAM roles like read/write to various storage products
   * Dataflow access; etc...

jlewi:
* Create a service account to be used to authorize TFJobs and other work
  within the cluster.

* Create a helper script to download service account keys and turn
  them into K8s keys

* Fix some bugs in the docs.

* Fix #878 create a GCP service account for the user.

* IAP script needs a GCP service account with network admin privileges.
* Add network admin privileges to the admin service account.
* Name the secrets in K8s so that be default the names are the same across
  the deployments. This way there's one less parameter to set for
  every deployment.

* VM service account should have a unique name per deployment so deployments
  are isolated.

* Need to grant the VM service account logs and monitoring access to support
  monitoring.

* I don't think there's any reason to allow user to specify name of the
  VM service account in the YAML file right now.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/904)

### DIFF
--- a/docs/gke/configs/cluster.jinja
+++ b/docs/gke/configs/cluster.jinja
@@ -41,8 +41,15 @@ limitations under the License.
 {% set STATEFULSETS_COLLECTION = '/apis/apps/v1/namespaces/{namespace}/statefulsets' %}
 {% set CLUSTER_ROLE_BINDING_COLLECTION = '/apis/rbac.authorization.k8s.io/v1/clusterrolebindings' %}
 
-{# Names for service accounts.#}
+{# Names for service accounts.
+   -admin is to be used for admin tasks
+   -user is to be used by users for actual jobs.
+   -vm is used for the VM service account attached to the GKE VMs.
+ #}
 {% set KF_ADMIN_NAME = NAME_PREFIX + '-admin' %}
+{% set KF_USER_NAME = NAME_PREFIX + '-user' %}
+{% set KF_VM_SA_NAME = NAME_PREFIX + '-vm' %}
+
 {# For most of the K8s resources we set the deletePolicy to abandon; otherwise deployment manager reports various errors.
    Since we delete the cluster all the K8s resources will be deleted anyway.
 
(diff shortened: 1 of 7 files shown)


docs/gke/create_k8s_secrets.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+#
+# A simple helper script to download secrets for Kubeflow service
+# accounts and store them as K8s secrets.
+set -ex
+export SA_EMAIL=${DEPLOYMENT_NAME}-admin@${PROJECT}.iam.gserviceaccount.com
+
+# TODO(jlewi): We should name the secrets more consistently based on the service account name.
+# We will need to update the component configs though
+gcloud --project=${PROJECT} iam service-accounts keys create ${SA_EMAIL}.json --iam-account $SA_EMAIL
```
ankushagarwal: nit: use ${SA_EMAIL}

docs/gke/configs/cluster.jinja
```diff
@@ -286,9 +300,40 @@ TODO(jlewi): Do we need to serialize API activation
           members:
             - {{ 'serviceAccount:' + env['project_number'] + '@cloudservices.gserviceaccount.com' }}
 
+        {# servicemanagement.admin is needed by CloudEndpoints controller
```
ankushagarwal: How do we manage deletion of IAM Role Bindings? When we delete the deployment, the service accounts go away but these IAM Role Bindings still lurk around.

jlewi: Good point. Created 
https://github.com/kubeflow/kubeflow/issues/910

PR #905, opened 2018-06-01
Upgrade jupyterhub and kube_spawner

### LINKED ISSUE #479: JupyterHub Spawner complains notebook is already spawning; upgrade JupyterHub to 0.9
Steps to repeat

1. Try to spawn a pod with an invalid image name
2. Pod for notebook doesn't start with error invalid image name
3. Now delete the pod for the notebook.

Try spawning a new notebook. JupyterHub complains that it is already spawning a notebook.

```
400 : Bad Request
accounts.google.com:jlewi@google.com is pending spawn
```

My suspicion is that JupyterHub is stuck waiting for the pod to timeout as part of the spawn and its not smart enough to realize it will never start.

jlewi:
* JupyterHub 0.9 is now available and 0.8 for Kubespawner.
* These updates include improvements that should help with UX when spawning
  notebooks by providing more information about progress.

* As part of this change we switch to using the JupyterHub image as our
  base image and only installing the additional modules that we care about.

* It looks like the original image was manually intalling nodejs presumably
  to support the javascript proxy inside JupyterHub. If nodejs is needed
  by jupyterhub then I expect it to be included in the official image
  so we shouldn't need to manually install it.

* I verified this manually by spinning up a new JupyterHub instance using this image
     * The spawner provides a much greater UI
     * The UI shows K8s events e.g. "notebook spawned"
     * It also has a status bar showing progress. I'm guessing the progress is a total fabrication but it will probably make users feel better.
   
Fix #479

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/905)

### DIFF
--- a/kubeflow/core/prototypes/all.jsonnet
+++ b/kubeflow/core/prototypes/all.jsonnet
@@ -11,8 +11,7 @@
 // @optionalParam tfDefaultImage string null The default image to use for TensorFlow.
 // @optionalParam tfJobUiServiceType string ClusterIP The service type for the UI.
 // @optionalParam jupyterHubServiceType string ClusterIP The service type for Jupyterhub.
-// @optionalParam jupyterHubImage string gcr.io/kubeflow/jupyterhub-k8s:v20180531-3bb991b1
- The image to use for JupyterHub.
+// @optionalParam jupyterHubImage string gcr.io/kubeflow/jupyterhub-k8s:v20180531-3bb991b1 The image to use for JupyterHub.
 // @optionalParam jupyterHubAuthenticator string null The authenticator to use
 // @optionalParam jupyterNotebookPVCMount string null Mount path for PVC. Set empty to disable PVC
 // @optionalParam reportUsage string false Whether or not to report Kubeflow usage to kubeflow.org.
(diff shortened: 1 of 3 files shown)


kubeflow/core/prototypes/all.jsonnet
```diff
@@ -11,7 +11,8 @@
 // @optionalParam tfDefaultImage string null The default image to use for TensorFlow.
 // @optionalParam tfJobUiServiceType string ClusterIP The service type for the UI.
 // @optionalParam jupyterHubServiceType string ClusterIP The service type for Jupyterhub.
-// @optionalParam jupyterHubImage string gcr.io/kubeflow/jupyterhub-k8s:1.0.1 The image to use for JupyterHub.
+// @optionalParam jupyterHubImage string gcr.io/kubeflow/jupyterhub-k8s:v20180531-3bb991b1
```
pdmack: @jlewi Think you got an unintended line split here.

jlewi: Thanks.

jlewi:
@ankushagarwal Could you take a look please? @pdmack is OOO.

/assign @ankushagarwal

PR #907, opened 2018-06-01
Install py2 to global conda env

### LINKED ISSUE #906: Regression Jupyter notebooks no longer include python2 runtime
I just tried gcr.io/kubeflow-images-public/tensorflow-1.7.0-notebook-cpu:v20180601-2464851a

And I didn't see an option for a python2 runtime which is needed for TFX/Beam support.

My guess is this was a regression introduced when refactored the notebook image to support mounting all of /home/jovyan on a PVC.

/cc @pdmack

pdmack:
Fixes #906 

PTAL
/cc @jlewi 
/cc @ankushagarwal 

Are we good with just dropping the `--user` py2 install and letting it get installed to `/opt/conda/envs`? Are there some other tests I should try?

Also, should we install the Beam pkg @jlewi was referring to in #906 while we're here?

screen shot 2018-06-01 at 11 47 20 am
screen shot 2018-06-01 at 1 01 54 pm

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/907)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -117,12 +117,12 @@ RUN conda create -n py2 python=2 && \
     tensorflow-serving-api \
     # ipykernel for python 2 jupyter notebook kernel
     && \
-    python -m ipykernel install --user && \
+    python -m ipykernel install && \
     # tensorflow-model-analysis is only supported for TF 1.6 and above
     if [[ $INSTALL_TFMA == "yes" ]]; then \
       pip install --no-cache-dir tensorflow-model-analysis && \
-      jupyter nbextension install --py --symlink tensorflow_model_analysis --user && \
-      jupyter nbextension enable --py tensorflow_model_analysis --user; \
+      jupyter nbextension install --py --symlink tensorflow_model_analysis && \
+      jupyter nbextension enable --py tensorflow_model_analysis ; \
     fi \
     && \
     # Install jupyterlab-manager


ankushagarwal:
Installing beam SGTM

ankushagarwal:
My bad. tensorflow-model-analysis seems to install apache-beam already : https://github.com/tensorflow/model-analysis/blob/master/setup.py#L189

pdmack:
Yep, I misinterpreted @jlewi's comment. Doing some more testing.

pdmack:
@ankushagarwal Do you/we have a particular build environment for the NB images when TFMA is enabled? I consistently get a SIGILL from your post-refactor build in my env:

```
/bin/bash: line 1:   283 Illegal instruction     (core dumped) jupyter nbextension install --py --symlink tensorflow_model_analysis
```

This happens even in the NB container itself if I try to manually install it.

pdmack:
OK figured out that my CPU env wasn't up to snuff. If there's no AVX support then that extension install fails because it tries to do a TF import. Will do more py2 testing on another env tomorrow.

jlewi:
Thank you!

pdmack:
I think this is OK now. @jlewi can you try this out and see if it's satisfactory?

`docker pull pdmack/tf-notebook:1.7.0`

jlewi:
Confirmed imports work in a Python 2 environment

```
import apache_beam
import tensorflow_model_analysis as tfma
```
/lgtm
/approve

PR #915, opened 2018-06-02
Ambassador to jupyterhub

### LINKED ISSUE #239: Use Ambassador/Envoy as proxy for JupyterHub

kkasravi:
Fixes #239 

I abandoned spawning any new service by using kubespawner's proxy (as [suggested](https://github.com/kubeflow/kubeflow/issues/239#issuecomment-365421864))  and went with registering tf-hub-lb with the prefixes /hub/ and /user/ with ambassador. This will route to tf-hub-0 which will use configurable-http-proxy to redirect to user's notebooks (eg /user/kam, /user/jeremy, /user/...). Now all notebooks will be behind ambassador. 

There is a SPOF with tf-hub-0 since deleting it will also delete it's local sql db it uses to track users and jupyter-<username> pods. I'll a separate issue regarding this and I think storing these mappings as a secret that can be fetched will allow recovery. Note that this would be a issue in kubespawner not kubeflow 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/915)


jlewi:
Thanks for your hard work; and especially for documenting why this is non-trivial.

/lgtm
/approve

PR #918, opened 2018-06-03
Run v1alpha1 and v1alpha2 TfJob tests in GKE workflow

ankushagarwal:
* Update deploy_kubeflow_gcp to add tfjob_version flag
* In GKE Testing workflow, bring up two k8s clusters - one to test v1alpha1 and other to test v1alpha2
* Deploy kubeflow on both clusters using bootstrapper - set tfjob_version different for both clusters
* Using different kubeConfig for both clusters
* Teardown both clusters at the end

Fixes https://github.com/kubeflow/kubeflow/issues/852

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/918)

### DIFF
--- a/testing/deploy_kubeflow_gcp.py
+++ b/testing/deploy_kubeflow_gcp.py
@@ -30,6 +30,10 @@ def parse_args():
     help="The bootstrapper image for deployment.")
 
   parser.add_argument(
+    "--tfjob_version", required=True, type=str,
+    help="The tfjob version to deploy.")
+
+  parser.add_argument(
     "--config", required=True, type=str, 
     help="The path to the YAML file for the deployment config to use.")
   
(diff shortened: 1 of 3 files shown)


ankushagarwal:
/cc @kunmingg 
/cc @jlewi 

This is ready for review

jlewi:
Did you consider parameterizing the workflow by TFVersion and then submitting two different workflows? One for v1alpha1 and v1alpha2?

ankushagarwal:
Yes. That wasn't a clean approach either. A couple of problems in that approach:
1. Tests would be repeated - pytorchjob tests, jsonnet tests - needed to add logic to skip tests
2. The workflow copies all the junit-xml files to a common artifacts directory at the end of the test. Running multiple copies of the test would cause the artifacts to be overridden - we would have to add logic to create files with different names for each of the workflow steps

PR #919, opened 2018-06-03
fix envoy config volumeMount

jramos:
Envoy fails to start due to `/etc/envoy/envoy-config.json` not being mounted properly. This corrects the volumeMount name.

### DIFF
--- a/kubeflow/core/iap.libsonnet
+++ b/kubeflow/core/iap.libsonnet
@@ -151,7 +151,7 @@
       volumeMounts: [
         {
           mountPath: "/etc/envoy",
-          name: "shared",
+          name: "config-volume",
         },
       ],
     },  // envoyContainer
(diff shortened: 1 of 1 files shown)


pdmack:
It appears correct in our dev.kubeflow.org env. That volume is shared between the envoy and iap containers but at different mount points.

```
# cat /var/shared/healthz.env
JWT_AUDIENCE=/projects/235037502967/global/backendServices/3766062064519193928
NODE_PORT=31556
BACKEND_ID=3766062064519193928
# cat /etc/envoy/healthz.env
JWT_AUDIENCE=/projects/235037502967/global/backendServices/3766062064519193928
NODE_PORT=31556
BACKEND_ID=3766062064519193928
```

What exactly are you seeing when you say envoy fails to start?

Issue #901, opened 2018-06-01
bootstrapper should not crash loop on error


jlewi:
Right now if the bootstrapper exits with an error the bootstrapper container will crashloop. Here's an error caused by an invalid config; in this case I forgot to set hostname.

```
[...]

PR #921, opened 2018-06-04
keep bootstrapper pod alive when error occurs

kunmingg:
fix https://github.com/kubeflow/kubeflow/issues/901

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/921)

### DIFF
--- a/bootstrap/cmd/bootstrap/main.go
+++ b/bootstrap/cmd/bootstrap/main.go
@@ -22,6 +22,7 @@ import (
 
 	"github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/app"
 	"github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/app/options"
+	"time"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


bootstrap/cmd/bootstrap/main.go
```diff
@@ -43,6 +44,14 @@ func main() {
 	}
 
 	if err := app.Run(s); err != nil {
-		log.Fatalf("%v\n", err)
+		if s.InCluster && s.KeepAlive {
+			log.Infof("Bootstrapper failed with error: %v\n", err)
+			log.Infof("Keeping pod alive...")
```
ankushagarwal: Please log a description of why we're doing this so that it's easier for the user to understand what's happening

bootstrap/cmd/bootstrap/main.go
```diff
@@ -43,6 +44,14 @@ func main() {
 	}
 
 	if err := app.Run(s); err != nil {
-		log.Fatalf("%v\n", err)
+		if s.InCluster && s.KeepAlive {
+			log.Infof("Bootstrapper failed with error: %v\n", err)
```
jlewi: Use log.Errorf here since you are logging an error.

jlewi:
One minor comment but otherwise looks good.

jlewi:
Looks like a test flake; filed
https://github.com/kubeflow/testing/issues/147

/test all

jlewi:
/test all

Looks like the most recent failure was a minikube error waiting for the VM to be deleted.

PR #923, opened 2018-06-04
Bootstrap image delete on workflow cleanup for e2e tests

ankushagarwal:
/cc @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/923)


ankushagarwal:
Discussed offline with @kunmingg. Looks like this only removed the tag and does not actually delete the image layers from GCS.

Issue #903, opened 2018-06-01
Deadlocks configuring envoy for IAP


jlewi:
I'm noticing deadlocks and other problems configuring envoy using the IAP script. Some of the problems I observe

1. iap.sh is never able to acquire the lock and therefore able to write the envoy-config.json
1. envoy container is crash looping - prevents GCP loadbalancer from detecting the backend is heathy

I think we should make the following changes

1. There should be a single pod responsible for enabling IAP and updating the envoy-config map as needed
   
    * We should move this out of the sidecar and into a separate deployment
    * Locking should be less important because there won't be contention

1. The envoy sidecars are now just responsible for updating envoy config based on the config map
    * They no longer need to acquire a lock
    * They can periodically check the config map and compute a hash to know when it changes

1. We should provide a default config that will allow envoy to startup but ensure non secure traffic is blocked 
    * This way we can avoid the problems with the ingress thinking the backend is unhealthy.

jlewi:
Closing this issue because #924 should fix the deadlocks and we have open issues for the other issues.

PR #924, opened 2018-06-05
Prevent deadlocks trying to setup IAP.

jlewi:
* We want a single process to update the backend service and IAP.
  Right now we do this in the envoy side car which is replicated.

* The script gets a lock to prevent multiple services from trying to update
  IAP simultaneously.

* But until the process acquires the lock it can't update the local copy
  of the envoy config to do proper JWT validation.

* So we move updating the backend into a separate deployment that has a single
  replica 1 (although we still use a lock).

* envoy sidecars now do not modify the backend but just get relevant info
  and update the envoy config.

* To support getting the ksonnet app of the bootstrapper add source repository
  IAM role to the admin service account and mount the service account
  into the bootstrapper.
     This will allow use to push the ksonnet app from the bootstrapper to
     a cloud source repository.

Related to #903

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/924)

### DIFF
--- a/kubeflow/core/setup_iap.sh
+++ b/kubeflow/core/setup_iap.sh
@@ -1,0 +1,125 @@
+#!/bin/bash
+#
+# A simple shell script to enable IAP and configure timeouts by using gcloud.
+[ -z ${CLIENT_ID} ] && echo Error CLIENT_ID must be set && exit 1
+[ -z ${CLIENT_SECRET} ] && echo Error CLIENT_SECRET must be set && exit 1
+[ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
+[ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
+
+apk add --update jq
+curl https://storage.googleapis.com/kubernetes-release/release/v1.9.4/bin/linux/amd64/kubectl > /usr/local/bin/kubectl && chmod +x /usr/local/bin/kubectl
+
+# Stagger init of replicas when acquiring lock
+sleep $(( $RANDOM % 5 + 1 ))
+
+# We acquire a lock because we want to ensure  there is a single process
+# trying to modify the backend at a time.
+kubectl get svc ${SERVICE} -o json > service.json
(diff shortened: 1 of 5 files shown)


kubeflow/core/setup_iap.sh
```diff
@@ -0,0 +1,125 @@
...
+# Stagger init of replicas when acquiring lock
+sleep $(( $RANDOM % 5 + 1 ))
+
+# We acquire a lock because we want to ensure  there is a single process
+# trying to modify the backend at a time.
+kubectl get svc ${SERVICE} -o json > service.json
+LOCK=$(jq -r ".metadata.annotations.iaplock" service.json)
+
+NOW=$(date -u +'%s')
+if [[ -z "${LOCK}" || "${LOCK}" == "null" ]]; then
```
kunmingg: why we still need lock here if replicas=1?

jlewi: Because there could still be more than 1 pod at a given time; K8s doesn't guarantee that there is a single pod all the time; it just guarantees that in steady state there will be 1 pod.

kubeflow/core/setup_iap.sh
```diff
@@ -0,0 +1,125 @@
...
+
+# For healthcheck compare.
+mkdir -p /var/shared
+echo "JWT_AUDIENCE=${JWT_AUDIENCE}" > /var/shared/healthz.env
+echo "NODE_PORT=${NODE_PORT}" >> /var/shared/healthz.env
+echo "BACKEND_ID=${BACKEND_ID}" >> /var/shared/healthz.env
+
+# TODO(jlewi): We should publish the modified envoy config as a config map
+# and use that in the envoy sidecars.
+kubectl get configmap -n ${NAMESPACE} envoy-config -o jsonpath='{.data.envoy-config\.json}' | \
```
danisla: No need to write out the envoy-config.json to the shared directory since there is no envoy sidecar in the "iap-enabler" deployment.

ankushagarwal:
/lgtm
/approve

/hold

Feel free to run `/hold cancel` when ready to submit

Issue #927, opened 2018-06-05
Friction log for bootstrapper documentation


cwbeitel:
Worked through https://github.com/kubeflow/kubeflow/blob/master/docs/gke/gke_setup.md, here's some feedback, hope it helps!

- Trying to understand exactly which sections in the YAML I'm supposed to modify feels a little tedious. Does this need to involve modifying these directly? Can there be a simple python wrapper that takes deployment parameters and templates them in the appropriate places? This can still generate a YAML that gets checked into source control. If doing so there can be additional validation e.g. that acmeEmail is in the form of an email.

- Not sure what "Keep this in sync with the value of property ipName" means. Is this saying that the line "ipName: kubeflow-ip" should have a value that matches the same under "component: iap-ingress"? If so maybe the ipName line could have a #TODO comment on it like those in the bootstrapperConfig section?

- Had to look for env-kubeflow.sh. Maybe more clear to refer to "configs/cluster-kubeflow.yaml" and "configs/env-kubeflow.sh" to underline that these can be found in the configs/ dir.

- On https://github.com/kubeflow/kubeflow/blob/master/docs/gke/configs/env-kubeflow.sh#L10 maybe better to have project=\<your project name here\> than project=kubeflow as it draws attention to what people need to modify in this file. gke_setup.md just says to modify the file so it might be more clear for gke_setup.md to say they need to set the project if that's the only modification that's needed.

- With the command:
```
gcloud deployment-manager --project=${PROJECT} deployments create ${DEPLOYMENT_NAME} --config=${CONFIG_FILE}
```
I get the error:
```
ERROR: (gcloud.deployment-manager.deployments.create) Unable to read file 'cluster-\<name-of-my-project\>.yaml'. [Errno 2] No such file or directory: u'cluster-\<name-of-my-project\>.yaml'
```
Because the file in configs/ is named "cluster-kubeflow.yaml" not "cluster-\<name-of-my-project\>.yaml". If using a python wrapper you could generate a file with the right name before making an API call to deployment manager.

- After fixing the file name I see the following error:

```
...
Waiting for create [operation-1528221248255-56de8bca3f419-860f372d-fc711944]...failed.                                                                                                                                                                                                                          
ERROR: (gcloud.deployment-manager.deployments.create) Error in Operation [operation-1528221248255-56de8bca3f419-860f372d-fc711944]: errors:
- code: MANIFEST_EXPANSION_USER_ERROR
[...]

PR #931, opened 2018-06-05
fix vm service account config, doc update

kunmingg:
related to https://github.com/kubeflow/kubeflow/issues/927.
This bug should be covered by e2e test.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/931)

### DIFF
--- a/docs/gke/configs/cluster-kubeflow.yaml
+++ b/docs/gke/configs/cluster-kubeflow.yaml
@@ -75,7 +75,7 @@ resources:
     # You should set the following parameters in bootstrapperConfig.
     #    acmeEmail - Set this to your email address; this is used with lets-encrypt.
     #    ipName - This is the name of the static ip address reserved in GCP to be used with ingress.
-    #             For each deployment, value of ipName should be same among all configs
+    #             Keep the value of ipName in sync with the value in the bootstrapper config section below.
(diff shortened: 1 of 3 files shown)


jlewi:
Can you sync please? I think VM service account was fixed by #928

docs/gke/configs/cluster-kubeflow.yaml
```diff
@@ -60,16 +60,16 @@ resources:
...
+    # Each Kubeflow deployment in your project should use one unique ipName among all configs.
     ipName: kubeflow-ip
     # Provide the config for the bootstrapper. This should be a string
     # containing the YAML spec for the bootstrapper.
     #
     # You should set the following parameters in bootstrapperConfig.
     #    acmeEmail - Set this to your email address; this is used with lets-encrypt.
     #    ipName - This is the name of the static ip address reserved in GCP to be used with ingress.
-    #             keep this in sync with the value of property ipName
+    #             For each deployment, value of ipName should be same among all configs
```
jlewi: This comment is confusing. It makes it sound like every deployment should use the same value for ipName which isn't correct because we don't want different Kubeflow deployments to use the same name.

I would say something like

Keep the value of ipName in sync with the value in the bootstrapper config section below.

PR #948, opened 2018-06-07
ksonnet prototype for automated image release

kunmingg:
Initial version of image release prototype.
For https://github.com/kubeflow/kubeflow/issues/941

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/948)

Issue #781, opened 2018-05-09
Make release process cover bootstrapper image


kunmingg:
Release process should publish bootstrap image tagged with release version:
1. Update kubeflow version in [Dockerfile](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/Dockerfile)
2. Update bootstrap image on public registry.

PR #951, opened 2018-06-08
hide registries config in bootstrapper image to save user effort

kunmingg:
1. Save user effort and avoid confusion:
Adding new file ```image_registries.yaml```, which defines registries to be included in boostrap image and will be baked in to image during build.
When bootstrapper execute, it read ```image_registries.yaml``` from image and load all local registries.

2. Step 1 of https://github.com/kubeflow/kubeflow/issues/781: modify build_image.sh of bootstrapper so it follows config defined in https://github.com/kubeflow/kubeflow/pull/948.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/951)


bootstrap/build.py
```diff
@@ -44,19 +44,23 @@ def main(unparsed_args=None):
...
+    test_reg_path = args.test_registry.split(":")[1]
+
   for reg in conf['registries']:
-    if "repo" in reg and "branch" in reg:
+    if test_reg_name == reg["name"]:
+      src_registry = os.path.join(test_reg_path, reg["path"])
+      reg_path = os.path.join(tmp_dir, *os.path.join(reg["name"], reg["path"]).split('/')[:-1])
+      if not os.path.exists(reg_path):
+        os.makedirs(reg_path)
+      print("cp -r %s %s" % (src_registry, reg_path))
```
jlewi: nit use logging not print

jlewi:
Can a user override the registry baked into the container image? e.g. if I specify the Kubeflow registry in the bootstrapper config, will that pull the latest version of the registry as opposed to using the version baked into the image?

kunmingg:
@jlewi Not supported for now, currently baked-in registries have privilege. 
We can support overwrite registries version / source later, which will require user to provide github token.

jlewi:
Lets file a feature request for that. I think we will want this to be able to support private registries in a consistent fashion. It will also be useful for development. For development we'd like to be able to use the latest configs without having to rebuild the bootstrapper.

/lgtm
/approve

/hold If you want to submit as is and leave this for a future PR. IMO though it might be useful to just implement in this PR because I think it will be straightforward. This will also create a consistent experience for the use of private repositories. 

Here's my expectation of how this should work.

* We should be using ksonnet's support for file registries
* User can explicitly refer to a registry e.g
   name: kubeflow
   registry: file://opt/registries kubeflow

or
    name: kubeflow
    registry: http://github.com/kubeflow/kubeflow

http registries should always pull from the remote repository. We shouldn't automagically map "http" to a local "file" 

* For convenience if user doesn't specify registry, the bootstrapper will use the name of the registry to map it to the local file registry and use that.

* Note: a GitHub token isn't necessarily required to use remote repositories; in many cases the default quota should be sufficient.

PR #956, opened 2018-06-08
A bunch of fixes to the DM configs for GCP to work with latest bootstrapper

jlewi:
* docks/gke/configs/cluster-kubeflow.yaml was out of date and needed to
  be updated to work with the latest bootstrapper to image.
* Update the bootstrapper image so we pull in the latest registry configs.
* Make the instructions cleaner about what steps need to be performed to
  configure gcp.
* create_k8s_secrets.sh needs to create the kubeflow namespace because the
  namespace won't exist yet because the bootstrapper hasn't run because
  its blocked waiting for the GCP key.

* Don't include git tags in the image tag for the bootstrapper image.
* cluster-kubeflow.yaml is using the wrong name for the GCP service account.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/956)

### DIFF
--- a/docs/gke/configs/cluster-kubeflow.yaml
+++ b/docs/gke/configs/cluster-kubeflow.yaml
@@ -65,7 +65,7 @@ resources:
       # - user:john@acme.com
       # - group:data-scientists@acme.com
     # Path for the bootstrapper image.
-    bootstrapperImage: gcr.io/kubeflow-images-public/bootstrapper:latest
+    bootstrapperImage: gcr.io/kubeflow-images-public/bootstrapper:v20180607-v0.1.1-109-g4867d01b-dirty-601834
     # This is the name of the GCP static ip address reserved for your domain.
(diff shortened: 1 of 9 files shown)


docs/gke/configs/cluster-kubeflow.yaml
```diff
@@ -83,12 +83,20 @@ resources:
     #                 deployment in your project
     #          Project - This is the name of your GCP provided project.
     bootstrapperConfig: |
-      # Apps only apply if on GKE
+      registries:
```
kunmingg: We don't need registries here once https://github.com/kubeflow/kubeflow/pull/951 merged.

jlewi: Ok. We can consider removing them once #951 is merged.

jlewi:
@kunmingg PTAL

PR #965, opened 2018-06-11
Source admin role should be granted to KF service accounts.

jlewi:
* Source admin permission was granted to cloud services account but
  that service account is only used by deployment manager and shouldn't
  need source admin.

* We want the Kubeflow service accounts to have access to source admin
  so that we can push the ksonnet app to source control and so that users
  can access their code in pods.

/assign @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/965)


jlewi:
@ankushagarwal Could you re-lgtm please? I needed to sync to get tests to pass.

PR #967, opened 2018-06-11
Delete tf-cnn-benchmarks.jsonnet; there is now an example prototype.

jlewi:
* We no longer need the tf-cnn-benchmarks.jsonnet prototype; we now
  have an examples package that has a simple TF job that uses
  TFCNN by default.
* Update the instructions on running the TFCNN job.

/assign @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/967)


jlewi:
@pdmack @ankushagarwal Could someone please re-LGTM? I had to sync to get latest changes.

Issue #954, opened 2018-06-08
[gcp] bootstrapper fails to create CloudEndpoints resource


jlewi:
When I deploy on GKE I'm seeing the following error and no cloudendpoints resource is created.

```
{"filename":"app/server.go:243","level":"info","msg":"Using existing namespace: kubeflow","time":"2018-06-08T05:01:00Z"}
{"filename":"app/server.go:402","level":"info","msg":"Cluster version: v1.9.6-gke.1","time":"2018-06-08T05:01:00Z"}
{"filename":"app/server.go:217","level":"info","msg":"Storage class: standard","time":"2018-06-08T05:01:00Z"}
{"filename":"app/server.go:231","level":"info","msg":"StorageClass standard is default true","time":"2018-06-08T05:01:00Z"}
{"filename":"app/server.go:417","level":"info","msg":"Using K8s host https://10.7.240.1:443","time":"2018-06-08T05:01:00Z"}
{"filename":"env/create.go:78","level":"info","msg":"Creating environment \"default\" with namespace \"kubeflow\", pointing to cluster at address \"https://10.7.240.1:443\"","time":"2018-06-08T05:01:00Z"}
{"filename":"lib/lib.go:129","level":"info","msg":"Generating ksonnet-lib data at path '/opt/bootstrap/default/lib/v1.7.0'","time":"2018-06-08T05:01:01Z"}
{"filename":"app/server.go:441","level":"info","msg":"Successfully initialized the app /opt/bootstrap/default.","time":"2018-06-08T05:01:01Z"}
{"filename":"app/server.go:296","level":"info","msg":"Installing package kubeflow/core","time":"2018-06-08T05:01:01Z"}
{"filename":"registry/cache.go:83","level":"info","msg":"Retrieved 33 files","time":"2018-06-08T05:01:01Z"}
{"filename":"app/server.go:296","level":"info","msg":"Installing package kubeflow/tf-serving","time":"2018-06-08T05:01:01Z"}
{"filename":"registry/cache.go:83","level":"info","msg":"Retrieved 5 files","time":"2018-06-08T05:01:01Z"}
{"filename":"app/server.go:296","level":"info","msg":"Installing package kubeflow/tf-job","time":"2018-06-08T05:01:01Z"}
{"filename":"registry/cache.go:83","level":"info","msg":"Retrieved 6 files","time":"2018-06-08T05:01:01Z"}
[...]

PR #976, opened 2018-06-11
use ks apply in bootstrapper to avoid creation ordering issue

kunmingg:
1. fix https://github.com/kubeflow/kubeflow/issues/954:
```ks apply``` allow us to define CRDs and use them during same apply action,
while ```ks show | kubectl apply``` doesn't guarantee so.

2. edit gcp_prototype.yaml following new format to make sure it works.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/976)


jlewi:
When you say ks apply considers resource dependency order; what do you mean?
Are you just saying that if we do ks apply we don't seem to run into this particular error?

kunmingg:
@jlewi 
```ks apply``` allow us to define CRDs and use them during same apply action, 
while ```ks show | kubectl apply``` doesn't guarantee so.

PR #978, opened 2018-06-12
Fix permissions for jupyter-role

wmuizelaar:
The permissions handed out in the jupyter-role were allowing "*" in the
apiGroups, in stead of "". This means that it gives permissions to all
different kinds of apiGroups, in stead of the default one, which is
actually sufficient.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/978)


wmuizelaar:
Any idea what's going on with the test? It failed first, but it now states 'still running'?

pdmack:
@wmuizelaar can you rebase this? I'm assuming there's some OOD tests.

wmuizelaar:
Sure, done! Looks better now :)

PR #985, opened 2018-06-13
Consolidate GKE deployment script

activatedgeek:
/cc @jlewi @kunmingg @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/985)


docs/gke/configs/deploy.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
+
+set -e
```
ankushagarwal: `set -e` might be an issue because `kubectl create namespace` can fail if namespace is already created by bootstrapper.

docs/gke/configs/deploy.sh
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env bash
```
ankushagarwal: Can you add a comment explaining what this script does

docs/gke/configs/deploy.sh
activatedgeek: I did see that comment in the other script but it appears to be working even with setting this flag. I can add an `OR` flag after this command.

activatedgeek: I add an `unset -e` just before the namespace and secrets creation towards the end of the script.

docs/gke/configs/env-kubeflow.sh
```diff
@@ -19,5 +19,6 @@ export ZONE=us-east1-d
 # Set config file to the YAML file defining your deployment manager configs.
 export CONFIG_FILE=cluster-kubeflow.yaml
 
-# Get the project number
-export PROJECT_NUMBER=`gcloud projects describe ${PROJECT} --format='value(project_number)'`
+# Set OAuth2 Credentials
+export CLIENT_ID=
```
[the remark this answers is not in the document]
activatedgeek: I was actually considering this file as a reference for the variables needed rather than something to be checked into source control. Perhaps, we should put a warning in the documentation regarding the sensitivity of these credentials?

PR #987, opened 2018-06-13
Retry on ssl.SSLError in vm_util - this resolves test flakiness

ankushagarwal:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/987)

### DIFF
--- a/testing/vm_util.py
+++ b/testing/vm_util.py
@@ -47,6 +47,7 @@ def wait_for_operation(client,
                                            operation=op_id).execute()
     except ssl.SSLError as e:
       logging.error("Ignoring error %s", e)
+      continue
     status = op.get("status", "")
     # Need to handle other status's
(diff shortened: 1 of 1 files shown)


testing/vm_util.py
```diff
@@ -37,13 +38,15 @@ def wait_for_operation(client,
...
-
+    try:
+      if zone:
+        op = client.zoneOperations().get(
+          project=project, zone=zone, operation=op_id).execute()
+      else:
+        op = client.globalOperations().get(project=project,
+                                           operation=op_id).execute()
+    except ssl.SSLError as e:
+      logging.error("Ignoring error %s", e)
```
lluunn: should continue?

ankushagarwal: Done

Issue #953, opened 2018-06-08
[gcp] backend services can't be listed insufficient IAM permissions


jlewi:
In the scripts for setting IAM, I'm noticing that after creating a new deployment I'm getting the error

```
Waiting for backend id PROJECT=cloud-ml-dev NAMESPACE=kubeflow SERVICE=envoy...
ERROR: (gcloud.compute.backend-services.list) Some requests did not succeed:
 - Required 'compute.backendServices.list' permission for 'projects/cloud-ml-dev'
```
I believe in my case the service account already existed prior to the deployment being created because when I deleted the previous instance of the deployment the service accounts weren't deleted. The IAM permissions look correct.

I find that if I delete the service account and then update the deployment so its recreated it works.

My suspicion is that the problem is the service accounts aren't deleted when the previous deployment was deleted. 

I suspect the service accounts might not get deleted because the roles aren't deleted #910 and I suspect that might block service account creation. 

Although I believe @ankushagarwal mentioned that he observed the service account but not role bindings being deleted.

jlewi:
I'm hitting this again.

I created a delete script to try to prevent this by cleaning up service accounts and roles when deleting the deployment
https://github.com/jlewi/kubeflow-dev/blob/master/kubeflow-deployments/delete_deployment.sh

I also tried fixing it by doing this after the cluster had been created but just recreating the service account and roles 
https://github.com/jlewi/kubeflow-dev/blob/master/kubeflow-deployments/fix_admin_sa.sh

But when I tried it didn't fix the error.

lluunn:
@abhi-g hit this issue again with deploy app
In the iap container:
```
Waiting for backend id PROJECT=abhishek-codelab NAMESPACE=kubeflow SERVICE=envoy filter=name~k8s-be-30019-...
ERROR: (gcloud.compute.backend-services.list) Some requests did not succeed:
 - Required 'compute.backendServices.list' permission for 'projects/abhishek-codelab'
```

lluunn:
cc @kunmingg

kunmingg:
Seems caused by IAM binding leftover bug.
Web app fixed the problem before: https://github.com/kubeflow/kubeflow/pull/1595
We can implement same logic in script or wait till it switch to go binary cli.

jlewi:
How did the webapp fix this?

jlewi:
If we delete the policy bindings for the service account and then add them back doesn't that fix things?

kunmingg:
[...]

Issue #910, opened 2018-06-01
GCP Deployment Manager needs to delete IAM roles when DM is deleted


jlewi:
Our deployment manager configs for GCP create service accounts and IAM roles.

When the deployment is deleted we delete the service accounts but the roles persist.

To fix this we may need to add actions to remove the roles but we need to execute it only on delete.

jlewi:
I think we can deal with this by adding actions to remove the roles and setting runtimePolicy to DELETE.
https://github.com/GoogleCloudPlatform/deploymentmanager-samples/tree/master/examples/v2/cloudbuild

PR #1001, opened 2018-06-14
fix iam role delete for deployment manager

kunmingg:
Fix https://github.com/kubeflow/kubeflow/issues/910.
Fix https://github.com/kubeflow/kubeflow/issues/953.

Deploy manager iam role works similar to k8s multi-pod locks: one version can only have one consumer.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1001)


ankushagarwal:
/lgtm

/hole
In case @jlewi has comments

PR #1002, opened 2018-06-14
Use docs/gke/configs to run e2e tests

### LINKED ISSUE #950: [gcp] cluster-kubeflow.yaml isn't tested
This is the recommended DM and bootstrapper config for GKE deployments.
https://github.com/kubeflow/kubeflow/blob/master/docs/gke/configs/cluster-kubeflow.yaml

It doesn't like that YAMl file is used by our E2E tests because it wasn't updated to specify the registry when that change was made to 
https://github.com/kubeflow/kubeflow/tree/master/bootstrap/config

There is also another GCP bootstrapper config in that directory

Our E2E tests are using this DM config
https://github.com/kubeflow/kubeflow/blob/master/testing/dm_configs/cluster-kubeflow.yaml

We do need to make changes to the DM config in order to pull the registry at head.

We have two options

1. We could create a separate test for https://github.com/kubeflow/kubeflow/blob/master/docs/gke/configs/cluster-kubeflow.yaml
1. We could make the necessary modifications to https://github.com/kubeflow/kubeflow/blob/master/docs/gke/configs/cluster-kubeflow.yaml

    * We could use a simple shell script that runs jq to make the necessary field changes

I prefer #2. if we use jq then we could update the [instructions](https://github.com/kubeflow/kubeflow/blob/master/docs/gke/gke_setup.md) to use the same jq commands.

@ankushagarwal thoughts?

ankushagarwal:
* Use docs/gke/configs to bring up test clusters
* Update default bootstrapper image to gcr.io/kubeflow-images-public/bootstrapper:v20180614-052b002-dirty-f2d8e3
* Add pytorch-job to packages
* Create deploy_kubeflow_gcp.sh to deploy kubeflow 
* Add update_dm_config.py - this is used to update params in deployment manager config
* Update workflows.libsonnet with the new workflow
Fixes #950

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1002)


ankushagarwal:
dev.kubeflow.org resources were provisioned manually - before deployment manager was a thing. 

What exactly was the secret/sa change needed for dev.kubeflow.org?

pdmack:
admin-gcp-sa

jlewi:
@pdmack At some point we should bring up a new d.k.o cluster using deployment manger; checking in the DM yaml files into the repository. 

/lgtm
/approve

jlewi:
Looks like a weird PRC error pulling from git in checkout step

```
+ REPO_DIR=/mnt/test-data-volume/kubeflow-presubmit-kubeflow-e2e-minikube-1002-22f2128-2153-947f/src/kubeflow/kubeflow
Job Name = kubeflow-presubmit
+ cd /mnt/test-data-volume/kubeflow-presubmit-kubeflow-e2e-minikube-1002-22f2128-2153-947f/src/kubeflow/kubeflow
+ '[' '!' -z 1002 ']'
+ git fetch origin pull/1002/head:pr
fatal: The remote end hung up unexpectedly
```

jlewi:
@ankushagarwal Maybe you should set the retryConfig for the checkout step?

ankushagarwal:
Retrying `git clone` will fail if the directory already exists :-/

ankushagarwal:
I am seeing a lot of issues with deployment manager when it comes to updating iam policy in a project in parallel deployments. Some of the deployments get stuck in an unrecoverable / un-deletable stage. This makes it really hard to reliably test.

ankushagarwal:
@jlewi Should we try to use separate project per e2e-test run?

jlewi:
I'd like to avoid separate project per e2e run if we can avoid it.

ankushagarwal:
Bug in argo retryStrategy can cause this workflow to be marked as failed despite passing
https://github.com/argoproj/argo/issues/885

ankushagarwal:
FYI - this seems to be stable now, but it takes at least 25 minutes for each run

jlewi:
Review status: 0 of 12 files reviewed, 6 unresolved discussions (waiting on @ankushagarwal, @wbuchwalter, @kunmingg, and @jlewi)

---

*[docs/gke/configs/cluster-kubeflow.yaml, line 126 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKk-_0C4CRaU9KFXqA:-LFKk-_1Bjm5bHGQJMCx:bx12y6r) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/docs/gke/configs/cluster-kubeflow.yaml#L126)):*
> ```YAML
>             name: jupyterHubAuthenticator
>             value: iap
>           - component: kubeflow-core
> ```

Is there a reason we need to override tfJobVersion here? The default should now be v1alpha2 right?

---

*[docs/gke/configs/deploy.sh, line 14 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKk53xBdSfnTKjJinK:-LFKk53xBdSfnTKjJinL:b-9755fn) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/docs/gke/configs/deploy.sh#L14)):*
> ```Shell
> #
> 
> set -xe
> ```

Thank you!

---

*[testing/deploy_kubeflow_gcp.sh, line 47 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKkJU5DlQnIQgH5CEw:-LFKkJU5DlQnIQgH5CEx:b-938vfq) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/testing/deploy_kubeflow_gcp.sh#L47)):*
> ```Shell
> set +e
> # Delete the existing deployment in case we end up retying this script
> gcloud deployment-manager deployments delete ${DEPLOYMENT_NAME} --quiet --project=kubeflow-ci
> ```

Can you add a TODO. I think it would be better if deploy.sh tested if the deployment existed and if it did issued
an update rather than a create.

---

*[testing/run_with_retry.py, line 7 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKkbLsA_lEsWxc9dfA:-LFKkbLsA_lEsWxc9dfB:bvsoi33) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/testing/run_with_retry.py#L7)):*
> ```Python
> python run_with_retry --retries=5 -- bash my_flaky_script.sh
> 
> This runs bash my_flaky_script.sh upto 5 times till it succeeds
> ```

Is this only intended for bash scripts and not python scripts?

Is it not possible to do retries with Argo?

---

*[testing/teardown_kubeflow_gcp.sh, line 16 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKkpsJCCT-TpScenPk:-LFKkpsKC9gi9pEOikVN:b2891co) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/testing/teardown_kubeflow_gcp.sh#L16)):*
> ```Shell
> # Add a jitter to reduce the chance of deployments updating at the same time
> # since tests are run in parallel
> sleep $((${RANDOM} % 30))
> ```

Is this because you might have two E2E tests running simultaneously so they are in "sync"?

---

*[testing/teardown_kubeflow_gcp.sh, line 20 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKkkgE6A29pIWSmSM9:-LFKkkgE6A29pIWSmSMA:b41gppk) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/testing/teardown_kubeflow_gcp.sh#L20)):*
> ```Shell
> # We need to run an update because for deleting IAM roles,
> # we need to obtain a fresh copy of the IAM policy. A stale
> # copy of IAM policy causes issues while deletion.
> ```

during deletion

---


*Comments from [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1002)*

jlewi:
Review status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @ankushagarwal, @jlewi, @wbuchwalter, and @kunmingg)

---

*[testing/run_with_retry.py, line 7 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKkbLsA_lEsWxc9dfA:-LFMvhnC6BkjClqpZmq-:bbaoxjc) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/testing/run_with_retry.py#L7)):*
<details><summary><i>Previously, ankushagarwal (Ankush Agarwal) wrote…</i></summary><blockquote>

Can be used with any scripts. retryStrategy in argo has a bug https://github.com/argoproj/argo/issues/885
</blockquote></details>

nit can you add a comment explaining this? Wrapping a python script with run_with_retry seems really inelegant compared to adding retries directly to the python code. So I'd like a comment explaining that to prevent people from doing that. And link to the argo bug as that seems more elegant if it works.

---


*Comments from [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1002)*

ankushagarwal:
Review status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on @ankushagarwal, @jlewi, @wbuchwalter, and @kunmingg)

---

*[testing/run_with_retry.py, line 7 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/1002#-LFKkbLsA_lEsWxc9dfA:-LFNayeYAfMGmpeWVt2n:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/35706b6fef64401f9c73d37c33eec87a734c42a1/testing/run_with_retry.py#L7)):*

nit can you add a comment explaining this? Wrapping a python script with run_with_retry seems really inelegant compared to adding retries directly to the python code. So I'd like a comment explaining that to prevent people from doing that. And link to the argo bug as that seems more elegant if it works.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1002)*

Issue #1006, opened 2018-06-14
Bootstrapper should apply components in an order


ankushagarwal:
Right now when bootstrapper applies components, order is not enforced - this creates issues sometimes when the CRD is defined after the object is created - it causes bootstrapper to fail un-recoverably. 

We should apply the components in the order in which they are defined in the config file.

For kubeflow, this correct order is `cert-manager, cloud-endpoints, iap-ingress, kubeflow-core`

/assign @kunmingg

PR #1007, opened 2018-06-14
Order the kubeflow components correctly

ankushagarwal:
Having them out of order can cause bootstrapper to fail

Related to https://github.com/kubeflow/kubeflow/issues/1006
/assign @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1007)


kunmingg:
Seems current config order is working, and ```kubeflow-core``` should be first since we don't want to have  ```kubeflow-core``` having dependencies on other optional components.

ankushagarwal:
Sounds good. Closing this PR.

PR #1008, opened 2018-06-15
update ks version; run ks apply following config order

kunmingg:
1. update ks version in bootstrap to 0.11.0
2. run ks apply for each component following orders in config file
fix https://github.com/kubeflow/kubeflow/issues/1006.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1008)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/server.go
+++ b/bootstrap/cmd/bootstrap/app/server.go
@@ -25,9 +25,6 @@ import (
 	"regexp"
 	"strconv"
 	"time"
-	"os/exec"
-	"bytes"
-
 	"github.com/ghodss/yaml"
 	"github.com/ksonnet/ksonnet/actions"
 	kApp "github.com/ksonnet/ksonnet/metadata/app"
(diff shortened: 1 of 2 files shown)


bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -534,19 +533,23 @@ func Run(opt *options.ServerOption) error {
...
+			rawCmd := fmt.Sprintf(
+				"ks apply default -c %v --token=$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)",
+				component.Name)
+			applyCmd := exec.Command("bash", "-c", rawCmd)
+
+			var out bytes.Buffer
+			var stderr bytes.Buffer
+			applyCmd.Stdout = &out
+			applyCmd.Stderr = &stderr
+			if err := applyCmd.Run(); err != nil {
```
ankushagarwal: Can you print both stdout and stderr regardless of whether there was an error or not?

It helps understand what's going right/wrong

kunmingg: updated

Issue #462, opened 2018-03-20
E2E test for tf-job-simple prototype


jlewi:
See #458. 

The TFCnn prototype was broken by a change to our templates.

We should add some ksonnet tests for the TFCnn prototype that would have caught this.

jlewi:
We should consider updating it to use v1alpha2 as well.

jlewi:
We no longer have a prototype for TFCNN. We now just have tf-job-simple which runs TFCNN by default.

We should an E2E test to verify we can actually run jobs using this prototype.

At a minimum we should verify that we can start a job using this prototype and make sure the pods enter the running state. 

The job takes a long time to run so we may not want to wait for it to complete.

PR #1010, opened 2018-06-15
Add test for tf-job-simple prototype

ankushagarwal:
tf_job_simple_test.py tests the tf-job-simple prototype by creating a tfjob from that prototype and checking that the appropriate pods and service accounts get created

Fixes https://github.com/kubeflow/kubeflow/issues/462

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1010)

### DIFF
--- a/testing/tf_job_simple_test.py
+++ b/testing/tf_job_simple_test.py
@@ -42,13 +42,11 @@ def parse_args():
 def wait_for_tf_job():
   out = util.run(["kubectl", "get", "pods", "-l",
                   "tf_job_name=mycnnjob", "-ndefault"])
-  logging.info("outlen = %d", len(out.split("\n")))
   if "No resources found" in out or len(out.split("\n")) != 3:
     raise Exception("Could not find pods with label tf_job_name=mycnnjob")
   logging.info("Found pods with label tf_job_name=mycnnjob")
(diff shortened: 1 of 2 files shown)


ankushagarwal:
/cc @lluunn @kunmingg 

This is ready for review

testing/tf_job_simple_test.py
```diff
@@ -0,0 +1,76 @@
...
+    default="",
+    type=str,
+    help="The kubeflow src directory")
+  args, _ = parser.parse_known_args()
+  return args
+
+@retry(wait_fixed=5000, stop_max_attempt_number=20)
+def wait_for_tf_job():
+  out = util.run(["kubectl", "get", "pods", "-l",
+                  "tf_job_name=mycnnjob", "-ndefault"])
```
lluunn: If namespace is default, do we need to delete them? I guess jobs will gone after finish, how about the service?

ankushagarwal: We will be deleting the cluster at the end so I did not bother deleting this.

lluunn: Are we using kubeflow-testing cluster in kubeflow-ci?

ankushagarwal: No, we'll be creating a new cluster for every test

PR #1028, opened 2018-06-19
Facilitate specifying seperate requests and limits

wmuizelaar:
When starting the jupyter notebook pods, it is good to also be able to
specify limits and not just only requests.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1028)

### DIFF
--- a/kubeflow/core/kubeform_spawner.py
+++ b/kubeflow/core/kubeform_spawner.py
@@ -27,12 +27,20 @@ class KubeFormSpawner(KubeSpawner):
     </datalist>
     <br/><br/>
 
-    <label for='cpu_guarantee'>CPU</label>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+    <label for='cpu_guarantee'>CPU Request</label>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
     <input name='cpu_guarantee' placeholder='200m, 1.0, 2.5, etc'></input>
     <br/><br/>
 
-    <label for='mem_guarantee'>Memory</label>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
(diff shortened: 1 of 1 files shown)


ankushagarwal:
Hi @wmuizelaar 

I am not sure if adding a limit to the UI is a very good idea. Firstly it adds the overhead of explaining to users what the difference b/w request and limit is - it is not very obvious for users who are not k8s experts. Even after that, for a compute and memory intensive pod like a jupyter notebook it is very hard for a user to estimate "after how much memory consumption should my pod be OOM killed".

wmuizelaar:
@ankushagarwal yes - in the clusters we have running, we have a default limitrange set in place, with default request and limit values. 

If you then start a notebook with for example 4 CPU's as a request, but your default limit is set to 1 CPU, then the pod doesn't start, and you will get a nice red warning that the pod cannot be scheduled.

But I agree that it gives extra overhead in explaining. Maybe hide it behind a feature-flag?

ankushagarwal:
Hiding behind a feature flag SGTM. 

1. Can you add a [tooltip](https://jqueryui.com/tooltip/) to the fields which explains what Limit and Request is
2. Can you update the logic so that when the Limit fields are empty, limits are not set for the Pods

jlewi:
@wmuizelaar Any interest in updating the PR to address comments? If not we'll close it out.

wmuizelaar:
Yes, I want to update the PR! Have been out of office for a bit, but please leave this open.

Thanks,
Wietse

Sent from my iPhone

On 27 Jul 2018, at 6:52 AM, Jeremy Lewi <notifications@github.com<mailto:notifications@github.com>> wrote:


@wmuizelaar<https://github.com/wmuizelaar> Any interest in updating the PR to address comments? If not we'll close it out.

ankushagarwal:
Closing this because of lack of activity

PR #1037, opened 2018-06-20
[gcp-deployer] Move to Typescript, clear dead code

yebrahim:
image

Also:
- Clears some dead code in HTML, CSS, and TSX.
- Removes unneeded test. We can re-add this later if we think it's useful, and if we add it to Travis as well.
- Separate out `NameForm` into its own file and define types for it.
- Removed unneeded service worker.

Review guide:
- (Always) Skip `package-lock.json`.
- `module_types.d.ts`, `tsconfig.*`, and `tslint.json` are all needed by the Typescript compiler and linter, no impact on functionality.
- Other than this, no change in functionality to make reviewing a little easier. These will come in alter PRs.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1037)


jlewi:
Thank you.

/lgtm
/approve

PR #1035, opened 2018-06-20
Update examples for 0.2.

jlewi:
* Make tf-job-simple a v1alpha2 prototype
* Rename the existing tf-job-simple-v1alpha1 prototype
* Add a prototype for tensorboard.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1035)

PR #1043, opened 2018-06-20
add mpi operator ksonnet package

rongou:
/cc @jlewi @everpeace @gaocegege @alsrgv 

Closes https://github.com/kubeflow/mpi-operator/issues/17

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1043)


kubeflow/mpi-job/prototypes/mpi-job-custom.jsonnet
```diff
@@ -0,0 +1,26 @@
+// @apiVersion 0.1
+// @name io.ksonnet.pkg.mpi-job-custom
+// @description A custom MPI Job.
+// @shortDescription A custom MPI Job that allows you to specify number of replicas and GPUs per replica.
+// @param name string Name to give to each of the components
+// @param replicas number Total number of replicas for the job
+// @param gpusPerReplica number Total number of GPUs per replica
+// @param image string The image to use for the job
```
kunmingg: Which image should user use here? Is it publicly available? If it is we can set default image here.
Maybe also set default value for replicas and gpusPerReplica?

kubeflow/mpi-job/prototypes/mpi-job-simple.jsonnet
```diff
@@ -0,0 +1,24 @@
+// @apiVersion 0.1
+// @name io.ksonnet.pkg.mpi-job-simple
+// @description A simple MPI Job.
+// @shortDescription A simply MPI Job that allows you to specify the total number of GPUs.
+// @param name string Name to give to each of the components
+// @param gpus number Total number of GPUs for the job
+// @param image string The image to use for the job
```
kunmingg: Same here

kubeflow/mpi-job/prototypes/mpi-job-custom.jsonnet
rongou: This is a user's job, so I suppose it should use whatever image the user has built? We can put in the tensorflow benchmark image here as the default, but I'm not sure if that's very sensible.

What would be good defaults for replicas and gpusPerReplica? Set them both to 1?

rongou: OK, make them optional and added defaults.

kubeflow/mpi-job/prototypes/mpi-job-simple.jsonnet
rongou: ditto

kunmingg:
@rongou Got error when try to deploy it:
```ERROR handle object: can't update clusterroles mpi-operator: clusterroles.rbac.authorization.k8s.io "mpi-operator" is forbidden: attempt to grant extra privileges:```

Did you test deploy on k8s cluster with rbac on?

**Comment by rongou** [comment]
@kunmingg yes I just tested it again both locally on my desktop and on an Amazon EKS cluster. I think the error message means your kubernetes credential doesn't give you enough permissions? Are you running as an admin? The cluster role for the operator does need a lot of access since it's managing rbac for mpi jobs.

File: kubeflow/mpi-job/mpi-operator.libsonnet [hunk 3]
```diff
@@ -0,0 +1,276 @@
...
+          resources: [
+            "events",
+          ],
+          verbs: [
+            "create",
+            "patch",
+          ],
+        },
+        {
+          apiGroups: [
```
kunmingg: Can we remove rule for ```rbac.authorization.k8s.io```?
This one will give permission which equivalent to ```cluster-admin```.

rongou: This is needed to create new roles/rolebindings to lock down the MPI jobs since they need to communicate with other pods and do remote execution. We actually create a new role/serviceaccount/rolebinding for each MPIJob. There is some information about it in the design proposal: https://github.com/kubeflow/community/blob/master/proposals/mpi-operator-proposal.md

kubeflow/mpi-job/mpi-job.libsonnet
```diff
@@ -0,0 +1,58 @@
+local k = import "k.libsonnet";
+
+{
+  parts:: {
+    simple(namespace, name, gpus, image, command, args):: {
```
jlewi: The current thinking is that for the jobs the whole spec should be in the prototype file and not an import from a library. This way ks generate is the equivalent of copying an example. This makes it easier for users to do advanced modifications (e.g. set volume mounts).

Example:

kubeflow/mpi-job/prototypes/mpi-job-custom.jsonnet
```diff
@@ -0,0 +1,26 @@
...
+local namespace = if params.namespace != "null" then params.namespace else env.namespace;
+local name = import "param://name";
+local replicas = import "param://replicas";
+local gpusPerReplica = import "param://gpusPerReplica";
+local image = import "param://image";
+local command = import "param://command";
+local args = import "param://args";
+
+std.prune(k.core.v1.list.new([
+  mpi.parts.custom(namespace, name, replicas, gpusPerReplica, image, command, args),
```
jlewi: I'd suggest putting the full spec inside the jsonnet file as in

This way the user will have complete access to the spec inside the resulting jsonnet file and can easily customize it for their use case.

kubeflow/mpi-job/prototypes/mpi-operator.jsonnet
```diff
@@ -0,0 +1,26 @@
...
+// @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
+// @optionalParam image string mpioperator/mpi-operator:latest The image for the MPI Operator.
+// @optionalParam kubectlDeliveryImage string mpioperator/kubectl-delivery:latest The image for delivering kubectl.
+// @optionalParam gpusPerNode number 8 The maximum number of GPUs per node.
+
+local k = import "k.libsonnet";
+local operator = import "kubeflow/mpi-job/mpi-operator.libsonnet";
+
+local namespace = if params.namespace != "null" then params.namespace else env.namespace;
+local name = import "param://name";
```
jlewi: You can just do
local name=params.name

params will be defined in the resulting jsonnet file
https://github.com/kubeflow/kubeflow/pull/1035/files

I think its cleaner than doing import "param:/...."

kubeflow/mpi-job/prototypes/mpi-operator.jsonnet
```diff
@@ -0,0 +1,26 @@
...
+// @param name string Name to give to each of the components.
+// @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
+// @optionalParam image string mpioperator/mpi-operator:latest The image for the MPI Operator.
+// @optionalParam kubectlDeliveryImage string mpioperator/kubectl-delivery:latest The image for delivering kubectl.
+// @optionalParam gpusPerNode number 8 The maximum number of GPUs per node.
+
+local k = import "k.libsonnet";
+local operator = import "kubeflow/mpi-job/mpi-operator.libsonnet";
+
+local namespace = if params.namespace != "null" then params.namespace else env.namespace;
```
jlewi: I suggest
local namespace = env.namespace

The rest harkens back to earlier versions of ksonnet in which we couldn't access environment namespace.

kubeflow/mpi-job/prototypes/mpi-operator.jsonnet
```diff
@@ -0,0 +1,26 @@
+// @apiVersion 0.1
+// @name io.ksonnet.pkg.mpi-operator
+// @description MPI Operator.
+// @shortDescription MPI Operator for running MPI jobs.
+// @param name string Name to give to each of the components.
+// @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
+// @optionalParam image string mpioperator/mpi-operator:latest The image for the MPI Operator.
+// @optionalParam kubectlDeliveryImage string mpioperator/kubectl-delivery:latest The image for delivering kubectl.
+// @optionalParam gpusPerNode number 8 The maximum number of GPUs per node.
+
```
everpeace: Sometimes, cluster-admin wanted to manage RBAC related resources on their own or to use existing service account.  In the case, admins don't want to let the prototype create them.  So, 

How about introducing `rbacCreate` and `serviceAccountName` parameters??  It's popular way in helm packages.  In this case, we need explicit guidance so that admins can refer to required roles when they choose `rbacCreate=false`.

rongou: Good idea, but it'll involve some code changes. Future enhancement? Tracking it at https://github.com/kubeflow/mpi-operator/issues/20.

kubeflow/mpi-job/mpi-operator.libsonnet
kunmingg: OK, in this case can we remove ```clusterRole``` and just do ```clusterRoleBinding``` to ```cluster-admin```?
With access to role and role binding user can get themselves admin access.
So let's save effort of define custom role.

jlewi: I think its better to have a specific role. This makes it easier to add/remove permissions needed by this role.

@rongou Do you create job specific roles/rolebindings?

jlewi:
/lgtm
/approve
/hold 

My comments have been addressed thank you.

@kunmingg any additional comments?

kubeflow/mpi-job/mpi-operator.libsonnet
rongou: Right, the role is very specific to the individual `MPIJob`. The launcher has `pods/exec` permissions only to the worker pods. See https://github.com/kubeflow/mpi-operator/blob/master/pkg/controllers/mpi_job_controller.go#L821

Here is a sample role:
```yaml
kind: Role
metadata:
  creationTimestamp: 2018-06-11T23:14:22Z
  labels:
    app: tensorflow-benchmarks-8
  name: tensorflow-benchmarks-8-launcher
  namespace: default
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: MPIJob
    name: tensorflow-benchmarks-8
    uid: 2baf54b7-6dcd-11e8-9b38-029ed2ab0d38
  resourceVersion: "466996"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/default/roles/tensorflow-benchmarks-8-launcher
  uid: 2cf3b945-6dcd-11e8-b6ec-06d57706f108
rules:
- apiGroups:
  - ""
  resourceNames:
  - tensorflow-benchmarks-8-worker-0
  - tensorflow-benchmarks-8-worker-1
  - tensorflow-benchmarks-8-worker-2
  - tensorflow-benchmarks-8-worker-3
  - tensorflow-benchmarks-8-worker-4
  - tensorflow-benchmarks-8-worker-5
  - tensorflow-benchmarks-8-worker-6
  - tensorflow-benchmarks-8-worker-7
  resources:
  - pods
  verbs:
  - get
- apiGroups:
  - ""
  resourceNames:
  - tensorflow-benchmarks-8-worker-0
  - tensorflow-benchmarks-8-worker-1
  - tensorflow-benchmarks-8-worker-2
  - tensorflow-benchmarks-8-worker-3
  - tensorflow-benchmarks-8-worker-4
  - tensorflow-benchmarks-8-worker-5
  - tensorflow-benchmarks-8-worker-6
  - tensorflow-benchmarks-8-worker-7
  resources:
  - pods/exec
  verbs:
  - create
```

kunmingg:
Can not deploy with ks 0.11.0, get:
```panic: unknown Expr1 type *ast.Conditional```

kunmingg:
Filed an issue in ksonnet: https://github.com/ksonnet/ksonnet/issues/637
I guess there are some bug in prototype which ksonnet cannot catch.

rongou:
@kunmingg should be fixed now.

kunmingg:
@rongou Looks good, please update doc to remind user mpi k8s version >= 1.9, when have time.
/hold cancel

PR #1044, opened 2018-06-20
WIP - IAP

ankushagarwal:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1044)

### DIFF
--- a/docs/gke/configs/deploy.sh
+++ b/docs/gke/configs/deploy.sh
@@ -18,15 +18,11 @@ export PROJECT=${PROJECT:-}
 export DEPLOYMENT_NAME=${DEPLOYMENT_NAME:-}
 export ZONE=${ZONE:-}
 export CONFIG_FILE=${CONFIG_FILE:-}
-export CLIENT_ID=${CLIENT_ID:-}
-export CLIENT_SECRET=${CLIENT_SECRET:-}
 
 if [ -z "${PROJECT}" ] || \
    [ -z "${DEPLOYMENT_NAME}" ] || \
    [ -z "${ZONE}" ] || \
-   [ -z "${CONFIG_FILE}" ] || \
-   [ -z "${CLIENT_ID}" ] || \
-   [ -z "${CLIENT_SECRET}" ]; then
+   [ -z "${CONFIG_FILE}" ]; then
   echo 'Required variables missing. Please check again!'
   exit 1
 fi
@@ -76,4 +72,3 @@ kubectl create namespace ${K8S_NAMESPACE}
 kubectl create secret generic --namespace=${K8S_ADMIN_NAMESPACE} admin-gcp-sa --from-file=admin-gcp-sa.json=./${SA_EMAIL}.json
 kubectl create secret generic --namespace=${K8S_NAMESPACE} admin-gcp-sa --from-file=admin-gcp-sa.json=./${SA_EMAIL}.json
 kubectl create secret generic --namespace=${K8S_NAMESPACE} user-gcp-sa --from-file=user-gcp-sa.json=./${USER_EMAIL}.json
-kubectl create secret generic --namespace=${K8S_NAMESPACE} kubeflow-oauth --from-literal=CLIENT_ID=${CLIENT_ID} --from-literal=CLIENT_SECRET=${CLIENT_SECRET}
--- a/docs/gke/configs/env-kubeflow.sh
+++ b/docs/gke/configs/env-kubeflow.sh
@@ -4,7 +4,7 @@
 # This is script defines values for all the variables used in
 # the instructions.
 
-# Bucket and project must be unique for each project 
+# Bucket and project must be unique for each project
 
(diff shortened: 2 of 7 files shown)


kubeflow/core/setup_iap.sh
```diff
@@ -68,9 +71,29 @@ echo BACKEND_ID=${BACKEND_ID}
 NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[0].nodePort}')
 BACKEND_SERVICE=$(gcloud --project=${PROJECT} compute backend-services list --filter=name~k8s-be-${NODE_PORT}- --uri)
 # Enable IAP on the backend service:
-gcloud --project=${PROJECT} compute backend-services update ${BACKEND_SERVICE} \
-    --global \
-    --iap=enabled,oauth2-client-id=${CLIENT_ID},oauth2-client-secret=${CLIENT_SECRET}
+iap_patch='.iap = {
+  "enabled": true,
+  "oauth2ClientInfo": {
+    "developerEmailAddress": "iap@kubeflow.org",
```
jlewi: Can we use an environment variable to pass in the user's email address and not a kubeflow.org email?

kubeflow/core/setup_iap.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 #
 # A simple shell script to enable IAP and configure timeouts by using gcloud.
+
+set -x
+
 [ -z ${CLIENT_ID} ] && echo Error CLIENT_ID must be set && exit 1
```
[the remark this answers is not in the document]
ankushagarwal: Done

kubeflow/core/setup_iap.sh
ankushagarwal: Done

docs/gke/configs/deploy.sh
```diff
@@ -76,4 +73,4 @@ kubectl create namespace ${K8S_NAMESPACE}
 kubectl create secret generic --namespace=${K8S_ADMIN_NAMESPACE} admin-gcp-sa --from-file=admin-gcp-sa.json=./${SA_EMAIL}.json
 kubectl create secret generic --namespace=${K8S_NAMESPACE} admin-gcp-sa --from-file=admin-gcp-sa.json=./${SA_EMAIL}.json
 kubectl create secret generic --namespace=${K8S_NAMESPACE} user-gcp-sa --from-file=user-gcp-sa.json=./${USER_EMAIL}.json
-kubectl create secret generic --namespace=${K8S_NAMESPACE} kubeflow-oauth --from-literal=CLIENT_ID=${CLIENT_ID} --from-literal=CLIENT_SECRET=${CLIENT_SECRET}
+kubectl create secret generic --namespace=${K8S_NAMESPACE} kubeflow-oauth-email --from-literal=EMAIL_ADDRESS=${OAUTH_EMAIL_ADDRESS}
```
jlewi: nit this shouldn't be a secret; it can just be a config map since email address doesn't need to be secret.

docs/gke/configs/deploy.sh
```diff
@@ -18,15 +18,12 @@ export PROJECT=${PROJECT:-}
...
-export CLIENT_ID=${CLIENT_ID:-}
-export CLIENT_SECRET=${CLIENT_SECRET:-}
 
 if [ -z "${PROJECT}" ] || \
    [ -z "${DEPLOYMENT_NAME}" ] || \
    [ -z "${ZONE}" ] || \
-   [ -z "${CONFIG_FILE}" ] || \
-   [ -z "${CLIENT_ID}" ] || \
-   [ -z "${CLIENT_SECRET}" ]; then
+   [ -z "${EMAIL_ADDRESS}" ] || \
```
jlewi: I'm guessing the email address doesn't actually matter?
So lets not make EMAIL_ADDRESS required; lets set it to some default that makes it clear its some dummy value
e.g.
johndoe@acme.com

This way its one less thing and parameter we have to set. And we can get rid of the extra secret below.

We could make emailAddress a parameter of the component and then just set the environment variable directly on the deployment so we don't have to create a config map.

docs/gke/configs/env-kubeflow.sh
```diff
@@ -4,11 +4,15 @@
...
 
-# Bucket and project must be unique for each project 
+# Bucket and project must be unique for each project
 
 # Set PROJECT to the project you want to use with Kubeflow.
 export PROJECT=<your_project>
 
+# This should be the cluster admin's email address. This is used for
+# OAuth Login screen.
+export OAUTH_EMAIL_ADDRESS=user@example.com
```
jlewi: Just make this a parameter on the ksonnet component and set the environment variable directly on the K8s deployment without creating a configmap

jlewi:
Thank you; so excited about finally being able to automate this.

ankushagarwal:
Closing since this is obsolete

PR #1045, opened 2018-06-20
test-dir-delete should depend upon teardown to complete

ankushagarwal:
If test directory gets deleted, teardown won't work, so we need to
ensure that test-dir-delete happens only after teardown steps

Add test-dir-delete for minikube workflow as well

/assign @lluunn


lluunn:
Should minikube test_dir_delete depend on tear down as well?

PR #1047, opened 2018-06-20
A bunch of fixes for deploying with private ks registry

IronPan:
- use package registry name instead of hard coded "kubeflow"
- Avoid passing component parameter when creating component. The subsequent step will instead pass it. 

Got following error otherwise
{"filename":"bootstrap/main.go:48","level":"error","msg":"Bootstrapper failed with error: There was a problem creating component FOO: parse preview args: unknown flag: --BAR\n","time":"2018-06-20T19:39:44Z"}


IronPan:
I revert one of the fix, as it turned out it's a bug on the private ks registry side.

PR #1052, opened 2018-06-21
Update default notebook image for spawner

### LINKED ISSUE #1041: Jupyter can't start pod; the default spawner image is way too old
I got the following error when trying to start a notebook

```
kubectl logs jupyter-accounts-2egoogle-2ecom-3ajlewi-40google-2ecom
[FATAL tini (7)] Executing child process 'start-singleuser.sh' failed: 'Permission denied'
```
Notebook image is
image: gcr.io/kubeflow/tensorflow-notebook-cpu:latest
imageID: docker-pullable://gcr.io/kubeflow/tensorflow-notebook-cpu@sha256:964dbfa5d435aa61b2d3b90cf4c7992040b83d2dd9525fc3f518316b03f64b62

I didn't specify any image when I clicked spawn.

It looks like our default image used by the spawner is "latest" which is pointing to a very old image (even the registry is wrong).

We should probably be pinning to a specific image as our default image so we don't get broken by moving the latest tag.

We should fix this in 0.2.

@pdmack Could you take a look?

/assign @pdmack 

@kunmingg We should cherry pick thia after its fixed.

pdmack:
Fixes #1041 

/assign @jlewi 
/assign @kunmingg 

@cheyang do you have a mirror image that you want me to add here for alibaba?

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1052)

### DIFF
--- a/kubeflow/core/kubeform_spawner.py
+++ b/kubeflow/core/kubeform_spawner.py
@@ -57,7 +57,7 @@ class KubeFormSpawner(KubeSpawner):
         if cloud == 'ack':
             image = 'registry.aliyuncs.com/kubeflow-images-public/tensorflow-notebook-cpu'
         else:
-            image = 'gcr.io/kubeflow/tensorflow-notebook-cpu'
+            image = 'gcr.io/kubeflow-images-public/tensorflow-1.8.0-notebook-cpu:v20180619-c79195b3'
         if self.user_options.get('image'):
             image = self.user_options['image']
         return image


jlewi:
Thanks

It would be nice if Alibaba was using the same image by default just to avoid confusion created by slightly different behavior on different clouds.

/lgtm
/approve
/hold

jlewi:
Removing the hold because I want this to be in the next RC. We can do a follow on PR if desired for Alibaba.

/hold cancel

Issue #288, opened 2018-02-24
Include GPU daemonset in GKE configs?


jlewi:
On GKE users need to deploy a daemonset to configure GPU nodes

```
https://cloud.google.com/kubernetes-engine/docs/concepts/gpus#installing_drivers
```

I think the main advantage is eliminating another step users have to do get a GKE cluster configured.

I'm guessing we can probably already start the daemonset as it only runs on GPU nodes so starting it on a CPU cluster is a null op.

I'm wondering if having multiple copies of the daemonet is actually problematic (e.g. 1 started manually and 1 via Kubeflow). My guess is one of them would succeed on a given node and mark the node as processed.

One problem could be is if the daemonsets are installing different versions of the drivers users might want up with a random mix of driver versions on nodes.

/cc @vishh

bhack:
Seems that we are quite stalled for other providers or very frequent Kops deployments on different providers https://github.com/kubernetes/kubernetes/issues/54011

jlewi:
Thanks @bhack.

swiftdiaries:
Just a thought, I have a script that installs drivers manually. [Install drivers](https://github.com/swiftdiaries/dl-kops/blob/master/scripts/setupgpu.sh).
Additionally, it also installs CUDA and CuDNN. 
We could version it using environment variables and containerize it to deploy it as daemon set (using ksonnet) with a default set of parameters. This could be editable using the 'ks param set', so that users can choose to install different versions if they want to. 
This can be done on any Ubuntu machine and not just GKE.
The script is a bit of a mess, but any thoughts on the idea?

bhack:
See also https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/51 and https://github.com/kubernetes/kops/pull/4971

lluunn:
Can we just always deploy the daemonset as it has tolerations on gpu:
https://github.com/GoogleCloudPlatform/container-engine-accelerators/blob/master/daemonset.yaml#L45?

bhack:
https://github.com/GoogleCloudPlatform/container-engine-accelerators/pull/51#issuecomment-399187623

lluunn:
Thanks @bhack , looks really cool.

IIUC, using this, we can just run a (driver) container and avoid installing the driver?
Is this going to be the suggested pattern?

Paste the link here:
[...]

PR #1059, opened 2018-06-21
[WIP] bootstrapper deploying gpu driver

lluunn:
For #288

### DIFF
--- a/bootstrap/cmd/bootstrap/app/server.go
+++ b/bootstrap/cmd/bootstrap/app/server.go
@@ -345,6 +345,22 @@ func appGenerate(opt *options.ServerOption, kfApp *kApp.App, fs *afero.Fs, bootC
 	return err
 }
 
+func runBashCommand(rawCmd string, description string) error {
+  applyCmd := exec.Command("bash", "-c", rawCmd)
+
+  var out bytes.Buffer
+  var stderr bytes.Buffer
+  applyCmd.Stdout = &out
+  applyCmd.Stderr = &stderr
+  if err := applyCmd.Run(); err != nil {
+    log.Infof(description + " failed: " + fmt.Sprint(err) + "\n" + out.String() + "\n" + stderr.String())
+    return err
+  } else {
+    log.Infof(description + " successfully.\n" + out.String() + "\n" + stderr.String())
+    return nil
+  }
+}
+
 // Run the tool.
 func Run(opt *options.ServerOption) error {
 	// Check if the -version flag was passed and, if so, print the version and exit.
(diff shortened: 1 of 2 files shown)


bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -538,20 +554,19 @@ func Run(opt *options.ServerOption) error {
...
-			}
+
+      if err := runBashCommand(rawCmd, "Component apply"); err != nil {
+        return err
+      }
 		}
+
+    // Deploy GKE driver daemonset.
+    err := runBashCommand(
+      "kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/stable/nvidia-driver-installer/cos/daemonset-preloaded.yaml",
```
jlewi: Can we make this a K8s package/component that the bootstrapper installs?

lluunn: Do you mean make it into a ksonnet component as part of kubeflow package?

jlewi: Yes. This way the daemonset becomes part of the ks app so you aren't dependent on bootstrapper in order to have all the configs needed for the cluster.

lluunn:
make gpu driver part of kubeflow core. PTAL

lluunn:
@kunmingg PTAL

ankushagarwal:
Shouldn't we install this only for GKE instances?

lluunn:
ah, good catch, thanks.

kubeflow/core/prototypes/all.jsonnet
```diff
@@ -22,6 +22,7 @@
 // @optionalParam usageId string unknown_cluster Optional id to use when reporting usage to kubeflow.org
 // @optionalParam tfJobVersion string v1alpha2 which version of the TFJob operator to use
 // @optionalParam centralUiImage string gcr.io/kubeflow-images-public/centraldashboard:v20180618-v0.2.0-rc.0-5-g715aafc8-e3b0c4 Image to use for Central UI.
+// @optionalParam deployGpuDriver string false Whether to deploy the GPU driver daemonset.
```
ankushagarwal: Rename to `deployGkeGpuDriver`?

lluunn: done.

jlewi:
Should we just make it is own component rather than making it part of the all component? And then in our GKE configs e.g our deploy.sh for GKE we could just install that component?

jlewi:
An even simpler solution might be to just add 
kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/stable/nvidia-driver-installer/cos/daemonset-preloaded.yaml

to 
https://github.com/kubeflow/kubeflow/blob/master/docs/gke/configs/deploy.sh

lluunn:
Good idea, thanks.

Issue #1060, opened 2018-06-21
Simplify Image Tag management for releases


jlewi:
Right now we tag our docker images with v${DATE}-${SHA}. There are two problems with this approach

1. This format isn't very informative for users
1. It's really cumbersome to update images as part of an RC

    * We need to build the images
    * Then create a PR to update the images in the ksonnet configs.

I propose we tag images with v${RELEASE} e.g. v0.2.0 and v.0.2.1 and this work as follows

1. The tags v0.2.0 will be mutable until a release candidate is promoted to actual release
    * As we cut RC's the tag for the upcoming release will be updated to point at the most recent RC
    * Once an RC is promoted to the actual release the tag is no longer mutated

1. When we cut the release branch; we update all the images to use the tag of the next release e.g. 0.2.1
     * This means we don't have to update the configs each time we cut a new image; we can just move the release tag (v0.2.1) to the new image

1. To provide an audit trail we use a simple YAML that maps immutable sha based images to tags
    e.g.

```
images:
   - image: gcr.io/some-image/image:sha256@abcd...
        tags: 
            - gcr.io/some-image/image:v0.2.0
```

1. We can use scripts to apply the tags to a given image
1. To push a new image as part of an RC we just need to move the tags by
   i. updating the YAML file
  ii.  rerunning scripts to reapply the tags

/cc @pdmack @kunmingg

pdmack:
This sounds awesome!

jlewi:
#1061 is an initial stab.

I think what we'll need some scripts to manage that YAML file for example

- Get the shas for the most recent tensorflow notebook images
- Apply some release tag to them
- Remove some release tag from some set of images.

jlewi:
Closing this issue since we've committed a bunch of scripts to make this easier and there is no immediate follow on work.

PR #1066, opened 2018-06-22
Tooling to make it easier to tag images and update the ksonnet prototypes

jlewi:
* Provide some python scripts to match images and then apply a tag to
  them like "v0.2.0". This makes it possible to easily apply a new release
  tag to set of images like the Jupyter images.

* Creeate a shell script to use sed and other twos to update images
  in our ksonnet prototypes.

* Add instructions for doing this.

* I used the scripts to add the v0.2.0 tag to our Jupyter images.

Related to #1060

/assign @pdmack

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1066)

### DIFF
--- a/docs_dev/releasing.md
+++ b/docs_dev/releasing.md
@@ -6,7 +6,7 @@
   - [Create Release Workflow](#create-release-workflow)
   - [Update Release Config](#update-release-config)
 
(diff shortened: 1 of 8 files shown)


pdmack:
love it

/approve
/lgtm

PR #1085, opened 2018-06-26
Add katib tag to images

inc0:
This change will make easier to pin Katib to any version we have images
for. Once we tag images with 0.2.0 for Katib, we should pin this to
0.2.0 in Kubeflow branch as well

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1085)

### DIFF
--- a/kubeflow/katib/modeldb.libsonnet
+++ b/kubeflow/katib/modeldb.libsonnet
@@ -218,7 +218,7 @@
                     value: "/katib",
                   },
                 ],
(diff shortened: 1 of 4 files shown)


pdmack:
@inc0 just need to format `katib/vizier.libsonnet`

jlewi:
Is there a reason you are taking the tag as separate parameter rather than just having a single parameter containing the full URL of the image?

Issue #845, opened 2018-05-22
TF-Serving 1.8 Images


jlewi:
Can we build a 1.8 version of TF Serving images

/assign @lluunn 
/priority p1

lluunn:
TF serving has not released 1.8 https://github.com/tensorflow/serving/releases

jlewi:
Still no 1.8 release.

I'll leave this on 0.2.0 release but it will get punted if TFServing doesn't do a release.

lluunn:
https://github.com/tensorflow/serving/issues/899

We should have 1.8 soon.

lluunn:
We have 1.8 now
https://github.com/tensorflow/serving/releases

lluunn:
I am having trouble building TF 1.8 gpu images.
The command: `docker build --pull --build-arg cuda_config_version=9-1 --build-arg cudnn_version=7 --build-arg tf_cuda_comp
atibility=3.5,5.2,6.1,7.0 --build-arg tf_version=1.8.0 --build-arg cuda_version=9.1 --build-arg base_image=nvidia/cuda:9.1-cudnn7-devel-ubuntu16.04 --bui
ld-arg bazel_version=0.11.1 -t gcr.io/kubeflow-images-public/tensorflow-serving-1.8gpu:latest -f Dockerfile.gpu .`

Failed with:

```
INFO|2018-06-15T11:54:11|build_image.py|46| external/org_tensorflow/tensorflow/core/kernels/cwise_ops.h(199): warning: __host__ annotation on a defaulted
 function("scalar_right") is ignored                                                                                                                     
INFO|2018-06-15T11:54:11|build_image.py|46|                                                                                                              
INFO|2018-06-15T11:54:11|build_image.py|46| external/org_tensorflow/tensorflow/core/kernels/cwise_ops.h(199): warning: __device__ annotation on a default
ed function("scalar_right") is ignored                                                                                                                   
INFO|2018-06-15T11:54:11|build_image.py|46|                                                                                                              
INFO|2018-06-15T11:54:12|build_image.py|46| Target //tensorflow_serving/model_servers:tensorflow_model_server failed to build                            
INFO|2018-06-15T11:54:12|build_image.py|46| INFO: Elapsed time: 1713.230s, Critical Path: 490.98s                                                        
INFO|2018-06-15T11:54:12|build_image.py|46| FAILED: Build did NOT complete successfully
```
[...]

Issue #1031, opened 2018-06-19
unflake TF serving testing


lluunn:
https://github.com/kubeflow/kubeflow/blob/master/testing/test_tf_serving.py#L106

We are comparing the result as text. We should compare the result numbers with some tolerance.

E.g. see #1030

PR #1090, opened 2018-06-28
[wip] TF serving 1.8 gpu

lluunn:
#845

Using this Dockerfile: https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/Dockerfile.devel-gpu

One modification: cp to /usr/bin instead of /usr/local/bin.
https://github.com/tensorflow/serving/blob/master/tensorflow_serving/tools/docker/Dockerfile.devel-gpu#L121

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1090)


lluunn:
The image is working. Just the result is a bit different #1031

PR #1091, opened 2018-06-28
Delete the tf-job package.

### LINKED ISSUE #1049: TFJob prototype's default TFVersion should be v1alpha2
TFOperator defaults to v1alpha2 but TFJob defaults to v1alpha1 
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/prototypes/tf-job-operator.jsonnet#L12
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/tf-job/tf-job.libsonnet#L47

Defaults should be same for consistency.

jlewi:
* The tf-job package only contains an outdated prototype for TFJobs. This is
  currently broken because the version was changed to v1alpha2 but the spec
  wasn't actually updated.

* Per discussions elsewhere we are moving towards structuring templates for
  TFJobs as prototypes in the examples package. We want the full spec
  to be in the prototype so that when a component is generated from the
  prototype users can easily modify it to do advanced customizations.

Fix #1049

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1091)


johnugeorge:
Can you clarify ? What is the recommended way to create a tf job now ? I
think, we still need a easy way that can be used to create a basic job.
 If we are going ahead with deleting  TFjob, we have to update readme and
user guide.

jlewi:
The preferred way to create a TFJob v1alpha2 is

```
ks generate tf-job-simple tf-job-simple
```

jlewi:
Sent kubeflow/website#71 to update the website.

jlewi:
/test all

Looks like there was a timeout waiting for workflows but the workflows actually ran too completion.

jlewi:
@kunmingg ping

kunmingg:
And the 2 config file [here](https://github.com/kubeflow/kubeflow/tree/master/bootstrap/config)

kunmingg:
Will cherry-pick to release branch once merged. Thanks!

PR #1095, opened 2018-06-28
add chainer-job/chainer-operator ksonnet package

everpeace:
this package contains three prototypes:
- `io.ksonnet.pkg.chainer-perator`: this deploys chainer-operator
- `io.ksonnet.pkg.chainer-job-simple`
  - this is intended to use as quick start.  this supports only single pod job.
- `io.ksonnet.pkg.chainer-job`
  - this is standard chainer job prototype.  this supports non-distributed and distributed jobs

fixes kubeflow/chainer-operator#12

/cc @jlewi @rongou  @gaocegege @alsrgv

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1095)


kubeflow/chainer-job/README.md
```diff
@@ -0,0 +1,104 @@
...
+
+If you haven't yet created a [ksonnet application](https://ksonnet.io/docs/tutorial#1-initialize-your-app), do so using `ks init <app-name>`.
+
+Finally, in the ksonnet application directory, run the following:
+
+```shell
+# Expand prototype as a Jsonnet file, place in a file in the
+# `components/` directory. (YAML and JSON are also available.)
+$ ks pkg install kubeflow/chainer-job@master
+$ ks prototype use io.ksonnet.pkg.chainer-operator chainer-operator \
```
rongou: The kubeflow docs mostly use `ks generate` which is just an alias of `ks prototype use`.

kubeflow/chainer-job/README.md
```diff
@@ -0,0 +1,104 @@
...
+  --namespace default \
+  --name chainer-operator
+
+$ ks prototype use io.ksonnet.pkg.chainer-job chainer-job-simple \
+  --namespace default \
```
rongou: I think the `namespace` is now inherited from the environment, so probably can't set it here any more?

kubeflow/chainer-job/chainer-job.libsonnet
```diff
@@ -0,0 +1,127 @@
+local k = import "k.libsonnet";
```
jlewi: We are moving in the direction of putting the entire manifest in the jsonnet prototype as opposed to splitting it between a libsonnet and jsonnet file.See
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/examples/prototypes/tf-job-simple-v1alpha1.jsonnet

This way the result of ks generate is a jsonnet file that contains the complete manifest that users can easily modify. 
(i.e ks generate is just fancy copy and paste).

We are moving away from using complex jsonnet features to parameterize templates in complex ways because this hurts readability.

kubeflow/chainer-job/README.md
everpeace: Thanks for pointing out!  I will change this.

kubeflow/chainer-job/README.md
everpeace: you're right.  I will remove this.

kubeflow/chainer-job/chainer-job.libsonnet
everpeace: Thanks for letting me know the direction!  I will change it 👍

kubeflow/chainer-job/README.md
```diff
@@ -0,0 +1,104 @@
...
+#### Parameters
+
+```shell
+# This outputs available parameters
+$ ks prototype describe io.ksonnet.pkg.chainer-job-simple
+```
+
+### io.ksonnet.pkg.chainer-job
+
+A [Chainer Job](https://github.com/kubeflow/chainer-operator) which supporte distributed and non-distributed jobs.
```
disktnk: supports

everpeace:
@rongou @jlewi @disktnk Thank you for your review.  I fixed about reviewed comments.  Please take a look again 🙇 

@jlewi I simplified `chainer-job` and `chainer-job-simple` prototypes and remains `libsonnet` for `chainer-operator` because I think this prototype would not have many chances for users to need to change.

jlewi:
/approve

For thinks  like the operator using a libsonnet file is the right thing to do.

Issue #574, opened 2018-04-03
TFJobs UI doesn't work behind IAP; React APP needs support IAP?


jlewi:
TFJobs UI is deployed on dev.kubeflow.org.

The UI shows up behind IAP but its doesn't work
- No TFJobs are listed
- Creating a job via the UI doesn't work.

Looking at the developer console we see requests to

```
https://accounts.google.com/o/oauth2/v2/auth?client_id=235037502967-9cpmvs4ljbiqb3ojtnhnhlkkd8d562rl.apps.googleusercontent.com&response_type=code&scope=openid+email&redirect_uri=https://dev.kubeflow.org/_gcp_gatekeeper/authenticate&state=Ci1odHRwczovL2Rldi5rdWJlZmxvdy5vcmcvdGZqb2JzL2FwaS9uYW1lc3BhY2USMEFOa2F0U2dYVjdSYnlHMzVXeGFwR1gxNURTS29TNjNiYnc6MTUyMjc4MDQxODEwOA
```

Which suggests to me the request is hitting the loadbalancer and being directed to do auth verification to sign in and its getting rejected.

So I think one of two things is happening

1. The request is coming from the server running in K8s and incorrectly being redirected to the external loadbalncer and thus hitting IAP when it shouldn't be
1. The request is coming from the client and the client needs to be updated to support IAP.

@wbuchwalter Do you know where the request is coming from?

You should be able to access it at
https://dev.kubeflow.org/tfjobs/ui/

wbuchwalter:
@jlewi @ankushagarwal Could one of you grant me access to dev.kubeflow.org so I can test this?
`wbuchwalter at gmail`.
Thanks!

@jlewi This is most likely `2`

jlewi:
I think @wbuchwalter started looking at this last week.

jlewi:
@wbuchwalter any update?

jlewi:
Punting to 0.3 because no one is actively working on this so I don't think its going to make the cut.

kkasravi:
@jlewi I can look at this - given my recent wrestling with jupyterhub I think it should be straigtforward.

jlewi:
@kkasravi That would be fantastic. 

Related; it looks like recent changes to the central UI also created problems for IAP (kubeflow/kubeflow#957)

I'd like to better understand the proper way to write web apps that are compatible with reverse proxies and IAP.

jlewi:
There's some guidance here about handling IAP refresh
[...]

PR #1097, opened 2018-06-28
Create a version of echo-server to echo headers.

jlewi:
* This is intended to support debugging IAP; we want to see what headers
  are on resulting requests.

* See kubeflow/kubeflow#574

* While creating this I ran into an issue with ksonnet not formatting the
  Ambassador mapping correctly unless we import it from a libsonnet file see
  ksonnet/ksonnet#670

/assign @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1097)

### DIFF
--- a/components/echo-server/Makefile
+++ b/components/echo-server/Makefile
@@ -16,9 +16,20 @@
 #   https://github.com/mattrobenolt/jinja2-cli
 #   pip install jinja2-clie
 IMG = gcr.io/kubeflow-images-staging/echo-server
-
-TAG := $(shell date +v%Y%m%d)-$(shell git describe --always --dirty)-$(shell git diff | shasum -a256 | cut -c -6)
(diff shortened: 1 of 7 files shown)


components/echo-server/Makefile
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Requirements:
+#   https://github.com/mattrobenolt/jinja2-cli
+#   pip install jinja2-clie
+IMG = gcr.io/kubeflow-images-staging/echo-server
```
ankushagarwal: kubeflow-images-public?

PR #1098, opened 2018-06-28
add seldon to ks config; pre install all pkg

kunmingg:
To smooth user experience, including https://github.com/kubeflow/examples

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1098)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/server.go
+++ b/bootstrap/cmd/bootstrap/app/server.go
@@ -95,20 +95,30 @@ type BootConfig struct {
 	App  AppConfig
 }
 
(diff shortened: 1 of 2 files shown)


ankushagarwal:
Does this mean that we no longer need to specify packages in the config file : https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/default.yaml#L5 ?

kunmingg:
@ankushagarwal 
Yes
I didn't delete packages config in this commit to avoid backward compatibility problems.

PR #1099, opened 2018-06-28
Tfjob create fails in tfjob UI

kkasravi:
See https://github.com/kubeflow/tf-operator/issues/701

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1099)

### DIFF
--- a/bootstrap/image_registries.yaml
+++ b/bootstrap/image_registries.yaml
@@ -4,6 +4,6 @@
 registries:
   - name: kubeflow
     repo: https://github.com/kkasravi/kubeflow.git
-    branch: tfjobs_iap
+    branch: tfjob_create
     # relative path to registry from git repo.
     path: kubeflow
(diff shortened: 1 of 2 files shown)


bootstrap/image_registries.yaml
```diff
@@ -4,6 +4,6 @@
 registries:
   - name: kubeflow
```
jlewi: Why are you upgrading this file? Can you revert it back to the original value and not use kkassarvi?

Here's the process I follow to do debugging

1. Follow kubeflow.org to deploy on GCP
2. Run this command to copy your ksonnet app to your local machine
https://www.kubeflow.org/docs/started/getting-started-gke/
3. Replace vendor/kubeflow with a symbolic link to your code; this way you will pick up changes to your ksonnet component
4. Recreate the K8s resources to pick up your changes.

jlewi:
Thanks.

/lgtm
/approve

PR #1100, opened 2018-06-29
Create a script to update a ksonnet app to the latest Kubeflow package

jlewi:
* Related to: kubeflow/website#57
* Based on the script in ksonnet/ksonnet#237

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1100)

### DIFF
--- a/scripts/upgrade_ks_app.py
+++ b/scripts/upgrade_ks_app.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python
+import argparse
+import logging
+import os.path
+import shutil
+import subprocess
+import yaml
+
+# The path used by the bootstrapper
+BOOTSTRAPPER_REGISTRY = "/opt/registries/kubeflow/kubeflow"
+
+# The current release of Kubeflow. This should be upgraded on every release.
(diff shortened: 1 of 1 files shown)


jlewi:
Looks like a test flake
/test all

scripts/upgrade_ks_app.py
```diff
@@ -0,0 +1,95 @@
...
+
+  with open('app.yaml', 'w') as f:
+    yaml.dump(app, f, default_flow_style=False)
+
+  for name, registry in registries.iteritems():
+    if name != registry_name:
+      logging.info("Skipping registry %s", name)
+      continue
+    logging.info("Adding registry %s point at %s", name, registry_url)
+    subprocess.call(['ks', 'registry', 'add', name, registry_url])
```
kunmingg: We need to either delete old registry info from app.yaml or use a new name like ```kubeflow-0.2.0```.
Otherwise ks will complain when add new version registry with same name.

jlewi:
@kunmingg PTAL. We were deleting entries in app.yaml but there were bugs in the logic; we were deleting all entries and not just the registry being upgraded.

Issue #1068, opened 2018-06-22
[gcp] GKE setup; do as much as possible in deploy.sh


jlewi:
Our getting started instructions: https://www.kubeflow.org/docs/started/getting-started-gke/

Depend on a deploy.sh script; but there are still other steps in the process that user has to do manually.
Since we already have deploy.sh; why not move more of the commands into that script and try to infer sensible defaults using gcloud.

Specifically

1. deploy.sh can download the DM configs

1. deploy.sh can modify the DM configs
    * Set project and endpoints name
    * Set zone
    * Set email using gcloud auth config list

1. Create K8s resources including the bootstrapper

    * I think we should stop creating K8s resources with DM
    * We already issue kubectl commands in deploy.sh we might as well move any other resources there
    * Creating K8s resources in DM is weird because you probably don't want to update/manage ks resources using DM

1. Download the ksonnet app from the bootstrapper

     ```
     kubectl cp kubeflow-admin/kubeflow-bootstrapper-0:/opt/bootstrap/default ~/kf-app
     ```

1. deploy.sh can create clusterrolebindings to give user necessary K8s permissions
1. deploy.sh should trigger the creation of KF K8s resource creation (rather than having bootstrapper do this automatically)

   * I think its helpful if a user can see all the steps in deploy.sh 
   * Deploy.sh can trigger ks apply doing

   ```
   kubectl exec kubeflow-admin kubeflow-bootstrapper-0 cd /opt/bootstrapper/default && ks apply ...
   ```
   * We can have deploy.sh pass a bearer token to the bootstrapper so bootstrapper doesn't need elevated permissions

/cc @kunmingg @aronchick @ankushagarwal

kunmingg:
Dependency: https://github.com/kubeflow/kubeflow/pull/1044?

PR #1111, opened 2018-06-30
Create a deployment script for gke and minikube

ankushagarwal:
* Move away from using bootstrapper by doing everything locally
* Try to use gcloud to infer as much as possible (PROJECT, ZONE, EMAIL, etc.)
* Delete k8s resources from DM configs
* Require prerequisite of gcloud, kubectl and ks

TODO(ankushagarwal): Add integration test for this. 
TODO(ankushagarwal): Delete https://github.com/kubeflow/kubeflow/tree/master/docs/gke/configs

Fixes https://github.com/kubeflow/kubeflow/issues/1068

After this PR the getting started experience for kubeflow on GKE:
```
git clone https://github.com/kubeflow/kubeflow/
cd kubeflow
git checkout v0.2.0-rc.1
# export CLIENT_ID and CLIENT_SECRET
./scripts/gke/deploy.sh
```
We can get rid of `CLIENT_ID and CLIENT_SECRET` step soon

After this the getting started experience for kubeflow on GKE:
```
git clone https://github.com/kubeflow/kubeflow/
cd kubeflow
git checkout v0.2.0-rc.1
./scripts/minikube/deploy.sh
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1111)

### DIFF
--- a/scripts/gke/deploy.sh
+++ b/scripts/gke/deploy.sh
@@ -1,0 +1,133 @@
+#!/bin/bash
+# This script creates a kubeflow deployment on GCP
+# It checks for kubectl, gcloud, ks
+# Uses default PROJECT, ZONE, EMAIL from gcloud config
+# Creates a deployment manager config copy and edits appropriate values
+# Adds user to the IAP role
+# Creates the deployment
+# Creates the ksonnet app, installs packages, components and then applies them
+
+set -xe
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${SCRIPT_DIR}/../util.sh"
+KUBEFLOW_REPO=$(cd "${SCRIPT_DIR}/../.."; pwd)
+check_install gcloud
+check_install kubectl
+# TODO(ankushagarwal): verify ks version is higher than 0.11.0
+check_install ks
+
+check_variable "${CLIENT_ID}" "CLIENT_ID"
+check_variable "${CLIENT_SECRET}" "CLIENT_SECRET"
+# Name of the deployment
+DEPLOYMENT_NAME=${DEPLOYMENT_NAME:-"kubeflow"}
+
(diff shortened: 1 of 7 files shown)


scripts/gke/deploy.sh
```diff
@@ -0,0 +1,132 @@
...
+# Creates the ksonnet app, installs packages, components and then applies them
+
+set -xe
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "${SCRIPT_DIR}/../util.sh"
+KUBEFLOW_REPO=$(cd "${SCRIPT_DIR}/../.."; pwd)
+check_install gcloud
+check_install kubectl
+check_install ks
```
kunmingg: Might want to check ks version since only new one support local registry.

ankushagarwal: This is non-trivial to do in bash. Added a TODO. We will mention in the docs that the requirements are `ks > 0.11`

jlewi:
Can we move the clone into the deploy script?

So the experience is something like

```
curl https://raw.githubusercontent.com/kubeflow/kubeflow/master/docs/gke/configs/deploy.sh | bash
```

jlewi:
This is fantastic. Thank you.

scripts/gke/deploy.sh
```diff
@@ -0,0 +1,143 @@
...
+cd $(dirname "${KUBEFLOW_KS_DIR}")
+ks init $(basename "${KUBEFLOW_KS_DIR}")
+cd "${KUBEFLOW_KS_DIR}"
+
+ks env set default --namespace "${K8S_NAMESPACE}"
+# Add the local registry
+ks registry add kubeflow "${KUBEFLOW_REPO}/kubeflow"
+
+# Install all required packages
+ks pkg install kubeflow/core
```
jlewi: Should we make this a function so that other deploy scripts could use it?

PR #1119, opened 2018-07-03
Put the full PyTorch prototype in the jsonnet file.

### LINKED ISSUE #1114: PyTorch job prototype should contain the full job spec
The PyTorch job prototype
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/pytorch-job/prototypes/pytorch-job.jsonnet

Currently defines the spec in a .libsonnet file which is imported.

We want to move the complete spec into the jsonnet file. This way when a user generates a component they have complete access to the spec and can easily modify it.

This is the pattern we are now following with other spec e.g.

jlewi:
* The current pattern in Kubeflow is to put the complete prototype in
  the jsonnet file. This way the result of ks generate is a .jsonnet
  file containing a full spec. This makes it easy for users to do
  complex modifications starting with the prototype as an example.

* Configure the prototype to do mnist by default.

Fix #1114

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1119)


ankushagarwal:
Cleaned up some old test gpu pods from the cluster. tests should pass now

Issue #955, opened 2018-06-08
[gcp] monitoring agent need to emit events/status information particularly related to IAP


jlewi:
It can take a while for ingress + IAP to be setup

We need to wait for a whole host of events

   * Lets encrypt gets certificate
   * Cloud Endpoints Registers Hostname
   * Loadbalancer reports the backend as healthy

We need a way to provide status to the user to let them know when things are finally ready. 

The first step is probably to just document how to monitor the individual steps. We could then create an agent (perhaps the iap_enabler.sh) which monitors the various services and emits appropriate status information.

One way to emit status information would be K8s events.

Ideally we would emit a very clear event when things are ready.

jlewi:
Lets try to create a simple agent that just emits K8s events.

jlewi:
One thing we could do is just write a health check agent that pings the central UI on its public endpoint using IAP.

PR #1123, opened 2018-07-03
add service monitor prototype for monitoring deployment status

kunmingg:
fix https://github.com/kubeflow/kubeflow/issues/955

Provide stackdriver/prometheus compatible data exporter, monitoring kubeflow status after deployment: 
the signal ```kubeflow_availability==1``` means https url behind IAP can be logged in with credentials.

How to integrate:
1. directly access port 8000 on pod.
2. read through stackdriver (if beta feature turned on for cluster).
3. read through prometheus engine (if deployed in cluster).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1123)

### DIFF
--- a/metric-collector/service-readiness/kubeflow-readiness.py
+++ b/metric-collector/service-readiness/kubeflow-readiness.py
@@ -1,0 +1,140 @@
+import argparse
+from time import sleep, time
+
+import google.auth
+import google.auth.app_engine
+import google.auth.compute_engine.credentials
+import google.auth.iam
+from google.auth.transport.requests import Request
+import google.oauth2.credentials
+import google.oauth2.service_account
+from prometheus_client import start_http_server, Gauge
+import requests
+
+IAM_SCOPE = 'https://www.googleapis.com/auth/iam'
+OAUTH_TOKEN_URI = 'https://www.googleapis.com/oauth2/v4/token'
+METHOD = 'GET'
+KUBEFLOW_AVAILABILITY = Gauge('kubeflow_availability', 'Signal of whether IAP protected kubeflow is available')
+
+def metric_update(args, google_open_id_connect_token):
+  resp = requests.request(
+    METHOD, args.url,
+    headers={'Authorization': 'Bearer {}'.format(
+      google_open_id_connect_token)})
+  if resp.status_code == 200:
+    KUBEFLOW_AVAILABILITY.set(1)
+  else:
+    KUBEFLOW_AVAILABILITY.set(0)
+
+def main(unparsed_args=None):
+  parser = argparse.ArgumentParser(
+    description="Output signal of kubeflow service readiness.")
+
+  parser.add_argument(
+    "--url",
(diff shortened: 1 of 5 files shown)


metric-collector/service-readiness/kubeflow-readiness.py
```diff
@@ -0,0 +1,140 @@
...
+import google.auth.app_engine
+import google.auth.compute_engine.credentials
+import google.auth.iam
+from google.auth.transport.requests import Request
+import google.oauth2.credentials
+import google.oauth2.service_account
+from prometheus_client import start_http_server, Gauge
+import requests
+
+IAM_SCOPE = 'https://www.googleapis.com/auth/iam'
```
jlewi: Why do we need to hard code these URLs aren't they part of the client library?

metric-collector/service-readiness/kubeflow-readiness.py
```diff
@@ -0,0 +1,140 @@
...
+  # Construct OAuth 2.0 service account credentials using the signer
+  # and email acquired from the bootstrap credentials.
+  service_account_credentials = google.oauth2.service_account.Credentials(
+    signer, signer_email, token_uri=OAUTH_TOKEN_URI, additional_claims={
+      'target_audience': args.client_id
+    })
+
+  token_refresh_time = 0
+  while True:
+    if time() > token_refresh_time:
```
jlewi: Either in this PR or in a follow on PR; could we emit Kubernetes events as well indicating the status of the endpoint.

metric-collector/service-readiness/kubeflow-readiness.py
```diff
@@ -0,0 +1,140 @@
...
+
+  args = parser.parse_args(args=unparsed_args)
+
+  if args.url == "":
+    sleep(2000)
+    return
+
+  # Figure out what environment we're running in and get some preliminary
+  # information about the service account.
+  bootstrap_credentials, _ = google.auth.default(
```
jlewi: nit: Why did you call it bootstrap_credentials and not just credentials? What is bootstrapping referring to?

metric-collector/service-readiness/kubeflow-readiness.py
kunmingg: Surprisingly seems all user guide use hard code URL directly.

metric-collector/service-readiness/kubeflow-readiness.py
kunmingg: For all service / deployment, will fetch with prometheus and store in stackdriver by default.
Will do in separate PR.

metric-collector/service-readiness/kubeflow-readiness.py
kunmingg: This name was used in IAP code examples, I think they were referring to fetch credential aggressively from environment and use whatever available (gcloud / service account key / ...).

Issue #1135, opened 2018-07-06
Docker image building workflows are failing 


jlewi:
See #1130 

It looks like the latest runs of most of the release workflows for our images failed; e.g. pytorch operator; tf-serving-release, notebook-release.

jlewi:
It looks to me like the dind sidecar got a sig term.

jlewi:
main container logs

```
gcloud beta logging read --project=${PROJECT} --freshness=1d "resource.labels.cluster_name=\"${CLUSTER}\"  resource.labels.container_name=\"${CONTAINER}\" resource.labels.pod_id=\"${POD_ID}\"" --format="table(timestamp,textPayload)" --order=asc
TIMESTAMP             TEXT_PAYLOAD
2018-07-06T03:43:49Z  + DOCKERFILE=/mnt/test-data-volume/image-release-pytorch-operator-release-a50c2fb-3dd6-1f9d/src/kubeflow/pytorch-operator/Dockerfile
2018-07-06T03:43:49Z  ++ dirname /mnt/test-data-volume/image-release-pytorch-operator-release-a50c2fb-3dd6-1f9d/src/kubeflow/pytorch-operator/Dockerfile
2018-07-06T03:43:49Z  + CONTEXT_DIR=/mnt/test-data-volume/image-release-pytorch-operator-release-a50c2fb-3dd6-1f9d/src/kubeflow/pytorch-operator
2018-07-06T03:43:49Z  + IMAGE=gcr.io/kubeflow-images-public/pytorch-operator
2018-07-06T03:43:49Z  + TAG=v20180706-a50c2fb0
2018-07-06T03:43:49Z  + ROOT_DIR=/mnt
2018-07-06T03:43:49Z  + export GOPATH=/mnt
2018-07-06T03:43:49Z  + GOPATH=/mnt
2018-07-06T03:43:49Z  + GO_DIR=/mnt/src/github.com/kubeflow/pytorch-operator
2018-07-06T03:43:49Z  + gcloud auth activate-service-account --key-file=/secret/gcp-credentials/key.json
2018-07-06T03:43:51Z  Activated service account credentials for: [kubeflow-releasing@kubeflow-releasing.iam.gserviceaccount.com]
[...]

Issue #1132, opened 2018-07-06
Building the Jupyter images took over 4 hours


jlewi:
I did a run of the workflow to build the Jupyter images. This took over 4 hours. The full workflow spec is below. This seems unreasonably slow.

```
 startedAt: 2018-07-05T17:45:30Z
 finishedAt: 2018-07-05T22:06:08Z
```
```
apiVersion: argoproj.io/v1alpha1
kind: Workflow
metadata:
  annotations:
[...]

PR #1136, opened 2018-07-06
Set requests and limits for RAM and CPU in TF notebook image releaser.

jlewi:
* See if this makes the build more reliable and faster

* We should really be setting disk space because it looks like (see #1135)
  that is one resource that is under pressure. Need to figure out
  if that is easy to control with K8s.

Related to:
  #1135 Docker image building workflows are failing
  #1132 Building Jupyter images took over 4 hours.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1136)


jlewi:
Test failure looks like:
kubeflow/testing#169

jlewi:
kubeflow/testing#172 should increase the timeout to 20 minutes; so lets see if that fixes it.

/test all

PR #1151, opened 2018-07-08
Turn bootstrapper into an RPC server for ksonnet.

### LINKED ISSUE #962: Bootstrapper should support file and http registries in a consistent manner
See [comment](https://github.com/kubeflow/kubeflow/pull/951#issuecomment-395930919) in #951.

Here's how I think specification of registries in the bootstrapper config should work

We should be using ksonnet's support for file registries
User can explicitly refer to a registry e.g
name: kubeflow
registry: file://opt/registries kubeflow
or
name: kubeflow
registry: http://github.com/kubeflow/kubeflow

http registries should always pull from the remote repository. We shouldn't automagically map "http" to a local "file"

For convenience if user doesn't specify registry, the bootstrapper will use the name of the registry to map it to the local file registry and use that.

Note: a GitHub token isn't necessarily required to use remote repositories; in many cases the default quota should be sufficient.

jlewi:
We have two use cases.

We want to create a click to deploy web app. This will be a client side
javascript app. This app needs a way to run ksonnet. We will do that
by deploying the bootstrapper on the K8s cluster and then sending RPCs.

Using RPCs will allow the web app to send a bearer token containing user
credentials to do authorization so that we don't need to give the RPC
elevated permissions.

We also want a microservice to do ksonnet substitution that can be used
with hyperparameter tuning to fill in the hyperparamters into a template.

Fix #962 handle file and http registries in a consistent manner.

  * The AppConfig can now specify registries consisting of a (name, URI)
    pair. URI can be anything ks registry add supports (e.g. file or git link).

  * Registries that have been prefetched to local directories are specified
    separately

  * In AppConfig if a registry doesn't specify a URI, the server will
    try to match that registry to one of the known regstries based on
    the name. This allows us to avoid needing a GitHub token
    by using prefetched repositories.

* Delete the code to create the namespace and RBAC roles.

  * This is a legacy of when bootstrapper was running locally in
    a docker container.

    * The code to create the RBAC role was only invoked when running locally
      and not in cluster; if its running in cluster some process would
      have had to deploy the bootstrapper on the cluster and that
      process presumably already has credentials to deploy it.

  * Kubeflow setup now has two modes

    1. Running a deploy.sh script locally; the namespace and rbac
       roles should be created by the deploy script

       The deploy scripts most likely won't use the bootstrapper; there's
       not much reason to.

    2. Click to deploy

      * We use bootstrapper as a ksonnet server since the webapp can't
        otherwise run ksonnet.

      * The webapp however can create the minimal K8s resources needed
        to run the bootstrapper e.g. RBAC roles, namespace etc...

* Add a script to manually test sending requests to the server.

* Add script to run gofmt and goimports on all files.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1151)

### DIFF
--- a/bootstrap/Dockerfile
+++ b/bootstrap/Dockerfile
@@ -1,6 +1,6 @@
 # bootstrapper-builder contains the tool chain we need to build the image.
 # This primarily means glide.
-FROM gcr.io/kubeflow-images-public/bootstrapper-builder:v20180709-2ef5fe7a-dirty-39cf08 as builder
+FROM gcr.io/kubeflow-images-public/bootstrapper-builder:v20180712-4adef8c0 as builder
 
 RUN mkdir -p /opt/kubeflow
 
@@ -17,7 +17,7 @@ RUN cd $GOPATH/src/github.com/kubeflow/kubeflow/bootstrap/ && \
 
 RUN go build -i -o /opt/kubeflow/bootstrapper ${GOPATH}/src/github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/main.go
 
-FROM golang:1.8.2
+FROM golang:1.10.3
 
 # We need gcloud to get gke credentials.
 RUN wget -q https://dl.google.com/dl/cloudsdk/channels/rapid/google-cloud-sdk.tar.gz && \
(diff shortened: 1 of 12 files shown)


jlewi:
Tests are passing. This is now ready for review.

/assign @kunmingg

jlewi:
We setup from WebApp the sequence will be

* Make request to DM to create GKE cluster
* Call K8s to deploy the bootstrapper/ksonnet server
* Call the bootstrap service via the K8s APIServer proxy to create and deploy Kubeflow

Making it a CRD is one possibility but I think we'll have to see whether that makes sense. It depends on how we want to integrate it into other services like Katib.

PR #1152, opened 2018-07-09
Don't check in bootstrap/vendor.

jlewi:
* This bloats the repository size by 150M +.
* Since we want to be able to quickly clone the repo using to facilitate
  getting our ksonnet registry; we want to avoid bloating the repository.

* We create a builder image which is Go + glide since we don't introduce
  flakiness trying to fetch glide.

Fix: 1051

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1152)


jlewi:
Marking this as WIP until tests pass.

PR #1159, opened 2018-07-10
Create a doc to describe the deployment process.

jlewi:
* This is not a proposal but a description of the current state.

/assign @ankushagarwal 
/assign @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1159)

### DIFF
--- a/docs_dev/kubeflow_deployment.md
+++ b/docs_dev/kubeflow_deployment.md
@@ -1,0 +1,282 @@
+# Deploying Kubeflow
+
+Authors:
+
+  * jlewi@google.com
+
+Status:
+
+  * 2018-07-08 Created
+
+## TL;DR
+
+This document describes how we are approaching the problem of deploying Kubeflow.
+
+This document is primarily intended for Kubeflow contributors and in particular
+contributors interested in
+
+- Improving the user experience
+- Creating an optimized experience for a particular platform.
+
+## Goals For Deploying Kubeflow
+
+Kubeflow is a Kubernetes native platform for machine learning.
+When it comes to deploying Kubeflow the primary goals are
+
+1. Creating a very simple getting started experience 
+1. Allow Kubeflow to be customized and optimized for different platforms
+
+## Glossary
+
+Platform - We use platform to refer to the entirety of a user's solution. This
+  includes not only Kubeflow and Kubernetes distribution but supporting services
+  that may be external to Kubernetes (e.g. S3/GCS).
+
+## Getting Started Experience: one click or one command
+
+With our 0.3 release we'd like to offer users a one command or one click 
+experience.
+
+1. User deploys Kubeflow on their platform using one click or one command
+
+   * Using the command line (with required tools preinstalled) the experience should 
+     be something like
+
+     ```
+     curl https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh | bash
+     ```
+
+   
+   * Alternatively, we'd like to offer a web-app that allows deployments with one click
+
+     * A web-app creates an opportunity to offer a wizard that can help walk users through 
+       the setup process
+
+     * A web-app can also eliminate the need to install any tools client side
+
+
+   * We'd like to offer scripts for various platforms including major clouds as well
+     as solutions for deploying Kubeflow locally on one's machine (e.g. docker, minikube, microk8s) 
+
+   * The script should take care of setting up Kubernetes and supporting services
+     for each platform
+
+1. Scripts/users use standard K8s patterns to determine when Kubeflow is ready
+
+   ```
+   kubectl get application kubeflow
+   ```
+
+   * We use the sig apps [application resource](https://github.com/kubernetes-sigs/application)
+     to represent Kubeflow and provide status information
+
+   * K8s events will be used to surface information about the deployment process.
+
+1. Users can easily navigate to JupyterLab after deploying Kubeflow
+
+   * JupyterLab provides a suitable environment for running Kubeflow codelabs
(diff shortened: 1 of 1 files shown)


docs_dev/kubeflow_deployment.md
```diff
@@ -0,0 +1,261 @@
...
+One of the goals of Kubeflow is to run anywhere Kubernetes runs. 
+**However we also want to optimize Kubeflow to take advantage of features in
+different K8s versions and platforms.**
+
+We've tried a variety of approaches but the approach that is gaining traction
+is
+
+1. We provide simple, platform deployment scripts like this [one for GKE](https://github.com/kubeflow/kubeflow/blob/master/docs/gke/configs/deploy.sh)
+
+1. A corresponding platform specific getting started page ([see here]](https://github.com/kubeflow/website/tree/master/content/docs/started) provides platform specific instructions
```
ankushagarwal: Typo near `see here`

docs_dev/kubeflow_deployment.md
```diff
@@ -0,0 +1,261 @@
...
+One of the goals of Kubeflow is to run anywhere Kubernetes runs. 
+**However we also want to optimize Kubeflow to take advantage of features in
+different K8s versions and platforms.**
+
+We've tried a variety of approaches but the approach that is gaining traction
+is
+
+1. We provide simple, platform deployment scripts like this [one for GKE](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh)
+
+1. A corresponding platform specific getting started page ([see here]](https://github.com/kubeflow/website/tree/master/content/docs/started) provides platform specific instructions
```
ankushagarwal: Replace `[see here]]` with `[see here]`

docs_dev/kubeflow_deployment.md
```diff
@@ -0,0 +1,282 @@
...
+
+So far this idea hasn't gained much traction. 
+
+Most of the setup to date has focused on users starting from scratch. In this situation we 
+have complete control of the setup so there's no reason to optimize with respect to an
+existing deployment.
+
+The scaffolding/prototype for this (bootstrapper) is still in place and we haven't
+rejected this idea completely. So contributions pursuing this idea further would
+be welcome.
```
kkasravi: @jlewi I liked the design of the bootstrapper running in the 'kubeflow' namespace with elevated privileges. It did seem to have a limitation of only creating/deploying one kubeflow instance/namespace with no specific rbac mapping to users in that namespace. i wonder if this design might be continued with a local command communicating with the bootstrapper for create, delete, list, add <user/group>, and other subcommands. The bootstrapper would leverage the `application` as a bundling construct for different types of applications that have different users, cloud-providers and components.

docs_dev/kubeflow_deployment.md
```diff
@@ -0,0 +1,282 @@
...
+We've tried a variety of approaches but the approach that is gaining traction
+is
+
+1. We provide simple, platform deployment scripts like this [one for GKE](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh)
+
+1. A corresponding platform specific getting started page ([see here](https://github.com/kubeflow/website/tree/master/content/docs/started) provides platform specific instructions
+
+Here are some **guidelines (not requirements)** for creating the above scripts and instructions
+
+* Platform scripts should assume users are starting from scratch
```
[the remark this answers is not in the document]
jlewi: See comment below about the scripts being linear. My thinking is that providing a single script optimizes for getting started "one command". If the scripts are linear it should be pretty straightforward.

If we can have a module approach where the scripts are split up into separate scripts that can then be called from a single uber script that might be nice too. That would probably help with code reuse as well. But I also don't want to make our scripts super complicated.

docs_dev/kubeflow_deployment.md
```diff
@@ -0,0 +1,282 @@
...
+Monitoring deployment is an open area. The initial focus is:
+
+  1. Informing the user when Kubeflow is fully deployed and ready to use
+  1. Surfacing relevant events (e.g. as K8s events)
+
+For more info please take a look at some of the open issues and consider
+opening more. 
+
+The current thinking is to follow the guidance of sig-apps and use an [application resource](https://github.com/kubernetes-sigs/application)
+to represent Kubeflow and attach events, status, and other metrics to that application as appropriate.
```
kkasravi: In addition to the application crd noted above - could the [SecurityProfile](https://github.com/easeway/community/blob/a8cb2060dc621664c86b185c7426367994b181b5/keps/draft-20180418-security-profile.md) also be leveraged?

jlewi: Assuming you mean for security purposes (as opposed to monitoring) then I think so.  It looks like SecurityProfile is still very early though and not even supported by sig-apps yet.

kkasravi: yes. It is early but it seems like it could be combined with application resource - where a UI could allow different components and security profiles to be bundled and deployed. 

Is the bootstrapper assuming that only one kubeflow instance would be deployed or many? EG:
- kubeflow-admin
above privileged namespace could deploy kubeflow 'instances' like below and have different directories (that could include gitops) for each deployment
- kubeflow-teamX
- kubeflow-datascientistY
- kubeflow-pytorch-gpu-1

jlewi:
@kkasravi Would you mind LGTM'ing since you're the only who provide significant feedback? We can continue to discuss/iterate but it would be good to get it checked in.

/assign @kkasravi

jlewi:
Rebase'd to pick up improvements to the tests.

PR #1172, opened 2018-07-11
Make GKE VM service account storage.objectViewer to have read access of gcr #1164

jlewi:
Cherry pick #1164

* Sync the contents of scripts/gke to what's on upstream/master because
  we want to use the new deploy.sh script with v0.2. This doesn't depdend
  on the bootstrapper.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1172)


jlewi:
Merging manually since CLA check is an issue.

PR #1173, opened 2018-07-11
GCFS support

ankushagarwal:
GCFS support in GKE deploy script. Running `deploy.sh` will create a GCFS instance and make it available in jupyter notebooks spawned by jupyterhub

* remove nfs.libsonnet as it relied on GCP PDs for NFS
* Create a ksonnet prototype: google-cloud-filestore-pv
  * This creates the PV and PVC for GCFS
* Update deploy.sh to create GCFS instance and mount it in Jupyter Notebooks via disks param
* refactor jupyterhub.libsonnet and move logic into spawner.py

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1173)


kubeflow/core/jupyterhub.libsonnet
```diff
@@ -220,6 +157,19 @@ c.RemoteUserAuthenticator.header_name = 'x-goog-authenticated-user-email'",
...
+                    name: "KF_AUTHENTICATOR",
+                    value: jupyterHubAuthenticator,
+                  },
+                  {
+                    name: "KF_PVC_LIST",
+                    value: disks,
+                  },
+                  if cloud == "gke" then
+                    {
+                      name: "GCP_SECRET_NAME",
```
jlewi: Why does JupyterHub need a GCP secret to add PVs to pods?

jlewi: Is this to add the GCP secret to the JupyterPods?

kubeflow/core/kubeform_spawner.py
```diff
@@ -102,6 +112,7 @@ def extra_resource_limits(self):
 repoName = os.environ.get('REPO_NAME')
 c.JupyterHub.spawner_class = KubeFormSpawner
 c.KubeSpawner.singleuser_image_spec = '{0}/{1}/tensorflow-notebook'.format(registry, repoName)
+c.KubeSpawner.image_spec = '{0}/{1}/tensorflow-notebook'.format(registry, repoName)
```
jlewi: What's this for?

jlewi:
Good stuff.

kubeflow/core/jupyterhub.libsonnet
ankushagarwal: This secret is mounted to all jupyter notebook pods. It is not used by jupyterhub.

kubeflow/core/kubeform_spawner.py
ankushagarwal: I should've clarified this. Some fields have been deprecated in `KubeSpawner` class[1]. I've added the new field names so that this config file is forward compatible with future jupyterhub versions.

[1] https://github.com/jupyterhub/kubespawner/blob/master/kubespawner/spawner.py#L986

kubeflow/core/prototypes/google-cloud-filestore-pv.jsonnet
```diff
@@ -0,0 +1,100 @@
+// @apiVersion 0.1
+// @name io.ksonnet.pkg.google-cloud-filestore-pv
+// @description Creates PV and PVC based on Google Cloud Filestore NFS
+// @shortDescription Creates PV and PVC based on Google Cloud Filestore NFS
+// @param name string Name for the component
+// @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
```
jlewi: Do we need a namespace parameter? Can't we just assume its the same as that of the environment. If users want to set it they can directly customize the generated component.

kubeflow/core/prototypes/google-cloud-filestore-pv.jsonnet
```diff
@@ -0,0 +1,100 @@
...
+// @optionalParam storageCapacity string 1T Storage Capacity
+// @optionalParam path string /kubeflow Path in NFS server
+// @param serverIP string Google Cloud Filestore Server IP
+
+// updatedParams uses the environment namespace if
+// the namespace parameter is not explicitly set
+local updatedParams = params {
+  namespace: if params.namespace == "null" then env.namespace else params.namespace,
```
jlewi: I think we just always do
namespace = env.namespace

kubeflow/core/prototypes/google-cloud-filestore-pv.jsonnet
ankushagarwal: This is just following the namespace param convention in all of our other components. We can refactor it out from our components if needed.

PR #1190, opened 2018-07-12
Use PV for /home/jovyan by default

### LINKED ISSUE #1187: Use PV by default mounted at /home/jovyan
Now that PV works with Jupyter I think we should make the default to use PV and mount it at
/home/jovyan.

jlewi:
* PV is working reliably now and its a better default experience to have
  notebooks backed up. Users who don't have a default storage class can disable
  it.

Fix #1187

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1190)


jlewi:
Closing this; I wanted to merge with the v0.2-branch

PR #1192, opened 2018-07-12
Support private GKE clusters in gke deploy.sh script

ankushagarwal:
* Update gke/deploy.sh to support Private GKE Clusters using PRIVATE_CLUSTER env variable
* Minor changes to cluster-kubeflow.yaml

/cc @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1192)

### DIFF
--- a/scripts/gke/deployment_manager_configs/cluster-kubeflow.yaml
+++ b/scripts/gke/deployment_manager_configs/cluster-kubeflow.yaml
@@ -53,11 +53,9 @@ resources:
       # Whether to enable Pod Security Policy Admission Controller
       # Use v1beta1 api
       podSecurityPolicy: false
-      masterAuthorizedNetworksConfig:
-          cidrBlocks:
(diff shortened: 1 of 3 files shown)


scripts/gke/deployment_manager_configs/cluster-kubeflow.yaml
```diff
@@ -53,11 +53,9 @@ resources:
       # Whether to enable Pod Security Policy Admission Controller
       # Use v1beta1 api
       podSecurityPolicy: false
-      masterAuthorizedNetworksConfig:
-          cidrBlocks:
```
lluunn: just curious: what's this?

ankushagarwal: By default any IP is allowed to connect to the master. `masterAuthorizedNetworksConfig` is a fancy name for firewall rules for master.

Issue #1193, opened 2018-07-13
deploy.sh is broken; wrong directory for the unpack?


jlewi:
It looks like the directory it got unpacked to didn't match the move

```
+ pwd
+ KUBEFLOW_REPO=/home/jlewi/git_kubecon_gh_demo/kubeflow_repo
+ KUBEFLOW_VERSION=0.2.1
+ [[ ! -d /home/jlewi/git_kubecon_gh_demo/kubeflow_repo ]]
+ '[' 0.2.1 == master ']'
+ TAG=v0.2.1
+ curl -L -o /tmp/kubeflow.0.2.1.tar.gz https://github.com/kubeflow/kubeflow/archive/v0.2.1.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   125    0   125    0     0     78      0 --:--:--  0:00:01 --:--:--    78
100 5307k    0 5307k    0     0   703k      0 --:--:--  0:00:07 --:--:-- 1382k
+ tar -xzvf /tmp/kubeflow.0.2.1.tar.gz -C /tmp
+ mv /tmp/kubeflow-v0.2.1 /home/jlewi/git_kubecon_gh_demo/kubeflow_repo
mv: cannot stat '/tmp/kubeflow-v0.2.1': No such file or directory
...

It looks like the v is being stripped out of the unpacked files.

**Comment by jlewi** [comment]
Workaround is too manually download it before running the script

```
export KUBEFLOW_REPO=....
curl -L -o /tmp/kubeflow.${KUBEFLOW_VERSION}.tar.gz https://github.com/kubeflow/kubeflow/archive/${TAG}.tar.gz
tar -xzvf /tmp/kubeflow.${KUBEFLOW_VERSION}.tar.gz  -C /tmp
mv /tmp/kubeflow-${KUBEFLOW_VERSION} "${KUBEFLOW_REPO}"

${KUBEFLOW_REPO}/scripts/deploy.sh
```

PR #1194, opened 2018-07-13
Fix deploy.sh moving the unpacked repo.

jlewi:
* It looks like GitHub is stripping out the v in the name of the top level
  directory.

Related to #1193

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1194)


jlewi:
/hold 
going to do some refactoring

PR #1199, opened 2018-07-13
Add a script to cleanup leaked kubeflow-ci ingress resources

ankushagarwal:
This script cleans up ingress related compute engine resources like backend-services,
forwarding rules, health-checks, etc. from kubeflow-ci project which
do not belong to the kubeflow-testing cluster

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1199)

### DIFF
--- a/scripts/cleanup_kubeflow_ci.sh
+++ b/scripts/cleanup_kubeflow_ci.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+# This script cleans up ingress related compute engine resources like backend-services,
+# forwarding rules, health-checks, etc. from kubeflow-ci project which
+# do not belong to the kubeflow-testing cluster
+
+set -xe
+
+PROJECT="--project=kubeflow-ci"
+
+# This id comes from ingress objects in kubeflow-testing cluster
+# They use a common suffix for GCP resources
+KUBEFLOW_TESTING_ID=c3cb19bff97cde34
+
+function cleanup() {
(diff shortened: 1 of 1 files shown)


scripts/cleanup_kubeflow_ci.sh
```diff
@@ -0,0 +1,28 @@
...
+# forwarding rules, health-checks, etc. from kubeflow-ci project which
+# do not belong to the kubeflow-testing cluster
+
+set -xe
+
+PROJECT="--project=kubeflow-ci"
+
+# This id comes from ingress objects in kubeflow-testing cluster
+# They use a common suffix for GCP resources
+KUBEFLOW_TESTING_ID=c3cb19bff97cde34
```
jlewi: Don't we have ingress that we don't want to delete? Don't we have an ingress exposing argo?

scripts/cleanup_kubeflow_ci.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# This script cleans up ingress related compute engine resources like backend-services,
+# forwarding rules, health-checks, etc. from kubeflow-ci project which
+# do not belong to the kubeflow-testing cluster
+
+set -xe
+
+PROJECT="--project=kubeflow-ci"
```
jlewi: nit: PROJECT should be the project name. The "--project" should be in the command.

ankushagarwal: Done

scripts/cleanup_kubeflow_ci.sh
ankushagarwal: All ingresses in the `kubeflow-testing` cluster create GCP objects with the suffix of `3cb19bff97cde34`. When we filter out this suffix from the GCP object names, we are protecting all the ingress resources of `kubeflow-testing` cluster

PR #1200, opened 2018-07-13
change {username}{servername} to {userid} to avoid volume length restrictions of 63 characters

### LINKED ISSUE #1177: unable to spawn jupyter notebook - volume name is too long
using IAP and a G suite account kam.d.kasravi@intel.com I get the following error when attempting to create the POD

Error: (422) Reason: error HTTP response headers: HTTPHeaderDict({'Audit-Id': '18fc2109-861e-4c94-943e-1946b9dbc2e5', 'Content-Type': 'application/json', 'Date': 'Wed, 11 Jul 2018 14:53:33 GMT', 'Content-Length': '942'}) HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Pod \"jupyter-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom\" is invalid: [spec.volumes[0].name: Invalid value: \"volume-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom\": must be no more than 63 characters, spec.containers[0].volumeMounts[0].name: Not found: \"volume-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom\"]","reason":"Invalid","details":{"name":"jupyter-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom","kind":"Pod","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: \"volume-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom\": must be no more than 63 characters","field":"spec.volumes[0].name"},{"reason":"FieldValueNotFound","message":"Not found: \"volume-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom\"","field":"spec.containers[0].volumeMounts[0].name"}]},"code":422}

kkasravi:
Fixes #1177

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1200)


jlewi:
How is userid set?

Could we add pod labels for the username (but truncate that to appropriate lengths) so its easy for people find their pods.

kkasravi:
i've kept the pod name as is which for IAP ends up being accounts.google:kam.d.kasravi@intel.com (where '.:@' are escaped). I've only changed the pvc name which is the attribute with the 63 character length description. The calculation of userid is a simple integer beginning from 1. So the pod has mount with names like volume-1, ... The userid is taken from the local sqllite db. Should tf-hub-0 die then these userid's are rebuilt.

kkasravi:
I could try labeling pods by using the email from IAP however it would still be escapified - it wouldn't have the accounts.google: prefix though.

jlewi:
Sorry I missed that it was only the PVC.

Don't we want to change the pod name template? I assume they will run into the same length issues.

It looks like the pod name is

```
jupyter-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom
```

Whereas volume name is

```
volume-accounts-2egoogle-2ecom-3akam-2ed-2ekasravi-40intel-2ecom
```

I think pod names are DNS names 
https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/identifiers.md

So aren't they subject to the same 63 character limit? So I think we will eventually hit the same issue if we don't truncate those names as well.

kkasravi:
Good point on recognizing both as being identifiers. 
I'll lop off the prefix 'accounts.google.com:' if it exists and truncate to 63?

kkasravi:
@jlewi thanks - i didn't think so either

kubeflow/core/kubeform_spawner.py
```diff
@@ -111,6 +113,33 @@ def get_env(self):
             env['GOOGLE_APPLICATION_CREDENTIALS'] = '{}/{}.json'.format(SERVICE_ACCOUNT_SECRET_MOUNT, gcp_secret_name)
         return env
 
+    def _parse_user_name(self, username):
```
jlewi: Can you add a TODO to add a unittest?

kkasravi:
this error about tfjobs is probably coming from the tf-operator repo

jlewi:
@lluunn is looking into the test failure (#1218).

Issue #1160, opened 2018-07-10
Create Prometheus Component in Kubeflow Core.


kunmingg:
Will integrate with stackdriver.
Support usage cases:

1. Fetch metrics from TF Serving. (https://github.com/kubeflow/kubeflow/issues/1036)
1. Fetch metrics from kf healthiness metrics_collector. (https://github.com/kubeflow/kubeflow/issues/1142)
1. Fetch metrics from Envoy to get network traffic metrics.

kunmingg:
@lluunn 
@jlewi

lluunn:
cc @ynqa

lluunn:
https://cloud.google.com/monitoring/kubernetes-engine/prometheus

jlewi:
#1204 was checked in. What is the remaining work?

Do we just need to document it?

kunmingg:
Only docs needed, this one we can close.

PR #1204, opened 2018-07-13
prometheus prototype

kunmingg:
Provide stackdriver compatible prometheus ks prototype to support monitoring within kubeflow.

For https://github.com/kubeflow/kubeflow/issues/1160

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1204)

### DIFF
--- a/kubeflow/core/prometheus.yml
+++ b/kubeflow/core/prometheus.yml
@@ -54,12 +54,9 @@ scrape_configs:
   - role: pod
 
   relabel_configs:
-#  - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_scrape]
-#    action: keep
-#    regex: true
-  - source_labels: [__meta_kubernetes_namespace]
+  - source_labels: [__meta_kubernetes_namespace,__meta_kubernetes_service_annotation_prometheus_io_scrape]
     action: keep
-    regex: kubeflow
+    regex: kubeflow;true
   - source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_path]
     action: replace
     target_label: __metrics_path__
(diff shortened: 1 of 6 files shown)


kunmingg:
@lluunn 
As a prototype yes, but it haven't integrate with our services yet

kubeflow/core/prometheus.yml
```diff
@@ -0,0 +1,143 @@
...
+# * `prometheus.io/path`: If the metrics path is not `/metrics` override this.
+# * `prometheus.io/port`: If the metrics are exposed on a different port to the
+# service then set this appropriately.
+- job_name: 'kubernetes-service-endpoints'
+
+  kubernetes_sd_configs:
+  - role: endpoints
+
+  relabel_configs:
+#  - source_labels: [__meta_kubernetes_service_annotation_prometheus_io_scrape]
```
lluunn: Is this intended to leave here or should be removed?

kubeflow/core/prometheus.yml
```diff
@@ -0,0 +1,143 @@
...
+    action: replace
+    target_label: __metrics_path__
+  - source_labels: [__address__,__meta_kubernetes_service_annotation_prometheus_io_port]
+    action: replace
+    target_label: __address__
+    regex: (.+)(?::\d+);(\d+)
+    replacement: $1:$2
+
+remote_write:
+- url: "https://monitoring.googleapis.com:443/"
```
lluunn: what's this url?

kubeflow/core/prometheus.yml
kunmingg: I think we might need to uncomment it and use as whitelist flag. 
And only scrape on signals we care about.

kubeflow/core/prometheus.yml
kunmingg: We write metrics data to stackdriver through this url.

kunmingg:
/hold
Need more change

lluunn:
jsonnet format failed

PR #1207, opened 2018-07-13
Split kubeflow-core into individual prototypes

### LINKED ISSUE #1107: Split all prototype into separate prototypes
Right now we have the "all" prototype which combines TFJob components JupyterHub and some others. We should split this into separate prototypes and get rid of all.

There are probably better patterns for grouping a set of related components (e.g. ksonnet modules or maybe the application itself).

ankushagarwal:
Splits into
* Ambassador
* Jupyterhub
* Central Dashboard
* Spartakus
* tf-job-operator

Also
* Update bootstrapper where kubeflow-core is hardcoded
* Other updates to READMEs

Changes to the website are here : https://github.com/kubeflow/website/pull/91

Fixes #1107

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1207)


ankushagarwal:
We do use cloud parameter for some components 
* jupyterhub - cloud is used to set the images. alibaba uses its own registry instead of gcr
* ambassador - cloud is used to determine the `isDashboardTls` parameter

PR #1208, opened 2018-07-14
[gcp-deployer] Require user sign in, fetch email, add splash screen

yebrahim:
This PR uses react router to isolate protected paths (only /deploy for now, which goes to the deploy form), and requires the user to login first before viewing the form, and fetches the user's email to populate its form field. It also adds a splash screen with the Google sign in button, and reworks the design a little. There are still more design changes to come.

Sorry for the long PR, I was refactoring the code into multiple react components, and didn't have the time to break it down properly. Here are a couple of screenshots, please file bugs if you see any:

Not signed in:
image

After sign in, redirects to /deploy:
image

Bad pathname:
image

/assign @jlewi 
/assign @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1208)


jlewi:
nit: Maybe squash all your commits. They will be squashed when the PR gets merged; but if you manually squash you will be able to create a cleaner commit message than the concatenation of all the commit messages.

jlewi:
Do you mind rebasing as noted above otherwise LGTM

yebrahim:
That's an interesting way to merge. I thought prow would automatically squash merge, and maybe use the PR's title as a commit message.
Anyway, fixed.

jlewi:
@yebrahim prow does squash merge but I think it takes the commit messages from all the commits and concatenates them; I don't think it uses the PR description.

PR #1209, opened 2018-07-14
Seldon 0.2 update

cliveseldon:
* Updates the CRD to v1alpha2. This is a breaking change and kubeflow examples will need updating. See https://github.com/SeldonIO/seldon-core#important-v1alpha2-update
 * Simplifies the raw json by creating it from the seldon-core helm chart by template to ensure both are the same.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1209)


cliveseldon:
Any chance of review  @jlewi @kunmingg

jlewi:
@cliveseldon Sorry Clive I missed this. You need to do /assign so it shows up in PR dashboard
/assign jlewi

kubeflow/seldon/json/README.md
```diff
@@ -0,0 +1,7 @@
+The template.json is generated from the helm source chart at https://github.com/SeldonIO/seldon-core/tree/master/helm-charts/seldon-core
+
+```
+helm template ${SELDON-CORE-HOME}/helm-charts/seldon-core > template.yaml
```
jlewi: Did you forget to add the .json files? Based on the diff it looks like a lot of .json files are deleted and I don't see new ones.

jlewi:
@cliveseldon Since this is a breaking change and we need to update the examples, can we maintain v1alpha1 and v1alpha2 and control with an option whether v1alpha1 or v1alpha2 is installed? This way we can update the examples before flipping the default.

jlewi:
I believe K8s 1.11 can install multiple versions of  a CRD. To support that it would be good to maintain both versions of seldon and have an option to install both. This would facilitate upgrading our examples in a sane fashion.

cliveseldon:
@jlewi Do you know of any examples. The [k8s docs](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#specify-multiple-versions) seem to have 1 CRD file. We have CRDs with different validation specs so this couldn't be specified in 1 CRD  file for the 2 versions.

cliveseldon:
I think multiple validation/OpenAPI specs is for a latter version of k8s, see [here](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/api-machinery/customresources-versioning.md) so at present we could have a more manual parameter to choose which version you want to install, but having both would not be possible I think.

kubeflow/seldon/json/README.md
cliveseldon: All the json is now in template.json which is generated from our centralized helm chart using ```helm template```

jlewi:
@cliveseldon So the process will need to be manual. You can follow what we did for TFJob operator. We added a parameter that controls whether we install v1alpha1 or v1alpha2 (kubeflow/tf-operator#598). 

kubernetes/features#544 is tracking versioning of CRDs. When this is implemented we would be able to install multiple versions of a CRD in a cluster. I think this made it into K8s 1.11.

So for 1.11 clusters I believe we should be able to install both CRDs so it would be good to have an option that also allows both CRDs to be installed simultaneously.

cliveseldon:
* The version of seldon to install can now be controlled via a seldonVersion parameter which will install appropriate components and the correct CRD depending on whether you ask for 0.1.X  or 0.2.X.

@jlewi

jlewi:
@cliveseldon Can rebase off the latest master please? That should fix the tests

cliveseldon:
* great idea about examples. I have modified the existing ones and will add some more as part of this PR: AB-test, Multi-Armed Bandit etc.
 * I have updated image pull policy

Can you point to me how I update docs for seldon in kubeflow? @jlewi

jlewi:
@cliveseldon 

The source for the seldon is here
https://github.com/kubeflow/website/blob/master/content/docs/guides/components/seldon.md

Which will render here
https://www.kubeflow.org/docs/guides/components/seldon/

The presubmit check will have a link to a preview website; GitHub markdown render is different from Netify so you should check the preview.

You can prepare a PR but not merge it yet because we don't have a way yet to pin kubeflow.org to the 0.2.0 version of the docs while continuing to make updates.

cliveseldon:
I have added Seldon templates for:
 *  an A-B Test
 * a Multi-Armed Bandit over 2 models
 * an Outlier Detection with Model.

jlewi:
/hold

I would suggest removing the namespace as a parameter and always use the value in the environment but I'll leave that decision to you.

/lgtm
/approve

cliveseldon:
I have removed the namespace optionalParam and use the env.namespace instead.

Issue #1000, opened 2018-06-14
Serving path should support logging request input/output


lluunn:
This will enable continuous training and skew detection etc.

jlewi:
Should we do this in TF serving? Can we do it in a sidecar? Does ISTIO provide any relevant functionality?

lluunn:
I think we can do in sidecar.
I remember some discussion that we should not do it in istio as they are very light weight. Will try to find them and propose a solution.

lluunn:
TF Serving seems to support this: https://github.com/tensorflow/serving/blob/master/tensorflow_serving/core/request_logger.h
Will investigate more.

lluunn:
To do it in TF Serving, we need to implement [LogCollector](https://github.com/tensorflow/serving/blob/master/tensorflow_serving/core/log_collector.h) for the backend (e.g. BigQuery)

lluunn:
I think doing it with sidecar is better
1. it can be easily reused by other frameworks
1. it's easier to make progress (put code in kubeflow v.s. in tf serving)

Fluentd should already support logging to BQ, GCS etc. I will try using that.

jlewi:
Closing this issue since initial PRs have been merged and we have issues filed for additional work.

PR #1229, opened 2018-07-18
TF serving support request logging

lluunn:
For #1000 

/cc @jlewi

- Using fluentd as a sidecar container to export logs.
- Choosing bigquery to begin with. Should be easy to add other backends using fluentd plugins.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1229)


lluunn:
/hold
need md

jlewi:
Please update the PR description to describe the approach being taken in this PR to solve the request logging problem.

components/k8s-model-server/request-logger/logging_worker.py
```diff
@@ -0,0 +1,62 @@
...
+  table_ref = dataset_ref.table(args.table)
+  table = client.get_table(table_ref)
+
+  # rotating files.
+  target_file = args.file_path + '.1'
+  while True:
+    time.sleep(10)
+    if not os.path.isfile(target_file):
+      continue
+    print('found the file\n')
```
jlewi: Use logging not print.

components/k8s-model-server/request-logger/requirements.txt
```diff
@@ -0,0 +1 @@
+google-cloud-bigquery
```
jlewi: Why do you need bigquery?

components/k8s-model-server/request-logger/logging_worker.py
```diff
@@ -0,0 +1,62 @@
...
+"""
+from __future__ import print_function
+
+import argparse
+import ast
+import json
+import os
+import time
+
+from google.cloud import bigquery
```
jlewi: Why do we need to write our own logging agent as opposed to using fluentd?
For example, https://github.com/kaizenplatform/fluent-plugin-bigquery

components/k8s-model-server/http-proxy/server.py
```diff
@@ -19,10 +19,12 @@
 from itertools import repeat
 import base64
 import logging
-import grpc
+import random
```
jlewi: If we do it in the http server proxy we won't get this for gRPC requests; right?

jlewi:
Are you emitting the logs in json format?

lluunn:
Using fluentd now, PTAL
@jlewi

lluunn:
/hold cancel

It's working now

jlewi:
This is great.

Can you update the PR description to explain the design choices; i.e. the fact that we are using fluentd and BigQuery?

lluunn:
Some of the comment are regarding how we present the prototype:
1. use tf-serving package, setting parameters, adding volumes/env if XXX
2. use examples package, ks generate copies the whole jsonnet, and let users modify directly.

I think the current libsonnet is already messy (filed #1264 ), so I choose option 2 here.

WDYT?

jlewi:
>use tf-serving package, setting parameters, adding volumes/env if XXX
use examples package, ks generate copies the whole jsonnet, and let users modify directly.

The prototype belongs in the tf-serving package regardless of how you generate it; i.e. you can still copy the whole jsonnet.

Telling a user to change certain lines in a text file is much less convenient then doing

```
ks param set fluent-conf project myproject
```

Since these parameters are required we should try to expose them as `ks param set` commands rather than requiring users to modify the conf.

lluunn:
Remove the conf file and make it in the prototype. project etc can be set via `ks param set`.
PTAL, thanks!

lluunn:
test failure: #1266

kubeflow/tf-serving/prototypes/tf-serving-with-request-log.jsonnet
```diff
@@ -0,0 +1,229 @@
...
+              },
+              requests: {
+                cpu: "1",
+                memory: "1Gi",
+              },
+            },
+          },
+          // Http proxy
+          {
+            name: "mnist-http-proxy",
```
jlewi: remove mnist in the name.

kubeflow/tf-serving/prototypes/tf-serving-with-request-log.jsonnet
```diff
@@ -0,0 +1,229 @@
...
+              fsGroup: 1000,
+            },
+            volumeMounts: [
+              {
+                name: "request-logs",
+                mountPath: "/tmp/logs",
+              },
+            ],
+          },
+          // Logging container.
```
jlewi: Lets file an issue to create an admission controller to inject the side car.

kubeflow/tf-serving/prototypes/tf-serving-with-request-log.jsonnet
```diff
@@ -0,0 +1,229 @@
...
+          {
+            name: "mnist-logging",
```
jlewi: Don't use mnist in the name.

jlewi:
Sorry for the slow reply; GCP next.

Minor comment about replacing the name "mnist" somewhere.

The duplication of code with the other templates is unfortunate. I think the best option is probably to create an admission controller that can inject the relevant information and then just make that its own prototype. Lets file a follow on issue.

PR #1232, opened 2018-07-18
commit for v0.1

TheJaySmith:
This is v0.1 of the KubeFlow ksonnet package, enabling Weaveworks Flux on Kubeflow clusters.

### DIFF
--- a/kubeflow/kubeflux/all.libsonnet
+++ b/kubeflow/kubeflux/all.libsonnet
@@ -1,0 +1,283 @@
+{
+  // Define the various prototypes you want to support.
+  // Each prototype should be a list of different parts that together
+  // provide a userful function such as WeaveWorks Flux
+  kubeflux(params, env):: [
+    $.parts(params, env).nodeport,
+    $.parts(params, env).fluxlb,
+    $.parts(params, env).flux,
+    $.parts(params, env).memcachedep,
+    $.parts(params, env).memcachesvc,
+    $.parts(params, env).serviceAccount,
+    $.parts(params, env).role,
+    $.parts(params, env).roleBinding,
+    $.parts(params, env).secret,
+
+  ],
+
+
+  // Parts should be a dictionary containing jsonnet representations of the various
+  // K8s resources used to construct the prototypes listed above.
+  parts(params, env):: {
+    // All ksonnet environments are associated with a namespace and we
+    // generally want to use that namespace for a component.
+    // However, in some cases an application may use multiple namespaces in which
+    // case the namespace for a particular component will be a parameter.
+    local namespace = if std.objectHas(params, "namespace") then params.namespace else env.namespace,
+    local giturl = if std.objectHas(params, "giturl") then params.giturl else env.giturl,
+
+    serviceAccount:: {
+        "apiVersion": "v1",
+        "kind": "ServiceAccount",
+        "metadata": {
+          "labels": {
+            "name": "flux",
+          },
+          "name": "flux"
+        }
+      },
+
+    role:: {
+        "apiVersion": "rbac.authorization.k8s.io/v1beta1",
+        "kind": "ClusterRole",
+        "metadata": {
+          "labels": {
+            "name": "flux",
+          },
+          "name": "flux"
+        },
+        "rules": [
+          {
+            "apiGroups": [
+              "*"
+            ],
+            "resources": [
+              "*"
+            ],
+            "verbs": [
+              "*"
+            ]
+          },
+          {
+            "nonResourceURLs": [
+              "*"
+            ],
+            "verbs": [
+              "*"
+            ]
+          }
+        ]
+      },
+
+      roleBinding:: {
+        "apiVersion": "rbac.authorization.k8s.io/v1beta1",
+        "kind": "ClusterRoleBinding",
+        "metadata": {
+          "labels": {
+            "name": "flux",
+          },
+          "name": "flux"
+        },
+        "roleRef": {
+          "apiGroup": "rbac.authorization.k8s.io",
+          "kind": "ClusterRole",
+          "name": "flux"
+        },
+        "subjects": [
+          {
+            "kind": "ServiceAccount",
+            "name": "flux",
+            "namespace": namespace
+          }
+        ]
+      },
+
+    secret:: {
+      // K8s Deployment,
+      "apiVersion": "v1",
+      "kind": "Secret",
+      "metadata": {
+        "name": "flux-git-deploy",
+      },
+      "type": "Opaque"
+    },
+
+    flux:: {
+      // K8s Deployment,
+      "apiVersion": "apps/v1beta1",
+      "kind": "Deployment",
+      "metadata": {
+        "name": "flux",
+      },
+      "spec": {
+        "replicas": 1,
+        "strategy": {
+          "type": "Recreate"
+        },
+        "template": {
+          "metadata": {
+            "labels": {
+              "name": "flux"
+            }
+          },
+          "spec": {
+            "containers": [
+              {
+                "args": [
+                  "--ssh-keygen-dir=/var/fluxd/keygen",
+                  "--git-url="+giturl,
+                  "--git-branch=master"
+                ],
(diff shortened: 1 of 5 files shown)


kubeflow/kubeflux/prototypes/kubeflux.jsonnet
```diff
@@ -0,0 +1,24 @@
...
+// the namespace parameter is not explicitly set
+local updatedParams = params {
+  namespace: if params.namespace == "null" then env.namespace else params.namespace,
+  giturl: if params.giturl == "null" then env.giturl else params.giturl,
+};
+
+
+local giturl = updatedParams.giturl;
+local namespace = updatedParams.namespace;
+//local imageTag = import "param://imageTag";
```
jlewi: Please delete commented out code.

kubeflow/kubeflux/prototypes/kubeflux.jsonnet
```diff
@@ -0,0 +1,24 @@
...
+// @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
+// @optionalParam giturl string null Your default git URL.
+
+local k = import "k.libsonnet";
+local all = import "kubeflow/kubeflux/all.libsonnet";
+
+// updatedParams uses the environment namespace if
+// the namespace parameter is not explicitly set
+local updatedParams = params {
+  namespace: if params.namespace == "null" then env.namespace else params.namespace,
```
jlewi: You can use the env namespace always you should not take namespace as a parameter.

kubeflow/kubeflux/prototypes/kubeflux.jsonnet
```diff
@@ -0,0 +1,24 @@
...
+  giturl: if params.giturl == "null" then env.giturl else params.giturl,
```
jlewi: You can't get giturl from env. It will always come from params.
env refers to values set as part of your ksonnet enviornment like namespace.

kubeflow/kubeflux/all.libsonnet
```diff
@@ -0,0 +1,283 @@
...
+          {
+            "port": 3030,
+            "protocol": "TCP",
+            "targetPort": 3030
+          }
+        ],
+        "selector": {
+          "name": "flux"
+        },
+        "type": "LoadBalancer"
```
jlewi: This would expose a port by default that's open to the world. We don't want to do this. This should be a parameter and users should have to explicitly choose to set the type to LoadBalancer in order to open up a port.

kubeflow/kubeflux/all.libsonnet
```diff
@@ -0,0 +1,283 @@
...
+                },
+                "name": "git-keygen"
+              }
+            ]
+          }
+        }
+      }
+    },
+
+    nodeport:: {
```
jlewi: Why do we have two different services?

kubeflow/kubeflux/all.libsonnet
```diff
@@ -0,0 +1,283 @@
...
+          "metadata": {
+            "labels": {
+              "name": "flux"
+            }
+          },
+          "spec": {
+            "containers": [
+              {
+                "args": [
+                  "--ssh-keygen-dir=/var/fluxd/keygen",
```
jlewi: How are keys handled? Can the user provide a deploy key via a K8s secret?

kubeflow/kubeflux/prototypes/kubeflux.jsonnet
TheJaySmith: Fixed

kubeflow/kubeflux/prototypes/kubeflux.jsonnet
TheJaySmith: This SHOULD be fixed

kubeflow/kubeflux/prototypes/kubeflux.jsonnet
TheJaySmith: Think I got it

kubeflow/kubeflux/all.libsonnet
TheJaySmith: It may just be a remnant from the conversion of Flux from yaml to ksonnet. Removed

kubeflow/kubeflux/all.libsonnet
TheJaySmith: Suggestions on how to fix this? I am thinking of a 

ks param set <deployment> LoadBalancer Deploy

I am just not sure how the code would look in the parameters and all.libsonnet

Also, should we give them more than one option for say a proxy? 

ks param set <deployment> port (Proxy/LoadBalancer)?

kubeflow/kubeflux/all.libsonnet
TheJaySmith: I don't see why they couldn't create an additional secret for the key. 
I am looking at the documentation here: https://github.com/weaveworks/flux/blob/master/site/standalone/setup.md

There are two options
"fluxctl identity" can generate one after deployment

kubectl create secret generic flux-git-deploy --from-file=identity=/path/to/private_key

Maybe create a parameter for option two?

jlewi: Right; They can create a key but they need some to  specify that the provided secret should be added to the pod. So you probably need a parameter (secret name) and if that is set mount that into the pod.

TheJaySmith: There is an existing "flux-git-deploy" secret created in this process. It looks like it mounts to a volume. Should we add a secret in addition to that or is there a reason we should create a new one.

jlewi: So the user has to download the key and then add it to their GITHUB repository as a deploy key? Is that documented somewhere.

kubeflow/kubeflux/all.libsonnet
jlewi: Looks like this still needs to be fixed.

jlewi:
Took a quick look; seems like some comments still need to be addressed.

kubeflow/kubeflux/all.libsonnet
TheJaySmith: So I am looking here: 
https://github.com/weaveworks/flux/blob/master/site/daemon.md

and "flux-git-deploy" seems to be used to store the private SSH key as a secret for GitHub. 

This documentation (https://www.weave.works/docs/cloud/latest/tasks/deploy/manual-configuration/) suggests that one use "fluxctl identity" to generate a public key while the private key is stored in "flux-git-deploy". You then go to Github and add the Key in the settings.

TheJaySmith:
The comments should be removed.

kubeflow/kubeflux/all.libsonnet
TheJaySmith: We need to expose the flux service to the outside world so that the "fluxctl" CLI tool can access the tool. The LoadBalancer was to prevent a user from having to do a portforward. I think a better option may be a proxy option as the service does need to be exposed

jlewi:
Jay it looks like the package is still named Kubeflux and not WeaveFlux. While I do like the name Kubeflux I think it will lead to confusion and we should just call it WeaveFlux because that's what it is.

Also Can rebase off of master and squash your changes. This will pull in the latest changes on master which should pull in some fixes to the E2E tests.

You will also need to run scripts/autoformat_jsonnet.sh to autoformat the jsonnet.

jlewi:
@TheJaySmith It looks like the autoformat worked. But it doesn't look like you rebased off the latest code of master because the minikube and tf serving tests are running. If you do that it will probably fix the tests.

kubeflow/weaveflux/README.md
```diff
@@ -0,0 +1,38 @@
+# KubeFlux
```
jlewi: you missed this "Kubeflox"

jlewi:
/lgtm
/approve

@TheJaySmith  Doesn't look like you rebased on the latest master make sure you pull in the latest changes to your remote branch before rebasing. So assuming its called origin

```
git fetch origin
git rebase -i origin/master
```

PR #1242, opened 2018-07-19
Adds check to see if Katib is deployed and updates CentralUI accordingly.

swiftdiaries:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1242)


components/centraldashboard/dashboard.go
```diff
@@ -0,0 +1,78 @@
...
+
+//SearchKatibPods searches across the cluster to see if Katib is deployed.
+func SearchKatibPods(clientset *kubernetes.Clientset) bool {
+
+	pods, err := clientset.CoreV1().Pods("").List(metav1.ListOptions{})
+	if err != nil {
+		fmt.Printf("Error in regexp")
+	}
+
+	katibpod := "vizier*"
```
toshiiw: You don't understand the regexp syntax and matching. "*" denotes a zero or more occurrences. It matches on "vizie". It must be "^vizier", or better use strings.HasPrefix.

 func example() {
    r, _ := regexp.Compile("foo*")

    fmt.Println(r.MatchString("food"))
    fmt.Println(r.MatchString("seafood"))
    fmt.Println(r.MatchString("info"))
    fmt.Println(strings.HasPrefix("food", "foo"))
    fmt.Println(strings.HasPrefix("seafood", "foo"))
}

swiftdiaries:
Will fix and update. Thank you for the review.

PR #1252, opened 2018-07-20
Enable new SD agents by default

lluunn:
Now that we have k8s 1.10.

/cc @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1252)


jlewi:
@lluunn You probably need to rebase to get the tests fixed.

/assign @kunmingg

scripts/gke/deployment_manager_configs/cluster.jinja
```diff
@@ -66,20 +66,10 @@ resources:
       # Create a very small minimal pool. Actual nodes will be managed
       # as additional node pools. This makes it easier to
       initialNodeCount: {{ properties['cpu-pool-initialNodeCount'] }}
-      {% if properties['stackdriver-kubernetes'] %}
-      # TODO: remove alpha when 10.2 is public.
-      # https://github.com/kubeflow/kubeflow/issues/821
-      enableKubernetesAlpha: true
       # We need 1.10.2 to support Stackdrivier GKE.
       initialClusterVersion: 1.10.2-gke.0
```
kunmingg: Now most recent version is 1.10.5-gke.3, and 1.10.2-gke.0 is no long available from UI.

scripts/gke/deploy.sh
```diff
@@ -159,6 +159,10 @@ if ${KUBEFLOW_DEPLOY}; then
 
   # Install the GPU driver. It has no effect on non-GPU nodes.
   kubectl apply -f https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/stable/nvidia-driver-installer/cos/daemonset-preloaded.yaml
+
+  # Install Stackdriver Kubernetes agents.
```
jlewi: Why do we need to do this? Aren't the stackdriver agents installed by default?

jlewi: The beta instructions don't mention installing these agents
https://cloud.google.com/monitoring/kubernetes-engine/installing?

scripts/gke/deployment_manager_configs/cluster.jinja
```diff
@@ -66,20 +66,10 @@ resources:
...
-      # Logging and monitoring have default value [logging/monitoring].googleapis.com
-      # if not set. We are using the new Stackdricer Kubernetes agents here.
-      # See cloud.google.com/monitoring/kubernetes-engine/.
-      loggingService: logging.googleapis.com/kubernetes
-      monitoringService: monitoring.googleapis.com/kubernetes
-      {% else %}
-      initialClusterVersion: 1.9.6-gke.1
-      {% endif %}
+      initialClusterVersion: 1.10.5-gke.3
+      loggingService: none
```
jlewi: Why don't we need to  set loggingService and monitoringService anymore?

scripts/gke/deploy.sh
lluunn: The instruction there are using gcloud or pantheon. We are doing manually here.

scripts/gke/deployment_manager_configs/cluster.jinja
lluunn: Those are legacy ones.

lluunn:
/hold
I might be wrong.

scripts/gke/deploy.sh
jlewi: Should we take advantage of the YAML support in ksonnet and start a package for all the YAML specs for GKE?

We could do this in a follow on a PR. I think the question is how can we make it easier to keep in sync between our web app and deploy.sh

Right now we'd have to specify the YAML spec in two different locations.

lluunn: https://github.com/kubeflow/kubeflow/issues/1288

PR #1255, opened 2018-07-20
insert service account keys to GKE cluster

kunmingg:
Store service account keys are secret to GKE cluster during deployment.

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -109,6 +109,8 @@ const theme = createMuiTheme({
   },
 });
 
+const k8s = require('@kubernetes/client-node');
+
 export default class DeployForm extends React.Component<any, DeployFormState> {
 
   constructor(props: any) {
@@ -402,8 +404,43 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
         });
       });
 
+    // Step 4: In cluster resource set up
+    const endpoint = Gapi.getClusterEndpoint(project, this.state.zone, deploymentName);
+    k8s.Config.defaultClient();
+    var k8sApi = k8s.Core_v1Api(`https://${endpoint}`);
+    const token = await Gapi.getToken();
+    if (!token) {
+      this.setState({
+          error: 'You are not signed in',
+          errorMessage: 'You must be signed in to use deploy Kubeflow to your GCP account.',
+      });
+      return;
+    }
+    k8sApi.setApiKey(k8s.Core_v1ApiApiKeys.BearerToken, token);
+
+    k8sApi.createNamespacedSecret('kubeflow',
+        this._generateServiceAccountSecret(project, deploymentName, 'admin'));
+    k8sApi.createNamespacedSecret('kubeflow',
+        this._generateServiceAccountSecret(project, deploymentName, 'user'));
+
   }
 
+  private _generateServiceAccountSecret(project: string, deploymentName: string, role: string) {
+    const roleSecret = new k8s.V1Secret();
+    roleSecret.apiVersion = 'v1';
+    roleSecret.metadata = new k8s.V1ObjectMeta();
+    roleSecret.metadata.name = `${role}-gcp-sa`;
+    roleSecret.metadata.namespace = 'kubeflow';
+    roleSecret.type = 'Opaque';
+    roleSecret.data = {};
+
+    const saKey = Gapi.iam.createKey(
+        project,
+        `${deploymentName}-${role}@${project}.iam.gserviceaccount.com`
+    );
+    roleSecret.data[`${role}-gcp-sa.json`] = Buffer.from(JSON.stringify(saKey)).toString("base64");
+    return roleSecret;
+  }
   /**
    * Returns a list of services that are needed but not enabled for the given project.
    */
@@ -412,6 +449,8 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
 
     const servicesToEnable = new Set([
       'deploymentmanager.googleapis.com',
+      'servicemanagement.googleapis.com',
+      'container.googleapis.com',
       'cloudresourcemanager.googleapis.com',
       'endpoints.googleapis.com',
       'iam.googleapis.com',
--- a/components/gcp-click-to-deploy/src/Gapi.ts
+++ b/components/gcp-click-to-deploy/src/Gapi.ts
@@ -47,6 +47,26 @@ export default class Gapi {
 
   }
 
+  public static iam = class {
+
+    public static async createKey(projectId: string, serviceAccounts: string) {
+        await this._load();
+        return gapi.client.projects.serviceAccounts.keys.create({
+            name: `projects/${projectId}/serviceAccounts/${serviceAccounts}`
+        }).then(response => response.result,
+                badResult => {
+          throw new Error('Errors creating key for serviceAccount ${serviceAccounts}');
+        });
+    }
+
+    private static async _load() {
+        await Gapi.load();
+        return new Promise(resolve =>
+            gapi.client.load('iam', 'v1', () => resolve()))
+    }
+
+  }
+
   public static servicemanagement = class {
 
     public static async list(project: string) {
@@ -90,6 +110,22 @@ export default class Gapi {
     }
 
   }
+
+  public static async getToken(): Promise<string | null> {
+      await this.load();
+      const user = await this._getCurrentUser();
+      return user ? user.getAuthResponse().id_token : null;
+  }
+
+    public static async getClusterEndpoint(projectId: string, zone: string, clusterId: string) {
+        await this.load();
+        return gapi.client.request({
+            path: `https://container.googleapis.com/v1/projects/${projectId}/zones/${zone}/clusters/${clusterId}`
+        }).then(response => (response.result as any).endpoint,
+            badResult => {
+                throw new Error('Errors getting cluster endpoint: ' + flattenDeploymentOperationError(badResult.result));
+            });
+    }
 
   public static async signIn(doPrompt?: boolean): Promise<void> {
     const rePromptOptions = 'login consent select_account';
--- a/components/gcp-click-to-deploy/package.json
+++ b/components/gcp-click-to-deploy/package.json
@@ -3,6 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@kubernetes/client-node": "^0.3.0",
     "@material-ui/core": "^1.3.1",
     "glamor": "^2.20.40",
     "glamorous": "^4.13.1",
@@ -22,6 +23,7 @@
     "@types/gapi": "0.0.35",
     "@types/gapi.auth2": "0.0.47",
     "@types/gapi.client.deploymentmanager": "^2.0.0",
+    "@types/gapi.client.iam": "^1.0.0",
     "@types/jest": "^23.0.0",
     "@types/js-yaml": "^3.11.1",
     "@types/node": "^10.3.2",


jlewi:
/lgtm
/assign @yebrahim 

Looks good to me but lets wait for Yasser.

@kunmingg Why is there a hold on the PR?

kunmingg:
@yebrahim @jlewi 
Thanks for reviewing, tested locally, found one blocker:
chrome won't allow requests to self-signed-cert address (like our k8s cluster).
Haven't found neat workaround...

How about give master ip a DNS though cloud endpoint?
Thoughts?

jlewi:
@kunmingg That won't work; we need the secret in order to be able to create the endpoint and get the signed certificate. In chrome if you try to navigate to a page with a self-signed cert it will give a warning and you can click through to accept it. is it not possible to do that in JS code?

jlewi:
@kunmingg @yebrahim Could we prompt the user to install the self signed cert for the K8s master in their browser?

kunmingg:
Found an workaround, now need another service:
https://github.com/kubeflow/kubeflow/issues/1276

yebrahim:
What kind of auth is currently used for the kubeflow cluster by default? If it's just basic username/password auth, then you don't want to open CORS for anyone to talk to the k8s master.

yebrahim:
@jlewi I don't know if that's doable, never come across this before. This seems like just a CORS issue though? If we add localhost and kubeflow.org to CORS whitelist origins, and use Bearer token for auth, should this take care of it?

kunmingg:
@yebrahim 
To add our hosts to CORS whitelist origins, I plan to host a CORS proxy on kubeflow.org (#1276). 
(or let user use chrome extension) 
Sounds good?

yebrahim:
If you're planning to add the hosts to the origins whitelist, why do you need the proxy? The browser should just be able talk directly to the k8s api server.

jlewi:
A proxy is highly undesirable. We don't want users to have to provide their credentials to a backend service they don't control.

@yebrahim We are connecting to the K8s master e.g.
https://XX.XXX.XXX

And we authenticate using a bearer token in the authorization header. So its not localhost; its the ip of the K8s master.

kunmingg:
@yebrahim 
We don't own k8s api server, can not directly edit from there.

kunmingg:
@yebrahim 
I think on GKE we do not have access to k8s master config.

yebrahim:
If this is the case, that's a huge bummer!
We should make sure of this before we write a server component to proxy requests. From brief googling, I can see there is a [CORS annotation for nginx ingress controller](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#enable-cors), but I don't know if kubeflow is using that.

Otherwise, seems like a server piece is unavoidable, @jlewi. The API server just won't accept requests coming from the browser without the CORS annotation.

jlewi:
@kunmingg @yebrahim My suggestion is to not block this PR on the CORS issue. I believe its possible to disable CORS in the browser using various extensions. 

I think here's one for chrome
https://chrome.google.com/webstore/detail/allow-control-allow-origi/nlfbmbojpeacfghkpbjhddihlkkiljbi?hl=en

So my suggestion is that we use such extensions as a workaround for now so we can continue to make progress.

Thoughts?

kunmingg:
@jlewi @yebrahim 
I can deploy k8s components through deployment manager:
we might need to deploy multi times per click.

We can worry about API endpoint later.

jlewi:
Which K8s resources? I think using DM to create a single resource e.g. the bootstrapper/click to deploy APP might be ok.

Overall though I think using DM to create K8s resources leads to bad UX because its unclear if a user wants to update those resources whether to do so using DM or by talking directly to K8s ApiServer.

kunmingg:
@jlewi @yebrahim 
Update:
1. Use deployment manager to insert secrets to unblock web app.
2. Add client id / secret fields.

Now web app should deploy kubeflow successfully on any projects with following steps:
1. (GCP UI) create oauth client id/secret in project
2. (GCP UI) Add ```Project IAM Admin``` role to ```Google APIs Service Agent``` 
3. (kubeflow.org UI) Click to deploy.

Note: will start another doc to discuss best way to integrate with new bootstrapper logic.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -412,6 +449,8 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
 
     const servicesToEnable = new Set([
       'deploymentmanager.googleapis.com',
+      'servicemanagement.googleapis.com',
```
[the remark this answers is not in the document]
yebrahim: Are we keeping this? I think it'll cause the service enable API to keep looping forever.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -404,43 +456,69 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+    let getAttempts = 0;
+    const getTimeout = 15000;
+    do {
+      getAttempts++;
+      const curStatus = await Gapi.deploymentmanager.get(this.state.project, deploymentName)
+            .catch(err => {
+              this._appendLine("Cluster endpoint not available yet.")
+            });
+      if (!curStatus) {
+        wait(getTimeout);
```
yebrahim: You need to `await` this.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -404,43 +456,69 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+        continue;
+      }
+      status = curStatus.operation!.status!
+    } while (status !== "DONE" && getAttempts < 20);
 
-    k8sApi.createNamespacedSecret('kubeflow',
-        this._generateServiceAccountSecret(project, deploymentName, 'admin'));
-    k8sApi.createNamespacedSecret('kubeflow',
-        this._generateServiceAccountSecret(project, deploymentName, 'user'));
+    const saKey = await Gapi.iam.createKey(
```
yebrahim: nit: use a more expressive name? maybe `serviceAccountKey`?

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -404,43 +456,69 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+      project,
+      `${deploymentName}-admin@${project}.iam.gserviceaccount.com`
+    );
+    const kubeflowUtil = this.state.clusterUtilSpec.resources[0]
 
-  }
+    kubeflowUtil.properties.clusterType = this.state.deploymentName + '-type';
+    kubeflowUtil.properties.saKey = saKey.privateKeyData;
+    kubeflowUtil.properties.clientId = window.btoa(this.state.clientId);
```
yebrahim: nit: you don't need the `window.` bit, you can call `btoa` directly.

components/gcp-click-to-deploy/src/Gapi.ts
```diff
@@ -50,21 +55,18 @@ export default class Gapi {
...
-        return gapi.client.projects.serviceAccounts.keys.create({
-            name: `projects/${projectId}/serviceAccounts/${serviceAccounts}`
-        }).then(response => response.result,
-                badResult => {
-          throw new Error('Errors creating key for serviceAccount ${serviceAccounts}');
+      await Gapi.load();
+      return gapi.client.request({
+          method: 'POST',
+          path: `https://iam.googleapis.com/v1/projects/${projectId}/serviceAccounts/${serviceAccounts}/keys`,
+        }).then(response => {
```
yebrahim: nit: you can simplify this to one line: `.then(response => response.result as CreateKeyResponse)`

components/gcp-click-to-deploy/src/Gapi.ts
```diff
@@ -50,21 +55,18 @@ export default class Gapi {
...
+                return response.result as CreateKeyResponse;
+            },
+          badResult => {
+          throw new Error('Errors creating key for serviceAccount ${serviceAccounts}: '
```
yebrahim: Use backticks "`" for string formatting.

components/gcp-click-to-deploy/src/configs/cluster-kubeflow-util.jinja
```diff
@@ -0,0 +1,27 @@
...
+  type: {{ CLUSTER_TYPE }}:{{ SECRETS_COLLECTION }}
+  properties:
+    apiVersion: v1
+    kind: Secret
+    namespace: {{ properties["namespace"] }}
+    metadata:
+      name: admin-gcp-sa
+    type: Opaque
+    data:
+      admin-gcp-sa.json: {{ properties["saKey"] }}
```
jlewi: This is a smart idea; using DM to insert the service account key into the cluster. Unfortunately this means the secret key would be stored in DM and viewable by anyone with access to the project. I don't think we want that.

An alternative approach would be to just use the service account as the service account for the GKE nodes.
The downside of that approach would be that it means all pods on that node would run with elevated privileges which would be undesirable.

Its possible that we could use RBAC to restrict what can run on a specific node pool so that other processes couldn't run on that node pool.

kunmingg: Currently our code assume those keys will be stored in k8s cluster, so anyone with project access can view from cluster secrets?
Another way is maybe we can expire those keys once we finish deploy?

kunmingg: By using VM account for nodes, we should be able to avoid exposing secrets.
Handled in separate PR: https://github.com/kubeflow/kubeflow/pull/1302

jlewi: I think putting them in DM might be an enhanced risk because K8s secrets are namespace scoped resource so not everyone might have access to the namespace. I'm not even sure project viewers do.

jlewi:
@kunmingg it looks like this PR is doing two things

1. Inserting Service accounts via DM
2. Adding OAuthClientID and OAuthSecret  for IAP

Per our discussion today I think the plan is to have a go backend. So I think we can just have the go backend create the secrets and insert them in the cluster. The go backend can talk directly to the K8s master.

kunmingg:
Update:
per our discussion, will use backend to deal with k8s resource management.
Separate PRs will take care backend part.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -25,6 +26,9 @@ import clusterJinjaPath from './configs/cluster.jinja';
 // selects auto domain then we should automatically supply the suffix
 // <hostname>.endpoints.<Project>.cloud.goog
 
+// Assume user access app via kubectl proxy
+const BackendAddress = 'http://127.0.0.1:8001/api/v1/namespaces/default/services/kubeflow-controller:8080/proxy/';
```
jlewi: Why do we need to make this assumption? Can we parameterize this based on how the user accessed the click to deploy app?

What if we had our go backend serve the click to deploy app? i.e. can we just add a handler to our go server that returns the javascript click to deploy app?

jlewi:
Thanks. Can you update the PR description as well please?

components/gcp-click-to-deploy/src/DeployForm.tsx
kunmingg: Will make it a parameter in yaml spec.

I think having go backend serve the app won't help with resolving backend api address: backend itself won't know from which address will user access it unless configured.

jlewi:
Per our discussion I think all the logic for deploying things should now move into the go backend; but we can do that in a follow on PR.

Issue #1266, opened 2018-07-25
[Test Flake] simple tf job failing; Job not found waiting for job


jlewi:
https://github.com/kubeflow/kubeflow/pull/1159

https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/1159/kubeflow-presubmit/2691/

Clearly the PR is not at fault because the PR is just adding a markdown doc.

The tf_job_client is returning a 404 waiting for the job

```
ERROR|2018-07-25T12:51:38|/mnt/test-data-volume/kubeflow-presubmit-kubeflow-e2e-gke-1159-922209d-2691-1f3e/src/kubeflow/tf-operator/py/tf_job_client.py|21
5| There was a problem waiting for Job simple-tfjob-gke.simple-tfjob-gke; Exception; (404)
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Date': 'Wed, 25 Jul 2018 12:51:38 GMT', 'Audit-Id': '240ad140-fdd1-4c52-a35f-446c30d3ebaf', 'Content-Length': '240
', 'Content-Type': 'application/json'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"tfjobs.kubeflow.org \"simple-tfjob-gke\" not found","re
ason":"NotFound","details":{"name":"simple-tfjob-gke","group":"kubeflow.org","kind":"tfjobs"},"code":404}
ERROR|2018-07-25T12:51:38|/mnt/test-data-volume/kubeflow-presubmit-kubeflow-e2e-gke-1159-922209d-2691-1f3e/src/kubeflow/tf-operator/py/test_runner.py|577|
 There was a problem running the job; Exception (404)
Reason: Not Found
```

jlewi:
The logs indicate the client is using v1alpha2

```
INFO|2018-07-25T12:51:38|/mnt/test-data-volume/kubeflow-presubmit-kubeflow-e2e-gke-1159-922209d-2691-1f3e/src/kubeflow/tf-operator/py/test_runner.py|415|
tfjob_version=v1alpha2
INFO|2018-07-25T12:51:38|/mnt/test-data-volume/kubeflow-presubmit-kubeflow-e2e-gke-1159-922209d-2691-1f3e/src/kubeflow/tf-operator/py/test_runner.py|423|
Wait for conditions Running, Succeeded, or Failed
[...]

PR #1267, opened 2018-07-25
Fix tfjob test; the simple tfjob test prototype was renamed.

jlewi:
* kubeflow/tf-operator#749 renamed the prototype
* ksonnet just fails silently if the component doesn't exist so the test just ends up failing with 404 trying to wait for the job.
* Fix 1266

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1267)

### DIFF
--- a/testing/workflows/components/workflows.libsonnet
+++ b/testing/workflows/components/workflows.libsonnet
@@ -417,7 +417,7 @@
               "--component=simple_tfjob_v1alpha2",
               // Name is used for the test case name so it should be unique across
               // all E2E tests.
-              "--params=name=simple-tfjob-" + platform + ",namespace=" + stepsNamespace + ",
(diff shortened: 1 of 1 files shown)


lluunn:
it's still failing with job not found

lluunn:
ref #1266

testing/workflows/components/workflows.libsonnet
```diff
@@ -418,7 +418,7 @@
               "--project=" + project,
               "--app_dir=" + tfOperatorRoot + "/test/workflows",
               "--tfjob_version=v1alpha1",
-              "--component=simple_tfjob",
+              "--component=simple_tfjob_v1alpha2",
```
lluunn: alpha1?

lluunn: and line 437 alpha2?

PR #1270, opened 2018-07-25
Small fix to tf serving component

### LINKED ISSUE #360: HTTP Proxy and TFServing should not use the same resource defaults
Right now HTTPProxy and TFServing both use the same resource defaults.

https://github.com/kubeflow/kubeflow/blob/master/kubeflow/tf-serving/tf-serving.libsonnet#L12

This doesn't make much sense as HTTPProxy probably needs much fewer resources. For example it probably doesn't need as much RAM since it doesn't load the model or GPUs in the case of GPU serving.

It might be worth looking at some examples of using with and mixins e.g.
https://github.com/ksonnet/parts/blob/master/incubator/memcached/memcached.libsonnet

It might nice if we could have a withResources option.

lluunn:
- remove unnecessary logic of http proxy image
- http proxy requests lower resource

Fix #360 

/cc @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1270)

### DIFF
--- a/kubeflow/tf-serving/tf-serving.libsonnet
+++ b/kubeflow/tf-serving/tf-serving.libsonnet
@@ -18,12 +18,7 @@
     deployIstio: false,
 
     deployHttpProxy: false,
-    defaultHttpProxyImage: "gcr.io/kubeflow-images-public/tf-model-server-http-proxy:v20180606-9dfda4f2",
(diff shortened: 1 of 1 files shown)


lluunn:
need lgtm again @ankushagarwal .
thanks!

Issue #337, opened 2018-03-03
Create a simple TF model to test TF Serving


jlewi:
For E2E testing of TF serving it would be great to have a very simple model; e.g. a model that would just do a matrix multiple.

We should create a simple python script that can produce a model in a format suitable for consumption by TF serving.

For reference we use a similarly simple TF program for testing training
https://github.com/kubeflow/tf-operator/tree/master/examples/tf_sample/tf_sample

nishi-t:
Hi @jlewi,
I want to try this issue for my first contribution. I write simple codes for this issue and put it [my gist](https://gist.github.com/nishi-t/b9002119de7af7c0d23084878b797624) now. 
Where should I put it in the repository for making PR ?

Thnak you,

lluunn:
Since it's for testing, I think it should go to kubeflow/testing.
@nishi-t do you want to send a PR?

jlewi:
@lluunn Won't it be easier to keep the model in the same repository as the rest of the TFServing code? This way you make changes to the code and the test in the same PR?

lluunn:
Ah good point! Sounds good.

analyticalmonk:
Hi @jlewi,
Are you still looking for contribution towards this issues? If yes, I'll like to work on it.

jlewi:
That would be great

CSP197:
Hello @jlewi
I see there is a merged PR referencing this issue (#1271)
Has this issue been fixed? Or are you still accepting contributions for this issue?
-CSP

PR #1271, opened 2018-07-26
Improve tf serving e2e test

### LINKED ISSUE #198: e2e test for http-proxy
Follow up of this https://github.com/kubeflow/kubeflow/pull/183

To Do
  - [x] ksonnet template
  - [ ] E2E tests
  - [x] Documentation explaining how to use it
  - [ ] Discussion of the API

### LINKED ISSUE #1031: unflake TF serving testing

lluunn:
Fix #1031 
  - now compares the result with tolerance

Fix #198
  - predict with http. Http proxy send gRPC requests, so it tests both grpc and http.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1271)


jlewi:
DId you consider creating a simple deterministic model ( #337) that does something like x + 1?  I think this is fine as a short term solution but I think that approach might be better longer term.

lluunn:
Makes sense. I will do that in a followup PR.

PR #1297, opened 2018-08-01
Add basic permissions to read storage and launch tfjobs inside the containers

gaoning777:
We assume that these permissions are basic permissions to be granted during initialization.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1297)

### DIFF
--- a/scripts/gke/deployment_manager_configs/cluster.jinja
+++ b/scripts/gke/deployment_manager_configs/cluster.jinja
@@ -21,7 +21,7 @@ limitations under the License.
    that will be created to represent Kubernetes objects.
    There is type corresponding to each API endpoint.
 #}
-{% set VM_OAUTH_SCOPES = ['https://www.googleapis.com/auth/logging.write', 'https://www.googleapis.com/auth/monitoring'] %}
+{% set VM_OAUTH_SCOPES = ['https://www.googleapis.com/auth/logging.write', 'https://www.googleapis.com/auth/monitoring', 'https://www.googleapis.com/auth/cloud-platform' ] %}
 
 
 {# Names for service accounts.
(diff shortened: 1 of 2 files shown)


gaoning777:
@IronPan I updated the cluster.jinja for permission grant. Can you also review the codes, please.

IronPan:
Do you also need to change default node to 3?
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster-kubeflow.yaml#L39

scripts/gke/deployment_manager_configs/cluster.jinja
```diff
@@ -21,7 +21,7 @@ limitations under the License.
    that will be created to represent Kubernetes objects.
    There is type corresponding to each API endpoint.
 #}
-{% set VM_OAUTH_SCOPES = ['https://www.googleapis.com/auth/logging.write', 'https://www.googleapis.com/auth/monitoring'] %}
+{% set VM_OAUTH_SCOPES = ['https://www.googleapis.com/auth/logging.write', 'https://www.googleapis.com/auth/monitoring', 'https://www.googleapis.com/auth/cloud-platform' ] %}
```
kunmingg: If we add the third one, then first two is included.

@jlewi will it be too open to give anyone in cluster full access to all resources and services?

jlewi: Agree with Kunming why can't you use the namespace scoped secret.

IronPan: what's the risk of having cloud platform scope?

Issue #1293, opened 2018-07-31
ack_guide.md out of date


pdmack:
The Alibaba-specific user guide needs updating based on the re-organization of kubeflow components and prototypes.

/assign @cheyang

cheyang:
@pdmack, I will update it, thanks.

cheyang:
already fixed in #1388

PR #1298, opened 2018-08-01
support customized docker registry

cheyang:
Fix issue #1293 

Provide environment variables to update default docker registry. This can both help when using private registry and Alibaba Cloud.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1298)

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -10,6 +10,7 @@ KUBEFLOW_VERSION=${KUBEFLOW_VERSION:-"master"}
 KUBEFLOW_DEPLOY=${KUBEFLOW_DEPLOY:-true}
 K8S_NAMESPACE=${K8S_NAMESPACE:-"kubeflow"}
 KUBEFLOW_CLOUD=${KUBEFLOW_CLOUD:-"minikube"}
+KUBEFLOW_DOCKER_REGISTRY=${KUBEFLOW_DOCKER_REGISTRY:-""}
 
 if [[ ! -d "${KUBEFLOW_REPO}" ]]; then
   if [ "${KUBEFLOW_VERSION}" == "master" ]; then
@@ -37,6 +38,11 @@ DEPLOYMENT_NAME=${DEPLOYMENT_NAME:-"kubeflow"}
 
 KUBEFLOW_KS_DIR=${KUBEFLOW_KS_DIR:-"`pwd`/${DEPLOYMENT_NAME}_ks_app"}
 
+if [[ ! -z "$KUBEFLOW_DOCKER_REGISTRY" ]]; then
+  find $KUBEFLOW_REPO -name "*.libsonnet" -o -name "*.jsonnet" | xargs sed -i -e "s%gcr.io%$KUBEFLOW_DOCKER_REGISTRY%g"
+  find $KUBEFLOW_REPO -name "*.libsonnet" -o -name "*.jsonnet" | xargs sed -i -e "s%quay.io%$KUBEFLOW_DOCKER_REGISTRY%g"
+fi
+
 cd $(dirname "${KUBEFLOW_KS_DIR}")
 
 set +e
@@ -44,6 +50,7 @@ kubectl create ns ${K8S_NAMESPACE}
 set -e
 
 ks init $(basename "${KUBEFLOW_KS_DIR}")
+
 cd "${KUBEFLOW_KS_DIR}"
 
 ks env set default --namespace "${K8S_NAMESPACE}"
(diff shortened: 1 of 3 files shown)


cheyang:
In our case, you can follow the such steps:

```
curl -O http://kubeflow.oss-cn-beijing.aliyuncs.com/ks_0.11.0_linux_amd64.tar.gz
tar -xvf ks_0.11.0_linux_amd64.tar.gz
mv ks_0.11.0_linux_amd64/ks /usr/local/bin/
export KUBEFLOW_VERSION=master
export KUBEFLOW_CLOUD=ack
export KUBEFLOW_DOCKER_REGISTRY=registry.aliyuncs.com
curl https://raw.githubusercontent.com/cheyang/kubeflow/features/customized_image_repo/scripts/deploy.sh | bash
```

cheyang:
Thanks for your good suggestion. I think it's a big refactor for the kubeflow project. I think it's the way to go when developing kubeflow ksonnet asset. We also need such guidance in kubeflow community. But because I'm not familiar with ksonnet, can anyone who is good at this to do such refactor?

scripts/deploy.sh
```diff
@@ -10,6 +10,7 @@ KUBEFLOW_VERSION=${KUBEFLOW_VERSION:-"master"}
 KUBEFLOW_DEPLOY=${KUBEFLOW_DEPLOY:-true}
 K8S_NAMESPACE=${K8S_NAMESPACE:-"kubeflow"}
 KUBEFLOW_CLOUD=${KUBEFLOW_CLOUD:-"minikube"}
+KUBEFLOW_DOCKER_REGISTRY=${KUBEFLOW_DOCKER_REGISTRY:-""}
```
jlewi: All this environment variables are creating another layer of indirection and config. Basically we are creating a giant params file.

I don't think we want to keep going down that road. In fact I have a PR removing a bunch of the variables we added to start configuring GCP options.

scripts/deploy.sh
```diff
@@ -37,13 +38,19 @@ DEPLOYMENT_NAME=${DEPLOYMENT_NAME:-"kubeflow"}
 
 KUBEFLOW_KS_DIR=${KUBEFLOW_KS_DIR:-"`pwd`/${DEPLOYMENT_NAME}_ks_app"}
 
+if [[ ! -z "$KUBEFLOW_DOCKER_REGISTRY" ]]; then
```
jlewi: I think this should be a separate script.

I think the flow we want is

1. User creates an APP with default configs
2. User performs any necessary customizations
3. User deploys

scripts/deploy.sh
cheyang: Please let me know which PR it is. Thanks.

cheyang: I agree. I think I have to hold on my PR when the changes are done, so I can follow the principle of community.

jlewi: I think if you pull this out into a separate script you should be ok for now.

So the instructions would be something like

1. Follow Kubeflow guide to create a ksonnet ap
2. cd to ksonnet app
3. Run script to change all the images.

cheyang: May I put this environment variable in the deploy.sh? Because I think the only difference between the separate script and the current deploy script is updating docker registry. and if we don't set `KUBEFLOW_DOCKER_REGISTRY`, there will no impact to the current deploy script.

The real challenge for me is how to name the new script. :)  deploy_with_customized_reigstry.sh?

jlewi: I want to get rid of deploy.sh; see #1308 

Users will do something like the following

kfctl.sh init ${NAME} --platform=aliyuncs
cd ${NAME}
kfctl.sh generate k8s
kfctl.sh apply k8s

Under the hood when the user calls kfctl.sh generate k8s that will check $PLATFORM and can invoke a platform dependent function. In this case we can invoke your function to customize the images.

ks init will create "env.sh" . So you can persist environment variables there during the init stage.
https://github.com/kubeflow/kubeflow/pull/1308/files#diff-8a6f2d237f052b07ca594c20830ed9d2R37

So I guess the best thing to do would be to wait for kfctl.sh to be checked in.

cheyang: Thank you. I will keep eyes on that.

jlewi:
Is the plan to wait for #1308  to be submitted? Or did you want to submit this first?

cheyang:
I decide to wait for #1308 .

jlewi:
#1308 is now submitted.

I think what you want to do is add a function here:
https://github.com/kubeflow/kubeflow/blob/master/scripts/kfctl.sh#L214

that based on your platform configures the Docker images appropriately in the ksonnet app.

You might want to implement this by calling a script that provides more flexibility. This way if a user wants to customize all the images but for a different registry they can do that by reusing your script.

PR #1302, opened 2018-08-01
allow IAP setup using VM service account when no service account was set

kunmingg:
So we can setup IAP without inserting service account key into cluster.
 
Related: discussion in https://github.com/kubeflow/kubeflow/pull/1255


jlewi:
Why do we need this?

Now that we are going with a go backend I think we can just create secrets via RPC using credentials in the incoming RPC.

See my comment in your other PR
https://github.com/kubeflow/kubeflow/pull/1255

PR #1308, opened 2018-08-03
Turn deploy.sh into kfctl.sh

jlewi:
Simplify deployment script and make it more rentrant

* The scripts are becoming more complicated because we are creating
  more environment variables to control things like whether or not to use
  GCFS.

* Those environment variables are creating an unnecessary layer of indirection;
  if users want to configure things like private cluster, or GCFS, they
  should just edit the ksonnet or DM configs.

* To support that we need to make the following flow easier

  create DM and ksonnet configs
  user manually edits configs
  user deploys

* Refactor the deploy.sh script to make the steps more easier and add a
  command function so user can select which step they want to perform.

* Only copy a subset of the files in the Kubeflow repo into the directory;
  the repo is quite large and we expect user to check their directory into
  source control and we don't want to bloat it.

* Remove GCFS from the setup; I think it makes sense for GCFS setup to
  a separate step. If users want to use GCFS they should

  1. Create a GCFS instance if they don't have one
  2. Update the ksonnet app to use GCFS
  3. Redeploy the ksonnet app

  * This removes the complexity caused by giving users options to skip to fix bugs in deploy.sh.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1308)


jlewi:
This should be ready for review once the tests pass.

jlewi:
/retest
/hold cancel

/assign @ankushagarwal 

@ankushagarwal I think you can start reviewing this; the test failures look like transient failures. I will add retries to run.sh to try to deal with that.

jlewi:
@ankushagarwal  Tests are passing. PTAL

PR #1311, opened 2018-08-04
Update deploy.sh

372046933:
kubeflow requires that k8s version should be >= 1.8. Now kubernetes has 1.11 released.  Now user can specify k8s version to `ks init`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1311)


372046933:
@googlebot I signed it!

372046933:
@jlewi  `ks init` defaults to `v1.8.0`. Besides, ks cannot judge the version of k8s from `.kube/config` because there is no version info there. You can have a try.

jlewi:
Thanks 

We need to wait for #1308 which replaces deploy.sh and we'll need to figue out how to set the version consistent with that PR.

Is it possible to set the version using the ks command after the app is created?

The idea of #1308 is that we don't want to keep adding another config layer on top of ksonnet and plumbing more options into that config.

So the idea would be we create an app with default config and then user can customize it; the K8s version being one of the customizations

372046933:
Great, I hope that day will come soon!

372046933:
Hooray

jlewi:
I'm going to go ahead and close this; please reopen if its still relevant.

Issue #896, opened 2018-05-31
Investigate using tensorflow serving's built-in http server


lluunn:
https://github.com/tensorflow/serving/pull/902

https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/http_server.h

yupbank:
i did tried a bit... i think most of the stuff we provide, c++ version already have, except for they are most strict on the data type and less friendly on the error message. i did even did some bench mark with the same model, the python version would add ~ 10ms overhead .

PR #1321, opened 2018-08-06
Add Rest API Port to TF Serving

### LINKED ISSUE #1299: Use tensorflow/serving images
https://hub.docker.com/r/tensorflow/serving/

We can retire some infra for building and releasing those images.

activatedgeek:
This adds the ability to expose the Rest API port directly from Tensorflow Model Server instead of adding a sidecar proxy container. See [Tensorflow docs](https://www.tensorflow.org/serving/api_rest).

We also move away from custom TF Serving images and move to official ones. (Fixes #1299)

/cc @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1321)

### DIFF
--- a/kubeflow/tf-serving/tf-serving.libsonnet
+++ b/kubeflow/tf-serving/tf-serving.libsonnet
@@ -118,9 +118,7 @@
       args: [
         "/usr/bin/tensorflow_model_server",
         "--port=9000",
-      ] +
-      if $.params.exposeRestApi then ["--rest_api_port=8000"] else []
-      + [
+        if $.params.exposeRestApi then "--rest_api_port=8000",
         "--model_name=" + $.params.modelName,
         "--model_base_path=" + $.params.modelPath,
       ],
@@ -129,8 +127,8 @@
           containerPort: 9000,
         },
         {
-          containerPort: 8000,
-        },
+          [if $.params.exposeRestApi then "containerPort"]: 8000
+        }
       ],
       // TODO(jlewi): We should add readiness and liveness probes. I think the blocker is that
       // model-server doesn't have something we can use out of the box.
@@ -365,7 +363,7 @@
           spec+: {
             containers: [
               $.gcpParts.tfServingContainer,
-              if $.util.toBool($.params.deployHttpProxy) then
+              if $.util.toBool($.params.deployHttpProxy) && !$.params.exposeRestApi then
                 $.parts.httpProxyContainer,
             ],
             volumes: [
--- a/kubeflow/tf-serving/tf-serving.libsonnet
+++ b/kubeflow/tf-serving/tf-serving.libsonnet
@@ -117,12 +117,16 @@
       args: [
         "/usr/bin/tensorflow_model_server",
         "--port=9000",
+        "--rest_api_port=8000",
         "--model_name=" + $.params.modelName,
         "--model_base_path=" + $.params.modelPath,
       ],
       ports: [
         {
           containerPort: 9000,
+        },
+        {
+          containerPort: 8000,
         },
       ],
       // TODO(jlewi): We should add readiness and liveness probes. I think the blocker is that
--- a/kubeflow/tf-serving/tf-serving.libsonnet
+++ b/kubeflow/tf-serving/tf-serving.libsonnet
@@ -17,7 +17,7 @@
 
     deployIstio: false,
 
-    exposeRestApi:: false,
+    exposeRestApi:: true,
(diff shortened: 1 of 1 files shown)


kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -117,13 +118,19 @@
...
+      + [
         "--model_name=" + $.params.modelName,
         "--model_base_path=" + $.params.modelPath,
       ],
       ports: [
         {
           containerPort: 9000,
         },
+        {
+          containerPort: 8000,
```
lluunn: should this be conditional?

lluunn:
ref https://github.com/kubeflow/kubeflow/issues/896

jlewi:
/unassign @jlewi
I'll let @lluunn  handle this

kubeflow/tf-serving/tf-serving.libsonnet
activatedgeek: I've let this be there because it does not impact how the service behaves.

jlewi: is the port actually exposed though? I think one point of setting the port is to let people/services know which ports are open. So if the port isn't open we probably shouldn't list it.

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -17,6 +17,7 @@
 
     deployIstio: false,
 
+    exposeRestApi:: false,
```
jlewi: The default should probably be true.

lluunn: https://github.com/tensorflow/serving/blob/9e960b37607da80b8c4c06754e0873c71ae1cd2f/tensorflow_serving/model_servers/main.cc#L334

The http server is only started if it's set, so I think I should only set it when needed.

kubeflow/tf-serving/tf-serving.libsonnet
lluunn: Also, if the http proxy is deployed, will it cause problem?

kubeflow/tf-serving/tf-serving.libsonnet
jlewi: I'm suggesting that by default we should deploy TFServing with the REST API enabled and people can disable it if they want to.

activatedgeek: My reason for keeping it an opt-in variable was because the older versions don't have the REST API feature.

Alternatively, we could update the default images in the lib file to use TF 1.8. I suspect anyone would want to use a lower version anyways now that 1.10 is out so this should be a safe choice.

jlewi: I think that makes sense.

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -358,7 +363,8 @@
           spec+: {
             containers: [
               $.gcpParts.tfServingContainer,
-              if $.util.toBool($.params.deployHttpProxy) then
+              if $.util.toBool($.params.deployHttpProxy) &&
+                 $.util.toBool($.params.exposeRestApi) == false then
```
lluunn: This will break the existing behavior: set deployHttpProxy to true won't deploy http proxy.

So in test http proxy is not deployed: https://github.com/kubeflow/kubeflow/blob/master/components/k8s-model-server/images/releaser/components/workflows.libsonnet#L95

activatedgeek:
/hold

REST API and the HTTP Proxy need to co-exist because HTTP Proxy allows us the new logging feature.

activatedgeek:
@lluunn Should we just always expose the Rest API. TF is now 1.10 and since it supports upto 3 versions until 1.8, all of these support rest api. Shall we just remove the conditional and expose it at a non-conflicting port?

activatedgeek:
@lluunn From the [releases page](https://github.com/tensorflow/serving/releases) of TF Serving 1.10

> TensorFlow Serving API now uses gRPC's GA release. The beta gRPC API has been deprecated, and will be removed in a future version of TensorFlow Serving. Please update your gRPC client code

Is this relevant?

lluunn:
we are using gRPC beta to send request in test: https://github.com/kubeflow/kubeflow/blob/master/testing/test_tf_serving.py#L27

But I isn't that 1.10 and you are using 1.9?

PR #1329, opened 2018-08-08
[gcp-deployer] add service account key handler and a check health handler

kunmingg:
Add 2 handlers to web app go backend:

service account key handler: 
given service account and tmp token, create service account key and insert into GKE cluster.

check health handler
check if backend respond to post request.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1329)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/SaKey.go
+++ b/bootstrap/cmd/bootstrap/app/SaKey.go
@@ -66,7 +66,7 @@ func (s *ksServer) InsertSaKey(ctx context.Context, request InsertSaKeyRequest) 
 
 	c, err := iamadmin.NewIamClient(ctx, option.WithTokenSource(ts))
 	if err != nil {
-		log.Errorf("Can now create iam admin client: %v", err)
+		log.Errorf("Cannot create iam admin client: %v", err)
 		return err
 	}
 	createServiceAccountKeyRequest := admin.CreateServiceAccountKeyRequest{
--- a/bootstrap/cmd/bootstrap/app/SaKey.go
+++ b/bootstrap/cmd/bootstrap/app/SaKey.go
@@ -1,0 +1,100 @@
+package app
+
+import (
+	"golang.org/x/net/context"
+	"cloud.google.com/go/container/apiv1"
+	iamadmin "cloud.google.com/go/iam/admin/apiv1"
+	"google.golang.org/api/option"
+	"golang.org/x/oauth2"
+	"k8s.io/client-go/rest"
+	containerpb "google.golang.org/genproto/googleapis/container/v1"
+	"google.golang.org/genproto/googleapis/iam/admin/v1"
+	"fmt"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"encoding/base64"
+)
+
+
+type InsertSaKeyRequest struct {
+	Cluster string
+	Namespace string
+	Project string
+	SecretKey string
+	SecretName string
+	ServiceAccount string
+	Token string
+	Zone string
+}
+
+func buildClusterConfig(ctx context.Context, ts oauth2.TokenSource, request InsertSaKeyRequest) (*rest.Config, error) {
+	c, err := container.NewClusterManagerClient(ctx, option.WithTokenSource(ts))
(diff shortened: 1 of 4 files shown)


bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -534,6 +566,7 @@ func decodeCreateAppRequest(_ context.Context, r *http.Request) (interface{}, er
 
 // The same encoder can be used for all RPC responses.
 func encodeResponse(_ context.Context, w http.ResponseWriter, response interface{}) error {
+	log.Info("encode response: ")
```
yebrahim: Did you meant to add this? It'll be very noisy.

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -550,6 +583,35 @@ func (s *ksServer) StartHttp(port int) {
...
+			if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
+				return nil, err
+			}
+			return request, nil
+		},
+		encodeResponse,
+	)
+
 	http.Handle("/apps/create", createAppHandler)
+	http.Handle("/liveness", livenessHandler)
```
yebrahim: Use `/healthz`?

yebrahim: Similarly, can `makeLivenessEndpoint` be `makeHealthzEndpoint`.. etc?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -515,8 +525,30 @@ func makeCreateAppEndpoint(svc KsService) endpoint.Endpoint {
...
+		}
+		return r, nil
+	}
+}
+
+func makeLivenessEndpoint(svc KsService) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		req := request.(LiveRequest)
+		r := &LiveResponse{}
+		r.Reply = req.Msg + "accepted! Sill alive!"
```
yebrahim: Are there no other checks we want to perform here to make sure the entire system is up and running? If the endpoint returns a success code, can the caller then assume that all other services are up.. etc?

bootstrap/cmd/bootstrap/app/SaKey.go
```diff
@@ -0,0 +1,100 @@
...
+	})
+	k8sConfig, err := buildClusterConfig(ctx, ts, request)
+	if err != nil {
+		log.Errorf("Failed getting GKE cluster info: %v", err)
+		return err
+	}
+
+	c, err := iamadmin.NewIamClient(ctx, option.WithTokenSource(ts))
+	if err != nil {
+		log.Errorf("Can now create iam admin client: %v", err)
```
yebrahim: "Cannot"*

bootstrap/cmd/bootstrap/app/SaKey.go
```diff
@@ -0,0 +1,100 @@
...
+	secretData := make(map[string][]byte)
+	secretData[request.SecretKey] = createdKey.PrivateKeyData
+	_, err = k8sClientset.CoreV1().Secrets(request.Namespace).Create(
+		&v1.Secret{
+			ObjectMeta: meta_v1.ObjectMeta{
+				Namespace: request.Namespace,
+				Name:      request.SecretName,
+			},
+			Data: secretData,
+		})
```
yebrahim: Is this call synchronous? What does the caller of the SA key endpoint expect to happen after making the API call? Do they then have to wait or poll somehow if they want to block on this operation?

jlewi:
I took a look; I don't have anything to add beyond @yebrahim's comments.

bootstrap/cmd/bootstrap/app/ksServer.go
kunmingg: Currently this one only indicate backend is reachable and have cycles to respond to requests.

We can keep a server status variable to indicate backend status, which should kill the pod if needed so we have auto-start.
Let's create issue for it.

bootstrap/cmd/bootstrap/app/SaKey.go
kunmingg: It's synchronous. When handler returns, either the sa key already inserted into GKE or an error happened and included in response.

yebrahim:
Awesome, thanks!
/lgtm

Issue #1325, opened 2018-08-07
Reusability and composability of steps in our E2E tests


jlewi:
Our E2E tests often have subgraphs/steps that we'd like to reuse across different E2E tests.
To date we've largely followed one of two approaches

1. We either copy/paste the code between multiple workflows

    * e.g. [tf-operator/workflows](https://github.com/kubeflow/tf-operator/blob/master/test/workflows/components/workflows.libsonnet) & [kubeflow/workflows](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/workflows.libsonnet)

2. We create a single workflow and parameterize it it with conditionals e.g. [kubeflow/workflows](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/workflows.libsonnet) is used for minikube and GKE.

Both of these approaches have drawbacks and it would be nice to come up with better patterns of reuse among our E2E tests. Specifically, we'd like to make it easy to define and reuse steps or DAGS across E2E tests.

Example: We'd like to define the steps/dag to deploy Kubeflow and make it easy to reuse that across all component tests that need to deploy Kubeflow.

I think we can leverage

  * Argo [Nested Workflows](https://github.com/argoproj/argo/blob/master/examples/nested-workflow.yaml)
 *  ksonnet to make this more reusable.

In a ".libsonnet" file we can define Argo DAGs corresponding to specific functions e.g.
  * Deploy Kubeflow
  * Teardown Kubeflow
  * Run a particular test (e.g TFJob test)

The libsonnet file can then be imported into individual E2E tests and the DAG can be used as a nested workflow.

If needed we can take advantage of ksonnet features (e.g. late binding) to make it easy to override parameters.

jlewi:
I might take a stab at defining a workflow to run deploy Kubeflow using kfctl.sh based on #1308.

lluunn:
+1

PR #1333, opened 2018-08-08
[wip] Add TFJob test to the Kfctl test; refactor workflows to start to use 

jlewi:
Start adding component tests to the new kfctl workflow.

* This PR adds the TFJob simple to that workflow.

* To support that we start refactoring workflows.libsonnet to make it easy to define an Argo Dag that can be nested in other workflows so we can reuse subgraphs across tests.

* Related to #1325 Reusability of subgraphs in our tests.


lluunn:
Do you want to add testing/ to prowconfig https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml#L21
so that the ktctl_test is triggered?

Also, I think the variable names in kfTests is confusing.
We have argoDagTemplate, argoTemplates, dagTemplates, argoTemplate (in the buildTemplate).
Can we make it more clear what's the different usage of each one? Maybe better variable names, or at least some comments.

jlewi:
The test is already being triggered by prow
https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml#L14

jlewi:
I take that back looks like its not actually getting triggered.
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/1333/kubeflow-presubmit/2862/

jlewi:
I think its not being triggered because we aren't including the directory in the prow_config will update.

lluunn:
tide shows not mergeable, not sure why.

jlewi:
Looks like
https://github.com/kubernetes/test-infra/issues/9001

PR #1335, opened 2018-08-09
add bindRoleHandler which bind roles to service accounts

kunmingg:
Provide this handler so we can edit service account roles used by kubeflow through backend server.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1335)


jlewi:
What is the reasoning behind deciding which RPCs to expose?

Why not simply expose a high level function that performs all the needed steps? 

If you look at kfctl.sh
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/README.md

There are two commands
generate - creates the configs
apply - applies the config (calls DM, creates service accounts, creates secrets, calls ks apply etc...).

Why not just have two similar RPCs? And treat specific RPCs (like role bindings as internal RPCs).

kunmingg:
@jlewi 
We'll have the 2 high level handler you mentioned.
But this one is necessary because without editing IAM first (through user's token), deployment manager as a service, don't have sufficient permission to edit project IAM.

PR #1341, opened 2018-08-10
Update TFJob operator to the latest image.

jlewi:
/assign @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1341)


jlewi:
kubeflow/tf-operator#779 attempts to make the test more reliable.

Issue #1348, opened 2018-08-10
GKE version "foo" is unsupported.


kunmingg:
Currently in our deployment manager config we use GKE version "1.9.7-gke.3" as default, which recently got removed by GKE and not available any more.
Will remove GKE version from deployment manager config (when deploy, DM will use GKE's current default version), since GKE's version update and deprecate is quite frequent. To avoid this error happening again.    
Leave one conditional version for stackdriver which requires ^1.10. Should remove once ^1.10 becomes default.

kunmingg:
@richardsliu @jlewi 
This currently blocks all e2e tests.

jlewi:
I don't think we we want to upgrade the major version by default; we've seen breaking changes when moving from 1.9.7 to 1.10. So we should only upgrade the patch level. 
e.g. from 1.9.7-gke.3 to 1.9.7-gke.5
but not from 1.9.7 to 1.10

jlewi:
I think what we should do is

1. Have a default value in cluster-kubeflow.yaml
2. Check during project initialization if its valid
3. If its not valid change the default with a log message to a supported version.

sarahmaddox:
Hallo @kunmingg I'm getting this error at the moment.  I'm using Kubeflow v0.2.2

What I ran:
export KUBEFLOW_VERSION=0.2.2
curl https://raw.githubusercontent.com/kubeflow/kubeflow/v${KUBEFLOW_VERSION}/scripts/gke/deploy.sh | bash

Error:
- code: RESOURCE_ERROR
  location: /deployments/sarahmaddox-kubeflow-gcp/resources/sarahmaddox-kubeflow-gcp
  message: '{"ResourceType":"container.v1.cluster","ResourceErrorCode":"400","ResourceErrorMessage":{"code":400,"message":"Master
    version \"1.9.6-gke.1\" is unsupported.","status":"INVALID_ARGUMENT","statusMessage":"Bad
    Request","requestPath":"https://container.googleapis.com/v1/projects/ml-writers/zones/us-central1-c/clusters","httpMethod":"POST"}}'

Can you tell me how I can work around this problem until it's fixed? It's blocking my deployment, and I can't figure out where to reset the version.

jlewi:
@sarahmaddox  try the following

```
export KUBEFLOW_VERSION=0.2.2
export KUBEFLOW_DEPLOY=false
[...]

PR #1349, opened 2018-08-10
remove cluster version from DM config, use GKE default one 

kunmingg:
To avoid GKE version deprecation errors happening again.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1349)


kunmingg:
https://github.com/kubeflow/kubeflow/issues/1348

kunmingg:
There might be breaking changes when GKE jump to next major version. Need more logic to handle it.

PR #1351, opened 2018-08-10
Create ks prototype for tf-batch-predict

yixinshi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1351)


jlewi:
I think we'd just like to have a simple prototype for the k8s/job to run the Dataflow job.
The prototype should go in the examples package.
You can take a look at the example here

We want to put all the logic in the jsonnet file so that when the user runs ks generate they get the full spec and can easily customize it.

kubeflow/examples/prototypes/tf-batch-predict.jsonnet
```diff
@@ -0,0 +1,193 @@
...
+          cpu: "1",
+        },
+        limits: {
+          memory: "4Gi",
+          cpu: "4",
+        },
+      },
+    },  // tfBatchPredictContainerBase
+
+    tfBatchPredictContainer+: base.parts.tfBatchPredictContainerBase +
```
jlewi: Lets simplify the prototype further. There's not much reason to pull out tfBatchPredictContainer and then override by merging in a dictionary; just paste the relevant dictionary describing the container into the container arrays below line 111.

We want the examples to be optimized for readability. The idea is that if people want to modify them (e.g. they want to set appropriate keys for AWS they would delete and replace the appropriate lines.

In other words, think of this as an "example" that people will copy and modify.

Ideally the example should run under some suitable set of conditions; e.g. assuming there is a valid GCP credential either as a VM service account or as a secret key if secretName is set.

I think we should keep conditionals for setting the following based on parameters

1. num GPUs
2. GCP secret.

PR #1355, opened 2018-08-14
upgrade gke version for v0.2

lluunn:
For #1348 

/cc @kunmingg 
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1355)

### DIFF
--- a/scripts/gke/deployment_manager_configs/cluster.jinja
+++ b/scripts/gke/deployment_manager_configs/cluster.jinja
@@ -78,7 +78,7 @@ resources:
       loggingService: logging.googleapis.com/kubernetes
       monitoringService: monitoring.googleapis.com/kubernetes
       {% else %}
-      initialClusterVersion: 1.9.6-gke.1
+      initialClusterVersion: 1.9.7-gke.5
       {% endif %}
       {% if properties['gkeApiVersion'] == 'v1beta1' %}
       podSecurityPolicyConfig:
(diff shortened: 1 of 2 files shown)


jlewi:
@lluunn looks like you didn't pick the correct version

```
 "endTime": "2018-08-13T17:08:14.627-07:00",
 "error": {
   "errors": [
     {
       "code": "RESOURCE_ERROR",
       "location": "/deployments/e2e-59de/resources/e2e-59de",
       "message": "{\"ResourceType\":\"container.v1.cluster\",\"ResourceErrorCode\":\"400\",\"ResourceErrorMessage\":{\"code\":400,\"message\":\"Master ve
rsion \\\"1.9.6-gke.1\\\" is unsupported.\",\"status\":\"INVALID_ARGUMENT\",\"statusMessage\":\"Bad Request\",\"requestPath\":\"https://container.googleap
is.com/v1/projects/kubeflow-ci/zones/us-east1-d/clusters\",\"httpMethod\":\"POST\"}}"
```

jlewi:
Looking at the logs it looks like there is a failure due to an incorrect import of JobTimeoutError
https://github.com/kubeflow/tf-operator/blob/master/py/test_runner.py#L571

But that looks like a red herring and not the root cause of why the job failed.

Created kubeflow/tf-operator#780

jlewi:
I'm going to ignore the test failure and just merge this manually.

The test failure is with v1alpha1 which we don't really care about. It doesn't look like the job actually runs v1alpha2 tests.

The cluster is created successfully.

So the test failure is unlikely related to this PR.

PR #1362, opened 2018-08-15
Use gcr.io images for argo, ambassador and cert-manager

ankushagarwal:
When using private gke clusters, dockerhub images are not available in
the cluster. So I have mirrored the images from dockerhub and quay to
gcr.io/kubeflow-images-public registry and updated prototypes to use
those images instead of dockerhub

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1362)

### DIFF
--- a/kubeflow/argo/prototypes/argo.jsonnet
+++ b/kubeflow/argo/prototypes/argo.jsonnet
@@ -4,7 +4,6 @@
 // @shortDescription Argo workflow engine
 // @param name string Name to give to the component
 // @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
-// @optionalParam imageTag string v2.1.1 Docker image tag to use. It is automatically inherited from the environment if not set.
 // @optionalParam workflowControllerImage string argoproj/workflow-controller:v2.1.1 workflowControllerImage
 // @optionalParam uiImage string argoproj/argoui:v2.1.1 uiImage
 // @optionalParam executorImage string argoproj/argoexec:v2.1.1 executorImage
@@ -19,6 +18,5 @@ local updatedParams = params {
 };
 
 local namespace = updatedParams.namespace;
-local imageTag = import "param://imageTag";
 
 std.prune(k.core.v1.list.new(argo.parts(updatedParams).all))


kubeflow/argo/prototypes/argo.jsonnet
```diff
@@ -4,7 +4,7 @@
 // @shortDescription Argo workflow engine
 // @param name string Name to give to the component
 // @optionalParam namespace string null Namespace to use for the components. It is automatically inherited from the environment if not set.
-// @optionalParam imageTag string latest Docker image tag to use. It is automatically inherited from the environment if not set.
+// @optionalParam imageTag string v2.1.1 Docker image tag to use. It is automatically inherited from the environment if not set.
```
jlewi: image parameters should always be the full image path i.e. registry and tag. Any logic to combine a tag and image name should happen upstream.

jlewi:
Long term; I think we need a better long term solution. I think we need to treat the private GCR cluster as an optional customization. When that is turned on we need to swap in the appropriate images.

So there's two pieces to this:

1. We need tooling to help sync images to the registry
1. We need to be able to overwrite the images to point to a different registry when needed.

Can we make incremental progress on #2? Specifically, can we create a script that will overwrite the docker images via calls to ks param.

So for a private cluster the steps would be

```
kfctl init myapp
kfctl generate platform
kfctl generate k8s
cd myapp/ks_app
./use_gcr_for_all_images.sh
```


I don't think we can rely on being able to set the default images; we need a consistent way to overwrite the images. For example, we need to be able to point to images in DockerHub when deploying in China. 
See: kubeflow/kubeflow#1298

kubeflow/argo/prototypes/argo.jsonnet
ankushagarwal: Done.

ankushagarwal:
Addressed comments - created a new script to set the images.

kubeflow/argo/prototypes/argo.jsonnet
jlewi: Can we delete the imageTag param now?

jlewi:
One minor comment; 
Can you also update the docs in the website to explain that we need to run the script?

ankushagarwal:
Website changes are WIP https://github.com/kubeflow/website/pull/165/files#diff-7383ce26651e4a7a8084724286fd7835R372

kubeflow/argo/prototypes/argo.jsonnet
ankushagarwal: Done

PR #1364, opened 2018-08-15
Disable spartakus usage reporting in our ci clusters.

jlewi:
Related to: kubeflow/testing#191


lluunn:
Some changes are from #1357 , is that intended to move here? If so, please approve.
/lgtm

PR #1371, opened 2018-08-15
Add user service account developer access to k8s cluster

activatedgeek:
This PR updates the base GKE deployment.

The User Service Account provided with the cluster deployment may need access to deploy resources on to GKE. The addition of this role allows that. This role is most restrictive possible.

Since, we aim to allow users to interact with the cluster from Jupyter notebooks, this change has implications on general usage and not a particular use case.

/cc @jlewi @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1371)

### DIFF
--- a/scripts/gke/deployment_manager_configs/cluster.jinja
+++ b/scripts/gke/deployment_manager_configs/cluster.jinja
@@ -228,6 +228,10 @@ resources:
           members:
             - {{ 'serviceAccount:' + KF_USER_NAME + '@' + env['project'] + '.iam.gserviceaccount.com' }}
 
+        - role: roles/container.developer
+          members:
+            - {{ 'serviceAccount:' + KF_USER_NAME + '@' + env['project'] + '.iam.gserviceaccount.com' }}
+
         - role: roles/logging.logWriter
           members:
             {# VM service account is used to write logs. #}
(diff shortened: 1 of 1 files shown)


jlewi:
/unassign @kunmingg 
/assign  @ankushagarwal 

Assigning Ankush since he's been working on our overall enterprise / security story.

jlewi:
I think we have two options

1. We can use IAM (what this PR is doing) but this gives access on a per project basis
https://cloud.google.com/kubernetes-engine/docs/how-to/role-based-access-control

1. We can use RBAC which should allow finer grained permissions. 
For example we can create a role binding that is scoped to particular namespaces and even certain permissions within the namespace.

I think I lean towards using RBAC so we can create finer grained permissions. I think Kubeflow should define appropriate RBAC roles that can then easily be bound to one or more users to give them appropriate permissions.

I think Ankush already defined the RBAC role jupyter-notebook-role for this purpose
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/jupyterhub.libsonnet#L223

My guess is the following is what's happening.

There are two service accounts in the pod
  1. A K8s service account (bound to the RBAC role above)
  1. The GCP service account 

My hunch is you are activating the GCP service account in your notebook (in order to talk to GCP services). At which point you start using the GCP identity with RBAC. So the problem is that that the GCP service account isn't bound to jupyter-notebook-role.

@activatedgeek Is that the problem? Can you try creating a role binding for the GCP service account and seeing if that works? 

If that's the case then I think the thing to do would be to add that role binding always for GCP.

ankushagarwal:
Discussed offline with @activatedgeek. We can use jupyter-notebook k8s service account for this and don't need to create a new GCP service account.

activatedgeek:
Thanks @ankushagarwal!

activatedgeek:
See #1374 for a follow up update which is a simpler way of achieving this objective.

PR #1372, opened 2018-08-15
[kfctl web app] frontend & backend update to agree with api changes

kunmingg:
Update API calls to backend service.
Now all manual steps are covered by backend:

Update: 
  1. one-click-to-deploy works now on any projects.
  2. use DM config based on the one used by kfctl.
  3. backend security enhancement: make sure auth token only shorted lived in memory. 
  4. ingress compatibility: simplify ingress rule & handle health check.

Follow up:
1. (recommended) auto redirect when detect that IAP is up.
2. (optional) polish UI elements

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1372)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -116,7 +116,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
       dialogBody: '',
       dialogTitle: '',
       hostName: '<HOST>.endpoints.<PROJECT>.cloud.goog',
-      ipName: 'kubeflow',
+      ipName: 'kubeflow-ip',
       project: 'cloud-ml-dev',
       showLogs: false,
       zone: 'us-east1-d',
@@ -279,6 +279,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     kubeflow.properties.zone = this.state.zone;
     kubeflow.properties.clientId = btoa(this.state.clientId);
     kubeflow.properties.clientSecret = btoa(this.state.clientSecret);
+    kubeflow.properties.ipName = this.state.ipName;
 
     const state = this.state;
     const email = await Gapi.getSignedInEmail();
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -344,6 +344,26 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     // enabling, make requests to enable them, then we repeat this in a loop
     // until we have no more services left, or we try too many times.
     this._appendLine(`Getting enabled services for project ${project}..`);
+    await request(
+      {
+        body: JSON.stringify(
+          {
+            Msg: 'Echo from web app?',
+          }
+        ),
+        headers: { 'content-type': 'application/json' },
+        method: 'PUT',
+        uri: this._configSpec.appAddress + '/healthz',
+      },
+      (error, response, body) => {
+        if (!error) {
+          const msg = JSON.parse(response.body).Reply;
+          this._appendLine('From Backend' + msg);
+        } else {
+          this._appendLine('Backend error: ' + response.statusCode);
+        }
+      }
+    );
 
     let servicesToEnable: string[] = [];
     let enableAttempts = 0;
@@ -412,6 +432,30 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     }
 
     this._appendLine('Proceeding with project number: ' + projectNumber);
+    const token = await Gapi.getToken();
+    await request(
+      {
+        body: JSON.stringify(
+          {
+            Project: project,
+            ProjectNumber: projectNumber,
+            Token: token,
+          }
+        ),
+        headers: { 'content-type': 'application/json' },
+        method: 'PUT',
+        uri: this._configSpec.appAddress + '/initProject',
+      },
+      (error, response, body) => {
+        if (!error) {
+          const msg = JSON.parse(response.body).Reply;
+          this._appendLine('project init succeeded: ' + msg);
+        } else {
+          this._appendLine('Backend error: ' + JSON.parse(response.body).Reply);
+          return;
+        }
+      }
+    );
 
     // Step 3: Create GCP Deployment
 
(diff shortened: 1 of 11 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -344,6 +344,26 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     // enabling, make requests to enable them, then we repeat this in a loop
     // until we have no more services left, or we try too many times.
     this._appendLine(`Getting enabled services for project ${project}..`);
+    await request(
+      {
+        body: JSON.stringify(
+          {
+            Msg: 'Echo from web app?',
```
yebrahim: I would prefer the communication with the Go backend is transparent to the user. Echoes would seem strange here, we can just display a message saying "Could not reach server" or something.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -344,6 +344,26 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+        headers: { 'content-type': 'application/json' },
+        method: 'PUT',
+        uri: this._configSpec.appAddress + '/healthz',
+      },
+      (error, response, body) => {
+        if (!error) {
+          const msg = JSON.parse(response.body).Reply;
+          this._appendLine('From Backend' + msg);
+        } else {
+          this._appendLine('Backend error: ' + response.statusCode);
```
yebrahim: We should probably fail the deployment at this point, no? What's the point of continuing if we won't be able to pass on the token and create the service account key, etc?

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -412,6 +432,30 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+            Token: token,
+          }
+        ),
+        headers: { 'content-type': 'application/json' },
+        method: 'PUT',
+        uri: this._configSpec.appAddress + '/initProject',
+      },
+      (error, response, body) => {
+        if (!error) {
+          const msg = JSON.parse(response.body).Reply;
```
yebrahim: Let's always wrap `JSON.parse` calls in a `try/catch`, especially since we don't have string API typings (protos or such). If the server decides to return something else here, the client should handle this gracefully by showing an error dialog, instead of things just stopping to work.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -440,36 +484,38 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     // Step 4: In-cluster resources set up
     let status = '';
     let getAttempts = 0;
-    const getTimeout = 15000;
+    const getTimeout = 20000;
     do {
       getAttempts++;
+      await wait(getTimeout);
```
yebrahim: Why move the wait up here? We should only wait after each call fails right? I think that's what the old code does.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -440,36 +484,38 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+              this._appendLine('Deployment status not available yet.');
             });
       if (!curStatus) {
-        await wait(getTimeout);
         continue;
       }
       status = curStatus.operation!.status!;
-    } while (status !== 'DONE' && getAttempts < 20);
+      this._appendLine(`Status of ${deploymentName}: ` + status);
+    } while (status !== 'DONE' && getAttempts < 30);
```
yebrahim: We should add one more check on `getAttempts` after the `while` loop is done. If it's equal to 30, that means we've waited too long, and we should show a suitable timeout error message.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -440,36 +484,38 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
```
yebrahim: We need to check other values of the deployment status, and exit the loop with an error if the deployment failed for example.

components/gcp-click-to-deploy/src/DeployForm.tsx
kunmingg: We want to always wait:
even call succeeded, if the returned status is "RUNNING", we will still wait for 20s.
So in total we wait up to 20*30 (10 min), till deployment manager successfully set up every thing (usually finish within 5 min).

kunmingg:
/hold
Will update DM configs and add apply call.

jlewi:
Thanks.

kunmingg:
@jlewi 
About DM configs:
Because our priority is making deploy app works properly, this PR focus on making everything works.
(Tested and now the app works properly).
So take a shortcut here to unblock @lluunn 's web app update.

DM config update is the only piece left on frontend side, will migrate to backend, and merge with script's DM config in following PRs of this sprint.

jlewi:
Makes sense thanks.

/lgtm
/approve

This looks good on my end.

@yebrahim Any additional comments?

/hold.

jlewi:
Looks like an Argo test failure
/test all

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -412,9 +426,51 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+        }
+      }
+    );
+
+    let initAttempts = 0;
+    const initTimeout = 2000;
+    do {
+      initAttempts++;
+      await wait(initTimeout);
+    } while (!readiness && initAttempts < 10);
```
yebrahim: How does this work? `readiness` isn't being changed in the loop body. This will always either wait for 10*2000 ms or not wait at all.

yebrahim:
Just the one comment. Otherwise
/lgtm

components/gcp-click-to-deploy/src/DeployForm.tsx
kunmingg: ```readiness``` is changed in previous call back function, here the loop is to make sure call back updated the status.

yebrahim: If you're `await`ing the request call (line 426), then the next code block will execute only when that call is finished and the callback is executed. It seems to me what you're trying to do here is run the request in parallel to this code block (starting line 456), which isn't how you should do this.

yebrahim: In fact, I just noticed the `request` library isn't installed as a dependency, only it's typings. This should not be working.
You can just use the built-in `fetch` API. It's [supported](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility) by all major browsers.

kunmingg: Which one is not working? This PR has been deployed to kubeflow.dev

yebrahim: This code that calls `request(...` has not been deployed to kubeflow.dev. It's been a while since there's been a deployment as far as I'm aware.

The code I meant by not working is the call using `request`, that library isn't installed as part of the app, and this call will fail if the execution gets to it.

kunmingg: We deployed this PR earlier on a new domain name. ```request``` seems works.

jlewi:
@kunmingg Has @yebrahim's comment about the request package been addressed? I looked and couldn't tell if the dependencies had been cleaned up.

I'll lgtm it and if there are no more changes needed you can just cancel the hold

/lgtm
/approve
/hold

components/gcp-click-to-deploy/src/DeployForm.tsx
yebrahim: - I see, that makes sense then, although we should explicitly install `request` as a dependency rather than rely on its existence by chance.
- Can you send me the new domain name?
- My original comment is still valid I believe. If you want to wait for the request to come back use a Promise (either using `request` or using the build-in `fetch`), rather than manually wait in a loop.

kunmingg: new address is: kfdeploy.kubeflow.dev
merge PR to unblock @lluunn 
Will use promise in following PRs.

PR #1378, opened 2018-08-17
Change the default branch used for the configs to v0.2-branch and not master

jlewi:
* This is better UX since we want the deploy script to be in sync with
  the version of the configs.

Related to #1348

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1378)


jlewi:
Same test failure as in
https://github.com/kubeflow/kubeflow/pull/1355#issuecomment-413091845

Going to ignore it as in that PR; and merge manually.

PR #1380, opened 2018-08-17
Fix groups for various resources

activatedgeek:
Role groups for a few resources were not correct.

- `deployments` belongs to `apps` group (and also `extensions` for compatibility with older versions)
- `jobs` belongs to `batch` group
- Allow all resource verbs

/cc @ankushagarwal

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1380)


ankushagarwal:
/lgtm
/approve

Thanks!

PR #1387, opened 2018-08-19
Added a minikube setup script.

abhi-g:
Adding a shell script to make it easier for someone trying to setup local Minikube based Kubeflow deployment. 
Some features:
 - Downloads kubectl, minikube and installs if necessary
 - Dynamic configuration of minikube VM based on host resources and user input
 - Allows a local host directory to be mounted into the Jupyter notebooks to allow for access to local datasets and persisting notebooks across minikube restarts / complete reconfigurations 

/hold

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1387)


scripts/setup-minikube.sh
```diff
@@ -0,0 +1,354 @@
...
+  if [ $svc_ready = false ]
+  then
+    return 1
+  fi
+}
+
+
+# if user requested a local fs path to be mounted, make it accessible via
+# Jupyter Notebooks
+function mount_local_fs() {
```
jlewi: To better fit this into the kfctl.sh pattern;
pv.yaml should be dumped to the k8s_specs directory during the generate phase of the kfctl command. And it should then be created during the kfctl apply phase

scripts/setup-minikube.sh
```diff
@@ -0,0 +1,354 @@
...
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 10Gi
+  volumeName: local-volume
+EOF
+  kubectl create -n kubeflow -f ./pv-claim.yaml
+
+  # update tf-hub stateful set env to use the claim
+  kubectl -n kubeflow set env statefulset tf-hub -e KF_PVC_LIST=local-notebooks
```
jlewi: Can we do this in kfctl.sh during the generate phase?
Can we check if platform is minikube and then set ksonnet parameters to set these parameters on tf-hub?

jlewi:
Right now this script is wrapping kfctl.sh; can we instead have kfctl.sh call out to the functions in this script?

One time setup functions can be called during the init phase; e.g.
  1. call install_kubectl_minikube
  1. infer_minikube_settings - this should be called during init or generate
      * The values should be persisted to a file e.g. env.sh
  1. cleanup_and_deploy_minikube this should be called during kfctl.sh apply
      * env.sh will be sourced so any environment variables defined in that file will be loaded and available.

scripts/setup-minikube.sh
```diff
@@ -0,0 +1,354 @@
...
+
+  #download_kfctl_scripts
+  KUBEFLOW_REPO=/usr/local/google/home/abhishek/kubeflow ./kfctl.sh init localapp --platform minikube
+  pushd .
+  cd localapp
+  ../kfctl.sh generate all
+  ../kfctl.sh apply all
+  popd
+
+  if is_kubeflow_ready
```
jlewi: Per other comments; if we call this from kfctl.sh then most of the functionality up to this line should already be handled by kfctl.sh

abhi-g:
@jlewi I think its ready for review now.

Updated based on your comments.

- kfctl init  phase now will read in required parameters from user, install ks, kubectl, minikube if needed and start minikube.
- kfctl generate phase will create the local volume claim specs in the ksonnet dir, and setup params for jupyterhub
- kfctl apply phase will deploy everything, wait for services to get ready, mount the local fs, setup tunnels.

When a new notebook is spawned it should automatically mount the local fs directory if specified and shows up as local-notebooks alongside work directory.

I haven't been able to test the parameterizing the jupyterhub ksonnet part. Since the script pulls all the sonnets etc. from github, I will retest after this is merged and send any fixes if needed.

kubeflow/core/prototypes/jupyterhub.jsonnet
```diff
@@ -13,6 +13,8 @@
 // @optionalParam repoName string kubeflow-images-public The repository name for JupyterNotebook.
 // @optionalParam disks string null Comma separated list of Google persistent disks to attach to jupyter environments.
 // @optionalParam gcpSecretName string user-gcp-sa The name of the secret containing service account credentials for GCP
+// @optionalParam notebookUid string 1000 UserId of the host user for minikube local fs mount
```
jlewi: Are we changing the default behavior?  It doesn't look like KubeFormSpawner was setting the UID/GID before but now it is.

kubeflow/core/kubeform_spawner.py
```diff
@@ -177,6 +177,26 @@ def _expand_user_properties(self, template):
...
 volumes = []
 volume_mounts = []
+
+# Allow environment vars to override uid and gid.
+# This allows local host path mounts to be read/writable
+env_uid = os.environ.get('NOTEBOOK_UID')
+if env_uid and env_uid != 'null':
+    c.KubeSpawner.singleuser_uid = int(env_uid)
+env_gid = os.environ.get('NOTEBOOK_GID')
+if env_gid and env_gid != 'null':
```
jlewi: Won't this break by default? It looks like you are defaulting env_gid to 100. So you will now add a post pod hook that tries to mount local notebooks into the container and that command will presumably fail.

kubeflow/core/kubeform_spawner.py
```diff
@@ -177,6 +177,26 @@ def _expand_user_properties(self, template):
...
+    c.KubeSpawner.singleuser_fs_gid = int(env_gid)
+    def modify_pod_hook(spawner, pod):
```
jlewi: Should we guard the lifecycle with its own hook? Maybe define an environment variable for the directory containing the notebooks.

I worry it might be brittle to infer this based on gid.

kubeflow/core/kubeform_spawner.py
```diff
@@ -177,6 +177,26 @@ def _expand_user_properties(self, template):
 c.KubeSpawner.singleuser_working_dir = '/home/jovyan'
 volumes = []
 volume_mounts = []
+
+# Allow environment vars to override uid and gid.
```
jlewi: @pdmack Would you mind taking a look at this code? I remember it taking us a while to get uid and gid to work correctly when using a PV backing the home directory?

kubeflow/core/kubeform_spawner.py
```diff
@@ -177,6 +177,26 @@ def _expand_user_properties(self, template):
...
+# This allows local host path mounts to be read/writable
+env_uid = os.environ.get('NOTEBOOK_UID')
```
jlewi: @abhi-g How does this compare to the current behavior? i.e. if you run on GKE with a PV for /home/jovyan how does this code changing the UID and GID change what user and group we run the notebook as?

We had a lot of trouble getting the permissions correct when using a PV for the home directory.
The thing to do would be to run it on GKE with the default settings and verify that the user had RW access to the home directory /home/jovyan

scripts/kfctl.sh
```diff
@@ -15,9 +15,14 @@ WHAT=$2
 
 ENV_FILE="env.sh"
 
+KUBEFLOW_VERSION=${KUBEFLOW_VERSION:-"master"}
```
jlewi: Why do you need KUBEFLOW_VERSION?

scripts/kfctl.sh
```diff
@@ -15,9 +15,14 @@ WHAT=$2
...
+KUBEFLOW_DEPLOY=${KUBEFLOW_DEPLOY:-true}
```
jlewi: We shouldn't need KUBEFLOW_DEPLOY anymore

scripts/kfctl.sh
```diff
@@ -15,9 +15,14 @@ WHAT=$2
...
+K8S_NAMESPACE=${K8S_NAMESPACE:-"kubeflow"}
+KUBEFLOW_CLOUD=${KUBEFLOW_CLOUD:-"minikube"}
```
jlewi: Can we get rid of KUBEFLOW_CLOUD? This should be provided by the parameter --platform passed on kfctl init

scripts/kfctl.sh
```diff
@@ -15,9 +15,14 @@ WHAT=$2
 
 ENV_FILE="env.sh"
 
+KUBEFLOW_VERSION=${KUBEFLOW_VERSION:-"master"}
+KUBEFLOW_DEPLOY=${KUBEFLOW_DEPLOY:-true}
+K8S_NAMESPACE=${K8S_NAMESPACE:-"kubeflow"}
```
jlewi: Why do we need K8S_NAMESPACE to be provided as an environment variable?
This should probably be a parameter of kfctl init if its really needed

scripts/kfctl.sh
```diff
@@ -35,6 +40,16 @@ createEnv() {
 
 	if [ "${PLATFORM}" == "minikube" ]; then
 	  echo KUBEFLOW_CLOUD=minikube >> ${ENV_FILE}
+    set +x
+    infer_minikube_settings
+    set -x
+    download_kubeflow_source
```
jlewi: Why are you calling download_kubeflow_source here?
The needed source will already be there; source is downloaded by
https://github.com/kubeflow/kubeflow/blob/master/scripts/download.sh

The source is located based on the location of kfctl.sh

Once you get rid of it it will use your local code and you can test it.

scripts/kfctl.sh
```diff
@@ -35,6 +40,16 @@ createEnv() {
...
+    echo HOST_PLATFPORM=${HOST_PLATFPORM} >> ${ENV_FILE}
+    echo MOUNT_LOCAL=${MOUNT_LOCAL} >> ${ENV_FILE}
+    echo MINIKUBE_CMD=\"${MINIKUBE_CMD}\" >> ${ENV_FILE}
+    echo KUBEFLOW_REPO="$(pwd)/kubeflow-${KUBEFLOW_VERSION}" >> ${ENV_FILE}
```
jlewi: KUBEFLOW_REPO is already defined on line 24

scripts/kfctl.sh
```diff
@@ -92,6 +107,17 @@ createEnv() {
   fi
 }
 
+# For minikube single script download experience
+function download_kfctl_scripts() {
```
jlewi: You can delete this;
We have a separate script to do the download.sh
https://github.com/kubeflow/kubeflow/blob/master/scripts/download.sh

scripts/kfctl.sh
```diff
@@ -112,6 +138,16 @@ if [ "${COMMAND}" == "init" ]; then
       esac
       shift
 	done
+
+  # Download minikube dependencies
+  if [ "${PLATFORM}" == "minikube" ]; then
+    download_kfctl_scripts
+  fi
+
+  source "${DIR}/util.sh"
```
jlewi: Move this back to the top; they will already exist because kfctl.sh doesn't download the scripts.

scripts/kfctl.sh
```diff
@@ -112,6 +138,16 @@ if [ "${COMMAND}" == "init" ]; then
       esac
       shift
 	done
+
+  # Download minikube dependencies
```
jlewi: You can delete this.

scripts/kfctl.sh
```diff
@@ -148,6 +184,9 @@ if [ "${COMMAND}" == "init" ]; then
 fi
 
 source ${ENV_FILE}
+source "${DIR}/util.sh"
```
jlewi: Didn't we already source the files?

scripts/kfctl.sh
jlewi: This is already defined in CREATE_ENV

scripts/util-minikube.sh
```diff
@@ -0,0 +1,320 @@
...
+fi
+
+MINIKUBE_CMD="minikube start"
+MOUNT_LOCAL=false
+
+GB1=$(( 1024 * 1024 * 1024 ))
+MB1=$(( 1024 * 1024 ))
+KB1=$(( 1024 ))
+
+function install_ks_kubectl_minikube() {
```
jlewi: Should we make this a separate script?

scripts/util-minikube.sh
```diff
@@ -0,0 +1,320 @@
...
+  echo -n "If you'd like to access a local directory in JupyterHub, please enter the full path: "
+  read l
+  if [[ $l != "" ]]
+  then
+    MINIKUBE_CMD=$MINIKUBE_CMD" --mount-string=$l:/mnt/local --mount"
+    MOUNT_LOCAL=true
+  fi
+}
+
+function download_kubeflow_source() {
```
jlewi: No longer needed handled by separate script download.sh

kubeflow/core/prototypes/jupyterhub.jsonnet
abhi-g: Yes, it is doing this now only in minikube case. This is to make sure that the locally mounted filesystem is read/writable via Jupyter Notebook

kubeflow/core/kubeform_spawner.py
abhi-g: You are correct. Updating above default values of params to null

kubeflow/core/prototypes/jupyterhub.jsonnet
abhi-g: updating to nulls.

kubeflow/core/kubeform_spawner.py
abhi-g: Added a separate env var for this.

kubeflow/core/kubeform_spawner.py
abhi-g: By the way, I have been able to test that I can access the local directory when specified via the Jupyter Notebook and create/modify notebooks.
Also, if not specified, the notebook behaves as earlier. 
I haven't tested after creating the corresponding jsonnet params.

kubeflow/core/kubeform_spawner.py
abhi-g: I'll definitely test that. Except, last time I tried, I still get stuck with IAM issues on GKE

scripts/kfctl.sh
abhi-g: To download the appropriate repository archive.

scripts/kfctl.sh
abhi-g: removed

scripts/kfctl.sh
abhi-g: removed.

scripts/kfctl.sh
abhi-g: This needs to be set here, because createEnv is called after downloading corresponding shell scripts that are needed by kfctl.sh.

scripts/kfctl.sh
abhi-g: removed K8S_NAMESPACE here as well

scripts/kfctl.sh
abhi-g: For the local experience, I don't like the idea of a user downloading the kubeflow source first. In fact, in its current form, they only need to download kfctl.sh, and then:
kfctl.sh init appName --platform minikube
cd appName
../kfctl.sh generate all
../kfctl.sh apply all

scripts/kfctl.sh
abhi-g: removed the irrelevant comment line.

scripts/kfctl.sh
abhi-g: This was restructured, hence the need to move the source statements.

scripts/util-minikube.sh
abhi-g: Let's discuss how to structure it. I am really looking for a single entry point script that does all of the needed installs etc. which is what the original setup-minikube.sh was. It looks like we need to build out two different experiences: 
1. straightforward one script does it all
2. With all the dev setup similar to GKE, ie. kubeflow repo etc. being present, kfctl to deploy on minikube with its split phases.

If so, I'd need to do another refactoring, moving around bunch of this stuff.

My primary goal is to get #1 ready as described in the local CUJ

abhi-g:
Pushed the current fixes based on your comments. But I think I need another
restructuring to separate out the minikube installation etc.

jlewi:
Just a couple nits and then this should be ready to go.

abhi-g:
Addressed your last couple of comments @jlewi

abhi-g:
adding back lgtm label. Only fixed a 1 line conflict with upstream. 

/lgtm

abhi-g:
Good to know @k8s-ci-robot

kubeflow/core/kubeform_spawner.py
pdmack: Sorry, missed this. Merged but I'll take a closer look this weekend anyway.

PR #1395, opened 2018-08-21
util.libsonnet has methods that allow a <component>.part to be modified

### LINKED ISSUE #1231: provide the ability to add imagePullSecrets to different ServiceAccounts so that private images can be fetched
ServiceAccounts should have an imagePullSecrets that references a docker-registry name when the image is coming from a different registry or different gcr.io account.

For example for jupyterhub - adding a parameter jupyterHubImageSecret which is then added to the jupyterhub ServiceAccount as 
```yaml
imagePullSecrets: 
- name: gcr-accountfoo
```
would reference this secret. The secret itself could be added to deploy.sh so that it does something like 
```bash
kubectl create secret docker-registry --namespace=${KUBEFLOW_NAMESPACE} gcr-accountfoo --docker-server=https://gcr.io --docker-username=_json_key --docker-email=${SA_EMAIL} --docker-password="$(cat ./${SA_EMAIL}.json)"
```

kkasravi:
Fixes #1231 

After `kfctl.sh generate k8s` the user would update components/jupterhub.jsonnet 

```
local util = import "kubeflow/core/util.libsonnet";
util.update(jupyterhub, updatedParams, 'StatefulSet/tf-hub',
  {
    template+: {
      spec+: {
        imagePullSecrets: [
          {name: "mySecret",}
        ],
      },
    },
  }
)
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1395)


jlewi:
See comment here:
https://github.com/kubeflow/kubeflow/issues/1231#issuecomment-413869590

It looks like the pattern you are suggesting is to have parts define a list and then convert that to a map so you can modify an element by name.  That's a useful function.

But why wouldn't we just rewrite the libsonnet file to be a map?

e.g suppose the libsonnet file looked like

```
{
    statefulset :: {
     ...
   }

   parts :: [$.statefulset, ....]
}
```

Then I think in the .jsonnet file you could do something like

```
local updated = jupterhub +  {
   statfulset +: {
        template+: {
      spec+: {
        imagePullSecrets: [
          {name: "mySecret",}
        ],
      },
    },
  }
 },
)

std.prune(updated.parts)
```

kubeflow/core/util.libsonnet
```diff
@@ -31,5 +33,39 @@
       else [],
   }.result,
 
+  // Convert an object's properties into an array of values
+  toList:: [$[key] for key in std.objectFields($)],
+
+  // Convert an array of kubernetes resources into a  map[kind/name, resource]
+  toMap:: function(params) {
```
jlewi: Why is toMap taking params as an argument?

kkasravi: I believe I understand. Does the following make sense?

1. Remove params from toMap ([prometheus.libsonnet](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/prometheus.libsonnet) is an example where parts is a map and all is parts)
2. Build the map first, then the parts which reference the map eg: [map.values()]  

Question: What about repeating resources within a libsonnet? For example jupyterhub.libsonnet has 2 Services, 2 ServiceAccounts, 2 Roles, 2 RoleBindings. Should we deal with it only if needed? For example:
```
local jup = {
  Service: {
    "tf-hub-0": {foo:"bar",},
    "tf-hub-lb": {bar:"baz",},
  },
};
```
so the update would look like
```
jup +  {
   Service+: {
    "tf-hub-0"+: {one: "two",},
  },
}
```
yielding
```
{
   "Service": {
      "tf-hub-0": {
         "foo": "bar",
         "one": "two"
      },
      "tf-hub-lb": {
         "bar": "baz"
      }
   }
}
```
for non-repeating resources it would look like your example with StatefulSet or would that be too confusing?

jlewi: I don't think the top level items in the dictionary should be based on type. They should be based on what they do; e.g. if there are multiple services each service should be its own item.

kkasravi: so how bout
```
{
  ConfigMap:: 
  HubService::
  HubServiceAccount::
  HubRole::
  HubRoleBinding::
  NotebookService::
  NotebookServiceAccount::
  NotebookRole::
  NotebookRoleBinding::  
  StatefulSet::
}
```

where the map is a set of functional hidden names

kkasravi: parts would be in the top level mapping
```
{
  ConfigMap:: 
  HubService::
  HubServiceAccount::
  HubRole::
  HubRoleBinding::
  NotebookService::
  NotebookServiceAccount::
  NotebookRole::
  NotebookRoleBinding::  
  StatefulSet::

  parts:: [ ConfigMap:: HubService:: ... ]
}

**Comment by kkasravi** [comment]
@jlewi 

This now works 
```
  StatefulSet+: {
     spec+: {
       template+: {
         spec+: {
          imagePullSecrets: [
            {name: "mySecret",}
          ],
        },
      },
    },
  },
};
updated.all(updatedParams).list
```

I can update the other libsonnets if this looks ok. Note that params is now references as $.params within the resources.

File: kubeflow/core/jupyterhub.libsonnet [hunk 2]
```diff
@@ -1,342 +1,345 @@
...
+  all(_params):: self + {
+    params+: _params,
+  },
+  list:: [
+    self[key]
+    for key in std.objectFieldsAll(self)
+    if key != "list" && key != "params" && key != "all"
+  ],
+
+  ConfigMap:: {
```
jlewi: nit: Change the name and StatefulSet name to be more descriptive 
e.g
ConfigMap -> KubeSpawnerConfig

StatefulSet -> HubStatefulSet

jlewi:
Looks great to me. I had one nit about the names but otherwise this is ready IMO.

kkasravi:
thanks @jlewi

PR #1327, opened 2018-08-07
Fix iap-ingress certificate creation and use BackendConfig for IAP

danisla:
- Added k8s job for patching ingress with the TLS spec after the certificate is acquired.
- Refactored IAP enabler to use BackendConfig to enable IAP. Script still has to set timeout and health check. BackendConfig is only available on GKE __1.10.5__.
- Updated cert-manager to v0.4.0 which no longer depends on container for ingress-shim. 

Fix #1146 Use BackendConfig to enable IAP

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1327)

PR #1396, opened 2018-08-21
Updates to iap component to support private clusters

ankushagarwal:
Docs changes are here : https://github.com/kubeflow/website/pull/173

* Creates a new component: Ingress Setup
* Updates all various ingress related bash scripts to use Ingress Setup docker image
* Don't create certificate CRD when using private cluster
* Add `privateGKECluster` and `ingressSetupImage` flags to iap prototype

/cc @jlewi @danisla

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1396)

### DIFF
--- a/kubeflow/core/configure_envoy_for_iap.sh
+++ b/kubeflow/core/configure_envoy_for_iap.sh
@@ -7,10 +7,6 @@
 [ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
 [ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
 
-apk add --update jq
-curl https://storage.googleapis.com/kubernetes-release/release/v1.9.4/bin/linux/amd64/kubectl > /usr/local/bin/kubectl && chmod +x /usr/local/bin/kubectl
-
-
 PROJECT=$(curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/project/project-id)
 if [ -z ${PROJECT} ]; then
 echo Error unable to fetch PROJECT from compute metadata
@@ -54,7 +50,7 @@ curl -s ${ENVOY_ADMIN}/quitquitquit
 
 function checkIAP() {
 # created by init container.
-. /var/shared/healthz.env 
+. /var/shared/healthz.env
 
 # If node port or backend id change, so does the JWT audience.
 CURR_NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[0].nodePort}')
--- a/components/ingress-setup-image/Dockerfile
+++ b/components/ingress-setup-image/Dockerfile
@@ -1,0 +1,5 @@
+FROM google/cloud-sdk:alpine
+
+RUN apk add --update jq openssl
+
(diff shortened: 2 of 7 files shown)


components/iap-enabler/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM google/cloud-sdk:alpine
+
+RUN apk add --update jq
```
danisla: We should also move the curl to install kubectl from the shell script to the Dockerfile.

kubeflow/core/setup_iap.sh
```diff
@@ -1,12 +1,12 @@
...
 #
 # A simple shell script to enable IAP and configure timeouts by using gcloud.
+
 [ -z ${CLIENT_ID} ] && echo Error CLIENT_ID must be set && exit 1
 [ -z ${CLIENT_SECRET} ] && echo Error CLIENT_SECRET must be set && exit 1
 [ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
 [ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
 
-apk add --update jq
 curl https://storage.googleapis.com/kubernetes-release/release/v1.9.4/bin/linux/amd64/kubectl > /usr/local/bin/kubectl && chmod +x /usr/local/bin/kubectl
```
danisla: Add this curl to the Dockerfile and delete from this script.

kubeflow/core/configure_envoy_for_iap.sh
```diff
@@ -9,7 +9,6 @@
 [ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
 [ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
 
-apk add --update jq
 curl https://storage.googleapis.com/kubernetes-release/release/v1.9.4/bin/linux/amd64/kubectl > /usr/local/bin/kubectl && chmod +x /usr/local/bin/kubectl
```
danisla: Add this curl to the Dockerfile and delete from this script. Also bump it to v1.11.0.

ankushagarwal:
Done

components/iap-enabler/README.md
```diff
@@ -0,0 +1,10 @@
+# IAP Enabler
+
+IAP Enabler is a docker image which is used to enable iap. We build this and push it to gcr so that it can be used in private GKE clusters
```
jlewi: I think we use BackendConfig now to enable IAP. I think this is only used to set the timeout on the loadbalancer.

jlewi:
nit: please update PR description to describe changes that are needed.

components/iap-enabler/README.md
ankushagarwal: I'll rebase this once https://github.com/kubeflow/kubeflow/pull/1327 gets merged

ankushagarwal:
This is updated and ready for review

danisla:
@ankushagarwal it looks like we now omit installing the Certificate resource for private clusters, should we also omit installing the cert-manager component for private clusters since it won't be used?

ankushagarwal:
I've added it to docs here : https://master.kubeflow.org/docs/started/getting-started-gke/ `Remove components which are not useful in private clusters`

danisla:
Great! 

/LGTM

PR #1397, opened 2018-08-21
Add namespaces to the tf-job-dashboard role

ankushagarwal:
This fixes the 500 errors in tf-job-dashboard

Related to kubeflow/tf-operator#754

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1397)


ankushagarwal:
It fixes the 500s when tfjob is being created. I'll have to check if this fixes the other issue mentioned in the bug where `Name, Status, Logs` field values are missing

PR #1399, opened 2018-08-22
Fix annotations in ambassador and centraldashboard

### LINKED ISSUE #1306: CentralDashboard returns 404; Ambassador can't parse the route
On master Ambassador shows the following error trying to parse the route for the central dashboard

```
2018-08-03 06:35:54 diagd 0.30.1 ERROR: k8s-dashboard-kubeflow.yaml.1: <RichStatus BAD error=k8s-dashboard-kubeflow.yaml: could not parse YAML hostname=ambassador-7d8f87f6dc-2mbk2 resolvedname=10.20.0.10 version=0.30.1>
```
As a result no mapping is defined and trying to access the central dashboard behind Ambassador gives a 404.

This is ksonnet/ksonnet#670

We should be able to workaround this.

richardsliu:
Fix #1306

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1399)

### DIFF
--- a/kubeflow/core/ambassador.libsonnet
+++ b/kubeflow/core/ambassador.libsonnet
@@ -1,4 +1,8 @@
 {
+  // Due to https://github.com/ksonnet/ksonnet/issues/670, escaped characters in
+  // jsonnet files are not interpreted correctly by ksonnet, which causes runtime
+  // parsing failures. This is fixed in ksonnet 0.12.0, so we can merge this back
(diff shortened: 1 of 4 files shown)


jlewi:
nit: Can you add comments in both .libsonnet files explaining why we do this and the fact that we can remove it once we are willing to depend on ksonnet 0.12

richardsliu:
Done.

Issue #1331, opened 2018-08-08
Update E2E test to use kfctl.sh and delete gke/deploy.sh;


jlewi:
We want to delete the deploy.sh scripts
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh

It is replaced by kfctl.sh

deploy.sh is still being used by our E2E test
https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/workflows.jsonnet

We added a new E2E workflow for kfctl.sh
https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_test.jsonnet
but this test isn't running any validation steps (e.g. running a PyTorch or TFJob test).

Per #1325 we'd like to refactor our tests to make subgraphs/steps more reusable

jlewi:
kubeflow/kubeflow#1342

Updated our tests to use kfctl.sh

deploy.sh is still being called by deploy_kubeflow_gcp.sh which is referenced in workflows.jsonnet.

I don't think its still being used though; so we could probably just delete it and remove the references.

jlewi:
Post-submit run:
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_chainer-operator/21/kubeflow-chainer-operator-presubmit/57/

Looks like minikube broke because it depends on deploy_kubeflow.py which we deleted in #1406.

jlewi:
Post submits are green:
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-postsubmit

jlewi:
gke/deploy.sh wasn't actually deleted.
see #1481

PR #1406, opened 2018-08-23
Remove deploy_gcp.sh and update workflows.libsonnet.

jlewi:
* These scripts are no longer used to deploy KF on GKE in tests; we now
  use kfctl.

Related to: #1331

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1406)


jlewi:
Test failure is about the async keyword; should be fixed by kubeflow/tf-operator#790

jlewi:
kubeflow/tf-operator#790 is now submitted; should fix TF upgrade.

/test all

Issue #1401, opened 2018-08-22
Don't use DM for IAM policy management


jlewi:
Right now our DM config [cluster.jinja](
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster.jinja#L161)

Adds a bunch of IAM role bindings.

This creates two problems

1. The optimistic locking on setting IAM roles can block DM creation. It also makes deleting deployments really complicated because we need to do an update before doing a delete.
1. it requires give the DM service account the ability to set project IAM which significantly escalates its permission.

We could create a separate DM resource to set IAM roles; this would partially address #1 but #2.

I think to avoid granting the DM elevated permissions we should stop using DM to create appropriate role bindings.

We would still like to be declarative though in the way we manage role bindings.

I think there's a couple ways to do this

1. During the generate phase of kfctl we could create add and remove shell scripts that would call out to gcloud add-iam-policy-binding and gcloud remov-iam-policy-binding

1. We could define an IAM policy patch in JSON/YAML and then create a python script to apply it.

kunmingg:
Agree, current IAM role manage logic is fragile.

ankushagarwal:
IAM management is problematic in DM, but I don't think that either #1 or #2 is a good idea because we will end up creating our own deployment manager for declaratively managing IAM configs.

jlewi:
@ankushagarwal is that duplication less desirable than the current situation?

In particular, granting DM ability to set IAM policies seems highly undesirable. 

In current situation, we need to create our own delete wrapper because we can't issue a DM delete without issuing an update first.  This has a bunch of drawbacks; including 
   * not being able to easily delete via U
   * If a DM is partially deleted eg. cluster is deleted; doing an update will recreate the resources before doing the delete.

We already need to implement our own delete script to properly clean up the policy bindings.
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/delete_role_bindings.py

ankushagarwal:
I guess it makes sense if we have already added code to manage IAM policies outside of DM.

ankushagarwal:
FYI I've started taking a stab at this : https://github.com/kubeflow/kubeflow/pull/1408/

I'm going with suggestion #2

PR #1408, opened 2018-08-23
Python script to declaratively manage IAM binding patches

ankushagarwal:
* Remove IAM related config from cluster.jinja
* Add `iam_bindings_template.yaml` - this is a template which is used to generate `iam_bindings.yaml`. `iam_bindings.yaml` is used to declaratively manage IAM bindings
* Add `scripts/gke/iam_patch.py` - a python script which uses `iam_bindings.yaml` to declaratively manage IAM bindings
  * `scripts/gke/iam_patch.py` is idempotent and retries if IAM updates fail
  * It uses a single API call to update all bindings instead of going through the bindings one by one
  * Add a simple unit test
* Update `scripts/gke/util.sh` and `scripts/gke/delete_deployment.sh` to use `scripts/gke/iam_patch.py`
* Delete `scripts/gke/delete_role_bindings.py`

Fixes https://github.com/kubeflow/kubeflow/issues/1401

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1408)


scripts/gke/iam_patch.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+if __name__ == "__main__":
+  logging.getLogger().setLevel(logging.INFO)
+  args = parse_args()
+
+  if args.action not in ["add", "remove"]:
+    raise ValueError("invalid --action. Valid values are add, remove")
+
+  for i in range(5):
```
jlewi: Can we use the retrying module?

ankushagarwal: I intentionally tried to avoid using `retrying` module because this script will be run on an end-users computer and I wanted to avoid an additional step of doing a `pip install`

ankushagarwal:
Addressed all issues. PTAL

jlewi:
Mostly nit picky things;

Can you also remove this line so we no longer grant setIam permission to the cloud services account?

scripts/gke/deployment_manager_configs/iam_bindings_template.yaml
```diff
@@ -0,0 +1,40 @@
+# This config is used by iam_patch.py. It is not a DM config.
+bindings:
+- members:
+  - set-kubeflow-admin-service-account
+  roles:
+  # Grant permissions needed to push the app to a cloud repository
+  - roles/source.admin
```
kunmingg: Can we follow binding schema defined here: https://cloud.google.com/iam/reference/rest/v1/Policy?
Each binding has one role and a list of members.

Ark-kun:
@ankushagarwal @jlewi 
Does the test script verify the failure handling behavior?

PR #1411, opened 2018-08-24
Fix minikube test and trigger on presubmit when modified.

### LINKED ISSUE #1350: Trigger minikube E2E test on presubmit when minikube test is modified
Right now we only run minikube E2E test on postsubmit.

But we should run the minikube test when the minikube test (i.e. the .jsonnet file defining the Argo workflow test) changes on presubmit so that we know the test is valid.

jlewi:
* In #1406 we deleted deploy_kubeflow.py but this file is still used
  by the minikube test.

* Partially reverts commit a4e4a8e9caf7bfa40958d9b8c9094fd7f9c7534e.

* Related to #1331

* Trigger minikube on presubmits if files used by minikube E2E test are changed.
  run_e2e_workflow.py uses fnmatch so we can specify individual files
  or other expressions

* Fix #1350 trigger minikube test on presubmit when test modified.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1411)

### DIFF
--- a/testing/deploy_kubeflow.py
+++ b/testing/deploy_kubeflow.py
@@ -60,8 +60,6 @@ def deploy_kubeflow(test_case):
     [
       "ks", "generate", "tf-job-operator", "tf-job-operator",
       "--namespace=" + namespace,
-      "--tfJobImage=gcr.io/kubeflow-images-public/tf_operator:v20180522-77375baf",
-      "--tfJobVersion=v1alpha1"
     ],
     cwd=app_dir)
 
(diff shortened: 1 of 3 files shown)


testing/deploy_kubeflow.py
```diff
@@ -0,0 +1,127 @@
+"""Deploy Kubeflow and wait for it to be deployed.
+
+TODO(jlewi): This script is outdated. Its no longer used for GKE.
+It is still used by minikube. For minikube we should be using kfctl.sh
```
richardsliu: Did you mean "for GKE we should be using kfctl.sh"?

jlewi: No I meant we should switch to using kfctl.sh for our minikube E2E test; at which point we shouldn't need this script anymore. GKE already uses kfctl.sh to deploy so its not using this script.

PR #1423, opened 2018-08-25
Use JupyterLab as default instead of Jupyter Tree interface

ankushagarwal:
This is configurable via the ksonnet parameter - useJupyterLabAsDefault

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1423)

### DIFF
--- a/kubeflow/core/kubeform_spawner.py
+++ b/kubeflow/core/kubeform_spawner.py
@@ -114,7 +114,7 @@ class KubeFormSpawner(KubeSpawner):
         return env
 
     # TODO(kkasravi): add unit test
-    def _parse_user_name(self, username): 
+    def _parse_user_name(self, username):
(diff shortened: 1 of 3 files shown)


kubeflow/core/kubeform_spawner.py
```diff
@@ -237,6 +237,9 @@ def _expand_user_properties(self, template):
 else:
     c.JupyterHub.authenticator_class = 'dummyauthenticator.DummyAuthenticator'
 
+if os.environ.get('DEFAULT_JUPYTERLAB') == 'true':
```
jlewi: Can you ignore case please so that "TRUE" will also work?

ankushagarwal: Done

PR #1425, opened 2018-08-25
Fix kfctl.sh gcpInitProject is always skipped

sambaiz:
- Fix if statements

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1425)


pdmack:
Not sure about those test failures but could you rebase your changes?

PR #1427, opened 2018-08-25
Add logging statements to help figure out why test results aren't reported

### LINKED ISSUE #1426: tf_job_simple_test results not being report
Here's a postsubmit run
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/kubeflow_kubeflow/kubeflow-postsubmit/619

There are 10 passed tess
test_jsonnet
deploy-kubeflow-deploy_argo-test-argo-deploy
deploy-kubeflow-deploy_minikube
deploy-kubeflow-deploy_model-mnist-cpu
deploy-kubeflow-deploy_pytorchjob-pytorch-job
deploy-kubeflow-teardown
deploy-kubeflow-teardown_minikube
smoke-tfjob-gke
test_jsonnet_formatting test_jsonnet_formatting
tf-serving-image-mnist-cpu

1 Failed test
deploy-kubeflow-deploy_model-mnist-gpu

There is no report for the simple TFJob prototype test
Looking at Argo it looks like the test ran
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-postsubmit-kfctl-a41ba72-619-0fc2?tab=workflow&nodeId=kubeflow-postsubmit-kfctl-a41ba72-619-0fc2-203334643

There's no indication that the test completed successfully; i.e. 
https://github.com/kubeflow/kubeflow/blob/master/testing/tf_job_simple_test.py#L111
we should print out "TFJob launched successfully."

There's also no indication in the logs that we saved the results/failure to GCS as an example file

jlewi:
Fix the E2E test for the TFJob simple prototype test.

* Fix #1426

There are two problems with the test

  1. Test isn't properly reporting results to gubernator; so test failures
     aren't being noticed.
  2. Test needs to be updated to work with v1alpha2.

* The TestSuite name needs to be set because this is used as the name
  of the junit XML file.

* simple-prototype-test should set test_dir and artifacts_dir.

* Fix the test; use tf_job_client to wait for the job to be in the Running
  condition. This should be more reliable than checking for actual pods.

* The test has probably been broken for a while but this went unnoticed
  because results weren't being properly surfaced in test grid because
  the XML file is improperly named. I suspect things broke as part of
  the switch to v1alpha2 which changed the names of the pods.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1427)

### DIFF
--- a/testing/tf_job_simple_test.py
+++ b/testing/tf_job_simple_test.py
@@ -25,7 +25,10 @@ import logging
 import os
 import re
 import subprocess
+
+from kubernetes import client as k8s_client
 from kubeflow.testing import test_helper, util
+from py import tf_job_client
 from retrying import retry
 
 NAMESPACE = "default"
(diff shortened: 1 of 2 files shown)


jlewi:
/test all

Filed #1436 for the test flake

PR #1434, opened 2018-08-27
Weaveflux version 0.1.2. 

TheJaySmith:
My last PR somehow lost data in rebase. Attempt to fix this

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1434)

### DIFF
--- a/kubeflow/weaveflux/prototypes/weaveflux.jsonnet
+++ b/kubeflow/weaveflux/prototypes/weaveflux.jsonnet
@@ -4,7 +4,7 @@
 // @shortDescription A Flux meets Kubeflow
 // @param name string Name to give to each of the components
 // @optionalParam giturl string null Your default git URL.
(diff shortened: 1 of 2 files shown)


kubeflow/weaveflux/prototypes/weaveflux.jsonnet
```diff
@@ -4,7 +4,7 @@
 // @shortDescription A Flux meets Kubeflow
 // @param name string Name to give to each of the components
 // @optionalParam giturl string null Your default git URL.
-// @optionalParam serviceType string ClusterIP The service type for WeaveFlux.
+// @optionalParam serviceType string NodePort The service type for WeaveFlux.
```
pdmack: I think this is the same as @jlewi point before. We should default to ClusterIP.

PR #1437, opened 2018-08-28
standardize remaining <component>.{jsonnet,libsonnet} files 

### LINKED ISSUE #1414: standardize remaining <component>.{jsonnet,libsonnet} files
jupyterhub.{[jsonnet](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/prototypes/jupyterhub.jsonnet),[libsonnet](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/jupyterhub.libsonnet)} allows one to easily add, update or remove resources or parts of resources. For example if one wanted to add imagePullSecrets to jupyterhub's HubStatefulSet one would do:

```
  HubStatefulSet+: {
     spec+: {
       template+: {
         spec+: {
          imagePullSecrets: [
            {name: "mySecret",},
          ],
        },
      },
    },
  },
};
updated.all(updatedParams).list
```

This pattern should be standardized in the other components

kkasravi:
fixes #1414

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1437)


kubeflow/core/centraldashboard.libsonnet
```diff
@@ -1,18 +1,151 @@
...
-        "---",
-        "apiVersion: ambassador/v0",
-        "kind:  Mapping",
-        "name: centralui-mapping",
-        "prefix: /",
-        "rewrite: /",
-        "service: centraldashboard." + namespace,
-      ]),
+  local util = import "kubeflow/core/util.libsonnet",
+  new(params):: self + util + {
```
jlewi: What's the reason we set new to self + util?

kubeflow/core/util.libsonnet
```diff
@@ -31,5 +41,11 @@
       else [],
   }.result,
 
-
+  local e = self,
+  local k = import "k.libsonnet",
```
jlewi: Move import to top of file?

kubeflow/core/util.libsonnet
```diff
@@ -31,5 +41,11 @@
...
+  list:: std.prune(k.core.v1.list.new([
```
jlewi: Could we make this and then in the .libsonnet files just define a list property that is the result of calling this function?

Would that lead to infinite recursion?

jlewi: I wonder if we'd be better off just manually defining the list as opposed to auto-constructing it from all the keys in the dict?

Its possible in the future we'd want to have more than 1 prototype defined in a .libsonnet file composed of different subsets of elements defined in the .libsonnet file. In which case we wouldn't want to add all the iems in the dictionary.

kkasravi: There is an example of this in spartacus.libsonnet of a 'prototype' (see AllResources::) and i'm also converting tensorboard.libsonnet to use these patterns. Tensorboard is a great use case since it has {params, gcpParams, s3params} and {parts, s3parts, gcpParts}. But to answer your question:
> I wonder if we'd be better off just manually defining the list as opposed to auto-constructing it from all the keys in the dict?

For organization by 'prototype' (or grouping) you basically do:
```
{
  PrototypeA:: if params.emitPrototypeA then (self + {
    ConfigmapA: {
    },
    ServiceA: {
    },
  }),
  PrototypeB:: if params.emitPrototypeB then (self + {
    ConfigmapB: {
    },
    ServiceB: {
    },
  }),
}
```

This still allows auto-construction of lists and may be better than how tensorboard is currently grouping prototypes (which is quite clever). I think a side-by-side comparison of this pattern versus tensorboard's current approach will be helpful. 

Your other question:
> Could we make this and then in the .libsonnet files just define a list property that is the result of calling this function? Would that lead to infinite recursion?
 
All lower case fields (leading character) including hidden functions are ignored in util.list - similar to golang where exported fields have a leading upper character. It's also difficult to access util's list unless you do it explicitly (eg something like List:: util.list) since list is not added to self until 'new' returns. 

I understand why you are suggesting explicit lists instead of auto-construction (opaqueness, risk of infinite recursion). I'm hoping that we can eliminate current "bad" practices sneaking into components that involve adding params, env vars, etc (which you've suggested is a goal of this work). This is why i'm using certain conventions like upper/lower in lieu of being explicit. This auto-construction may do no more than obfuscate manifest generation. I think tensorboard will be a good use case to examine to see if these patterns (including auto-construction) are making things easier, harder or just plain confusing.

Would it be ok to wait for my update of tensorboard.libsonnet?

kkasravi:
@ankushagarwal @jlewi any idea where k.libsonnet is? - I've tried adding a few additional srcDirs in workflows/components/unit_tests.jsonnet but test_util.jsonnet still can't load k.libsonnet.

kubeflow/core/centraldashboard.libsonnet
jlewi: Why use a mix in rather than just calling those functions on util when you need them?

jlewi:
```
find ./ -name "k.libsonnet"
./lib/v1.10.6/k.libsonnet
./lib/v1.9.7/k.libsonnet
```

kkasravi:
@jlewi @ankushagarwal 

Update: 
- incorporating mixins for some of the components like tensorboard.
- in util.list the std.objectFieldsAll(obj) does not preserve order so i may remove this 
implicit generation of a components values or do it in a different way. 
- Hope to have this ready by Monday including an updated docs_dev/ksonnet_packages.md.

jlewi:
Thanks.

Do we still need to include self e.g.

"new(_env, _params):: self + { "

What is self referring to in this case?

> in util.list the std.objectFieldsAll(obj) does not preserve order so i may remove this
implicit generation of a components values or do it in a different way.

  * I think manual construction of the list would be just fine. That's what we've done until now and its generally been fine. We can even use late binding to allow users to modify one element without having to recreate the list.

jlewi:
Did you add tensorboard.libsonnet to the PR? I'm not seeing it. For tensorboard it might make sense to put the full depoyment and service into the prototype so that ks generate is effectively just a copy. Users will likely want to modify it in order to attach different volumes.

kkasravi:
Tensorboard.libsonnet was added.

@jlewi Question:
Do we want to construct all objects using k.libsonnet - as done in tensorboard? I can do this everywhere - it should take part of a day.

It looks like in
"new(_env, _params):: self + { "

self isn’t required, i’ll remove

kubeflow/core/tensorboard.libsonnet
```diff
@@ -1,247 +1,153 @@
...
-    },  // tbContainer
-
-    tfDeployment: {
-      apiVersion: "extensions/v1beta1",
-      kind: "Deployment",
-      metadata: {
-        name: $.params.name,
-        namespace: $.params.namespace,
-        labels: $.params.labels,
+  local k = import "k.libsonnet",
```
jlewi: Should we move all of this into the .jsonnet file? This will make it easier for people to customize.

tensorboard is different from components like JupyterHub because end users will need to create many instances of tensorboard whereas most people won't be modifying JupyterHub.

PR #1445, opened 2018-08-29
Fix readme for local run

lluunn:
/cc @kunmingg 
/cc @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1445)

### DIFF
--- a/components/gcp-click-to-deploy/README.md
+++ b/components/gcp-click-to-deploy/README.md
@@ -23,12 +23,18 @@ in the web app and used in requests to GCP services when needed.
 
 ## Development
 
-You can run the app locally for development/testing
-
+You can run the app locally for development/testing.
 ```
 cd components/gcp-click-to-deploy
 # Install dependencies
 npm install
+```
+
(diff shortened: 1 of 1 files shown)


components/gcp-click-to-deploy/README.md
```diff
@@ -23,12 +23,18 @@ in the web app and used in requests to GCP services when needed.
...
-You can run the app locally for development/testing
-
+You can run the app locally for development/testing.
 ```
 cd components/gcp-click-to-deploy
 # Install dependencies
 npm install
+```
+
+Before starting the server, make a config file `user_config/app-config.yaml` under `src/` with the content from
```
yebrahim: @jlewi @kunmingg why is this needed? Why can't the file by symlinked, or just loaded from that location?
Also, the location is outdated.

yebrahim: So is the deployer's url. This is currently the only place I can find easily that points to the deployer.

PR #1449, opened 2018-08-31
Click to deploy ui: remove ipName and make hostName optional

lluunn:
/cc @kunmingg 
/cc @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1449)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -31,7 +31,6 @@ interface DeployFormState {
   deploymentName: string;
   dialogTitle: string;
   dialogBody: string;
-  hostName: string;
   project: string;
   showLogs: boolean;
   zone: string;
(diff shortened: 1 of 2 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -319,16 +321,22 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
         return;
       }
     }
 
+    if (this.state.hostName === '') {
+      this.setState({
+        hostName: this.state.deploymentName + '.endpoints.' + this.state.project + '.cloud.goog',
+      });
+    }
+
```
abhi-g: Does this mean that if a user wants they can specify a completely different hostname?

kunmingg: Yes, currently we only support the default one.
We can add support to custom hostname later.

kunmingg: We can remove hostname field from UI and print it in logs?

lluunn:
Removed hostname field from UI, PTAL

lluunn:
screen shot 2018-08-31 at 11 01 41

PR #1460, opened 2018-08-31
implement gcpUtils which takes care of integrating deployment manager API (part of PR 1458)

kunmingg:
See previous PR:
https://github.com/kubeflow/kubeflow/pull/1458

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1460)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/gcpUtils.go
+++ b/bootstrap/cmd/bootstrap/app/gcpUtils.go
@@ -1,0 +1,86 @@
+package app
+
+import (
+	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


jlewi:
Please give the PR a more descriptive name

jlewi:
It looks like information is still stored about the application in a map[project + deployment] name.
Presumably this is because we are still storing the app on disk.

I assume this is only short term until we can make it truly state less by pulling the app on each request from git?

kunmingg:
Yes, will do in following steps.

Issue #1432, opened 2018-08-27
[gcp] VM account should have GCS read only scope to support pulling from GCR


jlewi:
Our DM configs should add readonly scope to the VM service account so that we can pull private images.

https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster.jinja#L24

We should include
https://www.googleapis.com/auth/devstorage.read_only

sarahmaddox:
Confirmation: I've just successfully pulled a container into GKE from GCR, after manually adjusting the OAuth settings in cluster.jinja as described in this issue. 

Before making that change, I reliably received the following error in GKE:
ImagePullBackOff rpc error: code = Unknown desc = unauthorized: authentication required: ErrImagePull
Failed to pull image "gcr.io/ml-writers/sarahmaddox-kf-gcp-train:1535231259": rpc error: code = Unknown desc = unauthorized: authentication required

After making the change to the cluster.jinja file, the image pull works.

jlewi:
Applications that need read/write access to GCS storage should use the "-user" service account
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/cluster.jinja#L43

And not the VM service account.

We only want to give the VM service account read access as a convenience for pulling private GCR images.

jlewi:
The fix was pushed in v0.2.5
https://github.com/kubeflow/kubeflow/releases/tag/v0.2.5

PR #1466, opened 2018-09-03
Add GCS read only scope to GKE VM scopes

gindeleo:
Regarding #1432 I've added the read only scope to the Google deployment manager templates (cluster.jinja). 

This will allow the GKE to pull private images from gcr.io/<project>.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1466)


jlewi:
Thanks for the fix!

/ok-to-test
/lgtm
/approve

PR #1467, opened 2018-09-03
Add Spark Job

holdenk:
Initial work in progress attempt at adding Spark to kubeflow using the spark-on-k8s-operator as a base starting point. This is super early but I'd love peoples feedback on the direction with this.

cc @texasmichelle 

Known TODOs:
- Tests that run on more than just my specific directory layout
- Documentation
- Cleanup

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1467)

### DIFF
--- a/kubeflow/spark/all.libsonnet
+++ b/kubeflow/spark/all.libsonnet
@@ -1,0 +1,348 @@
+{
+  // Define the various prototypes you want to support.
+  // Each prototype should be a list of different parts that together
+  // provide a userful function such as serving a TensorFlow or PyTorch model.
+  all(params, name, env):: [
+    $.parts(params, name, env).operatorServiceAccount,
+    $.parts(params, name, env).operatorClusterRole(params.deploymentScope),
+    $.parts(params, name, env).operatorClusterRoleBinding(params.deploymentScope),
+    $.parts(params, name, env).deployment,
+  ],
+
+  sparkJob(params, name, env):: [
+    $.parts(params, name, env).jobServiceAccount,
+    $.parts(params, name, env).jobClusterRole,
+    $.parts(params, name, env).jobClusterRoleBinding,
+    $.parts(params, name, env).sparkJob,
+  ],
+
+  // Parts should be a dictionary containing jsonnet representations of the various
+  // K8s resources used to construct the prototypes listed above.
+  parts(params, name, env):: {
+    // All ksonnet environments are associated with a namespace and we
+    // generally want to use that namespace for a component.
+    // However, in some cases an application may use multiple namespaces in which
+    // case the namespace for a particular component will be a parameter.
+    local namespace = if std.objectHas(params, "namespace") then params.namespace else env.namespace,
+    local mainClass = if params.mainClass == "null" then "" else params.mainClass,
+    local jobArguments = if params.jobArguments == "null" then [] else std.split(params.jobArguments, ","),
+    local sparkVersion = params.sparkVersion,
+
+    jobServiceAccount:: {
+      apiVersion: "v1",
+      kind: "ServiceAccount",
+      metadata: {
+        name: name,
+        namespace: namespace,
+      },
+    },
+
+    jobClusterRole:: {
+      apiVersion: "rbac.authorization.k8s.io/v1beta1",
+      kind: "Role",
+      metadata: {
+        namespace: namespace,
+        name: name,
+      },
+      rules: [
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "pods",
+          ],
+          verbs: [
+            "*",
+          ],
+        },
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "services",
+          ],
+          verbs: [
+            "*",
+          ],
+        },
+      ],
+    },
+    jobClusterRoleBinding:: {
+      apiVersion: "rbac.authorization.k8s.io/v1beta1",
+      kind: "RoleBinding",
+      metadata: {
+        name: name,
+        namespace: namespace,
+      },
+      subjects: [
+        {
+          kind: "ServiceAccount",
+          name: name,
+          namespace: namespace,
+        },
+      ],
+      roleRef: {
+        kind: "Role",
+        name: name,
+        apiGroup: "rbac.authorization.k8s.io",
+      },
+    },
+    operatorServiceAccount:: {
+      apiVersion: "v1",
+      kind: "ServiceAccount",
+      metadata: {
+        name: name,
+        namespace: namespace,
+      },
+    },
+    operatorClusterRole(deploymentScope):: {
+      local roleType = if deploymentScope == "cluster" then "ClusterRole" else "Role",
+      kind: roleType,
+      apiVersion: "rbac.authorization.k8s.io/v1beta1",
+      metadata: {
+        labels: {
+          app: "spark-operator",
+        },
+        name: name,
+        [if deploymentScope == "namespace" then "namespace"]: namespace,
+      },
+      rules: [
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "pods",
+          ],
+          verbs: [
+            "*",
+          ],
+        },
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "services",
+            "configmaps",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "delete",
+          ],
+        },
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "nodes",
+          ],
+          verbs: [
+            "get",
+          ],
+        },
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "events",
+          ],
+          verbs: [
+            "create",
+            "update",
+            "patch",
+          ],
+        },
+        {
+          apiGroups: [
+            "apiextensions.k8s.io",
+          ],
+          resources: [
+            "customresourcedefinitions",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "update",
+            "delete",
+          ],
+        },
+        {
+          apiGroups: [
+            "admissionregistration.k8s.io",
+          ],
+          resources: [
+            "mutatingwebhookconfigurations",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "update",
+            "delete",
+          ],
+        },
+        {
+          apiGroups: [
+            "sparkoperator.k8s.io",
+          ],
+          resources: [
+            "sparkapplications",
+            "scheduledsparkapplications",
+          ],
+          verbs: [
+            "*",
+          ],
+        },
+      ],
+    },
+    operatorClusterRoleBinding(deploymentScope):: {
+      apiVersion: "rbac.authorization.k8s.io/v1beta1",
+      local bindingType = if deploymentScope == "cluster" then "ClusterRoleBinding" else "RoleBinding",
+      local roleType = if deploymentScope == "cluster" then "ClusterRole" else "Role",
+      kind: bindingType,
+      metadata: {
+        name: name,
+        [if deploymentScope == "namespace" then "namespace"]: namespace,
+      },
+      subjects: [
+        {
+          kind: "ServiceAccount",
+          name: name,
+          namespace: namespace,
+        },
+      ],
+      roleRef: {
+        kind: roleType,
+        name: name,
+        apiGroup: "rbac.authorization.k8s.io",
+      },
+    },
+    deployment:: {
+      apiVersion: "apps/v1beta1",
+      kind: "Deployment",
+      metadata: {
+        name: name,
+        namespace: namespace,
+        labels: {
+          "app.kubernetes.io/name": name,
+          "app.kubernetes.io/version": sparkVersion,
+        },
+      },
+      spec: {
+        replicas: 1,
+        selector: {
+          matchLabels: {
+            "app.kubernetes.io/name": name,
+            "app.kubernetes.io/version": sparkVersion,
+          },
+        },
+        strategy: {
+          type: "Recreate",
+        },
+        template: {
+          metadata: {
+            annotations: {
+              "prometheus.io/scrape": "true",
+              "prometheus.io/port": "10254",
+              "prometheus.io/path": "/metrics",
+            },
+            labels: {
+              "app.kubernetes.io/name": name,
+              "app.kubernetes.io/version": sparkVersion,
+              name: name,
+            },
+            initializers: {
+              pending: [
+
+              ],
+            },
+          },
+          spec: {
+            serviceAccountName: name,
+            containers: [
+              {
+                name: name,
+                image: params.image,
+                imagePullPolicy: "Always",
+                command: [
+                  "/usr/bin/spark-operator",
+                ],
+                ports: [
+                  {
+                    containerPort: 10254,
+                  },
+                ],
+                args: [
+                  "-logtostderr",
+                  "-enable-metrics=true",
+                  "-metrics-labels=app_type",
+                ],
+              },
+            ],
+          },
+        },
+      },
+    },
+    // Job specific configuration
+    sparkJob:: {
+      apiVersion: "sparkoperator.k8s.io/v1alpha1",
+      kind: "SparkApplication",
+      metadata: {
+        name: params.jobName,
+        namespace: namespace,
+      },
+      spec: {
+        type: params.type,
+        mode: "cluster",
+        image: params.image,
+        imagePullPolicy: "Always",
+        mainClass: mainClass,
+        mainApplicationFile: params.applicationResource,
+        arguments: jobArguments,
+        volumes: [
+          {
+            name: "test-volume",
+            hostPath: {
+              path: "/tmp",
+              type: "Directory",
+            },
+          },
+        ],
+        driver: {
+          cores: params.driverCores,
+          memory: params.driverMemory,
+          labels: {
+            version: sparkVersion,
+          },
+          serviceAccount: params.name,
+          volumeMounts: [
+            {
+              name: "test-volume",
+              mountPath: "/tmp",
+            },
+          ],
+        },
+        executor: {
+          cores: params.executorCores,
+          instances: params.numExecutors,
+          memory: params.executorMemory,
+          labels: {
+            version: params.sparkVersion,
+          },
+          volumeMounts: [
+            {
+              name: "test-volume",
+              mountPath: "/tmp",
+            },
+          ],
+        },
+        restartPolicy: "Never",
+      },
+    },
+  },
+}
--- a/kubeflow/spark/all.libsonnet
+++ b/kubeflow/spark/all.libsonnet
@@ -23,7 +23,7 @@
     // generally want to use that namespace for a component.
     // However, in some cases an application may use multiple namespaces in which
     // case the namespace for a particular component will be a parameter.
-    local namespace = env.namespace;
+    local namespace = env.namespace,
     local mainClass = if params.mainClass == "null" then "" else params.mainClass,
     local jobArguments = if params.jobArguments == "null" then [] else std.split(params.jobArguments, ","),
     local sparkVersion = params.sparkVersion,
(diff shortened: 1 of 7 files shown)


kubeflow/spark/parts.yaml
```diff
@@ -0,0 +1,27 @@
+{
+   "name": "spark",
+   "apiVersion": "0.0.1",
+   "kind": "ksonnet.io/parts",
+   "description": "An empty package used as a stub for new packages.\n",
```
pdmack: "Holden's awesome Spark Job prototype\n"

pdmack:
/ok-to-test

@holdenk did you autoformat the *sonnet files yet? Guess we'll find out...

pdmack:
winner, winner...

Please use `scripts/autoformat_jsonnet.sh` to fix up the sonnet files.

kubeflow/spark/parts.yaml
holdenk: <3 :D

kubeflow/spark/all.libsonnet
```diff
@@ -0,0 +1,339 @@
...
+        template: {
+          metadata: {
+            annotations: {
+              "prometheus.io/scrape": "true",
+              "prometheus.io/port": "10254",
+              "prometheus.io/path": "/metrics",
+            },
+            labels: {
+              "app.kubernetes.io/name": name + "-sparkoperator",
+              "app.kubernetes.io/version": "v2.3.1-v1alpha1",
```
inc0: should this be a parameter with default ver?

kubeflow/spark/all.libsonnet
```diff
@@ -0,0 +1,339 @@
...
+      kind: "RoleBinding",
+      metadata: {
+        name: name + "-spark-role-binding",
+        namespace: namespace,
+      },
+      subjects: [
+        {
+          kind: "ServiceAccount",
+          name: name + "-spark",
+          namespace: "default",
```
inc0: this should be param

kubeflow/spark/all.libsonnet
holdenk: Yes, good catch,

kubeflow/spark/all.libsonnet
holdenk: thanks for catching.

holdenk:
Hey @texasmichelle & @kunmingg I'm wondering about the best way to test this operator - looking at the `test_deploy.py` script it seems to use KF components @ master but since this is a PR I'm not sure that would work well.

kunmingg:
@holdenk 
1. You can edit tmp logic change in ```test_deploy.py``` like changing version to use.
2. presubmit test will then test under your new logic.
3. When test looks good you can revert change in 1. and merge this PR.

holdenk:
Ok so I can point it at this branch. Is there a reason this isn't auto
pointed at the branch in PR builds (seems like it could give us bad results
on PRs in general)? Unless I'm missing something which is quite possible
since it's my first pass through this code.

kunmingg:
@holdenk 
For tests running on GKE I believe now they are covered by https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml#L15
which should deploy components from current PR code.
It would be easier to add new test here.

```test_deploy.py``` is more focusing on minikube integration now.

File: kubeflow/spark/all.libsonnet [hunk 4]
```diff
@@ -0,0 +1,339 @@
...
+    operatorServiceAccount:: {
+      apiVersion: "v1",
+      kind: "ServiceAccount",
+      metadata: {
+        name: name + "-sparkoperator",
+        namespace: namespace,
+      },
+    },
+    operatorClusterRole:: {
+      kind: "ClusterRole",
```
kris-nova: Why `ClusterRole` instead of `Role`? Looks like we are binding on a single namespace only, and opening up the broader permissions might be unnecessary? 

https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole

holdenk: Not super sure since the RBACS are based on the ones from the spark-operator project from GCP folks, my gut is that if we wanted to support having jobs in different namesapces wed need the operator have a `ClusterRole` but I'm not super sure. The RBAC file for the spark-operator is at https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/8c7fdbb306dfd656093c1b2a4ede901d651c9bd5/manifest/spark-operator-rbac.yaml , I could try and scope it down though and see if it still works?

kris-nova: Ah I didn't realize it was a port from the `spark-operator` project. I can ask there. It's not necessarily a concern, just more of wandering why we needed the broader scope. If we are spanning namespaces that makes sense. No change needed, thanks for clarifying.

testing/deploy_kubeflow.py
```diff
@@ -56,6 +57,8 @@ def deploy_kubeflow(test_case):
   # TODO(jlewi): We don't need to generate a core component if we are
   # just deploying TFServing. Might be better to refactor this code.
   # Deploy Kubeflow
+  print("Generate operators.")
```
jlewi: Is this script called by kfctl_test.jsonnet?

I think kfctl_test.jsonnet is just calling kfctl.sh and not using these script..

testing/workflows/components/kfctl_test.jsonnet
```diff
@@ -215,6 +215,11 @@ local dagTemplates = [
         kfCtlPath,
         "generate",
         "k8s",
+        // Generate spark operator
+        "&&",
+        kfCtlPath,
+        "generate",
```
jlewi: Did you modify kfctl to add this?

components/k8s-model-server/images/releaser/components/workflows.libsonnet
```diff
@@ -247,6 +247,11 @@
           template: "test-tf-serving",
           dependencies: ["deploy-tf-serving"],
         },
+        {
+          name: "deploy-spark-job",
```
jlewi: Why did you modify this file?
I think this E2E test is specific for K8s model server.

jlewi:
It doesn't look like the test is quite right. 

Here's some guidance.

You want to create an Argo Workflow that does something like the following

1. Use kfctl.sh to create and deploy Kubeflow
2. Run the ksonnet command to generate the spark job component and apply it
3. Run some script e.g. the python script you have to verify the deployment is working.

For now I don't think we should install the sparkjob operator by default; so for now users should have to do something like

ks generate spark-job spark-job
ks apply ${ENV} -c spark-job

So your Argo workflow should have steps executing these commands.

In terms of which Argo workflow to modify I think you want to look at this one
https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_test.jsonnet

You have 2 options

1. You can add an optional step to that Argo workflow to deploy/test the spark job
1. You could create a separate Argo workflow to just test deploy the SparkJob CRD.

testing/deploy_kubeflow.py
holdenk: No, this is called by the e2e workflow test.

testing/deploy_utils.py
```diff
@@ -97,14 +97,22 @@ def setup_kubeflow_ks_app(dir, namespace, github_token, api_client):
   util.run(["ks", "env", "set", "default", "--namespace=" + namespace_name],
     cwd=app_dir)
 
-  kubeflow_registry = "github.com/kubeflow/kubeflow/tree/master/kubeflow"
+  # IDK what the env looks like lets just dump everything
```
holdenk: This hard coded registry to master was confusing to me, and using the one that's checked out seems to work -- is there a reason for the master registry?

jlewi:
In terms of getting the logs. If you add labels to the argo steps as described here
https://github.com/kubeflow/testing#argo-spec

That should make it more convenient to get pod ids.

Unfortunately most of our existing workflow tests e.g. kfctl_test.jsonnet don't do this so you'd have to add it.

Assuming you have the right access (I think you added yourself to ci-team?) you should be able to access the kubeflow-testing cluster so you should be able to then get the pod id from the labels.

Instructions for fetching the logs are here:
https://github.com/kubeflow/testing#debugging-failed-tests

As an example here's a query to fetch one of the pod logs

```
resource.type="container"
resource.labels.cluster_name="kubeflow-testing"
resource.labels.pod_id="kubeflow-tf-operator-presubmit-v1beta1-910-ec293c3-1353-4037-3768977243"
```

holdenk:
Ok, I'd been trying to find the pod ids in the gubernator logs as mentioned in https://github.com/kubeflow/testing#debugging-failed-tests but it wasn't really working out ( see https://gubernator.k8s.io/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/1467/kubeflow-presubmit/4917/?log#log ). Connecting to the kubeflow-testing cluster though I see only pods > 10 days old which doesn't seem right.

holdenk:
It sounds like it might be easier to test as a separate workflow so I guess I'll explore how to do that assuming that `deploy_kubeflow.py` isn't meant to be general purpose.

holdenk:
Looks like the E2E tests finally pass, I'll rework the commit history into something more manageable but if you've got some cycles to take a look at how it ended up would be rad cc @texasmichelle

testing/deploy_utils.py
jlewi: Why are you modifying this script?

testing/spark_temp/simple_test.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
```
jlewi: What is this script for? Can you add a comment please?

testing/deploy_utils.py
jlewi: Are you using this script test_deploy.py to deploy the spark operator in your cluster?

testing/workflows/components/kfctl_test.jsonnet
```diff
@@ -230,6 +230,53 @@ local dagTemplates = [
     ),
     dependencies: ["kfctl-apply-gcp"],
   },
+  {
+    template: buildTemplate(
+      "install-spark-operator",
```
jlewi: So here in your test you are running ks directly to install the package so are you using deploy_kubeflow.py or test_deploy.py?

testing/workflows/components/workflows.libsonnet
```diff
@@ -278,6 +278,25 @@
         },
         dependencies: ["wait-for-kubeflow"],
       },  // pytorchjob - deploy,
+      {
```
jlewi: I think this is legacy; I think this workflow is only used by the minikube test because it hasn't been updated to use kfctl.sh. 

I think its fine not to run the test on minikube.

jlewi:
Really appreciate your persistence; I know this has been a slog through our test infra.

I think you are making some unnecessary changes related to testing.

Some background:

There are currently two different E2E tests for deploying Kubeflow and verifying it works

1. [kfctl_test.jsonnet](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_test.jsonnet) - Uses GKE and kfctl.sh
1. [workflows.jsonnet](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/workflows.jsonnet) - Uses minikube and kfctl.sh

Both of these use the code in workflows.libsonnet.

The minikube way of doing things is outdated as described in this TODO
https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/workflows.libsonnet#L325

We want to follow the GKE test and use kfctl.sh to deploy things.

As result, you should only test on GKE not on minikube because as the above indicates the minikube test infra is not well maintained.

This means you shouldn't need to modify deploy_kubeflow.py.

So the next thing you need to figure out is how will you add steps to the argo workflow to install the spark job package and verify it works. There are two ways to do this

1. You can add argo steps that just shell out to ks
    * This appears to be what you are doing in kfctl_test.sh https://github.com/kubeflow/kubeflow/pull/1467/files#diff-4718be91c949a9485ca0cf3fa589267aR234

1. You can create a script (bash or python) to run the commands

    * It seems like you are partially going down this road by creating simple_test.sh but it doesn't seem like this script gets used anywhere.

My suggestion would be to go the route of the second option and use python. The main advantage is that you can then run the test locally against your K8s cluster to see if its working.

For an example of what this python script might look like you can take a look at

https://github.com/kubeflow/examples/blob/master/mnist/testing/deploy_test.py

   * This script runs some ks commands to deploy a TFServing instance
   * Runs a simple check to verify that its deployed successfully
   * You'd have to adapt the commands of course

The main advantage is that as the instructions in that file show; it should be fairly straightforward to run it against your existing K8s cluster to develop and debug.

Once you have the script working you can just invoke it from your Argo workflow. This way integration into Kubeflow CI is largely just a matter of getting the paths, working directory, etc... set correctly.

kubeflow/spark/all.libsonnet
```diff
@@ -0,0 +1,348 @@
...
+  ],
+
+  // Parts should be a dictionary containing jsonnet representations of the various
+  // K8s resources used to construct the prototypes listed above.
+  parts(params, name, env):: {
+    // All ksonnet environments are associated with a namespace and we
+    // generally want to use that namespace for a component.
+    // However, in some cases an application may use multiple namespaces in which
+    // case the namespace for a particular component will be a parameter.
+    local namespace = if std.objectHas(params, "namespace") then params.namespace else env.namespace,
```
holdenk: Chatted with @texasmichelle about this and the weird behaviour I had scene back with the minikube tests and it make sense now so I'll keep as-is

kubeflow/spark/all.libsonnet
```diff
@@ -0,0 +1,348 @@
...
+              mountPath: "/tmp",
+            },
+          ],
+        },
+        executor: {
+          cores: 1,
+          instances: 1,
+          memory: params.executorMemory,
+          labels: {
+            version: "2.3.1",
```
holdenk: Make the version come from params?

kubeflow/spark/all.libsonnet
```diff
@@ -0,0 +1,348 @@
...
+          serviceAccount: params.name,
+          volumeMounts: [
+            {
+              name: "test-volume",
+              mountPath: "/tmp",
+            },
+          ],
+        },
+        executor: {
+          cores: 1,
```
holdenk: TODO: paramatirize this

testing/deploy_kubeflow.py
```diff
@@ -114,6 +114,7 @@ def deploy_kubeflow(test_case):
 
 
 def main():
+  print("Calling deploy kubeflow")
```
holdenk: REVERT THIS

testing/spark_temp/simple_test.sh
holdenk: This was for local testing, I can remove it.

testing/deploy_utils.py
holdenk: So I figured that hard coding master was a bad idea and installing Spark as part of the e2e minikube tests make sure it at least can be installed. I don't feel strongly about this since we don't need it for the full workflow tests so I'm happy to revert if this complicates matters.

testing/test_deploy.py
```diff
@@ -126,7 +126,7 @@ def setup_kubeflow_ks_app(args, api_client):
     ["ks", "registry", "add", "kubeflow", kubeflow_registry], cwd=app_dir)
 
   # Install required packages
-  packages = ["kubeflow/common", "kubeflow/tf-serving", "kubeflow/tf-job", "kubeflow/pytorch-job", "kubeflow/argo"]
+  packages = ["kubeflow/common", "kubeflow/tf-serving", "kubeflow/tf-job", "kubeflow/pytorch-job", "kubeflow/argo", "kubeflow/spark"]
```
holdenk: TODO: revert this

testing/workflows/components/kfctl_test.jsonnet
holdenk: Yeah I can revert those if we want, I figured it made sense to see Spark installed on minikube even if we only used the operator on the full version in e2e workflow.

testing/workflows/components/workflows.libsonnet
holdenk: So this part seems to be triggered outside of the minikube tests.

That being said, I think it might make sense, in the future, to test this and other operator one-by-one on minikube, what do you think?

jlewi:
@holdenk Anything I can do to help?

testing/deploy_utils.py
```diff
@@ -97,14 +97,22 @@ def setup_kubeflow_ks_app(dir, namespace, github_token, api_client):
...
+    REGISTRY_PATH)
   util.run(
     ["ks", "registry", "add", "kubeflow", kubeflow_registry], cwd=app_dir)
 
   # Install required packages
-  packages = ["kubeflow/common", "kubeflow/gcp", "kubeflow/jupyter", "kubeflow/tf-serving", "kubeflow/tf-job", "kubeflow/tf-training", "kubeflow/pytorch-job", "kubeflow/argo"]
+  packages = ["kubeflow/common", "kubeflow/gcp", "kubeflow/jupyter", "kubeflow/tf-serving", "kubeflow/tf-job", "kubeflow/tf-training", "kubeflow/pytorch-job", "kubeflow/argo", "kubeflow/spark"]
 
   # Instead of installing packages we edit the app.yaml file directly
+  # ^ Why?
```
holdenk: I wish I knew why, but it is unrelated so I'll get rid of this.

kubeflow/spark/all.libsonnet
```diff
@@ -0,0 +1,339 @@
...
+        },
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "nodes",
+          ],
+          verbs: [
+            "get",
```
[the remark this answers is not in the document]
holdenk: So ClusterRole versus Role is now user configurable, if folks don't need to run jobs outside of the namespace where they created the operator we'll just do a Role, but if they want to have the operator and jobs sit in different namespaces we use clusterrole.

holdenk:
Thanks so much for all your help with reviews @jlewi , @texasmichelle  @inc0 , @pdmack, @kris-nova & @kunmingg it should be ready for review again I've squashed it down and rebased on master.

jlewi:
FYI looks like the test is failing

```
+ [[ 1 -eq 0 ]]
+ echo Command failed: ks apply -c spark-operator --verbose
+ for i in '$(seq 1 3)'
+ set +e
+ ks apply -c spark-operator --verbose
level=debug msg="setting log verbosity" verbosity-level=1
level=debug msg="loading application configuration from /mnt/test-data-volume/kubeflow-presubmit-kfctl-1467-c45c69c-5289-6200/kfctl-6200/ks_app"
level=error msg="environment is not set; use `env list` to see available environments"
```

holdenk:
cc @jlewi can you tag this as OK to test again? CI lost the tag :(

holdenk:
Thanks @richardsliu !

holdenk:
Looks like the spark operator is applying successfully now, so it should be ready for review again. cc @jlewi @texasmichelle

jlewi:
Thanks @holdenk. It looks like the spark apply test is still failing

```
+ ks apply default -c spark-operator --verbose
level=debug msg="setting log verbosity" verbosity-level=1
level=debug msg="loading application configuration from /mnt/test-data-volume/kubeflow-presubmit-kfctl-1467-39b503c-5575-5aca/kfctl-5aca/ks_app"
level=debug msg="creating ks pipeline for environment \"default\""
level=debug msg="building objects" action=pipeline module-name=/
level=debug msg="jsonnet evaluate snippet" elapsed=59.03316ms name=applyGlobals
level=error msg="find objects: open /mnt/test-data-volume/kubeflow-presubmit-kfctl-1467-39b503c-5575-5aca/kfctl-5aca/ks_app/environments/default/params.li
bsonnet: no such file or directory"
```

I think the problem is you aren't creating the default environment.
The default environment is created by the kfctl-apply-k8s step. So I think you want to make the spark-operator step depend on that step.

holdenk:
@jlewi so I think the apply operator is succeeding, the part which is failing is the part which depended on the Python script so I don't think it the env issue (although that was possibly the issue before).

jlewi:
Woo Hoo!
Thanks for pushing this through.
/lgtm
/approve

holdenk:
Test failures look unrelated, /retest

jlewi:
We were having quota issues earlier. Should be fixed now.

/lgtm
/approve

holdenk:
It's still failing before getting to the Spark specific code, and looks like a quota failure so lets try:
/retest

holdenk:
Although IDK if the bot listens to me might need someone else to tell it to retest

holdenk:
Ok the spark jobs passed but the notebooks test failed ?

PR #1482, opened 2018-09-05
Delete deploy.sh scripts; we use kfctl.sh now.

### LINKED ISSUE #1331: Update E2E test to use kfctl.sh and delete gke/deploy.sh;

### LINKED ISSUE #1481: Multiple issues with gke/deploy.sh
There are multiple small issues with deploying kubeflow to GKE via [deploy.sh] on HEAD.(https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh)

1. `KUBEFLOW_REPO` must be an absoulute path to specify a preexisting local kubeflow package. Should this path be absolute as the default is set to `kubeflow_repo` (which then gets augmented to its abolutute path contrary to the externally set env variable? See:
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh#L18

2. If `GCFS_INSTANCE` is not set but the GCFS instance already exists then the `wait` [here](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh#L245) will terminate the script as there is no PID existing. Checking upfront if the instance exist (as mentioned in @jlewi 's comment [here](https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh#L149) should fix that. 
Also, how much time is saved by creating the GCFS instance directly (and in parallel/async) as opposed to via deployment manager?

3. Creating the local variable here: https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deploy.sh#L149
raises an error as it's not called in a function.

jlewi:
* Fix #1331
* Fix #1481

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1482)


jlewi:
/test all

Looks like the there was a problem creating the PR symlink

PR #1485, opened 2018-09-05
Simplify tensorboard

### LINKED ISSUE #1477: simplify tensorboard as separate aws, gcp prototypes
the tensorboard.{libsonnet, jsonnet} files will be used for this

1. Show how the use of k.libsonnet simplifies creation and modification of manifests in tensorboard.{libsonnet, jsonnet}
2. Show the pattern exposed in k.libsonnet

kkasravi:
fixes #1477

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1485)


kkasravi:
> Why create a tensorboard package and not a more generic tensorflow package?
I think I lean towards a single tensorflow package because I suspect once a user starts using tensorflow they will likely use multiple components (e.g. TFJob, TFServing, Tensorboard) so its more convenient if they can just install a single ks package.

I agree. This was to see how {gcp, aws} prototypes could be built using mixins. This could be combined with TF in a subsequent issue/PR. For tensorboard, reducing the {aws,gcp} differences down to a function called by deployment.mapContainers where the function(container) just mixes in gcp or aws functionality is cleaner (IMO) than the original.

> What's the benefit of using "k.libsonnet" over just writing the manifest as json? Looking at the tensorboard-gcp prototype, ITSM to that using the "k.libsonnet" makes it harder to understand and doesn't really simplify the code.


Creating the base tbDeployment and tbService using k.core.v1.service and  k.apps.v1beta1.deployment  isn't a clear win unless you've gotten familiar with k.libsonnet. There is a small reduction of code and an argument that indentation errors or other common manifest mistakes are reduced. 

Going back to the original [use-case](https://github.com/kubeflow/kubeflow/issues/1414#issue-353856861) shown below:
```
  HubStatefulSet+: {
     spec+: {
       template+: {
         spec+: {
          imagePullSecrets: [
            {name: "mySecret",},
          ],
        },
      },
    },
  },
};
```

Using k.libsonnet it would be 
```
local instance = jupyterhub.new(env, params);
local statefulSet = k.apps.v1beta1.statefulSet;
instance.list(instance.hubStatefulSet + 
  statefulSet.mixin.spec.template.spec.withImagePullSecretsMixin({name: "mySecret"}))
```

One could argue that the latter is easier and less error prone. Once you've learned the pattern 
that k.libsonnet uses, it becomes pretty powerful. But I tend to agree with your earlier [assessment](https://github.com/kubeflow/kubeflow/pull/1437#issuecomment-418523083) that only base objects which are modified by ksonnet/jsonnet experts should take advantage of k.libsonnet.

pdmack:
need to rebase for #1494

PR #1487, opened 2018-09-05
fixes add ability to only generate parts of a component in the jsonnet file

### LINKED ISSUE #1486: add ability to only generate parts of a component in the jsonnet file
databricks style-guide suggests classes be defined as local variables but exported as methods.
Within the jsonnet file, we should have the flexibility to export some of the classes

kkasravi:
fixes #1486

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1487)


pdmack:
need to rebase for #1494

PR #1501, opened 2018-09-10
fixes update tensorboard to use the same pattern as kubeflow/core/prototypes

### LINKED ISSUE #1500: update tensorboard to use the same pattern as kubeflow/core/prototypes
small changes are needed to agree with the other prototypes

kkasravi:
fixes #1500

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1501)


kkasravi:
assign @ankushagarwal

PR #1515, opened 2018-09-12
Integrate with cloud source repos to save ks app config.

kunmingg:
1. make backend server pods truly stateless by saving state to cloud source repos (as source of truth)
2. integrate with cloud source repos as private git repo which user can easily clone their kubeflow config after deploy succeeded.
3. with new approach, can implement robust kubeflow edit / update logic into backend server since now we have versioned / trackable state / app config.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1515)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -500,6 +500,7 @@ func GetRepoName(project string) string {
 }
 
 // Not thread-safe, make sure project lock is on.
+// Clone project repo to local disk, which contains all existing ks apps config in the project
 func (s *ksServer) CloneRepoToLocal(project string, token string) error {
 	repoDir := path.Join(s.appsDir, GetRepoName(project))
 	_, err := s.fs.Stat(repoDir)
@@ -552,7 +553,7 @@ func (s *ksServer) GetApp(project string, appName string, token string) (*appInf
 	}, nil
 }
 
-// Save ks app config to cloud source repos
+// Save ks app config local changes to project source repo.
 // Not thread safe, be aware when call it.
 func (s *ksServer) SaveAppToRepo(project string) error {
 	repoDir := path.Join(s.appsDir, GetRepoName(project))
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -65,6 +65,7 @@ type ksServer struct {
 
 	fs afero.Fs
 
+	// project-id -> project lock
 	projectLocks map[string]*sync.Mutex
 	serverMux sync.Mutex
 }
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -3,7 +3,6 @@ package app
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"path"
 	"sync"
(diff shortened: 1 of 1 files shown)


kunmingg:
/assign @lluunn 
@abhi-g

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -63,7 +65,7 @@ type ksServer struct {
 
 	fs afero.Fs
 
-	apps      map[string]*appInfo
+	projectLocks map[string]*sync.Mutex
```
lluunn: I guess the map key is the project name and the mutex for each project? Could you add comment.

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -228,25 +196,31 @@ func (s *ksServer) CreateApp(ctx context.Context, request CreateRequest) error {
...
 	}
-	a, err := func() (*appInfo, error) {
+	func() {
 		s.serverMux.Lock()
 		defer s.serverMux.Unlock()
-
-		if request.Name == "" {
-			return nil, fmt.Errorf("Name must be a non empty string.")
+		_, ok := s.projectLocks[request.Project]
+		if !ok {
```
lluunn: not ok means the project lock not exists?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -527,18 +493,75 @@ func (s *ksServer) autoConfigureApp(kfApp *kApp.App, appConfig *AppConfig, names
...
-	mapKey := fmt.Sprintf(" %s-%s", project, appName)
-	info, ok := s.apps[mapKey]
-	if !ok {
+
+func GetRepoName(project string) string {
+	return fmt.Sprintf("%s-kubeflow-config", project)
+}
+
+// Not thread-safe, make sure project lock is on.
+func (s *ksServer) CloneRepoToLocal(project string, token string) error {
```
lluunn: one project can have multiple deployments. Are they all in the same repo?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -527,18 +494,75 @@ func (s *ksServer) autoConfigureApp(kfApp *kApp.App, appConfig *AppConfig, names
...
+		App: kfApp,
+	}, nil
 }
+
+// Save ks app config to cloud source repos
 // Not thread safe, be aware when call it.
-func (s *ksServer) InsertApp(project string, appName string, app *appInfo) {
-	mapKey := fmt.Sprintf(" %s-%s", project, appName)
-	s.apps[mapKey] = app
+func (s *ksServer) SaveAppToRepo(project string) error {
```
lluunn: Can you add some comment on this. It will sync all the app dir of the project right?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -339,6 +305,7 @@ func (s *ksServer) CreateApp(ctx context.Context, request CreateRequest) error {
 		s.autoConfigureApp(&a.App, &request.AppConfig, request.Namespace, config)
 	}
 
+	s.SaveAppToRepo(request.Project)
```
lluunn: This only add the new app right? Do you need to cd into the app dir?

bootstrap/cmd/bootstrap/app/ksServer.go
lluunn: Can you add comment that this will clone all the apps of the project

bootstrap/cmd/bootstrap/app/ksServer.go
kunmingg: It will sync the project repo, and each project has one repo.

PR #1516, opened 2018-09-12
edit gke version to 1.10.6

kunmingg:
some regions deprecated 1.10.5

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1516)


pdmack:
Why do we have different versions here and at #1494?

kunmingg:
@pdmack We have two sets of configs for deployment, will dedup to one set asap.

PR #1523, opened 2018-09-13
NVIDIA TensorRT Inference Server

deadeyegoodwin:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1523)


lluunn:
Thank you for contributing this!
I will take a look tomorrow morning.

The presubmit test failure is about jsonnet format. You can auto format it like https://github.com/kubeflow/kubeflow/blob/master/scripts/autoformat_jsonnet.sh#L89

PR #1535, opened 2018-09-14
fixes 'refactor tf-job-operator to match style-guide of libsonnet'

### LINKED ISSUE #1534: refactor tf-job-operator to match style-guide of libsonnet
tf-job-operator use a similar pattern as other libsonnets:
```
kubeflow/argo/argo.libsonnet
kubeflow/core/ambassador.libsonnet
kubeflow/core/centraldashboard.libsonnet
kubeflow/core/cert-manager.libsonnet
kubeflow/core/cloud-endpoints.libsonnet
kubeflow/core/echo-server.libsonnet
kubeflow/core/google-cloud-filestore-pv.libsonnet
kubeflow/core/iap.libsonnet
kubeflow/core/jupyterhub.libsonnet
kubeflow/core/metric-collector.libsonnet
kubeflow/core/prometheus.libsonnet
kubeflow/core/spartakus.libsonnet
kubeflow/tensorboard/tensorboard.libsonnet
```

kkasravi:
Fixes #1534

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1535)


jlewi:
What's the motivation for refactoring the TFJob operator spec to use k8s.libsonnet mixins as opposed to just writing the manifest directly? Is this intended to help readability?

kkasravi:
Thanks @jlewi - so happy you're back. I'll convert to the style you've suggested above.

For tf-serving - should i pick that up since it's similar to tensorboard?

jlewi:
Thanks for the warm regards :)

If you could pick up tf-serving that would be great; but I'd suggest letting in @lluunn because I think that was also on his radar.

Note to get the tests passed you'll have to update them to work with ks 0.12 see
https://github.com/kubeflow/kubeflow/issues/1540#issuecomment-422253521

jlewi:
I think this PR requires ks 0.12. So I don't think this will pass until you make the changes I described here
https://github.com/kubeflow/kubeflow/issues/1540#issuecomment-422253521

Feel free to ping me on slack if you want to discuss.

jlewi:
Looks like some jsonnet test failures; so I assume an issue with this PR.

kkasravi:
Note - there is a subtle issue in the crd definition - see #1606

kubeflow/core/tests/tf-job_test.jsonnet
```diff
@@ -1,12 +1,29 @@
 local tfjob = import "../tf-job-operator.libsonnet";
-local params = {
-  namespace:: "test-kf-001",
+local paramsv1alpha1 = {
+  name:: "tf-job-operator",
+  tfJobImage:: "gcr.io/kubeflow-images-public/tf_operator:v20180226-403",
```
jlewi: Please change to 
gcr.io/kubeflow-images-public/tf_operator:v0.3.0 so it won't collide with #1608

PR #1537, opened 2018-09-15
Prevent ambassador getaddrinfo error logs

### LINKED ISSUE #945: Ambassador Version 0.34.0 causing DNS Issues on Worker Node
We recently updated to ambassador 0.34. When I deployed kubeflow on a single node cluster (n1-standard-8) kube-dns seems to be running at 100% CPU. Other processes are not able to resolve DNS on the host. Deleting the ambassador deployment fixes the DNS issue.

sambaiz:
This prevents that appear the following error logs in large quantities
```
socat[1581671] E getaddrinfo("statsd-sink", "NULL", {1,0,2,17}, {}): Name does not resolve
```

This error always occurs.
It is hardcoded [here](https://github.com/datawire/ambassador/blob/0.37.0/statsd/Dockerfile#L5) in ambassador and it is mismatched so it seems to fail to resolve name.

It may resolve #945 (or it maybe already resolved)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1537)

### DIFF
--- a/kubeflow/core/ambassador.libsonnet
+++ b/kubeflow/core/ambassador.libsonnet
@@ -40,7 +40,7 @@
         labels: {
           service: "ambassador",
         },
-        name: "ambassador-exporter",
+        name: "statsd-sink",
         namespace: params.namespace,
         annotations: {
           "prometheus.io/scrape": "true",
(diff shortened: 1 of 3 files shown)


sambaiz:
@kunmingg 
This error always occurs.
It is hardcoded [here](https://github.com/datawire/ambassador/blob/0.37.0/statsd/Dockerfile#L5) on ambassador and it is mismatched so it seems to fail to resolve name.

kunmingg:
/lgtm
/approve

Thanks!

Issue #1320, opened 2018-08-06
Automated sync of Kubeflow between GCR and DockerHub


TheJaySmith:
We have a rising popularity in China with Kubeflow however they are unable to access GCR. We also have other users who may prefer using DockerHub over GCR for images as it works better for their workflow. We need to sync Kubeflow Dockerfiles between GCR and DockerHub

TheJaySmith:
Created kubeflowrobot account

swiftdiaries:
For reference / background: https://github.com/kubeflow/kubeflow/issues/211.

jlewi:
Here's Jay's code
https://github.com/TheJaySmith/kubeflowpush/blob/master/push.py

1. Its using docker. which will make this a pain to run on K8s.
1. It looks like it uses a hand written list of images to sync

Jay I suggest putting relevant code here:
https://github.com/kubeflow/kubeflow/tree/master/releasing

Jay I don't think we need to use docker. We can

1. Shell out to gcloud to get a list of images in our GCR registry

   * https://github.com/kubeflow/kubeflow/blob/master/releasing/add_image_shas.py#L43

1. We can use GCB to push from GCR to DockerHub

The GCB workflow could look like this:
https://github.com/jlewi/kubeflow-dev/blob/master/kaggle-image/retag_image.yaml

TheJaySmith:
Let me play with it this weekend to use gCloud and I will upload it to the https://github.com/kubeflow/kubeflow/tree/master/releasing location. Will update you before Monday

jlewi:
@TheJaySmith Any update/progress on this?

TheJaySmith:
#1548

jlewi:
Thanks.

Should we mark this issue as a Duplicate of #211 (Publish Docker Images to DockerHub)?

jlewi:
Duplicate of #211

PR #1548, opened 2018-09-18
HubSync v1.0

TheJaySmith:
This will sync gcr to dockerhub per #1320 . It functions in this way

1) runs gcloud to pull down all image repos within gcr.io/kubeflow-images-public
2) runs a follow-up gcloud command to pull down image tags (latest)
3) Build cloudbuild.yaml file to authenticate with DockerHub, pull the image from gcr.io/kubeflow-images-public, tag the image, and push it to DockerHub

This is base code and is pre-coded largely just to pull from GCR to DockerHub. Users could modify the code in order to push to their own GCR or DockerHub. keys.yaml is the file where you store your credentials. 

I am hoping for a version 2.0 with a CLI. This would break it down into steps so that a user can build and sync what they want. However, for now, this will move all images from gcr.io/kubeflow-images-public to DockerHub

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1548)

### DIFF
--- a/releasing/hubsync/hubsync.py
+++ b/releasing/hubsync/hubsync.py
@@ -1,0 +1,88 @@
+# Copyright 2015 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# [START app]
+
+
+import os
+import subprocess
+import logging
+import yaml
+import json
+import re
+import argparse
+
+#declaring some variables
+images = []
+timeout = '7200s'
+filename = 'cloudbuild.yaml'
+builder = 'gcr.io/cloud-builders/docker'
+kfRepo = 'gcr.io/kubeflow-images-public/'
+myRepo = 'gcr.io/<my_repo>'
+
+#Get Auth
+with open('keys.yaml', 'r') as keyfile:
+    kcfg = yaml.load(keyfile)
+
+login = kcfg['username']
+pswd= kcfg['password']
+
+#build a json file with all files.
+repos = subprocess.run(["gcloud","--project=kubeflow-images-public","container","images","list","--format=json"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+my_json = json.loads(repos.stdout.decode('utf8').strip().replace("'", '"'))
+for data in my_json:
+    for name, image in data.items():
+        #get Tags and Repos
+        raw_images = subprocess.run(["gcloud","container","images","list","--repository="+image+"","--format=json"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        imgData = raw_images.stdout.decode("utf-8")
+        if "[]" not in imgData:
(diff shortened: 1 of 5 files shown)


kubeflow/weaveflux/README.md
```diff
@@ -25,8 +25,17 @@ Finally, we will deploy the application
 ks apply default -c weaveflux
 ```
 
+<<<<<<< HEAD
```
jlewi: Looks like you have some merge conflicts in the file.

jlewi:
Please use "yaml" not "yml" as the extension.

Can you reference in the PR description the issue tracking syncing the Docker images?

Can you expand the PR description? What I'd like to understand is what this PR is doing in terms of the overall plan to sync our images to DockerHub.

It looks like this is a script that uses gcloud to get a list of the images and then creates a YAML file to retag those images using GCB?

TheJaySmith:
Hello, 

Fixed merge conflicts and yaml and updated comment

/retest

jlewi:
Thanks for the really great description. Sorry for the slow reply been busy preparing for the summit and helping with the 0.3 release.

I don't think we need to resolve this as part of this PR, but I thought the primary audience for this tool was the Kubeflow release team to make it easy to sync our images from GCR to DockerHub to support users who can't use GCR. Would we still need a CLI in that case?

Ideally I'd like to automate this. The current thinking is we are using Prow to manage building and pushing our Docker images. Prow has a periodic job which is equivalent to a cron job. So my initial thought is that we would setup a prow periodic job to run this script regularly so that we could sync all our images to DockerHub.

releasing/hubsync/hubsync.py
```diff
@@ -0,0 +1,94 @@
...
+                #print(imgData)
+                    images.append(b)
+        else:
+            #print(image)
+            images.append(image)
+
+
+for item in images:
+    getTags = subprocess.run(["gcloud","--project=kubeflow-images-public","container","images","list-tags",item,"--format=json","--limit=1"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    #print(getTags)
```
jlewi: Delete commented out code.

releasing/hubsync/requirements.txt
```diff
@@ -0,0 +1,6 @@
+<<<<<<< HEAD
+=======
```
jlewi: Looks like you have some merge conflicts?

jlewi:
It looks like this would push all of our images; even if they already exist in DockerHub. Is this something we will need to fix later?

jlewi:
/test all

Test failure looks like a flake since this PR isn't actually changing any code that would affect kfctl.

TheJaySmith:
@jlewi it should just add an extra image tag in the future. 

My thoughts around a CLI are more for the end user. The script as it exists today takes all images and uploads them to another source. It's possible an end user might want to do some form of version control of the images for their own use or only use certain images for compliance or other policy. The idea behind the CLI would be to make each step it's own command so that an end user can only do what the feel they need.

jlewi:
Looks like the tests are still failing.

TheJaySmith:
@jlewi is there a particular element I can fix? I am having difficulty determining from the log why it's failing.

TheJaySmith:
@jlewi I took a deeper look at the logs and couldn't find anything specific. I may need a second pair of eyes to confirm.

jlewi:
Here' are the pod logs

```
version \"1.10.7-gke.1\" is unsupported.","status":"INVALID_ARGUMENT","statusMessage":"Bad
```

Your GKE deployment is failing because your GKE version is out of date. This is because your code is out of date. You need to sync and then rebase on master to pull in the fix.

jlewi:
@TheJaySmith you've suffered enough.

The test failures are unrelated to this PR since the only code modified is in hubsync which doesn't contain any jsonnet files.

I'll ignore the presubmit failures and merge manually.

PR #1550, opened 2018-09-18
Webapp: Don't use DM for IAM.

lluunn:
/cc @kunmingg 

iam_bindings_template.yaml is the same as https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/deployment_manager_configs/iam_bindings_template.yaml so it will be easy to merge later.

cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1550)


bootstrap/cmd/bootstrap/app/gcpUtils.go
```diff
@@ -83,4 +104,98 @@ func (s *ksServer)GetDeploymentStatus(ctx context.Context, req CreateRequest) (s
...
+			if exists {
+				binding.Members = append(binding.Members, member)
+			}
+		}
+		newPolicy.Bindings = append(newPolicy.Bindings, &binding)
+	}
+	return newPolicy
+}
+
+func (s *ksServer)ApplyIamPolicy(ctx context.Context, req ApplyIamRequest) error {
```
kunmingg: Let's add retry into ApplyIamPolicy.
There's a small chance that concurrent iam change might occur for same project.

lluunn: Done, PTAL

bootstrap/cmd/bootstrap/app/gcpUtils.go
```diff
@@ -187,11 +196,18 @@ func (s *ksServer)ApplyIamPolicy(ctx context.Context, req ApplyIamRequest) error
...
 	// Get the updated policy and apply it.
 	newPolicy := GetUpdatedPolicy(saPolicy, &iamConf, req)
-	_, err = resourceManager.Projects.SetIamPolicy(
-		req.Project,
-		&cloudresourcemanager.SetIamPolicyRequest{
-			Policy: &newPolicy,
-		}).Do()
+	retry = 0
+	if retry < 5 {
+		_, err = resourceManager.Projects.SetIamPolicy(
```
kunmingg: If the failure is caused by concurrent iam policy change, we need to re-get the changed iam policy as refresh start and then apply configs to it and try to set iam again.

lluunn: I see, updated. Thanks

Issue #1563, opened 2018-09-19
Image Auto Release Cron Job is failing


jlewi:
The cron job to regularly release our Kubeflow images is failing.
It runs in the kubeflow-releasing cluster;

Here is the error:

```
INFO|2018-09-19T00:05:44|/src/kubeflow/testing/py/kubeflow/testing/util.py|62| level=info msg="Creating environment \"image-release-centraldashboard-release-d8973b1-f6c7-427a\" with namespace \"default\", pointing to cluster at address \"https://35.226.49.107\""
INFO|2018-09-19T00:05:44|/src/kubeflow/testing/py/kubeflow/testing/util.py|62| level=error msg="load configuration: Current app uses unsupported spec version '0.2.0' (this client only supports 0.1.0)"
Traceback (most recent call last):
  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 349, in <module>
    final_result = main()
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 339, in main
    return run(args, file_handler)
  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 189, in run
    util.run(["ks", "env", "add", env], cwd=w.app_dir)
  File "/src/kubeflow/testing/py/kubeflow/testing/util.py", line 74, in run
    " ".join(command), process.returncode), "\n".join(output))
subprocess.CalledProcessError: Command 'cmd: ks env add image-release-centraldashboard-release-d8973b1-f6c7-427a exited with code 1' returned non-zero exit status 1
```

I think this indicates that ksonnet is too old in the image.

jlewi:
The pod used image:
gcr.io/kubeflow-ci/test-worker:latest
gcr.io/kubeflow-ci/test-worker@sha256:66fc77fc0f2cb1e388707da2e6f212c58e8778d21e02ea6194d563376c5a8fb2

jlewi:
It looks like ks 0.11 is installed in that image

ks version
ksonnet version: 0.11.0
jsonnet version: v0.10.0
client-go version:

jlewi:
[...]

PR #1565, opened 2018-09-19
Run Jupyter image and centraldashboard image release on postsubmit

jlewi:
* The autoimage releaser cron job is broken because the recently added centraldashboard
  workflow is using ksonnet app version 0.2.0 (which requires ksonnet 0.12) while
  the other release workflows are using version 0.1.0 of the APi (ksonnet 0.11)

* The simplest fix is to use 0.1.0.

* To make this easier consolidate all the release workflows in a single
  ksonnet app.

Related to #1563

* Trigger the centraldashboard image release workflow on presubmit.
  Related to #1574 (Use prow to manage image release workflows.)

  * Shorten the dashboard name to avoid hitting pod naming limits.

* Set CPU and RAM on the containers in release.libsonnet to see if this speeds up building
  the docker images.

* Add building of Jupyter Docker images to pre and postsubmits so we verify this is working. This will also give us something to compare against w.r.t the central dashboard images.

* Disable image builds on presubmits until they are passing.

Related to #1576 Jupyter image builds failing
Related to #1575 Central dashboard image builds aren't working.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1565)


prow_config.yaml
```diff
@@ -52,3 +52,45 @@ workflows:
...
+
+  # Image Auto Release workflows.
+  # The workflows below are related to auto building our Docker images.
+  # We have separate pre/postsubmit jobs because we want to use different
+  # registries
+  - app_dir: kubeflow/kubeflow/releasing/releaser
+    component: centraldashboard
+    name: dashboard-release
+    job_types:
+      - presubmit
```
kunmingg: Maybe we want to run release workflows only on postsubmit.

prow_config.yaml
```diff
@@ -52,3 +52,45 @@ workflows:
       - kubeflow/tf-serving/*
       - components/k8s-model-server/*
       - testing/*
+
+  # Image Auto Release workflows.
+  # The workflows below are related to auto building our Docker images.
```
kunmingg: Should we delete the current release cron job if prow will take over?

jlewi: We should but lets do that in a follow on PR.

Issue #1567, opened 2018-09-19
Certificate not working


lluunn:
I am seeing error in certificate: 
` ACME server does not allow selected challenge type or no provider is configured for domain `

certificate spec:
```
spec:
    acme:
      config:
      - domains:
        - kfdeploy10.endpoints.kai-test2.cloud.goog
        http01:
          ingress: envoy-ingress
```
issuer spec:
```
spec:
    acme:
      email: lunkai@google.com
      privateKeySecretRef:
        key: ""
        name: letsencrypt-prod-secret
      server: https://acme-v02.api.letsencrypt.org/directory
```
I am guessing the `http01: {}` got trimmed from issuer spec

lluunn:
Result of doing ks show locally:
```
apiVersion: certmanager.k8s.io/v1alpha1
kind: Issuer
metadata:
  labels:
    ksonnet.io/component: vm
  name: letsencrypt-prod
  namespace: default
spec:
  acme:
    email: abc
    privateKeySecretRef:
      name: letsencrypt-prod-secret
    server: https://acme-v02.api.letsencrypt.org/directory
```

lluunn:
Verified after I patched the issuer with `http01: `, the certificate works now.

Issuer spec:
```
spec:
    acme:
      email: lunkai@google.com
      http01: {}
      privateKeySecretRef:
        key: ""
        name: letsencrypt-prod-secret
      server: https://acme-v02.api.letsencrypt.org/directory
```

sambaiz:
#1544 includes a fix that resolve this issue.

PR #1580, opened 2018-09-19
remove prune from kubeflow core which delete required fields

kunmingg:
std.prune will delete key with empty value, which could be dangerous as we have some required keys with empty value.

related: 
https://github.com/kubeflow/kubeflow/issues/1567
https://github.com/kubeflow/kubeflow/pull/1544

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1580)


kunmingg:
@abhi-g

kunmingg:
prune against ks prototype configs will break consistency between what user configured in prototype and what ks app actually output.

Issue #1264, opened 2018-07-25
Make TF serving component more readable and extendable


lluunn:
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/tf-serving/tf-serving.libsonnet
It's currently not very readable, and adding other features will make it worse (e.g. request logging).

Some previous discussion in #376 .

Here are some example packages: https://github.com/ksonnet/parts/tree/master/incubator
From this I think the general pattern is for each component, define 
1. default [param](https://github.com/ksonnet/parts/blob/master/incubator/nginx/nginx.libsonnet#L8)
1. [base](https://github.com/ksonnet/parts/blob/master/incubator/nginx/nginx.libsonnet#L32)
1. different [extensions](https://github.com/ksonnet/parts/blob/master/incubator/nginx/nginx.libsonnet#L16)

We can utilize ksonnet [mixins](https://github.com/ksonnet/parts/blob/master/incubator/redis/redis.libsonnet#L3).

jlewi:
Given #1229 I think the use case is to support things like injecting a config map and sidecar for request logging.

I think there are two patterns we should explore

1. Using custom admission controllers
1. Using overlays 
   * I think this was introduced by [kustomize](https://github.com/kubernetes-sigs/kustomize)
   * But I believe its supported in ksonnet with the "+" operator

lluunn:
#1589 is the new pattern. We can improve on that

PR #1589, opened 2018-09-20
New TF Serving template

lluunn:
For #1264

Move things into jsonnet instead of relying on import.
It's easier to customize.

- previous params remain the same, just the component name is tf-serving-template
- e2e test is now using the new tf-serving-template

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1589)


jlewi:
I think we're headed in the direction of having multiple prototypes (we already have 2) this adds a 3rd.
I think if we follow my original suggestion of putting everything into the prototype (.jsonnet) file we will get a lot of unnecessary code duplication that will be difficult to maintain overtime.

@kkasravi's recent prototype #1501 suggests to me (see [comment](https://github.com/kubeflow/kubeflow/issues/376#issuecomment-422377216)) that we should define a base libsonnet file and then use k8s libsonnet to modify it for each prototype.

I considered having a basic prototype with all the logic in the .jsonnet file that people can use as a simple example so they don't have to learn the k8s lib. But even that seems like it would get pretty unruly.

My initial thought is that we should have the following prototypes 

* TFServing using S3
* TFServing using GCS
* TFServing using PVC

So just supporting those 3 different use cases will lead to a lot of code duplication if we put all the code into the .jsonnet file.

lluunn:
I was thinking the we can use only this one jsonnet later.
The optional features are not mutually exclusive. They (GCP credential, S3, istio, request log) can co-exist, so have their own "enabled" flag.

Here the jsonnet structure has two parts (before and after line 75):
1. Flags and customization.
    Flags for enabling features and can be easily set. Customization for each feature can be set by flags 
    or by modify the component directly (the component is the copy of whole jsonnet)
1. Service/Container/Deployment
    This part may look a little complicated as features being added, `env: [] + if XX then [a] else [] + ...`
    But it should not need to be changed by users.

WDYT?

jlewi:
One of the reasons for having parameters is for making it easy for users to figure out what they have to set.
The problem with the way its currently configured is that the parameters don't define the things the user has to set for S3 vs. GCS. They'd have to read the code to realize there are parameters like "gcpCredentialSecretName" and "s3SecretName".

If we have separate prototypes for GCP and S3. Then in the prototype for GCP we can explicitly define all the relevant parameters e.g.

```
// @optionalParam gcpSecretName string "user-sa" Name of the secret containing the GCP service account
```

Then when the user invokes help on the component they can get a meaningful list of parameters that they could set.

lluunn:
But how about istio and request logs? They can co-exist with gcp credentials and S3.
We can't make one prototype just for istio and another just for request logs because users might want both (and gcp credential as well).

We can add both gcpSecretName and s3Enabled as flags in this prototype, and add better documentation. WDYT?

kubeflow/tf-serving/prototypes/tf-serving.jsonnet
```diff
@@ -0,0 +1,240 @@
...
+        port: 8000,
+        targetPort: 8000,
+      },
+    ],
+    selector: {
+      app: appName,
+    },
+    type: "ClusterIP",
+  },
+};  // service
```
kkasravi: I would add a 
service:: service

as described [here](https://github.com/databricks/jsonnet-style-guide#defining-classes)

kubeflow/tf-serving/prototypes/tf-serving.jsonnet
```diff
@@ -0,0 +1,240 @@
...
+  image: modelServerImage,
+  imagePullPolicy: "IfNotPresent",
+  name: name,
+  ports: [
+    {
+      containerPort: 9000,
+    },
+  ],
+  env: []
+       + if util.toBool(params.s3Enable) then s3Env else []
```
kkasravi: I would move this into deployment.mapContainers  in the kubeflow/tf-serving/prototypes/tf-serving-aws.jsonnet file as described above

kubeflow/tf-serving/prototypes/tf-serving.jsonnet
```diff
@@ -0,0 +1,240 @@
...
+        },
+      },
+      spec: {
+        containers: [
+          modelServerContainer,
+        ] + if util.toBool(params.deployHttpProxy) then [
+          httpProxyContainer,
+        ] else [],
+        volumes: []
+                 + if gcpParams.gcpCredentialSecretName != "" then
```
kkasravi: move into a kubeflow/tf-serving/prototypes/tf-serving-gcp.jsonnet file

lluunn:
Thanks @jlewi @kkasravi , PTAL

lluunn:
argo test failed in kfctl
Failed to run argo workflow

lluunn:
argo test failed again, there might be an issue
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-kfctl-1589-636f49f-3514-3e66?tab=workflow

lluunn:
still argo test failure, but I don't see how this PR is related to that.

lluunn:
https://pantheon.corp.google.com/logs/viewer?project=kubeflow-ci&minLogLevel=0&expandAll=false&timestamp=2018-09-26T20:45:52.695000000Z&customFacets=&limitCustomFacetWidth=true&interval=NO_LIMIT&resource=container%2Fcluster_name%2Fkubeflow-testing&scrollTimestamp=2018-09-26T18:07:57.000000000Z&advancedFilter=resource.type%3D%22container%22%0Aresource.labels.cluster_name%3D%22kfctl-7247%22%0Aresource.labels.pod_id%3D%22hello-world-p822r%22%0A

I can see the argo pod up at 11:07:47, but it timeout at 07:03. So I am extending the wait time.

kubeflow/tf-serving/util.libsonnet
```diff
@@ -1,5 +1,7 @@
 // Some useful routines.
 {
+  local k = import "k.libsonnet",
+
   // Convert a string to upper case.
   upper:: function(x) {
```
kkasravi: std library has std.asciiUpper(str)
http://jsonnet.org/ref/stdlib.html

kkasravi:
@lluunn jfmt errors

lluunn:
Thanks @kkasravi

jlewi:
This looks great.
/lgtm
/approve

lluunn:
not sure why notebook release workflow is triggered...

PR #1592, opened 2018-09-21
Update Seldon to 0.2.3 version

cliveseldon:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1592)


jlewi:
Thanks@cliveseldon

Would you mind cherrypicking the same change onto the 0.3 branch?

Issue #1588, opened 2018-09-20
Envoy unable to read config


kunmingg:
Saw error message from envoy pod:
external/envoy/source/server/server.cc:71] error initializing configuration '/etc/envoy/envoy-config.json': unable to read file: /etc/envoy/envoy-config.json

@abhi-g

jlewi:
@kunmingg What was the context in which this error appeared? Did envoy end up recovering?

Which envoy were you looking at? Is this the Ambassador envoy instance or the envoy we use to do JWT checking for IAP?

kunmingg:
https://github.com/kubeflow/kubeflow/pull/1595
This one fix it.
Root cause is IAM leftovers from previous deployment.

kunmingg:
@connected-bsamadi 
There are one more issue can could cause it:
https://github.com/kubeflow/kubeflow/issues/1807

Can you retry with latest master?
Will include fix in v0.3.2

PR #1595, opened 2018-09-21
Force update IAM policy; print app url on web UI.

kunmingg:
Need to force update IAM policy in case there are existing orphan IAM bindings for related service accounts in target project (accounts deleted but bindings are still there).

fix https://github.com/kubeflow/kubeflow/issues/1588

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1595)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -447,7 +447,8 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
               'deployment failed with error:' + flattenDeploymentOperationError(r.operation!));
             clearInterval(monitorInterval);
           } else if (r.operation!.status! && r.operation!.status === 'DONE') {
-            this._appendLine('Deployment is done');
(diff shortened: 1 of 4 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -447,7 +447,8 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
               'deployment failed with error:' + flattenDeploymentOperationError(r.operation!));
             clearInterval(monitorInterval);
           } else if (r.operation!.status! && r.operation!.status === 'DONE') {
-            this._appendLine('Deployment is done');
+            this._appendLine('Deployment is done, your kubeflow app url should be ready within 15 minutes: '
+              + this.state.deploymentName + '.endpoints.' + this.state.project + '.cloud.goog');
```
abhi-g: should this add 'https://' before the this.state.deploymentName?

abhi-g:
looks good.

PR #1612, opened 2018-09-23
fixes 'TFJob will not run if the TFJob CRD includes scope: Namespaced'

### LINKED ISSUE #1606: TFJob will not run if the TFJob CRD includes scope: Namespaced
@jlewi There is a subtle issue in tf-job-operator

this works and results in tf-job-simple running
```
---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  labels:
    ksonnet.io/component: tf-job-operator
  name: tfjobs.kubeflow.org
spec:
  group: kubeflow.org
  names:
    kind: TFJob
    plural: tfjobs
    singular: tfjob
  validation:
    openAPIV3Schema:
      properties:
        spec:
          properties:
            tfReplicaSpecs:
              properties:
                Chief:
                  properties:
                    replicas:
                      maximum: 1
                      minimum: 1
                      type: integer
                PS:
                  properties:
                    replicas:
                      minimum: 1
                      type: integer
                Worker:
                  properties:
                    replicas:
                      minimum: 1
                      type: integer
  version: v1alpha2
```
and this doesn't (difference is the inclusion of the scope attribute).
There are no errors that I can see and tf-job-simple never runs.

```
---
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  labels:
    ksonnet.io/component: tf-job-operator
  name: tfjobs.kubeflow.org
spec:
  group: kubeflow.org
  names:
    kind: TFJob
    plural: tfjobs
    singular: tfjob
  scope: cluster
  validation:
    openAPIV3Schema:
      properties:
        spec:
          properties:
            tfReplicaSpecs:
              properties:
                Chief:
                  properties:
                    replicas:
                      maximum: 1
                      minimum: 1
                      type: integer
                PS:
                  properties:
                    replicas:
                      minimum: 1
                      type: integer
                Worker:
                  properties:
                    replicas:
                      minimum: 1
                      type: integer
  version: v1alpha2
```

Since the default for CustomResourceDefinition scope is [Namespaced](https://godoc.org/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions) is seems like this would be a bug. I couldn't find in tf-operator repo where this is referenced.

kkasravi:
fixes #1606

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1612)

### DIFF
--- a/kubeflow/core/prototypes/tf-job-operator.jsonnet
+++ b/kubeflow/core/prototypes/tf-job-operator.jsonnet
@@ -9,7 +9,7 @@
 // @optionalParam tfDefaultImage string null The default image to use for TensorFlow.
 // @optionalParam tfJobUiServiceType string ClusterIP The service type for the UI.
 // @optionalParam tfJobVersion string v1alpha2 which version of the TFJob operator to use
-// @optionalParam deploymentScope string cluster The scope at which tf-job-operator should be deployed - valid values are cluster, namespace.
+// @optionalParam deploymentScope string namespace The scope at which tf-job-operator should be deployed - valid values are cluster, namespace.
 // @optionalParam deploymentNamespace string null The namespace to which tf-job-operator should be scoped. If deploymentScope is set to cluster, this is ignored.
 
 local tfJobOperator = import "kubeflow/core/tf-job-operator.libsonnet";
--- a/kubeflow/core/tests/tf-job_test.jsonnet
+++ b/kubeflow/core/tests/tf-job_test.jsonnet
@@ -11,8 +11,8 @@ local paramsv1alpha2 = {
   name:: "tf-job-operator",
   tfJobImage:: "gcr.io/kubeflow-images-public/tf_operator:v20180226-403",
   tfDefaultImage:: "null",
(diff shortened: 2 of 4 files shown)


jlewi:
@kkasravi Can you explain what this PR is doing?

DeploymentScope refers to whether we want to our TFJob controller to monitor all namespaces or just a single namespace that it is deployed in; see kubeflow/tf-operator#759

This is different from the definition of scope in the CRD. The value of scope in the CRD should always be namespace because TFJobs should always belong in a namespace.

When you refactored the TFJob prototype did we lose support for scoping the operator to a particular naemspace?

kkasravi:
the deploymentScope and deploymentNamespace are the same as the earlier version and scoping the operator to a particular namespace should be the same. What's on master works and i've run tf-job-simple as well. This PR was trying to do a better job of checking if deploymentScope == namespace then deploymentNamespace shouldn't be null. I can close this for now until I have a bit more time to look at the failure.

PR #1621, opened 2018-09-24
make builder image version consistent with glide.lock hash

kunmingg:
To simplify test and local dev.

Old method:
go pkg was compiled in builder image:
1. every line of change requires build two images: builder and final image. 
2. resulting in need of edit builder image version in Makefile every time before run ```make build``` 
3. resulting in frequent version change of builder image.

New method:
builder image is cache of glide & vendor dir, ```glide.lock``` file 100% determines the content of builder image. So use hash code of ```glide.lock``` as builder image version. And resolve 1,2,3 of old method

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1621)

### DIFF
--- a/bootstrap/Dockerfile
+++ b/bootstrap/Dockerfile
@@ -4,6 +4,11 @@ ARG GOLANG_VERSION=1.10.3
 ARG BUILDER_IMG=gcr.io/kubeflow-images-public/bootstrapper-builder
 ARG BUILDER_IMG_VERSION=latest
 FROM ${BUILDER_IMG}:${BUILDER_IMG_VERSION} AS build_base_remote
+# Copy the src
+COPY ./ $GOPATH/src/github.com/kubeflow/kubeflow/bootstrap/
+
+RUN go build ${GOLANG_GCFLAGS} -i -o /opt/kubeflow/bootstrapper \
+    ${GOPATH}/src/github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/main.go
 
 FROM golang:${GOLANG_VERSION} AS build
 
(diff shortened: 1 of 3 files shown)


bootstrap/Dockerfile
```diff
@@ -4,6 +4,11 @@ ARG GOLANG_VERSION=1.10.3
 ARG BUILDER_IMG=gcr.io/kubeflow-images-public/bootstrapper-builder
 ARG BUILDER_IMG_VERSION=latest
 FROM ${BUILDER_IMG}:${BUILDER_IMG_VERSION} AS build_base_remote
+# Copy the src
+COPY ./ $GOPATH/src/github.com/kubeflow/kubeflow/bootstrap/
+
+RUN go build ${GOLANG_GCFLAGS} -i -o /opt/kubeflow/bootstrapper \
+    ${GOPATH}/src/github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/main.go
 
 FROM golang:${GOLANG_VERSION} AS build
```
lluunn: what's this line for?
Do we need to build after this line?

bootstrap/Dockerfile.Builder
```diff
@@ -18,10 +18,3 @@ RUN mkdir -p $GOPATH/src/github.com/kubeflow/kubeflow/bootstrap
 COPY ./glide* $GOPATH/src/github.com/kubeflow/kubeflow/bootstrap/
 RUN cd $GOPATH/src/github.com/kubeflow/kubeflow/bootstrap/ && \
     glide install -v
-
```
lluunn: So builder only containers vendor, and we only need to rebuild it if glide is changed, right?

bootstrap/Dockerfile
kunmingg: Which line?

bootstrap/Dockerfile.Builder
kunmingg: Yes

Issue #1435, opened 2018-08-27
Update CentralUI image used at head and 0.3 branch


jlewi:
We committed a number of improvements to the central UI.

We need to update the docker image used at head in order to test out these changes.

jlewi:
@swiftdiaries Are we continuously building the central UI image? If we have an updated image can you create a PR to update the image used in ksonnet?

swiftdiaries:
I don't think we are building it continuously. I'll get a PR in to fix that.

Issue #1617, opened 2018-09-24
[test flake] pre and postsubmit failures deploying mnist


jlewi:
Postsubmit failure:

https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/logs/kubeflow_kubeflow/kubeflow-postsubmit/700

Error message is

```
Test failed; Timeout waiting for deployment mnist-cpu-v1 in namespace kubeflow-postsubmit-tf-serving-image-b518499-700-07b2
```

Also saw presubmit failures in
kubeflow/kubeflow#1614

which is surprising because in that PR the tf serving test shouldn't have been triggered based on the files that were modified.

/cc @lluunn

lluunn:
I think the cluster is having some issue:

```
Unschedulable pods
kubeflow-postsubmit-dashboard-release-00d2921-707-01a8-2504870139
kubeflow-postsubmit-dashboard-release-875e315-703-6404-1430270421
kubeflow-postsubmit-dashboard-release-a65bb97-710-b8ff-750234357
kubeflow-postsubmit-dashboard-release-ced3d0c-708-61ee-3757777039
kubeflow-postsubmit-dashboard-release-f254e0a-709-91a3-2584187888
and other 25 unschedulable pods.
```

lluunn:
Deleted some leaked workflow (not many)

There are 3 cpu nodes
```
kubectl get pods --field-selector="status.phase==Running,spec.nodeName=gke-kubeflow-testing-default-pool-90dc0402-bj6f" --all-namespaces
NAMESPACE             NAME                                                                  READY     STATUS    RESTARTS   AGE
devstats              cert-manager-555c87df98-ts7wn                                         2/2       Running   532        136d
devstats              grafana-945fbcf7d-ccpvr                                               1/1       Running   0          4d
kube-system           fluentd-gcp-v2.0.17-kwc9b                                             2/2       Running   3          5d
kube-system           heapster-v1.5.2-68b6f5dc47-fr4pv                                      3/3       Running   3          5d
kube-system           kube-dns-56dc69f5bd-rxjv2                                             4/4       Running   4          5d
kube-system           kube-proxy-gke-kubeflow-testing-default-pool-90dc0402-bj6f            1/1       Running   4          167d
kube-system           kubernetes-dashboard-7b5c8f9668-gbkcc                                 1/1       Running   5          124d
kube-system           metrics-server-v0.2.1-7f8dd98c8f-42rdc                                2/2       Running   39         124d
[...]

PR #1623, opened 2018-09-24
Build image for centralui part of presubmit

swiftdiaries:
Addresses: https://github.com/kubeflow/kubeflow/issues/1435

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1623)


lluunn:
I think the presubmit cluster is having some issue
```
Unschedulable pods
kubeflow-postsubmit-dashboard-release-00d2921-707-01a8-2504870139
kubeflow-postsubmit-dashboard-release-875e315-703-6404-1430270421
kubeflow-postsubmit-dashboard-release-a65bb97-710-b8ff-750234357
kubeflow-postsubmit-dashboard-release-ced3d0c-708-61ee-3757777039
kubeflow-postsubmit-dashboard-release-f254e0a-709-91a3-2584187888
and other 26 unschedulable pods.

```

lluunn:
See #1617

swiftdiaries:
Thank you! @lluunn

jlewi:
Most recent test flake looks like an issue with teardown. Looks like it couldn't find the kubeconfig file. 

```
File "/usr/local/lib/python2.7/dist-packages/kubernetes/config/kube_config.py", line 458, in _get_kube_config_loader_for_yaml_file
   with open(filename) as f:
IOError: [Errno 2] No such file or directory: '/root/.kube/config'
```

This is a bug in the release workflow. The release workflow shouldn't be calling "teardown" because the release workflow doesn't create any resources (e.g. there is no setup step).

I think this is a bug in the release workflow
https://github.com/kubeflow/kubeflow/blob/320c37903970340a795bf49f3a1a54a121c336de/kubeflow/automation/release.libsonnet#L281

We probably shouldn't be running test_deploy on teardown. That looks like a bug in the release workflow.
I think we need to remove that step in kubeflow/automation/release.libsonnet
You'll then want to make a copy of that updated file in the vendor directory of the releasing ksonnet app.

components/centraldashboard/build_image.sh
```diff
@@ -3,21 +3,24 @@
...
 #
-# build_image.sh ${DOCKERFILE} ${IMAGE} ${TAG} (optional | ${TEST_REGISTRY})
+# build_image.sh ${DOCKERFILE} ${IMAGE} ${TAG}
 set -ex
 
 DOCKERFILE=$1
 IMAGE=$2
 TAG=$3
+CONTEXT_DIR=$(dirname "$DOCKERFILE")
+PROJECT="{GCP_PROJECT}"
```
jlewi: Is there a reason you are passing these as environment variables and the others as command line arguments?

jlewi: Actually looking at the workflow I think it makes sense to continue to pass as environment variables since they're optional; e.g. only needed in case of use of GCB.

releasing/releaser/vendor/kubeflow/automation/release.libsonnet
```diff
@@ -108,6 +108,14 @@
                 },
               },
             },
+            {
```
jlewi: can you make the same change in
kubeflow/automation/release.libsonnet

Otherwise your changes will be overwritten if we ever upgrade the release prototype

swiftdiaries: Okay :)

swiftdiaries: Thanks for the review

components/centraldashboard/build_image.sh
swiftdiaries: 👍

swiftdiaries:
Yeah. There's an image pull failure in the checkout step. Any idea what might be causing that?

prow_config.yaml
```diff
@@ -62,7 +62,10 @@ workflows:
     name: dashboard-release
     job_types:
       # TODO(jlewi): Reenable on presubmit once its passing.
-      #- presubmit
+      - presubmit
+      # TODO(jlewi): We should probably create a separate postsubmit job that pushes
```
richardsliu: This can be removed. I recently added the postsubmit job below that pushes the image to kubeflow-images-public.

swiftdiaries:
Hitting this issue on teardown: https://github.com/kubeflow/testing/issues/209.
Argo logs: http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-dashboard-release-1623-496efe6-3597-fd73?tab=workflow&nodeId=kubeflow-presubmit-dashboard-release-1623-496efe6-3597-fd73-1556967763&sidePanel=logs%3Akubeflow-presubmit-dashboard-release-1623-496efe6-3597-fd73-1556967763%3A

Issue #1601, opened 2018-09-22
Update Jupyter images on 0.3. branch


jlewi:
We need to update the Jupyter images on the 0.3 branch

The Jupyter images should now be built by postsubmits. The postsubmits aren't being marked as succesful though but that should be fixed by #215.

Nonetheless if we have successfully built Jupyter images we should go ahead and get them updated on the release branch.

Related to #1541 0.3 release

/assign @richardsliu

jlewi:
Lets make sure we have docker images with TF.Data validation
https://github.com/kubeflow/kubeflow/pull/1629

PR #1624, opened 2018-09-24
Add postsubmits to push centralui and Jupyter notebook images to kubeflow-images-public

richardsliu:
This change adds post-submit jobs that push CentralDashboard and Jupyter notebook images to gcr.io/kubeflow-images-public.

Related to #1601 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1624)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -63,9 +63,6 @@ workflows:
     job_types:
       # TODO(jlewi): Reenable on presubmit once its passing.
       #- presubmit
-      # TODO(jlewi): We should probably create a separate postsubmit job that pushes
-      # to kubeflow-images-public.
-      - postsubmit
     params:
       cluster: "kubeflow-testing"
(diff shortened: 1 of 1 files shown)


prow_config.yaml
```diff
@@ -77,14 +74,25 @@ workflows:
...
       - components/centraldashboard/*
       - releasing/releaser/components/centraldashboard.jsonnet
+  # The postsubmit run publishes the docker images to gcr.io/kubeflow-images-public
+  - app_dir: kubeflow/kubeflow/releasing/releaser
+    component: centraldashboard
+    name: dashboard-release
+    job_types:
+      - postsubmit
+    params:
+      cluster: "kubeflow-testing"
```
jlewi: Nit here and below. Only set the parameters that we want to override. So in this case I think we only want to set registry. For everything else we can use the defaults.

richardsliu:
Done.

Issue #1604, opened 2018-09-22
Update Katib image in 0.3 branch


jlewi:
Tracking bug for updating the Katib images on the 0.3 branch

#1597 updates it on master. Will need to cherry pick it onto the release branch.

PR #1625, opened 2018-09-24
Update katib image URIs

ankushagarwal:
Some of these don't have the katib- prefix
/assign @jlewi

Related to #1604  Update Katib images

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1625)


jlewi:
Looks like there are some failures
/lgtm
/approve

PR #1628, opened 2018-09-26
Use JupyterNotebook as default instead of JupyterLab

ankushagarwal:
JupyterLab extensions work differently that JupyterNotebook
Once we have tested and validated JupyterLab, we can switch to it as
default.

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1628)


Ark-kun:
> Once we have tested and validated JupyterLab, we can switch to it as default.

Has this been validated?

Issue #1483, opened 2018-09-05
Test to verify we can deploy Katib


jlewi:
We should have an E2E test for our katib ksonnet component. We can start by ensuring that Katib is actually deployed.

See #1478 for an example bug where a bug in the ksonnet component prevented Katib from being deployed.

Ideally we'll have this in 0.3.

jlewi:
The E2E test in kubeflow/katib uses
https://github.com/kubeflow/katib/blob/master/test/scripts/run-tests.sh

Which uses some YAML manifests.

A good starting point would just be deploying Katib and then verifying all the pods start.

We should be able to piggy back on our existing ksonnet lib for deploying kubeflow using kfctl.
https://github.com/kubeflow/kubeflow/blob/764d28e520c643f6596931cc64fe9e1b2e0a7800/testing/workflows/components/workflows.libsonnet#L58

For an example of how this is used see:
https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_test.jsonnet

We could then just extend the workflow to include the steps to deploy Katib. We could copy the verification steps from the existing katib E2E test.

PR #1646, opened 2018-09-27
[Cherry-pick] Add a simple e2e-test for katib

ankushagarwal:
Master PR #1638
Related to #1483

This is run as part of the kfctl e2e test after setting up the kubeflow on the gke cluster.

It uses kubectl to verify the various deployments made by katib ksonnet component.

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1646)

### DIFF
--- a/scripts/kfctl.sh
+++ b/scripts/kfctl.sh
@@ -35,7 +35,7 @@ createEnv() {
   # Namespace where kubeflow is deployed
   echo K8S_NAMESPACE=${K8S_NAMESPACE:-"kubeflow"} >> ${ENV_FILE}
 
-  case "$PLATFORM" in 
+  case "$PLATFORM" in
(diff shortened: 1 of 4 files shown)


scripts/kfctl.sh
```diff
@@ -210,6 +210,7 @@ ksApply () {
   ks apply default -c centraldashboard
   ks apply default -c tf-job-operator
   ks apply default -c argo
+  ks apply default -c katib
```
abhi-g: This would mean that katib will get deployed by default into the installations. Is that what we want?

ankushagarwal: I am assuming that 0.3 onwards we want to install katib by default

PR #1650, opened 2018-09-28
Update initial clsuter version in cluster.jinja based on what gcloud get-server-config returns

### LINKED ISSUE #1359: kfctl.sh needs to get initial cluster version based on getServerConfig
GKE deprecates patch versions for the master (see #1348) e.g.  1.9.7-gke.3 -> 1.9.7-gke.5 quite frequently.

We don't want to auto-upgrade to the latest minor version because that can also cause breaks. For example, in 1.9 to 1.10 the behavior of GCE ingress changed breaking certmanager (kubernetes/ingress-gce#388).

The short term solution is to have kfctl.sh call 

```
gcloud --project=cloud-ml-dev container get-server-config
```

and select an appropriate version.

ashahba:
fixes #1359

### DIFF
--- a/scripts/gke/deployment_manager_configs/cluster.jinja
+++ b/scripts/gke/deployment_manager_configs/cluster.jinja
@@ -22,10 +22,10 @@ limitations under the License.
    There is type corresponding to each API endpoint.
 #}
 
-{% set VM_OAUTH_SCOPES = ['https://www.googleapis.com/auth/logging.write', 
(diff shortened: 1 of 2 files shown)


ashahba:
/cc @gaocegege @richardsliu @jlewi 
This is ready.
For now we grab the first one from within `validMasterVersions` list while sticking with `1.10` still.

jlewi:
This has been a huge pain point; thank you so much fixing this.
/lgtm
/approve

PR #1661, opened 2018-09-29
Change kubebench 0.3 image

xyhuang:
The previous 0.3 image was not working properly due to an issue in build, this updates the reference to the new image.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1661)


xyhuang:
Please let me know if we should create a new tag or just re-tag.

jlewi:
We can just retag until the v0.3.0 is finalized (right now its just RC's.).

After this is merged someone with appropriate permissions will need to run the command to apply the images.

PR #1674, opened 2018-09-30
gke/deploy.sh: test uuidgen exists before using it

rabierp:
line 170: uuidgen is used for generating a unique identifyer for optional usage metrics collection.
But uuidgen isn't available in Cloud Shell, and deploy.sh exists with an error

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1674)


rabierp:
Hi Jeremi,

Is kfctl.sh also the recommend install script for KF 2.x?
I would keep deploy.sh for 2.x because there are lots of articles/how-tos
already published that use it.
And even quite recent ones describe installing 2.5 with deploy.sh.

Pascal

jlewi:
The jsonnet test failures should be unrelated to this PR so I'll go ahead and merge this.

PR #1680, opened 2018-10-01
fix katib metrics collector bug

YujiOshima:
The manifest of metrics collector needs namespace.

Signed-off-by: YujiOshima <yuji.oshima0x3fd@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1680)

### DIFF
--- a/kubeflow/katib/studyjobcontroller.libsonnet
+++ b/kubeflow/katib/studyjobcontroller.libsonnet
@@ -303,7 +303,7 @@
                   - name: {{.WorkerId}}
                     image: alpine
                   restartPolicy: Never
-          |||% {ns: namespace},
+          ||| % { ns: namespace },
         },
       },
     ],
--- a/kubeflow/katib/studyjobcontroller.libsonnet
+++ b/kubeflow/katib/studyjobcontroller.libsonnet
@@ -97,7 +97,7 @@
           namespace: namespace,
         },
         data: {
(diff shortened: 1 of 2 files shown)


kubeflow/katib/studyjobcontroller.libsonnet
```diff
@@ -97,7 +97,7 @@
...
+          "defaultMetricsCollectorTemplate.yaml": 'apiVersion: batch/v1beta1\nkind: CronJob\nmetadata:\n  name: {{.WorkerId}}\n  namespace: {{.NameSpace}}  spec:\n  schedule: "*/1 * * * *"\n  successfulJobsHistoryLimit: 1\n  failedJobsHistoryLimit: 1\n  jobTemplate:\n    spec:\n      template:\n        spec:\n          serviceAccountName: metrics-collector\n          containers:\n          - name: {{.WorkerId}}\n            image: katib/metrics-collector\n            args:\n            - "./metricscollector"\n            - "-s"\n            - "{{.StudyId}}"\n            - "-t"\n            - "{{.TrialId}}"\n            - "-w"\n            - "{{.WorkerId}}"\n            - "-n"\n            - "{{.NameSpace}}"\n          restartPolicy: Never',
```
jlewi: Can you use a verbatim string
https://jsonnet.org/learning/tutorial.html
using |||
This should allow it to be formated as a multiline string which is more readable.

kubeflow/katib/studyjobcontroller.libsonnet
```diff
@@ -261,7 +290,20 @@
...
+            metadata:
+              name: {{.WorkerId}}
+            spec:
+              template:
+                spec:
+                  containers:
+                  - name: {{.WorkerId}}
+                    image: alpine
+                  restartPolicy: Never
+          |||% {ns: namespace},
```
jlewi: Is this a typo? Why is "% {ns:namespace} outside the verbatim string?

jlewi:
Looks like your jsonnet isn't formatted.
/mnt/test-data-volume/kubeflow-presubmit-unittests-1680-35c2780-3655-1551/src/kubeflow/kubeflow/kubeflow/katib/studyjobcontroller.libsonnet is not formatted

You can run
https://github.com/kubeflow/kubeflow/blob/master/scripts/autoformat_jsonnet.sh

YujiOshima:
@jlewi Thanks! I updated!

kubeflow/katib/studyjobcontroller.libsonnet
YujiOshima: This is not typo. It is string formatting same as example 3 in https://jsonnet.org/ .

PR #1703, opened 2018-10-03
Add TOS and Privacy links and some other UI improvements.

jlewi:
* Improve the instructions
  * Turn instructions into a numbered list

* Move the Kubeflow version field to the bottom of the form since in most
  cases users won't need to set it.

Here's a screen shot of the updated UI

webpage_ui

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1703)

PR #1705, opened 2018-10-03
config & dockerfile update

kunmingg:
want to be included in 0.3 release if possible.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1705)


jlewi:
/lgtm
/approve

It looks like the katib prototype isn't properly formatted (not as a result of this PR).
Should be fixed by #1703 but you could also patch in those changes.

#1703 will need your lgtm to be submitted.

jlewi:
@kunmingg 

```

```

Issue #1309, opened 2018-08-03
Use Istio 1.0


lluunn:
We integrated istio for TF Serving with istio 0.7.
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/examples/prototypes/tf-serving-with-istio.jsonnet#L17
https://www.kubeflow.org/docs/guides/components/istio/

Now that istio is 1.0, we should migrate to that.

lluunn:
#1823 actually doesn't work: I set the traffic to v1:100,v2:0, but still see requests going to v2.

And the reason is that istio uses client-side proxy to route requests, but in this case the client (ambassador) service is not in istio mesh.

So I think options are:
- add one additional hop
- don't use ambassador

lluunn:
Another option: ambassador supports percentage traffic routing:
https://www.getambassador.io/reference/canary

jlewi:
@lluunn Can you clarify the scope of this issue?

Is this issue just about changing the version of ISTIO from 0.7 to 1.0? Or is it more involved then that?

What work remains on this issue?

lluunn:
I think we can close this one and file separate issues.
My original thought was only to change Istio version from 0.7 to 1.0, which is done now.

We have #1844 for rollout.
Filed #1907 for jwt validation.

PR #1714, opened 2018-10-04
TF serving template cleanup

lluunn:
1. remove http proxy as we have http from TF serving. We should be able use istio for request logging [1], so http proxy is not needed anymore.
2. Add istio annotation. Will update documentation for this later (in website repo). Related: #1309 

[1]: https://istio.io/docs/tasks/telemetry/fluentd/#configure-istio

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1714)


jlewi:
@lluunn How do we use ISTIO for request and response logging?

lluunn:
From this: https://istio.io/docs/tasks/telemetry/fluentd/#configure-istio
I think we can configure istio's "logentry" to log the payload.

jlewi:
Lets open an issue to track whether we can use ISTIO to log requests and responses.
/lgtm
/approve

PR #1716, opened 2018-10-05
Update gcp credentials filename

### LINKED ISSUE #1715: [GCP] Update credentials filename
TF-serving sets the value of GOOGLE_APPLICATION_CREDENTIALS to the contents of [key.json](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/tf-serving/tf-serving.libsonnet#L342), which is incompatible with the secret as setup by the deploy script (user-gcp-sa.json). This needs to be updated to match the filename used by the automated deploy.

texasmichelle:
Fixes #1715 

Replace key.json with user-gcp-sa.json


lluunn:
/lgtm
/approve
thanks

Issue #1721, opened 2018-10-05
jupyter_console needs update in the latest notebooks


pdmack:
```
ModuleNotFoundError: No module named 'prompt_toolkit.formatted_text'
```

`pip install -U jupyter_console`

/cc @richardsliu 
/cc jlewi

leoncamel:
I check the `Dockerfile`, I think the `pip install jupyter` package will lead to `prompt-toolkit-1.5.x'. After I remove [the jupyter package](https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/Dockerfile#L100), this image works here.

jlewi:
@pdmack Any chance you could provide a PR to update our images?

pdmack:
AFAIK, we just need to pick up jupyter/jupyter_console#161 (jupyter-console==6.0.0) in new notebook builds.

jlewi:
@pdmack Can you provide a PR updating the Docker builds?

PR #1726, opened 2018-10-08
fix tensorflow-notebook's prompt_toolkit bug (#1721)

leoncamel:
Since the `jupyter` package is quite out-of-date. It will install `prompt-toolkit-1.5.x` for some reason(the python3). The python2 will also use `prompt-toolkit-1.5.x`, which is works well.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1726)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -97,7 +97,6 @@ RUN pip install --upgrade pip && \
     # Tensorflow
     ${TF_PACKAGE} \
     # Jupyter Stuff
-    jupyter \
     jupyterhub \
     jupyterlab \
     # Cleanup


components/tensorflow-notebook-image/Dockerfile
```diff
@@ -97,7 +97,6 @@ RUN pip install --upgrade pip && \
     # Tensorflow
     ${TF_PACKAGE} \
     # Jupyter Stuff
-    jupyter \
```
jlewi: Why are you removing Jupyter; Don't we need that to run Jupyter?

pdmack:
This should be effectively addressed by #1746. Recommend closing unless I'm missing something.

components/tensorflow-notebook-image/Dockerfile
leoncamel: It seems `jupyter` package is an meta package, it looks quite out of date for me.

But, it is OK, if we add `jpyter-console==0.6.0`.

Let's close this PR.

leoncamel:
Since [PR 1746](https://github.com/kubeflow/kubeflow/pull/1746) can solve this issue. Let's close this PR.

Issue #1450, opened 2018-08-31
Does CentralUi need ClusterScope?


jlewi:
See #1213 

We'd like to be able to scope Kubeflow to a particular namespace to better support multi-tenant environments.

It looks like CentralUI is creating a ClusterRole right now. Do we need ClusterScope? Could we use a namespace scoped role?

/cc @swiftdiaries

swiftdiaries:
If Kubeflow components (Argo, job-operator, jupyter) are deployed to same namespace then CentralUI doesn't need ClusterScope. A namespaced scope should be good enough.

jlewi:
@swiftdiaries Until proven otherwise, I think we can assume central dashboard is in the same name space as the other components.

So lets replace the cluster scoped role with a namespace scoped role.

swiftdiaries:
Sounds good 👍  Will get a PR in with the change.

swiftdiaries:
Closing because a PR has already been submitted and merged for this.
/close

PR #1728, opened 2018-10-08
change clusterrole to role

swiftdiaries:
Addresses issue: https://github.com/kubeflow/kubeflow/issues/1450. 
Assuming all kubeflow components are going to installed to a single namespace, CentralUI does not need cluster level scope. This PR is to restrict CentralUI's scope down to a single namespace.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1728)


jlewi:
Would you mind adding some unittests as part of this PR to verify that the manifests are correct?
Here's an example
https://github.com/kubeflow/kubeflow/tree/master/kubeflow/core/tests

jlewi:
Looks like you just need to format the jsonnet files.

swiftdiaries:
Thanks 👍

PR #1754, opened 2018-10-11
fixes tensorboard generated under kubeflow/tensorboard/prototypes/tensorboard-aws.jsonnet is incomplete

### LINKED ISSUE #1753: tensorboard generated under kubeflow/tensorboard/prototypes/tensorboard-aws.jsonnet is incomplete
The generated deployment for tensorboard only contains the environment variables in the container.
See below
```
  "Deployment.apps/v1beta1": {
    "tensorboard-aws": {
      "apiVersion": "apps/v1beta1",
      "kind": "Deployment",
      "metadata": {
        "labels": {
          "app": "tensorboard-aws"
        },
        "name": "tensorboard-aws",
        "namespace": "kf-103-user"
      },
      "spec": {
        "replicas": 1,
        "template": {
          "metadata": {
            "labels": {
              "app": "tensorboard-aws"
            }
          },
          "spec": {
            "containers": [
              {
                "env": [
                  {
                    "name": "AWS_ACCESS_KEY_ID",
                    "valueFrom": {
                      "secretKeyRef": {
                        "key": "AWS_ACCESS_KEY_ID",
                        "name": "secret-sauce"
                      }
                    }
                  },
                  {
                    "name": "AWS_SECRET_ACCESS_KEY",
                    "valueFrom": {
                      "secretKeyRef": {
                        "key": "AWS_SECRET_ACCESS_KEY",
                        "name": "secret-sauce"
                      }
                    }
                  },
                  {
                    "name": "AWS_REGION",
                    "value": "us-east-1"
                  },
                  {
                    "name": "S3_REGION",
                    "value": "us-east-1"
                  },
                  {
                    "name": "S3_USE_HTTPS",
                    "value": "true"
                  },
                  {
                    "name": "S3_VERIFY_SSL",
                    "value": "true"
                  },
                  {
                    "name": "S3_ENDPOINT",
                    "value": "http://s3.us-east-1.amazonaws.com"
                  }
                ]
              }
            ]
          }
        }
      }
    }
  },
```

kkasravi:
Related to #1753

Will need to cherry pick onto v0.3-branch

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1754)

### DIFF
--- a/kubeflow/tensorboard/gcp.libsonnet
+++ b/kubeflow/tensorboard/gcp.libsonnet
@@ -8,37 +8,37 @@
 
   local tbDeployment =
     super.tbDeployment +
-      deployment.mixin.spec.template.spec.withVolumesMixin(
-        if params.gcpCredentialSecretName != "" then (
-          [{
-            name: "gcp-credentials",
-            secret: {
-              secretName: params.gcpCredentialSecretName,
-            },
-          }]
-        ) else [],
-      ) +
-      deployment.mapContainers(
-        function(c) {
-          result::
-            c + c.withEnvMixin(
-              if params.gcpCredentialSecretName != "" then (
-                [{
-                  name: "GOOGLE_APPLICATION_CREDENTIALS",
-                  value: "/secret/gcp-credentials/key.json",
-                }]
-              )
-            ) +
-            c.withVolumeMountsMixin(
-              if params.gcpCredentialSecretName != "" then (
-                [{
-                  name: "gcp-credentials",
-                  mountPath: "/secret/gcp-credentials",
-                }]
-              )
-            ),
-        }.result,
-      ),
+    deployment.mixin.spec.template.spec.withVolumesMixin(
+      if params.gcpCredentialSecretName != "" then (
+        [{
+          name: "gcp-credentials",
+          secret: {
+            secretName: params.gcpCredentialSecretName,
+          },
+        }]
+      ) else [],
+    ) +
+    deployment.mapContainers(
+      function(c) {
+        result::
+          c + c.withEnvMixin(
+            if params.gcpCredentialSecretName != "" then (
+              [{
+                name: "GOOGLE_APPLICATION_CREDENTIALS",
+                value: "/secret/gcp-credentials/key.json",
+              }]
+            )
+          ) +
+          c.withVolumeMountsMixin(
+            if params.gcpCredentialSecretName != "" then (
+              [{
+                name: "gcp-credentials",
+                mountPath: "/secret/gcp-credentials",
+              }]
+            )
+          ),
+      }.result,
+    ),
   tbDeployment:: tbDeployment,
 
   all:: [
--- a/kubeflow/tensorboard/gcp.libsonnet
+++ b/kubeflow/tensorboard/gcp.libsonnet
@@ -2,11 +2,13 @@
   local k = import "k.libsonnet",
   local util = import "kubeflow/core/util.libsonnet",
   local deployment = k.apps.v1beta1.deployment,
+  // super resolves to any object which has a params field.
   local params = super.params,
 
+  // super resolves to any object which has a tbService field
   tbService:: super.tbService,
 
-  local tbDeployment =
+  tbDeployment::
     super.tbDeployment +
     deployment.mixin.spec.template.spec.withVolumesMixin(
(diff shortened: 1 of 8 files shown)


kkasravi:
added a servicePort instead of it being hardcoded to 9000, fixed an error on the targetPort and updated kubeflow/examples/prototypes/tensorboard.jsonnet to be similar to kubeflow/tensorboard/prototypes/{tensorboard-aws.jsonnet,tensorboard-gcp.jsonnet}. I tested that I can connect to the UI in kubeflow/examples/prototypes/tensorboard.jsonnet

jlewi:
Looks like you need to reformat to fix the tests.

jlewi:
Also, lets add a unittest before merging this

kkasravi:
@jlewi I've created a pattern that can also be applied to tf-serving that I think improves on how variations are done. If you look at kubeflow/tensorboard/gcp.libsonnet and kubeflow/tensorboard/prototypes/tensorboard-gcp.jsonnet you'll see how I add just what is needed in gcp and reference super from within it. This is a great feature of mixins, where super is resolved as part of adding gcp to tensorboard. It also makes it easy to do tests otherwise we would have to copy the definitions of tensorboard in kubeflow/tensorboard/prototypes/tensorboard-{gcp,aws} to kubeflow/core/tests/{tensorboard-aws_test.jsonnet, tensorboard-gcp_test.jsonnet}.

jlewi:
Looks like the test failure is

```
"    was \"logging.googleapis.com/kubernetes\" but must be one of \"\", \"none\", \"logging.googleapis.com\".","status":"INVALID_ARGUMENT","statusMessage":"Bad
```

It looks like this was added in #1772 it looks like that PR didn't actually trigger the presubmit; Will revert that PR.

jlewi:
I think this was fixed by
https://github.com/kubeflow/kubeflow/commit/7f59af3b2747ec01ff2639ae4c1999c503cf3780

Looks like you are 1 commit behind that so I think if you just rebase that will fix the issue.

kubeflow/tensorboard/gcp.libsonnet
```diff
@@ -0,0 +1,50 @@
+{
+  local k = import "k.libsonnet",
+  local util = import "kubeflow/core/util.libsonnet",
+  local deployment = k.apps.v1beta1.deployment,
+  local params = super.params,
+
+  tbService:: super.tbService,
+
+  local tbDeployment =
+    super.tbDeployment +
```
jlewi: Why do you need super.tbDeployment?

IIUC this gets resolved in tensorboard-gcp.jsonnet when we say

```
local instance = tensorboard.new(env, params) + gcp;
````

So super inside gcp gets resolved to tensorboard.new(env, params).

How is this better then defining mixins for tbDeployment and tfService in gcp.libsonnet

And then explictly adding those mixins to the deployment and service defined in tensorboard.libsonnet? (I think if we did it this way we wouldn't have to use the super keyword).

One thing I notice is that the individual parts e.g. tbDeployment and tbService end up being the sum of the part defined in tensorboard.libsonnet and the mixin defined in tensorboard-gcp.libsonnet because of the use of the super keyword (I presume).  

But why doesn't the field "all" end up having 4 items since the list is defined in tensorboard.libsonnet and tensorboard-gcp.libsonnet and we are adding them? Did you verify that it only has 2 items in the case of the gcp prototype?



I'm fine with what you have here (assuming we don't duplicate 4 items). I honestly don't know what the right style for ksonnet should be, so I'm happy iterating

Can you add some comments at least explaining where super gets resolved?

kkasravi: Thanks for the review @jlewi . I'll add notes on how super is resolved.

The improvements are 
1. we only need to define one 'new(env,params)' method (in tensorboard.libsonnet).
2. it's easier to test gcp, aws since these subtypes are defined in gcp.libsonnet, aws.libsonnet
3. any attribute/method defined in the subtype overrides the super attribute/method, including 'all' (i tested this to make sure only 2 where defined)
4. it's matches jsonnet best practices more than the old way of creating an instance of the base type inside the subtype. When referring to super.tbDeployment, it will resolve to any object that has a tbDeployment attribute.

jlewi:
/lgtm
/approve

Can you cherry pick this onto v0.3-branch please?

Issue #1745, opened 2018-10-11
TFMA Dependency


zlcnju:
Related to:  [1576](https://github.com/kubeflow/kubeflow/issues/1576)

>I think it's not elegant idea to just modify install_tfma to no. 
It  is easy to add TFMA_VERSION and pip install the Specified version when build image. And this not only benefits the people use tf<1.9 with tfma,  and will make it easy when tfma dependecy changes.

jlewi:
I agree any interest in picking this up?

zlcnju:
@jlewi i will pick it up

jlewi:
This has been fixed.

Issue #1751, opened 2018-10-11
Pin notebook image requirements


pdmack:
Related to #1746, we should consider pinning installs and requirements in the notebooks to avoid issues like #1721, #1726.

May also want to consider the [OSS plan for dependabot](https://github.com/marketplace/dependabot/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW4zNTQ=#pricing-and-setup) too.

PR #1759, opened 2018-10-12
add tfma version

zlcnju:
Related to [1754](https://github.com/kubeflow/kubeflow/issues/1745)
1.modify tfma version to 0.6.0 when tf < 1.9
2. add tfma_version config to control tfma version

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1759)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -123,7 +123,9 @@ RUN conda create -n py2 python=2 && \
     && \
     python -m ipykernel install && \
     # tensorflow-model-analysis is only supported for TF 1.6 and above
-    if [[ $TFMA_VERSION ]]; then \
+    # TODO temporarily remove tensorflow-model-analysis because of gpu problem 
+    # https://github.com/kubeflow/kubeflow/pull/1759
+    if [[ $TFMA_VERSION ]] && [[ $TF_PACKAGE  =~ "cpu" ]]; then \
       # if you find any dependency problems in the future
       # please visit https://github.com/tensorflow/model-analysis#compatible-versions
       # and fix the TFMA_VERSION with compatible version number in the config json
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -118,7 +118,6 @@ RUN conda create -n py2 python=2 && \
     ${TF_PACKAGE_PY_27} \
     # Tensorflow packages which only supports python 2
     tensorflow-transform \
-    tensorflow-serving-api \
     # ipykernel for python 2 jupyter notebook kernel
     && \
     python -m ipykernel install && \
@@ -136,6 +135,10 @@ RUN conda create -n py2 python=2 && \
       jupyter nbextension enable --py --system tensorflow_model_analysis ; \
     fi \
     && \
+    # TODO a quick fix for tensorflow_serving_api when install gpu
+    # https://github.com/tensorflow/serving/issues/1142
+    # now the dep in tensorflow can meet the require of tensorflow-serving-api but not ensure future
+    pip install --no-cache-dir --no-deps tensorflow-serving-api && \
     # Install jupyterlab-manager
     conda install --quiet --yes \
     # nodejs required for jupyterlab-manager
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -8,6 +8,7 @@ FROM $BASE_IMAGE
 ARG TF_PACKAGE=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.7.0-cp36-cp36m-linux_x86_64.whl
 ARG TF_PACKAGE_PY_27=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.7.0-cp27-none-linux_x86_64.whl
 ARG TFMA_VERSION
+ARG TFDV_VERSION
 
 USER root
 
@@ -127,7 +128,10 @@ RUN conda create -n py2 python=2 && \
       # please visit https://github.com/tensorflow/model-analysis#compatible-versions
       # and fix the TFMA_VERSION with compatible version number in the config json
       pip install --no-cache-dir tensorflow-model-analysis==$TFMA_VERSION && \
-      pip install --no-cache-dir tensorflow-data-validation && \
+      if [[ $TFDV_VERSION ]]; then \
+        pip install --no-cache-dir tensorflow-data-validation==$TFDV_VERSION ; \
+      fi \
+      && \
       # We use --system because if we don't the config is written to the home directory
       # and the changes are lost when we mount a PV over them.
       jupyter nbextension enable --py --system widgetsnbextension && \
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -5,10 +5,9 @@ ARG BASE_IMAGE=ubuntu:latest
 
 FROM $BASE_IMAGE
 
-ARG INSTALL_TFMA=yes
 ARG TF_PACKAGE=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.7.0-cp36-cp36m-linux_x86_64.whl
 ARG TF_PACKAGE_PY_27=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.7.0-cp27-none-linux_x86_64.whl
-ARG TFMA_VERSION=default
+ARG TFMA_VERSION
 
 USER root
 
@@ -124,16 +123,11 @@ RUN conda create -n py2 python=2 && \
     && \
     python -m ipykernel install && \
     # tensorflow-model-analysis is only supported for TF 1.6 and above
-    if [[ $INSTALL_TFMA == "yes" ]]; then \
+    if [[ $TFMA_VERSION ]]; then \
       # if you find any dependency problems in the future
       # please visit https://github.com/tensorflow/model-analysis#compatible-versions
       # and fix the TFMA_VERSION with compatible version number in the config json
-      if [[ $TFMA_VERSION == "default" ]]; then \
-        pip install --no-cache-dir tensorflow-model-analysis; \
-      else \
-        pip install --no-cache-dir tensorflow-model-analysis==$TFMA_VERSION; \
-      fi \
-      && \
+      pip install --no-cache-dir tensorflow-model-analysis==$TFMA_VERSION && \
       pip install --no-cache-dir tensorflow-data-validation && \
       # We use --system because if we don't the config is written to the home directory
       # and the changes are lost when we mount a PV over them.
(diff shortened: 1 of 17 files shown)


zlcnju:
I signed it!

zlcnju:
I signed it!

zlcnju:
I signed it!

zlcnju:
I signed it!

components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
```diff
@@ -140,10 +140,15 @@
           else
             "nvidia/cuda:9.0-cudnn7-devel-ubuntu16.04",
         local installTfma =
-          if tf_version < "1.9" then
+          if tf_version < "1.6" then
             "no"
           else
             "yes",
+        local tfmaVersion =
```
jlewi: Can we get rid of this? and "INSTALL_TFMA"? If TFMA_VERSION is specified then that should mean we should install TFMA and if it isn't specified then we shouldn't install TFMA.

components/tensorflow-notebook-image/Dockerfile
```diff
@@ -124,7 +125,15 @@ RUN conda create -n py2 python=2 && \
     python -m ipykernel install && \
     # tensorflow-model-analysis is only supported for TF 1.6 and above
     if [[ $INSTALL_TFMA == "yes" ]]; then \
-      pip install --no-cache-dir tensorflow-model-analysis && \
+      # if you find any dependency problems in the future
+      # please visit https://github.com/tensorflow/model-analysis#compatible-versions
+      # and fix the TFMA_VERSION with compatible version number in the config json
+      if [[ $TFMA_VERSION == "default" ]]; then \
```
jlewi: Can we simplify this 

1) Why do we need to support "default"
2) Get rid of INSTALL_TFMA and just rely on TFMA_VERSION.

components/tensorflow-notebook-image/Dockerfile
```diff
@@ -124,7 +125,15 @@ RUN conda create -n py2 python=2 && \
...
+        pip install --no-cache-dir tensorflow-model-analysis; \
+      else \
+        pip install --no-cache-dir tensorflow-model-analysis==$TFMA_VERSION; \
+      fi \
+      && \
       pip install --no-cache-dir tensorflow-data-validation && \
```
jlewi: Can we follow the same pattern for TF data-validation?

components/tensorflow-notebook-image/Dockerfile
zlcnju: 1. For example, if the tfma has a new version with for  tf==1.9 and we specified the tfma version, now we are possible to ignore the tfma update. But if we have the default, we will update the tfma as the pip update. And until the tfma does not support the version in the latest release, such as tf < 1,9, we specified its version.
2. I will combine this two and use TFMA_VERSION to 'disbale_install' when tf < 1.6. I haven't combine the two at first because i don't want to the TFMA_VERSION has the other meaning, it hurt the concise of code.

jlewi: I'm not sure we want to auto-update. That could break things. I think we are moving in the direction of pinning things.

See for example #1751 

I would suggest removing "default". If TFMA releases a new version for existing TF versions then we can explicitly upgrade.

components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
```diff
@@ -336,20 +331,21 @@
...
-            buildImageTemplate("1.7.0", "1-7-0", "cpu"),
-            buildImageTemplate("1.7.0", "1-7-0", "gpu"),
-            buildImageTemplate("1.8.0", "1-8-0", "cpu"),
-            buildImageTemplate("1.8.0", "1-8-0", "gpu"),
-            buildImageTemplate("1.9.0", "1-9-0", "cpu"),
-            buildImageTemplate("1.9.0", "1-9-0", "gpu"),
-            buildImageTemplate("1.10.1", "1-10-1", "cpu"),
-            buildImageTemplate("1.10.1", "1-10-1", "gpu"),
+            //TODO tfma_version should be read from config.json. And so as the tf_package ...
+            buildImageTemplate("1.4.1", "1-4-1", "cpu", "disable_install"),
```
jlewi: Why do we need to specify the TFMA version as arguments to buildImageTemplate? Why isn't it being ready from config.json as the TODO mentions?

jlewi: Can we just rewrite build_image.sh to take in the config file?
https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/build_image.sh

components/tensorflow-notebook-image/Dockerfile
```diff
@@ -123,8 +123,16 @@ RUN conda create -n py2 python=2 && \
     && \
     python -m ipykernel install && \
     # tensorflow-model-analysis is only supported for TF 1.6 and above
-    if [[ $INSTALL_TFMA == "yes" ]]; then \
-      pip install --no-cache-dir tensorflow-model-analysis && \
+    if [[ $TFMA_VERSION != "disable_install" ]]; then \
+      # if you find any dependency problems in the future
+      # please visit https://github.com/tensorflow/model-analysis#compatible-versions
+      # and fix the TFMA_VERSION with compatible version number in the config json
```
richardsliu: Extra character in TFMA_VERSION

components/tensorflow-notebook-image/Dockerfile
```diff
@@ -123,8 +123,16 @@ RUN conda create -n py2 python=2 && \
     && \
     python -m ipykernel install && \
     # tensorflow-model-analysis is only supported for TF 1.6 and above
-    if [[ $INSTALL_TFMA == "yes" ]]; then \
-      pip install --no-cache-dir tensorflow-model-analysis && \
+    if [[ $TFMA_VERSION != "disable_install" ]]; then \
```
richardsliu: How about just 'none'?

components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
richardsliu: +1

components/tensorflow-notebook-image/Dockerfile
zlcnju: I think 'disable_install' is clear, somebody may think the none as any version. I think 'disable_install' is a more elegant code style.

components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
zlcnju: @jlewi  I will try it. As a fact, i am not familiar with scripts language. But  it will not make too much difficulty for me i think.

components/tensorflow-notebook-image/Dockerfile
zlcnju: Ok, it's a better way if we have the auto way to modify the pinning version.  I will remove it.

components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
zlcnju: @jlewi le I have a problem, why don't we use the file build_image.py in the build_image.sh？ It will Eliminate duplication and make code clear. If there is no special thoughts, i will use the python install scripts in build_image.sh

components/tensorflow-notebook-image/Dockerfile
jlewi: I think using magic values makes this unnecessarily complicated.

We'd like builds to be repeatable. So pinning python packages to specific versions is the best approach.

So lets check if the value is specified and if its is that should be the version of the package to install.

We can use the empty string to indicate don't install this.

I'd really like to get this into the 0.3.1 release so lets make this changes and if we want to discuss whether that's the right solution lets do it in a follow on issue.

components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
jlewi: Lets leave it as a TODO.

jlewi: It would be good to fix build_image.sh and not duplicate it; but this is blocking 0.3.1 so lets just get a quick fix to allow us to build the images by adding the library versions for TFMA and TF.DV as extra arguments and then fix build_image.sh or build_image.py in a follow on PR.

jlewi:
Looks like there are some jsonnet errors;
You should be able to fix these with
./scripts/autoformat_jsonnet.sh

zlcnju:
Now, there are some errors:
```
ImportError: Traceback (most recent call last):
 File "/opt/conda/envs/py2/lib/python2.7/site-packages/tensorflow/python/pywrap_tensorflow.py", line 58, in <module>
   from tensorflow.python.pywrap_tensorflow_internal import *
 File "/opt/conda/envs/py2/lib/python2.7/site-packages/tensorflow/python/pywrap_tensorflow_internal.py", line 28, in <module>
   _pywrap_tensorflow_internal = swig_import_helper()
 File "/opt/conda/envs/py2/lib/python2.7/site-packages/tensorflow/python/pywrap_tensorflow_internal.py", line 24, in swig_import_helper
   _mod = imp.load_module('_pywrap_tensorflow_internal', fp, pathname, description)
ImportError: libcuda.so.1: cannot open shared object file: No such file or directory
Failed to load the native TensorFlow runtime.
```
It looks like loss file of libcuda.so.1. I guess it because the gpu envioronment is not ready and these images should be build by nvida-docker or map the device files into docker environment.

jlewi:
Is it possible you are trying to use a GPU version of TF with a non GPU base image?

We've been using this workflow to build GPU images successfully so I think its probably related to this PR.

jlewi:
Looks like something about installing the widgets ends up trying to import tensorflow which fails
because we can't import CUDA

```
Enabling notebook extension jupyter-js-widgets/extension...
      - Validating: OK
```

@richardsliu Any ideas? I guess one option would be to add a GPU node pool to the CI cluster and then build on GPU nodes.

/cc @pdmack

zlcnju:
@jlewi  Before, it could build successfully because it will not use tensorflow package int the build process. But now, 'jupyter nbextension enable --py --system widgetsnbextension' use it.  Another things strangely is that pip install --no-cache-dir tensorflow-serving-api, this will cause the reinstall of tensorflow when it turns to gpu. I met the same problem in the previous kubeflow version. So, i think it couldn't have any possibility to build gpu version recently.  [The related issue in tensorflow serving](https://github.com/tensorflow/serving/issues/1142), And i clear the dep install in this pr as a quick fix

jlewi:
Regarding the formatting error it seems like your script is getting the wrong repo name.
Are you running the script in your git directory for kubeflow/kubeflow

Do you have multiple remote repos defined e.g kubeflow/kubeflow and kubeflow/testing

What is the output of

```
git remote -v
```

You can see what the script is doing here
https://github.com/kubeflow/kubeflow/blob/master/scripts/autoformat_jsonnet.sh#L73

richardsliu:
@zlcnju The script assumes you have Git SSH setup: https://help.github.com/articles/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent/#generating-a-new-ssh-key

Afterwards you need to change your Git remote hosts to this format:
```
origin	git@github.com:zlcnju/kubeflow.git (fetch)
origin	git@github.com:zlcnju/kubeflow.git (push)
upstream	git@github.com:kubeflow/kubeflow.git (fetch)
upstream	git@github.com:kubeflow/kubeflow.git (push)
```

jlewi:
It looks like there are merge conflicts now.

I started #1815 based on this PR intending to try to fix the GPU builds.

I ended up having to resolve the merge conflicts.

So @zlcnju you have two options

1. You can resolve the mere conflicts and we can merge this PR
1. We can close this PR in favor #1815

zlcnju:
> It looks like there are merge conflicts now.
> 
> I started #1815 based on this PR intending to try to fix the GPU builds.
> 
> I ended up having to resolve the merge conflicts.
> 
> So @zlcnju you have two options
> 
> 1. You can resolve the mere conflicts and we can merge this PR
> 2. We can close this PR in favor #1815
OK, I think the second is bettler, since some todo lists is in.

jlewi:
#1815 is submitted. Closing this PR since that should fix it.

PR #1770, opened 2018-10-13
Fix TF data validation and TFMA install for 1.10

### LINKED ISSUE #1718: TF data validation and tfma are in 1.9 images but not 1.10
The imports

failed in 
gcr.io/kubeflow-images-public/tensorflow-1.10.1-notebook-cpu:v0.3.0

but succeeded in
gcr.io/kubeflow-images-public/tensorflow-1.9.0-notebook-cpu:v0.3.0

But we should be installing TFMA in 1.10 as well

richardsliu:
Fixes #1718

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1770)

### DIFF
--- a/components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
+++ b/components/tensorflow-notebook-image/releaser/components/workflows.libsonnet
@@ -19,6 +19,30 @@
       )
     else [],
 
+  // Function to parse and compare version strings. Assumes version strings are in the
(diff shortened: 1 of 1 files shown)


jlewi:
is #1759 a better fix? Specifically that PR moves in the direction of allowing us to specify different versions of the pip packages in different images.

richardsliu:
Sounds good.

PR #1773, opened 2018-10-14
Cleanup the OWNERs files.

jlewi:
* Move some top level approvers into subdirectory OWNERs files that better
  match their areas of expertise
* Remove some reviewers that haven't been that active.
* This will make blunderbuss more efficient.
* Data can be seen here
https://devstats.kubeflow.org/d/46/user-reviews-repository-groups?orgId=1&var-period=d7&var-repo_name=All&var-repo=all&var-reviewers=ScorpioCPH&var-reviewers=willb

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1773)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,7 +2,6 @@
 # list approvers for individual components (e.g. Seldon folks for Seldon component)
 approvers:
   - abhi-g
-  - gaocegege
   - jlewi
   - kunmingg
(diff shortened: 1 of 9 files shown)


pdmack:
/approve

Easy enough to add OWNERS as contributions merit.

Issue #1766, opened 2018-10-12
Enable TPU in GKE


dsdinter:
It would be great to be able to leverage TPUs in our Kubeflow cluster if deploying in GKE.
I can work on that feature as I enabled that option already in my local version.
Believe we just need to modify the below two files within deployment_manager_configs:
cluster-kubeflow.yaml -> Add boolean option "tpu: false"
cluster.jinja -> Add settings to enable TPU and IP Aliases if tpu: true

jlewi:
That would be great!

PR #1774, opened 2018-10-14
Enable TPU in GKE (#1766)

dsdinter:
Modify the below two files within deployment_manager_configs to enable TPU if using API v1beta1:
cluster-kubeflow.yaml -> Add boolean option "tpu: false"
cluster.jinja -> Add settings to enable TPU and IP Aliases if tpu: true and API v1beta1


jlewi:
Thanks!

It would be great if you could add docs to our website about enabling TPUs.

PR #1805, opened 2018-10-16
Add support for using host network for OpenMPI worker pods

hmizuma:
Some workloads are impacted by overhead of overlay network and I believe can benefit from running on the host network.
Allows passing in `useHostNetwork true` to configure worker pods to run on the host network.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1805)

### DIFF
--- a/kubeflow/openmpi/util.libsonnet
+++ b/kubeflow/openmpi/util.libsonnet
@@ -14,4 +14,27 @@
       [std.splitLimit(keyValue, "=", 1)[0]]: std.splitLimit(keyValue, "=", 1)[1]
       for keyValue in $.toArray(str)
     } else {},
+
+  // Convert a string to upper case.
+  upper:: function(x) {
+    local cp(c) = std.codepoint(c),
+    local upLetter(c) = if cp(c) >= 97 && cp(c) < 123 then
+      std.char(cp(c) - 32)
+    else c,
+    result:: std.join("", std.map(upLetter, std.stringChars(x))),
+  }.result,
+
+  // Convert non-boolean types like string,number to a boolean.
+  // This is primarily intended for dealing with parameters that should be booleans.
+  toBool:: function(x) {
+    result::
+      if std.type(x) == "boolean" then
+        x
+      else if std.type(x) == "string" then
+        $.upper(x) == "TRUE"
+      else if std.type(x) == "number" then
+        x != 0
+      else
+        false,
+  }.result,
 }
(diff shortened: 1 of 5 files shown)


jlewi:
Please sign the CLA

kubeflow/openmpi/util.libsonnet
```diff
@@ -14,4 +14,27 @@
       [std.splitLimit(keyValue, "=", 1)[0]]: std.splitLimit(keyValue, "=", 1)[1]
       for keyValue in $.toArray(str)
     } else {},
+
+  // Convert a string to upper case.
+  upper:: function(x) {
```
everpeace: I think jsonnet has `std.asciiUpper`.  Are there any specific reason to use custom method??

hmizuma: The reason for this was to be consistent with a number of other places where upper and toBool appear such as https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/util.libsonnet#L15-L22

hmizuma:
I signed the CLA

kubeflow/openmpi/util.libsonnet
jlewi: I think std.asciiUpper was added in a later version of jsonnet. We should get rid of our custom function once we a require a version of ksonnet with a sufficiently new version of the jsonnet library.

Please a TODO to that effect.

jlewi: It would be great to file an issue for that and reference the issue in the TODO

hmizuma:
Regarding performance, we were testing Horovod workloads on four p3.16xlarge instances on AWS. Performance over host network was equivalent to running the same work outside of Kubernetes but we saw about 30-60% performance hit when using various CNIs.
We saw good results on kubenet, but it is not available or convenient to configure in all environments. I'll see if I can provide more details and numbers.

kubeflow/openmpi/util.libsonnet
hmizuma: I created an issue https://github.com/kubeflow/kubeflow/issues/1826 and updated commit with corresponding TODO

everpeace:
> Performance over host network was equivalent to running the same work outside of Kubernetes but we saw about 30-60% performance hit when using various CNIs.  We saw good results on kubenet, but it is not available or convenient to configure in all environments. 

Wow, It's really interesting.

> I'll see if I can provide more details and numbers.

I'm looking forward to see the numbers.

everpeace:
/assign @pdmack

Would you mind if you took a look this pr and gave an approval?? 🙇 

I reviewed and approved this PR.  But, this contains a tiny change (just add an comment) in `kubeflow/core/util.libsonnet`.  So we need your approval.

Thanks in advance.

PR #1808, opened 2018-10-17
use tf serving 1.11

lluunn:
This should enable getModelStatus http api, and prometheus exporter.

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1808)


lluunn:
Yes. Let's just close for now

PR #1816, opened 2018-10-17
Add "services/proxy" resources to notebook role.

qimingj:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1816)


qimingj:
We don't need it since we can access the service via DNS without going through proxy.

PR #1817, opened 2018-10-17
add ssl cert reuse logic in e2e & prober tests

kunmingg:
Add ssl cert reuse logic in e2e & prober tests

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1817)

### DIFF
--- a/testing/test_deploy_app.py
+++ b/testing/test_deploy_app.py
@@ -3,8 +3,11 @@ import argparse
 import base64
 import logging
 import os
+import shutil
 from time import sleep
 from google.auth.transport.requests import Request
+from googleapiclient import discovery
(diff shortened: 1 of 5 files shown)


testing/test_deploy_app.py
```diff
@@ -63,13 +69,43 @@ def make_deploy_call(args):
...
+        continue
+    except Exception as e:
+      print("Deployment hasn't started")
+      continue
+    break
+
+  if os.path.exists(SSL_DIR):
+    shutil.rmtree(SSL_DIR)
+  os.mkdir(SSL_DIR)
+  print("donwload ssl cert and insert to GKE cluster")
```
jlewi: Can we use logging module?

PR #1820, opened 2018-10-18
fixes 'add the metacontroller component #1819'

### LINKED ISSUE #1819: add the metacontroller component
this is a prerequisite to #1769

kkasravi:
fixes #1819

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1820)


jlewi:
Would you mind filing an issue to remove meta controller from here and using this package instead?
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/core/cloud-endpoints.libsonnet

kubeflow/metacontroller/metacontroller.libsonnet
```diff
@@ -0,0 +1,186 @@
+{
+  local k8s = import "k8s.libsonnet",
+  local util = import "kubeflow/core/util.libsonnet",
+  local crd = k8s.apiextensions.v1beta1.customResourceDefinition,
+
+  new(_env, _params):: {
+    local params = _env + _params {
```
jlewi: nit: should we just get namespace from env and not params?

The only reason we ever got it from params was because the early version of ksonnet didn't support getting it from env.

I think this snippet just keeps getting copied around. I think we should try to stick with the pattern env maps to namespace.

I think if we start letting different components deploy in different namespaces its confusing to users so I think we should treat that as an exception.

jlewi:
Do we want to put this in its own package or should we create a package that would be a little bit more generic?

I don't have a strong oppinion either way.

We can always refactor later.

/lgtm
/approve

/hold
In case you want to make changes based on comment above. Otherwise just remove the hold

kkasravi:
@jlewi   #1824,  #1825

kkasravi:
i think having the metacontroller package is ok for now.

PR #1834, opened 2018-10-20
fixes 'metacontroller's compositecontrollers and decoratorcontrollers need to be at Cluster scope'

### LINKED ISSUE #1833: metacontroller's compositecontrollers and decoratorcontrollers need to be at Cluster scope
See metacontroller's [yaml](https://raw.githubusercontent.com/GoogleCloudPlatform/metacontroller/master/manifests/metacontroller.yaml)

kkasravi:
fixes #1833

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1834)

### DIFF
--- a/kubeflow/metacontroller/metacontroller.libsonnet
+++ b/kubeflow/metacontroller/metacontroller.libsonnet
@@ -1,7 +1,5 @@
 {
-  local k8s = import "k8s.libsonnet",
   local util = import "kubeflow/core/util.libsonnet",
-  local crd = k8s.apiextensions.v1beta1.customResourceDefinition,
 
   new(_env, _params):: {
     local params = _env + _params,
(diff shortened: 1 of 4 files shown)


jlewi:
So metacontroller is a cluster scoped resource? 

It looks like this PR is installing metacontroller by default. Do we already have a component that depends on metacontroller or is this in preparation for the new Jupyter CRD?

jlewi:
So metacontroller is cluster scoped?

It looks like you are installing metacontroller by default now with kfctl.sh. Do we already have something that is using metacontroller by default? Or is this in preparation for the Jupyter CRD.

kkasravi:
Yes, metacontroller in 0.2 is cluster scoped. 

We could wait until jupyter CRD is ready. Also we need to upgrade cloud-endpoints.

kkasravi:
Since this is in preparation for jupyterhub replacement which i'll open a PR this week and it doesn't conflict with cloud-endpoints (which isn't using any of metacontroller's 0.2 crds), I think this could go in now.

PR #1836, opened 2018-10-20
Add some latency metrics for GKE & KF deployments

abhi-g:
- Flow start time of request handling through the context
- create metrics around time to process the requests.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1836)


kunmingg:
Currently stackdriver dashboard force user to choose aligner (between delta or rate) when add a new metrics. This might mess up our histogram metrics.
@jlewi shall we set up standalone dashboard?

PR #1841, opened 2018-10-23
Separate logic for setup_backend.sh

### LINKED ISSUE #1145: Problems upgrading services of type NodePort; spec.clusterIP: Invalid value: "":
I notice that if I run ks apply on an existing component that has services I get errors like the following

```
ks apply default -c iap-ingress
ERROR handle object: merging object with existing state: Service "envoy" is invalid: [metadata.resourceVersion: Invalid value: "": must be specified for an update, spec.clusterIP: Invalid value: "": field is immutable] 
```
I get the same error if I do ks show followed by kubectl apply

```
ks show default -c iap-ingress | kubectl apply -f -
ingress.extensions "envoy-ingress" unchanged
certificate.certmanager.k8s.io "envoy-ingress-tls" configured
serviceaccount "envoy" unchanged
clusterrolebinding.rbac.authorization.k8s.io "envoy" configured
clusterrole.rbac.authorization.k8s.io "envoy" configured
deployment.extensions "envoy" configured
deployment.extensions "iap-enabler" unchanged
configmap "envoy-config" created
service "whoami-app" unchanged
deployment.extensions "whoami-app" unchanged
The Service "envoy" is invalid: 
* metadata.resourceVersion: Invalid value: "": must be specified for an update
* spec.clusterIP: Invalid value: "": field is immutable
```

Here is the output of ks show

```
apiVersion: v1
kind: Service
metadata:
  labels:
    service: envoy
  name: envoy
  namespace: kubeflow
spec:
  ports:
  - name: envoy
    port: 8080
    targetPort: 8080
  selector:
    service: envoy
  type: NodePort
```

My conjecture is that this issue is specific to services with NodePort and is the result of 0.11 switching to using merge patching to try to preserve node ports:
https://github.com/ksonnet/ksonnet/releases

I suspect this is not a ksonnet issue.

Related issues:
   #980 - ks apply complains about optimistic locking

r2d4:
"setup_backend.sh is doing two things

1. updating the backend service
2. updating the envoy config in each instance with the info needed for
JWT validation
"
This PR separates those two concerns into two separate scripts.

Fixes #1145

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1841)

### DIFF
--- a/kubeflow/core/iap.libsonnet
+++ b/kubeflow/core/iap.libsonnet
@@ -251,6 +251,87 @@
     },  // deploy
     deploy:: deploy,
 
+    // Run the process to update the backend service
+    local backendUpdater = {
+      apiVersion: "apps/v1",
+      kind: "StatefulSet",
+      metadata: {
+        name: "backend-updater",
+        namespace: params.namespace,
+        labels: {
+          service: "backend-updater",
+        },
+      },
+      spec: {
+        selector: {
+          matchLabels: {
+            service: "backend-updater",
+          },
+        },
+        template: {
+          metadata: {
+            labels: {
+              service: "backend-updater",
+            },
+          },
+          spec: {
+            serviceAccountName: "envoy",
+            containers: [
+              {
+                name: "backend-updater",
+                image: params.ingressSetupImage,
+                command: [
+                  "bash",
+                  "/var/envoy-config/update_backend.sh",
+                ],
+                env: [
+                  {
+                    name: "NAMESPACE",
+                    value: params.namespace,
+                  },
+                  {
+                    name: "SERVICE",
+                    value: "envoy",
+                  },
+                  {
+                    name: "GOOGLE_APPLICATION_CREDENTIALS",
+                    value: "/var/run/secrets/sa/admin-gcp-sa.json",
+                  },
+                ],
+                volumeMounts: [
+                  {
+                    mountPath: "/var/envoy-config/",
+                    name: "config-volume",
+                  },
+                  {
+                    name: "sa-key",
+                    readOnly: true,
+                    mountPath: "/var/run/secrets/sa",
+                  },
+                ],
+              },
+            ],
+            volumes: [
+              {
+                configMap: {
+                  name: "envoy-config",
+                },
+                name: "config-volume",
+              },
+              {
+                name: "sa-key",
+                secret: {
+                  secretName: "admin-gcp-sa",
+                },
+              },
+            ],
+          },
+        },
+      },
+
+    },  // backendUpdater
+    backendUpdater:: backendUpdater,
+
     // Run the process to enable iap
     local iapEnabler = {
       apiVersion: "extensions/v1beta1",
@@ -631,6 +712,7 @@
       data: {
         "envoy-config.json": std.manifestJson(envoyConfig(params)),
         "setup_backend.sh": importstr "setup_backend.sh",
+        "update_backend.sh": importstr "update_backend.sh",
         "configure_envoy_for_iap.sh": importstr "configure_envoy_for_iap.sh",
       },
     },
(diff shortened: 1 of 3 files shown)


kubeflow/core/iap.libsonnet
```diff
@@ -251,6 +251,78 @@
     },  // deploy
     deploy:: deploy,
 
+    // Run the process to update the backend service
+    local backendUpdater = {
```
jlewi: Should we run this as a StatefulSet  or Deployment?

It could be desirable; if we periodically rerun the commands to update the backend. This way if someone messes with the backend it will eventually be reset to the desired state.

kubeflow/core/pod.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: v1
```
jlewi: What's this for? Aren't use using the K8s job in iap.libsonnet to run update_backend.sh?

kubeflow/core/update_backend.sh
```diff
@@ -0,0 +1,38 @@
...
+[ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
+
+PROJECT=$(curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/project/project-id)
+if [ -z ${PROJECT} ]; then
+echo Error unable to fetch PROJECT from compute metadata
+exit 1
+fi
+
+# Activate the service account
+gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS}
```
jlewi: I think we need to add retries here:
https://github.com/kubeflow/kubeflow/blob/7a84feb3f0caf3bf7c282c31008073d90edceaf0/kubeflow/core/configure_envoy_for_iap.sh#L22

To allow for the fact that service account might not work immediately.

kubeflow/core/update_backend.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#
+# A simple shell script to configure the backend timeouts and health checks by using gcloud.
+[ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
```
jlewi: As noted in my comment about using a stateful set vs. K8s job,

should we add a while loop around this code and repeat the activation every hour or so?

scripts/util.sh
```diff
@@ -76,7 +76,7 @@ function createKsApp() {
   # ks component rm spartakus
   # Generate a random 30 bit number
   local usageId=$(((RANDOM<<15)|RANDOM))
-  ks generate spartakus spartakus --usageId=${usageId} --reportUsage=true
+  # ks generate spartakus spartakus --usageId=${usageId} --reportUsage=true
```
jlewi: Did you intend to comment this out?

jlewi:
Thanks @r2d4 

You should be able to follow these instructions to get the logs for the Argo step that failed in your most recent presubmit
https://github.com/kubeflow/testing#no-logs-in-argo-ui-or-pod-id-missing-in-argo-logs

However, the reason the logs don't show up in Argo is probably just a transient error; so its likely if you rerun the tests the Argo logs will be available and you can see the failure that way.

jlewi:
@r2d4 ping? Is this blocked on anything if not would be good to get it updated and submitted?

kubeflow/core/iap.libsonnet
r2d4: Updated as Statefulset

kubeflow/core/update_backend.sh
r2d4: I added a `sleep 3600` at the end of the script, after the container completes, k8s should spin up another pod

r2d4:
cc @jlewi this should be ready for review

PR #1843, opened 2018-10-23
create a self-serve environment for data-scientists

### LINKED ISSUE #1842: create a self-serve component for data-scientists
- [x] have all infra-components in a common namespace like 'kubeflow'
- [x] provide a CRD and controller so a datascientist can create one or more protected namespaces 
- [x] protected namespaces created by the data scientist allow 
  - [ ] jupyter notebooks to be run
  - [ ] tf-jobs to be run
  - [ ] tf-serving to be run 
  - [x] is secure such that only the data scientist can see it

kkasravi:
fixes #1842

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1843)


kubeflow/metacontroller/metacontroller.libsonnet
```diff
@@ -135,7 +135,8 @@
                   "-v=4",
                   "--discovery-interval=20s",
                 ],
-                image: "metacontroller/metacontroller:0.2.0",
+                #image: "metacontroller/metacontroller:latest",
```
jlewi: Delete commented out code.

Should the image be a parameter?

kubeflow/registry.yaml
```diff
@@ -4,6 +4,12 @@ libraries:
   application:
     version: master
     path: application
+  metacontroller:
```
jlewi: Didn't you already add metacontroller in another PR?

scripts/kfws
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
jlewi: What's this script for?

Why not just tell users to create the YAMl directly and do kubectl apply?

If we have a script that creates and applies a resource, then we hide the declarative management.
With kfctl.sh we deliberately split this into two steps generate and apply so that it was easy for users to customize.

kubeflow/projects/projects.libsonnet
```diff
@@ -0,0 +1,962 @@
...
+          {
+            port: 80,
+            targetPort: 8080,
+          },
+        ],
+      },
+    },
+    projectsService:: projectsService,
+
+    local syncProject =
```
jlewi: Can you add a comment explaining what this is?

kubeflow/projects/projects.libsonnet
```diff
@@ -0,0 +1,962 @@
...
+              std.objectHas(resource.metadata, 'name') && 
+              std.objectHas(request, 'parent') &&
+              std.objectHas(request.parent, 'spec') &&
+              std.objectHas(request.parent.spec, 'namespace') &&
+              resource.metadata.name == request.parent.spec.namespace then
+                true
+              else
+                false,
+          }.return,
+          //local foundChildren = std.filter(existingResource,
```
jlewi: Delete commented out code.

kubeflow/projects/projects.libsonnet
```diff
@@ -0,0 +1,962 @@
...
+      metadata: {
+        name: "workspaces-controller",
+      },
+      spec: {
+        generateSelector: true,
+        parentResource: {
+          apiVersion: "kubeflow.org/v1alpha1",
+          resource: "workspaces",
+        },
+        childResources: [
```
jlewi: Delete commented out code.

jlewi:
Thank you. As I mentioned yesterday I think this is a great leap forward in our multi-tenant story.

Please expand the PR description.

It looks like this PR is introducing 3 new resources

* Projects
* Workspaces
* Permissions

How are projects and workspaces different? Is there a 1:1 relationship between Projects or Workspaces and namespaces? 

What is permission? Is it a collection of resources?

IIUC metacontroller is implemented using a user supplied lambdas. Are the lambdas in this case implemented as jsonnet?

kkasravi:
this is closed in lieu of #1872

PR #1845, opened 2018-10-23
prober test, kubeflow testing part

kunmingg:
add prober test logic to test_deploy_app.py
which for loop:
1. clean up resource
2. make deploy call
3. check result

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1845)

### DIFF
--- a/testing/test_deploy_app.py
+++ b/testing/test_deploy_app.py
@@ -265,6 +265,8 @@ def prober_clean_up_resource(args):
 
   # Delete target-http-proxies
   delete_gcloud_resource(args, 'target-http-proxies')
+  # Delete target-http-proxies
+  delete_gcloud_resource(args, 'target-https-proxies')
   # Delete url-maps
   delete_gcloud_resource(args, 'url-maps')
   # Delete backend-services
--- a/testing/test_deploy_app.py
+++ b/testing/test_deploy_app.py
@@ -10,7 +10,7 @@ from time import sleep
 from google.auth.transport.requests import Request
 from googleapiclient import discovery
 from oauth2client.client import GoogleCredentials
-from prometheus_client import start_http_server, Gauge
+from prometheus_client import start_http_server, Gauge, Counter
 
 import requests
 import yaml
@@ -27,9 +27,11 @@ IAM_SCOPE = 'https://www.googleapis.com/auth/iam'
 OAUTH_TOKEN_URI = 'https://www.googleapis.com/oauth2/v4/token'
 METHOD = 'GET'
 SERVICE_HEALTH = Gauge('deployment_service_status',
-                              '0: normal; 1: deployment not successful; 2: service down')
+                       '0: normal; 1: deployment not successful; 2: service down')
 PROBER_HEALTH = Gauge('prober_health',
-                              '0: normal; 1: not working')
+                      '0: normal; 1: not working')
+SUCCESS_COUNT = Counter('deployment_success_count', 'accumulative count of successful deployment')
+FAILURE_COUNT = Counter('deployment_failure_count', 'accumulative count of failed deployment')
 
 def may_get_env_var(name):
   env_val = os.getenv(name)
@@ -111,9 +113,9 @@ def insert_ssl_cert(args):
   credentials = GoogleCredentials.get_application_default()
 
   service = discovery.build('deploymentmanager', 'v2', credentials=credentials)
-  retry_credit = 100
-  while retry_credit > 0:
-    retry_credit -= 1
+  # Wait up to 10 minutes till GKE cluster up and available.
+  end_time = datetime.datetime.now() + datetime.timedelta(minutes=10)
+  while datetime.datetime.now() < end_time:
     sleep(5)
     try:
       request = service.deployments().get(project=args.project, deployment=args.deployment)
@@ -131,6 +133,7 @@ def insert_ssl_cert(args):
   os.mkdir(SSL_DIR)
   logging.info("donwload ssl cert and insert to GKE cluster")
   try:
+    # TODO: switch to client lib
     util_run(("gsutil cp gs://%s/%s/* %s" % (SSL_BUCKET, args.mode, SSL_DIR)).split(' '))
   except Exception:
     logging.warning("ssl cert for %s doesn't exist in gcs" % args.mode)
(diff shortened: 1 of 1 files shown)


testing/test_deploy_app.py
```diff
@@ -23,7 +26,10 @@
 IAM_SCOPE = 'https://www.googleapis.com/auth/iam'
 OAUTH_TOKEN_URI = 'https://www.googleapis.com/oauth2/v4/token'
 METHOD = 'GET'
-logging.getLogger('googleapiclient.discovery_cache').setLevel(logging.ERROR)
+SERVICE_HEALTH = Gauge('deployment_service_status',
```
jlewi: The indentation of '0: normal..." looks off, Can you fix please?

testing/test_deploy_app.py
```diff
@@ -33,7 +39,7 @@ def may_get_env_var(name):
   else:
     raise Exception("%s not set" % name)
 
-def make_deploy_call(args):
+def prepare_request_data(args):
```
jlewi: Why does the prototype need to load gcp_prototype.yaml? Shouldn't that be handled by the backend?

testing/test_deploy_app.py
```diff
@@ -67,15 +71,40 @@ def make_deploy_call(args):
...
-    "SAClientId": "111670663612681935351",
+    "SAClientId": args.sa_client_id,
     "Token": access_token,
     "Zone": args.zone
   }
-  resp = requests.post("http://kubeflow-controller.%s.svc.cluster.local:8080/kfctl/e2eDeploy" % args.namespace, json=req_data)
+
+def make_e2e_call(args):
+  req_data = prepare_request_data(args)
+  resp = requests.post("http://kubeflow-controller.%s.svc.cluster.local:8080/kfctl/e2eDeploy" % args.namespace,
```
jlewi: Why are you contactig a locally running service? Shouldn't the prober be running in a different project and using the endpoint
deploy.kubeflow.cloud
or
deploy-staging.kubeflow.cloud

jlewi: Oh this is used for the E2E test.

testing/test_deploy_app.py
```diff
@@ -23,7 +26,10 @@
 IAM_SCOPE = 'https://www.googleapis.com/auth/iam'
 OAUTH_TOKEN_URI = 'https://www.googleapis.com/oauth2/v4/token'
 METHOD = 'GET'
-logging.getLogger('googleapiclient.discovery_cache').setLevel(logging.ERROR)
+SERVICE_HEALTH = Gauge('deployment_service_status',
+                              '0: normal; 1: deployment not successful; 2: service down')
+PROBER_HEALTH = Gauge('prober_health',
```
jlewi: Why gauges and not a counter?
What if we have N instances of the prober deploying in parallel?

If we have a counter for these values e.g. 1 counter for successes and 1 for failure then we can just increment those counters and compute rates in order to setup alerts.

testing/test_deploy_app.py
```diff
@@ -101,13 +130,52 @@ def insert_ssl_cert(args):
     shutil.rmtree(SSL_DIR)
   os.mkdir(SSL_DIR)
   logging.info("donwload ssl cert and insert to GKE cluster")
-  os.system("gsutil cp gs://%s/%s/* %s" % (SSL_BUCKET, args.cert_group, SSL_DIR))
-  os.system("gcloud container clusters get-credentials %s --zone %s --project %s" % (args.deployment, args.zone, args.project))
-  os.system("kubectl create -f %s" % SSL_DIR)
+  try:
+    util_run(("gsutil cp gs://%s/%s/* %s" % (SSL_BUCKET, args.mode, SSL_DIR)).split(' '))
```
jlewi: nit: Add a TODO to switch to using the GCS client instead of shelling out to gsutil.

testing/test_deploy_app.py
```diff
@@ -101,13 +130,52 @@ def insert_ssl_cert(args):
...
 def check_deploy_status(args):
   logging.info("check deployment status")
+  service_account_credentials = get_service_account_credentials("CLIENT_ID")
+
+  google_open_id_connect_token = get_google_open_id_connect_token(
+    service_account_credentials)
+  # Wait up to 30 minutes for IAP access test.
+  retry_credit = 180
+  status_code = 0
+  while retry_credit > 0:
```
jlewi: Why not just use the retrying module?

jlewi: If your doing time based retrying then I would suggest doing something like

end_time = datetime.datetime.now() + datetime.timedelta(minutes=30)
while datetime.datetime.now() < end_time:
...

This way if you change the sleep; you don't affect the total time we wait.

testing/test_deploy_app.py
```diff
@@ -179,6 +217,105 @@ def get_google_open_id_connect_token(service_account_credentials):
     request, OAUTH_TOKEN_URI, body)
   return token_response['id_token']
 
+def delete_gcloud_resource(args, keyword, filter='', dlt_params=[]):
+  get_cmd = 'gcloud compute %s list --project=%s --format="value(name)"' % (keyword, args.project)
```
jlewi: Add a TODO to replace this with use of the client libraries rather than shelling out to gcloud.

testing/test_deploy_app.py
```diff
@@ -179,6 +217,105 @@ def get_google_open_id_connect_token(service_account_credentials):
...
+  credentials = GoogleCredentials.get_application_default()
+  service = discovery.build('deploymentmanager', 'v2', credentials=credentials)
+  delete_done = False
+  try:
+    request = service.deployments().delete(project=args.project, deployment=args.deployment)
+    response = request.execute()
+  except Exception as e:
+    logging.info("Deployment doesn't exist, continue")
+  # wait up to 500 seconds till delete finish.
+  for i in range(50):
```
jlewi: Same comment as above.

while datetime.datetime.now() < endtime

testing/test_deploy_app.py
```diff
@@ -210,20 +347,20 @@ def main(unparsed_args=None):
...
-    "--client_id",
-    default="",
-    type=str,
-    help="oauth client id")
-  parser.add_argument(
-    "--client_secret",
-    default="",
-    type=str,
-    help="oauth client secret")
+    help="target url which accept deployment request")
```
jlewi: Delete commented out code.

testing/test_deploy_app.py
kunmingg: Want to use value to indicate error type.
If we have n instance, then
```any(signal != 0)``` should trigger alert.

kunmingg: Added SUCCESS_COUNT and FAILURE_COUNT

kunmingg:
Need to rebase for presubmit test

PR #1846, opened 2018-10-23
add gauge metrics for dashboard & alerting; add metrics for invalid a…

kunmingg:
* add gauge metrics for dashboard & alerting
* add metrics for invalid request
* make kubeflow version a dropdown list
* fix request latency metrics

(Gauge metrics have more aligner choices than counter)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1846)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


kunmingg:
/hold
testing

lluunn:
Can you describe more on why we need both gauge and counter in the PR description?
e.g. what's the problem of counter's aligner? what do we want to see on dashboard/alert

It's weird to use gauge type for a count.

kunmingg:
agree its a bit weird:
because of stackdriver aligner limitation:

count: aligner only has rate/delta
* good for some alert rules

gauge: aligner has mean/min/max/std.dev/....
* good for dashboard and some alert rules

abhi-g:
Instead of duplicating the counters can we simply replace them with the gauge counters if they serve our purpose. Having duplicates seems like an overkill.

PR #1849, opened 2018-10-24
make alert metrics count type & add service heartbeat

kunmingg:
1. make alert metrics count type
2. add service heartbeat

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1849)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -187,6 +187,28 @@ type ApplyRequest struct {
 }
 
 var (
+	// Counter metrics
+	deployReqCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "deploy_requests",
+		Help: "Number of requests for deployments",
+	})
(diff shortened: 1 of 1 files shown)


bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -1137,5 +1165,6 @@ func (s *ksServer) StartHttp(port int) {
 	// add an http handler for prometheus metrics
 	http.Handle("/metrics", promhttp.Handler())
 
+	go countHeartbeat()
```
r2d4: nit: you could probably inline this one as

```go
go func() {
   time.Sleep(10 * time.Second)
   serviceHeartbeat.Inc()
}()
```

abhi-g: Sorry, added the labels too soon...

r2d4: It works the same either way :)

PR #1851, opened 2018-10-25
Fix broken presubmit tests

richardsliu:
https://github.com/kubeflow/tf-operator/pull/849 changed the way we launch TFJob E2E tests. This PR changes the workflow definition to the new format.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1851)

### DIFF
--- a/testing/deploy_kubeflow.py
+++ b/testing/deploy_kubeflow.py
@@ -47,7 +47,7 @@ def deploy_kubeflow(test_case):
   """Deploy Kubeflow."""
   args = parse_args()
   test_dir = test_case.test_suite.test_dir
-  namespace = args.namespace
+  namespace = "default"
   api_client = deploy_utils.create_k8s_client()
   app_dir = deploy_utils.setup_kubeflow_ks_app(test_dir, namespace, args.github_token, api_client)
 
--- a/testing/deploy_kubeflow.py
+++ b/testing/deploy_kubeflow.py
@@ -47,7 +47,7 @@ def deploy_kubeflow(test_case):
   """Deploy Kubeflow."""
   args = parse_args()
   test_dir = test_case.test_suite.test_dir
-  namespace = "default"
+  namespace = args.namespace
   api_client = deploy_utils.create_k8s_client()
   app_dir = deploy_utils.setup_kubeflow_ks_app(test_dir, namespace, args.github_token, api_client)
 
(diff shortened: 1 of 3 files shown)


lluunn:
/lgtm
/approve 
thanks

testing/deploy_kubeflow.py
```diff
@@ -47,7 +47,7 @@ def deploy_kubeflow(test_case):
   """Deploy Kubeflow."""
   args = parse_args()
   test_dir = test_case.test_suite.test_dir
-  namespace = args.namespace
+  namespace = "default"
```
lluunn: why do we use default?

richardsliu: Namespace isn't being passed anywhere, so all the components are deployed in the default namespace. See the logs: http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-kubeflow-e2e-minikube-1851-379d825-4045-4620?tab=workflow&nodeId=kubeflow-presubmit-kubeflow-e2e-minikube-1851-379d825-4045-4620-2307009077&sidePanel=logs%3Akubeflow-presubmit-kubeflow-e2e-minikube-1851-379d825-4045-4620-2307009077%3Amain

lluunn: Isn't namespace being used at line 52 to setup the app?

richardsliu: Right, but all the components are deployed under the default namespace. On line 106 we look for the deployment, so if we pass the wrong namespace then it will fail. We can either use the default namespace, or figure out how to actually pass the namespace to each component.

richardsliu: Maybe the right fix is to change line 83 to "namespace"?

jlewi: I think its only minikube that is using deploy_kubeflow.py.

I would expect deploy_utils.setup_kubeflow_ks_app to set the namespace of the environment which should correctly set the namespace for each component.

So we shouldn't need to set namespace when we generate each component.

So removing namespace as an argument seems correct.

jlewi: Can we pass in a value of "default" for --namespace rather than ignoring the argument and hardcoding "default"

richardsliu:
Fixed namespace issue.
/assign @jlewi

PR #1852, opened 2018-10-25
use broadwell

### LINKED ISSUE #1840: For GKE deployment, change default CPU to Broadwell
For the GKE deployment, change default CPU to Broadwell instead of Haswell (which is no longer as widely available).
https://cloud.google.com/compute/docs/regions-zones/

cc @jlewi

lluunn:
fix #1840

/cc @kunmingg 
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1852)


lluunn:
presubmit failure should be fixed by: https://github.com/kubeflow/kubeflow/pull/1851

kunmingg:
Do you need to rebase?

PR #1855, opened 2018-10-25
fixes 'create a notebook controller that can replace jupyterhub and uses k8 native auth'

### LINKED ISSUE #1769: create a notebook controller that can replace jupyterhub and uses k8 native auth
using metacontroller create a controller which takes a CRD that spawns a jupyter notebook for a user. Vet the design first, and demo the implementation. 

See related discussion #1630

kkasravi:
fixes #1769

This PR will launch a jupyter notebook in a manner similar to jupyterhub but has a 
few noteworthy improvements:
- a Notebook CRD defines the API where the Notebook.spec.template.spec (PodTemplateSpec) replaces `c.KubeSpawner.<attributes>`
- no separate authentication layer is imposed. Uses k8 authentication
- will launch a jupyter notebook in a user's own namespace if paired with the profiles component

See the README.md for details.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1855)


jlewi:
Thanks @kkasravi ; really excited to see this.

nit please provide a more descriptive PR title and description. Its very helpful if we don't have to click through to the issue.

jlewi:
Should the files go in the newly created jupyter package as opposed to creating a new notebooks package?

jlewi:
A couple questions

1. Based on the README it sounds like the controller is spawning a pod. Why not use a statefulset or deployment so the pod is resilient? Main advantage of using a statefulset is that the name of the pod is deterministic?

1. What are the annotations that are set in the Notebook resource (see notebook_example.yaml) for? Why wouldn't user just directly set the relevant options on the pod? e.g. PVC mounts.

1. Can you explain the dependency on the Profile resource? Is it really a dependency? I would assume that as long as the user has permissions to create notebooks in namespace they should be able to create jupyter notebooks in that namespace? The Profile resource is just one way to create that?

kubeflow/notebooks/notebook.example.yaml
```diff
@@ -0,0 +1,86 @@
...
+              "start-singleuser.sh",
+              "--ip='0.0.0.0'",
+              "--port=8888",
+              "--allow-root"
+            ],
+            "env": [],
+            "image": "gcr.io/kubeflow-images-public/tensorflow-1.10.1-notebook-cpu:v0.3.0",
+            "imagePullPolicy": "IfNotPresent",
+            "name": "notebook",
+            "ports": [
```
jlzhao27: will the CRD create the correct service if there are multiple containers on different ports?

We have a use case where the user would like to run the main jupyter container to write tensorflow code but also launch a `tensorboard` container with a shared volume mount for visualizing their training progress.

kubeflow/notebooks/notebook.example.yaml
```diff
@@ -0,0 +1,86 @@
...
+      "useJupyterLabAsDefault": true,
+      "ttlSecondsAfterFinished": 300
+    },
+    "name": "notebook",
+    "namespace": "resnet50"
+  },
+  "spec": {
+    "template": {
+      "metadata": {
+        "name": "notebook",
```
jlzhao27: feels weird that the metadata is redefined here for `name` and `namespace`, what if the user set them to be different values?

Can the pod spec (or job/deployment spec) just inherity the `metadata.labels` and `metadata.annotations` from the CRD?

kubeflow/notebooks/prototypes/README.md
```diff
@@ -0,0 +1,131 @@
...
+  selector:
+    app: training
+  type: ClusterIP
+```
+
+Subsequent browser requests to `https://<api-server>/<namespace>/<notebook>` are routed to the Service and Pod as shown below:
+
+![Jupyter Notebook](./docs/jupyter_notebook.png "Jupyter Notebook")
+
+The notebook component depends on the [profiles component](https://github.com/kubeflow/kubeflow/tree/master/kubeflow/profiles/prototypes) to provide a protected namespace for the user. This means that a Profile yaml was posted to the kubernetes api-server prior to posting the Notebook yaml. In this case a Profile yaml would look like:
```
jlzhao27: similar to @jlewi's question, is this a strict dependency? If my k8s client has permissions to create/modify namespaces, the CRD should just work right?

kubeflow/notebooks/notebook.example.yaml
```diff
@@ -0,0 +1,86 @@
+{
+  "apiVersion": "kubeflow.org/v1alpha1",
+  "kind": "Notebook",
+  "metadata": {
+    "annotations": {
+      "accessLocalFs": "false",
+      "disks": "null",
+      "image": "tensorflow-1.10.1-notebook-cpu:v0.3.0",
```
jlzhao27: why are all these necessary?

kubeflow/notebooks/prototypes/README.md
```diff
@@ -0,0 +1,131 @@
+## Goals
+
+- provide a k8 native mechanism to spawning jupyter notebooks for users
+
+- spawn a notebook within a RBAC protected namespace for a user
+
+- set a serviceAccountName within the spawning pod similar to jupyter-notebook
```
jlzhao27: nit: point (2), (3), (4) doesn't really seem related to this CRD. I feel like the namespace RBAC, service account attachment, and k8s auth is really external to this PR. 

The same CRD should also work on a k8s cluster if I use a `cluster-admin` role to create a `namespace` and launch it right?

This CRD feels like it is more wrapping the logic for handling networking and maybe populating jupyter specific default command, args, and env vars in pod templates.

kkasravi:
> nit please provide a more descriptive PR title and description. Its very helpful if we don't have to click through to the issue.

Done

kkasravi:
> Should the files go in the newly created jupyter package as opposed to creating a new notebooks package?

We could, though i would consider waiting until the UI is integrated and feature parity is there (plus the additional advantages of k8 authn/authz,  using the PodTemplateSpec API instead of KubeSpawner, etc. I agree that this should be the end goal.

kkasravi:
> 1. Based on the README it sounds like the controller is spawning a pod. Why not use a statefulset or deployment so the pod is resilient? Main advantage of using a statefulset is that the name of the pod is deterministic?

I've considered using a Job or Deployment. The path routing passed to ambassador in the service is `https://<api-server>/<namespace>/<notebookname>`. We do need a way of reaping terminated or idle notebooks using k8 mechanisms.

kubeflow/notebooks/prototypes/README.md
```diff
@@ -0,0 +1,131 @@
...
+        workingDir: "/home/jovyan"
+      ttlSecondsAfterFinished: 300
+      securityContext: 
+      - fsGroup: 100
+        runAsUser: 1000
+```
+
+### User Interaction
+
+The user submits a Notebook yaml either through a UI or CLI (eg `kubectl apply -f notebook.yaml`) and the Notebook yaml is handled by the notebook-controller. The controller will create a Service and Pod within the user's namespace. The Service uses ambassador to create a reverse proxy that will route subsequent browser requests to the Pod. An example Service is shown below:
```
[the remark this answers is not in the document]
kkasravi: yes, the controller could be changed to do something along these lines. Since our  common API Gateway is ambassador it replaces the (redundant) configurable-http-proxy in jupyterhub. In general we've wanted to move routing, authn up to the API gateway from these types of services.

kubeflow/notebooks/prototypes/README.md
```diff
@@ -0,0 +1,131 @@
...
+The Notebooks component is an alternative to jupyterhub. The component defines a Notebook CRD and provides a notebook-controller (based on metacontroller's [CompositeController](https://metacontroller.app/api/compositecontroller/)). The Notebook CRD schema is similar to a kubernetes [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#creating-a-deployment) where it defines a [Pod Template](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template) that is used by the notebook-controller to create a Pod. An example Notebook yaml is shown below:
+
+```yaml
+apiVersion: kubeflow.org/v1alpha1
+kind: Notebook
+metadata:
+  name: training
+  namespace: resnet50
+spec:
+  template:
```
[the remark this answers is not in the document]
kkasravi: yes, i'll make this change. I was leaning towards Job. Thanks.

kubeflow/notebooks/notebook.example.yaml
kkasravi: Let me think about that. The notebook crd could define something similar to an Ingress's rules or allow the user to supply ambassador annotations in the notebook crd which the controller then uses when creating the service/pod/containers.

kubeflow/notebooks/notebook.example.yaml
kkasravi: yes, i've removed notebook.example.yaml. The controller does use the notebooks metadata when creating the pod/job/deployment.

kubeflow/notebooks/prototypes/README.md
kkasravi: yes, correct - there shouldn't be any explicit dependency on profiles. i'll review any assumptions I've made in this regard. I should also add a test case for this.

kubeflow/notebooks/notebook.example.yaml
kkasravi: please ignore. They were hold-overs when I was fairly early in to implementation.

kubeflow/notebooks/prototypes/README.md
kkasravi: Thanks Jason, this is a nice summary and should be exactly what the notebooks component provides. I'll revisit the documentation and any assumptions I've made in the code.

kubeflow/notebooks/prototypes/README.md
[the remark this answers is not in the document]
jlewi: I suspect @jlzhao27's question was whether using Ambassador makes the CRD overly Kubeflow specific and whether we should generalize that by somehow letting customers use ingress.

I think one of the motivations for the CRD is to hide the creation of additional resources (e.g. the routes) needed to setup the notebook. So for this reason I think it makes sense to add the Ambasador annotation to the services.

If a user isn't using Ambassador then the annotations will simply be null op.

FWIW, I think we are moving in the direction of using ISTIO to dynamically manage routes. One of the nice things about this CRD is that we should be able to hide that change from other services (e.g. the notebook web app).

Once we switch to ISTIO all services would be behind the ISTIO gateway. So ideally the only thing that admins would have to figure out how to expose is the ISTIO gateway which could be done a number of ways (e.g. VPN or IAP).

kubeflow/notebooks/prototypes/README.md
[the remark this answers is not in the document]
jlewi: Why Job? The semantics of Job are run until completion. I think of Jupyter more as a server; so the semantics of a statefulset or deployment seem more appropriate. I lean towards statefulset just to give the pods predictable names.

kubeflow/notebooks/notebook.example.yaml
jlewi: @jlzhao27 What's the reasoning for running tensorboard in the same pod as Jupyter vs. a separate pod and using a shared filesystem to allow communication between the two? Is the advantage just eliminating the need to have a shared filesystem.

Can we follow the same model for TB as for Jupyter? i.e. having a simple web app to manage the TB instances just like we have for Jupyter?

jlewi:
Thanks @kkasravi.

I think there are only two remaining issues

1. The type of resource used for the notebooks. You mentioned using a Job; but I lean towards statefulset WDYT?

1. Package location; You mentioned using notebooks now and then moving to the newly created jupyter package once there is feature parity.
   
    * Why not just put it there now and avoid future churn?

kubeflow/notebooks/prototypes/README.md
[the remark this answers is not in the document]
kkasravi: Sure, I'll replace with a StatefulSet. 

I was thinking of ways to eliminate the long wait for notebook to come up - due to the cost of pulling the image. Are there ways to eliminate that time cost under a StatefulSet?

kkasravi:
> Thanks @kkasravi.
> 
> I think there are only two remaining issues
> 
> 1. The type of resource used for the notebooks. You mentioned using a Job; but I lean towards statefulset WDYT?
> 2. Package location; You mentioned using notebooks now and then moving to the newly created jupyter package once there is feature parity.
>    
>    * Why not just put it there now and avoid future churn?

Makes sense, however I'll need to add an httpserver to present the ui (similar to jupyterhub)

kubeflow/notebooks/prototypes/README.md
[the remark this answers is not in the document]
jlzhao27: Cool this makes sense, will istio be relying on ingresses or just raw service annotations?

Main reason I'm advocating for ingress as well is the current set of controllers we run internally both operate on ingress annotations instead of service. Since the semantic of k8s services is cluster internal communication where as ingress is cluster external. 

 ([nginx-ingress](https://github.com/kubernetes/ingress-nginx), [heptio/contour](https://github.com/heptio/contour)) both

kubeflow/notebooks/prototypes/README.md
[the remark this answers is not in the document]
jlzhao27: nit: might want to use `deployment` since stateful sets give you some weird guarantees. Namely:

```
 Deleting and/or scaling a StatefulSet down will not delete the volumes associated with the StatefulSet. This is done to ensure data safety, which is generally more valuable than an automatic purge of all related StatefulSet resources.
```
which means if a user specifies a PVC in a notebook, creating/deleting the notebook CRD will never remove that volume...which could be argued as intended behavior? But as it stands right now, the volume management is configured outside of the CRD which I like for its simplicitly and breaking down ownership of various components.

```
StatefulSets currently require a Headless Service to be responsible for the network identity of the Pods. You are responsible for creating this Service
```
Not really an issue, more of a nit to be careful around when creating the service.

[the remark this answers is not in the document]
jlzhao27: The image pulling cost is independent of the controller type so I dont think deployment/ss matters. What we do is we run a `daemonset` with the fat notebook images as init-containers and the daemonset itself just runs the pause container. This essentially creates an image pre-fetch layer and caches it on all the kubelets. You could add a template of that here if you want to simplify people's lives.

kubeflow/notebooks/notebook.example.yaml
jlzhao27: The main problem is configuring the shared filesystem. It is trivial to write a pod spec with a common emptyDir volume. Otherwise, you will need some kind of multi-attach PVC, host mounted paths, or external storage like s3. Do you have any tips on a simple k8s native shared fs?

All of those options are possible but just increases the complexity for the user iterating on the notebook itself. I suppose if there is a hosted tensorboard service that reads from a common data store, that could be a design that works. Given the flexibility of the pod spec today, we decided to just roll that into one and not worry about a hosted tensorboard service.

jlzhao27:
hey @kkasravi, this is the metacontroller docs right? https://metacontroller.app/

Might be worthwhile to call out in the README that the only requirements are `kubectl` access and an installed `metacontroller`, I think you mention it in the design part but it is a runtime requirement also.

kkasravi:
> hey @kkasravi, this is the metacontroller docs right? https://metacontroller.app/
> 
> Might be worthwhile to call out in the README that the only requirements are `kubectl` access and an installed `metacontroller`, I think you mention it in the design part but it is a runtime requirement also.

Sure, I'll update the README.

kkasravi:
> > Makes sense, however I'll need to add an httpserver to present the ui (similar to jupyterhub)
> 
> I'm not sure I understood this comment. What does this have to do with the ksonnet package organization of the new Jupyter CRD?

My misunderstanding I think ...

Were you saying that I should do I `git mv kubeflow/notebooks/* kubeflow/jupyter` and have 2 prototypes under kubeflow/jupyter/prototypes/ or completely replace the current kubeflow/jupyter with what's under kubeflow/notebooks?

jlewi:
> Were you saying that I should do I git mv kubeflow/notebooks/* kubeflow/jupyter and have 2 prototypes under kubeflow/jupyter/prototypes/ or completely replace the current kubeflow/jupyter with what's under kubeflow/notebooks?

I was thinking multiple prototypes in the kubeflow/jupyter package. Is there any downside to doing this? I was thinking this would just avoid additional churn later on as I think kubeflw/jupyter is the package where we'd like them to live long term.

kkasravi:
> > Were you saying that I should do I git mv kubeflow/notebooks/* kubeflow/jupyter and have 2 prototypes under kubeflow/jupyter/prototypes/ or completely replace the current kubeflow/jupyter with what's under kubeflow/notebooks?
> 
> I was thinking multiple prototypes in the kubeflow/jupyter package. Is there any downside to doing this? I was thinking this would just avoid additional churn later on as I think kubeflw/jupyter is the package where we'd like them to live long term.

I agree. No downside since they both in theory could be deployed. I'll try and get this out asap.

jlewi:
Thanks @kkasravi !

jlewi:
@kkasravi Think we can get this merged this week?

kkasravi:
yes, I'll push changes today which should be most of what's required.

kubeflow/notebooks/prototypes/README.md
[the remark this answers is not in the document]
kkasravi: i'm not sure what istio will rely on. It's possible after the switch we would need to replace the service with an ingress - or make it configurable.

jlewi:
Thanks @kkasravi

2 comments but I think these are best handled in follow on discussion and PRs as necessary.

1. Why deployment not statefulset?

The main advantage of a statefulset is that the pod name is stable. This is mostly useful from a debugging perspective; i.e. we can tell users to look for a pod with a predictable name as opposed to filtering by labels or some other mechanism.

Since the type of resource is internal to the CRD we could always fix it later.

1. Other question is about naming. I think notebook might refer to the ipynb file. Whereas the server is jupyter. So having the resource named "notebooks.kubeflow.org" might be confusing. It might lead users to think that the resource is tied to a single notebook.

The text on the website leaves me a bit confused; http://jupyter.org
They say "Jupyter Notebook".

/lgtm
/approve

PR #1862, opened 2018-10-25
make zone a dropdown list which has GPU available

kunmingg:


kunmingg:
Now UI looks like:

screen shot 2018-10-25 at 1 20 10 pm

abhi-g:
Will the deployment from the webapp continue to ensure there is a GPU pool? And this dropdown will ensure that the user selects a zone that has GPUs?

kunmingg:
@abhi-g 
The dropdown list only include zones with GPU available.

lluunn:
Do all the zones listed have broadwell?

Issue #1694, opened 2018-10-02
Argo UI doesn't work behind ambassador


jlewi:
Using: argoproj/argoui:v2.1.1

We currently add a reverse proxy route to Ambassador for the Argo UI but when I tried it I just get a blank screen. Using port-forward to the Argo pod works.

I see some javascript errors:

```
GET http://localhost:8080/main.ac6e091ac4a22318d76f.js net::ERR_ABORTED 404 (Not Found)
localhost/:1 Refused to execute script from 'http://localhost:8080/main.ac6e091ac4a22318d76f.js' because its MIME type ('text/plain') is not executable, and strict MIME type checking is enabled.
```
   * localhost:8080 is mapped via port-forwarding to Ambassador port 80

This looks like it might an issue with Argo not being proxy aware e.g. the URL

```
http://localhost:8080/main.ac6e091ac4a22318d76f.js 
```
Looks wrong; I think it should be

```
http://localhost:8080/argo/main.ac6e091ac4a22318d76f.js 
```

I found this issue on using a reverse proxy with Argo.
https://github.com/argoproj/argo/issues/716

Looks like we might need to explicitly set some parameter to control the base URL.

/cc @alexmt

TimZaman:
Still an issue? I installed kubeflow last week with the tutorial, and visiting `https://XXX.cloud.goog/argo/` gives me (console):

```GET https://kubeflow2.endpoints.kubeflowtest-225221.cloud.goog/main.f6e517525a3ff95bf2c8.js net::ERR_ABORTED 404```

PR #1865, opened 2018-10-25
Add base href for correct link in Argo UI

andreyvelich:
Fixes https://github.com/kubeflow/kubeflow/issues/1694

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1865)

### DIFF
--- a/kubeflow/argo/argo.libsonnet
+++ b/kubeflow/argo/argo.libsonnet
@@ -157,6 +157,10 @@
                     name: "IN_CLUSTER",
                     value: "true",
                   },
+                  {
+                    name: "BASE_HREF",
+                    value: "/argo/"
+                  },
                 ],
                 image: params.uiImage,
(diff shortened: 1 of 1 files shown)


Jeffwan:
@andreyvelich  just like to confirm this PR will only work behind Ambassador?  If I install argo individually, this will not work.

andreyvelich:
@Jeffwan Yes, that's true. Right now Argo UI only supports work behind Ambassador. 
f you want to use Argo UI individually you have to remove this variable: 
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/argo/argo.libsonnet#L161
and remove this annotation from the service: 
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/argo/argo.libsonnet#L201

Jeffwan:
Cool. Thanks for the confirmation.  @andreyvelich

PR #1872, opened 2018-10-26
create a self-serve component for data-scientists

### LINKED ISSUE #1842: create a self-serve component for data-scientists

kkasravi:
fixes #1842 

See the README.md in this PR under kubeflow/profiles/prototypes

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1872)


jlewi:
Why jsonnet for implementing the controller? Do you think we will want to stick with jsonnet long term, or switch to a real programming language?

kkasravi:
> I think I liked the name project/workspace better than profile. Is profile chosen to match K8s nomenclature (e.g. security profiles)?

Yes. The Security Profile in kubernetes is alluding to [Protection Profile](https://en.wikipedia.org/wiki/Protection_Profile) and the Target name alludes to [Security Target](https://en.wikipedia.org/wiki/Security_Target). I think using similar terminology is useful and makes this less about a user and more about what protection the user would like have (profile) and what environment the user operates in (target).

kkasravi:
> Can you explain what the Target and Permissions CRDs are doing? Why isn't the profile controller sufficient?

If Profile is like a Deployment, then Target is like a Pod, where the template (and repeated fields) are similar. This use case is pretty much captured in 2 proposals - the [Namespace Initializer](https://github.com/kubernetes/community/blob/cd160fd19b2ede1e11ae3406e6d0e693b8a470ce/keps/draft-20180806-namespace-initializer.md) and the [Namespace Population](https://github.com/kubernetes/community/pull/2177). Both are part of [Security Design Proposal](https://github.com/kubernetes/community/pull/2052). Take a look at [Automatic Policy Population](https://github.com/kubernetes/community/blob/cd160fd19b2ede1e11ae3406e6d0e693b8a470ce/keps/draft-20180806-namespace-initializer.md#automatic-policy-population) which describes this scenario. The reason this is split across 3 controllers is that metacontroller doesn't allow a CompositeController to create a namespace and then populate it with policy objects. 

This could be collapsed into a more declarative syntax captured in the above proposals.

kkasravi:
@jlewi 

An option would be to move more information of the type of 'Protection Profile' up into the Profile. Currently Role, RoleBinding information is not carried across the CRDs, rather the permissions-controller creates a [Role](https://github.com/kubeflow/kubeflow/pull/1872/files#diff-85de45c7549811721912c907ed0d22ccR9), RoleBinding in the target namespace w/o using the CR's.
This would make the implementation more similar to the Namespace Population proposal where Resource Quota, Role, RoleBinding are defined in the CRD [itself](https://github.com/kubernetes/community/blob/a8cb2060dc621664c86b185c7426367994b181b5/keps/draft-20180418-security-profile.md#namespace-population)

jlewi:
> Would we add client-side go (spf12/cobra) hooks in arena? If kfctl is also transitioning to golang, we wouldn't want to mix admin and user subcommands would we?

Why we would we need to add client side go code anywhere? Can't users just create their resource by doing

```
kubectl apply -f myprofile.yaml
```

(Or using ksonnet if they prefer)?

kubeflow/profiles/prototypes/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Each resource has an associated controller that is implemented in jsonnet using metacontroller's CompositeController.
+These are:
+
+sync-profile.libsonnet
+sync-target.libsonnet
+sync-permission.libsonnet
+
+## Client Integration
```
jlewi: Why do we need client integration?

kubeflow/profiles/prototypes/README.md
```diff
@@ -0,0 +1,97 @@
+## Overview
+
+### Goals
```
jlewi: Can you update the README.md to make it clear what an admin would do vs. what each user would do to create their own profile?

I'm a little confused because the SecurityProfile doc 
https://github.com/kubernetes/community/blob/a8cb2060dc621664c86b185c7426367994b181b5/keps/draft-20180418-security-profile.md

Makes it seem like you define a SecurityProfile and then you create resources according that profile.

So if we are following that design, I would expected there to be a SecurityProfile that would be defined by the admin and then individual resources (e.g. namespaces) would be created with that Profile.

In the SecurityProfile doc for example, the Profile seems to describe the security policy for a cluster and then multiple clusters are created with that profile.

Is the reason we have to define a different SecurityProfile for each user, is because each SecurityProfile needs to correspond to a different set of users to grant access to?

kubeflow/profiles/prototypes/README.md
```diff
@@ -0,0 +1,97 @@
...
+spec:
+  template:
+    metadata:
+      name: gan
+    spec:
+      namespace: gan
+      owner: alice
+```
+
+The Target resource is created by the controller using the information in the Profile Resource. The Target resource is created within the shared namespace. The target contains a template where the name of the namespace and the permissions are specified. An example is:
```
jlewi: So does the user create the Target or is the Target automatically created after the user creates the profile?

jlewi:
After reading the code more closely. It looks to me like user creates a Profile and then that creates the Target which creates the namespace and permission. The permission then creates the RBAC role and binding.

> The reason this is split across 3 controllers is that metacontroller doesn't allow a CompositeController to create a namespace and then populate it with policy objects.

Could we get by with 2 controllers rather than 3?

IIUC it seems like the most obvious thing to do would be to have a single CR (Profile) in which user would specify all the user supplied information (e.g. namespace name, users to grant access to, quota limits eventually), the controller would then create all the resources (namespace, rbac role & binding).

That won't work though per your comment about creating a namespace and resources within the namespace. 

So can we get by with 2 CRs?

kubeflow/profiles/prototypes/README.md
kkasravi: There are rolebindings created for each user in the new namespace. 
A user can have multiple namespaces (by submitting multiple Profiles), 
but each namespace contains a Role, Rolebinding.
All namespaces reference the user's ServiceAccount in kubeflow.

So if fred submits 2 profiles: resnet50, inceptionv3, there would 2 namespaces, each one having a {Role, RoleBinding}.
```
      apiVersion: "rbac.authorization.k8s.io/v1",
      kind: "RoleBinding",
      metadata: {
        name: "fred",
        namespace: "resnet50",
      },
      roleRef: {
        kind: "Role",
        name: "edit",
      },
      subjects: [{
        kind: "ServiceAccount",
        name: "fred",
        namespace: "kubeflow",
      }
```
```
      apiVersion: "rbac.authorization.k8s.io/v1",
      kind: "RoleBinding",
      metadata: {
        name: "fred",
        namespace: "inceptionv3",
      },
      roleRef: {
        kind: "Role",
        name: "edit",
      },
      subjects: [{
        kind: "ServiceAccount",
        name: "fred",
        namespace: "kubeflow",
      }
```

The 'edit' Role is the same for all namespaces created.
The RoleBinding just differs by the user's name.

The RoleBinding could be parameterized the type of Role

kubeflow/profiles/prototypes/README.md
kkasravi: The latter, the controller watches for Profile and then creates a Target.
```
- profiles-controller 
  - watches for Profile
  - creates Target
```
```
- targets-controller
  - watches for Target
  - creates Namespace, Permission
```
```
- permissions-controller
  - watches for Permission
  - creates Role, RoleBinding
```

kkasravi:
> After reading the code more closely. It looks to me like user creates a Profile and then that creates the Target which creates the namespace and permission. The permission then creates the RBAC role and binding.
> 
> > The reason this is split across 3 controllers is that metacontroller doesn't allow a CompositeController to create a namespace and then populate it with policy objects.
> 
> Could we get by with 2 controllers rather than 3?
> 
> IIUC it seems like the most obvious thing to do would be to have a single CR (Profile) in which user would specify all the user supplied information (e.g. namespace name, users to grant access to, quota limits eventually), the controller would then create all the resources (namespace, rbac role & binding).
> 
> That won't work though per your comment about creating a namespace and resources within the namespace.
> 
> So can we get by with 2 CRs?

Yes, I think we can. It would also make it more like the 'Namespace Population' proposal. Let me refactor. Thanks!

kubeflow/profiles/prototypes/README.md
kkasravi: > Can you update the README.md to make it clear what an admin would do vs. what each user would do to create their own profile?

Yes, will do

jlewi:
Thanks. Could we also add an option to the Profile to use a user account/IAM rather than ServiceAccount?

I'd like to be able to create a profile for user   bob@acme.com and just have the RBacBinding defined in terms of that user so that if I'm running kubectl on my laptop, I can already access K8s resources using bob@acme.com.

I think I'd still like the profile to define / create an appropriate service account. This would be used as a robot account by resources (TFJobs, Argo Workflows, etc...) that user bob might create.

e.g. User Bob might create an Argo workflow that used ServiceAccount "bobs-robot" so that the running ArgoWorkflow 

I think it would be reasonable to filing issues for some of this and doing it in follow on work.

kkasravi:
> Thanks. Could we also add an option to the Profile to use a user account/IAM rather than ServiceAccount?
> 
> I'd like to be able to create a profile for user [bob@acme.com](mailto:bob@acme.com) and just have the RBacBinding defined in terms of that user so that if I'm running kubectl on my laptop, I can already access K8s resources using [bob@acme.com](mailto:bob@acme.com).
> 
> I think I'd still like the profile to define / create an appropriate service account. This would be used as a robot account by resources (TFJobs, Argo Workflows, etc...) that user bob might create.
> 
> e.g. User Bob might create an Argo workflow that used ServiceAccount "bobs-robot" so that the running ArgoWorkflow
> 
> I think it would be reasonable to filing issues for some of this and doing it in follow on work.

Will incorporate. Thanks.

kkasravi:
@jlewi Ready for review.
- [x] Updated the README.md
- [x] removed Target 
- [x] Changed Profile's owner to be a __subject__ type which can either be
```yaml
owner: {
  kind: ServiceAccount
  name: joe
  namespace: kubeflow
}
```
or
```yaml
owner: {
  kind: User
  name: joe
}
```

and I verified that this works when I submitted a Profile that looks like

```yaml
kind: Profile
metadata:
  name: tfoob
  namespace: kubeflow
spec:
  template:
    metadata:
      namespace: tfoob
    spec:
      owner:
        kind: User
        name: kam.d.kasravi@intel.com
```

and one that looks like

```yaml
kind: Profile
metadata:
  name: cats
  namespace: kubeflow
spec:
  template:
    metadata:
      namespace: cats
    spec:
      owner:
        kind: ServiceAccount
        name: dean
        namespace: kubeflow
```

The schema for subject is almost the same as the swagger definition for 10.1:
```json
                          owner: {
                            type: "object",
                            required: [
                              "kind",
                              "name",
                            ],
                            properties: {
                              apiGroup: {
                                type: "string",
                              },
                              kind: {
                                enum: [
                                  "ServiceAccount",
                                  "User",
                                ],
                              },
                              namespace: {
                                type: "string",
                              },
                              name: {
                                type: "string",
                              },
                            },
                          }
```

jlewi:
Woo Hoo.

/lgtm
/approve

PR #1873, opened 2018-10-26
Fixing the conflict with the getting started guide

connected-bsamadi:
According to the getting started guide, the KUBEFLOW_REPO is already made before running this script.

* The script no longer uses the env ${KUBEFLOW_REPO} so there's no reason to check whether
   that directory exists.

* The script now downloads the source to the current directory.

* The check if directory ${KUBEFLOW_REPO} exists is a legacy of a previous version of the script

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1873)

### DIFF
--- a/scripts/download.sh
+++ b/scripts/download.sh
@@ -12,11 +12,6 @@ fi
 
 KUBEFLOW_TAG=${KUBEFLOW_TAG:-master}
 
-if [ -d "${KUBEFLOW_REPO}" ]; then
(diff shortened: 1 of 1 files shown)


connected-bsamadi:
I signed it!

jlewi:
Thanks for the fix!

/ok-to-test
/lgtm
/approve

PR #1887, opened 2018-10-30
Provide an error message in case GCP ZONE is not set and exit, also fix styles for kfctl.sh

### LINKED ISSUE #1697: [GCP] kfctl should provide a helpful error message if ZONE isn't set
If zone isn't set then when you run kfctl.sh init you get an error like the following

```
++ gcloud --project=myproject container get-server-config --zone=
++ head -1
++ grep '^1.10.[0-9]*[-d]gke.[0-9]*$'
ERROR: (gcloud.container.get-server-config) One of [--zone, --region] must be supplied: Please specify location..
+ CLUSTER_VERSION=
+ [[ '' == '' ]]
+ echo 'Setting cluster version to 1.10.7-gke.2'
```

We should provide a more helpful error message.

ashahba:
Fixes #1697

### DIFF
--- a/scripts/kfctl.sh
+++ b/scripts/kfctl.sh
@@ -53,8 +53,8 @@ createEnv() {
       echo KUBEFLOW_PLATFORM=gke >> ${ENV_FILE}
       echo PROJECT="${PROJECT}" >> ${ENV_FILE}
       if [ -z "${PROJECT}" ]; then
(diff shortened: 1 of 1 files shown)


ashahba:
We should not enforce users to use `us-east1-d` which we only need for testing.
However during test runs we already provide `--zone=us-east1-d`, so we are ok to delete it I guess.

PR #1888, opened 2018-10-30
Provide an error message in case GCP ZONE is not set and exit, also fix styles for kfctl.sh

### LINKED ISSUE #1697: [GCP] kfctl should provide a helpful error message if ZONE isn't set

ashahba:
fixes #1697

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1888)


ashahba:
We should not enforce users to use us-east1-d which we only need for testing.
However during test runs we already provide --zone=us-east1-d, so we are ok to delete it I guess.

scripts/kfctl.sh
```diff
@@ -137,6 +140,10 @@ if [ "${COMMAND}" == "init" ]; then
             shift
             PROJECT=$1
             ;;
+        --zone)
```
jlewi: Can we have only one way to explicitly set zone? e.g. if we set it as a command line argument then we should not set it using an environment variable.

ashahba: @jlewi Sure, but we are already doing same practice for `--platform`, `PLATFORM`, `--project`, `PROJECT`, `--email` and `EMAIL` so I thought that's kind of expected pattern here 🙂 

Let me know what you think. But if we are taking one out in the favor of other I would choose the `--zone` over environment variable.

Let me know your thoughts.

jlewi: I'm in favor of moving things to flags.

scripts/kfctl.sh
```diff
@@ -49,13 +49,64 @@ createEnv() {
       echo KUBEFLOW_DOCKER_REGISTRY=registry.aliyuncs.com >> ${ENV_FILE}
       ;;
     gcp)
-      PROJECT=${PROJECT:-$(gcloud config get-value project 2>/dev/null)}
+      while [[ $# -gt 0 ]]; do
```
jlewi: Thanks for doing this.

Do you think we could support the syntax?
--name value
as well as
--name=value

That could be a separate PR.

ashahba: We can @jlewi , I just have to put more logic to cover both scenarios 🙂 
Feel free to file the ticket and describe it the way we want it to be done and assign it to me and I'll pick it up after this one.
But first let's get the tests fixed🤞

scripts/kfctl.sh
```diff
@@ -121,78 +158,66 @@ createNamespace() {
...
-	  fi
-	fi
+      -h | --help)
+        usage
+        exit
+        ;;
+      --platform)
+        shift
+        PLATFORM=$1
+        mkdir -p ${DEPLOYMENT_NAME}
```
jlewi: Why did you move the mkdir and other commands here?

ashahba: Because this `DEPLOYMENT_NAME` is really only used by `gcp` specific `init` calls and Also I needed to be inside that directory to make `createEnv $*` call and then later parse args.
So `mkdir` needs to be called either here or within `createEnv` function. 🙂

jlewi: I"m still not following.

Should the instructions

mkdir -p ${DEPLOYMENT_NAME}
cd ${DEPLOYMENT}
createEnv

be executed for all platforms not just GCP?

The directory ${DEPLOYMENT_NAME} is used as the top level directory for all the Kubeflow related config (e.g. env.sh) so I don't think its GCP specific.

.gitignore
```diff
@@ -37,6 +37,8 @@ examples/.ipynb_checkpoints/
 
 **/.ipynb_checkpoints
 
+# env.sh generated by kfct.sh
```
jlewi: kfctl.sh

Missing .sh

scripts/kfctl.sh
jlewi: I see you're passing the command line arguments back to createEnv and then reparsing them there to persist them.

Can we just set a dictionary of environment variables here and then pass that to createEnv? That seems a little cleaner then parsing command line arguments in two places. We can also avoid putting logic in the argument parsing case statements.

ashahba: You mean do the whole arg parsing business here, and then pass them to `createEnv` with the parsed values?

We can but do we want that in this PR or a subsequent PR.

I was thinking of doing one PR to just go over all the shell scripts and enhance path detection, format and style, enhance user messages specially during failures and that sort of thing.

Let me know what you think.

jlewi: Yes that's what I meant. Specifically; I'd like to do flag parsing in a single location. Otherwise we will likely have problems later on because we update one location but not the other.

I'm ok with doing this in a follow on PR as long as you think you can to it in the next week or so. I'll LGTM this with a hold. If you want to submit this first and fix command line parsing later just remove the hold.

ashahba: Thanks @jlewi 
I'll do the next PR sooner than later this week.
I'll file a ticket and go from there.

PR #1893, opened 2018-10-30
Dm conf

kunmingg:
1. output deployment manger config to source repo
2. restruct source repo: 
* ks app dir: ```version/deploy_name/ks_app```
* deployment manger config dir: ```version/deploy_name/gcp_config```
3. UI update: 
* print timestamp of estimate ready time 
* check input format
* print user account when permission error occurs
* update instruction text to match new layout
4. security enhancement
5. give more retry buffer to GCP APIs 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1893)


bootstrap/cmd/bootstrap/app/gcpUtils.go
```diff
@@ -58,36 +58,34 @@ func init() {
...
 	err := LoadConfig(path.Join(regPath, "../deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml"), &dmconf)
 
 	if err == nil {
 		dmconf.Resources[0].Name = req.Name
 		dmconf.Resources[0].Properties["zone"] = req.Zone
-		dmconf.Resources[0].Properties["clientId"] = req.ClientId
-		dmconf.Resources[0].Properties["clientSecret"] = req.ClientSecret
 		dmconf.Resources[0].Properties["ipName"] = req.IpName
-		dmconf.Resources[0].Properties["isWebapp"] = true
+		dmconf.Resources[0].Properties["gpu-pool-max-nodes"] = 10
```
jlewi: Lets default this to 0 for now. What we should do is file a follow on issue to automatically check whether the user has GPU quota in the zone and if they do set it.

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -310,8 +329,11 @@ func (s *ksServer) CreateApp(ctx context.Context, request CreateRequest) error {
 	} else {
 		log.Infof("Creating app %v", request.Name)
 		log.Infof("Using K8s host %v", config.Host)
-
-		appDir := path.Join(s.appsDir, GetRepoName(request.Project), request.Name)
+		deployConfDir := path.Join(s.appsDir, GetRepoName(request.Project), kfVersion, request.Name)
```
jlewi: What is s.appsDir? Is that a directory in the volume where things are stored?

kunmingg: Yes, that's where we checkout source repos

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -652,16 +682,20 @@ func GetRepoName(project string) string {
...
 // Clone project repo to local disk, which contains all existing ks apps config in the project
-func (s *ksServer) CloneRepoToLocal(project string, token string) error {
-	repoDir := path.Join(s.appsDir, GetRepoName(project))
-	_, err := s.fs.Stat(repoDir)
-
-	// delete existing local repo, pull from cloud
-	if err == nil {
-		os.RemoveAll(repoDir)
-	}
+func (s *ksServer) CloneRepoToLocal(project string, token string) (string, error) {
```
jlewi: Why isn't this thread safe?

kunmingg: The comment is not accurate: it's only not thread-safe when its part of createApp.
createApp need to write to source repo, concurrent push might cause conflict on git side.

kunmingg: So createApp action in general is not thread safe when multi user using same project;
createApp currently protected by a project-lock

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -45,15 +48,18 @@ const CACHED_REGISTRIES = "/opt/versioned_registries"
 // key used for storing start time of a request to deploy in the request contexts
 const START_TIME = "startTime"
 
+const KUBEFLOW_REG_NAME = "kubeflow"
+const DM_FOLDER  = "deploymentmanager"
```
jlewi: Can we use the same folder layout that kfctl produces?

So we should have the directories:
gcp_config
k8s_specs
ks_app/

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -310,8 +334,11 @@ func (s *ksServer) CreateApp(ctx context.Context, request CreateRequest) error {
 	} else {
 		log.Infof("Creating app %v", request.Name)
 		log.Infof("Using K8s host %v", config.Host)
-
-		appDir := path.Join(s.appsDir, GetRepoName(request.Project), request.Name)
+		deployConfDir := path.Join(s.appsDir, repoDir, GetRepoName(request.Project), kfVersion, request.Name)
```
jlewi: Can we do a defer delete here?

kunmingg: Now should be covered by https://github.com/kubeflow/kubeflow/pull/1893/files#diff-98496810c3874df46e02c51893aa04c1R318?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -652,15 +682,22 @@ func GetRepoName(project string) string {
...
 // Clone project repo to local disk, which contains all existing ks apps config in the project
-func (s *ksServer) CloneRepoToLocal(project string, token string) error {
-	repoDir := path.Join(s.appsDir, GetRepoName(project))
-	_, err := s.fs.Stat(repoDir)
-
-	// delete existing local repo, pull from cloud
-	if err == nil {
-		os.RemoveAll(repoDir)
+func (s *ksServer) CloneRepoToLocal(project string, token string) (string, error) {
+	folderName := generateRandStr(20)
```
jlewi: Just add a comment here explaining why this is random.

jlewi:
@kunmingg Just a addd a comment explaining why we are using a random directory and then I think this PR is good.

As I mentioned in chat, I think we'll want to refactor the code more to better isolate the handler for each request and also to support CLI mode; but we can do that in a follow on PR.

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -997,20 +1053,28 @@ func makeDeployEndpoint(svc KsService) endpoint.Endpoint {
...
-			r.Err = err.Error()
-			deploymentFailure.Inc()
-			return r, err
+		retry := 0
+		for {
+			retry += 1
+			err := svc.BindRole(ctx, req.Project, req.Token, dmServiceAccount)
+			if err != nil {
+				if retry >= 5 {
+					r.Err = err.Error()
```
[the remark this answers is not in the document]
r2d4: You can see an example of a retry util here
https://github.com/kubernetes/minikube/blob/2621bb3a275c3c5b4f5be2fffa29263b7cec9734/pkg/util/utils.go#L93-L114

bootstrap/cmd/bootstrap/app/saKey.go
```diff
@@ -11,37 +11,94 @@ import (
...
+	"encoding/base64"
 )
 
-type InsertSaKeyRequest struct {
-	Cluster   string
-	Namespace string
-	Project   string
-	Token     string
-	Zone      string
+const OATHU_SECRET_NAME = "kubeflow-oauth"
```
r2d4: nit: typo

bootstrap/cmd/bootstrap/app/server.go
```diff
@@ -269,7 +269,7 @@ func processFile(opt *options.ServerOption, ksServer *ksServer) error {
 		Namespace:     opt.NameSpace,
 		AutoConfigure: true,
 	}
-	if err := ksServer.CreateApp(ctx, request); err != nil {
+	if err := ksServer.CreateApp(ctx, request, nil); err != nil {
```
r2d4: it's a little odd to pass in nil here

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -674,47 +714,68 @@ func (s *ksServer) CloneRepoToLocal(project string, token string) error {
...
 	cmd := exec.Command("sh", "-c", cloneCmd)
 	result, err := cmd.CombinedOutput()
 	log.Infof(string(result))
-	return err
+	return repoDir, err
 }
 
-func (s *ksServer) GetApp(project string, appName string, token string) (*appInfo, error) {
-	err := s.CloneRepoToLocal(project, token)
+func (s *ksServer) GetApp(project string, appName string, kfVersion string, token string) (*appInfo, string, error) {
```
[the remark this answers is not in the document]
kunmingg: appInfo is defined in dependency lib, better keep it as is?

bootstrap/cmd/bootstrap/app/server.go
kunmingg: This API is in maintain mode, would deprecate soon.

jlewi:
LGTM

I'll let Matt actually LGTM though since he has additional comments.

PR #1899, opened 2018-11-01
add retry to runCmd

kunmingg:
As source repo API has a chance of refusing push

### DIFF
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -507,12 +507,16 @@ func (s *ksServer) getRegistryUri(registry *RegistryConfig) (string, error) {
 }
 
 func runCmd(rawcmd string) error {
-	cmd := exec.Command("sh", "-c", rawcmd)
-	err := cmd.Run()
-	if err != nil {
-		log.Errorf("Error occrued during execute cmd %v. Error: %v", rawcmd, err)
+	for retry := 0; ; retry++ {
+		cmd := exec.Command("sh", "-c", rawcmd)
+		if err := cmd.Run(); err == nil || retry > 5 {
+			if err != nil {
+				log.Errorf("Error occrued during execute cmd %v. Error: %v", rawcmd, err)
+			}
+			return err
+		}
+		time.Sleep(5 * time.Second)
 	}
-	return err
 }
 
 // appGenerate installs packages and creates components.
@@ -794,6 +798,7 @@ func (s *ksServer) UpdateDmConfig(repoDir string, project string, appName string
 	return nil
 }
 
+
 // Save ks app config local changes to project source repo.
 // Not thread safe, be aware when call it.
 func (s *ksServer) SaveAppToRepo(project string, email string, repoDir string) error {
@@ -802,9 +807,16 @@ func (s *ksServer) SaveAppToRepo(project string, email string, repoDir string) e
 	if err != nil {
 		return err
 	}
-
-	return runCmd(fmt.Sprintf("git config user.email '%s'; git config user.name 'auto-commit'; git add .; "+
-		"git commit -m 'auto commit from deployment'; git pull --rebase; git push origin master", email))
+	if err = runCmd(fmt.Sprintf("git config user.email '%s'; git config user.name 'auto-commit'", email)); err != nil {
+		return err
+	}
+	if err = runCmd("git add .; git commit -m 'auto commit from deployment'"); err != nil {
+		return err
+	}
+	if err = runCmd("git pull --rebase; git push origin master"); err != nil {
+		return err
+	}
+	return nil
 }
 
 // Apply runs apply on a ksonnet application.


bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -507,12 +507,17 @@ func (s *ksServer) getRegistryUri(registry *RegistryConfig) (string, error) {
 }
 
 func runCmd(rawcmd string) error {
-	cmd := exec.Command("sh", "-c", rawcmd)
-	err := cmd.Run()
-	if err != nil {
-		log.Errorf("Error occrued during execute cmd %v. Error: %v", rawcmd, err)
+	for retry := 0; ; retry++ {
```
r2d4: why not `for retry := 0; retry < 6; retry++` like below?

jlewi: Should we use a library like: https://github.com/cenkalti/backoff to handle retries?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -507,12 +507,17 @@ func (s *ksServer) getRegistryUri(registry *RegistryConfig) (string, error) {
...
+		cmd := exec.Command("sh", "-c", rawcmd)
+		result, err := cmd.CombinedOutput()
+		if err == nil || retry > 5 {
```
r2d4: Maybe something like?

```golang
var errStr string
for retry := 0; retry < 6; retry++ {
  if err != nil {
    // log error
    errStr += err
    continue
  }
  // success case
  return result
}
// all retries have been used up
return nil, fmt.Errrof("retries exceeded: %s", errStr)
```

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -507,12 +507,17 @@ func (s *ksServer) getRegistryUri(registry *RegistryConfig) (string, error) {
...
+			if err != nil {
+				return fmt.Errorf("Error occrued during execute cmd %v. Error: %v", rawcmd, string(result))
+			}
+			return err
```
r2d4: This is a little confusing since it will always be nil here

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -507,12 +507,17 @@ func (s *ksServer) getRegistryUri(registry *RegistryConfig) (string, error) {
...
-	cmd := exec.Command("sh", "-c", rawcmd)
-	err := cmd.Run()
-	if err != nil {
-		log.Errorf("Error occrued during execute cmd %v. Error: %v", rawcmd, err)
+	for retry := 0; ; retry++ {
+		cmd := exec.Command("sh", "-c", rawcmd)
+		result, err := cmd.CombinedOutput()
+		if err == nil || retry > 5 {
+			if err != nil {
+				return fmt.Errorf("Error occrued during execute cmd %v. Error: %v", rawcmd, string(result))
```
r2d4: this will only give you the last error. you might want to collect the error strings for each retry?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -802,9 +809,16 @@ func (s *ksServer) SaveAppToRepo(project string, email string, repoDir string) e
...
-
-	return runCmd(fmt.Sprintf("git config user.email '%s'; git config user.name 'auto-commit'; git add .; "+
-		"git commit -m 'auto commit from deployment'; git pull --rebase; git push origin master", email))
+	if err = runCmd(fmt.Sprintf("git config user.email '%s'; git config user.name 'auto-commit'", email)); err != nil {
+		return err
+	}
+	if err = runCmd("git add .; git commit -m 'auto commit from deployment'"); err != nil {
+		return err
+	}
+	if err = runCmd("git pull --rebase; git push origin master"); err != nil {
```
r2d4: I'm not really sure what/where we're pushing here, but if we're doing this do we want to do a force push?

bootstrap/cmd/bootstrap/app/ksServer.go
kunmingg: Sounds good

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -802,9 +809,16 @@ func (s *ksServer) SaveAppToRepo(project string, email string, repoDir string) e
 	if err != nil {
 		return err
 	}
-
-	return runCmd(fmt.Sprintf("git config user.email '%s'; git config user.name 'auto-commit'; git add .; "+
-		"git commit -m 'auto commit from deployment'; git pull --rebase; git push origin master", email))
+	if err = runCmd(fmt.Sprintf("git config user.email '%s'; git config user.name 'auto-commit'", email)); err != nil {
```
r2d4: Another way to do this would be

```golang
	cmds := []string{
		fmt.Sprintf("git config user.email %s", email),
		"git config user.name auto-commit",
		"git add .",
		`git commit "commit name"`,
		"git pull --rebase",
		"git push -f origin master",
	}
	for _, cmd := range cmds {
		if err := runCmd(cmd); err != nil {
			return err
		}
	}
```

Then you can drop the `sh -c` from the runCmd

bootstrap/cmd/bootstrap/app/ksServer.go
kunmingg: This is a success return, we use nil to indicate it

bootstrap/cmd/bootstrap/app/ksServer.go
kunmingg: We are use cloud source repo.
force push might not be a good idea, just in case user have changes pushed around same time.

abhi-g:
I have a meta comment regarding this. So please feel free to ignore it specifically for this PR. But in general, I am a bit worried about all these retry logic with regards to robustness of the deployment app. Hardcoding trial numbers, sleep seconds generally makes me worried. Even if this PR goes in with those constants, we should think about how to make it more robust.

jlewi:
@abhi-g Are you suggesting a different backoff and retry policy? e.g. exponential backoff within some finite time window? Or are you suggesting that the fact that we depend on any type of retry mechanism is problematic.

kunmingg:
/hold 
till pass review

PR #1901, opened 2018-11-03
Gkeversion

kunmingg:
Set gke version to be ```1.10.6```, as current ```1.10``` will end up using ```1.10.9-gke.0```, which seems not stable (not on release notes yet). 
Also make GKE version a configurable param.

### DIFF
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -83,6 +83,10 @@ type ksServer struct {
 	// other information about the regisry.
 	knownRegistries map[string]RegistryConfig
 
+	//GKE version to use, suggest format 1.X.Y, avoid using most recent version
+	// https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters
+	gkeVersion string
+
 	fs afero.Fs
 
 	// project-id -> project lock
@@ -113,7 +117,7 @@ func (m MultiError) ToError() error {
 }
 
 // NewServer constructs a ksServer.
-func NewServer(appsDir string, registries []RegistryConfig) (*ksServer, error) {
+func NewServer(appsDir string, registries []RegistryConfig, gkeVersion string) (*ksServer, error) {
 	if appsDir == "" {
 		return nil, fmt.Errorf("appsDir can't be empty")
 	}
@@ -122,6 +126,7 @@ func NewServer(appsDir string, registries []RegistryConfig) (*ksServer, error) {
 		appsDir:         appsDir,
 		projectLocks:    make(map[string]*sync.Mutex),
 		knownRegistries: make(map[string]RegistryConfig),
+		gkeVersion:		 gkeVersion,
 		fs:              afero.NewOsFs(),
 	}
 
--- a/bootstrap/cmd/bootstrap/app/ksServer.go
+++ b/bootstrap/cmd/bootstrap/app/ksServer.go
@@ -126,7 +126,7 @@ func NewServer(appsDir string, registries []RegistryConfig, gkeVersion string) (
 		appsDir:         appsDir,
 		projectLocks:    make(map[string]*sync.Mutex),
 		knownRegistries: make(map[string]RegistryConfig),
-		gkeVersion:		 gkeVersion,
+		gkeVersion:      gkeVersion,
 		fs:              afero.NewOsFs(),
 	}
 
--- a/bootstrap/cmd/bootstrap/app/options/options.go
+++ b/bootstrap/cmd/bootstrap/app/options/options.go
@@ -29,7 +29,7 @@ type ServerOption struct {
 	AppDir               string
 	Config               string
 	Email                string
-	GkeVersion			 string
+	GkeVersion           string
 	NameSpace            string
 	RegistriesConfigFile string
 }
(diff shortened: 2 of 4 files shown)


jlewi:
What is the problem with 1.10.9 you are seeing?

If we pin to 1.10.6 won't we have problems caused by GKE's frequent rollout/deprecation of patch versions?
It looks like new patch versions are rolled out weekly.
https://cloud.google.com/kubernetes-engine/versioning-and-upgrades

Can we set the default to 1.10 but then provide a field that allows people to override it if they want to?

bootstrap/cmd/bootstrap/app/options/options.go
```diff
@@ -47,6 +48,7 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&s.JsonLogFormat, "json-log-format", true, "Set true to use json style log format. Set false to use plaintext style log format")
 	fs.IntVar(&s.Port, "port", 8080, "The port to use when running an http server.")
 	fs.StringVar(&s.AppDir, "app-dir", "/opt/bootstrap", "The directory for the ksonnet applications.")
+	fs.StringVar(&s.GkeVersion, "GkeVersion", "1.10.6", "Initial GKE master version")
```
jlewi: Why is GkeVersion being set as a flag? Doesn't this mean default is now being set in multiple places; e.g. the DM config and the flag variable?

If we want to give users a way to override it can we surface it in the UI?

kunmingg: With flag we can edit the default GKE version without changing image.
We can surface version in the UI, but seems not all GKE versions are equally stable, so for now how about set a stable version for both user and prober test so that the target GKE version is guaranteed to be tested?

kunmingg:
@jlewi 
I saw errors like gke master connection refused | config not found, seems related to version 1.10.9-gke.0.

Can we choose oldest sub version in 1.10 instead of newest? Should be more stable.

bootstrap/cmd/bootstrap/app/options/options.go
jlewi: But if we change the default in the click to deploy UI but not the configs; won't the behavior in the app and via kfctl.sh diverge?

jlewi:
There's no way to automatically pin to the oldest supported patch version; we'd have to write that logic ourselves.

jlewi:
Current valid master versions are

```
validMasterVersions:
- 1.11.2-gke.9
- 1.10.9-gke.0
- 1.10.7-gke.6
- 1.10.6-gke.6
- 1.9.7-gke.6
```
So it looks like patch versions might dis-appear. So if we pin to 1.10.6 we will break the day they drop support for 1.10.6.

kunmingg:
@jlewi 
feel worth adding logic to choose oldest supported patch version. should be more stable:
using newest version: every patch version has its day, from gke0 to gke 9
using oldest version: only top path version got chance, like gke.6, gke.9

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -83,6 +83,10 @@ type ksServer struct {
 	// other information about the regisry.
 	knownRegistries map[string]RegistryConfig
 
+	//GKE version to use, suggest format 1.X.Y, avoid using most recent version
```
jlewi: Please change the comment and variable name. I'd like to convey the fact that this should only be used as a temporary override when things are going wrong. 

so lets call it something like
gkeVersionOverride

and the comment should indicate that  this should only be used temporarily if we need to override the default value in the DM config file.

The DM config file should be the source of truth. We want to avoid setting defaults in multiple places. That will lead to all kinds of confusion. 

For example, we will upgrade the DM configs to use GKE 1.11 and make other changes based on 1.11; this will work with CLI and then break with the web app because the web app will be overriding the GKE version.

If we set this flag the procedure should be to file an issue to unset it and revert to the default value being set in the DM config.

bootstrap/cmd/bootstrap/app/options/options.go
```diff
@@ -47,6 +48,7 @@ func (s *ServerOption) AddFlags(fs *flag.FlagSet) {
 	fs.BoolVar(&s.JsonLogFormat, "json-log-format", true, "Set true to use json style log format. Set false to use plaintext style log format")
 	fs.IntVar(&s.Port, "port", 8080, "The port to use when running an http server.")
 	fs.StringVar(&s.AppDir, "app-dir", "/opt/bootstrap", "The directory for the ksonnet applications.")
+	fs.StringVar(&s.GkeVersion, "GkeVersion", "1.10", "Initial GKE master version")
```
jlewi: The default should be the empty string so that we use the value in the DM config by default.

bootstrap/cmd/bootstrap/app/ksServer.go
jlewi: Can you fix the comment. 
Delete "GKE version to use, suggest format 1.X.Y, avoid using most recent version"

Change it to something like

"gkeVersionOverride allows overriding the GKE version specified in the DM config. If not set the value in the DM config is sued."

jlewi:
One nit picky comment about the comment in ksServer.go 
as soon as that's fixed this ready to go

PR #1902, opened 2018-11-03
Fix skip init project; use spaces consistently for indentation

jlewi:
* skipInitProject is a boolean so we need to use the syntax

if ${SKIP_INIT_PROJECT}

and not

if [ ]

* kfctl.sh was using a mix of spaces and tabs for indentation. Convert to spaces. spaces were being
  used more than tabs and this is consistent with python style.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1902)


ashahba:
This looks good @jlewi 
I tried to address the styles in this PR #1888 as well 🙂 
/lgtm

PR #1906, opened 2018-11-05
Envoy config change for istio

lluunn:
`/istio` will be routed to istio-ingressgatway

Related to
#1844 
#1309 

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1906)

### DIFF
--- a/docs/gke/gateway.yaml
+++ b/docs/gke/gateway.yaml
@@ -1,0 +1,15 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: kubeflow-gateway
+  namespace: kubeflow
+spec:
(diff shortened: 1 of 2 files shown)


docs/gke/gateway.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: Gateway
+metadata:
+  name: kubeflow-gateway
+  namespace: kubeflow
+spec:
+  selector:
+    istio: ingressgatway
```
kunmingg: typo: ingressgateway

lluunn: thanks!

PR #1908, opened 2018-11-05
make prober always sleep before execute

kunmingg:
1. make prober always sleep before execute
2. catch ```kubectl create``` exception when it happens

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1908)

### DIFF
--- a/testing/test_deploy_app.py
+++ b/testing/test_deploy_app.py
@@ -19,6 +19,7 @@ import google.auth.compute_engine.credentials
 import google.auth.iam
 import google.oauth2.credentials
 import google.oauth2.service_account
+from retrying import retry
 
 FILE_PATH = os.path.dirname(os.path.abspath(__file__))
 SSL_DIR = os.path.join(FILE_PATH, "sslcert")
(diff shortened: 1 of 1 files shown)


testing/test_deploy_app.py
```diff
@@ -138,11 +138,18 @@ def insert_ssl_cert(args):
     util_run(("gsutil cp gs://%s/%s/* %s" % (SSL_BUCKET, args.mode, SSL_DIR)).split(' '))
   except Exception:
     logging.warning("ssl cert for %s doesn't exist in gcs" % args.mode)
-    return
-  util_run(("gcloud container clusters get-credentials %s --zone %s --project %s" %
-            (args.deployment, args.zone, args.project)).split(' '))
-  util_run(("kubectl create -f %s" % SSL_DIR).split(' '))
-
+    return True
```
jlewi: Should we use the retrying module rather than writing our own loop for retries?
The retrying module allows you to specify various conditions. Typically, I suggest retrying for up to some maximum amount of time with exponential backoff between retries.

Typically its best to specify a max duration for retries e.g. we will give up after X minutes; because that makes it easier to reason about.

kunmingg: updated

PR #1916, opened 2018-11-06
fixes 'Create an openvino component'

### LINKED ISSUE #1913: Create an openvino component
This component will deploy the [openVINO model server](https://github.com/IntelAI/OpenVINO-model-server/)

kkasravi:
fixes #1913

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1916)


kubeflow/openvino/openvino.libsonnet
```diff
@@ -0,0 +1,101 @@
...
+      service.mixin.metadata.
+        withNamespace(params.namespace).
+        withAnnotationsMixin({
+        "getambassador.io/config":
+          std.join("\n", [
+            "---",
+            "apiVersion: ambassador/v0",
+            "kind:  Mapping",
+            "name: ov-mapping-" + params.name + "-get",
+            "prefix: / " + params.name + "/",
```
lluunn: should be prefix has something before name?
like openvino, or ov

kubeflow/openvino/openvino.libsonnet
```diff
@@ -0,0 +1,101 @@
...
+            "rewrite: /",
+            "method: GET",
```
lluunn: do you only want route for GET?

kubeflow/openvino/openvino.libsonnet
```diff
@@ -0,0 +1,101 @@
...
+
+    local ovContainer =
+      container.new(
+        params.name, params.imageURL
+      ).withImagePullPolicy("IfNotPresent").
+        withArgs([
+        "ie_serving",
+        "model",
+        "--model_path",
+        "/opt/ml/resnet",
```
lluunn: shouldn't model path be a param?
also the model name below.

kubeflow/openvino/openvino.libsonnet
kkasravi: removed GET

kubeflow/openvino/openvino.libsonnet
kkasravi: it will end up as /openvino/ - so similar to jupyter's /hub/ or /user/

lluunn: For hub, it has /hub/ in the prefix.
I think we want /openvino/ in the path, so the prefix should be something like
`"/openvino/" + params.name + "/"` , right?

kkasravi: i agree - but params.name is "openvino" ... where you thinking /openvino/openvino/ ?

kubeflow/openvino/openvino.libsonnet
kkasravi: added these as params

kubeflow/openvino/openvino.libsonnet
lluunn: Why would params.name be openvino?
Shouldn't that be more model specific, like myModelA?

Or you will only deploy one openvino deployment in the cluster?
Even in that case, I still think we should have /openvino/ because params.name is too arbitrary, and user can choose some name that will collide with other path.
WDYT?

kkasravi: I believe there is 1 openvino deployment in the cluster at this time

kkasravi:
Yes, I think we need openvino regardless of what params.name is. I’ll update later today. Thanks.

PR #1923, opened 2018-11-07
miscellaneous updates and enhacements for shell scripts

ashahba:
Initially when I worked on `kfctl.sh` recently I noticed the styles could be enhanced a bit.
Then as I looked at the rest of our `shell` scripts, I came to the conclusion that the same style enhancements can be applied to all.
So I ended up filing #1923 and here we are with the PR to resolve it.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1923)


ashahba:
It could be a bit overwhelming at first to see all these changes at once, but most of them are just add or removing spaces or correcting indentation around line breaks and code blocks.

jlewi:
I think you referenced the wrong issue.

It might be good to start a style guide for bash scripts (not necessarily this PR). And should we open an issue for lint checking of shell scripts?

Not sure where we would put a style guide but this seems like a reasonable location.
https://github.com/kubeflow/kubeflow/tree/master/docs_dev

jlewi:
Thanks @ashahba 

/lgtm
/approve

ashahba:
Thanks @jlewi I'll another PR to add something like `shell_styles.md` to https://github.com/kubeflow/kubeflow/tree/master/docs_dev

and once that's done we can close #1925

PR #1927, opened 2018-11-08
make IAP optional for click-deploy app

kunmingg:
1. make IAP optional for click-deploy app
2. update instructions for both cases (IAP / no IAP)
3. add button for both cases to smooth experience

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1927)

### DIFF
--- a/components/gcp-click-to-deploy/src/App.tsx
+++ b/components/gcp-click-to-deploy/src/App.tsx
@@ -141,8 +141,8 @@ class App extends React.Component<any, { signedIn: boolean }> {
                   <ul>
                     <ul>
                       <li>  Click Cloud Shell; run command in pop up window; kubeflow UI will be at
-                        <a href="http://localhost:8080/" style={{ color: 'inherit', marginLeft: 5 }} target="_blank"
-                        >localhost:8080</a>
+                        <a href="https://ssh.cloud.google.com/devshell/proxy?authuser=0&port=8081" style={{ color: 'inherit', marginLeft: 5 }} target="_blank"
+                        >here</a>
                       </li>
                     </ul>
                   </ul>
--- a/components/gcp-click-to-deploy/src/App.tsx
+++ b/components/gcp-click-to-deploy/src/App.tsx
@@ -140,9 +140,7 @@ class App extends React.Component<any, { signedIn: boolean }> {
                   </ul>
                   <ul>
                     <ul>
-                      <li>  Click Cloud Shell; run command in pop up window; kubeflow UI will be at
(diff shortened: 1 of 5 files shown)


components/gcp-click-to-deploy/src/App.tsx
```diff
@@ -107,22 +107,45 @@ class App extends React.Component<any, { signedIn: boolean }> {
...
+                <div>
+                  To connect to deployed Kubeflow cluster:
+                </div>
+                <div>
+                  <ul>
+                    <li> If you configured IAP Oauth Client ID and Secret: </li>
+                  </ul>
+                  <ul>
+                    <ul>
+                      <li>  Click IAP Access (might need to wait up to 20 minutes on network settings) </li>
```
jlewi: "20 minutes for domain and IAP to setup"

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -262,12 +271,53 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
       if (p.name === 'acmeEmail') {
         p.value = email;
       }
-    });
+
+      if (p.name === 'jupyterHubAuthenticator') {
```
jlewi: So the client side JS is responsible for setting the authenticator for JupyterHub?

jlewi:
This is great; just a minor comment about the instructions.

abhi-g:
Is it possible to add a checkbox for IAP setup, which is unchecked by default. When unchecked the IAP related fields can be hidden from the form, and visible if checked.

abhi-g:
I agree with Jeremy that we can still merge this PR, allowing users the
option to use kubectl port-forward.

kunmingg:
Sure, seems we might be able to automate cloud shell.
Will change frontend a bit and merge this PR.

components/gcp-click-to-deploy/src/App.tsx
```diff
@@ -107,22 +107,45 @@ class App extends React.Component<any, { signedIn: boolean }> {
...
+                    </ul>
+                  </ul>
+                  <ul>
+                    <li> If you leave IAP Oauth Client ID and Secret empty: </li>
+                  </ul>
+                  <ul>
+                    <ul>
+                      <li>  Click Cloud Shell; run command in pop up window; kubeflow UI will be at
```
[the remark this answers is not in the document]
kunmingg: We provide button on UI and prepare script in user's source repo,
when user click button, cloud shell will clone repo and cd into script dir.
User only need to run ```./conn.sh```

components/gcp-click-to-deploy/src/DeployForm.tsx
kunmingg: Ideally not, will take care in https://github.com/kubeflow/kubeflow/issues/1971

PR #1939, opened 2018-11-10
fixes 'ksonnet env should override params'

### LINKED ISSUE #1924: ksonnet env should override params
currently many libsonnet files have the following
```
    local params = _env + _params,
```

however this should be switched to 
```
    local params = _params + _env,
```
so that env values override param values

kkasravi:
fixes #1924

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1939)

### DIFF
--- a/kubeflow/application/application.libsonnet
+++ b/kubeflow/application/application.libsonnet
@@ -5,7 +5,7 @@
   local crd = k8s.apiextensions.v1beta1.customResourceDefinition,
 
   new(_env, _params):: {
-    local params = _env + _params {
+    local params = _params + _env {
       labels: {
         app: _params.name,
       },
--- a/kubeflow/argo/argo.libsonnet
+++ b/kubeflow/argo/argo.libsonnet
@@ -4,7 +4,7 @@
   local k = import "k.libsonnet",
   local util = import "kubeflow/core/util.libsonnet",
   new(_env, _params):: {
-    local params = _env + _params,
+    local params = _params + _env,
 
     // CRD's are not namespace scoped; see
     // https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/
--- a/kubeflow/core/ambassador.libsonnet
+++ b/kubeflow/core/ambassador.libsonnet
@@ -2,7 +2,7 @@
   local k = import "k.libsonnet",
   local util = import "kubeflow/core/util.libsonnet",
   new(_env, _params):: {
-    local params = _env + _params,
+    local params = _params + _env,
 
     local ambassadorService = {
       apiVersion: "v1",
--- a/kubeflow/core/centraldashboard.libsonnet
+++ b/kubeflow/core/centraldashboard.libsonnet
@@ -2,7 +2,7 @@
   local k = import "k.libsonnet",
   local util = import "kubeflow/core/util.libsonnet",
   new(_env, _params):: {
-    local params = _env + _params,
+    local params = _params + _env,
 
     local centralDashboardDeployment = {
       apiVersion: "extensions/v1beta1",
--- a/kubeflow/core/cert-manager.libsonnet
+++ b/kubeflow/core/cert-manager.libsonnet
@@ -1,7 +1,7 @@
 {
   local util = import "kubeflow/core/util.libsonnet",
   new(_env, _params):: {
-    local params = _env + _params,
+    local params = _params + _env,
 
     local certificateCRD = {
       apiVersion: "apiextensions.k8s.io/v1beta1",
--- a/kubeflow/core/cloud-endpoints.libsonnet
+++ b/kubeflow/core/cloud-endpoints.libsonnet
@@ -2,7 +2,7 @@
   local k = import "k.libsonnet",
   local util = import "kubeflow/core/util.libsonnet",
   new(_env, _params):: {
-    local params = _env + _params {
+    local params = _params + _env {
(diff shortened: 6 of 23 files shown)


jlewi:
I left a question in the issue; repasting here

> @kkasravi Can you explain the issue? Its not clear to my why env should override parameters? I would actually not expect env to contain parameters that would also be defined in params. So I wouldn't expect the order to matter.

kkasravi:
@jlewi i didn't expect their to be an intersection of attributes between env and params since we tried to eliminate things like namespace in an earlier issue/PR. But for workflows/components/tfserving.jsonnet namespace is being fed in through params. I think there are still a few places where this old override (params.namespace overrides env.namespace) is being assumed.

lluunn:
Shouldn't we just fix those component that use params.namespace instead?

kkasravi:
@jlewi agreed. 

For the e2e test cases (test_deploy.py) it looks like the `--namespace` argument is passed as parameters rather than through the env. I've changed this in test_deploy.py

Also, here is a common use case which overrides parameter values with env values and fails when 
> local params = _env + _params,

but succeeds when
> local params = _params + _env,

```
  ks generate ambassador ambassador
  ks env add default --namespace=kubeflow
  ks param set services.ambassador ambassadorServiceType LoadBalancer --env default
```

jlewi:
Sorry I'm still a bit confused.

Don't we want parameters to override env?

Parameters are component specific whereas environment is not component specific. So if env overrides parameters then there's no way to customize something on a per component basis (e.g. namespace).

In the example you gave

```
  ks generate ambassador ambassador
  ks env add default --namespace=kubeflow
  ks param set ambassador ambassadorServiceType LoadBalancer --env default
```

Why would this code be affected by the ordering of params and envs?  The "ambassadorServiceType" is a component parameter and I don't think it would ever be an env value. So why would the order matter?

kkasravi:
If you take a look at what is generated from 
```
  ks param set ambassador ambassadorServiceType LoadBalancer --env default
```

It's under ../ks_app/environments/default/params.libsonnet

```
local params = std.extVar('__ksonnet/params');
local globals = import 'globals.libsonnet';
local envParams = params + {
  components+: {
    "ambassador"+: {
      platform: 'null',
      ambassadorServiceType: 'LoadBalancer',
    },
  },
};

{
  components: {
    [x]: envParams.components[x] + globals
    for x in std.objectFields(envParams.components)
  },
}
```

Per the ksonnet [doc](https://github.com/ksonnet/ksonnet/blob/master/docs/concepts.md#parameter) , the environment params are intended to override the component params. In the above that has happened, but then we merge env + params in the prototype which undoes the override.

kkasravi:
> Parameters are component specific whereas environment is not component specific.

```
ks param set ambassador ambassadorServiceType LoadBalancer --env default
```

*`ks param set ambassador`* sets a component specific value *`--env default`* in the environment.

maybe i'm misunderstanding something...

jlewi:
Suppose we define the following component serviceParam.jsonnet. The purpose of this component is just to give us a way of seeing how ksonnet is setting env and params.

```
local env = std.extVar("__ksonnet/environments");
local params = std.extVar("__ksonnet/params").components.serviceParams;
{
   "apiVersion": "apps/v1beta1",
   "kind": "Deployment",
   "params": params,
   "env": env,
}
```

Now set some parameter and do ksshow

```
ks param set --env=default someParameter someValue
ks show default -c serviceParam
```

I get the following output

```
---
apiVersion: apps/v1beta1
env:
  namespace: kubeflow
  server: https://104.196.134.59
kind: Deployment
metadata:
  labels:
    ksonnet.io/component: serviceParams
params:
  containerPort: 80
  image: someimage
  name: serviceParams
  replicas: 1
  someParameter: someValue

```

So the value of the parameter **someParameter** is set in the **params** not **env** value. But more importantly there are no common fields in env and params. So whether we do **params + env** or **env+params** the result is the same. We can confirm this by changing the component to 

```
local env = std.extVar("__ksonnet/environments");
local params = std.extVar("__ksonnet/params").components.serviceParams;
{
   "apiVersion": "apps/v1beta1",
   "kind": "Deployment",
   "params": params,
   "env": env,
   "order1": env+params,
   "order2": params + env,
}
```

So I'm still not following why the order of the addition matters.

kkasravi:
@jlewi 

i went through the steps you did above and you're right - 'env' is disjoint from 'params' and typically just contains 2 members
```
    namespace: kubeflow
    server: https://35.185.234.170
```
 Mixing namespace, server into a component's params probably isn't the best idea since 
a component may want to set a namespace param that is different that the env namespace. 
For example the component may want to provide a namespace param where it deploys pod's to.
I think some components currently use 'deploymentNamespace'. But this mixin dates back to 
ksonnet's original bug/limitation of accessing the env.namespace.

In any case, as you've shown, the order doesn't matter as long as they're disjoint. 
I appreciate you taking time to show this above. The one change in this PR that may be 
worth considering is I pass namespace in separately in test_deploy rather than adding it to params. 
Also this PR could be refactored so that we don't mix _env + _params and explicitly refer to env for namespace, server in all the libsonnets. Your call on this. We have lots of other more important work to get to. Happy to close it as well. It wasn't a total waste of time for me - again thanks for your patience.

jlewi:
I'm fine with the changes in the PR; just wanted to make sure I understood  things.

Opened #1961 to stop merging params and env.

/lgtm
/approve

PR #1942, opened 2018-11-11
Parse all command line options in one place and within function

ashahba:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1942)

### DIFF
--- a/scripts/kfctl.sh
+++ b/scripts/kfctl.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # This script provides commands to initialize and manage Kubeflow
 # deployments.
 #
@@ -17,7 +17,7 @@ ENV_FILE="env.sh"
 SKIP_INIT_PROJECT=false
 CLUSTER_VERSION="1.10"
 
-DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
 source "${DIR}/util.sh"
 source "${DIR}/gke/util.sh"
 source "${DIR}/util-minikube.sh"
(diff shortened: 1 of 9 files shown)


ashahba:
The purpose of this PR is to address parsing all command line options in a single location as requested by @jlewi here:
https://github.com/kubeflow/kubeflow/pull/1888#discussion_r230949335

scripts/kfctl.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
jlewi: Why this change?

scripts/kfctl.sh
```diff
@@ -157,9 +107,63 @@ createNamespace() {
...
+# Generate all required components
+customizeKsApp() {
+  ks param set ambassador platform ${KUBEFLOW_PLATFORM}
+  ks param set jupyter platform ${KUBEFLOW_PLATFORM}
+}
+
+ksApply() {
+  pushd ${KUBEFLOW_KS_DIR}
+
+  if [ "${PLATFORM}" == "minikube" ]; then
```
jlewi: Should we always just create the Namespace? 
createNamespace should be a null op if the namespace doesn't exist.

scripts/kfctl.sh
```diff
@@ -157,9 +107,63 @@ createNamespace() {
...
+    fi
+  fi
+  set -x
+}
+
+
 if [ "${COMMAND}" == "init" ]; then
   DEPLOYMENT_NAME=${WHAT}
 
+  # Parse all command line options
```
jlewi: Should we move the code for argument parsing into a function 
parseArgs

and the remaining code into a function
named  main

and then just invoke main in the script?

jlewi:
Thank you so much for doing this.
Just a couple nitpicky comments.

scripts/kfctl.sh
ashahba: I think that'll make the code even cleaner.
I'll put those inside a function.

scripts/kfctl.sh
ashahba: Agreed. I think the `if` is not needed here.

ashahba:
@jlewi This one is ready to go let me know if anything else needs to be done.

scripts/kfctl.sh
jlewi: Seems good; thanks for the explanation.

jlewi:
Thanks.
/lgtm
/approve

PR #1945, opened 2018-11-12
Redirect deployer webapp page to Kubeflow dashboard after its ready

### LINKED ISSUE #1420: [kfctl] Web app redirect after IAP up.
Kfctl web app should redirect to the newly created kubeflow UI once IAP is available.

abhi-g:
fixes #1420 

Fetching a GET on the dashboard URI which belongs in a different domain from the deployer in general gets prevented by the browser due to CORS prevention. This fix gets around that issue by leveraging the fact that a static image fetch hosted at the other dashboard site would not be prevented by the browser. Once that image is able to load within the deployer, it can redirect the user to the dashboard.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1945)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -460,6 +460,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
   }
 
   private _monitorDeployment(project: string, deploymentName: string) {
+    const dashboardUri = 'https://' + this.state.deploymentName + '.endpoints.' + this.state.project + 'cloud.goog/';
     const monitorInterval = setInterval(() => {
       Gapi.deploymentmanager.get(this.state.project, deploymentName)
         .then(r => {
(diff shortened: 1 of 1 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -482,6 +483,29 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     }, 10000);
   }
 
+  private _redirectToKFDashboard(dashboardUri: string) {
+    // relying on JupyterHub logo image to be available when the site is ready.
```
jlewi: Is this because of the cors issue? Might be worth elaborating on that per the PR description.

jlewi:
This is great.

jlewi:
/lgtm
/approve

/hold 
in case @kunmingg wants to review.

components/gcp-click-to-deploy/src/DeployForm.tsx
abhi-g: done.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -482,6 +483,35 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+    // request and not an AJAX request, thus bypassing the CORS in this
+    // case.
+    const imgUri = dashboardUri + 'hub/logo';
+    const startTime = new Date().getTime() / 1000;
+    const img = document.createElement('img');
+    img.src = imgUri + '?rand=' + Math.random();
+    img.id = 'ready_test';
+    img.onload = () => { window.location.href = dashboardUri; };
+    img.onerror = () => {
+      const timeSince = (new Date().getTime() / 1000) - startTime;
```
kunmingg: Can we limit request frequency to like once per 10 seconds?
Just to avoid any rate limit.

abhi-g: Since its html load of the image, we don't have control over the loading. The browser itself controls the fetch of the image. Generally, the default timeout on a resource fetch by the browser is usually in minutes. For eg. in Chrome, it will wait 5min before it times out and raises an error.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -482,6 +483,35 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+      if (timeSince > 1500) {
+        this._appendLine('Could not redirect to Kubeflow Dashboard at: ' + dashboardUri);
+      } else {
+        const ready_test = document.getElementById('ready_test') as HTMLImageElement;
+        if (ready_test != null) {
```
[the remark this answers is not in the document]
abhi-g: done.

components/gcp-click-to-deploy/src/DeployForm.tsx
kunmingg: The DNS record is not up on initial minutes, so img load would fails immediately and keep busy waiting till DNS up. We might have huge amount of log output and img load requests.

PR #1951, opened 2018-11-14
Add pipeline to centaldashboard

yupbank:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1951)


abhi-g:
FYI /cc @yebrahim @IronPan

yebrahim:
Are we ready to do this yet? Should we wait until Pipelines is deployed with Kubeflow by default?

abhi-g:
It shouldn't hurt and might be helpful if someone has deployed pipelines
into their KF cluster.

PR #1959, opened 2018-11-15
Enable node autoprovisioning in v1beta1 clusters

richardsliu:
This enables the [node auto-provisioning](https://cloud.google.com/kubernetes-engine/docs/how-to/node-auto-provisioning) feature in Kubeflow clusters. Parameters are configured in the cluster-kubeflow.yaml file.

Note that GKE v1beta1 API is required because the feature is still in Beta. GA is expected early 2019.

Also bumps the cluster version to 1.11.

### DIFF
--- a/deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml
+++ b/deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml
@@ -31,7 +31,7 @@ resources:
     zone: SET_THE_ZONE
     # "1.X": picks the highest valid patch+gke.N patch in the 1.X version
     # https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters
-    cluster-version: "1.10"
+    cluster-version: "1.11"
     # Set this to v1beta1 to use beta features such as private clusterss
     # and the Kubernetes stackdriver agents.
     gkeApiVersion: v1
@@ -53,6 +53,12 @@ resources:
     gpu-pool-min-nodes: 0
     gpu-pool-max-nodes: 0
     gpu-type: nvidia-tesla-k80
+    # Autoprovisioning parameters (only supported in gkeApiVersion v1beta1)
+    autoprovisioning-config:
+      enabled: true
+      max-cpu: 20
+      max-memory: 200
+      max-nvidia-tesla-k80: 8
     # Whether to enable TPUs
     enable_tpu: false
     securityConfig:
--- a/deployment/gke/deployment_manager_configs/cluster.jinja
+++ b/deployment/gke/deployment_manager_configs/cluster.jinja
@@ -85,6 +85,18 @@ resources:
           {{ properties['securityConfig']['masterAuthorizedNetworksConfigCidr'] }}
         {% endif %}
       {% endif %}
+      # Autoprovisioning is only supported in v1beta1.
+      {% if properties['gkeApiVersion'] == 'v1beta1' and properties['autoprovisioning-config']['enabled'] %}
+      autoscaling:
+        enableNodeAutoprovisioning: true
+        resourceLimits:
+        - resourceType: 'cpu'
+          maximum: {{ properties['autoprovisioning-config']['max-cpu'] }}
+        - resourceType: 'memory'
+          maximum: {{ properties['autoprovisioning-config']['max-memory'] }}
+        - resourceType: 'nvidia-tesla-k80'
+          maximum: {{ properties['autoprovisioning-config']['max-nvidia-tesla-k80'] }}
+      {% endif %}
       nodePools:
       - name: default-pool
         initialNodeCount: {{ properties['cpu-pool-initialNodeCount'] }}


deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml
```diff
@@ -31,7 +31,7 @@ resources:
     zone: SET_THE_ZONE
     # "1.X": picks the highest valid patch+gke.N patch in the 1.X version
     # https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters
-    cluster-version: "1.10"
+    cluster-version: "1.11"
     # Set this to v1beta1 to use beta features such as private clusterss
     # and the Kubernetes stackdriver agents.
     gkeApiVersion: v1
```
kunmingg: Do you want to move this one to cluster.jinja, and make a mapping from GKE version?
1.10 -> v1
1.11 -> v1beta1

texasmichelle:
Once NAP is out of beta, these settings will affect clusters created with click-to-deploy, right? Are we adding this now so that it can be manually invoked?

Those seem like reasonable defaults to start with. We can adjust later based on user feedback.

deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml
richardsliu: These should not be mapped 1:1. And I think we need 1.11 anyway to support multiple CRD versions.

kunmingg: Or make:
```gkeApiVersion: v1beta1```
to have test coverage

**Comment by jlewi** [comment]
As Kunming suggests; any thoughts on how we can add E2E test coverage for the new settings?

I think with 1.11 they might be changing the behavior of Ingress so that the ingress won't be created if the secret doesn't exist. I think we already have a workaround for this but it would be good to verify.

**Comment by richardsliu** [comment]
We can quickly run into quota problems if we are testing v1beta1 clusters in every presubmit. Maybe just periodic?

**Comment by richardsliu** [comment]
@texasmichelle Correct, this is just adding the support for NAP so that it can be manually configured. This also allows the user to change provisioning parameters (such as GPU accelerator type) before the actual deployment. We can tune them according to user feedback when we enable it by default.

File: prow_config.yaml [hunk 2]
```diff
@@ -26,19 +26,35 @@ workflows:
       - deployment/*
     params:
       platform: gke
+      gkeApiVersion: v1
+  - app_dir: kubeflow/kubeflow/testing/workflows
+    component: kfctl_test
+    name: kfctl-beta
+    job_types:
+      - presubmit
```
jlewi: Per you comment; I thought we didn't want to test v1beta as a presubmit? 

Would it make sense to only test v1beta1 on presubmit if the GKE configs change?

scripts/kfctl.sh
```diff
@@ -178,6 +181,10 @@ parseArgs() {
         shift
         EMAIL=$1
         ;;
+      --gkeApiVersion)
```
jlewi: Why do we need to expose gkeApiVersion as a command line argument? Can't users just modify the YAML config to set it? 

In the E2E tests we can just use yq or similar to script modifying the config as needed.

In general, I'd like to avoid plumbing more and more of the options in the YAML config file as command line arguments. I don't think that really scales.

deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml
```diff
@@ -31,10 +31,10 @@ resources:
     zone: SET_THE_ZONE
     # "1.X": picks the highest valid patch+gke.N patch in the 1.X version
     # https://cloud.google.com/kubernetes-engine/docs/reference/rest/v1/projects.zones.clusters
-    cluster-version: "1.10"
+    cluster-version: "1.11"
     # Set this to v1beta1 to use beta features such as private clusterss
     # and the Kubernetes stackdriver agents.
-    gkeApiVersion: v1
+    gkeApiVersion: SET_GKE_API_VERSION
```
jlewi: This is outside the scope of this PR. The fact that we can't set meaningful default values because we rely on SED to do substitution is not good.

It would be better to use a tool like jsonnet or yq to allow more powerful substitution. We've avoided that so far because it would introduce additional dependencies.

We should be able to do that once we move to replacing kfctl.sh with a go binary.

jlewi:
Thanks @richardsliu I think only issue I see is whether or not gkeApiVersion needs to be a command line argument to kfctl.sh. I think I would prefer if possible to just tell users to modify the .yaml file rather than plumbing up more options as command line arguments.

deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml
richardsliu: Ack

prow_config.yaml
richardsliu: Changed - scoped this to when gke scripts or config files are changed.

PR #1981, opened 2018-11-21
add pipeline as part of kf deployment

IronPan:
And promote Argo as a required component
Fixed the parts.yaml format

Verified it works with kfctl.sh
TODO verify one-click-deployment works. 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1981)

### DIFF
--- a/bootstrap/config/default.yaml
+++ b/bootstrap/config/default.yaml
@@ -5,6 +5,10 @@ app:
   components:
     - name: jupyterhub
       prototype: jupyterhub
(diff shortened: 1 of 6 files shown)


IronPan:
not for now. the only parameters are images which have default values and should be updated when pipeline has new release. 
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/pipeline/prototypes/pipeline.jsonnet#L6

PR #1989, opened 2018-11-22
Make it easier to debug jsonnet tests

### LINKED ISSUE #1988: Better testing output for jsonnet
Right now are jsonnet tests work by having a jsonnet file that does std.assertEqual to raise an error.

This makes it very difficult to debug test failures. The output ends up being something like.

```
RUNTIME ERROR: Assertion failed. {"apiVersion": "apiextensions.k8s.io/v1beta1", "kind": "CustomResourceDefinition", "metadata": {"name": "tfjobs.kubeflow.org"}, "spec": {"group": "kubeflow.org", "names": {"kind": "TFJob", "plural": "tfjobs", "singular": "tfjob"}, "validation": {"openAPIV3Schema": {"properties": {"spec": {"properties": {"tfReplicaSpecs": {"properties": {"Chief": {"properties": {"replicas": {"maximum": 1, "minimum": 1, "type": "integer"}}}, "PS": {"properties": {"replicas": {"minimum": 1, "type": "integer"}}}, "Worker": {"properties": {"replicas": {"minimum": 1, "type": "integer"}}}}}}, "scope": "Namespaced"}}}}, "version": "v1alpha2"}} != {"apiVersion": "apiextensions.k8s.io/v1beta1", "kind": "CustomResourceDefinition", "metadata": {"name": "tfjobs.kubeflow.org"}, "spec": {"group": "kubeflow.org", "names": {"kind": "TFJob", "plural": "tfjobs", "singular": "tfjob"}, "scope": "Namespaced", "validation": {"openAPIV3Schema": {"properties": {"spec": {"properties": {"tfReplicaSpecs": {"properties": {"Chief": {"properties": {"replicas": {"maximum": 1, "minimum": 1, "type": "integer"}}}, "PS": {"properties": {"replicas": {"minimum": 1, "type": "integer"}}}, "Worker": {"properties": {"replicas": {"minimum": 1, "type": "integer"}}}}}}}}}}, "version": "v1alpha2"}}
	std.jsonnet:758:7-50	function <anonymous>
	./kubeflow/tf-training/tests/tf-job_test.jsonnet:(25:1)-(83:2)
```

A better solution would be to output pretty formatted objects that make it easy to see actual and expected.

We can accomplish this by outputting a json object representing a test suite

```
{
  "pass": true,
  "testCases": [
      {
        "actual": ...,
        "expected": ...,
      },
      ...
  ],
}
```

In python we can load the output of json eval and test the value of pass.
https://github.com/kubeflow/kubeflow/blob/fee939dcfc4530629be93e54bebba226e0dfbdbb/testing/test_jsonnet.py#L60

jlewi:
* jsonnet tests should be able to output a json object containing a field
  "pass"

* The python test wrapper can then check this value to see if the test passed.

* This allows the object to output information about expected and actual
  values which make it easier to debug.

Fix #1988

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/1989)

### DIFF
--- a/kubeflow/pytorch-job/tests/pytorch-job_test.jsonnet
+++ b/kubeflow/pytorch-job/tests/pytorch-job_test.jsonnet
@@ -224,4 +224,4 @@ local testSuite = suiteResult {
   testCases: testCasesWithResults,
 };
 
-testSuite.pass
+testSuite
--- a/kubeflow/pytorch-job/tests/pytorch-job_test.jsonnet
+++ b/kubeflow/pytorch-job/tests/pytorch-job_test.jsonnet
@@ -224,6 +224,4 @@ local testSuite = suiteResult {
   testCases: testCasesWithResults,
 };
 
-// TODO(https://github.com/kubeflow/kubeflow/issues/1988): Output testSuite
-// rather than std.assertEqual
-std.assertEqual(testSuite.pass, true)
+testSuite.pass
--- a/testing/test_jsonnet.py
+++ b/testing/test_jsonnet.py
@@ -61,7 +61,13 @@ def run(test_files_dirs, jsonnet_path_args, test_case):
             output = util.run(
               ['jsonnet', 'eval', full_path] + jsonnet_path_args,
               cwd=os.path.dirname(full_path))
-            parsed = json.loads(output)
+            try:
+              parsed = json.loads(output)
+            except AttributeError:
+              logging.error(
+                "Output of jsonnet eval could not be parsed as json; "
+                "output: %s", output)
+              parsed = {}
 
             test_passed = parsed.get("pass", false)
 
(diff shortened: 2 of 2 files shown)


kubeflow/pytorch-job/tests/pytorch-job_test.jsonnet
```diff
@@ -224,6 +224,4 @@ local testSuite = suiteResult {
   testCases: testCasesWithResults,
 };
 
-// TODO(https://github.com/kubeflow/kubeflow/issues/1988): Output testSuite
-// rather than std.assertEqual
-std.assertEqual(testSuite.pass, true)
+testSuite.pass
```
lluunn: should be testSuite only?
Since you are getting the "pass" key in python.

jlewi: You're right thanks.

PR #2000, opened 2018-11-27
Fixed formatting issues with new-package-stub

ellis-bigelow:
parts.yaml generated JSON. Json was also invalid, leading to development issues and slowdown. Simplified another hurdle, variable name does not need to be replaced if simply named "package"

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2000)

### DIFF
--- a/kubeflow/new-package-stub/parts.yaml
+++ b/kubeflow/new-package-stub/parts.yaml
@@ -1,22 +1,16 @@
-{
-   "name": "give-this-package-a-name",
-   "apiVersion": "0.0.1",
-   "kind": "ksonnet.io/parts",
-   "description": "An empty package used as a stub for new packages.\n",
-   "author": "kubeflow-team <kubeflow-discuss@googlegroups.com>",
-   "contributors": [
-   ],
-   "repository": {
-      "type": "git",
-      "url": "https://github.com/kubeflow/kubeflow"
-   },
-   "bugs": {
-      "url": "https://github.com/kubeflow/kubeflow/issues"
-   },
-   "keywords": [
-      "kubernetes",
-      "kubeflow",
-      "machine learning"
-   ],
-   "license": "Apache 2.0",
-}
+name: "${PACKAGE_NAME}"
+apiVersion: 0.0.1
+kind: ksonnet.io/parts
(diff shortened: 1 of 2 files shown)


ellis-bigelow:
I signed it!

ellis-bigelow:
I signed it!

kubeflow/new-package-stub/parts.yaml
```diff
@@ -1,22 +1,16 @@
...
-      "url": "https://github.com/kubeflow/kubeflow/issues"
-   },
-   "keywords": [
-      "kubernetes",
-      "kubeflow",
-      "machine learning"
-   ],
-   "license": "Apache 2.0",
-}
+name: "${PACKAGE_NAME}"
```
lluunn: interesting... parts.yaml is a json.
Do you know what this file is for?
I randomly checked other packages, and their parts.yaml are json and are fine?

ellis-bigelow: This is the file that is copied and pasted for creating new components. Everyone else has parts.yaml because they copied this file. I found it confusing to be json when the file is clearly named .yaml.

ellis-bigelow: Additionally, i tried to make it more clear which parts needed to substituted by the user.

lluunn: https://github.com/ksonnet/parts/blob/master/incubator/memcached/parts.yaml

The official ksonnet examples are also json though...

lluunn: Asking them: https://github.com/ksonnet/parts/issues/71

texasmichelle: Agree that changing this to yaml is less confusing

jlewi:
@lluunn @ellis-bigelow are we planning on merging this or should we close it out?

jlewi:
@ellis-bigelow ping?

jlewi:
@ellis-bigelow @lluunn ping?

lluunn:
I think we should be consistent with other ks packages for now.
Not so much benefit to rewrite all parts.yaml

Issue #2012, opened 2018-11-29
Grant kubeflow user service account CMLE permission


IronPan:
Pipeline uses kubeflow user account to interact CMLE in a few scenarios, e.g. upload model to CMLE through jupyter notebook.

jlewi:
Closing this issue because the PR has been merged.

PR #2013, opened 2018-11-29
grant KF user SA CMLE admin permission

IronPan:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2013)


IronPan:
fix #2012

PR #2019, opened 2018-11-29
move ks app config to backend & prepare config isolation for each kf version

kunmingg:
1. make frontend simpler and more portable
2. provide isolated kf config sets for each release as there'll be inconsistency between versions.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2019)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/gcpUtils.go
+++ b/bootstrap/cmd/bootstrap/app/gcpUtils.go
@@ -59,9 +59,45 @@ func init() {
 	prometheus.MustRegister(deploymentsStartedCounter)
 }
 
+func (s *ksServer) GenerateAppConfig(configPath string, req *CreateRequest) error {
+	var appConfigFile AppConfigFile
+	if err := LoadConfig(path.Join(configPath, req.KfVersion + ".yaml"), &appConfigFile); err != nil {
+		return err
+	}
+	appConfig := appConfigFile.App
+	authParamIdx := -1
+	for idx, param := range appConfig.Parameters {
+		if param.Name == "ipName" {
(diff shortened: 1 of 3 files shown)


jlewi:
So with this PR every KF Version has its own YAML file kfversion.yaml which specifies 
AppConfig
https://github.com/kubeflow/kubeflow/blob/97e481a807d0b2cce20b6f72aba1b1bf43e48b04/bootstrap/cmd/bootstrap/app/server.go#L84

which specifies the components and parameters to install.

Does this mean we need to generate a new YAML config file for every release of Kubeflow even minor releases for which the AppConfig doesn't change?

What if people want to deploy a version of KF that isn't explicitly supported by the click to deploy app?

We don't need to resolve it in this issue but it would be good to think about ways where people could specify arbitrary commits for the packages and deploy them. 

This relates to our discussion about composability (#2024).

jlewi:
From our discussion this a clear improvement over the current situation.

It would be good to think about ways to differentiate a config push from a binary push. So that we could just add new versions to the deploy app without changing the binary.

jlewi:
@kunmingg Any update on this PR?

kunmingg:
@jlewi 
Will resolve conflicts and test on v0.4.0 once release available.

jlewi:
@kunmingg reminder this PR is still pending.

kunmingg:
/hold
TODO:
Change to use new object struct & config files shared with kfctl.

jlewi:
@kunmingg I'm assuming this isn't ready for review. Let me know if that's not correct.

jlewi:
@kunmingg What's the status of this PR is it still relevant?

kunmingg:
Seems out of date, will close for now.

Issue #1909, opened 2018-11-05
Figure out Istio installation in Kubeflow


lluunn:
We probably have to checkin the manifest.

https://www.kubeflow.org/docs/guides/components/istio/

lluunn:
Things to configure:
- sidecar injection policy: I think we should label the kubeflow namespace so that sidecar is auto injected if we add proper annotation to the deployment
- egress: we should allow egress traffic to load model from GCS/S3. However, this needs to know the IP of the cluster, so differs from cluster to cluster, see https://istio.io/docs/tasks/traffic-management/egress/#calling-external-services-directly
- Istio ingressgateway: 
  - We want the gateway to sit behind ingress, so it should be NodePort instead of LoadBalancer. 
  - Maybe change the name
  - We will need some configuration setup for JWT validation later

jlewi:
@lluunn What is the remaining work to be done to close out this issue?

PR #2020, opened 2018-11-29
Istio manifest

lluunn:
For #1909 

/cc @jlewi


kunmingg:
Any plan or blocker to make it a ksonnet prototype?

lluunn:
No, the plan is to leave it as yaml.

lluunn:
/hold

I am seeing an error on my deployment with sidecar inject:

```
 Internal error occurred: admission webhook "sidecar-injector.istio.io" denied the request: error converting YAML to JSON: yaml: line 12: found unknown escape character
```

istio/install/README.md
```diff
@@ -0,0 +1,16 @@
+# Istio manifest for Kubeflow
+
+The manifest is taken from OSS Istio release 1.0.2.
+
+We need custom configuration as discussed in the
+[issue](https://github.com/kubeflow/kubeflow/issues/1909#issuecomment-438409215).
+
+Specifically, the things we changed are:
+
+1. The policy of configmap `istio-sidecar-injector` is changed from `enabled` to `disabled`.
```
ellis-bigelow: I thought Istio injector is disabled by default. IIUC, their suggested pattern is to enable it by namespace. E.G., you make a mesh with a gateway per namespace.

istio/install/README.md
```diff
@@ -0,0 +1,16 @@
+# Istio manifest for Kubeflow
```
ellis-bigelow: Consider nesting this file path one layer deeper. We want to be able to extend this pattern for other large deps (Thinking of Knative here).
E.G. dependencies/istio

istio/install/README.md
```diff
@@ -0,0 +1,16 @@
...
+[issue](https://github.com/kubeflow/kubeflow/issues/1909#issuecomment-438409215).
+
+Specifically, the things we changed are:
+
+1. The policy of configmap `istio-sidecar-injector` is changed from `enabled` to `disabled`.
+   According to this [table](https://github.com/istio/istio/issues/6476#issuecomment-399219937),
+   `policy=disabled` and `namespace label=enabled` is what we want.
+1. The service type of `istio-ingressgateway` is changed from `LoadBalancer` to `NodePort`.
+
+*TODO*: To allow egress, we need to know cluster specific IP ranges. Also, Istio's recommended way is to use
```
ellis-bigelow: Consider a service entry for www.googleapis.com. I was able to use this to make GCS work. Alternatively, we should provide IP ranges here for each cloud. They're static for all except GCP, which I think is static but not guaranteed.

istio/install/istio-noauth.yaml
```diff
@@ -0,0 +1,5159 @@
+apiVersion: v1
```
ellis-bigelow: Why noauth? Does "auth" refer to RBAC or sidecar?

istio/install/README.md
lluunn: The policy is enabled by default in the configmap.
You can check istio's original manifest.

Enabled doesn't mean it's always injecting, it's a combination of this enabled/disabled and namespace label to decide.

istio/install/README.md
lluunn: thanks, let's leave it as todo. I will try it out.

istio/install/istio-noauth.yaml
lluunn: noauth means no mutual TLS between service

istio/install/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+We need custom configuration as discussed in the
+[issue](https://github.com/kubeflow/kubeflow/issues/1909#issuecomment-438409215).
+
+Specifically, the things we changed are:
+
+1. The policy of configmap `istio-sidecar-injector` is changed from `enabled` to `disabled`.
+   According to this [table](https://github.com/istio/istio/issues/6476#issuecomment-399219937),
+   `policy=disabled` and `namespace label=enabled` is what we want.
+1. The service type of `istio-ingressgateway` is changed from `LoadBalancer` to `NodePort`.
```
ellis-bigelow: Do you plan to eventually revert this and do IAP with Istio?

istio/install/README.md
lluunn: done

ellis-bigelow:
lgtm

PR #2025, opened 2018-11-30
Grafana routing rule

lluunn:
This will expose grafana dashboard at
https://KFDEPLOY.endpoints.PROJECT.cloud.goog/istio/grafana/

/cc @jlewi 
/cc @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2025)

### DIFF
--- a/istio/grafana-virtual-service.yaml
+++ b/istio/grafana-virtual-service.yaml
@@ -1,0 +1,24 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: grafana-vs
+  namespace: kubeflow
+spec:
+  hosts:
+  - "*"
+  gateways:
+  - "kubeflow-gateway"
+  http:
+  - match:
+    - uri:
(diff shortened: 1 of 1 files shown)


istio/grafana-virtual-service.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: networking.istio.io/v1alpha3
+kind: VirtualService
+metadata:
+  name: grafana-vs
+  namespace: kubeflow
```
kunmingg: Is VirtualService namespace scoped or cluster-wise?

lluunn: namespaced

kunmingg: So should we make namespace a parameter or ```kubeflow``` should be good for most cases?

lluunn: I think kubeflow is good for now.

PR #2034, opened 2018-12-03
always convert numGpu to int

### LINKED ISSUE #1972: TFServing template needs to convert numGpus from string to int
Here
https://github.com/kubeflow/kubeflow/blob/41050ef854948600ef7f922f2a3c295b60b1f7c0/kubeflow/tf-serving/tf-serving-template.libsonnet#L15

We compare numGpus to a string. This check will fail if numGpus is an integer.

We should be converting numGpus to an int always (since parameters can be strings) and then use integer comparison.

lluunn:
Fix #1972 

/cc @jlewi 
/cc @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2034)


jlewi:
How difficult would it be to write a jsonnet test for this?

lluunn:
Added unit test, PTAL

PR #2037, opened 2018-12-04
add permission to get pod's log

IronPan:
Pipeline component needs the permission to  get tf serving pod log
https://github.com/kubeflow/pipelines/blob/master/components/kubeflow/deployer/src/deploy.sh#L202


r2d4:
Dupe #2015  :)

PR #2039, opened 2018-12-04
Remove Seldon Cluster Roles by default

### LINKED ISSUE #1452: Can we scope SeldonCore to a namespace
See kubeflow/kubeflow#1213
and kubeflow/tf-operator#759

We'd like to be able to scope Kubeflow to a particular namespace to better support multi-tenant environments.

It looks like SeldonCore is using ClusterRoles; could we add options to our ksonnet package to scope it to a particular namespace and use namespace scoped Roles?

/cc @cliveseldon

cliveseldon:
* Adds an optional parameter to include ClusterRoles, false by default.

Fixes #1452

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2039)

### DIFF
--- a/kubeflow/seldon/prototypes/core.jsonnet
+++ b/kubeflow/seldon/prototypes/core.jsonnet
@@ -59,7 +59,7 @@ local rbac1 = [
 
 local rbac3 = [
   core.parts(name, namespace, seldonVersion).rbacClusterRole(),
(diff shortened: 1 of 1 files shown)


jlewi:
Thanks @cliveseldon 
/lgtm
/approve

PR #2042, opened 2018-12-05
Easy for code reading

wangkeqiang123:
add blank and modify some words.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2042)

### DIFF
--- a/testing/vm_util.py
+++ b/testing/vm_util.py
@@ -35,7 +35,7 @@ def wait_for_operation(client,
     op: The final operation.
 
   Raises:
-    TimeoutError: Waiting for operation to complete timeout.
+    TimeoutError: if we timeout waiting for the operation to complete.
   """
   endtime = datetime.datetime.now() + timeout
   while True:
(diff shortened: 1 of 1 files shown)


testing/vm_util.py
```diff
@@ -35,7 +35,7 @@ def wait_for_operation(client,
     op: The final operation.
 
   Raises:
-    TimeoutError: if we timeout waiting for the operation to complete.
+    TimeoutError: Waiting for operation to complete timeout.
```
lluunn: I think the original one is more clear

testing/test_deploy.py
```diff
@@ -446,7 +446,7 @@ def deploy_minikube(args):
       raise
     content = json.loads(e.content)
     if content.get("error", {}).get("code") == requests.codes.CONFLICT:
-      # We don't want to keep going so we reraise the error after logging
+      # We don't want to keep going so we raise the error after logging
```
lluunn: this is intended to be reraise (or re-raise), right?

testing/vm_util.py
wangkeqiang123: Okay

testing/test_deploy.py
wangkeqiang123: 'reraise' may be the better choice.

wangkeqiang123: 'reraise' may be the better choice.

wangkeqiang123:
@lluunn @jlewi  Pls review my PR in your spare time, thanks a lot.

wangkeqiang123:
@lluunn @jlewi pls leaves an /ok-to-test message. Thans very much.

PR #2043, opened 2018-12-05
fixes 'Move components in core into a GCP specific ksonnet package'

### LINKED ISSUE #1996: Move components in core into a GCP specific ksonnet package
See #42 

We should move the components in the core package that are GCP specific into their own ksonnet package.

/cc @kunmingg

kkasravi:
fixes #1996

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2043)


kubeflow/gcp/OWNERS
```diff
@@ -0,0 +1,3 @@
+approvers:
+  - gaocegege
```
jlewi: Not the right set of approvers for GCP :)
Please list Googlers
jlewi, r2d4,  richardsliu, kunming, llunnn, abhi-g

jlewi:
Thanks so much. For doing this @kkasravi 

Test is likely failing because you need to update kfctl

kubeflow/gcp/OWNERS
kkasravi: my bad, i didn't realize that's where you look

kubeflow/gcp/OWNERS
```diff
@@ -0,0 +1,3 @@
...
+reviewers:
```
jlewi: If there are new reviewers just remove the reviewers section.

jlewi:
looks like the OWNERs file might still need to be updated.

kkasravi:
@jlewi OWNERs file should be good now

kubeflow/gcp/OWNERS
jlewi: Can you remove gaocegege please?

jlewi:
@kkasravi Can you remove gaocegege as an approver since this is GCP specific?

PR #2050, opened 2018-12-06
Add modeldb as a package for kubeflow

mpvartak:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2050)


kubeflow/modeldb/modeldb.libsonnet
```diff
@@ -0,0 +1,319 @@
...
+              component: "backend",
+            },
+            name: "modeldb-backend",
+          },
+          spec: {
+            containers: [
+              {
+                env: [
+                  {
+                    name: "MONGODB_HOST",
```
inc0: could you make this a param? Also make mongodb deployment optional. Some public clouds or on-prems have dedicated mongodb clusters that are replicated and it'd be good to allow to easily use them

mpvartak: Fixed to allow MONGODB_HOST to be specified in the params.

kubeflow/modeldb/prototypes/modeldb.jsonnet
```diff
@@ -0,0 +1,17 @@
+// @apiVersion 0.1
+// @name io.ksonnet.pkg.modeldb
+// @description Model management service for KubeFlow
+// @shortDescription Model management
+// @param name string Name to give to each of the components
+// @optionalParam modeldbImage string mitdbg/modeldb-backend:latest The image for modeldb
```
jlewi: Can we pin to immutable tags so that the deployment won't suddenly break because the latest tag was moved?

kubeflow/registry.yaml
```diff
@@ -70,6 +70,9 @@ libraries:
   kubebench:
     version: master
     path: kubebench
+  modeldb:
+    version: master
+    path: modeldb
   tf-batch-predict:
     version: master
-    path: tf-batch-predict
+    path: tf-batch-predict
```
jlewi: Did you add some extra whitespace here?

scripts/kfctl.sh
```diff
@@ -142,6 +142,7 @@ ksApply() {
   ks apply default -c spartakus
   ks apply default -c argo
   ks apply default -c pipeline
+  ks apply default -c modeldb
```
jlewi: Can we hold off installing modeldb by default? I think it would be good to get some usage as an optional package and then use that to install it by default if it makes sense.

scripts/util.sh
```diff
@@ -72,6 +72,7 @@ createKsApp() {
   ks pkg install kubeflow/metacontroller
   ks pkg install kubeflow/profiles
   ks pkg install kubeflow/application
+  ks pkg install kubeflow/modeldb
```
jlewi: You can keep this command so we always install modeldb into the app

scripts/util.sh
```diff
@@ -87,6 +88,7 @@ createKsApp() {
   ks generate pipeline pipeline
 
   ks generate katib katib
+  ks generate modeldb modeldb
```
jlewi: Per comment above I would prefer to make modeldb optional until we get some usage on it.

jlewi:
It looks like the test failure is just jsonnet formatting.
You can use this script to autoformat:
https://github.com/kubeflow/kubeflow/blob/master/scripts/autoformat_jsonnet.sh

scripts/kfctl.sh
mpvartak: Done

scripts/util.sh
mpvartak: Done

kubeflow/registry.yaml
mpvartak: Fixed.

PR #2051, opened 2018-12-06
More istio manifest, and README

lluunn:
/cc @jlewi 
/cc @ellis-bigelow

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2051)

### DIFF
--- a/dependencies/istio/README.yaml
+++ b/dependencies/istio/README.yaml
@@ -1,0 +1,7 @@
+# Manifest for Istio in Kubeflow
+
+- `install` dir contains manifest to install Istio CRDs and core components
+- kf-istio-resources.yaml has
+  - Gateway for routing
+  - VirtualService for Grafana
+  - ServiceEntry and VirtualService for egress traffic


dependencies/istio/kf-istio-resources.yaml
```diff
@@ -0,0 +1,109 @@
...
+      weight: 100
+---
+apiVersion: networking.istio.io/v1alpha3
+kind: ServiceEntry
+metadata:
+  name: google-storage-api-entry
+  namespace: kubeflow
+spec:
+  hosts:
+  - storage.googleapis.com
```
ellis-bigelow: Why is this one necessary?

ellis-bigelow: Strange. I was able to make this work using only www. What did you set as your GCS path?

lluunn: see the error:
```
2018-12-05 00:29:16.322501: I external/org_tensorflow/tensorflow/core/platform/cloud/retrying_utils.cc:77] The operation failed and will be automatically retried in 1.16501 seconds (attempt 1 out of 10), caused by: Unavailable: Error executing an HTTP request: libcurl code 35 meaning 'SSL connect error', error details: BoringSSL SSL_connect: SSL_ERROR_SYSCALL in connection to storage.googleapis.com:443
	 when reading gs://kubeflow-examples-data/mnist/1/saved_model.pb
```

ellis-bigelow: Interesting. I wonder if their routing sometimes has legacy DNS for some buckets. LGTM

PR #2055, opened 2018-12-06
Merge Pipeline integration change to v0.3

IronPan:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2055)


components/tensorflow-notebook-image/Dockerfile
```diff
@@ -8,6 +8,7 @@ FROM $BASE_IMAGE
 ARG INSTALL_TFMA=yes
```
r2d4: @kunmingg @richardsliu how should we handle a release of the notebook images for the 0.3 release?

Merge this and then release from the 0.3-branch? The UI will then need to be updated to show this as well right?

IronPan: Should we check in this and wait for the new image to be built and then do another PR in v0.3 to update image version? 

FYI this is an optional feature to bake the pipeline SDK into jupyter image. User can always do pip install manually. It's ok if we don't have it, i think @jlewi WDYT?

PR #2057, opened 2018-12-06
cherry pick central dashboard changes to v0.3 branch

kunmingg:


jlewi:
/test all
@kunmingg Now that we are cutting 0.4 does it still make sense to check this in or should we just close it?

Issue #421, opened 2018-03-14
Can not launch TensorFlow Serving because AVX not available on VM


ogre0403:
Hi all, 

I followed user guide try to [serve-a-model-using-tensorflow-serving](https://github.com/kubeflow/kubeflow/blob/master/user_guide.md#serve-a-model-using-tensorflow-serving),  but tf-serving pod can't be created. 
I setup kubeflow in my own kubernetes cluster, not using GKE or minikube. 
How can I fix this error ? Because there is no logs in Pod, I cannot show provide other useful information. 
Thanks. 

```
root@vagrant:~# kubectl get pod
NAME                              READY     STATUS             RESTARTS   AGE
ambassador-6ccb864c46-268xt       2/2       Running            0          10m
ambassador-6ccb864c46-9hrsh       2/2       Running            0          10m
ambassador-6ccb864c46-m7gc7       2/2       Running            0          10m
inception-858476d4c4-cr49s        0/1       CrashLoopBackOff   6          10m
tf-hub-0                          1/1       Running            0          10m
tf-job-operator-78757955b-nk457   1/1       Running            0          10m
```

```
Name:           inception-858476d4c4-cr49s
Namespace:      default
Node:           192.168.2.21/192.168.2.21
Start Time:     Wed, 14 Mar 2018 18:33:26 +0000
Labels:         app=inception
                pod-template-hash=4140328070
Annotations:    <none>
Status:         Running
IP:             172.20.0.152
Controlled By:  ReplicaSet/inception-858476d4c4
Containers:
  inception:
    Container ID:  docker://f59a9db5a3cf6f2e33671ba30ff71d3e34e32e1835ab908c4250f3c7f93f8c75
    Image:         gcr.io/kubeflow-images-staging/tf-model-server:v20180227-master
    Image ID:      docker-pullable://gcr.io/kubeflow-images-staging/tf-model-server@sha256:07ded66bc3a8e5ca582c3dc1871c2636a4ebc06b082e3afd76a68c45f1953385
    Port:          9000/TCP
    Args:
      /usr/bin/tensorflow_model_server
      --port=9000
      --model_name=inception
      --model_base_path=gs://kubeflow-models/inception
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       Error
      Exit Code:    132
      Started:      Wed, 14 Mar 2018 18:39:24 +0000
      Finished:     Wed, 14 Mar 2018 18:39:24 +0000
[...]

Issue #412, opened 2018-03-12
Executing child process 'start-notebook.sh' failed: 'Permission denied'


ernestmartinez:
I am trying to get Jupyterhub working. I get a 500 server error from the spawner. 

If I execute:
docker run  gcr.io/kubeflow/tensorflow-notebook-gpu 
[FATAL tini (7)] Executing child process 'start-notebook.sh' failed: 'Permission denied'

pdmack:
The release process is evolving so that image appears to be bad atm.
However,
`docker run gcr.io/kubeflow-images-staging/tensorflow-notebook-gpu` launches a notebook fine for me.
`[I 12:50:30.240 NotebookApp] The Jupyter Notebook is running at: ...` 

Please give that one a try.

ernestmartinez:
Thanks I'll give it a try.

jlewi:
For CPU you can use:
gcr.io/kubeflow-images-staging/tensorflow-notebook-cpu

jlewi:
Closing this issue because it looks like the problem was outdated images.

ernestmartinez:
The images are still failing with:
Failed to load the native TensorFlow runtime

PR #2059, opened 2018-12-07
fixes 'add a jobs component'

### LINKED ISSUE #2058: add a jobs component
The component kubeflow/jobs will include different prototypes to run common types of jobs. The single-job prototype takes a jobImage, jobName, jobCommand and jobArgs.

Other types of jobs will be added, like a job sequence

kkasravi:
fixes #2058

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2059)


kkasravi:
@jlewi this is being used to run a mnist training job using tensorflow with ngraph enabled that then shows the output in tensorboard. see [#412](https://github.com/kubeflow/examples/issues/412)

Although I can conceive of other types of prototypes in addition `single-job` like parallel jobs serial jobs, etc. I think these could be added over time.

jlewi:
@kkasravi So why not just make this part of the example in #421? Why create a reusable Jobs package?

Complex examples (e.g. code search) will often need to create a bunch of K8s resources. So we can just create manifests suitable for the example and parameterize them as needed to make it easier for people to run the example.

kkasravi:
@jlewi - yes I'll move this into the example and close this request

PR #2061, opened 2018-12-07
Fix a version number bug and add tf-batch-prediction into kubeflow registry

yixinshi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2061)


kubeflow/registry.yaml
```diff
@@ -70,3 +70,6 @@ libraries:
   kubebench:
     version: master
     path: kubebench
+  tf-batch-predict:
+    version: tf-batch-predict
```
lluunn: version should be master?

yixinshi: Good catch! Done.

PR #2062, opened 2018-12-07
use service account token to make request, set token life time to be 15 mins.

kunmingg:
* Using a service account is more secure because by using a new service account we can scope its access to only the resources needed to deploy the app.

* This also fixes the quota issues caused by requests made by the front end being charged to the GCP deployment project as opposed to the user's project.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2062)


kunmingg:
/hold
some APIs not ready.

kunmingg:
/hold cancel
should be good now

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -380,17 +393,68 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+        return undefined;
+      });
+
+    if (this.state.dialogTitle) {
+      return;
+    }
+
+    this._appendLine('Proceeding with project number: ' + projectNumber);
+
+    const deploymentName = this.state.deploymentName;
```
jlewi: Instead of creating a unique service account for each Kubeflow deployment should we just create a single service account and reuse it?

So the service account could be named something like
kubeflow-click-to-deploy@....

kunmingg: sounds good.

kunmingg:
/hold
change on backend side coming, want to make it in same PR.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -380,17 +393,68 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
+      await Gapi.iam.createServiceAccount(project, accountId);
+    }
+
+    const currProjPolicy = await Gapi.cloudresourcemanager.getIamPolicy(project);
+    const bindingKey = 'bindings';
+    currProjPolicy[bindingKey].push({
+      'members': ['serviceAccount:' + saEmail],
+      'role': 'roles/owner'
+    });
+    await Gapi.cloudresourcemanager.setIamPolicy(project, currProjPolicy);
```
yebrahim: We should probably add a `.catch` here to use the dialog to show the error properly, since we don't have a try/catch in parent scope. Same for all service calls, at least the ones that we think can fail, such as this one, and the `setServiceAccountIAM` call below.

yebrahim:
@jlewi sure thing, taking a look now.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -510,15 +560,29 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
   /**
    * Returns a list of services that are needed but not enabled for the given project.
    */
-  private async _getServicesToEnable(project: string) {
-    const enabledServices = await Gapi.servicemanagement.list(project);
+  private async _getServicesToEnable(project: string, token: string) {
+    // const enabledServices = await Gapi.servicemanagement.list(project);
+    const consumerId = encodeURIComponent(`project:${project}`);
+    const enabledServices = await rp(
```
yebrahim: What's the difference between this new code and the old `servicemanagement` object?
It's also good to keep all google APIs contained in the Gapi module, why not make these changes in there?

components/gcp-click-to-deploy/src/Gapi.ts
```diff
@@ -47,55 +36,116 @@ export default class Gapi {
...
+        });
+    }
+
+    public static async setIamPolicy(projectId: string, policy: object) {
       await Gapi.load();
-      const consumerId = `project:${project}`;
       return gapi.client.request({
         body: {
-          consumerId
+          'policy': policy
```
yebrahim: nit: use shorthand: `body: { policy }`

yebrahim: Same on lines 92 and 120.

kunmingg:
/hold cancel
Reuse params for prober test so avoid backend change.
Tested on new & existing projects, looks good.

jlewi:
Maybe update the docs here
https://github.com/kubeflow/kubeflow/blob/d72ba77aca9ad35a1fed98a2fb869e7bda83c2a3/components/gcp-click-to-deploy/src/App.tsx#L149

to explain that a service account is created and we delegate a credential for it?

Otherwise looks good to me.

yebrahim:
While updating docs, it'd be great if you can fix the Development section of this component's README. It's making an outdated reference to app-config.yaml file (why do developers have to copy that from anywhere, why not reference it directly in source code?)

components/gcp-click-to-deploy/app-config.yaml
```diff
@@ -0,0 +1,57 @@
+# Just for local test
```
yebrahim: Why not put this file directly under src/user_config as required? Why does the dev have to copy it before starting the app?

kunmingg: When run in k8s cluster, user need to provide ```app-config.yaml``` through config map.
By not creating user_config folder, if there's a bug or something caused config map failed mounting to container, user will notice their app doesn't work, instead of the app still running, with test/dev config.

kunmingg:
rebase to resolve errors

kunmingg:
squash all commits for easy rebase

Issue #1407, opened 2018-08-23
Argo deploy test failed


jlewi:
Seeing a lot of argo test failures on presubmits
https://k8s-gubernator.appspot.com/build/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/1399/kubeflow-presubmit/3081/#deploy-kubeflow-deploy_argo-test-argo-deploy

jlewi:
Looks like the test has been there for 4 months; added in #777

jlewi:
Event logs indicate the workflow exists
https://pantheon.corp.google.com/logs/viewer?project=kubeflow-ci&minLogLevel=0&expandAll=false&timestamp=2018-08-23T01:29:59.374000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2018-08-23T00:31:46.206Z&dateRangeEnd=2018-08-23T01:31:46.206Z&interval=PT1H&resource=container%2Fcluster_name%2Fkfctl-93f8&advancedFilter=resource.labels.cluster_name%3D%22kfctl-52c4%22%0AlogName%3D%22projects%2Fkubeflow-ci%2Flogs%2Fevents%22%20%0AjsonPayload.involvedObject.name%3D%22hello-world-sjlkj%22%0A&scrollTimestamp=2018-08-23T01:04:55.000000000Z

jlewi:
I think a better solution would be 

1. Have the python file test_deploy.py create the YAMl spec for the workflow so its always the same and the name is deterministic
1. Use https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/argo_client.py to wait for the workflow to complete successfully rather than polling pods.

PR #2067, opened 2018-12-10
Enforce use string type param for S3_USE_HTTPS and S3_VERIFY_SSL

Jeffwan:
By default, if we just enable s3 for model storage, we get string type from default template. 

```
ks param set ${MODEL_COMPONENT} s3Enable True

ks show default -c mnist-v3

...
        - name: S3_REGION
          value: us-west-1
        - name: S3_USE_HTTPS
          value: "true"
        - name: S3_VERIFY_SSL
          value: "true"
...

```

If we set true/false or 1/0 to these two parameters,  it will become boolean in libsonnet and kubernetes resource file. 

```
ks param set ${MODEL_COMPONENT} s3UseHttps true
ks param set ${MODEL_COMPONENT} s3VerifySsl true

# I tried using "true" and still got same result.

...
        - name: S3_REGION
          value: us-west-1
        - name: S3_USE_HTTPS
          value: true
        - name: S3_VERIFY_SSL
          value: true
...

```

In that case, since kubernetes environment require `String` value. deployment creation will fail as following log shows. 

I think if we enforce use string type in `tf-serving-aws.jsonnet` and we can get rid of this problem and make code more robust. User doesn't have to debug these kind of issues. 

```

➜  ks_app ks apply default -c mnist-v3
INFO Applying services kubeflow.mnist-v3
INFO Creating non-existent services kubeflow.mnist-v3
INFO Applying deployments kubeflow.mnist-v3
INFO Creating non-existent deployments kubeflow.mnist-v3
ERROR handle object: creating object: creating object: Deployment in version "v1beta1" cannot be handled as a Deployment: v1beta1.Deployment.Spec: v1beta1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found t, error found in #10 byte of ...|,"value":true},{"nam|..., bigger context ...|lue":"us-west-1"},{"name":"S3_USE_HTTPS","value":true},{"name":"S3_VERIFY_SSL","value":true},{"name"|...

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2067)

### DIFF
--- a/kubeflow/tf-serving/prototypes/tf-serving-aws.jsonnet
+++ b/kubeflow/tf-serving/prototypes/tf-serving-aws.jsonnet
@@ -47,8 +47,8 @@ local tfDeployment = base.tfDeployment +
                                    valueFrom: { secretKeyRef: { name: params.s3SecretName, key: params.s3SecretSecretaccesskeyKeyName } },
                                  },
                                  { name: "AWS_REGION", value: params.s3AwsRegion },
-                                 { name: "S3_USE_HTTPS", value: params.s3UseHttps },
-                                 { name: "S3_VERIFY_SSL", value: params.s3VerifySsl },
+                                 { name: "S3_USE_HTTPS", value: std.toString(params.s3UseHttps) },
+                                 { name: "S3_VERIFY_SSL", value: std.toString(params.s3VerifySsl) },
                                  { name: "S3_ENDPOINT", value: params.s3Endpoint },
                                ]
                              ) else [],


Jeffwan:
@ellis-bigelow @pdmack  do you have a chance to look at this issue?

Jeffwan:
@jlewi This is actually a breaking change.. Could we add to 0.4 release?

jlewi:
/lgtm
/approve

@Jeffwan Could you file a follow on issue to add a jsonnet unittest?

Jeffwan:
@jlewi Definitely, here it is #2067, I will try to add more unit tests on those components.

Jeffwan:
Hi @jlewi @r2d4  experienced some failures.  is this also related to #1407?

Jeffwan:
@jlewi @r2d4 looks like this one can not be automatically merged? Could you help merge this change?

Issue #1419, opened 2018-08-24
[kfctl] Create kfctl CLI 


kunmingg:
Should cover same user experience as https://github.com/kubeflow/kubeflow/blob/master/scripts/kfctl.sh

Can use https://github.com/spf13/cobra

Available APIs here:
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/cmd/bootstrap/app/ksServer.go

TheJaySmith:
Hello,

Go works too. Either way, do we have someone taking the action item?

kunmingg:
No, would be great if you have time for it?

TheJaySmith:
Sure. I am working on a CLI for some of the WeaveWorks Flux integration and it doesn't make sense to have two CLI's within the same product. My GoLang is rusty but what better way to improve it.

kkasravi:
@TheJaySmith take a look at [arena](https://github.com/kubeflow/arena/tree/master/cmd/arena/commands) which uses spf13/cobra and is a kubeflow CLI with a number of commands. You could possibly add the kfctl sub-commands there. I think at some point these CLI's would be combined (kfctl and arena).
@kunmingg @jlewi feel free to weigh in.

I'm doing some related work and may ping you on this stuff.

swiftdiaries:
I'd like to take this up if someone isn't already working on it. Also, would appreciate some direction as to where it would go? Part of arena or in kubeflow/components?

jlewi:
The go binary is here:
https://github.com/kubeflow/kubeflow/tree/master/bootstrap/

This is where we should add additional logic to make the binary a replacement for kfctl.sh.

@swiftdiaries I think a good place to start would be to just go through the commands that we execute with kfctl.sh and add entry points any missing logic; e.g.

```
kfctl init ...
kfctal generate platform ...
kfctl generate k8s ...
kfctl apply platform ...
kfctl apply k8s ...
```

jlewi:
@TheJaySmith if your working on this can you please clarify what exactly you are working on and when you plan on delivering it?

chrisheecho:
/assign @TheJaySmith 
Jay it seems like you are working on it? just want to make it official

TheJaySmith:
@chrisheecho yes. I have been a tad delayed but plan on having a deliverable by mid next week. Keep me to it

jlewi:
@TheJaySmith are you still working on this? I think @ashahba and @kkasravi expressed interest in working on this?

[...]

PR #2091, opened 2018-12-14
kfctl functional description and changes required

kkasravi:
Related to #1419 

kfctl functional description in README.md

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2091)

### DIFF
--- a/kfctl/README.md
+++ b/kfctl/README.md
@@ -1,0 +1,136 @@
+# kfctl golang client
+
+## Overview
+
+The kfctl golang client will provide a CLI similar to kfctl.sh but will 
+address the following deficits in the current tool
+
+1. No longer require a local install of ks (ksonnet)
+2. No longer require knowledge of the ks CLI 
+3. No longer save a ksonnet application to the user's filesystem
+4. Work in a similar way as the kubeflow UI [gcp-click-to-deploy](https://github.com/kubeflow/kubeflow/tree/master/components/gcp-click-to-deploy):
+  - Execution of subcommands `init`, `generate`, `apply` and `delete` are done by the bootstrapper.
+  - A local `<kf_app>.yaml` defines a kubeflow application. 
+  - This yaml file is submitted to the bootstrapper which will then generate the ksonnet application.
+
+## Usage
+
+The initial version of kfctl will seek parity with kfctl.sh by implementing the following subcommands:
+- `init`            Initialize a kubeflow application.
+- `generate`        Generate the k8 manifests of the kubeflow application.
+- `apply`           Submit the k8 manifests to the api-server
+- `delete`          Delete the kubeflow application
+
+The current usage of `kfctl.sh` is as follows:
+
+```sh
+kfctl.sh init myapp --platform generatic
+cd myapp
+kfctl.sh generate all
+kfctl.sh apply all
+```
+
+kfctl (golang) usage is similar:
+
+```sh
+kfctl init myapp 
+# edit myapp.yaml to set the appAddress and optionally add components and/or parameters
+kfctl generate --file myapp.yaml all
+kfctl apply --name myapp all
+```
+
+### Differences with kfctl.sh
+
+The `init` subcommand will create a `<name>.yaml` file that is used as input to the `generate` subcommand.
+The user modifies `<name>.yaml` to set the address of the bootstrapper service and optionally add 
+additional components and/or parameters. The `<name>.yaml` will use a similar field structure as what 
+the UI uses in [kf_app.yaml](https://github.com/kubeflow/kubeflow/blob/master/components/gcp-click-to-deploy/manifest/kf_app.yaml). This type will have a golang definition that is a kubernetes kind within the group `app.kubeflow.org`. 
+
+```go
+
+type Application struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty" 
+
+	AppAddress string     `json:"appaddress,omitempty"`
+	DefaultApp DefaultApp `json:"defaultapp:omitempty"`
+}
+
+type DefaultApp struct {
+	Components []KsComponent `json:"components,omitempty"`
+	Parameters []KsParameter `json:"parameters,omitempty"`
+	Registries []KsRegistry  `json:"registries,omitempty"`
+}
+
+type KsComponent struct {
+	Name      string `json:"name"`
+	Prototype string `json:"prototype"`
+}
+
+type KsParameter struct {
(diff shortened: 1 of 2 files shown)


kfctl/README.md
```diff
@@ -0,0 +1,140 @@
+# kfctl golang client
+
+## Overview
+
+The kfctl golang client will provide a CLI similar to kfctl.sh but will 
+address the following deficits in the current tool
+
+1. No longer require a local install of ks (ksonnet)
```
jlewi: Can you explain this goal? Do you intend to duplicate the functionality of ks? 
The intent with kfctl.sh was that if you wanted to customize individual components you would have to go into the ksonnet app directory?

kfctl/README.md
```diff
@@ -0,0 +1,140 @@
...
+2. No longer require knowledge of the ks CLI 
+3. No longer save a ksonnet application to the user's filesystem
```
jlewi: Why don't you want to save the ksonnet application to the user's filesystem?

kfctl/README.md
```diff
@@ -0,0 +1,140 @@
...
+4. Work in a similar way as the kubeflow UI [gcp-click-to-deploy](https://github.com/kubeflow/kubeflow/tree/master/components/gcp-click-to-deploy):
+  - Execution of subcommands `init`, `generate`, `apply` and `delete` are done by the bootstrapper.
```
jlewi: Do you mean they aren't running locally? Why aren't they running locally?

kfctl/README.md
```diff
@@ -0,0 +1,140 @@
...
+cd myapp
+kfctl.sh generate all
+kfctl.sh apply all
+```
+
+kfctl (golang) usage is similar:
+
+```sh
+kfctl init myapp 
+# edit myapp.yaml to set the appAddress and optionally add components and/or parameters
```
jlewi: Why are you introducing a myapp.yaml file here?

I thought the whole purpose of the CLI was to pretty much keep the existing semantics of kfctl.sh but just reuse the go library.

kfctl/README.md
```diff
@@ -0,0 +1,325 @@
...
+
+  Calls KsService.Apply //RELEVANT
+```
+
+- Note: Where there is a RELEVENT comment indicates where we need to insert a shared Interface that kfctl can use.
+- Note: kApp refers to the ksonnet interface.
+
+## Proposed Design
+
+### 1. Move the common interface and types required for the library to a pkg directory under bootstrap
```
jlewi: Should we define an interface for a Kubeflow App that would wrap ksonnet?

e.g

interface KsApp {
   generateK8s
  generateK8s
   applyK8s
   applyPlatform
}

This struct is what would be used by the CLI and webapp?

I guess an open question is how we would extend it to allow implementations for different platforms?

kkasravi: > Should we define an interface for a Kubeflow App that would wrap ksonnet?
> 
> e.g
> 
> interface KsApp {
> generateK8s
> generateK8s
> applyK8s
> applyPlatform
> }
> 
> This struct is what would be used by the CLI and webapp?
> 
> I guess an open question is how we would extend it to allow implementations for different platforms?

Yes I'll define a struct or interface. I'll post here when the implementation is slightly farther along (a day or so hopefully)

jlewi: SGTM. You could also just submit a skeleton PR with just the interface or struct but no implementation first.

jlewi:
In terms of directory structure should we put the readme in the bootstrap directory and then plan on renaming the bootstrap to kfctl later on?

Or does it make sense to put the readme in the kfctl directory for now and just move the code later?

kfctl/README.md
kkasravi: +1 on getting the skeleton PR in early for review.

kkasravi:
Maybe the former - bootstrap would look like:

bootstrap/cmd/kfctl
bootstrap/cmd/kfctl/README.md
bootstrap/cmd/kfctl/main.go
bootstrap/cmd/kfctl/cmd/{apply.go,generate.go, ...}
bootstrap/pkg/apis/apps/v1alpha1/kfapp_types.go
bootstrap/pkg/client/kfapi/typed/apps/v1alpha1/kfapp.go
bootstrap/pkg/server/main.go
bootstrap/pkg/server/app/{server.go,saKey.go, ...}
bootstrap/pkg/server/app/options/options.go

This is similar to how kubebuilder lays things out for CRDs (though it would have a controller subdir instead of server)

jlewi:
That seems good to me.

nit: wouldn't server/main.go be under cmd? e.g. 

bootstrap/cmd/kfctlWeb/main.go

We should plan on

1. Renaming "bootstrap" to kfctl

   * I think the name bootstrap is confusing so we should get it rid of it when possible

1. Probably moving it into its own repo at some point
   
    * At which point maybe the toplevel "bootstrap" directory is unnecessary?

I'd like the code for the webserver to be in a separate directory so that we can use OWNERs files to narrowly restrict the approver list. The directory structure above supports this since all the webapp code would be in the server directory and we can configure OWNERs files to not inherit parents see [here](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#owners). 

One of the things we need to ensure through code review is that there isn't any cross talk between requests; that could leak information. So we want to use a pattern like

```
func requestHandler() {
  app := NewKubeflowApp(...)
  defer app.Delete
  app.Generate()
}
```

This way its easy to check by inspection that we are being defensive in avoiding potential cross thread talking.


@kunmingg thoughts?

kkasravi:
> nit: wouldn't server/main.go be under cmd? e.g.
> 
> bootstrap/cmd/kfctlWeb/main.go

yes, I think that makes more sense.

> I'd like the code for the webserver to be in a separate directory so that we can use OWNERs files to narrowly restrict the approver list. The directory structure above supports this since all the webapp code would be in the server directory and we can configure OWNERs files to not inherit parents see [here](https://github.com/kubernetes/community/blob/master/contributors/guide/owners.md#owners).

+1

jlewi:
Thanks @kkasravi 
Should I LGTM this as is or there are additional changes you want to make first?

kunmingg:
@jlewi 
Create https://github.com/kubeflow/kubeflow/issues/2152 to track container-level isolation per request.

kkasravi:
@jlewi i think we're at a good point to close this PR. 
Additional issues can be opened that follow from this - like what @kunmingg is doing.

jlewi:
@kkasravi By close do you mean LGTM and merge or close this without merging?  (I usually interpret close as don't merge). Why wouldn't we merge in the README seems useful to me?

PR #2100, opened 2018-12-17
ks init: --skip-default-registries + append ${KS_INIT_EXTRA_ARGS}

### LINKED ISSUE #1615: On Premises deployment of Kubeflow fails unless ks flags are set v0.3
Please reference this [gist](https://gist.github.com/jtfogarty/511c2412466ec086f61046c6d554528f)

ksonnet is not picking up the http_proxy values or is expecting something other than what we have set.

Unless the changes detailed in the gist are made, the install process fails.  

```
 ks version
ksonnet version: 0.12.0
jsonnet version: v0.11.2
client-go version: kubernetes-1.10.4
```
Can the ks flags be added to the env.sh file and used in kfctl.sh / util.sh?

doodlesbykumbi:
## What are the goals of this PR ?

Fix  #1615

## How are the goals of the PR achieved ?

+ Avoid downloading from default registries, add --skip-default-registries to the ks init command here
https://github.com/kubeflow/kubeflow/blob/master/scripts/util.sh#L49

+ Define an extra environment variable ${KS_INIT_EXTRA_ARGS} to contain additional arguments to pass to the ks init command so that folks can specify additional arguments if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2100)

### DIFF
--- a/scripts/util.sh
+++ b/scripts/util.sh
@@ -46,7 +46,7 @@ createKsApp() {
   pushd .
   # Create the ksonnet app
   cd $(dirname "${KUBEFLOW_KS_DIR}")
-  ks init $(basename "${KUBEFLOW_KS_DIR}")
+  eval ks init $(basename "${KUBEFLOW_KS_DIR}") --skip-default-registries ${KS_INIT_EXTRA_ARGS}
   cd "${KUBEFLOW_KS_DIR}"
 
   # Remove the default environment; The cluster might not exist yet


jlewi:
Assuming tests pass
/lgtm
/approve

doodlesbykumbi:
@jlewi i'm wondering where is appropriate to document these changes ?

PR #2113, opened 2018-12-18
Update tf-serving.libsonnet for replicas

Youki:
Added replicas parameters in tf-serving.libsonnet so that multiple tf-serving pods can be deployed.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2113)

### DIFF
--- a/kubeflow/tf-serving/tf-serving.libsonnet
+++ b/kubeflow/tf-serving/tf-serving.libsonnet
@@ -5,6 +5,7 @@
   params:: {
     name: null,
     numGpus: 0,
+    replicas: 1,
     labels: {
       app: $.params.name,
     },
@@ -214,6 +215,7 @@
       },
       spec: {
         template: {
+          replicas: $.params.replicas,
           metadata: $.parts.tfServingMetadata,
           spec: {
             containers: [


jlewi:
/ok-to-test
/lgtm
/approve
@Youki sorry for the delay. If your reviewers aren't being responsive kubeflow.slack.com is a good place to call attention to it.

kubeflow/tf-serving/tf-serving.libsonnet
```diff
@@ -214,6 +215,7 @@
       },
       spec: {
         template: {
+          replicas: $.params.replicas,
```
nareshganesan: the 'replicas' field is part of DeploymentSpec and not the PodTemplateSpec?

nareshganesan:
Hi There,

I think there is a bug in the PR. the 'replicas' field is part of DeploymentSpec and not the PodTemplateSpec?

Please let me know.
Thanks!

Youki:
@nareshganesan 

Thanks. I will fix this bug in another PR, and paste the link to this conversation later.

nareshganesan:
@Youki , Thanks :thumbsup:

Youki:
@nareshganesan As shown in #2248, tf-serving.libsonnet will be replaced by newer version. So I won't fix the bug you pointed out, but newer version will not have the bug.

PR #2123, opened 2018-12-18
Create cloud sql database for ml pipeline 

IronPan:
Pipeline would use cloud sql for GCP scenarios. This change would give user the option to create a cloud sql instance, database and a root user as part of deployment.

The sql instance will have private IP enabled 
https://github.com/kubeflow/kubeflow/pull/2123/files#diff-a9c1d9db18e6d68f99216458ba3ab83fR203
so it can be accessed by GKE cluster by private IP. 
https://cloud.google.com/sql/docs/mysql/connect-kubernetes-engine#private-ip

This feature is currently disabled by a flag. 
https://github.com/kubeflow/kubeflow/pull/2123/files#diff-a9c1d9db18e6d68f99216458ba3ab83fR178
Will turn it on when pipeline fully integrate with cloud sql

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2123)


IronPan:
done. add todo

PR #2167, opened 2018-12-23
Add log timestamps, and do not wait for IAP if skipped IAP

### LINKED ISSUE #2165: Don't wait for IAP redirect when Skip IAP is selected
Even when Skip IAP is checked in the click-to-deploy app, the deployment app waits for the IAP url to become accessible.

### LINKED ISSUE #2166: Add timestamps in the log messages of the click to deploy app
Having timestamped log messages in the client is helpful specially when the same wait message is being printed and there is no way to tell in the log window if the app is progressing.

abhi-g:
fixes #2166 and fixes #2165

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2167)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -618,9 +618,9 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
               this._appendLine('your kubeflow app url should be ready within 20 minutes (by '
                 + readyTime.toLocaleTimeString() + '): https://'
                 + this.state.deploymentName + '.endpoints.' + this.state.project + '.cloud.goog');
-              clearInterval(monitorInterval);
               this._redirectToKFDashboard(dashboardUri);
             }
+            clearInterval(monitorInterval);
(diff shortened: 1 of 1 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -609,9 +618,9 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
               this._appendLine('your kubeflow app url should be ready within 20 minutes (by '
                 + readyTime.toLocaleTimeString() + '): https://'
                 + this.state.deploymentName + '.endpoints.' + this.state.project + '.cloud.goog');
+              clearInterval(monitorInterval);
+              this._redirectToKFDashboard(dashboardUri);
             }
-            clearInterval(monitorInterval);
```
kunmingg: We might want to do clearInterval() for both cases (skip or not skip).

abhi-g: thanks. makes sense.

PR #2180, opened 2018-12-27
Add the TF 1.11 & TF 1.12 Jupyter images to the spawner drop down box.

jlewi:
* Upgrade the images in the Arrikto ROK UI as well; it looks as though
  they hadn't been updated to the 0.4 images.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2180)


ioandr:
Thank you very much for this PR Jeremy - it was an oversight from our side not to update the TF images of the Arrikto Rok UI to the latest versions.

Just a nitpick on the PR: you might want to set the default value of field `Image` not to `1.10.0`, but to `1.12.0`, which is the latest one.

/lgtm

jlewi:
Done.
@ioandr PTAL.

Issue #603, opened 2018-04-06
Add auto-generated TOC for all docs


DjangoPeng:
I noticed that there is no completed TOC in our docs. It's unfriendly for readers to locate the section which they want.

I'd like to add TOC for all docs by auto-generation tool [doctoc](https://github.com/thlorenz/doctoc), which has been used in many repos.

/assign @DjangoPeng
/cc @jlewi

PR #2199, opened 2019-01-03
Add retry to loading the sample job

IronPan:
Loading the sample job has sporadically failures due to race condition with api server start up. They both can create the db foreign key. 

The gorm library handles foreign key existence but it won't be safe in race condition
https://github.com/jinzhu/gorm/blob/master/scope.go#L1221

This is a temporary fix to allow samples load correctly. In long term, the sample loading shouldn't be a job because job doesn't work well with ksonnet. It won't work during upgrade scenario. 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2199)

### DIFF
--- a/kubeflow/pipeline/pipeline-apiserver.libsonnet
+++ b/kubeflow/pipeline/pipeline-apiserver.libsonnet
@@ -216,7 +216,7 @@
             serviceAccountName: "ml-pipeline",
           },
         },
-        backoffLimit: 0,
+        backoffLimit: 2,
       },
     },  // loadSampleJob
 


IronPan:
Fix https://github.com/kubeflow/pipelines/issues/603

kubeflow/pipeline/pipeline-apiserver.libsonnet
```diff
@@ -216,7 +216,7 @@
             serviceAccountName: "ml-pipeline",
           },
         },
-        backoffLimit: 0,
+        backoffLimit: 2,
```
gaoning777: Why does the load sample require retries?
BTW, https://github.com/kubeflow/pipelines/issues/610 mentions that there are errors in ml-pipelines-load-samples.

jlewi: @IronPan Can you expand on what the problem is and how this fixes it?

I see two different problems
#603 problem creating the foreign key
#610 connection timeout.

If this is a retryable error; should the retry logic be in the go code as opposed to the K8s job? 

I think in other places in Kubeflow we've used this go library to implement exponential backoff.
https://github.com/cenkalti/backoff

IronPan: This is a temporary solution to resolve the foreign key issue that doesn't require a new pipeline release and can quickly unblock kf release. I'll work on a proper fix that doesn't use job to load samples. 

Add more details to the description

Issue #2209, opened 2019-01-04
Kfctl requires zone to be specified. Could be potentially automatically inferred.


abhi-g:
kfctl init requires that zone be specified manually with the command. It would be helpful to infer it automatically and yet allow the user to override it with the flag if needed.

jlewi:
@abhi-g Is there more work to be done here or can we close this? Do we need to cherry pick the fix onto the 0.4 branch?

abhi-g:
We can close this. This has already been cherrypicked into 0.4

Abhishek

PR #2210, opened 2019-01-04
kfctl: add default zone if not specified

r2d4:
/assign @abhi-g 

I'll send a follow up cherry-pick to v0.4-branch

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2210)

### DIFF
--- a/scripts/kfctl.sh
+++ b/scripts/kfctl.sh
@@ -19,6 +19,9 @@ SKIP_INIT_PROJECT=false
 # logging agents.
 GKE_API_VERSION="v1beta1"
 
+# Default GCP zone to deploy Kubeflow cluster if not specified
+GCP_DEFAULT_ZONE="us-east1-d"
+
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
 source "${DIR}/util.sh"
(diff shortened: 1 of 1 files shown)


scripts/kfctl.sh
```diff
@@ -211,9 +211,8 @@ parseArgs() {
     if [ -z "$ZONE" ]; then
       ZONE=$(gcloud config get-value compute/zone 2>/dev/null)
       if [ -z "$ZONE" ]; then
-        echo "GCP zone must be set either using --zone <ZONE>"
-        echo "or by setting a default zone in gcloud config"
-        exit 1
+        # Set default zone to us-east1-d if not specified
+        ZONE="us-east1-d"
```
abhi-g: Also echo a message suggesting that we are setting the default to us-east1d. but can be overriden using gcloud config or --zone

abhi-g: essentially the previous echo statements could be brought back though not the exit 1 :)

r2d4: Yep - made it a bit cleaner.

abhi-g:
fixes #2209 
/lgtm
/approve

PR #2218, opened 2019-01-04
simplify cloud shell; add instruction message

kunmingg:
Avoid dependency on source repo:
currently for multi deploy in same project, user will be asked whether or not ```git pull``` in cloud shell, cause confusions.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2218)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -195,7 +195,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
         </div>
 
         <Collapse in={!this.state.iap}>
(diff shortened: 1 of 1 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -194,6 +194,10 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
           </TextField>
         </div>
 
+        <Collapse in={!this.state.iap}>
+          <div style={styles.row}>Kubeflow UI Access: after Deployment done, click "Cloud Shell" and click "port forwarding" in new page.</div>
```
abhi-g: Add "is" I.e. Deployment is done

PR #2237, opened 2019-01-08
Fix katib repo for the docker registry which doesn't support sub namespace

cheyang:
The katib images like `gcr.io/kubeflow-images-public/katib/tfevent-metrics-collector:v0.4.0` uses sub namespace `kubeflow-images-public/katib` in `gcr.io`, but it's not supported by other docker image repo. We need to consider how to support it in other docker repos.

We will use the environment variable `DOCKER_REGISTRY_KATIB_NAMESPACE ` to replace `kubeflow-images-public/katib` if the variable is set.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2237)


cheyang:
@jlewi , Please also help to cherry pick this PR to v0.4-branch. Thanks.

PR #2252, opened 2019-01-09
Update Pipeline version to v0.1.7

IronPan:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2252)


IronPan:
@kunmingg Is the tests flaky or my change is actually bad? It's hard to tell from the log but I don't think it's caused by this change

Issue #1907, opened 2018-11-05
Investigate using Istio for jwt validation


lluunn:
related to #1309

lluunn:
notebook: #1994

lluunn:
Current architecture:

```
Ingress -->        Envoy        --> Ambassador --> Other services
              (JWT valdation)
```

The plan is to use istio-ingressgateway for both JWT validation and routing, and thus:
```
Ingress  --> istio-ingressgateway --> Other services
```

However, ingress is in kubeflow namespace, and istio-ingressgateway is in istio-system namespace, so currently it's not working. https://github.com/kubernetes/kubernetes/issues/17088

Question: should we move ingress to istio-system namespace, or move istio-ingressgateway to kubeflow namespace? Both involve not just service, but also things like secret, service account etc.

I think we can have a kubeflow-system namespace for server side resources, deployed at deployment time. These resources are rarely changed, and should only by cluster admin. We can move ingress/iap/istio core stuffs to this namespace.

Does it make sense @jlewi ?

lluunn:
it's probably easier to install istio in kubeflow namespace since the other is a breaking change.

lluunn:
covered by #2261

Issue #2052, opened 2018-12-06
[Click-to-deploy] install istio


lluunn:
See https://github.com/kubeflow/website/pull/319.
It should be a simple kubectl apply.

@jlewi 
@kunmingg

lluunn:
Our istio manifest is yaml not ksonnet because
1. it's large
1. we need to update it when istio has a new version
1. user won't need to change it after deployed, so no params need to be set

Then how do we install it with bootstrapper and save it with other manifest? I can think of several ways.
1. In ks_app, do `ks import` to drop in the yaml. But it will create a component per k8s resource, so there will be thousands of files which is not acceptable(?).
2. Use some hack to make it a ks component, e.g. a k8s job that deploys it.
3. save it as yaml, but not part of ks app, just like deployment manager config.

I think No.3 is better.
@kunmingg likes No.2 because it's more consistent with other components.
@jlewi thought?

jlewi:
kfctl.sh already supports using raw YAML for some components. These are stored in the k8s_specs directory
https://www.kubeflow.org/docs/started/getting-started-gke/#app-layout

So does this work for ISTIO? Do we just need to add this support to click to deploy?

lluunn:
Yes, this is what I meant by No.3

kunmingg:
Can we apply through client library instead of running kubectl from bash?
Potential risk of bash kubectl:
Need to take care of credential isolation when running kubectl in parallel. We can not use default ```.kubeconfig```

lluunn:
If each deployment creates its config, and we use kubectl with `--kubeconfig`, is this acceptable?

jlewi:
Here's a link to some code using the REST client:
https://github.com/kubeflow/tf-operator/blob/e4a436da92e198dcb88c89c33010608e0c8a23bf/pkg/util/k8sutil/tf_job_client.go#L88

lluunn:
IIUC, to use REST client, you need to set the api group/version, and the resource type in the request.
https://godoc.org/k8s.io/client-go/rest#Request

api group/version can be parsed from yaml file for each resource.
However, the resource type is the plural of the "Kind", which is not available by only looking at the yaml (I don't think there is a reliable kind-->pluralKind converter)
So I am not sure if REST client will help.

I found that client-go has dynamic client which can operate on arbitrary resource.
However, it's discouraged by the team to use it.
https://github.com/kubernetes/client-go/issues/394

jlewi:
[...]

PR #2259, opened 2019-01-11
use istio 1.1

lluunn:
We need 1.1 for JWT's trigger rule (excluded_paths: /healthz)

The manifest are from Istio pre-release 1.1-20190111-09-15.
Changed:
- policy is disabled instead of enabled (see README for explanation)
- istio-ingressgateway type is NodePort not LoadBalancer

/cc @kunmingg 

related: #1907

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2259)

### DIFF
--- a/dependencies/istio/install/README.md
+++ b/dependencies/istio/install/README.md
@@ -1,6 +1,7 @@
 # Istio manifest for Kubeflow
 
-The manifest is taken from OSS Istio pre-release 1.1-20190111-09-15.
+The manifest is taken from OSS Istio pre-release
+[1.1-20190111-09-15](https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/release-1.1-20190111-09-15/).
 
 We need custom configuration as discussed in the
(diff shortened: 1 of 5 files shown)


dependencies/istio/install/README.md
```diff
@@ -1,6 +1,6 @@
 # Istio manifest for Kubeflow
 
-The manifest is taken from OSS Istio release 1.0.2.
+The manifest is taken from OSS Istio pre-release 1.1-20190111-09-15.
```
jlewi: Can you provide a link where these manifests are obtained?

jlewi:
The ISTIO install still needs to be integrated into kfctl.sh? Do we have an issue tracking that?

lluunn:
https://github.com/kubeflow/kubeflow/issues/2052 this is for installing istio with click-to-deploy

dependencies/istio/install/README.md
lluunn: done

Issue #2149, opened 2018-12-20
Allow in-cluster traffic through ambassador while enable basic username/password login with ambassador 


kunmingg:
See https://github.com/kubeflow/kubeflow/issues/2094 for longer story.
Need to allow in cluster traffic while enable basic auth.

By current design, basic auth accept username/password or cookie, but in-cluster traffic would have none of them.

Auth server should recognize in-cluster traffic and allow them.
This should be transparent to individual components.

kunmingg:
One easy solution is to have separate ambassador services:
1. external facing: 
* used by user request, from browser or API call
* every request validate by auth service
2. internal facing: 
* used by in-cluster traffic.
* only accessible within cluster
* doesn't go through auth service

Issue #2148, opened 2018-12-20
Cookie based auth for basic username/password login with ambassador


kunmingg:
See https://github.com/kubeflow/kubeflow/issues/2094 for longer story.
Need to have cookie based auth once user login.

1. cookie generated, recorded and assigned once login.
2. client side: once clear cookie, redirect to login page.
3. server side: once cookie expires (1 day), clear and redirect to login page.

Issue #2147, opened 2018-12-20
login page for basic username/password login with ambassador


kunmingg:
See https://github.com/kubeflow/kubeflow/issues/2094 for longer story.
Need to have login page for basic auth case.

frontend: login page.
backend: redirect unauth traffic to login page.

PR #2262, opened 2019-01-12
basic auth service

kunmingg:
Backend part.

Test passed, ready to merge.

fix https://github.com/kubeflow/kubeflow/issues/2147
fix https://github.com/kubeflow/kubeflow/issues/2148
fix https://github.com/kubeflow/kubeflow/issues/2149

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2262)


kunmingg:
/hold 
Will manual test when ready.

components/gatekeeper/auth/AuthServer.go
```diff
@@ -0,0 +1,187 @@
...
+		cookies: make(map[string]time.Time),
+	}
+	return server
+}
+
+// Default auth check service
+func (s *authServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	log.Infof("Path check, url: %v, path: %v", r.URL, r.URL.Path)
+	// login page open to everyone; all other path requires auth with Password or cookie
+	if strings.HasPrefix(r.URL.Path, "/" + LoginPagePath) || s.authCookie(r) == true {
```
jlewi: Should we be doing an exact match here to path
r.URL.Path, "/" + LoginPagePath

What if by some coincedence we have a path for which this is a prefix but is not an exact match?

kunmingg: Browser need access resources like "/LoginPagePath/static/..." as well.
How about do exact match "/LoginPagePath" or prefix match "/LoginPagePath/"?

kunmingg: Should be safe to assume any thing behind "/LoginPagePath/" is public?

components/gatekeeper/auth/AuthServer.go
```diff
@@ -0,0 +1,187 @@
...
+		// Handle request from login page
+		if r.Header.Get(LoginPageHeader) != "" {
```
jlewi: Can you explain what this if block is doing? Is it modifying the request to indicate that it should be allowed?

components/gatekeeper/auth/AuthServer.go
```diff
@@ -0,0 +1,187 @@
...
+		username: opt.Username,
+		pwhash: string(data),
+		cookies: make(map[string]time.Time),
+	}
+	return server
+}
+
+// Default auth check service
+func (s *authServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	log.Infof("Path check, url: %v, path: %v", r.URL, r.URL.Path)
```
jlewi: Can we reject non https traffic?

We don't want user logging in if they aren't using https because that would send username/password in the clear?

components/gatekeeper/auth/AuthServer.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return false
+	}
+
+	namepw := strings.Split(string(upBytes), ":")
+
+	if len(namepw) != 2 {
+		return false
+	}
+	err = bcrypt.CompareHashAndPassword([]byte(s.pwhash), []byte(namepw[1]))
```
jlewi: So CompareHashAndPassword returns error if they don't match?

components/gatekeeper/auth/AuthServer.go
```diff
@@ -0,0 +1,187 @@
...
+
+		return true
+	}
+	return false
+}
+
+// auth with cookie
+func (s *authServer) authCookie(r *http.Request) bool {
+	if cookie, err := r.Cookie(CookieName); err == nil {
+		if val, ok := s.cookies[cookie.Value]; ok {
```
jlewi: What prevents someone from spoofing a cookie?

components/gatekeeper/auth/AuthServer.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return string(b)
+}
+
+func (s *authServer) addNewCookieValue(cookieVal string) {
+	s.serverMux.Lock()
+	defer s.serverMux.Unlock()
+	// Cookie expire after 12 hours
+	log.Info("cookie set: set new cookie value!")
+	s.cookies[cookieVal] = time.Now().Add(12 * time.Hour)
```
jlewi: So the cookie is stored in memory, so if the pod is restarted the cookie will no longer be valid. Will we automatically redirect user to log back in?

components/gatekeeper/auth/AuthServer.go
kunmingg: Can we only access https at gclb?

jlewi:
Thanks this looks great.

components/gatekeeper/auth/AuthServer.go
kunmingg: This part take care of user's re-login
They already have cookie in browser, so "StatusResetContent" bring them to kubeflow central dashboard.

components/gatekeeper/auth/AuthServer.go
kunmingg: ```SameSite: http.SameSiteStrictMode```
We restrict the cookie so can only be used for same site.
Should we encrypt value as well?

components/gatekeeper/auth/AuthServer.go
jlewi: I don't think there's any way to force GCLB to only accept https traffic. You can look at what we do in click to deploy. Basically we'd want to check if the protocol is http and then redirect to https

components/gatekeeper/auth/AuthServer.go
jlewi: You mean with bcrypt? So we only keep track of the encrypted value?

jlewi:
Will LetsEncrypt be an issue? Do we need to allow for the ACME challenge?

On GCP one option would be to migrate to Managed Certificates.

components/gatekeeper/auth/AuthServer.go
kunmingg: Yes, server side only keep encrypted password hash.

jlewi: If its not too difficult I think that would be valuable; for the same reason its valuable with the password.

components/gatekeeper/auth/AuthServer.go
jlewi: Might be worth a P2 issue to fix in a follow up.

kubeflow/common/basic-auth.libsonnet
```diff
@@ -0,0 +1,175 @@
...
+    local loginDeployment = {
+      apiVersion: "extensions/v1beta1",
+      kind: "Deployment",
+      metadata: {
+        name: ui_name,
+        namespace: params.namespace,
+
+      },
+      spec: {
+        replicas: 1,
```
jlewi: Add a comment here explaining this should always be 1 because the cookie is stored in memory so if we use more than 1 replica we will have problems because not all replicas will have the cookie.

jlewi:
On second thought I don't think LetsEncrypt is an issue because it adds a separate path to the ingress.

I had one more comment about adding a comment about number of replicas; otherwise this is good to go.

kubeflow/common/basic-auth.libsonnet
kunmingg: Comments updated

PR #2285, opened 2019-01-15
adjust param name for v0.3 compatibility

kunmingg:
customize params for v0.3, can be removed after https://github.com/kubeflow/kubeflow/pull/2019 fully functional.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2285)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -144,7 +144,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
   public render() {
     const zoneList = ['us-central1-a', 'us-central1-c', 'us-east1-c', 'us-east1-d', 'us-west1-b',
       'europe-west1-b', 'europe-west1-d', 'asia-east1-a', 'asia-east1-b'];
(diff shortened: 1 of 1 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -144,7 +144,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
   public render() {
     const zoneList = ['us-central1-a', 'us-central1-c', 'us-east1-c', 'us-east1-d', 'us-west1-b',
       'europe-west1-b', 'europe-west1-d', 'asia-east1-a', 'asia-east1-b'];
-    const versionList = ['v0.3.5', 'v0.4.0'];
```
lluunn: why is 0.4.0 removed?

kunmingg: We'll overwrite versions through env var. https://github.com/kubeflow/kubeflow/blob/master/components/gcp-click-to-deploy/src/DeployForm.tsx#L191
List here is for test only, and v0.4.0's registry file is not up to date

PR #2294, opened 2019-01-17
fixes 'Create kfctl CLI'

### LINKED ISSUE #1419: [kfctl] Create kfctl CLI

kkasravi:
fixes #1419

See bootstrap/cmd/kfctl/README.md for details and remaining work.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2294)


jlewi:
Thank you @kkasravi @ashahba so much for working on this.

I left some specific comments but here are some high level comments.

* Are we missing a high level type `KfApp` that encapsulates all the data and methods for managing the Kubeflow application

  * Primary methods would be `generate` and `apply` and mirror kfctl syntax e.g.

     ```
     func(a *KfApp) Generate(w KfNouns) ...
     func(a *KfApp) Apply(w KfNouns) ...
     ```

* It looks like right now we have kfApi but it looks like this is primarily managing the ksonnet application and kfctl is currently doing more than that.

* In the current implementation it looks like the logic that belongs in KfApp is currently in the anonymous functions invoked by the cobra commands

   * I would expect the cobra commands to just be invoking the corresponding method on KfApp

*  How do vendors extend the implementation to customize it to their platform?

   * If we follow kfctl.sh then Generate and Apply could have conditionals which invoke specific handlers based on the platform.

/cc @kunmingg

kkasravi:
> I left some specific comments but here are some high level comments.
> 
> * Are we missing a high level type `KfApp` that encapsulates all the data and methods for managing the Kubeflow application
>   
>   * Primary methods would be `generate` and `apply` and mirror kfctl syntax e.g.
>     ```
>     func(a *KfApp) Generate(w KfNouns) ...
>     func(a *KfApp) Apply(w KfNouns) ...
>     ```

Yes, this makes sense. We'll provide this before Friday and update the README.md

kkasravi:
> * 
> * In the current implementation it looks like the logic that belongs in KfApp is currently in the anonymous functions invoked by the cobra commands
>   
>   * I would expect the cobra commands to just be invoking the corresponding method on KfApp
> * How do vendors extend the implementation to customize it to their platform?
>   
>   * If we follow kfctl.sh then Generate and Apply could have conditionals which invoke specific handlers based on the platform.

We'll take a pass on this to allow vendor hooks

jlewi:
Thanks @kkasravi

jlewi:
@kkasravi Is this ready for an in depth review? If it is please remove WIP to make it clear.

kkasravi:
Hi @jlewi yes I believe it's close enough for a deeper review.

- we still need to add command args where needed
- add <registry|pkg|component> is one command where the args are being added

kkasravi:
> Thanks @kkasravi @ashahba the effort you guys are putting in is amazing.
> 
> I left a a number of comments so let me provide some pointers
> 
> * Look at README.md and kfapi.go
> * I think there are some design issues that still need to be resolved and those are captured in the comments in README.md and kfapi.go
> * I'd really like to slim this PR down, focus on core design and API and leave additional functionality for next PRs
>   
>   * I'd like to get to parity with kfctl.sh; and do the extra commands (e.g. add, show, list, remove) in a follow on PR
>   * I suggest not modifying the click to deploy app in this PR (or making only superficial changes)
>     
>     * I think once we get the core go structs written it will be easier to refactor the REST server to use them
>     * Creating a go binary that has parity with kfctl.sh should make that easier
> 
> >

Thanks @jlewi - we may want to unwind this PR - create a new branch and new PR with just the changes you've suggested in the README.md and kfapi.go. EG your latest review has 33 comments,
I've reviewed them and the new PR will reflect this guidance.
From a high level - we will: 

1. Just push the changes to get kfctl to parity with kfctl.sh 
  - Separate implementations from the interface - eg KfApp is the interface, KsApp, GCPApp are the implementations. GCPApp will be in a follow-on PR.
  - The KfApp interface is just init, generate, apply, delete (with a KsApp implementation) 
  - Replace env vars used in kfctl.sh with command line args in kfctl.
2. Make little to no changes to click-to-deploy at this point in time (but do some due diligence to validate that its calls to ksonnet can be replaced with GCPApp
3. Leave out all half-baked subcommands and extraneous files like debug.sh

I think this would narrow this down to a minimal set of files that could be merged to master to replace kfctl.sh. Subsequent PR's would tackle gcp-click-to-deploy and possibly other cloud-providers but at least we would have a pattern other cloud providers could follow.

My goal would be to have this ready to review by tomorrow.

Thoughts?

kkasravi:
@jlewi 
BTW - did you want me to branch off of this branch to do the above (and use rebase)? I think the new set of changes would have lots of deletes, unwinding gcp-click-todeploy changes. It might be cleaner to create a new branch from master.

jlewi:
Thanks @kkasravi 

I strongly prefer reusing the existing PR and removing unnecessary changes. I'm a big believer in preserving the entire discussion history.

For the same reason I would avoid rebasing until its done and we need to rebase to squash commits etc...

jlewi:
LGTM to the plan you outlined in your last comment.

kkasravi:
@jlewi - got it - I'll create a new branch based off of the current branch.

jlewi:
Thanks @kkasravi I added wip back to the title. Just remove it when its ready for me to take another look.

kkasravi:
superceded by #2331 which builds on this PR

jlewi:
Thank you so much @kkasravi and @ashahba I think this is very close to being done. I just had a couple design questions.

kkasravi:
@jlewi 

I've removed the wip label. I think this can be merged. Note below:

> > E8S? What does the E stand for ? I think kfctl.sh uses k8s.

Fixed

> > Why does Generate take ResourceEnum but not the other methods?

Fixed

Your remaining suggestions/guidance have been opened as new issues:

- [Remove --packages and --components from 'kfctl generate' (golang)](https://github.com/kubeflow/kubeflow/issues/2365)
- [Move --namespace parameter from init to generate in kfctl (golang)](https://github.com/kubeflow/kubeflow/issues/2366)
- [Have the different kfctl (golang) implementations of KfApp hold a map of KfApp children](https://github.com/kubeflow/kubeflow/issues/2367)
- [Replace CfgFile *viper.Viper in KsApp with a struct and move to group.go in kfctl (golang)](https://github.com/kubeflow/kubeflow/issues/2368)

Do we need a new label for kfctl (golang)?

jlewi:
Thanks.

I think we can just use existing kfctl label for both but if you want a new label just submit a PR to:
https://github.com/kubeflow/testing/blob/master/label_sync/kubeflow_label.yml

jlewi:
Thanks @kkasravi. I had a couple minor comments. I think the only remaining question I have is why are you writing out an app.yaml file in ${KF_APP} that contains the ksonnet application info? Doesn't that duplicate the information stored in the ksonnet app directory?

kkasravi:
To follow-up - the first app.yaml is produced after `kfctl init ...`, the second app.yaml is produced after `kfctl generate ...`. The ability of the user to change the set of packages, components and parameters is between `kfctl init ...` and `kfctl generate ...`. After `kfctl generate ...` the ks_app/app.yaml holds libraries, registries and environments, which is somewhat different that kfctl's app.yaml which holds components, packages, parameters, platform.

bootstrap/pkg/apis/apps/ksapp/v1alpha1/application_types.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+var DefaultRegistry = &RegistryConfig{
+	Name: "kubeflow",
+	Repo: "https://github.com/kubeflow/kubeflow.git",
+	Path: "kubeflow",
+}
+
+var DefaultPackages = []string{
+	"application",
```
kunmingg: Shall we store default values in config file so that if web service backend can update config without have to rebuild go binary / docker images?

bootstrap/pkg/client/ksapp/ksapp.go
```diff
@@ -0,0 +1,557 @@
...
+	}
+	clientConfig, clientConfigErr := kftypes.GetClientConfig()
+	if clientConfigErr != nil {
+		return fmt.Errorf("couldn't load client config Error: %v", clientConfigErr)
+	}
+	applyErr := ksApp.applyComponent([]string{"metacontroller"}, clientConfig)
+	if applyErr != nil {
+		return fmt.Errorf("couldn't create metacontroller component Error: %v", applyErr)
+	}
+	applyErr = ksApp.applyComponent([]string{"application"}, clientConfig)
```
kunmingg: We might miss error message from application CRD controller if any components failed to be created.
Still feel more confident to apply each ks component and check its status.

kkasravi:
changes address just about all recent comments in this PR
- remove cli options --packages, --components
- move --namespace to Init
- remove CfgFile and pass cli options into Apply, Delete, Init, Generate
- do not write out components, packages, parameters to app.yaml since it's redundant with ksonnet's app.yam
- update README.md

jlewi:
/lgtm
/approve
/hold 
@kunmingg any final comments?

@kkasravi You'll need to amend the commit history to get the CLA check to pass. Right now you have commits from multiple authors; options are

* Rebase and squash all the commits
* Use git commit amend to amend the commit history and change the author.

lluunn:
cc @lluunn

bootstrap/pkg/apis/apps/ksapp/v1alpha1/application_types.go
```diff
@@ -0,0 +1,255 @@
...
+	copy(pkgs, defaults)
+	for _, name := range names {
+		pkgs = RemoveItem(pkgs, name)
+	}
+	return pkgs
+}
+
+var DefaultRegistry = &RegistryConfig{
+	Name: "kubeflow",
+	Repo: "https://github.com/kubeflow/kubeflow.git",
```
kunmingg: User might hit Github API rate limit if they use default: always pull from github.
Shall we bake a release tag registry into our binary?
https://github.com/gobuffalo/packr

Or maybe follow up PRs? @jlewi

kunmingg: Another option is keep current logic: curl and download release tag registry into tmp dir.
Pros: serve multi release tags with same go binary
Cons: doesn't work for environment without public network: on prem usage case

kkasravi: @kunmingg i like baking in the release version as well as having versions kfctl could download. I think it would be best if we maintain default components, packages, parameters per version somewhere on github rather than having it has golang vars . @jlewi @ashahba thoughts?

kunmingg: Definitely. we can have file like https://github.com/kubeflow/kubeflow/blob/master/components/gcp-click-to-deploy/app-config.yaml in github, shared as default to deploy.
And we can update the file as we cut new release so each release tag has its own components set.

PR #2304, opened 2019-01-18
increase the number of mktemp template variables in download.sh

tnthornton:
The call to mktemp in the download.sh script results in an error similar to the following when running the script from Alpine, in my case I was using busybox:

```
mktemp: Invalid argument
```

Which is easily reproducible outside of the download.sh script:
```
bash-4.3# mktemp -d /tmp/tmp.kubeflow-repo-XXXX
mktemp: Invalid argument
```

Once the number of template variables was increased to _at least_ 6 places, the error was resolved:
```
bash-4.3# mktemp -d /tmp/tmp.kubeflow-repo-XXXXXX
/tmp/tmp.kubeflow-repo-XckJMKj
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2304)

### DIFF
--- a/scripts/download.sh
+++ b/scripts/download.sh
@@ -12,7 +12,7 @@ fi
 KUBEFLOW_TAG=${KUBEFLOW_TAG:-master}
 
 # Create a local copy of the Kubeflow source repo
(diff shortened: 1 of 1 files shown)


jlewi:
Thank you so much for the fix
/ok-to-test

PR #2305, opened 2019-01-18
Added context check before issuing delete

### LINKED ISSUE #1669: kfctl.sh delete doesn't check/change kubectl context before issuing delete
I don't think kfctl.sh delete verifies that kubectl is pointing at the right cluster before it issues a delete.

We should add a check and verify that we are pointing at the same cluster as the ksonnet app before deleting the namespace.

Here's some code to try to do that in bash.
https://github.com/kubeflow/testing/blob/master/deployment/redeploy_app.sh#L84

svalleru:
- Fixes #1669 
- Syntax improvements

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2305)

### DIFF
--- a/scripts/kfctl.sh
+++ b/scripts/kfctl.sh
@@ -139,7 +139,7 @@ createEnv() {
       export EMAIL=${EMAIL}
 
       # TODO: Do we need to make PROJECT_NUMBER also a flag like --project-number
-      if [ -z "${PROJECT_NUMBER}" ]; then
+      if [[ -z "${PROJECT_NUMBER}" ]]; then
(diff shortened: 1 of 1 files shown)


jlewi:
Thank you so much for doing this!
/ok-to-test

jlewi:
/lgtm
/approve

I think this would be worth cherry-picking into the v0.4-branch so it will be included in the next minor release.

svalleru:
> /lgtm
> /approve
> 
> I think this would be worth cherry-picking into the v0.4-branch so it will be included in the next minor release.

@jlewi here's the PR with cherrypicked changes into v0.4 branch -- #2314

PR #2315, opened 2019-01-23
bootstrapper install istio

lluunn:
For #2052

Bootstrapper will install Istio before the ksapp because the app will depend on Istio (e.g. istio-ingressgateway)

/cc @kunmingg 
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2315)


lluunn:
/hold
testing

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -349,6 +350,39 @@ func (s *ksServer) GetProjectLock(project string) *sync.Mutex {
 	return s.projectLocks[project]
 }
 
+// InstallIstio installs istio into the cluster.
+func (s *ksServer) InstallIstio(ctx context.Context, request CreateRequest) error {
+	regPath := s.knownRegistries["kubeflow"].RegUri
```
jlewi: How is the ISTIO manifest getting into bootstrap? Is it baked into the Docker image and we are using knownRegistries to locate the file?

jlewi:
One minor question but overall looks pretty good.

bootstrap/cmd/bootstrap/app/ksServer.go
lluunn: Yes, the same as deployment manager config

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -1145,6 +1205,14 @@ func finishDeployment(svc KsService, req CreateRequest, dmDeploy *deploymentmana
 		return
 	}
 
+	log.Infof("Installing Istio...")
```
kunmingg: Do we have a flag to control whether or not to install istio?

lluunn: Istio will be required for multi-user support, so I think we always want to install istio?
@jlewi WDYT?

jlewi: Ideally we'd like to specify this declaratively somewhere but I think that involves a bigger change. 

As a short term solution, can we condition this based on the Kubeflow version? For 0.4 we shouldn't install ISTIO by default because that would be a breaking change. But for newer versions of Kubeflow we should.

Longer term, do we check in the ISTIO manifests into our repo? Maybe that's something we should consider doing. This way each version of Kubeflow would have an associated set of versioned manifests just like we do for deployment manager configs.

lluunn: I made it a flag and default to false now.
We do check in istio manifest

lluunn:
/hold cancel
tested locally, able to install istio.

lluunn:
@kunmingg I also made another change: create the admin role binding earlier in configureCluster (was doing it in Apply) because install Istio also needs that. PTAL

jlewi:
/lgtm
/hold
@kunmingg Can you take another look and see whether you are OK with @lluunn's recent change?

kunmingg:
/hold cancel
lgtm, thanks!

PR #2328, opened 2019-01-24
add in-cluster nfs 

IronPan:
- Add in-cluster NFS that uses default storageclass in the cluster.
- switch minio to use in-cluster NFS

NFS ensure that the nfs-pvc is a RWM PVC that can be reused for other pipeline components to mount PV in the future. This will be the first step to deprecate minio in favor of K8s PV.

TODO: Since we don't expose functionality to allow user specify permanent storage, we use the default storageclass. The data will still be gone if cluster is deleted. As next step, expose this configuration as parameter to kfctl.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2328)

### DIFF
--- a/kubeflow/pipeline/all.libsonnet
+++ b/kubeflow/pipeline/all.libsonnet
@@ -2,6 +2,7 @@
   parts(_env, _params):: {
     local params = _env + _params,
 
+    local nfs = import "kubeflow/pipeline/nfs.libsonnet",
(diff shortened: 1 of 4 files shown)


neuromage:
cool!

/lgtm

TimZaman:
Nice. @IronPan could you share the roadmap for storage [on prem/local]?

Issue #2269, opened 2019-01-14
Go based implementation of Jupyter CR?


jlewi:
The Jupyter custom resource is currently implemented as a lambda using jsonnet and meta controller.

Should we replace this with a GoBased implementation of the lambda and continue to use meta controller?

We need to add more functionality to the CR e.g.

* Create ISTIO resources (#2172)

Furthermore, the ISTIO resources might vary depending on whether we want to validate the JWT or not (http://bit.ly/kf_jupyter_design_doc).

I think implementing this in Go might make the code easier to test and maintain in the long term?

@kkasravi What do you think?

Do we have a good example of a lambda for a CR that we could copy? I think the pipelines time has implemented a CR
https://github.com/kubeflow/pipelines/tree/master/backend/src/crd/controller/viewer

Not sure if its using metacontroller.

kkasravi:
I believe they used [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime).
AFAIK metacontroller doesn't do golang.

We have a few choices for a lambda CR

1. Migrate the jsonnet hooks to use controller-runtime in a similar way
2. Use kubebuilder which also leverages controller-runtime
3. Migrate the hook(s) to use kubeless golang hooks
4. Extend meta-controller to include golang

jlewi:
controller-runtime seems like a good choice. Any reason not to go with that and be consistent with kubeflow/pipelines?

kkasravi:
Not that I can think of. Would a good first case be profiles?

stpabhi:
We can use kubebuilder to generate the boilerplate code.

lluunn:
@kkasravi @jlewi I can take a stab at this one if you haven't started (or the profile one)

kimwnasptd:
Hi everyone, we've noticed with @ioandr that one issue with the current UX using JupyterHub is that when a user spawns a Notebook they don't get any feedback for the underlying Pod's status. Specifically, they would just wait for the Pod to start but wouldn't be notified if an error with the Pod's scheduling would occur.
We believe it could help if the Pod's status was mapped in the Notebook's CR. Then we could in turn show it and inform the user from the UI. Any thoughts on this @jlewi @lluunn ?

lluunn:
Yeah I think we can relay the pod status to notebook CR's status

lluunn:
Filed #2417 for followup

PR #2336, opened 2019-01-26
Golang notebook controller

lluunn:
For #2269 

Spec: The notebook CRD currently just provide a PodSpec
Resources: The controller will create attach the command (start.sh jupyter lab ....), port, env vars, and create a StatefulSet, as well as a service.

To review, most of change are in:
components/notebook-crd/pkg/controller/notebook/notebook_controller.go
components/notebook-crd/pkg/apis/notebook/v1alpha1/notebook_types.go

And also the ksonnet manifest

/cc @jlewi 
/cc @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2336)


jlewi:
Thanks @lluunn

lluunn:
Thanks for review @jlewi , PTAL

components/notebook-crd/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Build the manager binary
+FROM golang:1.11.2 as builder
```
kkasravi: you may want to make this an arg set by the Makefile

```
ARG GOLANG_VERSION=1.11.2
FROM golang:${GOLANG_VERSION} as builder
```

components/notebook-crd/Makefile
```diff
@@ -0,0 +1,61 @@
...
+vet:
+	go vet ./pkg/... ./cmd/...
+
+# Generate code
+generate:
+	go generate ./pkg/... ./cmd/...
+
+# Build the docker image
+docker-build: test
+	docker build . -t ${IMG}
```
kkasravi: could add 
```
--build-arg GOLANG_VERSION=$(GOLANG_VERSION)
```

components/notebook-crd/Makefile
```diff
@@ -0,0 +1,61 @@
+
+# Image URL to use all building/pushing image targets
+IMG ?= gcr.io/kubeflow-images-public/notebook-controller
+TAG ?= $(eval TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)-$(shell git diff | shasum -a256 | cut -c -6))$(TAG)
+
```
kkasravi: could add 
```
GOLANG_VERSION ?= 1.11.2
```

components/notebook-crd/cmd/manager/main.go
```diff
@@ -0,0 +1,84 @@
+/*
```
kkasravi: we probably want to name this components/notebook-controller since this is what it is doing

components/notebook-crd/config/crds/notebook_v1alpha1_notebook.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
kkasravi: all the yamls should be removed in lieu of the ksonnet component

components/notebook-crd/config/crds/notebook_v1alpha1_notebook.yaml
```diff
@@ -0,0 +1,47 @@
...
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          properties:
+            template:
```
kkasravi: i think needs fields which are used by the controller and are required

components/notebook-crd/config/default/kustomization.yaml
```diff
@@ -0,0 +1,49 @@
+# Adds namespace to all resources.
+namespace: notebook-crd-system
+
```
kkasravi: this is sort of a departure from our normal practice of using a ksonnet package. I might be nice if we wait until kustomize is supported by kfctl

components/notebook-crd/config/default/manager_image_patch.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: StatefulSet
+metadata:
+  name: controller-manager
+  namespace: system
+spec:
+  template:
+    spec:
+      containers:
+      # Change the value of image field below to your controller image URL
+      - image: IMAGE_URL
```
kkasravi: if this were in ksonnet we could use a parameter from the prototype

components/notebook-crd/pkg/apis/notebook/v1alpha1/notebook_types.go
```diff
@@ -0,0 +1,68 @@
...
+
+// NotebookSpec defines the desired state of Notebook
+type NotebookSpec struct {
+	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
+	// Important: Run "make" to regenerate code after modifying this file
+	Template NotebookTemplateSpec `json:"template,omitempty"`
+}
+
+type NotebookTemplateSpec struct {
+	Spec corev1.PodSpec `json:"spec,omitempty"`
```
kkasravi: in the original code I prevent certain fields in the PodSpec from being overloaded. This could be a security vulnerability if all the Spec fields could be replaced including the containers

components/notebook-crd/Dockerfile
lluunn: done

components/notebook-crd/Makefile
lluunn: done

components/notebook-crd/Makefile
lluunn: done

components/notebook-crd/cmd/manager/main.go
lluunn: makes sense, renamed

lluunn:
@kkasravi @jlewi Thanks for the review

I think there are two main questions:
1. The auto-generated yaml manifest. It's auto-gen by `make manifests`, and it's using kustomize. It's not directly usable since we are using ksonnet, but ksonnet does support yaml drop in and kustomize (I think?) I propose we can ignore the generated yamls for now and continue using our ksonnet. We can revisit in #2351 . WDYT?

2. The CRD spec. Which fields do we expose, and which fields are required?
-  should we provide all the fields with proper default values, and the users can specify nothing?
- what fields can be overwritten by users? I can think of: images, env vars, cmd, service account, volume.
  > in the original code I prevent certain fields in the PodSpec from being overloaded. This could be a security vulnerability if all the Spec fields could be replaced including the containers

jlewi:
@lluunn  Thanks 

> The auto-generated yaml manifest. It's auto-gen by make manifests, and it's using kustomize. It's not directly usable since we are using ksonnet, but ksonnet does support yaml drop in and kustomize (I think?) I propose we can ignore the generated yamls for now and continue using our ksonnet. We can revisit in #2351 . WDYT?

I agree; lets create a ksonnet manifest for now and then thinking about reusing the kustomize manifests in the context of using kustomize.

> The CRD spec. Which fields do we expose, and which fields are required?

should we provide all the fields with proper default values, and the users can specify nothing?
what fields can be overwritten by users? I can think of: images, env vars, cmd, service account, volume.

@kkasravi Can you elaborate on the security vulnerabilities? I thought there were better mechanisms for enforcing things like policies on pods rather than doing it at the controller level.

My expectation would be that the only required parameter would be the docker image and we would create defaults for everything else and that pretty much any field could be overwritten by the user.

lluunn:
> My expectation would be that the only required parameter would be the docker image and we would create defaults for everything else and that pretty much any field could be overwritten by the user.

Yes this is also what I think

components/notebook-controller/go.mod
```diff
@@ -0,0 +1,73 @@
+module github.com/kubeflow/kubeflow/components/notebook-crd
```
kkasravi: this should be changed to notebook-controller so it agrees with the new path

components/notebook-controller/note.txt
```diff
@@ -0,0 +1,308 @@
+	t := &envtest.Environment{
```
kkasravi: file needed?

components/notebook-controller/PROJECT
```diff
@@ -0,0 +1,3 @@
+version: "1"
+domain: kubeflow.org
+repo: github.com/kubeflow/kubeflow/components/notebook-crd
```
kkasravi: path change needed

components/notebook-controller/cmd/manager/main.go
```diff
@@ -0,0 +1,84 @@
...
+limitations under the License.
+*/
+
+package main
+
+import (
+	"flag"
+	"os"
+
+	"github.com/kubeflow/kubeflow/components/notebook-crd/pkg/apis"
```
kkasravi: once you change the go.mod you'll need to update these paths

components/notebook-controller/config/crds/notebook_v1alpha1_notebook.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
kkasravi: so kustomize would be using these files? Are they used now as part of deployment of the controller?

components/notebook-controller/config/crds/notebook_v1alpha1_notebook.yaml
```diff
@@ -0,0 +1,47 @@
...
+    controller-tools.k8s.io: "1.0"
+  name: notebooks.notebook.kubeflow.org
+spec:
+  group: notebook.kubeflow.org
+  names:
+    kind: Notebook
+    plural: notebooks
+  scope: Namespaced
+  validation:
+    openAPIV3Schema:
```
kkasravi: for the ksonnet package I had provided a notebook schema which is basically the PodTemplateSpec schema, might be nice to have something which can be validated

jlewi:
@kkasravi Regarding the differences between the go implementation and the jsonnet implementation? What changes to the go controller do you think we should make?

> My opinion is that overrides should not result in an inconsistent container and shouldn't be an easy back door.

Do you think we are creating a backdoor?

My thinking is that we don't need to worry about enforcing pod security in the controller because there better ways of doing this. e.g PodSecurityPolicies
https://kubernetes.io/docs/concepts/policy/pod-security-policy/

IIUC with PodSecurityPolicies we can define policies that will be applied to pods. So for example we could disallow containers running in privileged mode. So our controller doesn't need to worry about enforcing such policies.

kkasravi:
> My thinking is that we don't need to worry about enforcing pod security in the controller because there better ways of doing this. e.g PodSecurityPolicies

+1 I can open a issue for this...

components/notebook-controller/go.mod
lluunn: done

components/notebook-controller/note.txt
lluunn: removed, thanks

components/notebook-controller/cmd/manager/main.go
lluunn: done

lluunn:
- Fixed the import path.
- PodSecurity issue: #2378
- Will add a README

For the implementation detail, are we ok with:
- only require field: image
- providing all other fields with proper default values, and allowing users to overwrite

lluunn:
Added a README, with spec and todo. PTAL @kkasravi @jlewi 

@kkasravi let's ignore the generated yaml in config/ dir for now as we are using ksonnet.

jlewi:
@lluunn my only comment was about not setting default values for the container command or args because these will be highly dependent on the Docker image so I don't think we can provide sensible defaults.

lluunn:
@jlewi makes sense, removed default args. thanks!

jlewi:
/lgtm
/approve
/hold for @kkasravi 
@kkasravi Any aditional comments?

kkasravi:
> /hold for @kkasravi
> @kkasravi Any aditional comments?

/lgtm
nice work @lluunn !!

PR #2340, opened 2019-01-28
TF-Serving Service: Added param for custom HTTP and GRPC port

rummens:
I have added a new param for the HTTP and GRPC port of the ksonnet tf-serving service prototype. This makes it easier to change it from the defaults, since e.g. Azure only allows traffic on port 80 to the external IP of an Service. 

Port can be changed via:

```
ks param set $COMPONENT portHTTP 80
ks param set $COMPONENT portGRPC 90
```

Since this is my first Pull Request, feel free to correct me on anything :-)

Thanks
Marcel

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2340)

### DIFF
--- a/kubeflow/tf-serving/tf-serving-service-template.libsonnet
+++ b/kubeflow/tf-serving/tf-serving-service-template.libsonnet
@@ -50,12 +50,12 @@
         ports: [
           {
             name: "grpc-tf-serving",
-            port: 9000,
+            port: params.portGRPC,
             targetPort: 9000,
           },
           {
             name: "http-tf-serving",
-            port: 8500,
+            port: params.portHTTP,
             targetPort: 8500,
           },
         ],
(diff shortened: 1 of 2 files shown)


rummens:
> We found a Contributor License Agreement for you (the sender of this pull request), but were unable to find agreements for all the commit author(s) or Co-authors. If you authored these, maybe you used a different email address in the git commits than was used to sign the CLA ([login here](https://cla.developers.google.com/) to double check)? If these were authored by someone else, then they will need to sign a CLA as well, and confirm that they're okay with these being contributed to Google.
> In order to pass this check, please resolve this problem and have the pull request author add another comment and the bot will run again. If the bot doesn't comment, it means it doesn't think anything has changed.

Added gmail Adresse to github account (only had googlemail in it before).

rummens:
Please Run CLA Test again.

lluunn:
I think it's related to the commit user, you might need to amend the commit

rummens:
> I think it's related to the commit user, you might need to amend the commit

Thats strange, I only have one account and used that to commit and to create this PR. How can I ammend the commit? I cannot find the option :-/

lluunn:
https://cla.developers.google.com/clas
You can verify if it's signed.

kubeflow/tf-serving/tf-serving-service-template.libsonnet
```diff
@@ -50,12 +50,12 @@
         ports: [
           {
             name: "grpc-tf-serving",
-            port: 9000,
+            port: params.portGRPC,
```
lluunn: The e2e test failed, it might be that the type should be int not string (I am not sure)

rummens: I checked the cla is signed😅

I tried to use int instead of string but then it wouldn’t compile. Ist there another keyword for int in ks?

Issue #2261, opened 2019-01-11
Replace ambassador with Istio


lluunn:
We are moving toward Istio.
> Current architecture:
> 
> ```
> Ingress -->        Envoy        --> Ambassador --> Other services
>               (JWT valdation)
> ```
> The plan is to use istio-ingressgateway for both JWT validation and routing, and thus:
> 
> ```
> Ingress  --> istio-ingressgateway --> Other services
> ```

There are several tasks needed:

1. Ingress:
Ingress and the NodePort service right behind it must be in the same namespace (https://github.com/kubernetes/kubernetes/issues/17088). I think it's easier to move ingress to istio-system (comparing to moving istio stuffs to kubeflow namespace) because Istio manifest assumes the namespace is istio-system in many places.
- [ ] move ingress and related resources (BackendConfig, oauth secret etc) to istio-namespace

2. Routing
- [ ] For services that are routed by ambassador, remove the annotation and add Istio routes.

3. JWT validation
- [ ] migrate the current envoy config to istio-ingressgateway
- [ ] We need a job/deployment to get the backendService ID for JWT audience, and apply the Policy.

Example policy:
```
apiVersion: authentication.istio.io/v1alpha1
kind: Policy
metadata:
  name: ingress-jwt
  namespace: istio-system
spec:
  targets:
  - name: istio-ingressgateway
    ports:
    - number: 80
  origins:
  - jwt:
      issuer: https://cloud.google.com/iap
      jwksUri: https://www.gstatic.com/iap/verify/public_key-jwk
      audiences:
      - /projects/711890160434/global/backendServices/8282354174894887635
      jwtHeaders:
      - x-goog-iap-jwt-assertion
      trigger_rules:
      - excluded_paths: 
        - exact: /healthz
  principalBinding: USE_ORIGIN
```

cc @kunmingg @jlewi

iftachsc:
my question is related to the subject but not to kubeflow. as an experienced istio user, reviewing ambassador doesn't give me a reason why should i use ambassador and not the regular istio-ingress gateway, cant find the benefits. can some one enlighten   me?

lluunn:
we chose ambassador at the time istio didn't exist. Now we are migrating to istio

Jeffwan:
@lluunn do you have a timeline for this change?

lluunn:
targeting 0.5

kunmingg:
@gabrielwen

Jeffwan:
[...]

PR #2348, opened 2019-01-30
use Istio for JWT validation

lluunn:
Changes are in iap.libsonnet controlled by param: useIstio.
Major differences:
1. no envoy service and deployment, and thus no envoy-config
1. setup_backend.sh now creates policy for JWT validation (instead of sed envoy config)
1. update_backend.sh creates several routing rules (health check, ambassador)

/cc @jlewi

needs #2352 
related: #2261 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2348)


lluunn:
health check not working yet

lluunn:
Ready for review.
/assign @jlewi 
/assign @kunmingg

kunmingg:
If most changes happen around iap, can we have iap-istio component instead of whole new package?
Feels a bit confusing that both pkg contains same prototypes, as once user generate gcp components from pkg, later user can't tell which pkg they used unless check source code.

jlewi:
I like @kunmingg 's suggestion? Perhaps we could just add a bool parameter "use_istio" and then conditionally install various components and configure them differently based on that parameter?

lluunn:
make it a param now, PTAL

lluunn:
/hold
testing

lluunn:
/hold cancel
tested locally, will add e2e test

jlewi:
/lgtm
/hold 
for @kunmingg

kunmingg:
/hold cancel
lgtm
/approve

PR #2349, opened 2019-01-30
Prepare Notebook Controller for the new Jupyter UI

kimwnasptd:
This PR adds various fixes and updates to the `sync-notebook` Webhook of the Notebook Controller that is responsible for creating child resources for each Notebook CR. The suggested changes are a prerequisite to ensure that the new, forthcoming Jupyter UI can directly manage Notebook CRs and preserve existing functionality.

@jlewi @kam We would really appreciate your perspective on this to make sure we are all on the same page. Hereaupon, we briefly describe the issues we diagnosed while working with the existing Notebook Controller and its Webhook:

- The child `Service` does not use WebSockets and, thus, terminals and Python interpreters hang inside the JupyterLab UI. `HTTP GET` requests from inside the Notebook Pod result in `403 Forbidden` responses. 
- The child `Deployment` always overrides the Notebook CR `templateSpec` with the `podTemplateSpec` that contains certain default values. This is not desirable when Users want to spawn a Notebook that overrides default configuration.
- The child `Deployment` uses singled-quoted True/False values in two container spec args related to the Jupyter Notebook, so their values were treated as String and not as Boolean. This results in the Notebook Pod going to `CrashLoopBackOff` status with the following error in its logs: 
   ```
  [C 15:56:44.180 LabApp] Bad config encountered during initialization:                                                                                                                                                                                                           
  [C 15:56:44.180 LabApp] The 'allow_remote_access' trait of a LabApp instance must be a boolean, but a value of 'True' <class 'str'> was specified.   
  ```
  Related sources:
  https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L845
  https://github.com/jupyter/notebook/blob/master/notebook/notebookapp.py#L633
  
  **NOTE:**
  According to [this issue](https://github.com/kubeflow/kubeflow/issues/300), root access should not be allowed in a JupyterLab environment. However, the Notebook Controller currently sets the `allow_root` argument to `True`. Which is the desired behavior?

- The Notebook Pod does not use the latest stable Tensorflow Docker image, as the current JH Spawner UIs do

In addition, since the new Jupyter UI supports Volumes for Notebooks, we suggest the following additions to the Notebook Controller:

- Set the “volumes” field in `podTemplateSpec` and set its default value to an empty list, as fallback
- Explicitly set the `securityContext` (fsGroup) of the Notebook Pod to avoid permission errors when accessing attached volumes from inside the Notebooks
- Remove the unused optional `disks` and `pvcMount` ksonnet parameters from the `notebooks` KF component. In the future, we might need to go over the rest of the defined parameters and cross-check which are currently needed.

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="34" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/kubeflow/kubeflow/2349)
<!-- Reviewable:end -->


jlewi:
@kkasravi Any comment on this?

@kimwnasptd @ioandr @lluunn Has reimplemented the controller in go #2336 so I think you'll need to make the same changes there once its merged.

kimwnasptd:
Thanks for the heads up @jlewi. We will keep an eye on it and port these changes in a new PR once the new Notebook Controller  gets merged.

kimwnasptd:
We thought it was good to give an early warning because these are issues that @lluunn may also bump into, when testing the new Go operator.

lluunn:
Thanks for the heads up

kubeflow/jupyter/sync-notebook.jsonnet
```diff
@@ -46,13 +46,18 @@ function(request) {
         workingDir: "/home/jovyan",
       },
     ],
+    volumes: [],
     restartPolicy: "Always",
+    securityContext: {
```
kkasravi: should 100 be a param like params.notebookGid?

jlewi:
Obsolete.
We are now using a go based controller.

PR #2353, opened 2019-01-30
Add Deployment manager job to create permanent storage

IronPan:
Add a separate DM entry for creating all storages that should live beyond the lifecycle of the cluster.
When user delete the cluster DM, the data should be persisted. 

For the first iteration, add entry to create GCE PD for pipeline. 
Also move the existing CloudSQL entry to the new DM job.

The new DM entry is controlled by a feature flag and is disabled by default

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2353)


bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -1234,12 +1259,20 @@ func makeDeployEndpoint(svc KsService) endpoint.Endpoint {
 			return r, err
 		}
 
-		DmDeployment, err := svc.InsertDeployment(ctx, req)
+		if req.CreatePersistentStorage {
+			_, err := svc.InsertDeployment(ctx, req, StorageDmSpec)
```
kunmingg: Can we check if the deployment of storage actually succeeded?

IronPan: yeah i've verified and saw it created ok

kunmingg: I mean in "finishDeployment" can we check if both InsertDeployment succeeded?

IronPan: done

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -1130,28 +1154,29 @@ func timeSinceStart(ctx context.Context) time.Duration {
...
+			return err
 		}
 		if status == "DONE" {
 			if errMsg != "" {
 				log.Errorf("Deployment manager returned error message: %v", errMsg)
 				// Mark status "INVALID_ARGUMENT" as most deployment manager failures are caused by insufficient quota or permission.
 				// Error messages are available from UI, and should be resolvable by retries.
 				deployReqCounter.WithLabelValues("INVALID_ARGUMENT").Inc()
-				return
+				return err
```
kunmingg: here ```err == nil```,
but actually we might want to return an error containing errMsg, 
as the deployment action actually failed.

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -1130,28 +1154,29 @@ func timeSinceStart(ctx context.Context) time.Duration {
...
 			}
 			clusterDeploymentLatencies.Observe(timeSinceStart(ctx).Seconds())
 			log.Infof("Deployment is done")
```
kunmingg: Let's move 
```
clusterDeploymentLatencies.Observe(timeSinceStart(ctx).Seconds())
log.Infof("Deployment is done")
```
to ```finishDeployment```
so they are executed after both ```checkDeploymentFinished``` are done

bootstrap/cmd/bootstrap/app/ksServer.go
IronPan: done

bootstrap/cmd/bootstrap/app/ksServer.go
IronPan: done

kunmingg:
Thanks!
/lgtm
/approve

Issue #2359, opened 2019-01-31
[Dashboard v1][Tracker][Redesign] Landing Page v1


avdaredevil:
## General Idea
Create a landing page that provides all existing functionality while also starting to offer some basic help and information. This is a tracker issue. I will be adding these features!

## Features
Tentative list of features being added:
- Create a responsive and a11y enabled **sidebar**
  - Move links to a sidebar
  - Add link for Home
  - Add Kubeflow Logo
- Create a **Main View**
  - Either contains the page (**`Dashboard`** / **`Activity`**)
  - Or an Iframe view which loads all the help links!
- Contains a **_header toolbar_** (*in **Main View***)
  - Ability to swap between **Dashboard** and **Activity**

## Backend
Goals:
- Provide REST APIs to expose:
  - Namespaces
  - Build information
  - Any relevant metrics?

### Dashboard
- `paper-card` that contains action items to many of the common actions needed to be performed on the cluster
- Possibly a chart :)

jlewi:
@avdaredevil is actually working on this and has a PR pending.

avdaredevil:
Added some info on the backend info

jlewi:
@avdaredevil what is the remaining work on this issue? 

* Do we need to publish a docker image for the new app? Or are you just reusing the existing central dashboard image?

* Do we need to update the ksonnet component used to deploy it
  
  * We currently use this component to deploy the central dashboard:
    https://github.com/kubeflow/kubeflow/blob/master/kubeflow/common/centraldashboard.libsonnet

* Integrate the new dashboard into the kfctl go binary
   * I think the list of packages installed by default is defined in go here:
    https://github.com/kubeflow/kubeflow/blob/32aa25460d884dc73407ca6651535366ae41bf66/bootstrap/pkg/apis/apps/ksonnet/v1alpha1/application_types.go#L63

jlewi:
#2540 updated the image to 
gcr.io/kubeflow-images-public/centraldashboard:v20190222-v0.4.0-rc.1-152-g1de09b7a

But it looks like that image doesn't exist

sarahmaddox:
In iframing the docs into the UI, please see issue https://github.com/kubeflow/kubeflow/issues/2546

jlewi:
@avdaredevil @prodonjs what is the remaining work before we can close out this issue?

prodonjs:
[...]

PR #2361, opened 2019-02-01
Dashboard v1 UI rewrite done

avdaredevil:
#### Targets issue #2359

Before | After
--- | ---
Before design for Kubeflow Dashboard | AP's design for Kubeflow Dashboard

Iframe view |
--- |
image

### Responsive
Responsive Menu | Responsive Page with Hamburger
--- | ---
image | image

## TODOs
- Most UI work for now is done _(will confirm with UX folk)_
- Setup a backend to pass data to the dash

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2361)


jlewi:
@avdaredevil This is great. It would be good to demo this at one of the community meetings once its merged.

avdaredevil:
@jlewi Thanks! Sounds good to me

kunmingg:
/lgtm
/approve
Thanks!

abhi-g:
🎉🎉🎉

components/centraldashboard/frontend/index/components.html
```diff
@@ -0,0 +1,22 @@
+<base href="https://cdn.rawgit.com/download/polymer-cdn/2.6.0/lib/">
```
yebrahim: We should move away from using the CDN.
Separately, rawgit is going away, we shouldn't rely on it even in the interim.

components/centraldashboard/frontend/index/main-page.html
```diff
@@ -0,0 +1,191 @@
...
+                    {text: 'Deploy a solution', icon: 'donut-large'},
+                    {text: 'Compare run output', icon: 'donut-large'},
+                    {text: 'Share solution on AI Hub', icon: 'donut-large'},
+                ]},
+                sidebarItemIndex: {type: Number, value: 0},
+                primaryViewIndex: {type: Number, value: 0},
+                homeOrIframeViewIndex: {type: Number, value: 0},
+                url: {type: String, value: ""},
+            }}
+            static get observers() {return [
```
yebrahim: cleanup?

kunmingg:
@yebrahim 
Thanks!
Can we select framework in follow up PRs, especially when backend parts are involved?

yebrahim:
My comment was about the use of Polymer, I'm not aware of the backend parts.

avdaredevil:
@yebrahim Let's continue this conversation here: #2390

sarahmaddox:

PR #2372, opened 2019-02-02
upgrade conda to 4.5.12 and stick with pip version 19.0.1 for now

ashahba:
- Don't upgrade `pip` blindly since it's known to have caused issues in the past.
- Upgrade `conda` to latest version `4.5.12` since the current version is ancient.


ashahba:
Fixes #2374

ashahba:
This adds a good note about `Miniconda` and `pip` versions for future reference.
/assign @jlewi

PR #2373, opened 2019-02-02
deploy storage through kfctl.sh

IronPan:
Create kubeflow storage DM as part of kfctl.sh
This creates GCE PD as permanent storage and the storage is used by pipeline service for storing data permanently.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2373)

### DIFF
--- a/scripts/kfctl.sh
+++ b/scripts/kfctl.sh
@@ -12,6 +12,7 @@ set -xe
 
 ENV_FILE="env.sh"
 SKIP_INIT_PROJECT=false
+CREATE_PIPELINE_PERSISTENT_STORAGE=true
 
 # To enable GKE beta features we need to use the v1beta1 API.
 # https://cloud.google.com/kubernetes-engine/docs/reference/api-organization#beta
@@ -91,6 +92,7 @@ createEnv() {
   export DOCKER_REGISTRY_KATIB_NAMESPACE=${DOCKER_REGISTRY_KATIB_NAMESPACE:-""}
   # Namespace where kubeflow is deployed
   export K8S_NAMESPACE=${K8S_NAMESPACE:-"kubeflow"}
+  export CREATE_PIPELINE_PERSISTENT_STORAGE=${CREATE_PIPELINE_PERSISTENT_STORAGE:-"true"}
 
   case "$PLATFORM" in
     minikube)
@@ -119,7 +121,8 @@ createEnv() {
               'KUBEFLOW_ENDPOINT_NAME=$KUBEFLOW_ENDPOINT_NAME\n'
               'KUBEFLOW_HOSTNAME=$KUBEFLOW_HOSTNAME\n'
               'CONFIG_FILE=$CONFIG_FILE\n'
-              'GKE_API_VERSION=$GKE_API_VERSION\n')
+              'GKE_API_VERSION=$GKE_API_VERSION\n'
+              'CREATE_PIPELINE_PERSISTENT_STORAGE=$CREATE_PIPELINE_PERSISTENT_STORAGE\n')
       FORMAT+=('$PROJECT'
                '$ZONE'
                '$EMAIL'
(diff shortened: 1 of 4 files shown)


scripts/kfctl.sh
```diff
@@ -260,6 +264,9 @@ parseArgs() {
       --skipInitProject)
         SKIP_INIT_PROJECT=true
         ;;
+      --skipCreatePipelinePersistentStorage)
```
jlewi: Why the flag? Can we follow the pattern of
kfctl generate  
Performs the default desired behavior

For non-default behavior; customize the configs after running kfctl generate?

IronPan: do we expect user to change env.sh manually? If so I don't have to pass this flag here. 

I'm not sure what can be added to the args here what can't. it looks like there are a few args here are also in env.sh. If we expect user to change env.sh directly, when will they pass them directly?

jlewi: The user can customize the behavior via the YAML file storage-kubeflow.yaml right?
e.g. via the option createPipelinePersistentStorage and others?

So the pattern we'd like to follow is kfctl takes no arguments related to this.
kfctl.sh generate creates the storage-kubeflow.yaml file with suitable defaults.
User can then modify that YAML file as needed to get desired behavior; e.g. use PD or not.

IronPan: the flag in storage-kubeflow.yaml is specific for DM, while whether creating a persistent storage  would impact what parameter to pass to ks app downstream. The flag should be declared and used at a higher level. and that's why i put it in env variable. Is there any better location for that flag?

jlewi: Why do you need a higher level flag? 

Could you just tell users to do

ks init
ks generate
1. edit storage-kubeflow.yaml
2. Run  
    ks param set pipeline mysqlPd "${DEPLOYMENT_NAME}-storage-pipeline-db"
    ks param set pipeline nfsPd "${DEPLOYMENT_NAME}-storage-pipeline-nfs"

The problem with introducing a new command line flag is what's the source of truth?
Is the environment variable which is persisted in env.sh or is it the DM and ksonnet configs?

My preference would be to validate that customers are actually use it and then figure out how to add sugar to make it simpler.

IronPan: done.

kunmingg:
Causing https://github.com/kubeflow/kubeflow/issues/2401,
We need to verify after each pre-submit, all disks are deleted as well.

testing/workflows/components/kfctl_test.jsonnet
```diff
@@ -313,8 +313,30 @@ local deleteStep = if deleteKubeflow then
   }]
 else [];
 
+// Clean up all the permanent storages to avoid accumulated resource
+// consumption in the test project
+local deleteStorageStep = if deleteKubeflow then
+  [{
+    template: buildTemplate(
```
jlewi: I think it would be better to add it to this script
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/delete_deployment.sh

Possibly guarded by an option on whether or not to delete storage.

IronPan: I thought about the same but saw this
https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/delete_deployment.sh#L3
which looks fine to me. Since we don't provide option in kfctl.sh parameter to create a storage, it's consistent to me that we don't provide option in kfctl.sh delete.

Let me know if you feel strong about it.

IronPan:
verified the storage is deleted correctly. 
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-kfctl-2373-1f85581-5486-b8d0?tab=workflow&nodeId=kubeflow-presubmit-kfctl-2373-1f85581-5486-b8d0-184529030

IronPan:
`/test kubeflow-presubmit`

PR #2384, opened 2019-02-04
add query string to click-to-deploy app to support arbitrary version

IronPan:
A legit URL would look like 
http://localhost:3000/#/deploy?version=[commit-sha]

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2384)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -192,7 +201,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
               process.env.REACT_APP_VERSIONS.split(',').map((version, i) => (
                 <MenuItem key={i} value={version}>{version}</MenuItem>
               )) :
-              versionList.map((version, i) => (
+                this._versions.map((version, i) => (
```
kunmingg: Currently ```process.env.REACT_APP_VERSIONS``` has higher priority to set version list.
You probably want to use new param as high priority one for version list

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -121,6 +124,13 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     // be able to click submit until the fetches have succeeded. How can we do
     // that?
 
+    const values = queryString.parse(this.props.location.search);
```
yebrahim: You can instead do something like"
```suggestion
    const params = new URLSearchParams(this.props.location.search);
    if (params.get('version')) {
      ...
```

IronPan: done.

IronPan:
to @kunmingg i would actually prefer use plain text to start with, and for pipeline we would likely use commit id instead tag name, so we don't need to create tag. this would allow people deploy arbitrary commit easily (without base64 encode manually)
if we really see base64 is necessary we can always switch.

components/gcp-click-to-deploy/src/DeployForm.tsx
IronPan: when and how was it set? is it a config parameter in click-to-deploy application that's not exist in this repo?

IronPan: anyway always respect the query string as the first option. done

kunmingg: REACT_APP_VERSIONS is set through configmap in k8s cluster.

kunmingg:
/lgtm

/hold 
for @yebrahim

components/gcp-click-to-deploy/tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "suppressImplicitAnyIndexErrors": true,
-    "noUnusedLocals": true
+    "noUnusedLocals": true,
+    "allowSyntheticDefaultImports": true
```
yebrahim: You don't need this anymore.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -121,6 +123,13 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     // be able to click submit until the fetches have succeeded. How can we do
     // that?
 
+    const params = new URLSearchParams(this.props.location.search);
+    if (params.get('version')) {
+      this._query_string_version = params.get('version');
```
yebrahim: Why the extra variable? Why not just append/prepend to the versions array? Then the rest of the render logic can remain as is: render all items in the list, and auto-select the first/last one.

IronPan: kunming mentioned the REACT_APP_VERSIONS will take precedence. The code might get ugly without this extra variable.

components/gcp-click-to-deploy/tsconfig.json
IronPan: done

components/gcp-click-to-deploy/src/DeployForm.tsx
yebrahim: Suggestion: This logic should just be done at the beginning of the render function rather than here, it'd be cleaner to first compute the list then render the entire array here.

IronPan: done

Issue #2408, opened 2019-02-06
kfctl.sh delete all timeout failed to delete namespace; problems deleting minio


IronPan:
The log snippet parsed below. 
There are issues such as https://github.com/kubernetes/kubernetes/issues/60807 describing the similar thing. 

It happens in pipeline prow test that when test finish and want to do full delete, it's stuck in this step 
https://github.com/kubeflow/kubeflow/blob/master/scripts/kfctl.sh#L487
The test job is killed by prow after 2h timeout. However the cluster is not deleted and cause resource leak.

A short term mitigation is to skip delete namespace in GCP platform. 

To further debug and diagnose cluster problems, use '\''kubectl cluster-info dump'\''.' '[^\.0-9]*\([\.0-9]\+\)'
+ MASTER=35.226.133.205
+ echo MASTER=35.226.133.205
MASTER=35.226.133.205
+ [[ 35.226.133.205 != \3\5\.\2\2\6\.\1\3\3\.\2\0\5 ]]
+ echo 'kubectl context matches ks environment default'
kubectl context matches ks environment default
+ set +e
+ kubectl delete ns/kubeflow
namespace "kubeflow" deleted
+ kubectl get ns/kubeflow
NAME       STATUS        AGE
kubeflow   Terminating   23m
+ echo 'namespace kubeflow not yet deleted. sleeping 10 seconds...'
namespace kubeflow not yet deleted. sleeping 10 seconds...
+ sleep 10
+ kubectl get ns/kubeflow
NAME       STATUS        AGE
kubeflow   Terminating   23m
+ echo 'namespace kubeflow not yet deleted. sleeping 10 seconds...'
namespace kubeflow not yet deleted. sleeping 10 seconds...
+ sleep 10
+ kubectl get ns/kubeflow
NAME       STATUS        AGE
kubeflow   Terminating   23m
+ echo 'namespace kubeflow not yet deleted. sleeping 10 seconds...'
namespace kubeflow not yet deleted. sleeping 10 seconds...
+ sleep 10
+ kubectl get ns/kubeflow
NAME       STATUS        AGE
kubeflow   Terminating   24m
+ echo 'namespace kubeflow not yet deleted. sleeping 10 seconds...'
namespace kubeflow not yet deleted. sleeping 10 seconds...
+ sleep 10
+ kubectl get ns/kubeflow
NAME       STATUS        AGE
kubeflow   Terminating   24m
[...]

PR #2409, opened 2019-02-06
Wait for at most 100s on namespace deletion

IronPan:
Mitigate https://github.com/kubeflow/kubeflow/issues/2408

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2409)

### DIFF
--- a/scripts/kfctl.sh
+++ b/scripts/kfctl.sh
@@ -492,9 +492,9 @@ main() {
         sleep 10
       done
       if kubectl get ns/${K8S_NAMESPACE}; then
-        echo "namespace ${K8S_NAMESPACE} successfully deleted."
-      else
         echo "namespace ${K8S_NAMESPACE} failed to delete."
+      else
+        echo "namespace ${K8S_NAMESPACE} successfully deleted."
       fi
       set -e
     fi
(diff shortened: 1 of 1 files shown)


scripts/kfctl.sh
```diff
@@ -486,6 +486,9 @@ main() {
       kubectl delete ns/${K8S_NAMESPACE}
       while kubectl get ns/${K8S_NAMESPACE}; do
         echo "namespace ${K8S_NAMESPACE} not yet deleted. sleeping 10 seconds..."
+        # wait for 100s at most
+        # https://github.com/kubeflow/kubeflow/issues/2408
+        ((c++)) && ((c==10)) && break
```
kunmingg: We should give message ```delete failed``` instead of ```successfully deleted``` if we break here due to timeout

IronPan: done

scripts/kfctl.sh
```diff
@@ -486,9 +486,16 @@ main() {
...
         sleep 10
       done
-      echo "namespace ${K8S_NAMESPACE} successfully deleted."
+      if kubectl get ns/${K8S_NAMESPACE}; then
+        echo "namespace ${K8S_NAMESPACE} successfully deleted."
```
kunmingg: Seems we should swap these 2 messages?

IronPan: done

kunmingg:
/lgtm
/approve
thx!

PR #2428, opened 2019-02-07
fixes 'make it easy to configure the foo sample plugin as a plugin or statically linked'

### LINKED ISSUE #2414: make it easy to configure the foo sample plugin as a plugin or statically linked
kfctl can load plugins as dynamic libraries `<plugin>.so`. We should make it easy to enable the sample plugin foo as a plugin or as part of kfctl (statically linked into the kfctl binary). 

We should do this as a make target:

`make static`
`make plugins`

and document this feature along with the default configuration for foo (static).

kkasravi:
fixes #2414

this PR is based off of PR #2426

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2428)


jlewi:
Thanks @kkasravi 

Two questions

1. Is using plugins going to complicate things? Would we still publish a single binary or is the user going to have install the binary and multiple plugins?
   * Wouldn't we want to install GCP, minikube, and plane always and not have them be optional plugins?


1. Is there a better way to enable a debug mode that doesn't require changing the source code files? Could we just have a command line option.

jlewi:
> The use of plugins really won't complicate things. I moved everything to plugins based on a 1:1 conversion we had (https://kubeflow.slack.com/archives/D9GM6L6JE/p1548729168003700). Brew will let us install the kfctl binary and related .so's in known locations so the user won't be exposed to configuring any kind of plugin path.

What about people who aren't using brew? 

One of the goals with moving to a go binary was to simplify the process compared to what we have today a download.sh script. One of the pain points of the current shell script is that you need to download and unpack a directory and have all the paths work out.

Isn't the fact that you have two different code branches for debug and no debug mode indicate plugins complicate things?

How about we statically link in some core plugins like GCP and minikube for now and then try to find a specific plugin to test actual plugins before converting everything to plugins?

kkasravi:
@jlewi 
Hi Jeremy

Should be good to go.

jlewi:
Woo Hoo!

/lgtm
/approve

Issue #2364, opened 2019-02-01
Re-enable click_deploy_test


lluunn:
https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml#L6

I think it's not being run:
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-periodic-master

@kunmingg mentioned the cluster might need more quota.

kunmingg:
An easy approach is to make e2e test clean up before each run, we can reuse what prober test is doing today:
https://github.com/kubeflow/kubeflow/blob/master/testing/test_deploy_app.py#L363

lluunn:
It looks like it should be running in kubeflow-ci/kubeflow-testing cluster:
https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/click_deploy_test.jsonnet#L192

@kunmingg If it's quota problem, shouldn't the test still be run, and then failed?

kunmingg:
Let's setup a dedicated cluster for it.
Will double check periodic test config, maybe some PRs break the periodic test config

lluunn:
<img width="720" alt="screen shot 2019-02-04 at 14 41 35" src="https://user-images.githubusercontent.com/32310205/52242318-366a1d80-288b-11e9-9102-d6c2c4e5c3e4.png">

The log shows: the container main in pod XXX is not available: Bad Request

lluunn:
manually launched the workflow. Saw:

CreateContainerConfigError: secrets "kubeflow-oauth" not found

lluunn:
The plan is to deploy to a new project (kubflow_ci_bootstrap or something under kubeflow.org) so that quotas/resources are easier to cleanup.

lluunn:
Created project: kubeflow-ci-deployment

lluunn:
<img width="652" alt="screen shot 2019-02-07 at 10 42 28" src="https://user-images.githubusercontent.com/32310205/52434809-14f47600-2ac5-11e9-978b-0fef3ac890b9.png">

Still failing:
```
go: finding github.com/prometheus/client_model v0.0.0-20180712105110-5c3871d89910
go: finding google.golang.org/genproto v0.0.0-20180831171423-11092d34479b
go: finding go.opencensus.io v0.18.1-0.20181204023538-aab39bd6a98b
go: finding github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973
go: error loading module requirements
+ result=1
+ set -e
[...]

PR #2431, opened 2019-02-07
fix click_to_deploy test

lluunn:
- make `copy-artifacts` not depend on `deploy-delete` so that even delete failed we still have artifacts. @kunmingg Is this ok?
- when GetDeploymentStatus failed, retry
- update go version, and print out go version

for https://github.com/kubeflow/kubeflow/issues/2364

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2431)

### DIFF
--- a/bootstrap/Dockerfile
+++ b/bootstrap/Dockerfile
@@ -1,7 +1,7 @@
 #
 # bootstrap_base
 #
-ARG GOLANG_VERSION=1.11.2
+ARG GOLANG_VERSION=1.11.5
 FROM golang:$GOLANG_VERSION as bootstrap_base
 
 RUN apt-get update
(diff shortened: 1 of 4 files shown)


lluunn:
ready for review
/assign @kunmingg 
/assign @jlewi

jlewi:
@lluunn Does delete_deploy emit a junit test file to report the results in test grid?

If its not reporting a junit file then its probably ok to run copy-artifacts in parallel as delete-deploy won't emit any results anyway to upload to GCS.

If it does then it might be better to have deploy emit a junit that indicates failure but still exit with code 0. This way the next step copy-artifacts would still run.

lluunn:
it's won't.
it's doing https://github.com/kubeflow/kubeflow/blob/master/bootstrap/test_delete.sh

PR #2438, opened 2019-02-08
ingress entry for basic auth

kunmingg:
Create new entry for basic auth ingress on GCP.
How to use:
Apply both:
```
gcp/basic-auth-ingress
common/basic-auth
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2438)


lluunn:
basic-auth-ingress and iap-ingress have many duplicate code, should we merge them?
Controlled by param basicAuth?

kunmingg:
I don't param control here as:
1. they require different kind of input:
one is oauth pair
another is username/password pair
2. want to prevent user misunderstanding:
change param won't change underlying ingress.

PR #2371, opened 2019-02-02
Remove all unnecessary white spaces at the end of lines

ashahba:
Here are some of the commands to achieve this partially:
```
$ for file in $(git ls); do sed 's/[[:blank:]]*$//' ${file} > ${file}.new; cat ${file}.new > ${file}; rm -rf ${file}.new; done
```
Once done there would be a few files marked as untracked.
If all the files are truly untracked and not needed anywhere in the repo run this command:
```
$ git clean -f
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2371)

PR #2447, opened 2019-02-11
Update python code styles based on what's provided in .style.yapf

### LINKED ISSUE #2448: Update Python scripts based on ".style.yapf" config.

ashahba:
Fixes #2448 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2447)


ashahba:
/assign @jlewi 
Let me know what you think.
My idea is to be consistent. I don't have any preferences over `2` or `4` spaces for `Python` scripts.
I'm open to ideas.

jlewi:
I was wondering the same thing about 2 vs. 4 spaces. 

Is it easy to chang yapf settings and rerun so we can see how it looks?
I wonder what it would like if we enable
* continuation indent of 4
* and SPLIT_BEFORE_FIRST_ARGUMENT  https://stackoverflow.com/questions/45042530/how-to-configure-yapf-to-use-hanging-indentation-for-long-argument-lists

Is there a project whose style we should just copy?

ashahba:
@jlewi Although `Google` style suggests using `2` spaces for indentation improves performance but I'd imagine that's comes to play only if you deal with files consisting `100`'s of `1000`'s of lines.
Our top `10` longest `Python` files are:
```
     779 testing/test_deploy.py
     698 testing/test_deploy_app.py
     458 kubeflow/jupyter/ui/default/spawner.py
     384 components/k8s-model-server/http-proxy/server.py
     373 kubeflow/jupyter/ui/rok/spawner.py
     214 testing/katib_studyjob_test.py
     203 testing/deploy_utils.py
     202 tf-controller-examples/tf-cnn/create_job_specs.py
     188 components/build_image.py
     180 build/boilerplate/boilerplate.py
```
On the other hand a lot of developers prefer `4` space indentation and `80` chars long lines.
I like the idea of enabling `SPLIT_BEFORE_FIRST_ARGUMENT`.
I also don't have a project on top of my head to adopt their style.
However I tweak some knobs and came up with a new `.style.yapf` that showcases `4` spaces indentation: https://github.com/kubeflow/kubeflow/pull/2461
The code seems be more readable I think.

Let me know your thoughts.
Thanks.

jlewi:
This version #2447 looks better
Compare
https://github.com/kubeflow/kubeflow/blob/0c1503a63d581a6d1875f431a0f128612abbe5fc/components/build_image.py

```
def run(command,
        cwd=None,
        env=None,
        polling_interval=datetime.timedelta(seconds=1)):
  """Run a subprocess.
    Copied from kubeflow/test so it's easier to run locally.
    TODO(lunkai): refactor to dedup.
    Any subprocess output is emitted through the logging modules.
    Returns:
      output: A string containing the output.
  """
  logging.info("Running: %s \ncwd=%s", " ".join(command), cwd)
```

Compared to #2461 

https://github.com/kubeflow/kubeflow/blob/450ce57fdd5b51914170c1eae2f57baf172ed825/components/build_image.py

```
def run(
  command, cwd=None, env=None, polling_interval=datetime.timedelta(seconds=1)
):
    """Run a subprocess.
      Copied from kubeflow/test so it's easier to run locally.
      TODO(lunkai): refactor to dedup.
      Any subprocess output is emitted through the logging modules.
      Returns:
        output: A string containing the output.
    """
```

components/build_image.py
```diff
@@ -75,24 +80,28 @@ def wait_for_docker_daemon(timeout=60):
   # TODO(lunkai): use TimeoutError when we use py3.
   raise RuntimeError
 
+
 def get_build_args(config):
   """
-  Make the list of params for docker build from config.
+    Make the list of params for docker build from config.
```
jlewi: Is it normal for the comment to be indented with respect to """?

components/echo-server/main_test.py
```diff
@@ -15,68 +17,57 @@
...
 
 def test_echo(client):
-    r = client.post(
-        '/echo',
-        data='{"message": "Hello"}',
-        headers={
-            'Content-Type': 'application/json'
-        })
+  r = client.post(
+    '/echo',
```
jlewi: This looks a little weird; I would have expected

r = client.post('/echo',
                       data=....
..

components/k8s-model-server/http-proxy/server.py
```diff
@@ -125,25 +148,27 @@ def decode_b64_if_needed(data):
...
+    Returns:
+      The signature map of the model.
   """
   request = get_model_metadata_pb2.GetModelMetadataRequest()
   request.model_spec.name = model_name
   request.metadata_field.append("signature_def")
   try:
-    response = model_server_stub.GetModelMetadata(request, MODEL_SERVER_METADATA_TIMEOUT_SEC)
+    response = model_server_stub.GetModelMetadata(
+      request, MODEL_SERVER_METADATA_TIMEOUT_SEC)
```
jlewi: I would have preferred request to be indented 4 because of hanging indention.

components/k8s-model-server/http-proxy/server.py
```diff
@@ -125,25 +148,27 @@ def decode_b64_if_needed(data):
...
   except grpc.RpcError as rpc_error:
-    logging.exception("GetModelMetadata call to model server failed with code "
-                      "%s and message %s", rpc_error.code(),
-                      rpc_error.details())
+    logging.exception(
+      "GetModelMetadata call to model server failed with code "
```
jlewi: I preferred aligning on the "(" as it was before.

jlewi:
Overall this looks good to me. 

I left a couple comments but I don't know if its possible to fix those things.

I definitely thing think this is version is preferable #2461.

To clarify; I think 2 spaces indention is preferable for readability; otherwise you can't fit as much on a page.

I like 4 spaces only for continued lines; in the current pr continued lines are also indented 2 spaces.

ashahba:
Thanks @jlewi so we'll go with this version and I try to find a specific know that fixes your noted items above.

jlewi:
Thanks @ashahba.

This is already a big improvement so if we can't find solutions to my nitpicks I think that's fine; so I would time box how much time you spend on it.

components/build_image.py
ashahba: Nice catch, this is not normal, it'll be fixed in next commit.

components/echo-server/main_test.py
ashahba: Hmm, the recent version I have it looks like this:
```
def test_echo(client):
  r = client.post(
      '/echo',
      data='{"message": "Hello"}',
      headers={'Content-Type': 'application/json'})

  assert r.status_code == 200
  data = json.loads(r.data.decode('utf-8'))
  assert data['message'] == 'Hello'
```
which looks normal because it's continuation of line.

components/k8s-model-server/http-proxy/server.py
ashahba: With the recent update it does:
```
  try:
    response = model_server_stub.GetModelMetadata(
        request, MODEL_SERVER_METADATA_TIMEOUT_SEC)
```
So we should be good.

components/k8s-model-server/http-proxy/server.py
ashahba: This might be hard to achieve or possibly a conflicting `yapf` knobs' behavior.
On the other hand if we align to `(`, wouldn't that cause max line length to be consumed pretty quick when dealing with long strings like this one and we end up with too many lines.

I personally see this consistent because it's forcing the `4` space continuation policy 🙂

ashahba:
@jlewi This is one is ready to go. Looks like as soon os we merged #2483 this one passed so that one definitely helped with hardening tests 🙂

jlewi:
@ashaba did you mean to edit non python files like css files?

ashahba:
The `non-py` files are just `whitespace` fixes like the ones we did during #2371

jlewi:
SGTM thanks.

PR #2455, opened 2019-02-12
add basic auth as alternative of iap access to deploy app; 

kunmingg:
1. add basic auth as alternative of iap access to deploy app; 
2. makefile clean up

TODO: update docs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2455)


jlewi:
/lgtm
/approve
But lets file an issue to load the hashed password from a config map and not as a ksonnet parameter so it doesn't get checked into source control.

Issue #1525, opened 2018-09-13
Nodeport not accesible


PhilippeBarb:
Hello and sorry for my bad english :(

I create a deployment with yaml file 

`---
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   name: nginx
 spec:
   replicas: 2
   template:
     metadata:
       labels:
         app: nginx
     spec:
       containers:
         - name: nginx
           image: nginx
           ports:
             - containerPort: 80

---
apiVersion: v1
kind: Service
metadata:
  name: nginx 
  labels:
    name: nginx
spec:
  type: NodePort
  ports:
    - port: 80
      nodePort: 30080
  selector:
    name: nginx      `

I can made a curl on pod's ip but not with ip_cluster:30080

kubectl version
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T18:02:47Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T17:53:03Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}

In advance very thanks

Philippe

PhilippeBarb:
Other info 

 microk8s.kubectl get services
NAME         TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)          AGE
nginx       NodePort    10.152.183.221   <none>        3000:30500/TCP   2h

 microk8s.kubectl get pods

 kubectl get pods
NAME                       READY     STATUS    RESTARTS   AGE
nginx-578687d985-cxcjw   1/1       Running   0          2h
nginx-578687d985-tknf5   1/1       Running   0          2h

kunmingg:
Nodeport is accessible from any node within cluster.

Issue #630, opened 2018-04-10
Ingress/Networking for on prem K8s deployments


jlewi:
In on prem deployments; users may not have a load balancer. Their Kubernetes nodes might be reachable from their corporate network though. So one way to do ingress in this case would be 

- Expose the Envoy/Ambassador proxy service via NodePort
- User connects to envoy/Ambassador proxy using node IPs or configures their DNS server to point at those nodes

I think one issue we need to fix is that I don't think we currently define routes in our Envoy/Ambassador proxy for JupyterHub.

/cc @ddutta

PR #2457, opened 2019-02-12
Add optional parameter ambassadorNodePort. 

iankoulski:
Useful when ambassadorServiceType is NodePort. Allows fixing the port number to a specific value which can be used when configuring a load balancer or another application that connects to this port. This value should be set to a number between 30000 and 32767. By default it is set to 30080. If not set and ambassadorServiceType is NodePort, a random available port value within the nodePort range will be assigned.

Contributes to closing the following issues:
https://github.com/kubeflow/kubeflow/issues/630
https://github.com/kubeflow/kubeflow/issues/1525

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2457)


iankoulski:
CLA Signed!

iankoulski:
Added alternate email address to GitHub

iankoulski:
CLA Signed

kubeflow/common/prototypes/ambassador.jsonnet
```diff
@@ -5,6 +5,7 @@
 // @param name string Name
 // @optionalParam platform string none supported platforms {none|gke|minikube}
 // @optionalParam ambassadorServiceType string ClusterIP The service type for the API Gateway.
+// @optionalParam ambassadorNodePort number 30080 The port number to expose ambassador on when ambassadorServiceType is NodePort
```
kunmingg: Thanks for the PR!
Can you change default value of ambassadorNodePort to null and only add
```
nodePort: params.ambassadorNodePort
```
When ambassadorServiceType is NodePort and ambassadorNodePort is not null?

kubeflow/common/ambassador.libsonnet
```diff
@@ -20,6 +20,10 @@
             name: "ambassador",
             port: 80,
             targetPort: 80,
+            [if (params.ambassadorServiceType == 'NodePort') && 
+                (params.ambassadorNodePort >= 30000) && 
+                (params.ambassadorNodePort <= 32767) 
+             then 'nodePort']: params.ambassadorNodePort,
```
kunmingg: ```: params.ambassadorNodePort,```should be within if condition as well?

kubeflow/common/prototypes/ambassador.jsonnet
iankoulski: Hi @kunmingg, I've implemented the change with the conditions described above. I kept the ambassadorNodePort parameter as a number with default value 0. If the ambassadorServiceType is changed to NodePort, by default, a random available port will be picked by k8s. The nodePort will be included in the configuration only if the parameter is set to a valid number from the range.

kubeflow/common/ambassador.libsonnet
iankoulski: Apparently If the the if statement evaluates to false, libsonnet will not include the rest of the line in the yaml. I left the code this way and tested to make sure it behaves properly in the different scenarios.

iankoulski:
In order to correct the issue with my CLA and also catch up with the changes in the master branch, I had to create a new pull request: https://github.com/kubeflow/kubeflow/pull/2511
Closing this one, please review the new pull request instead.
Thank You!

PR #2462, opened 2019-02-12
fix test name on testgrid

lluunn:
for https://github.com/kubeflow/testing/issues/301

The test name on testgrid has requirement:
1. kfctl and kfctl-beta both running at the same time, the test name cannot collide
2. cannot use the name of workflow which is specific per run, e.g. kubeflow-presubmit-kubeflow-e2e-minikube-2447-0c1503a-5646-cec4

So I am creating a new param, workflowName, for each e2e test. kfctl and kfctl-beta will have different workflowName with this.

/cc @jlewi 
/cc @richardsliu 

/hold
waiting for the presubmit result

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2462)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -11,6 +11,8 @@ workflows:
     - bootstrap/*
     - kubeflow/*
     - testing/*
+    params:
+      workflowName: deployapp
   # kfctl test runs tests on gke.
   - app_dir: kubeflow/kubeflow/testing/workflows
     component: kfctl_test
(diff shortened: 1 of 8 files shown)


lluunn:
/hold cancel
See https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-presubmit&sort-by-name=
run 5662
The test name are fixed now.

Issue #2456, opened 2019-02-12
Issues with the current Go notebook controller


vkoukis:
Hello @lluunn @jlewi 

@kimwnasptd , @ioandr and myself are playing with the new Jupyter web app to make it work end-to-end with the new Go-based notebook controller, so we can then proceed with a final review and merge of #2357.

I am writing to summarize a few things we bumped into, it's a single GitHub issue to keep the discussion focused, but If you feel it's best to open distinct issues for them, please let us know.

* The controller seems to have a bug in the way it attempts to override the container spec ([line 114](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/pkg/controller/notebook/notebook_controller.go#L144)), which leads to the set values never having any effect. 
* The controller seems to hardcode the fact that the Pod always listens to port 8888 ([line 154](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/pkg/controller/notebook/notebook_controller.go#L154)). The deeper issue here is, "what can the controller expect of the application running in the pod?", and hardcoding the value to 8888 seems contrary to the goal of allowing arbitrary Images to work -- #2208. I will come back to this later, but the easiest way to solve this would be: The controller uses a default value of 8888 for `containerPort` , the user is *not* required to provide it with in the submitted Notebook CR, but they can always override it. If I understand correctly, the controller can access this value as `Notebook.spec.template.spec.containers[0].containerPort`. The controller should accept an explicit port choice by the user, because they know better what their presumably custom Image does, but should not require it.
* The controller creates a new Service but does not create annotations for Ambassador, thus there is no way for the user to reach their new Notebook from outside the cluster.
* This may be off-topic, but it was rather counter-intuitive to us: when Ambassador processes a new Mapping, it seems to not actually query the Service for its Endpoints, but instead expects the Mapping to specify the endpoint port explicitly. I understand the controller already know the (potentially custom) containerPort for the Notebook (see above), so it can just include it in the Ambassador-specific annotation that it sets on the new Service instance that it creates.
* When setting the Ambassador Mapping, the controller must make sure to enable `use_websocket: true`
because the Jupyter UI requires it and breaks without it.
[...]

Issue #2417, opened 2019-02-06
notebook controller Add CRD validation


lluunn:
The first PR is merged #2336, things to improve:

- [x] e2e test (we have one test for the jsonnet-metacontroller one, we should make it run on this one)
- [x] status field should reflect the error if there is any. See #2269.
- [ ] ttlSecondsAfterFinished: This is in the original jsonnet controller spec, but not being used yet. I think we want to cleanup the notebook after idle for a period?
- [ ] CRD validation. 

Istio integration: #2172 , #2173

jlewi:
Lets scope this issue to adding notebook CRD Validation.
We can use separate issues to track
  * Istio integration

I don't know what ttlSecondsAfterFinished was intended for; I'm not sure its relevant any more. We use statefulsets for jupyter and I don't see why we would need a TTL.

jlewi:
/assign @zabbasi 

Looks like @zabbasi has some PRs pending to fix the CRD validation.

lluunn:
The ttl was that we want to kill the idle notebook pod after some time; why statefulsets don't need to have ttl?

zabbasi:
There is another issue which takes care of idle notebooks: https://github.com/kubeflow/kubeflow/issues/1394

zabbasi:
> The ttl was that we want to kill the idle notebook pod after some time; why statefulsets don't need to have ttl?

There is another issue which takes care of idle notebooks: #1394

jlewi:
@lluunn I was thinking the TTL was used not to GC idle pods but to garbage collect terminated resources.

With Argo workflows, TFJobs, batch jobs and other resources where the resources are run to completion we run into problems because we eventually accumulate so many resources that the API server can no longer efficiently list those resources and get requests hang.

So to address that the pattern is to add a ttlSecondsAfterFinished which GC's completed resources after the TTL expires.

WIth notebooks we don't have that problem. As @zabbasi says I think we can use #1394 to track culling of idle notebooks which is a more involved change.

jlewi:
Fixed by #109.

MartinForReal:
related #5152

PR #2454, opened 2019-02-12
WIP notebook e2e test with golang implementation

lluunn:
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2454)

PR #2463, opened 2019-02-12
 Notebook controller fixes

lluunn:
for #2456

- ~~Add field, useLab, and set jupyterlab environment var when it's true~~
- set containerPort only when it's not set
- add ambassador annotation
- make e2e test run on the new golang impl
- fix the Service clusterIP immutable problem

/cc @jlewi @vkoukis

also related: #2417

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2463)


lluunn:
/hold
I will incorporate #2454 into this one so that we have e2e test running for go notebook controller

kimwnasptd:
Hi @lluunn, there are two last changes that are demo critical in order for the Jupyter UI to work seamlessly. 
* While we were testing, we realized that Ambassador doesn't resolve the annotated Service's Endpoints. This means that if we put only the Service's name in the annotation it won't send the traffic to Port 8888 in the Pod but in Port 80

kimwnasptd:
Hi @lluunn, there are two last changes that are demo critical in order for the Jupyter UI to work seamlessly.

* While we were testing, we realized that Ambassador doesn't resolve the annotated Service's Endpoints. This means that if we put only the Service's name in the annotation it won't map the traffic on Port 8888 in the Pod but in Port 80. I think we can solve this if we also specify the Port in the annotations, for ex. `service: <name>.<namespace>:<port>` (also the name should be prior to the namespace)
* We need to set the `securityContext` in the Pod. Specifically we need `fsGroup` to be 100 so that the volumes are owned from `users` GUID. This is needed for volumes to work properly on GKE

kimwnasptd:
Hi @lluunn, I noticed in the commits that in the annotations the Notebook URL will be under `/notebook/<namespace>/<name>`. Previously the URLs for notebooks was just under `/<namespace>/<name>`. Was the `/notebook/` prefixed discussed somewhere and I missed it?
Could you have me in the loop similar changes?

lluunn:
No I just added. I think other components might have service with same (namespace, name) pair, so it seems reasonable to me to have a prefix indicating it's a notebook.
WDYT?

kimwnasptd:
Totally agree with it! I will update `connect(url)` function in the UI to redirect to the new URL for the Notebooks

jlewi:
@lluunn just had 2 minor comments and then this is ready to go IMO.

lluunn:
Update the comment, and put back main in pytest, thanks

jlewi:
/lgtm
/approve
/hold
In case you want to move logging statements to top of module or into the test function so that logging works even if executed under pytest

lluunn:
The logging is already at top of the module.
/hold cancel
thanks!

PR #2474, opened 2019-02-14
Support multiple CRD versions for TFJob

richardsliu:
Fixes https://github.com/kubeflow/tf-operator/issues/932

This also gets rid of the tfJobVersion parameter in the tf-job-operator component. We should only need to deploy one operator that handles all API versions.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2474)


johnugeorge:
minikube tests failed at the same step

PR #2478, opened 2019-02-14
Refactor the Kubeflow Pipeline ksonnet to better support upgrade

IronPan:
Move the pipeline parameter to pipeline.libsonnet so only user specified parameter is persisted in 
components/params.libsonnet. This allow upgrading pipeline by deleting ksonnet package only. 
For more details on pipeline upgrade, check this script
https://github.com/kubeflow/kubeflow/pull/2479/files#diff-9e6f5278480335e834537d84293b656b

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2478)


IronPan:
cc @neuromage

jlewi:
The K8s cluster is created during the apply phase.

So if you add an environment during the generate phase how are you going to deal with the fact that a K8s cluster may not exist? Are you going to point it at a dummy cluster and then change that during the apply phase?

scripts/gke/util.sh
```diff
@@ -233,27 +233,17 @@ gcpGenerateKsApp() {
   ks generate cloud-endpoints cloud-endpoints
   ks generate cert-manager cert-manager --acmeEmail=${EMAIL}
   ks generate iap-ingress iap-ingress --ipName=${KUBEFLOW_IP_NAME} --hostname=${KUBEFLOW_HOSTNAME}
-  ks param set jupyter jupyterHubAuthenticator iap
-  ks param set pipeline mysqlPd "${DEPLOYMENT_NAME}-storage-pipeline-db"
-  ks param set pipeline nfsPd "${DEPLOYMENT_NAME}-storage-pipeline-nfs"
+  ks param set jupyter jupyterHubAuthenticator iap --env=default
```
jlewi: Why are you setting these parameters only in the environment default?

IronPan:
I thought the cluster is created at the "generate platform" step. This change is at "generate k8s" step. Please correct me if I miss anything (on-prem maybe?)

scripts/gke/util.sh
IronPan: the immediate reason is that i want to be able to persist the configuration if I upgrade the jupyter hub. otherwise it might be lost. 

does it make sense for all customizations(non default parameters) to set to a specific environment?

IronPan: this step would remove the customized parameter
https://github.com/kubeflow/kubeflow/pull/2479/files#diff-9e6f5278480335e834537d84293b656bR32

jlewi:
No infrastructure is created during generate. Only config files are created.

ksonnet tries to talk to the master when creating a new environment. So the K8s cluster needs to exist or you need to somehow fake it (or maybe there's some flag to disable that). The K8s cluster won't exist until affter we run apply platform.

IronPan:
sorry i had a typo in my previous reply. 

the cluster is created at the "**apply** platform" step. This change is at the next "generate k8s" step.

scripts/gke/util.sh
jlewi: Ok so you're depending on environment specific overrides to keep track of user defined overrides of parameters so you can preserve them on update.

Are you planning on adding an E2E test to verify that behavior to ensure refactorings don't break it?

jlewi:
So the intended order of operations is

1. generate platform
1. generate k8s
1. apply platform
1. apply k8s

In other words it should be possible to generate configs without depending on any infrastructure existing.

We don't have a good pattern right now for reoptimizing based on your actual cluster. 

So if you want to make that work I think you need to

1. During generate call `ks env add` with `--api-spec` and point it at the URL for the appropriate K8s version spec.
1. During apply reset the server associated with your environment.

scripts/gke/util.sh
jlewi: I think another option that might be better would be to define the default parameters inside your .libsonnet file e.g. in your libsonnet

```
params = defaultParams + overrides
```

Don't define optional parameters inside your prototype so parameters will only be defined when the user explicitly decides to override a parameter.

The advantage of this approach is that it doesn't depend on using specific environments so I think its less brittle.

IronPan: I am not sure I entirely follow the idea here, but here are some questions
> define the default parameters inside your .libsonnet file

If so would user not be able to `ks param list` to get the parameter default values?

> `params = defaultParams + overrides`

Where should I store the overrides and ensure the override value can be persist when I delete the component?

Given that we rely on default environment and we are on the way of deprecating ksonnet, it's ok to use default environment as the current approach?

IronPan:
My understanding how ksonnet app work is when initializing a ksonnet app with `ks init` https://github.com/kubeflow/kubeflow/blob/master/scripts/util.sh#L62
it requires a active kubeconfig context set correctly, otherwise I see
``` 
++ ks init ks_app --skip-default-registries
ERROR No current context found. Make sure a kubeconfig file is present
```
https://github.com/ksonnet/ksonnet/issues/251

The order you describe work because ks init was using with context from an old cluster. This is probably not what we intend to do. 

Should we stick with the order in the official document
https://www.kubeflow.org/docs/started/getting-started-gke/#deploy-kubeflow-on-gke-using-the-command-line

scripts/gke/util.sh
jlewi: > If so would user not be able to `ks param list` to get the parameter default values?
Yes.

> Where should I store the overrides and ensure the override value can be persist when I delete the component?

You would define the default values in the .libsonnet file for pipelines; e.g like we do for tf serving
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/tf-serving/tf-serving.libsonnet#L5

User defined overrides would still be defined in the params.libsonnet files but they wouldn't necessarily be tied to a specific environment. 

> Given that we rely on default environment and we are on the way of deprecating ksonnet, is it ok to use default environment as the current approach?

Up to you but I think the current approach is more brittle. Specifically, you are relying on logic in kfctl to correctly set the parameters for just the default environment so that defaults are still preserved in the components/params.libsonnet file. 

At a minimum you'll also need to reimplement this logic in the kfctl go binary.

The approach I suggest doesn't have that brittleness; and it doesn't rely on users only overriding parameters in a specific environment.

jlewi:
The desired order of operations is generate and then apply.
The instructions in the GKE guide are a work around because we never got around to implementing the fix to run ks init with out an existing cluster. We have other places where we do it.

I think it would be a mistake to assume the cluster exists; that will break things in the future.

scripts/gke/util.sh
IronPan: thanks for the sample. it's helped me understand the idea. done.

IronPan:
I've use the alternative approach you suggested and there is no need to create ks env upfront.
thanks

jlewi:
nit: Could you please update the PR description and title.

jlewi:
/lgtm
/approve
/hold 

Please update the PR title and description before canceling the hold and merging.

PR #2482, opened 2019-02-14
add config file entries for deployment on GCP

kunmingg:
Define ksonnet app used by kfctl to deploy kubeflow on GCP
Provide 2 entries:
* IAP
* Basic Auth

File can be loaded as [KsonnetSpec](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/apis/apps/ksonnet/v1alpha1/application_types.go#L201)

Package is not provided as [default pkg set](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/apis/apps/ksonnet/v1alpha1/application_types.go#L63) should be good now.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2482)


kunmingg:
@kkasravi 
Seems https://github.com/kubeflow/kubeflow/blob/master/bootstrap/cmd/kfctl/config/app.yaml is not compatible with https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/apis/apps/ksonnet/v1alpha1/application_types.go#L236

PR #2499, opened 2019-02-16
Readme added for Kubeflow Central-Dashboard Component

avdaredevil:
Added a readme for Kubeflow Landing Page

cc: @jlewi 
/assign @avdaredevil 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2499)


components/centraldashboard/README.md
```diff
@@ -0,0 +1,13 @@
+# Kubeflow Landing-Page
+This component is the dashboard for Kubeflow deployents
+
+## Getting Started
+For devmode, do these steps to spin up a stand-alone server locally:
+- Clone the repo
```
yebrahim: I'd just say you need npm, then you can do npm install in this directory, bower install, then serve the static files using any server.
Even better, how about including `bower install` in npm's post-install script, and including "serve" as a dev dependency, such that doing `npm start` serves the files using it?

jlewi:
Looks like you need to rebase.

Also it looks like you are doing slightly more than just adding a README; could you please update the PR title and description?

Could you also respond to @yebrahim's comment

/lgtm
/approve

components/centraldashboard/README.md
avdaredevil: This will happen with the new issue: #2512. For now, this is the best set of instructions for the site to load in dev mode

avdaredevil:
Rebased! Please /lgtm

components/centraldashboard/README.md
yebrahim: This is more about simplifying the UI requirements than the backend. Devs should not be required to install bower globally (and managing its version) if it can be done as part of npm. And owning an easy way to serve the static file is always a good idea.

This is a nit though, I won't block on it.

PR #2505, opened 2019-02-19
boostrapper install istio by default

lluunn:
For #2052 

/cc @kunmingg 
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2505)


jlewi:
@lluunn Is this going to affect what gets deployed via Click to Deploy? If it is how can we guard that change so that we don't change what gets deployed for 0.4? Only for later versions should we change what gets deployed.

lluunn:
ah good point
/hold

kunmingg:
@lluunn 
Do we have istio available on kfctl binary?

lluunn:
@kunmingg not yet
filed @2507

jlewi:
@lluunn In that case the best bet might be to add it to kfctl go binary first. @gabrielwen is going to work on adding new handlers to click to deploy to use the new go binary. So that might help guard it.

PR #2508, opened 2019-02-19
Update kubespawner property name

knkski:
kubespawner deprecated the `image_spec` property name in 0.10, in favor of just a plain `image` property name. Update the `KubeFormSpawner` property with that naming, and proxy the deprecated names to the new property.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2508)


jlewi:
/assign @jlewi 
@knkski We actually removed JupyterHub in 0.5
So if you wanted to make this change you might want to do it on one of the earlier release branches 0.4. 
On master I think we could just go ahead and delete this code.

Issue #2208, opened 2019-01-04
Make arbitrary Jupyter images work with Kubeflow


jlewi:
There's a lot of good jupyter images already out there.

For example #767 discusses a Jupyter image with swift support for TensorFlow.

We'd also like to support images with R.

As discussed in this [comment](https://github.com/kubeflow/kubeflow/issues/1643#issuecomment-425170924)

I think we want to

1. Document what requirements needed to be satisfied for an image to work with Kubeflow
1. Removing/reducing these requirements.

Ideally, any Jupyter runtime image should just work with Kubeflow. I think the only tweaks we've made are related to

1. filesystem permissions
1. support for backing /home/jovyan is backed by PV
1. requirements created by JupyterHub

I suspect backing /home/joyvan with a PV is the main compatibility issue with most arbitrary Jupyter images. I think we can work around this just by having an option by not mounting the PV at /home/jovyan but instead in a subdirectory. We already have an option at the ksonnet level that configures this for all jupyter instances; but we could potentially surface it in the spawner UI to allow it to be configured on a per instance database.

I think JupyterHub might impose some additional constraints; e.g. at a minimum I think JupyterHub needs to be installed in the Jupyter image itself and the version must match the JupyterHub version.

Its possible this requirement will go away in 0.5 if we remove JupyterHub.

jlewi:
The Jupyter notebook CR is available in 0.4. It would be good to try that out with some arbitrary notebook images starting with the stock TensorFlow images which include jupyter.

jlewi:
If someone wants to help out, a great way would be to try running a PyTorch notebook with our CR; see #1057

DavidSpek:
As said in the last part of my comment https://github.com/kubeflow/community/issues/344#issuecomment-697370531, the only change that is needed to run the standard jupyter images is the `CMD` command. After some playing around the only the only variables necessary to still get a working notebook are:
`CMD ["sh","-c", "jupyter lab --NotebookApp.token='' --NotebookApp.base_url=${NB_PREFIX}"]`
[...]

Issue #1130, opened 2018-07-05
TFMA plots don't render; GET tfma_widget_js.js returns 404


jlewi:
Jupyter image:
gcr.io/kubeflow-images-public/tensorflow-1.7.0-notebook-cpu:v20180607-476e150e

I tried running the Chicago example:
https://github.com/tensorflow/model-analysis/blob/master/examples/chicago_taxi/chicago_taxi_tfma_local_playground.ipynb

If you run all cells they all run successfully. The final cells are supposed to render plots. However, no plots show. If you look at the developer console we see the following javascript errors.

```
GET http://localhost:8000/user/jlewi/static/tfma_widget_js.js?v=20180705135002 404 (Not Found)

/user/accounts.google.com%3Ajlewi@google.com/nbextensions/jupyter-js-widgets/extension.js?v=20180705131450:21583 Error: Could not create a model.
    at promiseRejection (/user/accounts.google.com%3Ajlewi@google.com/nbextensions/jupyter-js-widgets/extension.js?v=20180705131450:21583)
promiseRejection @ /user/accounts.google.com%3Ajlewi@google.com/nbextensions/jupyter-js-widgets/extension.js?v=20180705131450:21583

Uncaught (in promise) Error: Script error for "tfma_widget_js"
http://requirejs.org/docs/errors.html#scripterror
    at makeError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:165)
    at HTMLScriptElement.onScriptError (require.js?v=6da8be361b9ee26c5e721e76c6d4afce:1732)
````

It looks like the TFMA Jupyter plugin isn't properly activated.

**Work Around**

Running the following commands to create a shell in the Jupyter pod and install the widget appears to fix the issue

```
kubectl exec -it <Jupyter pod>
source activate py2
jupyter nbextension install --py --symlink tensorflow_model_analysis
jupyter nbextension enable --py tensorflow_model_analysis 
```

These commands are supposed to be run when we build the docker image
https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/Dockerfile#L124

* Were these commands not run at build time?
* Is there a way to check which extensions are installed and enabled?

/cc @ojarjur @pdmack @amygdala

**Comment by jlewi** [comment]
```
jupyter nbextension list
Known nbextensions:
  config dir: /home/jovyan/.jupyter/nbconfig
    notebook section
      jupyter-js-widgets/extension  enabled 
      - Validating: OK
[...]

PR #2517, opened 2019-02-20
Change jupyter images to work with the new jupyter CR, issue# 2458

zabbasi:
Modified entrypoint command for jupyter images to launch jupyter on boot.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2517)

### DIFF
--- a/components/tensorflow-notebook-image/start-notebook.sh
+++ b/components/tensorflow-notebook-image/start-notebook.sh
@@ -15,5 +15,10 @@
 
 set -e
 
-jupyter notebook --notebook-dir=/home/jovyan --ip=0.0.0.0 --no-browser --allow-root --port=8888
+if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
+  # launched by JupyterHub, use single-user entrypoint
+  exec /usr/local/bin/start-singleuser.sh $*
+else
+  . /usr/local/bin/start.sh jupyter notebook $*
+fi
 
--- a/components/tensorflow-notebook-image/start-notebook.sh
+++ b/components/tensorflow-notebook-image/start-notebook.sh
@@ -15,9 +15,5 @@
 
 set -e
 
-if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
-  # launched by JupyterHub, use single-user entrypoint
-  exec /usr/local/bin/start-singleuser.sh $*
-else
-  . /usr/local/bin/start.sh jupyter notebook $*
-fi
+jupyter lab --notebook-dir=/home/jovyan --ip=0.0.0.0 --no-browser --allow-root --port=8888
+
--- a/components/tensorflow-notebook-image/start-notebook.sh
+++ b/components/tensorflow-notebook-image/start-notebook.sh
@@ -15,5 +15,5 @@
 
 set -e
 
-jupyter lab --notebook-dir=/home/jovyan --ip=0.0.0.0 --no-browser --allow-root --port=8888
+jupyter notebook --notebook-dir=/home/jovyan --ip=0.0.0.0 --no-browser --allow-root --port=8888
 
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -147,7 +147,7 @@ WORKDIR $HOME
 RUN rm -fr $(ls -A $HOME)
 
 # Copy over init scripts
-COPY --chown=jovyan:users start-singleuser.sh start-notebook.sh start.sh pvc-check.sh /usr/local/bin/
+COPY --chown=jovyan:users  pvc-check.sh /usr/local/bin/
 RUN chmod a+rx /usr/local/bin/*
 
 RUN docker-credential-gcr configure-docker && chown jovyan:users $HOME/.docker/config.json
@@ -155,4 +155,5 @@ RUN docker-credential-gcr configure-docker && chown jovyan:users $HOME/.docker/c
 # Configure container startup
 EXPOSE 8888
 ENTRYPOINT ["tini", "--"]
-CMD ["start-notebook.sh"]
+CMD ["sh","-c", "jupyter notebook --notebook-dir=/home/jovyan --ip=0.0.0.0 --no-browser --allow-root --port=8888"]
+


jlewi:
@zabbasi If this is ready for review please remove WIP from the title.

components/tensorflow-notebook-image/start-notebook.sh
```diff
@@ -15,9 +15,5 @@
 
 set -e
 
-if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
-  # launched by JupyterHub, use single-user entrypoint
-  exec /usr/local/bin/start-singleuser.sh $*
-else
-  . /usr/local/bin/start.sh jupyter notebook $*
-fi
+jupyter lab --notebook-dir=/home/jovyan --ip=0.0.0.0 --no-browser --allow-root --port=8888
```
jlewi: Is this starting the jupyter lab interface by default? I think we decided we wanted to stick with Jupyter by default.

vkoukis: I understand why you want to have the classic notebook interface be the default, and I agree.
It would be great if we can have a way for the user to choose.
Based on the discussion at https://github.com/kubeflow/kubeflow/pull/2463#pullrequestreview-204068952, you preferred an approach where we didn't have a Jupyter-specific section in the Notebook CRD.
Since we agree that the best way is to have a pre-baked default inside the images, for the type of environment to use, can we have two sets of images being built and published by Kubeflow automatically? One built with classic Jupyter notebook being the default, one built with JupyterLab being the default, so users can switch easily without having to rebuild images?

The proposed image will be the one with the classic interface, so the current TFX widgets work, but having an alternative readily available lifts a major obstacle to working with JupyterLab, having to rebuild the images.

components/tensorflow-notebook-image/start-notebook.sh
```diff
@@ -15,9 +15,5 @@
 
 set -e
 
-if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
-  # launched by JupyterHub, use single-user entrypoint
-  exec /usr/local/bin/start-singleuser.sh $*
-else
-  . /usr/local/bin/start.sh jupyter notebook $*
-fi
+jupyter notebook --notebook-dir=/home/jovyan --ip=0.0.0.0 --no-browser --allow-root --port=8888
```
lluunn: Why not just remove this file and put this line in Dockerfile?
https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/Dockerfile#L157

zabbasi: start-notebook.sh is being referenced in Dockerfile of kubeflow/componnets/contrib images. So I reverted changes for this file.

vkoukis: @lluunn , @zabbasi This is a great fix, because it exposes the full arguments to Jupyter in the entry point, without changing the actual contents of the image, if I understand correctly.

So building one more image with a new entrypoint, should be a very light operation, which does not require creating new layers. Maybe we can have our cake and eat it this way:
* All images are self-standing, they just start with default settings if I run `docker run`
* Since TFX widgets do not work right now with JupyterLab, Kubeflow can propose to users that they use images that start the old interface.
* Users can just choose another image in the web app from the list, and switch to a new image, which has the same layers, but a different entrypoint, to start the current JupyterLab interface.

WDYT @jlewi ?

lluunn:
The unittest jsonnet test failed, seems unrelated
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-unittests-2517-96759ac-5818-e739?tab=workflow&nodeId=kubeflow-presubmit-unittests-2517-96759ac-5818-e739-3717395489

components/tensorflow-notebook-image/start-notebook.sh
jlewi: @vkoukis Building multiple images seems like a heavy hammer. One of my hopes is that we can make it easier to reuse any jupyter image see #2208. With that in mind when we do publish jupyter images we should do that in the context of some specific use case and make choices suitable for that application.

For example, I think it might make sense to have a TFX focused Jupyter image that would have the TFX libraries installed and make Jupyter the default for that reason.

components/tensorflow-notebook-image/start-notebook.sh
```diff
@@ -21,3 +21,4 @@ if [[ ! -z "${JUPYTERHUB_API_TOKEN}" ]]; then
 else
   . /usr/local/bin/start.sh jupyter notebook $*
```
jlewi: Should we delete this file if we don't need it anymore.

components/tensorflow-notebook-image/start-notebook.sh
jlewi: @vkoukis See #2208. If we're building jupyter images I think the question we should ask is why are we building Kubeflow specific Jupyter images and then based on that figure out the right thing to do.

For example the TensorFlow docker image comes with Jupyter and I believe starts it automatically. So why can't users just use that why does Kubeflow need to build a new Jupyter image?

Once we pick a reason such as "Jupyter Environment suitable for TFX" or Jupyter environment for PyTorch or NVIDIA Rapids, i think it will be easier to figure out which libraries should be installed in that notebook and what environment should be used.

Historically, the reason we got into the business of building Jupyter images is because we were using JupyterHub and that required customization of the images.

components/tensorflow-notebook-image/start-notebook.sh
jlewi: I see you filed: #2526

components/tensorflow-notebook-image/start-notebook.sh
vkoukis: > For example the TensorFlow docker image comes with Jupyter and I believe starts it automatically. So why can't users just use that why does Kubeflow need to build a new Jupyter image?

I'm not sure. I think the biggest blocker is that these generic images will not contain Kubeflow-specific components, like Fairing, which are included in images built by Kubeflow. I'll look more into it though.

> Once we pick a reason such as "Jupyter Environment suitable for TFX" or Jupyter environment for PyTorch or NVIDIA Rapids, i think it will be easier to figure out which libraries should be installed in that notebook and what environment should be used.

Agreed. I understand that currently you opt for Kubeflow to use the older Jupyter interface, so you maintain compatibility with TFX widgets that misbehave with JupyterLab [Are you referring to [this](https://github.com/kubeflow/kubeflow/issues/1130)? Maybe this will be fixed eventually in the widget code? ]
This is a certainly a strong argument.

I would also argue that it's important for Kubeflow to showcase the latest JupyterLab UI along with its components. I'll make a note in the next community meeting to solicit feedback, and we'd be happy to make a PR to Kubeflow so we can build and maintain JupyterLab-enabled images that also contain its custom components.

components/tensorflow-notebook-image/start-notebook.sh
vkoukis: @jlewi just saw this, I'd missed it.
> Building multiple images seems like a heavy hammer. One of my hopes is that we can make it easier to reuse any jupyter image see #2208.
I agree with this in general. I think Kubeflow can focus only on building Images that contain Kubeflow-specific components, e.g., Fairing.

On whether this is heavy handed, I'll experiment a bit with building different images starting from existing ones as a base. so we can minimize the size of extra layers. Changing an entry point will be essentially free, both in build and download time, but can prove beneficial.

>For example, I think it might make sense to have a TFX focused Jupyter image that would have the TFX libraries installed and make Jupyter the default for that reason.

Agreed.
I'll experiment a bit and come back with a PR for discussion. Thanks!

PR #2518, opened 2019-02-21
Added a windows equivalent script for the makefile for central-dashboard

avdaredevil:
Just here in case, someone finds a use for it. I rewrote the makefile to work in a windows environment portably including environment level checks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2518)


avdaredevil:
Since I was using this to do all my `make`'s in windows. I accidentally made my hotfix branch from this PR, and as a result, PR #2544 contains this change as well.

cc: @jlewi

PR #2525, opened 2019-02-21
Add PD params back to prototypes

gabrielwen:
It's more user friendly to have them in prototypes so that we could define them during `ks generate`.  This shouldn't have impact on the project to upgrade pipelines.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2525)

### DIFF
--- a/kubeflow/pipeline/pipeline.libsonnet
+++ b/kubeflow/pipeline/pipeline.libsonnet
@@ -12,9 +12,6 @@
     mysqlPvName: null,
     minioPvName: null,
     nfsPvName: null,
-    mysqlPd: null,
-    minioPd: null,
-    nfsPd: null,
   },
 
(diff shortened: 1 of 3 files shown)


IronPan:
@gabrielwen I think you could still override all the parameters listed [here](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/pipeline/pipeline.libsonnet#L4) without exposing them in pipeline.jsonnet.

Just run 
```
ks param set pipeline mysqlPd [...]
```

The reason we dont want to expose them to pipeline.jsonnet is so that they are not stored in components/params.libsonnet by default. It leads to brittle experience during upgrade.

gabrielwen:
closing this PR as discussed - change go binary instead.

PR #2404, opened 2019-02-06
Profiles e2e test

lluunn:
Fix #2176 

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2404)

PR #2533, opened 2019-02-22
Golang profile controller 

### LINKED ISSUE #2270: Go based implementation of the profiles custom resource?
Currently the profiles custom controller is implemented using metacontroller with the sync function implemented in jsonnet.

Should we switch to a go based implementation? I think the primary reason for switching to go would be that it has a much higher ceiling; i.e. if we want to do more complicated things either; we won't find ourselves limited by jsonnet. It should also be easier to write unittests.

@kkasravi WDYT

lluunn:
Fix #2270 

Changed the spec a little bit to
```
kind: Profile
metadata:
  name: john
  namespace: kubeflow
spec:
  namespace: john
  owner:
    kind: User
    name: john@foo.com
```

- I haven't added resource quota field. I am not sure if it makes sense for user to specify their own quota? Or should controller determine the quota somehow?
- removed nested spec.template.spec.

The controller:
1. creates the namespace
2. creates the role and rolebinding for the user in that namespace

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2533)


lluunn:
/assign @kkasravi @jlewi 
Ready for review.

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -0,0 +1,464 @@
...
+		return reconcile.Result{}, err
+	}
+
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: instance.Spec.Namespace,
+		},
+	}
+	if err := controllerutil.SetControllerReference(instance, ns, r.scheme); err != nil {
+		return reconcile.Result{}, err
```
kunmingg: What if the namespace is already exist and owned by another profile?

components/profile-controller/Makefile
```diff
@@ -0,0 +1,63 @@
...
+
+all: test manager
+
+# Run tests
+test: generate fmt vet manifests
+	echo "Skip test..."
+	#go test ./pkg/... ./cmd/... -coverprofile cover.out
+
+# Build manager binary
+manager: generate fmt vet
```
kkasravi: I tried `make manager` and got
```
$ ☞  make manager
go generate ./pkg/... ./cmd/...
stat ../../vendor/k8s.io/code-generator/cmd/deepcopy-gen/main.go: no such file or directory
pkg/apis/apis.go:18: running "go": exit status 1
make: *** [generate] Error 1
```

I also tried `make generate` and `make docker-build` with similar errors.

The problem is deepcopy assumes a vendor directory exists which isn't the case for go modules.
This is due to the fact that kubebuilder doesn't generate a Makefile that works with go modules.

You may want to add the following targets to the Makefile
```
$(GOPATH)/bin/deepcopy-gen:
        GO111MODULE=off $(GO) get k8s.io/code-generator/cmd/deepcopy-gen

deepcopy: $(GOPATH)/bin/deepcopy-gen
        $(GOPATH)/bin/deepcopy-gen -i github.com/kubeflow/kubeflow/components/profile-controller/pkg/apis/... -O zz_generated.deepcopy
```

components/profile-controller/Makefile
```diff
@@ -0,0 +1,63 @@
...
+	go build -o bin/manager github.com/kubeflow/kubeflow/components/profile-controller/cmd/manager
```
kkasravi: I would change bin/manager to bin/profiler-controller

components/profile-controller/Makefile
```diff
@@ -0,0 +1,63 @@
...
+install: manifests
+	kubectl apply -f config/crds
+
+# Deploy controller in the configured Kubernetes cluster in ~/.kube/config
+deploy: manifests
+	kubectl apply -f config/crds
+	kustomize build config/default | kubectl apply -f -
+
+# Generate manifests e.g. CRD, RBAC etc.
+manifests:
```
kkasravi: this won't work since there is no vendor directory. I'll update with the correct command a bit later

components/profile-controller/config/default/kustomization.yaml
```diff
@@ -0,0 +1,49 @@
+# Adds namespace to all resources.
```
kkasravi: you may want to add these kustomization files to the new repo for kustomization manifests (kubeflow/manifests)

components/profile-controller/pkg/apis/kubeflow/v1alpha1/doc.go
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1alpha1 contains API Schema definitions for the kubeflow v1alpha1 API group
+// +k8s:openapi-gen=true
+// +k8s:deepcopy-gen=package,register
+// +k8s:conversion-gen=github.com/kubeflow/kubeflow/components/profile-controller/pkg/apis/kubeflow
+// +k8s:defaulter-gen=TypeMeta
+// +groupName=kubeflow.kubeflow.org
```
kkasravi: group name should probably be profiles.kubeflow.org

components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
```diff
@@ -0,0 +1,63 @@
...
+	// The namespace for this user
+	Namespace string `json:"namespace,omitempty"`
+
+	// The profile owner
+	Owner rbacv1.Subject `json:"owner,omitempty"`
+}
+
+// ProfileStatus defines the observed state of Profile
+type ProfileStatus struct {
+	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
```
kkasravi: we usually have something in here like 
```
        Conditions []ProfileCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type" protobuf:"bytes,6,rep,name=conditions"`
```

components/profile-controller/Makefile
lluunn: will it work if I just checkin vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go?

components/profile-controller/config/default/kustomization.yaml
lluunn: ack, will do later

components/profile-controller/pkg/apis/kubeflow/v1alpha1/doc.go
lluunn: I think it should be just kubeflow.org? updated

components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
lluunn: I will update status in a followup PR

components/profile-controller/config/crds/kubeflow_v1alpha1_profile.yaml
```diff
@@ -0,0 +1,46 @@
...
+  creationTimestamp: null
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: profiles.kubeflow.kubeflow.org
+spec:
+  group: kubeflow.kubeflow.org
+  names:
+    kind: Profile
+    plural: profiles
+  scope: Namespaced
```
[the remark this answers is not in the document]
lluunn: done. I make profile a cluster scope resource, and namespace's names = profile's name

components/profile-controller/pkg/controller/profile/profile_controller.go
lluunn: this is fixed now that profile is cluster resource

components/profile-controller/config/crds/kubeflow_v1alpha1_profile.yaml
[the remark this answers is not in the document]
jlewi: It looks like its still Namespace scoped.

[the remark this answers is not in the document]
lluunn: Sorry only updated ksonnet package.
yaml updated now

lluunn:
/hold
wait for test to be merged #2404

lluunn:
e2e test passed, ptal
@jlewi @kunmingg

kunmingg:
@jlewi @lluunn 
Seems we still have problem when the namespace is already exist and owned by another user (admin).
Create an issue to track: https://github.com/kubeflow/kubeflow/issues/2601

PR #2547, opened 2019-02-25
fixes 'kfctl (golang) Simplify the child/parent relationships in platforms'

### LINKED ISSUE #2536: kfctl (golang) Simplify the child/parent relationships in platforms
By making K8s a child of the platform KFApp we force every platform app to write logic to sequence the creation of the platform and the K8s resources.

I think a better approach would be to create an UberKFApp which would have a K8s KFApp and a Platform KFApp and know how to sequence the two.

kkasravi:
fixes #2536

This PR fixes the design flaw that we force every platform app to write logic to sequence ksonnet operations and platform operations for operations Apply, Delete, Generate and Init. For each platform (gcp, minikube) there was a high degree of code duplication.

This common code has been lifted into a 'Client' (./pkg/client/client.go) that sequences ksonnet and platform operations once.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2547)


kkasravi:
@jlewi responded to your review. 
Also not sure why this is failing in test-argo-deploy since the following PR succeeds.

jlewi:
Adding wip please remove it when its ready for another look.

kkasravi:
@jlewi ready for review - includes updates from your latest review

jlewi:
how bout [Compositor](https://en.wikipedia.org/wiki/Composite_pattern), Coordinator, Controller, Adaptor?

* I like **CompositeApp** better than **Compositor**; **Coordinator** works for me as well.
  * Not sure I have a preference. 
* I think the term **Controller** is overloaded; will create confusion with custom resource controller.

> The reason why I called it client is based on kubebuilder and k8 code generator [client-gen](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-api-machinery/generating-clientset.md#output-of-client-gen). When you specific // +genclient in application_types.go it will call client-gen which generates a client under pkg/client/clientset_generated. Since we're aspects of k8's code generation (deepcopy-gen) and our types are compatible with kubernetes runtime.Object (they have a metadata section, spec section) I thought having a client.go under pkg/client made sense.

Can you remind me why are using K8s code gen here?

jlewi:
/test all

You need to rebase to fix the authors.

PR #2548, opened 2019-02-25
fixes 'Add kustomize support to kfctl'

### LINKED ISSUE #2445: Add kustomize support to kfctl
Doc:
http://bit.ly/2RwMP8q

The immediate need is kfctl(go) needs to support installing ISTIO see #2052.
We currently install ISTIO using its YAML manifest. 
We could either add support to kfctl for installing YAML or we could just 

1. Add support for kustomize
1. Turn YAML manifests into kustomize packages by adding kustomize files 

To add support for kustomize we need to implement the KfApp interface for kustomize
https://github.com/kubeflow/kubeflow/tree/master/bootstrap/cmd/kfctl#kfapp-interface

See also #2351

kkasravi:
fixes 
- [x] #2445 Add kustomize support to kfctl
- [x] #2561 Move ksonnet, kustomize implementations out of KfApps.Children
- [x] #2560 Add a show command
- [x] #2653 kfctl generated yaml fails validation 
- [x] #2642 kfctl error in application component

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2548)


kkasravi:
additional functionality will follow

kkasravi:
@jlewi high number of files is because I had to rewrite author on an early commit I did so it needs to do it from one commit before to HEAD. I think that caused git to think merge from master files are part of this commit.

jlewi:
What if you rebase on master and squash all the commit? Does that get rid of unnecessary changes?

kkasravi:
This has too many spurious changes. See #2673

PR #2552, opened 2019-02-25
Updated Jupyter image paths for using in notebook controller

zabbasi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2552)


zabbasi:
Tested all cpu tagged images locally (and run some test on a Kubernete cluster) all could successfully launch the notebook and import tensorflow.

jlewi:
Thanks.

@zabbasi How come there's no 1.4 image?

zabbasi:
> Thanks.
> 
> @zabbasi How come there's no 1.4 image?

tensorflow could not be properly  imported in 1.4.1 . removed it per @lluunn suggestion as it is barely used.

lluunn:
@zabbasi I think you mentioned 1.4 has error? I think in that case we can just remove that.
I don't think we need to keep that many versions anyway.

PR #2555, opened 2019-02-26
Temporary change to pin kubeflow to the latest verified tf-operator commit 

zabbasi:
This is a temporary change so that we could merge [this PR](https://github.com/kubeflow/tf-operator/pull/945), which renames python top level package in tf-operator repo.  Once merged, it likely breaks kubeflow checkout tests since the tf-operator imports in external repos no longer are correct. 
To avoid this probelm, we temporarily pin checkout test to the latest tf-operator commit and will resume it once all import renaming in external repos are resolved.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2555)


zabbasi:
\assign @richardsliu

PR #2556, opened 2019-02-26
Add extra permissions to ClusterRole 'pipeline-runner'

elikatsis:
Hello,

as part of our work on kubeflow/pipelines#801. we need to include a few extra permissions to the service account which runs the pipeline (`pipeline-runner`) so it can `create`, `get` and `delete` `PVC`s and `VolumeSnapshot`s.

It can already `create` and `delete` `PVC`s, so we update the rule to add the verb `get`. We also make a new rule entry referring to `VolumeSnapshot` resources with the verbs `create`, `delete` and `get`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2556)

### DIFF
--- a/kubeflow/pipeline/pipeline-apiserver.libsonnet
+++ b/kubeflow/pipeline/pipeline-apiserver.libsonnet
@@ -230,6 +230,20 @@
           verbs: [
             "create",
             "delete",
+            "get",
+          ],
+        },
+        {
+          apiGroups: [
+            "snapshot.storage.k8s.io",
+          ],
+          resources: [
+            "volumesnapshots",
+          ],
+          verbs: [
+            "create",
+            "delete",
+            "get",
           ],
         },
         {


vkoukis:
Hello, a heads up for this!

It is a minor change to pipeline-runner's permissions so it can also handle PVCs and VolumeSnapshots, which we need so we can move with PR kubeflow/pipelines#926 and kubeflow/pipelines#801.

vkoukis:
@jlewi A heads up for this, as it has remained pending for quite some time, I should have mentioned it in the community meeting, but totally forgot. Any way we can push this forward?

hongye-sun:
@IronPan, it seems that I don't have permission to approve this PR. Could you grant the approve label? Thanks.

vicaire:
Assigning to @IronPan

vicaire:
Looks like I don't have permission either. @IronPan, could you please approve?

elikatsis:
The PR was a little old and I hadn't updated the branch, I believe that is the reason why the test failed.

I rebased on top of current master and force pushed.

PR #2558, opened 2019-02-26
Add status to notebook

lluunn:
- Add status condition "Ready" to notebook CR if no error
- e2e test verifies condition
- some refactoring

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2558)


lluunn:
ready for review
/assign @jlewi @kkasravi @richardsliu

jlewi:
@richardsliu Do you mind taking this review?

components/notebook-controller/pkg/controller/notebook/notebook_controller.go
```diff
@@ -123,18 +122,81 @@ func (r *ReconcileNotebook) Reconcile(request reconcile.Request) (reconcile.Resu
...
+		log.Info("Updating Service\n", "namespace", service.Namespace, "name", service.Name)
+		err = r.Update(context.TODO(), foundService)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	}
+
+	// Update the status if previous condition is not "Ready"
+	oldConditions := instance.Status.Conditions
+	if len(oldConditions) == 0 || oldConditions[0].Type != "Ready" {
```
richardsliu: Is the last condition at index 0 or -1?

kubeflow/jupyter/notebook_controller.libsonnet
```diff
@@ -23,6 +23,14 @@
           kind: "Notebook",
         },
       },
+      status: {
```
richardsliu: Consider using subresource status and then calling UpdateStatus(): https://github.com/kubeflow/tf-operator/pull/939

This prevents bugs where the previous status is unexpectedly overwritten as a result of an update.

lluunn: we are using status subresource. see line 17

components/notebook-controller/pkg/controller/notebook/notebook_controller.go
lluunn: I am doing prepending, (line 188), so should be 0.
I think prepending is convention, e.g. https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L136

lluunn:
Error: `gcloud auth activate-service-account` failed

components/notebook-controller/pkg/controller/notebook/notebook_controller.go
```diff
@@ -123,18 +122,81 @@ func (r *ReconcileNotebook) Reconcile(request reconcile.Request) (reconcile.Resu
...
+		err = r.Create(context.TODO(), ss)
+		justCreated = true
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	} else if err != nil {
 		return reconcile.Result{}, err
 	}
+	// Update the foundStateful object and write the result back if there are any changes
+	if !justCreated && util.CopyStatefulSetFields(ss, foundStateful) {
```
kunmingg: What if StatefulSet with name ```ss.Name``` already exist in the namespace and is created by other user / service?

lluunn: I think the reconcile will then fail, which is the intended behavior?

kunmingg: Ideally we should fail it and add message to status explaining why it fails.
Seems currently controller will just update StatefulSet?

lluunn: I think it will fail because we can't claim ownership on that statefulset.

PR #2564, opened 2019-02-27
new deploy API following Open API format

kunmingg:
New deployment API supports 2 types of entries:
* config entry
each entry maps to a set of config files stored in source repo or Github which can fully define a deployment entry. 
* deployment entry
each entry maps to a kubeflow instance on GCP which include resources like GKE, Endpoint, Storage, Service Account, IAM bindings. Deployment entry is created following definition in the corresponding config entry.

Goals of new API include simplify reuse configs and support configs customization. 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2564)

### DIFF
--- a/bootstrap/api/swagger.yaml
+++ b/bootstrap/api/swagger.yaml
@@ -20,54 +20,253 @@ tags:
 schemes:
 - "https"
 paths:
-  /projects/{project}/deployments:
+  /projects/{project}/configs:
     post:
       tags:
-      - "deployment"
-      summary: "Create a new kubeflow deployment on GCP"
-      description: ""
-      operationId: "createDeployment"
+        - "config"
+      summary: "Create a new kubeflow config entry on source repo"
+      security:
+        - BearerToken: []
+      operationId: "createConfig"
       consumes:
-      - "application/json"
+        - "application/json"
       produces:
-      - "application/json"
+        - "application/json"
       parameters:
-      - in: "path"
-        name: "project"
-        description: "Project needs to be created"
-        required: true
-        type: "string"
-      - in: "body"
-        name: "body"
-        description: "Deployment needs to be created"
-        required: true
-        schema:
-          $ref: "#/definitions/CreateDeployment"
-      responses:
-        200:
-          description: "successful operation"
+        - in: "path"
+          name: "project"
+          description: "gcp project string id"
+          required: true
+          type: "string"
+        - in: "body"
+          name: "body"
+          description: "Config needs to be created"
+          required: true
           schema:
-            $ref: "#/definitions/CreateOperation"
+            $ref: "#/definitions/KfConfig"
+      responses:
         405:
           description: "Invalid input"
           schema:
-            $ref: "#/definitions/InvalidInput"
+            $ref: "#/definitions/ErrorMessage"
+    get:
+      tags:
+        - "config"
+      summary: "List kubeflow config entries in target project"
+      security:
+        - BearerToken: []
+      operationId: "listConfigs"
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - in: "path"
+          name: "project"
+          description: "gcp project string id"
+          required: true
+          type: "string"
+      responses:
+        200:
+          description: "Kubeflow config entries"
+          schema:
+            $ref: "#/definitions/KfConfigEntries"
+        405:
+          description: "Invalid input"
+          schema:
+            $ref: "#/definitions/ErrorMessage"
+  /projects/{project}/configs/{config}/version/{version}/file:
+    get:
+      tags:
+        - "config"
+      summary: "Get kubeflow config entry file"
+      security:
+        - BearerToken: []
+      operationId: "getConfigFile"
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - in: "path"
+          name: "project"
+          description: "gcp project string id"
+          required: true
+          type: "string"
+        - in: "path"
+          name: "config"
+          description: "config entry name"
+          required: true
+          type: "string"
+        - in: "path"
+          name: "version"
+          description: "config entry kubeflow version"
+          required: true
+          type: "string"
+        - in: "query"
+          name: "filepath"
+          description: "file relative path"
+          required: true
+          type: "string"
+      responses:
+        200:
+          description: "Kubeflow config entries"
+          schema:
+            $ref: "#/definitions/KfConfigFile"
+        404:
+          description: "Entry not found"
+          schema:
+            $ref: "#/definitions/ErrorMessage"
+        405:
+          description: "Invalid input"
+          schema:
+            $ref: "#/definitions/ErrorMessage"
+    put:
+      tags:
+        - "config"
+      description: "Update kubeflow config entry file"
+      security:
+        - BearerToken: []
+      operationId: "putConfigFile"
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - in: "path"
+          name: "project"
+          description: "gcp project string id"
+          required: true
+          type: "string"
+        - in: "path"
+          name: "config"
+          description: "config entry name"
+          required: true
+          type: "string"
+        - in: "path"
+          name: "version"
+          description: "config entry kubeflow version"
+          required: true
+          type: "string"
+        - in: "body"
+          name: "body"
+          description: "file object to update"
+          required: true
+          schema:
+            $ref: "#/definitions/KfConfigFile"
+      responses:
+        405:
+          description: "Invalid input"
+          schema:
+            $ref: "#/definitions/ErrorMessage"
+  /projects/{project}/deployments:
+    post:
+      tags:
+        - "deployment"
+      summary: "Create a new kubeflow deployment on GCP"
+      security:
+        - BearerToken: []
+      operationId: "createDeployment"
+      consumes:
+        - "application/json"
+      produces:
+        - "application/json"
+      parameters:
+        - in: "path"
+          name: "project"
+          description: "gcp project string id"
+          required: true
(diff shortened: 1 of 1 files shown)


bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+        type: "string"
+      namespace:
+        type: "string"
+      cluster:
+        type: "string"
+      project:
+        type: "string"
+      zone:
+        type: "string"
+      token:
```
jlewi: Is the token the access token?
Should this be sent in the Auth header?

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+          schema:
+            $ref: "#/definitions/CreateOperation"
+        405:
+          description: "Invalid input"
+          schema:
+            $ref: "#/definitions/InvalidInput"
+definitions:
+  CreateDeployment:
+    type: "object"
+    properties:
```
jlewi: is this the body of the request?

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+      description: ""
+      operationId: "createDeployment"
+      consumes:
+      - "application/json"
+      produces:
+      - "application/json"
+      parameters:
+      - in: "body"
+        name: "body"
+        description: "Pet object that needs to be added to the store"
```
jlewi: Fix comment

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+tags:
+- name: "deployment"
+  description: "A kubeflow deployment on GCP"
+  externalDocs:
+    description: "Find out more"
+    url: "https://www.kubeflow.org/docs/started/getting-started-gke/"
+schemes:
+- "https"
+paths:
+  /e2eDeploy:
```
jlewi: Should we have a crud style API?  

We could model it on a combination of a K8s & GCP
https://cloud.google.com/deployment-manager/docs/reference/latest/deployments/insert

So should we have a create request at a URL like
/projects/${PROJECT}/deployments

And then a get at
/projects/${PROJECT}/deployments/${DEPLOYMENTNAME}

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - "application/json"
+      produces:
+      - "application/json"
+      parameters:
+      - in: "body"
+        name: "body"
+        description: "Pet object that needs to be added to the store"
+        required: true
+        schema:
+          $ref: "#/definitions/CreateDeployment"
```
jlewi: Following K8s and GCP style can the body of the Create request just be the spec for the deployment? 
i.e. a Create and Get request use the same object just with different fields.

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+        200:
+          description: "successful operation"
+          schema:
+            $ref: "#/definitions/CreateOperation"
+        405:
+          description: "Invalid input"
+          schema:
+            $ref: "#/definitions/InvalidInput"
+definitions:
+  CreateDeployment:
```
jlewi: Should we follow the K8s resource model?
So maybe the Deployment should have fields
metadata
    name
    namespace
spec:
    cluster
    zone
    ....

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+        type: "string"
+      ipName:
+        type: "string"
+      sAClientId:
+        type: "string"
+      extraComponents:
+        type: "array"
+        items:
+          type: "string"
+      parameters:
```
jlewi: Following K8s style should we add a status section and use conditions to report status?

bootstrap/api/swagger.yaml
jlewi: Perhaps instead of Create we should just have an Update method which would create it if it doesn't exist?

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - in: "body"
+        name: "body"
+        description: "Pet object that needs to be added to the store"
+        required: true
+        schema:
+          $ref: "#/definitions/CreateDeployment"
+      responses:
+        200:
+          description: "successful operation"
```
jlewi: One of the goals of the deployment process is to split it into separate phases generate and apply so that people can manually edit the manifests after generate in order to customize them.

Any thought how we could handle that?

One option would be to create custom verbs e.g generate & apply. But that goes against the K8s idea of having a common resource model with common verbs.

A second option would be to have a field in the spec that specifies the desiredState for the deployment. Allowed values could be something like "configured" "deployed"

The flow could then be something like
1. Call create with desiredState: configured
2. This creates the manifests and pushes them to source repo
3. User modifies the manifests
4. User calls Update setting the desiredState to deployed

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,317 @@
...
+      responses:
+        200:
+          description: "Kubeflow config entries"
+          schema:
+            $ref: "#/definitions/KfConfigEntries"
+        405:
+          description: "Invalid input"
+          schema:
+            $ref: "#/definitions/ErrorMessage"
+  /projects/{project}/configs/{config}/version/{version}/file:
```
jlewi: What is the version resource for?

Is it just fetching the config?

If users want to fetch the config why don't they talk to underlying storage (e.g source repo directly)? Is this just a helper API?

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,317 @@
...
+          schema:
+            $ref: "#/definitions/ErrorMessage"
+        405:
+          description: "Invalid input"
+          schema:
+            $ref: "#/definitions/ErrorMessage"
+    put:
+      tags:
+        - "config"
+      description: "Update kubeflow config entry file"
```
jlewi: Why do we need a put operation on config?

If people want to update the config couldn't they do this directly by modifying the file; e.g. in the source repo?

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,317 @@
...
+      clientId:
+        type: "string"
+      clientSecret:
+        type: "string"
+  KfConfig:
+    type: "object"
+    properties:
+      metadata:
+        $ref: "#/definitions/DeploymentMetadata"
+      extraComponents:
```
jlewi: Should KfConfig have a spec and status and follow the K8s resource model?

Status could provide information about things like version; e.g. if the underlying source has changed this could be reported via status.

Also it allows us to think of the resource as a controller where spec describes desired state and status provides actual state.

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,317 @@
...
+      storageOption:
+        $ref: "#/definitions/StorageOption"
+  KfConfigEntries:
+    type: "object"
+    properties:
+      configs:
+        type: "array"
+        items:
+          $ref: "#/definitions/DeploymentMetadata"
+  KfConfigFile:
```
jlewi: Per comment above what's this for?

bootstrap/api/swagger.yaml
```diff
@@ -0,0 +1,317 @@
...
+          name: "project"
+          description: "gcp project string id"
+          required: true
+          type: "string"
+        - in: "body"
+          name: "body"
+          description: "Deployment name"
+          required: true
+          schema:
+            $ref: "#/definitions/KfDeployment"
```
jlewi: How does kfDeployment relate to the Config?

Is KfDeployment just a pointer to the config?

bootstrap/api/swagger.yaml
kunmingg: This helper API provide read / write of single config file in underlying storage:
1. as of today for source repo there are not such API available.
2. existing edit UI is confusing because it aimed to solve more complicated day-to-day dev use case. 
3. If suggest user change config directly with underlying storage, user experience becomes:
    Create config entry in UI -> setup git and git clone -> edit config -> git push -> apply in UI

With this API we can have experience:
Create config entry in UI -> edit config file in UI -> apply in UI 
Then we provide whole experience within UI, would be much easier for usage like editing [GCP options](https://github.com/kubeflow/kubeflow/blob/master/deployment/gke/deployment_manager_configs/cluster-kubeflow.yaml#L52)

bootstrap/api/swagger.yaml
kunmingg: Relation between kfDeployment and KfConfig is like
relation between notebook instance and notebook image:
User can spawn up kfDeployment from existing KfConfig entries.

jlewi:
@kunmingg Thanks. Could you take a look at some of my previous comments such as adding a spec and status. 

If your blocked on getting started on implementation I think we can go ahead and merge this.

We will likely need to make some revisions in the future anyway after circulating it more broadly.

I think it would help to actually generate the API documentation (we could do that in a follow on PR) but  that is probably a lot more readable then the swagger spec.

/lgtm
/approve
/hold

kunmingg:
@jlewi 
To view API as docs:
https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/kubeflow/kubeflow/0f8a3babb6e0f8425dd5dff9432a68889911366a/bootstrap/api/swagger.yaml

jlewi:
Discussed with Kunming. We'll submit as is and iterate in follow on PRs

Some open questions

1. I'm not sure email should be a query parameter. 
    * I think it might be better to make it part of the KfConfig; e.g. have a YAML file somewhere with users/groups to be granted permission
    * For purposes of auditing the email query arg isn't good because it can't be audited.
    * Audit logs will automatically be generated for all actions based on the access token used.

1. Will move namespace into the KfConfig app.

PR #2571, opened 2019-02-27
WIP: Updating python package imports related to tf-operator

zabbasi:
We have renamed top level python package of tf-operator in [this PR](https://github.com/kubeflow/tf-operator/pull/945). Therefore, we need to update the related imports here.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2571)


testing/tf_job_simple_test.py
```diff
@@ -27,7 +27,7 @@
 import subprocess
 from kubeflow.testing import test_helper, util
 from kubernetes import client as k8s_client
-from py import tf_job_client
+arom kubeflow.tf_operator import tf_job_client
```
abhi-g: typo s/arom/from/

PR #2572, opened 2019-02-27
Refactor for Express + Webpack build stack.

prodonjs:
- Add webpack to process front-end code in public/
- Add TS and Express for server backend
- Update Dockerfile to run the server as Node.
- Fix issue with external links.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2572)


components/centraldashboard/Dockerfile
```diff
@@ -1,20 +1,13 @@
...
+COPY webpack.config.js *.json ./
 
-FROM alpine
-RUN apk --no-cache add ca-certificates
-WORKDIR /app
-COPY --from=builder /go/src/github.com/kubeflow/kubeflow/components/centraldashboard/ /app
-ENV PORT_1=8082
-EXPOSE 8082
-CMD [ "/app/centraldashboard" ]
+RUN npm install && npm run build
```
avdaredevil: Is there a reason to leave source in the image, when we already have the bundles?

components/centraldashboard/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "kubeflow-centraldashboard",
+  "version": "0.4.0",
```
avdaredevil: Let's use a seperate versioning for this component, doesn't make sense to link it to Kubeflow-Core's version.

Let's go with `v0.0.1`?

components/centraldashboard/package.json
```diff
@@ -0,0 +1,79 @@
...
+  "description": "Kubeflow Central Dashboard",
+  "scripts": {
+    "build": "NODE_ENV=production npx webpack && npm run tslint && npm run build-ts",
+    "build-ts": "tsc",
+    "build-webpack": "npx webpack",
+    "debug": "npm run build && npm run watch-debug",
+    "dev": "concurrently -k -p \"[{name}]\" -n \"Webpack,TypeScript,Express\" -c \"green.bold,yellow.bold,cyan.bold\" \"npx webpack-dev-server\" \"npm run watch-ts\" \"npm run watch-node\"",
+    "dev-debug": "concurrently -k -p \"[{name}]\" -n \"Webpack,TypeScript,Express\" -c \"green.bold,yellow.bold,cyan.bold\" \"npx webpack-dev-server\" \"npm run watch-ts\" \"npm run serve-debug\"",
+    "serve": "node dist/server.js",
+    "serve-debug": "nodemon --inspect dist/server.js",
```
avdaredevil: Maybe we should mention `nodemon` as a global dependency?
```
npm i -g nodemon
```

components/centraldashboard/public/components/main-page.js
```diff
@@ -0,0 +1,98 @@
+import { PolymerElement, html } from '@polymer/polymer/polymer-element.js';
```
avdaredevil: **_[nit]_** But can we do imports, and generally all objects like
```
import {PolymerElement, html} from '@polymer/polymer/polymer-element.js'
or
const {a, b} = obj
```
- No padding around imports / objects
- No semicolons

components/centraldashboard/public/components/main-page.js
```diff
@@ -0,0 +1,98 @@
...
+import '@polymer/neon-animation/animations/fade-in-animation.js';
+import '@polymer/neon-animation/animations/fade-out-animation.js';
+
+import css from './main-page.css';
+import template from './main-page.pug';
+
+export class MainPage extends PolymerElement {
+
+    static get template() {
+        return html([`<style>${css.toString()}</style> ${template()}`]);
```
avdaredevil: I really like this approach, good job!! 🎉

components/centraldashboard/public/components/main-page.pug
```diff
@@ -0,0 +1,43 @@
+app-drawer-layout.flex(narrow='{{narrowMode}}')
```
avdaredevil: Enforce pug templates to 4 spaces, so tabbing is more emphasized (since it's a tab based syntax)

components/centraldashboard/webpack.config.js
```diff
@@ -0,0 +1,161 @@
+'use strict';
+
+const { resolve } = require('path');
+const CleanWebpackPlugin = require('clean-webpack-plugin');
```
avdaredevil: We're not using harmony import statements here?

I'm assuming this file also runs via babel?

components/centraldashboard/Dockerfile
prodonjs: If your concern is the size of the build, I added `npm prune --production` to remove any of the devDependencies in node_modules/. That trims about 100MB from the image size. The base image for node:10 is about 900MB so we're not trimming a ton. We can optimize a bit more if we need to shrink this further.

components/centraldashboard/package.json
prodonjs: SGTM.

components/centraldashboard/package.json
prodonjs: It's included in devDependencies.

components/centraldashboard/public/components/main-page.js
prodonjs: That seems to work. I've always just stuck with the conventions from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import. Is there a reason why you explicitly want this permitted?

As an aside, we should probably add linting to the webpack configuration at some point to check these at build time.

components/centraldashboard/public/components/main-page.pug
prodonjs: Understood. I've updated this but I think this should eventually be something we lint for.

components/centraldashboard/webpack.config.js
prodonjs: As a matter of fact it doesn't. It gets loaded and used from the webpack process with is running in the standard Node runtime which doesn't support ES6 out of the box.

https://nodejs.org/docs/latest-v10.x/api/esm.html

components/centraldashboard/public/components/main-page.pug
avdaredevil: I agree! Looks good

components/centraldashboard/public/components/main-page.js
avdaredevil: If you read further down you'll see most examples mention it without the padding. It's kinda like the `{` on its own line or the same line. I think the latter _(without spaces)_ tends to be more common in OSS and is generally more condense, without sacrificing readability.

It really is not a big issue. Just something I am used to seeing/using in most codebases.

components/centraldashboard/Dockerfile
avdaredevil: We can look into it if it seems to be a concern. < 1GB is not bad

Issue #2574, opened 2019-02-27
presubmit error: ImportError: cannot import name tf_job_client


lluunn:
from https://github.com/kubeflow/kubeflow/pull/2404
```
from py import tf_job_client
ImportError: cannot import name tf_job_client
```

Seems related to https://github.com/kubeflow/tf-operator/pull/945?

lluunn:
seems temporary fix by #2555

lluunn:
We still need a proper fix and revert #2555 
/cc @richardsliu 
/assign @zabbasi

jlewi:
Is a proper fix to #2555 just editing the PYTHONPATH to include the new py directory in the tf-operator repository?

lluunn:
And also update test files' import

lluunn:
I still see presubmit failing with 
```
from py import tf_job_client
ImportError: cannot import name tf_job_client
```

lluunn:
maybe this line needs change: https://github.com/kubeflow/kubeflow/blob/f5dc02168240a6b4d4cf661adccc05d126fcb27d/testing/workflows/components/kfctl_test.jsonnet#L177

jlewi:
@zabbasi Is this fixed?

lluunn:
This is fixed

PR #2581, opened 2019-02-28
Updating python package imports related to tf-operator

zabbasi:
We have renamed top level python package of tf-operator in [this PR](https://github.com/kubeflow/tf-operator/pull/945). Therefore, we need to update the related imports here.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2581)

PR #2584, opened 2019-02-28
Fix presubmit

lluunn:
for #2574

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2584)


lluunn:
actually, @zabbasi is fixing that in #2581

Issue #2586, opened 2019-02-28
[kfctl] Consolidate kubeconfig and kfctl API calls


gabrielwen:
Currently we mixed uses of kubeconfig and cluster info from GCP.  Instead, in kfctl we should:
- Remove dependency of kubeconfig file from both kfctl CLI and Click-to-Deploy app.
- Add optionally kubeconfig support for kfctl CLI.  If user has both `gcloud` and `kubectl` installed, we help create cluster/context entry in user's local kubeconfig file.

gabrielwen:
Relates to #2370

gabrielwen:
Relates to #2535

jlewi:
What is click to deploy currently doing to get credentials to talk to the K8s master?

I think in click to deploy the incoming request with have an OAuth token which can just be forwarded to the K8s master.

With the CLI, I think we might need to depend on gcloud. Even if we are using the go client libraries the user would typically do `gcloud auth login` to setup the credentials that Go client libraries will use. So I think depending on gcloud is ok in the CLI setting.

We might need to think about how we make that pluggable so it can support CLI as well as click to deploy.

gabrielwen:
Right, I think click to deploy has an OAuth token.

I'm reading [this GoDoc](https://godoc.org/golang.org/x/oauth2/google), seems like we might be able to get auth token in CLI as well.  Would this be more feasible that we unify auth process by passing around auth tokens?  This also helps fix #2535

jlewi:
is this blocking for 0.5? Or should we punt it to 0.6 at this point?

gabrielwen:
it's not blocking.  I have some functions submitted for this but let's put it to 0.6

Issue #2585, opened 2019-02-28
[kfctl] Implements 3 way merge to CreateResourceFromFile or shell out to kubectl


gabrielwen:
Currently we implement `CreateResourceFromFile` in an elementary way such that:
- If deploying resource is there and we are not able to patch it, we try to delete and re-create it.
- If it's not there, we simply create it with `POST`.

We should at least consider:
- If the resource is there, we should take a look at config and see if we need to update it or just leave it as it is.

gabrielwen:
Related to #2370

jlewi:
I think we can shell out to `kubectl apply` if need be.

We definetly want the scripts to be re-entrant.

gabrielwen:
As discussed, we should:
- Check if `kubectl` exists, if it does, shell out to `kubectl apply`
- If it doesn't, should we bail out or give a warning message that it might not work perfectly?

jlewi:
@gabrielwen is this P1 or P2?

gabrielwen:
let's put it as P2 first.

jlewi:
Moving to 0.6 since its P2.

Issue #2535, opened 2019-02-22
kfctl (golang) Remove hard dependency on GOOGLE_APPLICATION_CREDENTIALS


kkasravi:
Won't google.DefaultClient automatically use GOOGLE_APPLICATION_CRENDETIALS if its defined?

Will users need to download a service account and set GOOGLE_APPLICATION_CREDENTIALS?

The desired behavior is that it automatically uses the credentials cached in the home directory created by gcloud. If a user sets GOOGLE_APPLICATION_CREDENTIALS then those should be used.

My expectation is that the go client libraries will automatically try a variety of methods (including the 2 I listed above) to get credentials. So our code

jlewi:
We create a DefaultClient here
https://github.com/kubeflow/kubeflow/blob/6fa88b7cf38430d17e35f968abbe0ea9d2272475/bootstrap/pkg/client/gcp/gcp.go#L275

Which should be reading the credentials from the gcloud .config file if GOOGLE_APPLICATION_CREDENTIALS isn't set.

That didn't seem to be the case when I tried basic auth and the release demo:
http://bit.ly/05_mid_release_demo

I'll debug further.

jlewi:
I suspect this might be a question of needing to run

```
 gcloud auth application-default login
```

Looking at the GCP Go client code it looks like it checks the file

```
.config/gcloud/application_default_credentials.json
```

gabrielwen:
I think the root cause is not just `DefaultClient` but [DefaultTokenSource](https://github.com/golang/oauth2/blob/HEAD/google/default.go#L43), which we also use in other places.  Let me do some testing and see if we could remove `DefaultTokenSource`.

gabrielwen:
So it seems we have several options:
### Without `gcloud` as dependency
- We download service account credential file as set with `GOOGLE_APPLICATION_CREDENTIALS`.
- We download user oauth secrets and supply as input to application: [doc](https://cloud.google.com/docs/authentication/end-user)
### With `gcloud` as dependency
- User needs to run `gcloud auth login` and `gcloud auth application-default login` to have `application_default_credentials.json`.

In any case it doesn't seem to have a proper way of avoiding manual process.  Discussed with @kunmingg , we think it is ok to make `gcloud` as required when platform is `gcp`.  And just like `kubectl`, `gcloud` will be the central control to auth related operations.

[...]

Issue #2507, opened 2019-02-19
kfctl installs istio


lluunn:
We have done it for click-to-deploy app, should extend to kfctl.

cc @kunmingg

jlewi:
@lluunn Is this work currently blocked?

lluunn:
yeah, waiting for Kam's fixes.

kunmingg:
@lluunn 
Any update?

jlewi:
@lluunn After #2755 what is the remaining work?

lluunn:
need testing to make sure it works

lluunn:
tested locally. With #2839, the endpoint is up and I can see the dashboard.

jlewi:
@lluunn Can we close this issue? Do we have an issue for adding an E2E test for Kubeflow using ISTIO?

lluunn:
filed #2891 for e2e test

PR #2587, opened 2019-02-28
Semi-working kfctl go binary

### LINKED ISSUE #2472: kfctl: support deploy with basic auth
User should be able to deploy cluster with go binary while provide username/password to later access their kubeflow service

### LINKED ISSUE #2491: kfctl go binary should have an option to skip initi project
The kfctl go binary should have an option to skip initializing the GCP project.

### LINKED ISSUE #2514: add additional config files to kfctl for iap, basic_auth, default components
in order to accommodate the different configuration possibilities we should allow kfctl to load config from 
- https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_iap.yaml
- https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_basic_auth.yaml
- https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_defaults.yaml

this will allow a different set of packages, components and parameters than standard as well as iap,  basic_auth

gabrielwen:
Works related to this PR:
- Fixes #2491
- Fixes #2472
- Fixes #2514
- Making #2370 ~80% finished.

Known WIPs:
- #2586 
- #2585 
- #2507 
- #2535
- #2391
- #2604
- When creating resources, some of resources has error message `roles.rbac.authorization.k8s.io [resource] is forbidden: attempt to grant extra privileges`
- Error message `Resource creation for metricscollector-template is failed at 1 attempt: ConfigMap "" is invalid: metadata.name: Required value: name or generateName is required`
- #2605 pod `iap` doesn't seem to be working - Having error message like #953

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2587)


lluunn:
could you add in PR description which part of gcp is done and what's remaining?

gabrielwen:
> could you add in PR description which part of gcp is done and what's remaining?

Added some, those are the WIPs that I know of.  I believe there are more as we dig into it, tho.

bootstrap/config/types.go
```diff
@@ -0,0 +1,19 @@
+package config
+
+import ()
```
lluunn: remove?

bootstrap/config/types.go
```diff
@@ -0,0 +1,19 @@
...
+
+type NameValue struct {
```
lluunn: add comment on the new types

bootstrap/pkg/apis/apps/group.go
```diff
@@ -37,9 +45,12 @@ const (
 	DefaultPlatform  = "ksonnet"
 	// TODO: find the latest tag dynamically
 	DefaultVersion  = "master"
-	DefaultGitRepo  = "https://github.com/kubeflow/kubeflow/tarball"
+	DefaultGitRepo  = "https://github.com/gabrielwen/kubeflow/tarball"
```
lluunn: revert?

bootstrap/cmd/kfctl/cmd/apply.go
```diff
@@ -69,4 +69,20 @@ func init() {
 		log.Errorf("couldn't set flag --%v: %v", string(kftypes.VERBOSE), bindErr)
 		return
 	}
+	applyCmd.Flags().String(string(kftypes.OAUTH_ID), "",
+		"OAuth Client ID, GCP only. Required if not ENV CLIENT_ID is not set. "+
```
lluunn: if not --> if. same below.

bootstrap/pkg/utils/k8utils.go
```diff
@@ -38,10 +44,152 @@ import (
...
+}
+
+func getResource(mapping *meta.RESTMapping, config *rest.Config, group string,
+	version string, namespace string, name string) error {
+	restClient, err := getRESTClient(config, group, version)
+	if err != nil {
+		return fmt.Errorf("getResource error: %v", err)
+	}
+
+	_, err = restClient.
```
lluunn: get should return the response?

bootstrap/config/types.go
gabrielwen: removed.

bootstrap/pkg/apis/apps/group.go
gabrielwen: reverted.

bootstrap/cmd/kfctl/cmd/apply.go
gabrielwen: updated.

bootstrap/config/types.go
gabrielwen: added.

lluunn:
if we plan to shell out the kubectl, do we still need the k8sutil thing?
One reason I added that was because we don't want to shell out in bootstrapper until we isolate different runs in containers.
@kunmingg

bootstrap/pkg/utils/k8utils.go
gabrielwen: currently we don't use the response.  added a TODO in case we need to do it.

gabrielwen:
> if we plan to shell out the kubectl, do we still need the k8sutil thing?
> One reason I added that was because we don't want to shell out in bootstrapper until we isolate different runs in containers.
> @kunmingg

I think we definitely should try to avoid shell out especially in bootstrapper.  It'll be very hard to poll the status and it's gonna stuck there for minutes - not ideal for server loading.

jlewi:
> When creating resources, some of resources has error message `roles.rbac.authorization.k8s.io [resource] is forbidden: attempt to grant extra privileges`

This is likely because you need to explicitly grant the current user cluster-admin privileges
Like we do here:
https://github.com/kubeflow/kubeflow/blob/544bf699370c59a019c7218fd06d72d42e377221/scripts/gke/util.sh#L179

gabrielwen:
> > When creating resources, some of resources has error message `roles.rbac.authorization.k8s.io [resource] is forbidden: attempt to grant extra privileges`
> 
> This is likely because you need to explicitly grant the current user cluster-admin privileges
> Like we do here:
> [kubeflow/scripts/gke/util.sh](https://github.com/kubeflow/kubeflow/blob/544bf699370c59a019c7218fd06d72d42e377221/scripts/gke/util.sh#L179)
> 
> Line 179 in [544bf69](/kubeflow/kubeflow/commit/544bf699370c59a019c7218fd06d72d42e377221)
> 
>  kubectl create clusterrolebinding default-admin --clusterrole=cluster-admin --user=${EMAIL}

agree - need to use impersonate as admin for these requests.

gabrielwen:
will address to other comments tomorrow :)

gabrielwen:
> Do we need to do impersonate? I don't think we use impersonation with kfctl.sh. What is click to deploy currently doing?
> 
> > _[Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2587#-:-LZu5DvABVe1XhOFI5RL:b-5dfhpd)_ status: 13 of 15 files reviewed, 21 unresolved discussions (waiting on @ellis-bigelow, @gabrielwen, @jlewi, and @lluunn)

We do: https://github.com/kubeflow/kubeflow/blob/master/scripts/gke/util.sh#L197

bootstrap/config/kfctl_iap.yaml
```diff
@@ -3,23 +3,46 @@
...
+  - application
+  - modeldb
+  - gcp
+components:
+  - pytorch-operator
   - ambassador
+  - openvino
+  - jupyter
+  - notebook-controller
+  - jupyter-web-app
```
kunmingg: openvino is not a default component, shouldn't be part of default config.
 We can add it on the fly via go logic.

bootstrap/config/kfctl_iap.yaml
```diff
@@ -3,23 +3,46 @@
...
   - centraldashboard
   - tf-job-operator
+  - tensorboard
+  - metacontroller
+  - profiles
+  - notebooks
```
kunmingg: We are still working on ```notebooks``` and ```notebook-controller``` components.
We can skip them in this PR and add later once work on that side is done.

kunmingg:
@gabrielwen 
Instead of impersonate you can bind ```default-admin``` to user before do apply.
click to deploy app is "Out of Cluster" as well.

bootstrap/config/kfctl_iap.yaml
gabrielwen: removed.

bootstrap/config/kfctl_iap.yaml
gabrielwen: removed.

bootstrap/config/kfctl_iap.yaml
```diff
@@ -30,13 +53,22 @@ parameters:
       value: ipName
     - name: hostname
       # TODO change value on the fly: replace with user-provide parameters. This need to be fully qualified domain name to use with ingress.
-      value: <deployName>.endpoints.<Project>.cloud.goog
+      value: <deployName>.endpoints.<project>.cloud.goog
+compParams:
```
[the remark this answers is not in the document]
gabrielwen: `protoParams` is called with `ks init`, while `compParams` is called with `ks param set`.  Explaned in bootstrap/config/types.go

gabrielwen:
> @gabrielwen
> Instead of impersonate you can bind `default-admin` to user before do apply.
> click to deploy app is "Out of Cluster" as well.

mm, it's weird that:
- I already bind "default-admin" to user.
- If I use `kubectl apply`, this kind of error doesn't show up.

gabrielwen:
> > @gabrielwen
> > Instead of impersonate you can bind `default-admin` to user before do apply.
> > click to deploy app is "Out of Cluster" as well.
> 
> mm, it's weird that:
> 
> * I already bind "default-admin" to user.
> * If I use `kubectl apply`, this kind of error doesn't show up.

gabrielwen:
> @gabrielwen
> Instead of impersonate you can bind `default-admin` to user before do apply.
> click to deploy app is "Out of Cluster" as well.

Also, from [here](https://github.com/kubeflow/kubeflow/blob/76230712d80805251f2ac54d66b0f02838de1c79/bootstrap/cmd/bootstrap/app/ksServer.go#L431) I think it's using In Cluster config.

PR #2609, opened 2019-03-04
Adding additional printer columns to display status

johnugeorge:
Display status field for kubectl get tfjobs/pytorchjobs. This outputs the type from the last condition in the status field.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2609)

### DIFF
--- a/kubeflow/pytorch-job/pytorch-operator.libsonnet
+++ b/kubeflow/pytorch-job/pytorch-operator.libsonnet
@@ -38,6 +38,18 @@
         subresources: {
           status: {},
         },
+        additionalPrinterColumns: [
+          {
+            JSONPath: ".status.conditions[-1:].type",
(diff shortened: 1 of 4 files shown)


kubeflow/pytorch-job/pytorch-operator.libsonnet
```diff
@@ -38,6 +38,18 @@
         subresources: {
           status: {},
         },
+        additionalPrinterColumns: [
+          {
+            JSONPath: ".status.conditions[-1:].type",
```
richardsliu: Can conditions be empty?

johnugeorge: @richardsliu I verified that field will be empty in that case.

Issue #2615, opened 2019-03-04
[Dashboard v1] Add endpoint to central dashboard server to surface events


prodonjs:
Add API endpoint to surface K8s events to the Central Dashboard.

Parent: #2512

/assign prodonjs
/area front-end
/priority p1

avdaredevil:
Close?

prodonjs:
Let's close when #2667 is merged.

Issue #2614, opened 2019-03-04
[Dashboard v1] Support deep-links


prodonjs:
Ensure that iframed pages and native component pages have unique URLs. 

This makes it possible to use browser buttons to land on specific pages of the dashboard.

Parent: #2359

prodonjs:
Closed by #2626 and #2651.

PR #2626, opened 2019-03-05
Separate front-end components and add mock activities service.

prodonjs:
- Adds deep-link routing behavior for all front-end pages.
  Iframed pages reflect a URL to the browser that makes them
  linkable but hides the internal page.
- Seperates the front-end into subcomponents, including an
  activities view.
- Fix iframe view behavior to have it cover the space voided
  by the layout header.
- Adds /api/activities endpoint to the Express server for
  activities to eventually be replaced with real data.
- Sets the _devMode flag to true based on the webpack build env.

Addresses #2614 and #2615

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2626)


components/centraldashboard/public/.eslintrc.json
```diff
@@ -6,6 +6,7 @@
     "extends": ["eslint:recommended", "google"],
     "globals": {
         "VERSION": true,
```
avdaredevil: Shouldn't version be a `string` perhaps `"dev-build"`?

components/centraldashboard/public/components/activity-view.js
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * Handles the Activities response to set date format and icon.
+     * @param {Event} responseEvent
+     */
+    _onResponse(responseEvent) {
+        const {status, response} = responseEvent.detail;
+        this.activities = [];
+        if (status === 200) {
```
avdaredevil: Can we invert this condition and decrease code depth:
```js
if (status != 200) return;
```

components/centraldashboard/public/components/main-page.js
```diff
@@ -27,88 +28,130 @@ import '@polymer/neon-animation/animations/fade-out-animation.js';
...
+            routeData: Object,
+            subRouteData: Object,
+            iframeRoute: Object,
+            menuLinks: {
                 type: Array,
                 value: [
-                    {text: 'Home', defaultPage: true, hasDivider: true},
                     {
-                        link: 'https://www.kubeflow.org/docs/about/kubeflow/',
+                        iframeUrl: 'https://www.kubeflow.org/docs/about/kubeflow/',
```
avdaredevil: Could we go with `href`? Not all are guaranteed to open in the iframe (ie. this one!)

components/centraldashboard/public/components/main-page.js
```diff
@@ -27,88 +28,130 @@ import '@polymer/neon-animation/animations/fade-out-animation.js';
...
                         text: 'Kubeflow docs',
+                        href: '/docs',
```
avdaredevil: Let's call this `route` so it's clear this is a route for the current page rather than a _Hypertext **REFerence**_.

components/centraldashboard/public/components/main-page.js
```diff
@@ -27,88 +28,130 @@ import '@polymer/neon-animation/animations/fade-out-animation.js';
...
+    /**
+     * Intercepts any external links and ensures that they are captured in
+     * the route and sent to the iframe source.
+     * @param {MouseEvent} e
+     */
+    openInIframe(e) {
+        const url = new URL(e.currentTarget.href);
+        window.history.pushState({}, null, `_${url.pathname}`);
+        window.dispatchEvent(new CustomEvent('location-changed'));
+        e.preventDefault();
```
avdaredevil: Does this actually work in the cluster? If it's a matter of unique routes, we can prefix all our routes with `d/` for `dashboard`.

components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,44 +1,40 @@
 app-drawer-layout.flex(narrow='{{narrowMode}}')
+    app-location(route='{{route}}')
+    app-route(route='{{route}}', pattern='/:page', data='{{routeData}}', tail='{{subRouteData}}')
+
+
```
avdaredevil: One too many extra whitespaces

components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,44 +1,40 @@
...
+            template(is='dom-repeat', items='[[menuLinks]]')
+                a(href$='[[item.href]]', on-click='openInIframe', tabindex='-1')
+                    paper-item.menu-item [[item.text]]
         aside.flex
         footer.footer
             section.privacy Privacy
-            section.build build version 
-                span(title="Build: v[[buildVersion]] | Dashboard: v[[dashVersion]]") v[[buildVersion]]
+            section.build build version
+                span(title="Build: v[[buildVersion]] | Dashboard: v[[dashVersion]]")  v[[buildVersion]]
```
avdaredevil: Revert change on this line

components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,44 +1,40 @@
...
-                            paper-item-body More coming soon
-                            paper-icon-button.button(icon='arrow-forward', diabled)
-                article#Activity
-            iframe#PageFrame.flex(src='[[url]]')
+                header(hides, hidden$='[[hideToolbar]]')
+                    paper-tabs.bottom(selected='[[page]]', attr-for-selected='page')
+                        paper-tab(page='dashboard', link)
+                            a.link(tabindex='-1', href='/') DASHBOARD
+                        paper-tab(page='activity', link,hidden$='[[!_devMode]]')
+                            a.link(tabindex='-1', href='/activity') ACTIVITY
```
avdaredevil: The case doesn't match the mock:
image

Also great job with simplifying the `neon-animatable-pages` with the `app-router`! 🎉

components/centraldashboard/app/server.ts
```diff
@@ -7,12 +7,47 @@ const {PORT, PORT_1} = process.env;
...
+    }
+    return activities.sort((a, b) => {
+        return a.time > b.time ? -1 : 1;
+    });
+}
+
+app.use(express.json());
 app.use(express.static(frontEnd));
 
 app.get("/api", (req: express.Request, res: express.Response) => {
```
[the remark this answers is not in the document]
prodonjs: Understood. I intend to break this apart with the PR that addresses #2615 as it will require some additional changes in the API layer.

components/centraldashboard/app/server.ts
```diff
@@ -7,12 +7,47 @@ const {PORT, PORT_1} = process.env;
...
+    for (let i = 0; i < 100; i++) {
+        activities.push({
+            time: new Date(now - (Math.random() * 86400000)),
+            event: `Event #${i}`,
+            isError: Math.random() * 10 <= 1, // 1/10 probability
+            source: systems[Math.floor(Math.random() * 3)]
+        });
+    }
+    return activities.sort((a, b) => {
+        return a.time > b.time ? -1 : 1;
```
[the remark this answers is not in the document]
prodonjs: Good suggestion. I like the succinctness.

components/centraldashboard/public/.eslintrc.json
prodonjs: Actually, this is just indicating to ESLint that the value is considered to be in-scope at runtime so the linter doesn't complain.

Based on the [docs](https://eslint.org/docs/user-guide/configuring#specifying-globals), it looks like the best option is to set them to **readonly**.

components/centraldashboard/public/components/activity-view.js
prodonjs: Good call. I think we'll want to add some kind of means for surfacing an request failure as well so that block will be expanded.

components/centraldashboard/public/components/main-page.js
prodonjs: I introduced `href` to be the attribute that is used when setting the anchor tag's href property on our UI to indicate the local "route". I changed link to `iframeUrl` since it was being assigned to the component's `iframeUrl` property. Other than the **Home** link which I moved out of the `menuLinks` list, all of the other ones opened in the iframe view in the existing code.

components/centraldashboard/public/components/main-page.js
prodonjs: See my above comment for my rationale. I'm not opposed to this change but LMK what you think.

components/centraldashboard/public/components/main-page.js
prodonjs: It does. I also considered the anchor links approach which can be configured in [app-location](https://www.webcomponents.org/element/PolymerElements/app-route/elements/app-location#property-useHashAsPath). That's a more futureproof solution if we are concerned about collisions.

components/centraldashboard/public/components/main-page.pug
prodonjs: Got it..

components/centraldashboard/public/components/main-page.pug
prodonjs: My editor is set to trim trailing whitespace which would be a common setting many users would have. I added the `&nbsp` to be explicit.

components/centraldashboard/public/components/main-page.pug
prodonjs: Thanks. I think I got caught up viewing the GCP console Dashboard and Activity tabs which use all caps. 
cczvgbouvu8

With respect to the namespace selector, ours has never included the rounded border shown in the mock, but I didn't make any changes to it in this PR.

prodonjs:
@avdaredevil PTAL when you can. This merges the changes from #2651. If possible, I'd like to get this change in since it separates the front-end components and makes it easier to divide work going forward.

prodonjs:
Also, the responsiveness doesn't work particularly well on mobile if that's a concern at this point. This was true of the code that I merged from as well. I don't think we need to address it here but it shouldn't be too tough to give it a flex layout with wrapping.

p7yjt6ey6zx

components/centraldashboard/public/.eslintrc.json
avdaredevil: I see, thanks!

avdaredevil:
Responsiveness isn't a required feature (in fact it's not even in the pipeline as most of our clients are on Desktop / Laptops). However, I decided to make it slightly responsive for those screens for the sake of 800px to 1600px screens

prodonjs:
@avdaredevil Merge conflict fixed. PTAL.

components/centraldashboard/public/components/main-page.js
avdaredevil: In your new PR please also move out the docs page, thanks!

Issue #2580, opened 2019-02-28
[click-to-deploy app] Load test with multiple projects 


zhenghuiwang:
Current [load test](https://github.com/kubeflow/kubeflow/blob/master/testing/test_deploy_app.py#L378) for click-to-deploy app doesn't generate enough load due to the API quota limit for a single project.

We need to

- Auto generate dozens of projects.
- Send requests to click-to-deploy to deploy KF to all these projects at the same time.

zhenghuiwang:
/cc @jlewi @kunmingg 

I created 30 projects at folder `kf-gcp`: `kf-load-test-project1`, `kf-load-test-project2`...`kf-load-test-project30` using https://github.com/kubeflow/testing/pull/325

I ran 20 deployments (10 project X 2 deployment per project) in for 1-click-deployment staging simultaneously using https://github.com/kubeflow/kubeflow/pull/2648. Here are the logs for the final result
```
......
INFO|2019-03-13T15:54:13|test_deploy_app.py|301| IAP is ready for https://kubeflow2.endpoints.kf-load-test-project10.cloud.goog!
INFO|2019-03-13T15:54:13|test_deploy_app.py|553| 2 out of 2 deployments succeed for project kf-load-test-project10
INFO|2019-03-13T15:54:13|test_deploy_app.py|562| 19 out of 20 deployments succeed in total
INFO|2019-03-13T15:54:13|test_deploy_app.py|410| Clean up project resource (source repo, backend service and deployment)
...
```
The only failed deployment is due to `envoy` keeps failing.

kunmingg:
Thanks!

PR #2648, opened 2019-03-07
Use multiple projects in the load test for 1-click-deployment

zhenghuiwang:
#2580 

This PR extends current load test to use multiple projects:
It now sends requests to `X` projects and  `Y` deployments per project, where X = `number_project` and Y=`number_deployments_per_project` from command line flags. In total that is X*Y concurrent requests for deploying Kubeflow.

In addition to that, I removed the `projectNumber` flag and get the value from the `projectID` flag. So there won't be mismatch by any chance.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2648)


testing/run_deploy_app.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#
+# Scripts to start a load test for 1-click-deployment service.
+#
+# GOOGLE_APPLICATION_CREDNTIALS and SERVICE_CLIENT_ID both point to
```
jlewi: What is this script for?
e.g. is this for manual testing or is it intended to be run in cluster as a cron job to run the load test regularly?

zhenghuiwang: This is for manually running the load test.

jlewi:
Mostly looks good. one question about the script.

testing/run_deploy_app.sh
jlewi: Could you update the comment please?

How come you are using a service account if you are running manually?

Why wouldn't you just use the credentials for @zhenghuiwang  configured via gcloud?

zhenghuiwang: I updated the comment.

I think the load test/ probe test are written with a cron job in mind. This script is just for me to test it actually works manually.

There are many env vars and cmd flags. So I think it is useful to document the entry point via a script.

testing/run_deploy_app.sh
```diff
@@ -0,0 +1,25 @@
...
+# service account "load-test-owner@kf-gcp-deploy0.iam.gserviceaccount.com"
+# https://pantheon.corp.google.com/iam-admin/serviceaccounts/details/112401461927766705527?organizationId=714441643818&project=kf-gcp-deploy0
+#
+# CLIENT_ID and CLIENT_SECRET both point to
+# https://pantheon.corp.google.com/apis/credentials/oauthclient/459682233032-a76ps35eh3j8odvudf4292bgq0jam74i.apps.googleusercontent.com?project=kf-gcp-deploy0&organizationId=714441643818
+
+GOOGLE_APPLICATION_CREDENTIALS=/usr/local/google/home/zhenghui/env/kf-gcp-deploy0-load-test.json \
+SERVICE_CLIENT_ID=112401461927766705527 \
+CLIENT_ID=459682233032-a76ps35eh3j8odvudf4292bgq0jam74i.apps.googleusercontent.com \
+CLIENT_SECRET=eEWmsMuOo7s1o04bRctfehuD \
```
kunmingg: CLIENT_SECRET should be passed as Env Var and not publicly available.

zhenghuiwang: Thanks for catching this security issue. I reset the client secret.
This is the env var for running a python cmd. I just change it to avoid putting the actual value here.

jlewi:
/lgtm
/approve
/hold 
@kunmingg Any additional comments?

jlewi:
Test failure is spark operator; looks like #2693

kunmingg:
Only 1 comment:

For ```insert_ssl_cert```, current GCS path is build from test_mode + deploy_name:
https://github.com/kubeflow/kubeflow/blob/master/testing/test_deploy_app.py#L239
Since in our load test we reuse deploy_name across projects, they will write to same GCS path and overwrite each other.

Can we add project as part of the GCS path as well? Or just use hostname as GCS folder name?

zhenghuiwang:
> Only 1 comment:
> 
> For `insert_ssl_cert`, current GCS path is build from test_mode + deploy_name:
> https://github.com/kubeflow/kubeflow/blob/master/testing/test_deploy_app.py#L239
> Since in our load test we reuse deploy_name across projects, they will write to same GCS path and overwrite each other.
> 
> Can we add project as part of the GCS path as well? Or just use hostname as GCS folder name?

good suggestion! added the project name into the GCS path.

kunmingg:
/lgtm
/approve
Thanks!

Issue #2546, opened 2019-02-24
Don't iframe the docs into the new Kubeflow UI


sarahmaddox:
Referring to PR https://github.com/kubeflow/kubeflow/pull/2361

Hallo @avdaredevil and @abhi-g I'd like to discuss the question of iframing the docs into the Kubeflow UI. I'd prefer not to iframe the doc site into the UI, as I think the experience is on the whole confusing and aesthetically unappealing. The doc website is designed as a standalone site. When iframed into the Kubeflow UI, we have double (and slightly different) branding and doubled/trebled navigation elements.

It may make sense for the Kubeflow webapps all to be iframed into the UI (I don't want to discuss that issue) but I think it makes a better experience for the docs to be left as a separate website, with links going from the various UI elements to the relevant doc page for contextual help.

People are used to having a separate doc site. If the docs are integrated, we need a fully contextual doc config which pops up exactly the right content at each phase of the UI.

abhi-g:
I agree that making docs site open in a separate tab is more desirable
experience for the users. AP, please let us know if we need any further
input from UX on this.

Abhishek

avdaredevil:
The sidebar in the dashboard will hide eventually when navigating to the external links, however, we don't want the user experience to change drastically if people are already used to working in a different way. I have no strong opinion either way. I'll make the change on Monday unless @ajayalfred has something to add.

Thanks @sarahmaddox

abhi-g:
Thanks AP.

Abhishek

ajayalfred:
I agree with Sarah's suggestion. Traditionally (looking at other product examples) help docs is an external site in a new tab. I think it's safe to follow the same convention and not iframe docs.

avdaredevil:
/assign @prodonjs 
Jason's change will include this update.

prodonjs:
@avdaredevil Is this issue resolved? We still load the **Kubeflow Docs** link from the sidenav inside the iframe, but if a user clicks the **Open docs** link from the main landing page it goes to a new tab. I don't think we've addressed that behavior in any previous change so let me know if it needs to be addressed.

avdaredevil:
Since we have completed the sidebar change, this issue is not up in the air, as there are no confusing UI experiences or duplicated sidebars, though I'll defer to @ajayalfred to make that decision.

ajayalfred:
We agreed (on a different git issue) to not iframe KF docs, but to link out to the page.

prodonjs:
[...]

PR #2651, opened 2019-03-07
Getting started card completed

### LINKED ISSUE #2562: [Dashboard v1] Getting started card on has no content
## About
Getting Started Card | Problem
--- | ---
image | There is no content available for these, and currently this card is hidden. Any relevant content would help, since this card will otherwise never show and be tested.

_**Note:** I see this card being used for navigation and actions done within the dashboard or view. Is this a correct assumption_

/assign @sarahmaddox 
/assign @abhi-g 
/assign @ajayalfred 
/assign @avdaredevil 

/cc @jlewi

avdaredevil:
## About
Getting started card completed

Card in normal View | Responsive View
--- | ---
image | image

#### Fixes #2562

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2651)


components/centraldashboard/webpack.config.js
```diff
@@ -81,7 +81,7 @@ module.exports = {
                     }
                 }
             },
-            {
+            (process.platform == 'win32'?{}:{
```
prodonjs: The Windows issues seem like they may be localized to your environment. I set up a Windows 2016 server on GCE, checked from master, and we able to get both the `dev` and `build` npm commands to work as expected without any changes to the eslint settings.

The site doesn't work in IE11 (per our Babel settings), but it worked perfectly fine in Chrome on Windows.

components/centraldashboard/public/components/main-page.js
```diff
@@ -54,11 +54,32 @@ export class MainPage extends PolymerElement {
...
                 value: [
-                    {text: 'Build a model in a notebook', icon: 'donut-large'},
-                    {text: 'Launch a pipeline', icon: 'donut-large'},
-                    {text: 'Deploy a solution', icon: 'donut-large'},
-                    {text: 'Compare run output', icon: 'donut-large'},
-                    {text: 'Share solution on AI Hub', icon: 'donut-large'},
+                    {
+                        text: 'Getting started with Kubeflow',
+                        desc: 'Quickly get running with your ML workflow on an existing Kubernetes installation',
+                        link: 'https://www.kubeflow.org/docs/started/getting-started/',
```
prodonjs: It seems a bit odd that these links open externally versus using the iframe view which would be activated if the user clicked Kubeflow Docs from the sidenav. My open PR #2614 will refactor this section and add the router, but I'm curious if you think we should make these links appear in the iframe as well.

avdaredevil: Your logic makes perfect sense, and that was my initial goal. However, that is no longer the desired behavior as mentioned in #2546. 

I am waiting for your PR to then refactor all `link`s to `href`s

components/centraldashboard/webpack.config.js
avdaredevil: Ok I'll work on this a little more to get it working. Seems really wierd

avdaredevil:
Correct me if I am wrong, but I18n isn't on the Agenda for landing page right? @ajayalfred

prodonjs:
I still don't understand why your specific environment is not letting us use the [standard method](https://eslint.org/docs/user-guide/configuring#using-configuration-files-1) of putting the configuration file in the directory where it will be applied. I don't want to hold things up though so we'll have to investigate that later.

/lgtm

PR #2647, opened 2019-03-07
 kfctl install Istio

lluunn:
#2507 

- breakdown gcp.go to multiple files
- add flag, `use_istio`, to init
- when `use_istio==true`
    - generate: copy the dependencies/istio manifests to the app
    - apply: apply istio manifest

Tested locally, istio can be installed.

Next step: change kubeflow components to work with Istio

/cc @kunmingg @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2647)

PR #2660, opened 2019-03-08
kfctl install istio -- components that work with istio

lluunn:
/hold
based on #2647 , will wait for that one to get merged first

- set param `useIstio=true` for iap-ingress component 
- create secret (admin-gcp-sa, kubeflow-oauth) in istio-system namespace

Also fix a bug in setNameVal (see code comment)

/cc @kunmingg @jlewi @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2660)


lluunn:
I will redo it since #2547 is merged

PR #2667, opened 2019-03-08
Add Namespaces and Activities from API server.

prodonjs:
- Work in progress changes for adding events surfacing to central dashboard.
- Adds Google typescript formatting guidelines.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2667)

### DIFF
--- a/components/centraldashboard/app/k8s_service.ts
+++ b/components/centraldashboard/app/k8s_service.ts
@@ -1,0 +1,34 @@
+import * as k8s from '@kubernetes/client-node';
+
+/** Wrap Kubernetes API calls in a simpler interface for use in routes. */
+export class KubernetesService {
+  private k8sApi: k8s.Core_v1Api;
+
+  constructor(private kubeConfig: k8s.KubeConfig) {
+    console.info('Initializing Kubernetes configuration');
+    this.kubeConfig.loadFromDefault();
+    this.k8sApi = this.kubeConfig.makeApiClient(k8s.Core_v1Api);
+  }
+
+  /** Retrieves the list of namespaces from the Cluster. */
+  async getNamespaces(): Promise<k8s.V1Namespace[]> {
+    try {
+      const {body} = await this.k8sApi.listNamespace();
+      return body.items;
+    } catch (err) {
+      console.error('Unable to fetch Namespaces', err.body ? err.body : err);
+      return [];
+    }
+  }
+
+  /** Retrieves the list of events for the given Namespace from the Cluster. */
+  async getEventsForNamespace(namespace: string): Promise<k8s.V1Event[]> {
+    try {
+      const {body} = await this.k8sApi.listNamespacedEvent(namespace);
+      return body.items;
+    } catch (err) {
+      console.error('Unable to fetch Namespaces', err.body ? err.body : err);
+      return [];
+    }
+  }
+}
--- a/components/centraldashboard/app/server.ts
+++ b/components/centraldashboard/app/server.ts
@@ -1,51 +1,22 @@
-import express from "express";
-import { resolve } from "path";
-import process from "process";
+import {KubeConfig} from '@kubernetes/client-node';
+import express from 'express';
+import {resolve} from 'path';
 
+import {api} from './api';
+
+const {PORT_1} = process.env;  // PORT_1 was defined in an earlier version
+const port: number = Number(PORT_1) || 8082;
+
+const frontEnd: string = resolve(__dirname, 'public');
 const app: express.Application = express();
-const {PORT, PORT_1} = process.env;
-const port: number = Number(PORT) || Number(PORT_1) || 8082;
-const frontEnd: string = resolve(__dirname, "public");
-
-interface Activity {
-    time: Date;
-    event: string;
-    isError: boolean;
-    source: string;
-}
-const actvities = _generateActivities();
-function _generateActivities(): Activity[] {
-    const activities: Activity[] = [];
-    const systems: string[] = [
-        "serving-system",
-        "gpu-system",
-        "training-system"
-    ];
-    const now = Date.now();
-    for (let i = 0; i < 100; i++) {
-        activities.push({
-            time: new Date(now - (Math.random() * 86400000)),
-            event: `Event #${i}`,
-            isError: Math.random() * 10 <= 1, // 1/10 probability
-            source: systems[Math.floor(Math.random() * 3)]
-        });
-    }
-    return activities.sort((a, b) => b.time.getTime() - a.time.getTime());
-}
+const kubeConfig = new KubeConfig();
 
 app.use(express.json());
 app.use(express.static(frontEnd));
-
-app.get("/api", (req: express.Request, res: express.Response) => {
-    console.info(`Request ${req.url} received`);
-    res.send("Hello World");
+app.use('/api', api(kubeConfig));
+app.get('/*', (_: express.Request, res: express.Response) => {
+  res.sendFile(resolve(frontEnd, 'index.html'));
 });
-app.get("/api/activities", (req: express.Request, res: express.Response) => {
-    res.send(actvities);
-});
-app.get("/*", (req: express.Request, res: express.Response) => {
-    res.sendFile(resolve(frontEnd, "index.html"));
-});
-
-app.listen(port,
+app.listen(
+    port,
     () => console.info(`Server listening on port http://localhost:${port}`));
--- a/components/centraldashboard/public/components/activity-view.js
+++ b/components/centraldashboard/public/components/activity-view.js
@@ -1,10 +1,11 @@
-import {PolymerElement, html} from '@polymer/polymer';
-
 import '@polymer/iron-flex-layout/iron-flex-layout-classes.js';
 import '@polymer/iron-ajax/iron-ajax.js';
 import '@polymer/iron-icon/iron-icon.js';
 import '@polymer/iron-icons/iron-icons.js';
 import '@polymer/paper-progress/paper-progress.js';
+
+import {html, PolymerElement} from '@polymer/polymer';
+import './activities-list.js';
 
 export class ActivityView extends PolymerElement {
     static get template() {
@@ -14,47 +15,26 @@ export class ActivityView extends PolymerElement {
             <style>
                 :host {
                     @apply --layout-vertical;
-                    background: #f1f3f4;
                     --accent-color: #007dfc;
                     --primary-background-color: #003c75;
                     --sidebar-default-color: #ffffff4f;
                     --border-color: #f4f4f6;
+                    background: #f1f3f4;
                     padding: 1em;
+                    overflow: auto;
                 }
                 paper-progress {
                     width: 100%;
                     --paper-progress-active-color: var(--accent-color)
                 }
-                #list {
-                    overflow-y: auto;
-                }
-                .activity-row {
+                p.message {
                     background: #fff;
                     box-shadow: 0 3px 3px rgba(0,0,0,.12);
                     transition: margin .2s cubic-bezier(0.4, 0, 0.2, 1);
                     font-size: 13px;
+                    margin: 0px;
                     padding: 5px;
                     align-items: center;
-                }
-                .activity-row:not(:first-of-type) {
-                    border-top: 1px solid rgba(0,0,0,.12);
-                }
-                .activity-row > .flex {padding: 0 5px}
-                .activity-row > .src {color: #616161}
-                .event {
-                    flex-basis: 50%;
-                }
-                .event-icon {
-                    height: 18px;
-                    width: 18px;
-                    padding: 0 10px;
-                }
-                .event-icon.build {
-                    transform: scale(-1, 1);
-                    color: rgb(28, 175, 199);
-                }
-                .event-icon.error {
-                    color: rgb(218, 66, 54);
                 }
                 [hidden] {
                     display: none;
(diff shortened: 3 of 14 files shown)


prodonjs:
In doing some in-cluster testing, it appears that the Service Account granted to the Central Dashboard component does not have the necessary access to Kubernetes APIs required to retrieve namespaces or events for a namespace.

```
body:
   { kind: 'Status',
     apiVersion: 'v1',
     metadata: {},
     status: 'Failure',
     message:
      'namespaces is forbidden: User "system:serviceaccount:kubeflow:centraldashboard" cannot list namespaces at the cluster scope',
     reason: 'Forbidden',
     details: { kind: 'namespaces' },
     code: 403 } }
```

```
body:
   { kind: 'Status',
     apiVersion: 'v1',
     metadata: {},
     status: 'Failure',
     message:
      'events is forbidden: User "system:serviceaccount:kubeflow:centraldashboard" cannot list events in the namespace "kubeflow"',
     reason: 'Forbidden',
     details: { kind: 'events' },
     code: 403 } }
```

prodonjs:
It looks like we will need to update https://github.com/kubeflow/kubeflow/blob/master/kubeflow/common/centraldashboard.libsonnet#L109 to include the additional roles. I'm trying to experiment on my own cluster to determine the change needed.

@avdaredevil If you know of anyone else in the project who is more familiar in managing these Ksonnet templates and would be able to quickly lend a hand, that would be most appreciated.

prodonjs:
The Jupyter FE also retrieves namespaces. I'll create a separate PR that adds the appropriate ClusterRole that allows accessing Namespaces and Events from those namespaces akin to what can be found in:
 https://github.com/kubeflow/kubeflow/blob/master/kubeflow/jupyter/jupyter-web-app.libsonnet#L36

prodonjs:
Demo video is available @ https://screencast.googleplex.com/cast/NjMxODkwOTAyNDEwODU0NHw4NTY3ZTg3Ny0xZA.

PR to address the permissions required to have things work properly in-cluster is #2684.

If possible, I'd like to get this in place to then be able to work on adding some unit tests and better failure behavior.

prodonjs:
A couple of further comments after some research.

1. The kind of information that can be found in the Activities tab in GCP 
Screenshot from 2019-03-12 14-50-56
comes from [auditing](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/) configured in the GKE runtime environment which uses a webhook to record the audit events as they occur. This data is not queryable via an API on the Kubeflow cluster.
2. Events in Kubernetes are only persisted for a brief time. The default is 1 hour per https://github.com/kubernetes/kubernetes/blob/master/cmd/kube-apiserver/app/options/options.go#L142

As such, it does not appear that there will ever be a substantial amount of events that span past a single day. Auto-refreshing on-scroll is likely not going to be necessary either. It seems it may be best to just set an interval to refresh when the view is active.

avdaredevil:
Thanks for figuring this out! For MVP we should just do what you mentioned and list the events available and poll every 5 seconds when the tab is active.

components/centraldashboard/app/k8s_service.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  /** Retrieves the list of namespaces from the Cluster. */
+  async getNamespaces(): Promise<k8s.V1Namespace[]> {
+    try {
+      const {body} = await this.k8sApi.listNamespace();
+      return body.items;
+    } catch (err) {
+      console.error('Unable to fetch Namespaces', err.body);
+      return [];
```
avdaredevil: _**[idiomatic js]**_
Shouldn't we reject the promise here, instead of returning an empty array?

components/centraldashboard/app/k8s_service.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  /** Retrieves the list of events for the given Namespace from the Cluster. */
+  async getEventsForNamespace(namespace: string): Promise<k8s.V1Event[]> {
+    try {
+      const {body} = await this.k8sApi.listNamespacedEvent(namespace);
+      return body.items;
+    } catch (err) {
+      console.error('Unable to fetch Namespaces', err.body);
+      return [];
```
avdaredevil: Same

components/centraldashboard/app/routes.ts
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Routes defined for the Express server.
+ */
+export function routes(k8sService: KubernetesService): express.Router {
+  const routes: express.Router = express.Router();
+
+  routes.use(express.json());
+  routes.use(express.static(frontEnd));
+  routes.get(
```
avdaredevil: Can we do
```js
routes.use('/api', apiRoutes)
```
And move all `/api` routes to a `apis.ts` file

components/centraldashboard/app/server.ts
```diff
@@ -1,51 +1,15 @@
-import express from "express";
-import { resolve } from "path";
-import process from "process";
+import express from 'express';
+import process from 'process';
```
avdaredevil: Wait why?

components/centraldashboard/public/components/activity-view.js
```diff
@@ -1,112 +1,109 @@
...
                 :host {
-                    @apply --layout-vertical;
-                    background: #f1f3f4;
+                    @apply --layout-fit;
                     --accent-color: #007dfc;
                     --primary-background-color: #003c75;
                     --sidebar-default-color: #ffffff4f;
                     --border-color: #f4f4f6;
+                    background: #f1f3f4;
+                    min-height: 50%;
```
avdaredevil: `Min-Height` not needed, and `overflow: auto` is missing

components/centraldashboard/public/components/activity-view.js
```diff
@@ -1,112 +1,109 @@
...
-                                icon="[[item.icon]]">
-                            </iron-icon>[[item.event]]
-                        </div>
-                        <div class="flex">[[item.source]]</div>
-                    </div>
-                </template>
-            </div>
+            <p hidden$="[[!message]]">[[message]]</p>
+            <activities-list activities="[[activities]]"></activities-list>
             `;
```
avdaredevil: Put backtick on the previous line?

components/centraldashboard/public/components/activity-view.js
```diff
@@ -1,112 +1,109 @@
...
-                        <div class="flex flex-auto event">
-                            <iron-icon class$="event-icon [[item.icon]]"
-                                icon="[[item.icon]]">
-                            </iron-icon>[[item.event]]
-                        </div>
-                        <div class="flex">[[item.source]]</div>
-                    </div>
-                </template>
-            </div>
+            <p hidden$="[[!message]]">[[message]]</p>
```
avdaredevil: This should have a class for semantic linkability with the css. IE. Make sense when CSS is being read, to know what this element does.

Good example:
`p.Message-Box` or `p.Error-Message`

components/centraldashboard/public/components/activity-view.js
```diff
@@ -1,112 +1,109 @@
...
+            },
         };
     }
 
+    /**
+     * Retrieves Events when namespace is selected.
+     * @param {string} newNamespace
+     */
+    _namespaceChanged(newNamespace) {
+        if (newNamespace) {
```
avdaredevil: Invert this condition to decrease code depth:
```suggestion
        if (!newNamespace) return;
        this.message = '';
        this.$['ajax'].generateRequest();
```

components/centraldashboard/public/components/activity-view.js
```diff
@@ -1,112 +1,109 @@
...
-        // TODO: Surface the error in some manner
-        if (status !== 200) return;
-        this.activities = response.map((a) => {
-            const activity = {
-                formattedTime: new Date(a.time).toLocaleString(),
-                icon: a.isError ? 'error' : 'build',
-            };
-            return Object.assign(activity, a);
-        });
+        this.splice('activities', 0);
```
avdaredevil: I'm sure this has a valid reason, but why do we do this?

avdaredevil:
If you can, could you also provide @ajayalfred with a link to your Cluster's dashboard, so he could take a look?

components/centraldashboard/public/components/activity-view.js
```diff
@@ -1,112 +1,109 @@
...
                 }
                 paper-progress {
                     width: 100%;
                     --paper-progress-active-color: var(--accent-color)
                 }
-                #list {
-                    overflow-y: scroll;
-                }
-                .activity-row {
+                p {
```
avdaredevil: Refer to my suggestion below in the HTML to make the css here more Semantically Understandable.

components/centraldashboard/app/k8s_service.ts
prodonjs: Normally I think that would be the best approach, but if we were to do so here we'd need to add some additional handling to the client code. With this approach, we can log the errors for investigation and not affect the behavior of the front-end for the read-only data being shown.

components/centraldashboard/app/routes.ts
prodonjs: Given the simplicity of the routes at this point and the fact that nothing is under test, I didn't factor it out too broadly. If I were to refactor this I would probably leave the static handling and json middleware in the main server module. What are your thoughts on that?

components/centraldashboard/app/server.ts
prodonjs: We moved the configuration of the static routes to the routes module so resolve is no longer needed here.

components/centraldashboard/public/components/activity-view.js
prodonjs: This seems to be the style followed by most of the canonical [Polymer elements](https://github.com/PolymerElements/paper-input/blob/master/paper-input.js#L187).

components/centraldashboard/public/components/activity-view.js
prodonjs: Understood. This is probably a case of me relying too much on the shadow DOM and private styles not needing as many qualifiers as one would expect to use when using global styles.

components/centraldashboard/public/components/activity-view.js
prodonjs: This makes sure that the the activities list is cleared and then we push the new events onto it later in the method. this.splice and this.push ensure that the changes are notified to other Polymer elements watching the mutable array.

components/centraldashboard/public/components/activity-view.js
prodonjs: Thanks for catching!

components/centraldashboard/public/components/activity-view.js
prodonjs: Done.

components/centraldashboard/public/components/activity-view.js
prodonjs: I've generally followed the "positive condition first" as I find it to be more readable, but I'm fine making this change.

components/centraldashboard/public/components/activity-view.js
prodonjs: Changed and added an error handling condition as well. Iron-ajax only fires on-response if the request is successful so I added on-error.

prodonjs:
@ajayalfred @avdaredevil 

https://kubeflow.endpoints.prodonjs-kubeflow-dev.cloud.goog/ is the Kubeflow cluster I've deployed on GKE with the current changes in this PR. Currently, you'll need to open Dev Tools, select the `<main-page>` tag and type `$0._devMode = true` into the Console to see them.

AP - I think we should consider dropping that flag from hiding these sections after this PR is merged.

components/centraldashboard/app/k8s_service.ts
avdaredevil: Fait enough, lgtm

components/centraldashboard/app/routes.ts
avdaredevil: I agree with that approach wholeheartedly. Sounds good, we can wait for post-NEXT to do those changes.

components/centraldashboard/app/server.ts
avdaredevil: Oh sorry, I meant why are we importing `process`? It's a node.js global

components/centraldashboard/public/components/activity-view.js
avdaredevil: Touché

avdaredevil:
@prodonjs Reviewed, looks good m8. Also I've removed these components from _`_devMode`_  in PR #2695. Once these 2 are in, we should be in a great spot!

@ajayalfred We'd love to get your thoughts on this, the code freeze kicks in this weekend :)

Thanks all!

/lgtm
/approve
/cuj landing-page
/area front-end

components/centraldashboard/app/server.ts
prodonjs: Good point. I think I was just expecting the Typescript compiler to complain if I didn't include it and then I referenced a global library. I removed it.

avdaredevil:
#### fixes #2615
/hold cancel

There's also a few merge conflicts, which I'm guessing your next commit will take care of

prodonjs:
Ok, I think this is ready to merge. That was a bit of an unusual rebase I had some CSS changes had been introduced by some earlier commits on this PR that predated #2695. It would probably would have been better to break this apart and get a few of those in. In any case, I think all is ready to go.

components/centraldashboard/app/k8s_service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  /** Retrieves the list of events for the given Namespace from the Cluster. */
+  async getEventsForNamespace(namespace: string): Promise<k8s.V1Event[]> {
+    try {
+      const {body} = await this.k8sApi.listNamespacedEvent(namespace);
+      return body.items;
+    } catch (err) {
+      console.error('Unable to fetch Namespaces', err.body ? err.body : err);
+      return [];
+    }
```
avdaredevil: Let's discuss this approach more after this PR

avdaredevil:
/hold cancel
/lgtm
/approve

G2G. Let's get this merged asap

components/centraldashboard/app/k8s_service.ts
prodonjs: SGTM, I'm starting on the tests, so it will be easier to start to define behaviors here.

Issue #2610, opened 2019-03-04
[kfctl] Need E2E test for go binary


jlewi:
We need to create a version of our E2E test which uses the new kfctl go binary.

kunmingg:
We need presubmit test covering kfctl for basic auth & IAP path.
@jlewi and @gabrielwen have coming PRs for it.

jlewi:
After #2705 is merged here is some follow on work.

* We need to actually extend the E2E workflow will the validation steps we currently run in kfctl_test
  * e.g. Try submitting a TFJob

* kfctl delete is failing;
  * Delete is failing because `kfctl delete` expects a kubeconfig file and there is none. 

* As a result we will be leaking deployments in the project `kubeflow-ci-deployment`
  * We will need to setup GC'ing of resources in that project.

jlewi:
The test is deployed and running.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kubeflow-periodic-master/1112425585537191937

Issue #2343, opened 2019-01-28
[basic auth] add E2E test


jlewi:
Related to #2094: Basic auth uber issue

We will need an E2E test for basic auth. We want to verify

1. kfctl can deploy using basic auth
1. user can login using basic auth and access central dashboard

jlewi:
Related issue: #2514

jlewi:
#2472 is the issue adding basic auth to kfctl.

jlewi:
We are running an E2E test on postsubmit and periodic to verify that kfctl deploys successfully with basic auth.

PR #2672, opened 2019-03-09
Scaffolding for E2E test for the new kfctl go binary.

jlewi:
* Create a new Argo workflow
* Add a simple python script to build the go binary.

Related to #2610  - E2E test for kfctl
Related to #2343 - Basic auth E2E test

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2672)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -38,7 +38,6 @@ workflows:
       - bootstrap/*
       - deployment/*
       - kubeflow/*
-      - scripts/*
       - testing/*
     params:
       platform: gke
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -26,6 +26,24 @@ workflows:
     params:
       installIstio: true
       workflowName: deployapp-istio
+  # kfctl test runs tests on gke using the new kfctl go binary
+  - app_dir: kubeflow/kubeflow/testing/workflows
+    component: kfctl_go_test
+    name: kfctl-go
+    job_types:
+      - presubmit
+      - postsubmit
+      - periodic
+    include_dirs:
+      - bootstrap/*
+      - deployment/*
+      - kubeflow/*
+      - scripts/*
+      - testing/*
+    params:
+      platform: gke
+      gkeApiVersion: v1
+      workflowName: kfctl-go
   # kfctl test runs tests on gke.
   - app_dir: kubeflow/kubeflow/testing/workflows
     component: kfctl_test
--- a/testing/workflows/components/kfctl_go_test.jsonnet
+++ b/testing/workflows/components/kfctl_go_test.jsonnet
@@ -1,5 +1,5 @@
 // E2E test for the new go based version of kfctl.
-local params = std.extVar("__ksonnet/params").components.kfctl_test;
+local params = std.extVar("__ksonnet/params").components.kfctl_go_test;
 
 local k = import "k.libsonnet";
 local util = import "workflows.libsonnet";
--- a/testing/workflows/components/kfctl_go_test.jsonnet
+++ b/testing/workflows/components/kfctl_go_test.jsonnet
@@ -215,44 +215,6 @@ local dagTemplates = [
 // to execute on exit
 local deleteKubeflow = util.toBool(params.deleteKubeflow);
 
-local deleteStep = if deleteKubeflow then
-  [{
-    template: buildTemplate(
-      "kfctl-delete",
-      [
-        runPath,
-        kfCtlPath,
-        "delete",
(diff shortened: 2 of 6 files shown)


kkasravi:
go.sum conflict is probably due to go1.12 vs go1.11.5

testing/workflows/components/kfctl_go_test.jsonnet
```diff
@@ -0,0 +1,372 @@
...
+      "create_pr_symlink",
+      "--bucket=" + bucket,
+    ]),  // create-pr-symlink
+    dependencies: ["checkout"],
+  },  // create-pr-symlink
+  {
+    template: buildTemplate(
+      "build-kfctl",
+      [        
+        "pytest",
```
lluunn: IIRC invoke it with pytest the  __name__ == "__main__" will not be called.

jlewi: Correct. That's why we use options like --log-cli info. We rely on pytest to set log level appropriately to capture the stout and error.

The main function is just to make it easy to run in a debugger.

lluunn:
> Any chance we can dedup code with the kfctl workflow?
> Or we can refactor later.

Just realized, we will probably just delete kfctl workflow later?

prow_config.yaml
```diff
@@ -26,6 +26,24 @@ workflows:
...
+    name: kfctl-go
+    job_types:
+      - presubmit
+      - postsubmit
+      - periodic
+    include_dirs:
+      - bootstrap/*
+      - deployment/*
+      - kubeflow/*
+      - scripts/*
```
kunmingg: Can we add a TODO for removing scripts/* in future?

testing/workflows/components/kfctl_go_test.jsonnet
```diff
@@ -0,0 +1,372 @@
...
+// outputDir is the directory to sync to GCS to contain the output for this job.
+local outputDir = testDir + "/output";
+local artifactsDir = outputDir + "/artifacts";
+// Source directory where all repos should be checked out
+local srcRootDir = testDir + "/src";
+// The directory containing the kubeflow/kubeflow repo
+local srcDir = srcRootDir + "/kubeflow/kubeflow";
+
+local runPath = srcDir + "/testing/workflows/run.sh";
+local kfCtlPath = srcDir + "/scripts/kfctl.sh";
```
kunmingg: Add TODO to remove script dependency once go CLI support delete?

testing/workflows/components/kfctl_go_test.jsonnet
```diff
@@ -0,0 +1,372 @@
...
+  },
+];
+
+// Each item is a dictionary describing one step in the graph
+// to execute on exit
+local deleteKubeflow = util.toBool(params.deleteKubeflow);
+
+local deleteStep = if deleteKubeflow then
+  [{
+    template: buildTemplate(
```
kunmingg: Seems previous step just build the binary, didn't actually deploy a cluster.
Do we need delete step here?

prow_config.yaml
jlewi: I went ahead and removed it; I don't think it should depend on scripts since its just go code.

testing/workflows/components/kfctl_go_test.jsonnet
jlewi: Did it now.

testing/workflows/components/kfctl_go_test.jsonnet
jlewi: Done.

ashahba:
Hi @jlewi 
I have a couple of questions about this PR:
1- Are you adding more `e2e` tests cases to this PR or you want us to bring more once this is merged?
2- In order to get the test passing locally yesterday we had to install the following 2 packages:
```
google-cloud-storage
google-api-python-client
```
I'm assuming Python dependencies are listed here: https://github.com/kubeflow/testing/blob/562e5e0a72edf2826a40f8181cfe4009a564f566/test-infra/auto-deploy/Pipfile

jlewi:
@ashahba I think the goal is to continue extending this test until its equivalent to the current E2E test.

I started extending the python code to start calling `generate` and `apply` in #2705

w.r.t to the Google python dependencies; yes the expectation is that the dependencies are available in the Docker image used by the E2E tests.

jlewi:
@ashahba @kunmingg @gabrielwen @kkasravi Tests are finally passing. PTAL.

PR #2679, opened 2019-03-11
Don't create PVCs if no default StorageClass is set

### LINKED ISSUE #2157: Notebook manager UI should disable New PVC option if no default storage class is defined.
Context: #2155 

Creating a PVC to back Jupyter pods can be a major friction point for new users just kicking the tires. 

As #2155 shows users may not realize they can just disable PVC.

We can configure kfctl to test if a cluster has a default storage class and if not disable the use of PVC with jupyter by default.

kimwnasptd:
closes #2157 

If no default StorageClass is set, then the user won't be able to create `New` PVCs for the Notebooks using the Jupyter UI.

/assign @jlewi 

/hold 
wait for #2671

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2679)


jlewi:
@avdaredevil 
/assign @prodonjs 

Could you review this please?

components/jupyter-web-app/rok/kubeflow/rokui/routes.py
```diff
@@ -46,15 +48,8 @@ def post_notebook_route():
...
-    pvc = create_pvc_template()
-    pvc['metadata']['name'] = body['ws_name']
-    pvc['metadata']['namespace'] = body['ns']
-    pvc['spec']['accessModes'].append(body['ws_access_modes'])
-    pvc['spec']['resources']['requests']['storage'] = \
-        body['ws_size'] + 'Gi'
-
     try:
-      create_pvc(pvc)
+      create_workspace_pvc(body)
```
prodonjs: The code here looks very similar to components/jupyter-web-app/default/kubeflow/jupyter/server.py. Again, I wouldn't want to request any significant changes with so little context but it does seem like it would advantageous to factor out this code into a single shared module.

components/jupyter-web-app/OWNERS
```diff
@@ -3,3 +3,5 @@ approvers:
   - kimwnasptd
 reviewers:
   - vkoukis
+  - prodonjs
+  - avdaredevil
```
avdaredevil: Please alphabetically sort

kimwnasptd: Done.

avdaredevil: Great! lgtm-ing

lluunn:
/approve
sorry missed this

Issue #2659, opened 2019-03-07
kfctl Not actually setting GCP IAM roles


jlewi:
I'm using the new kfctl go binary built from **v0.4.0-rc.1-199-g51b410ad**

The logs indicate that the IAM roles are being  set

```
INFO[0040] roles/dataproc.editor adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/compute.networkAdmin adding: Set{serviceAccount:jlewi-0307-basic-admin@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/bigquery.admin adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/cloudsql.admin adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/source.admin adding: Set{serviceAccount:jlewi-0307-basic-admin@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/viewer adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/dataflow.admin adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/cloudbuild.builds.editor adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/storage.objectViewer adding: Set{serviceAccount:jlewi-0307-basic-vm@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/iap.httpsResourceAccessor adding: Set{user:jlewi@google.com} 
INFO[0040] roles/logging.logWriter adding: Set{serviceAccount:jlewi-0307-basic-vm@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/storage.admin adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
INFO[0040] roles/ml.admin adding: Set{serviceAccount:jlewi-0307-basic-user@cloud-ml-dev.iam.gserviceaccount.com} 
[...]

PR #2681, opened 2019-03-12
fix IAM policy set

kunmingg:
fix https://github.com/kubeflow/kubeflow/issues/2659

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2681)

### DIFF
--- a/bootstrap/pkg/utils/iamutils.go
+++ b/bootstrap/pkg/utils/iamutils.go
@@ -26,6 +26,7 @@ import (
 	"google.golang.org/api/cloudresourcemanager/v1"
 	"io/ioutil"
 	"net/http"
+	"strings"
 )
 
 func getServiceClient(ctx context.Context) (*http.Client, error) {
@@ -74,7 +75,9 @@ func GetClearedIamPolicy(currentPolicy *cloudresourcemanager.Policy, pendingPoli
 	serviceAccounts := make(map[string]bool)
 	for _, binding := range pendingPolicy.Bindings {
 		for _, member := range binding.Members {
-			serviceAccounts[member] = true
+			if strings.HasPrefix(member, "serviceAccount:") {
+				serviceAccounts[member] = true
+			}
 		}
 	}
 	clearedPolicy := cloudresourcemanager.Policy{}
--- a/bootstrap/pkg/utils/iamutils.go
+++ b/bootstrap/pkg/utils/iamutils.go
@@ -53,19 +53,6 @@ func transformInterfaceToSlice(inter []interface{}) []string {
 	return ret
 }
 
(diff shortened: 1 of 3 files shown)


bootstrap/pkg/utils/iamutils.go
```diff
@@ -82,6 +69,31 @@ func GetIamPolicy(project string) (*cloudresourcemanager.Policy, error) {
 	return service.Projects.GetIamPolicy(project, req).Context(ctx).Do()
 }
 
+// Gets IAM plicy from GCP for the whole project.
```
lluunn: what does cleared mean? Could you explain in comment?

bootstrap/pkg/utils/iamutils.go
```diff
@@ -82,6 +69,31 @@ func GetIamPolicy(project string) (*cloudresourcemanager.Policy, error) {
...
+func GetClearIamPolicy(currentPolicy *cloudresourcemanager.Policy, pendingPolicy *cloudresourcemanager.Policy) *cloudresourcemanager.Policy {
```
lluunn: GetClearedIamPolicy?

bootstrap/pkg/utils/iamutils.go
kunmingg: updated

jlewi:
Would you mind adding a unittest to verify that ClearedIamPolicy returns the expected policy? We'd like to verify we don't delete accounts that shouldn't be deleted.

kunmingg:
@jlewi 
updated

bootstrap/pkg/utils/iamutils_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"google.golang.org/api/cloudresourcemanager/v1"
+)
+
+func Test(t *testing.T) {
+	tests := []struct {
```
jlewi: Thanks. So the typical style in go is to define a list of testCases where each testCase would specify the input and the expected output.

Would it be possible to write the test that way?

e.g. add a struct variable that should equal the expectedPolicy returned by GetClearediamPolicy. Then do a deepEqual to compare the expected and actual. 

Something like:
https://github.com/kubeflow/tf-operator/blob/cf01b5125e25239042e1caef1ef44f7aaf2337e3/pkg/apis/tensorflow/v1beta1/defaults_test.go#L117

kunmingg: updated

jlewi:
/lgtm
/approve
Thanks @kunmingg

PR #2687, opened 2019-03-12
Skip check to oauth_id/oauth_secret if using basic auth

### LINKED ISSUE #2657: kfctl generate won't succeed if first attempt failed
kfctl generate failed for me as follows

```
kfctl generate all
INFO[0000] generating kubeflow application              
ERRO[0002] couldn't generate KfApp: could not generate kssonnet under ks_app Error: Email is not set in default nor passed.
```

I then retry and specify the flag

```
/home/jlewi/git_kubeflow/bootstrap/bin/kfctl generate all --email=jlewi@google.com
INFO[0000] generating kubeflow application              
ERRO[0000] couldn't generate KfApp: could not generate deployment manager configs under gcp_config Error: cannot create directory mkdir /home/jlewi/git_kubeflow-dev/kubeflow-deployments/jlewi-0307-basic/gcp_config: file exists 
```

We should fix this. The desired behavior would be I should be able to rerun and specify the parameters and have it succeed.

/cc @kunmingg @gabrielwen @lluunn

### LINKED ISSUE #2658: kfctl apply requires oauth client id and secret even if using basic auth
I get the following error when trying to spin up Kubeflow using basic auth

```
ERRO[0062] couldn't apply KfApp: gcp apply could not create secrets Error At least one of --oauth_id or ENV `CLIENT_ID` needs to be set. 
```

But a client id and oauth id should only be required if using IAP; if using basic auth we shouldn't need this.

gabrielwen:
Fixes #2658
Fixes #2657

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2687)


bootstrap/go.sum
```diff
@@ -1,16 +1,19 @@
 bitbucket.org/ww/goautoneg v0.0.0-20120707110453-75cd24fc2f2c/go.mod h1:1vhO7Mn/FZMgOgDVGLy5X1mE6rq1HbkBdkF/yj8zkcg=
 cloud.google.com/go v0.0.0-20160913182117-3b1ae45394a2/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+cloud.google.com/go v0.34.0 h1:eOI3/cP2VTU6uZLDYAoic+eyzzB9YyGmJ7eIjl8rOPg=
```
lluunn: the change seems should not modify go.sum?

gabrielwen: reverted

Issue #2661, opened 2019-03-08
kfctl doesn't create secret with username and password for basic auth


jlewi:
I deployed Kubeflow using basic auth using the kfctl binary.

The secret ` kubeflow-login` containing the username and password isn't actually created.

It would be good to have a command in kfctl for setting the secret and username.
Something like

kfctl create-basic-auth-secret --username=<> --password=....

So users could easily update/change it if they forget.

gabrielwen:
do we expect people using other platforms also use basic auth?  e.g. do we need to add a new method to the interface?
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/apis/apps/group.go#L89-L94

jlewi:
Yes we expect people on other platforms to use basic auth.

@gabrielwen Is this fixed can we close it?

gabrielwen:
Still need to add the cmd `create-basic-auth-secret`.  Currently it's passed during `kfctl init`

jlewi:
@gabrielwen I also think we should avoid storing the encrypted password if possible.

We generally want to avoid storing sensitive information in the app directory because the expectation is people will check their app into source control.

The pattern we follow with other sensitive information like client_id and client_secret is to read it from an environment variable. This way we avoid storing it anywhere.

gabrielwen:
Yep, makes sense.  will make the change tomorrow first thing.

jlewi:
@gabrielwen Can we close this issue?

PR #2690, opened 2019-03-13
Create secret for basic login

gabrielwen:
Relates to #2661

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2690)


gabrielwen:
PR is ready.

bootstrap/pkg/client/gcp/gcp.go
```diff
@@ -719,7 +723,7 @@ func (gcp *Gcp) writeGcpSecret(client *clientset.Clientset, secretName string, d
 	return err
 }
 
-func (gcp *Gcp) createGcpSecret(ctx context.Context, client *clientset.Clientset,
+func (gcp *Gcp) createGcpServiceAcctSecret(ctx context.Context, client *clientset.Clientset,
```
lluunn: please add comment for func (and below).
Like what's created

bootstrap/cmd/kfctl/cmd/apply.go
```diff
@@ -43,8 +43,10 @@ var applyCmd = &cobra.Command{
 			return
 		}
 		options := map[string]interface{}{
-			string(kftypes.OAUTH_ID):     applyCfg.GetString(string(kftypes.OAUTH_ID)),
-			string(kftypes.OAUTH_SECRET): applyCfg.GetString(string(kftypes.OAUTH_SECRET)),
+			string(kftypes.OAUTH_ID):            applyCfg.GetString(string(kftypes.OAUTH_ID)),
```
lluunn: It's fine for now, but this should probably not be passed at apply.

bootstrap/pkg/client/gcp/gcp.go
```diff
@@ -802,13 +790,87 @@ func (gcp *Gcp) createSecrets(options map[string]interface{}) error {
...
+	userEmail := getSA(gcp.GcpApp.Name, "user", gcp.GcpApp.Spec.Project)
+	if err := gcp.createGcpServiceAcctSecret(ctx, k8sClient, adminEmail, ADMIN_SECRET_NAME); err != nil {
+		return fmt.Errorf("cannot create admin secret %v Error %v", ADMIN_SECRET_NAME, err)
+
+	}
+	if err := gcp.createGcpServiceAcctSecret(ctx, k8sClient, userEmail, USER_SECRET_NAME); err != nil {
+		return fmt.Errorf("cannot create user secret %v Error %v", USER_SECRET_NAME, err)
+
+	}
+	if err := gcp.createIapSecret(ctx, k8sClient, options); err != nil {
```
lluunn: shouldn't this be done conditionally (!UseBasicAuth?)
same as below

bootstrap/pkg/client/gcp/gcp.go
```diff
@@ -802,13 +790,87 @@ func (gcp *Gcp) createSecrets(options map[string]interface{}) error {
...
+	}
+	if username == "" {
+		return fmt.Errorf("At least one of --%v or ENV `%v` needs to be set.",
+			string(kftypes.BASIC_AUTH_USERNAME), BASIC_AUTH_USERNAME)
+	}
+
+	password := ""
+	if options[string(kftypes.BASIC_AUTH_PASSWORD)] != nil &&
+		options[string(kftypes.BASIC_AUTH_PASSWORD)].(string) != "" {
+		password = options[string(kftypes.BASIC_AUTH_PASSWORD)].(string)
```
lluunn: why do we have kftypes.BASIC_AUTH_PASSWORD and BASIC_AUTH_PASSWORD?

bootstrap/pkg/client/gcp/gcp.go
gabrielwen: added.

bootstrap/pkg/client/gcp/gcp.go
gabrielwen: metric-controller also uses it.  added a comment as TODO such that we should check not only `UseBasicAuth`, but also if `metric-controller` is part of components.

bootstrap/pkg/client/gcp/gcp.go
gabrielwen: it's part of `CliOptions` (string).

bootstrap/pkg/client/gcp/gcp.go
gabrielwen: also, `UseBasicAuth` checks is in `createBasicAuthSecret`

bootstrap/pkg/client/gcp/gcp.go
lluunn: write to GCP --> insert to cluster?

bootstrap/pkg/client/gcp/gcp.go
lluunn: I mean, why do we duplicate things at two places?

bootstrap/pkg/client/gcp/gcp.go
gabrielwen: done.

bootstrap/pkg/client/gcp/gcp.go
gabrielwen: was going to use ENV as well, thought twice, removed.

bootstrap/cmd/kfctl/cmd/apply.go
gabrielwen: thought twice, make them init params and write to app.yaml and encrypt password.

bootstrap/cmd/kfctl/cmd/init.go
```diff
@@ -48,7 +49,20 @@ or a <name>. If just <name> a directory <name> will be created in the current di
 		repo := initCfg.GetString(string(kftypes.REPO))
 		project := initCfg.GetString(string(kftypes.PROJECT))
 		init_gcp := initCfg.GetBool(string(kftypes.SKIP_INIT_GCP_PROJECT))
-		basic_auth := initCfg.GetBool(string(kftypes.USE_BASIC_AUTH))
+
+		useBasicAuth := initCfg.GetBool(string(kftypes.USE_BASIC_AUTH))
```
lluunn: I think these should not be at init, but at generate.
https://github.com/kubeflow/kubeflow/issues/2691#issuecomment-472970551

gabrielwen: Let's make a follow-up PR for it.  Want to submit it and test with another PR.

bootstrap/pkg/client/gcp/gcp.go
kunmingg: Should be safe to remove IapSecret when UseBasicAuth.
I'll change metric-controller to skip IapSecret when not available.

gabrielwen: Done.

PR #2692, opened 2019-03-13
moved the Profiles golang controller kustomize manifests around

swiftdiaries:
`kustomize build` currently works if the kustomization.yaml refers to resources in/within the same directory.
If it is above, it fails. 
This PR moves things around accordingly
/cc @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2692)


components/profile-controller/config/kustomization.yaml
```diff
@@ -9,24 +9,25 @@ namespace: profile-controller-system
...
+commonLabels:
+    kustomize.component: profiles-controller
 
 # Each entry in this list must resolve to an existing
 # resource definition in YAML.  These are the resource
 # files that kustomize reads, modifies and emits as a
 # YAML string, with resources separated by document
 # markers ("---").
 resources:
-- ../rbac/rbac_role.yaml
```
lluunn: Do you mean this ../ doesn't work?
Is this a known kustomize issue?
It's weird that they generate it this way and it's not working..

swiftdiaries: Yeah, I think the latest kustomize version removed it

swiftdiaries: kustomize 2.X

lluunn: I see. Is it intended, or is it a bug?
Do they have a issue to track?
The problem is if we run the generate script again, it will revert this change?

swiftdiaries: Related issue in kube-builder: https://github.com/kubernetes-sigs/kubebuilder/issues/595
Modified the directory layout as per instructions there
The generate script does not modify the manifests.

swiftdiaries: The latest change is recommended but oddly doesn't work for remote builds. I still can't use `../` for building manifests off of git repos.

PR #2697, opened 2019-03-14
Fix make build from within bootstrap

### LINKED ISSUE #2696: "make build" under "bootstrap" module fails due to some go modules missing under /tmp inside Docker.
While running `make build` inside `bootstrap` module I get:
```
The command '/bin/sh -c go mod download' returned a non-zero code: 1
make: *** [build] Error 1
```
I noticed some required `go` modules under `hack/v2.zip` are missing from within Docker.

ashahba:
Fixes #2696

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2697)

### DIFF
--- a/bootstrap/Dockerfile
+++ b/bootstrap/Dockerfile
@@ -4,7 +4,9 @@
 ARG GOLANG_VERSION=1.11.5
 FROM golang:$GOLANG_VERSION as bootstrap_base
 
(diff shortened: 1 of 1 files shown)


ashahba:
Looks like a recent PR fixed this and it's not longer needed.

Issue #2691, opened 2019-03-13
kfctl config issue


lluunn:
## Background
Currently, a KFApp has two layers:
1. parent app, the coordinator
2. child apps, a platform (gcp, minikube) and a packageManager (ksonnet, kustomize)

and different actions: init, generate, apply, delete, ..

When we run a command, e.g. `kfctl generate --A=B ...`, coordinator calls GcpApp.Generate and KsApp.Generate with the flags (the `option` variable in the code). 
I think it's not working properly now:
- some flag (e.g. UseBasicAuth) is needed is both platform and packageManager, and both generate and apply. 
- Gcp.Spec and KsApp.Spec stores different fields.
- I think it's confusing that whether a field, basicAuth, should be in `option`, or gcp.spec, or ks.spec.
- ksonnet.Generate right now has no information on whether to use basicAuth or not.

## Proposal
One possible solution is to save everything in `coordinator.spec`. And for any action (init or generate or else), coordinate passes `coordinator.spec` to platform and packageManager.
`gcpApp.Generate(coordinator.Spec)`

In addition, we can dump coordinator.spec to app.yaml. Then for any action, coordinator will
1. read app.yaml in as the spec
2. update the spec with new flags
3. pass spec to GcpApp and ksApp

This way users can just modify the app.yaml, and that will take effect.

cc @kkasravi @jlewi @kunmingg

lluunn:
@gabrielwen

kkasravi:
I think this makes sense. Each subcommand would call a function that takes the action and flags. This is merged with the app.yaml, unmarshalled into one Spec and given to coordinator. All platform and package managers share this one spec which may be mutated by the platform prior to passing to  the package manager.

kkasravi:
We suggested a design doc for kfctl, this would be a good time to produce one, since it is a service with now a wider audience @lluunn @kunmingg @gabrielwen @swiftdiaries

gabrielwen:
Discussed offline with @lluunn , we agree that information should be able to be shared between different components and be accessed easily.  Thanks for pointing this issue out - I'm gonna need it as well.

@kkasravi : Totally agree that we need a design doc for kfctl.  We have reached a stage where it's not only about kfctl CLI, we also need to consider compatibility with click-to-deploy app and extensibility to other platforms.  Let's make kfctl work for 0.5 for now and pivot back to the design doc :)

swiftdiaries:
+1 on the design doc.

[...]

PR #2705, opened 2019-03-15
Kfctl go E2E test should deploy kubeflow

### LINKED ISSUE #2706: kfctl tries to call ks init even on ks generate platform
I'm using #2705 which has  a WIP E2E test for kfctl.

It looks like when I call `kfctl generate platform` it ends up calling `ks init` which fails because there is no cluster to connect to.

Here are the logs.

```
util.py                     45 INFO     Running: /home/jlewi/git_kubeflow/bootstrap/bin/kfctl generate -V platform 
cwd=/tmp/kfctl-2145-4086
util.py                     60 INFO     Subprocess output:

util.py                     71 INFO     time="2019-03-14T21:47:01-07:00" level=info msg="reading from /tmp/kfctl-2145-4086/app.yaml"
util.py                     71 INFO     time="2019-03-14T21:47:01-07:00" level=info msg="Setting acmeEmail to "
util.py                     71 INFO     time="2019-03-14T21:47:01-07:00" level=info msg="Setting ipName to kfctl-2145-4086-ip"
util.py                     71 INFO     time="2019-03-14T21:47:01-07:00" level=info msg="Setting hostname to kfctl-2145-4086.endpoints.kubeflow-ci-deployment.cloud.goog"
util.py                     71 INFO     time="2019-03-14T21:47:01-07:00" level=info msg="Setting mysqlPd to kfctl-2145-4086-storage-metadata-store"
util.py                     71 INFO     time="2019-03-14T21:47:01-07:00" level=info msg="Setting minioPd to kfctl-2145-4086-storage-artifact-store"
util.py                     71 INFO     time="2019-03-14T21:47:01-07:00" level=info msg="Ksonnet.Generate Name kfctl-2145-4086 AppDir /tmp/kfctl-2145-4086 Platform gcp"
util.py                     71 INFO     time="2019-03-14T21:47:16-07:00" level=fatal msg="couldn't get server version info. Error: Get https://35.229.112.22/version: dial tcp 35.229.112.22:443: connect: connection timed out"
```

It looks to me like the bug is in `kfctl generate`
https://github.com/kubeflow/kubeflow/blob/b42734af7845087b6bd42e36d01422cb34ec8aa0/bootstrap/pkg/client/coordinator/coordinator.go#L386

It looks like the switch statement is always falling through and calling generate on the platform.

jlewi:
E2E test for kfctl should deploy kubeflow with basic auth
    
* The E2E test attempts to call delete but delete isn't working.
* kfctl delete fails because there is no kubeconfig file created.
* The delete step is marked as an expected failure.
    
Improve logging in kfctl with go binary
* Log the name and status of GCP DM operations
 * Print out DM creation errors so users see things like quota issues.
    
Fix #2706 - The coordinator commands should only invoke the commands apply/g
enerate/delete for the resources specified.
      * There was a bug in the switch statements and we were always calling
        generate/apply/delete for platform & k8s and not respecting the parameter
    
Attempt to fix #2706
    
* We want to eliminate the need to talk to a K8s server when calling ks init
* Hardcode the server address to 127.0.0.1
* Hardcode K8s version to 1.11.7
* The communication with the K8s server was coming because we were trying
        to talk to the K8s master to get the K8s version but we don't need to
        do that if we specify it.
    
    * Add filename log hook to kfctl so that we can emit the filename and line
      number where errors occur.
    
    * On init we Don't need to call GetConfig which tries to read the kubeconfig file on generate.
    
    Add retries to generate and apply because running under argo kfctl seems
    to randomly exit.

Related to #2610  E2E test for kfctl go binary

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2705)


bootstrap/pkg/client/ksonnet/ksonnet.go
```diff
@@ -487,13 +487,15 @@ func (ksApp *ksApp) Init(resources kftypes.ResourceEnum, options map[string]inte
 func (ksApp *ksApp) initKs(config *rest.Config) error {
 	newRoot := path.Join(ksApp.KsApp.Spec.AppDir, ksApp.KsName)
 	ksApp.KsEnvName = kstypes.KsEnvName
-	k8sSpec := kftypes.GetServerVersion(kftypes.GetClientset(config))
+	// We hard code the K8s spec because we won't have a cluster to talk to when calling init.
```
kkasravi: If we add serverVersion to the ClientSpec then it could be set by the platform and the package managers would use this value rather than attempting to contacting the server by using credentials. 

Similarly if the platform sets client credentials in the ClientSpec then the package managers would use this rather than $HOME/.kube/config's context. 

This would align with how the platform updates other values in ClientSpec like components, componentParameters and the package managers use this updated ClientSpec. 

I had discussed with @gabrielwen at the summit about this approach for fetching credentials. 
I have the PR for just using the ClientSpec about ready. (https://github.com/kubeflow/kubeflow/issues/2691)

jlewi: Only ksonnet needs to know the K8s version and that's so it can generate an appropriate k8s.libsonnet file from the corresponding swagger spec for the K8s API.

I think we can just hard code it.

* Only ksonnet needs to know the K8s API version; kustomize doesn't
  * ksonnet will likely be removed before the K8s API version is an issue
  * I suspect ksonnet might even generate a new lk8s lib file when a new environment is added (which we do later).

> Similarly if the platform sets client credentials in the ClientSpec then the package managers would use this rather than $HOME/.kube/config's context.

Why would client credentials need to be set in ClientSpec? How is that different then using a .kubeconfig file?

jlewi:
When running kfctl under Argo as part of the test; I'm seeing random failures where it just appears to exit 
on generate or apply. It doesn't seem like an error occurred or it was preempted.

If we look at the logs last line just indicates it was waiting for the deployment.

```
2019-03-15T14:32:04.473205048Z  main            util.py                     71 INFO     time="2019-03-15T14:31:40Z" level=warning msg="Deployment operation name: operation-1552660191586-58422df2d2a4e-6bf2031f-027c70b5 status: RUNNING" filename="gcp/gcp.go:331"
```

[jlewi-kfctl-test-2705-0315-072101-3799861166.log.txt](https://github.com/kubeflow/kubeflow/files/2971577/jlewi-kfctl-test-2705-0315-072101-3799861166.log.txt)

[jlewi-kfctl-test-2705-0315-072101-3799861166.event.log.txt](https://github.com/kubeflow/kubeflow/files/2971575/jlewi-kfctl-test-2705-0315-072101-3799861166.event.log.txt)

jlewi:
In the most recent run kfctl delete all failed because there is no kubeconfig file.

```
time="2019-03-15T17:48:33Z" level=fatal msg="could not open /mnt/test-data-volume/kubeflow-presubmit-kfctl-go-2705-05adf28-6330-dfdf/kfctl_test/.kube/kube
config Error stat /mnt/test-data-volume/kubeflow-presubmit-kfctl-go-2705-05adf28-6330-dfdf/kfctl_test/.kube/kubeconfig: no such file or directory" filenam
e="apps/group.go:180"
```

bootstrap/pkg/client/ksonnet/ksonnet.go
kkasravi: Different platforms will set different credentials. If the package managers use the credentials set by the platform then we avoid package managers using .kube/config when they should be using the platform credentials.

PR #2726, opened 2019-03-18
Launch trtserver so that CUDA driver compatibility is enabled

deadeyegoodwin:
The container entrypoint must run to ensure driver compatibility is
enabled correctly. Fix spec to not use "command" so that container
launch runs entrypoint.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2726)


deadeyegoodwin:
Rebase to master in attempt to pick up presubmit fixes

PR #2739, opened 2019-03-19
Add first unit tests for server components.

prodonjs:
This will be the first change to start introducing testing practices into the Central Dashboard component.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2739)


prodonjs:
/assign @avdaredevil 

Hey AP, I'm just starting the process of adding some tests here. The server tests are straightforward so I'm starting there and adding dependencies to the **Makefile** to ensure that tests are run as part of the presubmit process. I'm going to start looking at adding unit tests for the client-side components and then ensuring those are invoked when `npm test` is called.

prodonjs:
Closing this in favor of #2758 which includes testing for both server and client-side code.

PR #2746, opened 2019-03-20
add proxy component to ksonnet registry

IronPan:
This change add the proxy agent to the kubeflow cluster as part of gcp suit.
This change doesn't include the change to the new deployment go binary. I'll add it in a separate PR

Currently waiting for image to be released

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2746)


bootstrap/config/kfctl_basic_auth.yaml
```diff
@@ -28,7 +28,7 @@ components:
   - basic-auth
   - basic-auth-ingress
   - centraldashboard
-  - cert-manager
+  - proxy-agent
```
kunmingg: You can add another entry to config folder like 
kfctl_proxy.yaml

components/gcp-click-to-deploy/app-config.yaml
```diff
@@ -18,6 +18,8 @@ defaultApp:
     prototype: cert-manager
   - name: iap-ingress
     prototype: iap-ingress
+  - name: proxy-agent
+    prototype: proxy-agent
```
kunmingg: proxy-agent can be an optional components based on user's choice of how to access UI.
We should make the change on js side.

IronPan: is this file for local test only? if not adding it, will local test be able to deploy proxy agent component?

bootstrap/config/kfctl_basic_auth.yaml
IronPan: done

bootstrap/config/kfctl_proxy.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - seldon
+  - tensorboard
+  - tf-serving
+  - tf-training
+components:
+  - ambassador
+  - argo
+  - centraldashboard
+  - cert-manager
+  - cloud-endpoints
```
kunmingg: We might just remove cert-manager and cloud-endpoints as well.

bootstrap/config/kfctl_proxy.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - notebook-controller
+  - pipeline
+  - profiles
+  - proxy-agent
+  - pytorch-operator
+  - spartakus
+  - tensorboard
+  - tf-job-operator
+componentParams:
+  cert-manager:
```
kunmingg: Same here and below

components/gcp-click-to-deploy/app-config.yaml
kunmingg: Yeah, this one is for local test only.
We can edit config in javascript based on a param?

scripts/gke/util.sh
```diff
@@ -246,7 +247,7 @@ gcpKsApply() {
   createKsEnv
 
   if [[ -z $DEFAULT_KUBEFLOW_COMPONENTS ]]; then
-    export KUBEFLOW_COMPONENTS+=',"cloud-endpoints","cert-manager","iap-ingress"'
+    export KUBEFLOW_COMPONENTS+=',"cloud-endpoints","cert-manager","iap-ingress", "proxy-agent"'
```
kunmingg: If use ```proxy-agent```, then ```cloud-endpoints","cert-manager","iap-ingress``` is not needed.
Can we control by parameter?
Or we can just implement the logic in kfctl-go which should be easier

jlewi:
@IronPan Can we close this PR?

PR #2751, opened 2019-03-20
fix make build; set IAM policy following read/modify/write

kunmingg:
1. go binary need to import k8s 1.13.4 (kustomize) and 1.10.4 (ksonnet)
2. set IAM policy need to follow read/modify/write mode

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2751)


kunmingg:
updated

PR #2755, opened 2019-03-21
kfctl install istio

### LINKED ISSUE #2549: Create secret in istio namespace when using istio
When we deploy kubeflow with istio installed, the ingress and related resources are in istio-system namespace. So we should create secrets needed by those resources in istio-system namespace, including:
- kubeflow-oauth for clientID, clientSecret
- admin-gcp-sa for backend_updater pod

cc @kunmingg @jlewi

lluunn:
fix #2549
for #2507

/cc @kunmingg @gabrielwen @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2755)


lluunn:
/approve
thanks Gabriel

PR #2756, opened 2019-03-21
Fix error in iap component

lluunn:
previously, 
```
    + if params.privateGKECluster == "false" then [
      self.certificate,
    ] else [] + if isCloudEndpoint(params.hostname) then [
      self.cloudEndpoint,
    ] else [] + if !params.useIstio then [
      self.service,
      self.deploy,
    ] else [],
```
`else [] + if isCloudEndpoint` is in the same `else` clause.
The cloudEndpoint won't be there if privateGKECluster==false.

/cc @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2756)


lluunn:
test error kfctl: #2760

Issue #2618, opened 2019-03-04
[Front-End][Tracker] Create front-end tests for dashboard and notebooks


avdaredevil:
## About:
Tracker issue for all unit and integration testing we want to add to various front-end projects within Kubeflow

### Dashboard
General functionality and user flows

### Notebooks
We would like to test the notebooks page in selenium to test the construction of a notebook, which will eventually be made into a presubmit test. For now, it should be able to run manually

/assign @avdaredevil @prodonjs 
/cc @jlewi

prodonjs:
The code for the notebooks front-end is in [components/jupyter-web-app](https://github.com/kubeflow/kubeflow/tree/master/components/jupyter-web-app/default/kubeflow/jupyter). It probably makes sense to have tests included and run against that component rather than working through centraldashboard but ideally, our test framework can be used for all of the front-end components that need to be tested.

avdaredevil:
Yeah I was wondering if we want to make another Kubeflow repo for testing all different components, or configure at the component level?

jlewi:
So we have tests for central dashboard but not notebooks.

I think we should file a separate issue for notebooks that we add to 0.6 and close this issue.

avdaredevil:
Closing since Notebooks is not needed for this release

PR #2758, opened 2019-03-21
Add unit tests to Central Dashboard component

prodonjs:
This PR adds the Karma test runner framework and uses our existing Webpack configuration with Jasmine to be able to run unit tests on components.

I added a single test here to demonstrate the structure and usage going forward. We'll have to be mindful of how it's used in the build process since Karma depends on Chrome (or other browsers).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2758)


prodonjs:
This won't build in Docker at the moment but for unit testing purposes, I think we should be able to use the [Karma PhantomJS launcher](https://github.com/karma-runner/karma-phantomjs-launcher).

prodonjs:
It looks like PhantomJS is no longer maintained and Puppeteer is the way to go.

https://developers.google.com/web/updates/2017/06/headless-karma-mocha-chai

This also looks like it might be an interesting tool to use for integration testing.

prodonjs:
The last thing that we will want to do is figure out how to get these to run as part of pres

prodonjs:
Mistakenly closed the PR there. The last thing we're going to want to figure out is how to include these tests in the presubmit job. There are some issues with running Puppeteer inside of Docker (see https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker). I don't think we want to add all of those commands to our Dockerfile within this directory so we'll probably need to figure out how to make sure the container that runs our tests has this stuff configured.

prodonjs:
@avdaredevil The latest commit restructures the tests to ensure that they are properly isolated from each other and they provide a single Webpack entry point to avoid loading the custom elements polyfills more than once in the Karma environment.

Also, take a look at the most recent commit in particular the [namespace-selector_test.js](https://github.com/kubeflow/kubeflow/pull/2758/commits/736a6ec7147388e6c031dbbf41450a6732db7179#diff-78b8296af9f7b4e46eced179c1e1f5f8) file. Testing these components is a little tricky if they rely on an automatically invoked AJAX call. We need to intercept the AJAX to be able to return a mock response. Then, we also need to wait on the response to be received by the component and have its DOM stamped out based on the response before making any assertions. We can probably make some improvements to this test code including the use of async/await, but I'm also wondering if we need to add some additional events to our component to make it easier to test.

I've never written tests for Polymer but I've tried following the guide at https://polymer-library.polymer-project.org/3.0/docs/tools/tests#async as much as possible. Once we establish a pattern for this stuff it should become pretty mechanical to write tests going forward.

components/centraldashboard/app/k8s_service.ts
```diff
@@ -16,7 +16,7 @@ export class KubernetesService {
       const {body} = await this.k8sApi.listNamespace();
       return body.items;
     } catch (err) {
-      console.error('Unable to fetch Namespaces', err.body ? err.body : err);
+      console.error('Unable to fetch Namespaces:', err.body ? err.body : err);
```
avdaredevil: _**Succinct**_
```suggestion
      console.error('Unable to fetch Namespaces:', err.body || err);
```

components/centraldashboard/app/k8s_service.ts
```diff
@@ -27,7 +27,9 @@ export class KubernetesService {
       const {body} = await this.k8sApi.listNamespacedEvent(namespace);
       return body.items;
     } catch (err) {
-      console.error('Unable to fetch Namespaces', err.body ? err.body : err);
+      console.error(
+          `Unable to fetch Events for ${namespace}:`,
+          err.body ? err.body : err);
```
avdaredevil:
```suggestion
          err.body || err);
```

components/centraldashboard/public/components/activities-list_test.js
```diff
@@ -0,0 +1,102 @@
...
+    <activities-list id="${ACTIVITIES_LIST_ID}"></activities-list>
+  </template>
+</test-fixture>
+`;
+
+describe('Activities List', () => {
+    let activitiesList;
+
+    beforeAll(() => {
+        const div = document.createElement('div');
```
avdaredevil: We could re-use `lit-html` here!

avdaredevil:
Looks great! I think we can improve/modularize once we have more tests. Feel free to condense if you like the approaches.

/lgtm
/approve

components/centraldashboard/public/components/activity-view.js
```diff
@@ -78,7 +78,7 @@ export class ActivityView extends PolymerElement {
             },
             activities: {
                 type: Array,
-                value: [],
+                value: () => [],
```
avdaredevil: Damn, I knew about this. I should have caught it in my review

avdaredevil:
Hey @prodonjs you have conflicts

components/centraldashboard/public/components/activity-view.js
prodonjs: Ahh no biggie. I needed to initialize the array to something in order to use the splice and push methods but it never dawned on me that I'd be setting it up to use the same object for all instances of the element. It wouldn't have really mattered since we'd only ever have one `activity-view` on a page, but it made my tests fail when they ran in a particular order.

prodonjs:
@avdaredevil I fixed all the merge conflicts this morning but missed your suggestions. I'll add those and have it ready first thing tomorrow morning. Let me know if you see anything else in this final revision.

components/centraldashboard/public/components/main-page.js
```diff
@@ -28,6 +28,7 @@ import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 
 import css from './main-page.css';
 import template from './main-page.pug';
+import logo from '../assets/kf-logo_64px.svg';
```
avdaredevil: Awesome! Though this change could have been in a separate PR?

prodonjs: Actually I changed it because it made for easier testability. Otherwise, I would have had to mock window.fetch and make it resolve a promise with an object with a .text() method which would resolve to something that could be injected into the DOM. This seemed like an easier path and it eliminated the request when deployed.

components/centraldashboard/karma.conf.js
```diff
@@ -0,0 +1,40 @@
...
+    enforce: 'post',
+    test: /\.js$/,
+    use: {
+        loader: 'istanbul-instrumenter-loader',
+        options: {esModules: true},
+    },
+    exclude: /node_modules|_test\.js$/,
+});
+
+module.exports = function(config) {
```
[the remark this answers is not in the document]
prodonjs: Yeah I like that change. I just copy/pasted from Karma's docs and tweaked but it probably makes sense to simplify with arrows and other ES6+ stuff when we can.

components/centraldashboard/Makefile
```diff
@@ -17,9 +17,20 @@ endif
 TAG := $(shell date +v%Y%m%d)-$(GIT_VERSION)
 all: build
```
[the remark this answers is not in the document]
prodonjs: I'm could probably hack my way through writing Powershell, but honestly I'd be a lot slower and need to consult a reference the entire time. Do you want to chain that on as a later PR? Or perhaps I can take the Makefile changes out of this change and they can become part of that subsequent PR?

components/centraldashboard/public/components/activities-list_test.js
prodonjs: I played around with this a bit and I think we probably could eliminate this pattern going forward. It will probably make sense to add some additional helper functions for setup and tear down work in unit tests.

components/centraldashboard/app/k8s_service.ts
prodonjs: Addressed.

components/centraldashboard/app/k8s_service.ts
prodonjs: Addressed.

prodonjs:
@avdaredevil I think everything that was left open should be addressed other than the question of how to handle the Makefile and the Windows version. With respect to makefiles on Windows, I didn't notice any other ports for other component directories. Would those have to be authored to build those artifacts from source? Are there any alternatives that would allow a Windows machine to interpret and execute the contents of a standard Makefile? (I saw https://chocolatey.org/packages/make but I have no recent experience developing on Windows).

avdaredevil:
When running the make command which I have installed locally some binaries weren't behaving the same way on windows, and things like `>/dev/null` are not cross-platform. We could use node as the shared context to perform the windows builds, since that has to be a requirement to do dev-work with this component, and is easy to write cross-platform scripting with.

components/centraldashboard/Makefile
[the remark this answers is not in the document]
avdaredevil: Either works, though I have access to your fork I think, so I can just commit that in this?

prodonjs:
@avdaredevil This is now rebased and ready to merge. Needs your LGTM again though.

avdaredevil:
#### child of #2618

PR #2765, opened 2019-03-21
cache k8s spec in kubeflow so it can be used by ksonnet lib

kunmingg:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2765)


kunmingg:
rebase to resolve conflict

PR #2772, opened 2019-03-21
replace download k8s manifest with ksonnet entry

kunmingg:
The goal is to avoid shell out and run ```kubectl apply```:
To use gcp.go/ksonnet.go as a library, it need to support concurrent run and cannot use local kubeconfig. 

For stackdriver, will config with gcp/prometheus

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2772)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -22,6 +22,7 @@ import (
 	"github.com/cenkalti/backoff"
 	"github.com/deckarep/golang-set"
 	"github.com/ghodss/yaml"
+	bootstrap "github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/app"
 	configtypes "github.com/kubeflow/kubeflow/bootstrap/config"
 	kftypes "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps"
 	kfdefs "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps/kfdef/v1alpha1"
@@ -42,7 +43,6 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


kunmingg:
rebase 

@lluunn Can you review istio part when have time?
I imported your function from k8sUtil

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
```diff
@@ -117,6 +118,27 @@ func (ksApp *ksApp) Apply(resources kftypes.ResourceEnum) error {
 			return fmt.Errorf("could not change directory to %v Error %v", ksApp.Spec.AppDir, err)
 		}
 	}
+	// Install Istio
```
lluunn: Why move istio to ksonnet part?

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -637,65 +590,6 @@ func (gcp *Gcp) generateDMConfigs() error {
...
-	if k8sSpecsDirErr != nil {
-		return fmt.Errorf("cannot create directory %v Error %v", k8sSpecsDir, k8sSpecsDirErr)
-	}
-	daemonsetPreloaded := filepath.Join(k8sSpecsDir, "daemonset-preloaded.yaml")
-	url := "https://raw.githubusercontent.com/GoogleCloudPlatform/container-engine-accelerators/stable/nvidia-driver-installer/cos/daemonset-preloaded.yaml"
-	urlErr := gogetter.GetFile(daemonsetPreloaded, url)
-	if urlErr != nil {
-		return fmt.Errorf("couldn't download %v Error %v", url, urlErr)
-	}
-	rbacSetup := filepath.Join(k8sSpecsDir, "rbac-setup.yaml")
```
lluunn: Where does stackdriver part go?

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
kunmingg: One benefit is to apply all resources in k8s within ksonnet by default? 
We build k8s config for accessing cluster in ksonnet.

bootstrap/pkg/kfapp/gcp/gcp.go
kunmingg: It's not needed if we apply gcp/prometheus.
Will add in separate PR.

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
lluunn: I guess it makes sense as we are moving to kustomize. (I don' think we should convert istio manifest to ks)

One thing needs to be done in GCP is create istio namespace.
gcp apply will create secret in namespace `istio-system`, and the namespace is created in `dependencies/istio/install/crds.yaml`

So you need to create that namespace in gcp.apply, and remove the namespace part in  crds.yaml

kunmingg:
updated

kunmingg:
@lluunn 
Now move istio to gcp.go so logic can be shared to kustomize.

Issue #2641, opened 2019-03-06
Inject common config into Jupyter Pods


jlewi:
We need a way for admins to define common data to be injected into Jupyter pods.

Example: Adding environment variables and secrets (e.g. for GCP.)

There are a couple options

1. Use a custom MutatingWebAdmissionController 
1. Use a PodPreset

    * See https://github.com/kubeflow/kubeflow/tree/master/kubeflow/credentials-pod-preset
    * https://kubernetes.io/docs/concepts/workloads/pods/podpreset/
    * I think one issue on GCP at least is that in K8s 1.11 you need an alpha GKE cluster; not sure if that's true in later GCP clusters

1. Jupyter CR explicitly supports letting users configure this.

/cc @ioandr @kimwnasptd @r2d4

jlewi:
Example webhook for sidecar injection.
https://github.com/morvencao/kube-mutating-webhook-tutorial

We could probably modify that to do environment variables.

lluunn:
Next step: testing, install by kfctl and boostrapper

lluunn:
Followup issues:
Generalize the admission webhook: #2844 
Update notebook controller: #2770

PR #2776, opened 2019-03-22
kfctl/bootstrapper installs admission webhook

lluunn:
/cc @kunmingg 

#2641

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2776)


kunmingg:
Do you want to move this package into GCP?
Just noticed its only for GCP.

lluunn:
ah good point

PR #2778, opened 2019-03-22
Add NotFoundView to catch invalid paths

### LINKED ISSUE #2759: Central Dashboard - Provide a not found view for links that aren't valid
Current behavior allows a user to directly navigate to a path like `/bad-page` and still see the home page content.

This should be amended to provide some view indicating that the path is not a valid path. This can be addressed by updating the `_routePageChanged` method in the `main-page` component.

prodonjs:
Fixes #2759.

This PR adds a not-found view that stylistically matches the look of the message on the Activity page before a Namespace is selected.

Screenshot from 2019-03-22 09-24-25

Here are some screencasts of the behavior:
- This shows the primary behavior running locally on the Webpack Dev Server. If a user explicitly navigates to an unknown route, the view is shown indicating that the page is invalid.
  - https://screencast.googleplex.com/cast/NDc4OTAzMjM0NTE0MTI0OHxlOWI4YzdkMy0zMA
- This shows the behavior when deployed in a Kubeflow cluster on GKE. The behavior is the same.
  - https://screencast.googleplex.com/cast/NjAwMjAwMzc4ODQzMTM2MHxmZDI5OWFjNC1jYw

It should no longer be possible to land on the dashboard unless using the base URL of the cluster.

With respect to the iframe-inception behavior, I believe that should only be possible when running the Webpack Dev Server via `npm run dev` since it rewrites all all incoming requests to the entry point (index.html). I experimented with adding additional entries to the `proxy` object to see if I could proxy those paths to a working GKE cluster but it doesn't work properly due to same-origin issues. That might be different if you were working with a cluster on your local machine. In any case, I don't think that's a huge issue at the moment since it only affects development and can't occur in cluster.

/assign @avdaredevil 
/area front-end
/area centraldashboard

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2778)

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -91,8 +91,8 @@ export class MainPage extends PolymerElement {
             buildVersion: {type: String, value: BUILD_VERSION},
             dashVersion: {type: String, value: VERSION},
             inIframe: {type: Boolean, value: false, readOnly: true},
-            notFoundInIframe: {type: Boolean, value: false},
-            _devMode: {type: Boolean, value: DEVMODE},
+            hideTabs: {type: Boolean, value: false, readOnly: true},
+            notFoundInIframe: {type: Boolean, value: false, readOnly: true},
         };
     }
 
@@ -157,7 +157,7 @@ export class MainPage extends PolymerElement {
      */
     _routePageChanged(newPage) {
         let isIframe = false;
-        this.notFoundInIframe = false;
+        this._setNotFoundInIframe(false);
         switch (newPage) {
         case 'activity':
             this.sidebarItemIndex = 0;
@@ -176,9 +176,11 @@ export class MainPage extends PolymerElement {
             this.page = 'not_found';
             // Handles case when an iframed page requests an invalid route
             if (window.location !== window.parent.location) {
-                this.notFoundInIframe = true;
+                this._setNotFoundInIframe(true);
             }
         }
+        this._setHideTabs(!(this.page === 'activity' ||
+            this.page === 'dashboard'));
         this._setInIframe(isIframe);
         // If iframe <-> [non-frame OR other iframe]
         if (isIframe !== this.inIframe || isIframe) {
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -157,11 +157,13 @@ export class MainPage extends PolymerElement {
      */
     _routePageChanged(newPage) {
         let isIframe = false;
-        this._setNotFoundInIframe(false);
+        let notFoundInIframe = false;
+        let hideTabs = true;
         switch (newPage) {
         case 'activity':
             this.sidebarItemIndex = 0;
             this.page = 'activity';
+            hideTabs = false;
             break;
         case '_': // iframe case
             this._setIframeFromRoute(this.subRouteData.path);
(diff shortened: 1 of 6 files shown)


avdaredevil:
I think this will still lead to the inception looking issue. Since iframe -> iframe -> 404 will look really weird.

prodonjs:
I guess I'm not clear how you could obtain the inception behavior when deployed in cluster with the other routes pointing to real endpoints. My apologies for the screencasts cutting off the URL panel.

Can you provide me with the steps that reproduces the iframe behavior on this PR in-cluster?

https://kubeflow.endpoints.prodonjs-kubeflow-dev.cloud.goog has this version deployed.

I'm not too concerned about the behavior when running local development since those iframed pages won't be available.

prodonjs:
I suppose if one of the links inside one of the other apps that are iframed goes to a page that doesn't exist it could still occur. That's going to be a bit more complicated of an issue to resolve.

avdaredevil:
Yeah exactly, the issue is that some components may not necessarily come up, or may error out / not be accessible. A subsequent 404 will just re-render index.html. We could have the index.html be context-aware, and rather than handling it server side (_preferable_), we could possibly make the polymerApp at `main-app` level convert itself to a 404 page.

prodonjs:
I think this latest change addresses the issue of a route being loaded within the iframe that doesn't correspond to a valid page and is not already being handled by client-side routing within that application.

Screen recording when running locally is at: https://screencast.googleplex.com/cast/NjI0NjY1MjAzNzgyNDUxMnw4NjM3Y2RlMi01Zg

Essentially, when the main page is loaded inside of an iframe, we detect it and then force the sidebar and navbar to be hidden. As this should hopefully be an unlikely scenario, I think this should suffice.

prodonjs:
Here's the behavior after the latest commit:
- Direct navigation to a not found URL
Screenshot from 2019-03-25 16-01-39
- Navigation to a bad path inside an iframe
Screenshot from 2019-03-25 16-02-09

prodonjs:
@avdaredevil LMK if you have a chance to take a look at this or if you think it needs any other changes.

avdaredevil:
Yes I did, wasn't sure if it was final
/lgtm
/approve
/hold

components/centraldashboard/public/components/main-page.js
```diff
@@ -90,7 +91,8 @@ export class MainPage extends PolymerElement {
             buildVersion: {type: String, value: BUILD_VERSION},
             dashVersion: {type: String, value: VERSION},
             inIframe: {type: Boolean, value: false, readOnly: true},
-            _devMode: {type: Boolean, value: DEVMODE},
```
avdaredevil: Why are we removing this?
I assume we will be hiding off functionality for new releases as well, if they're longer running, or require work with other components (eg. pipelines, notebooks, etc)

components/centraldashboard/public/components/main-page.js
```diff
@@ -155,6 +157,7 @@ export class MainPage extends PolymerElement {
      */
     _routePageChanged(newPage) {
         let isIframe = false;
+        this._setNotFoundInIframe(false);
```
avdaredevil: Try to only propagate the value once, ie. track the value in a variable, and set it once.

Since polymer will instantaneously fire all property handlers as soon as you set the value. Does that make sense?

components/centraldashboard/public/components/main-page.js
avdaredevil: Why are we removing this?
I assume we will be hiding off functionality for new releases as well, if they're longer running, or require work with other components (eg. pipelines, notebooks, etc)

components/centraldashboard/public/components/main-page.js
prodonjs: Good point I hadn't really thought much about the effects of setting the property more than once. For some reason I had it in my mind that any change detection was deferred and push to the event loop but I guess I should have read the docs more closely where it states that [Data flow is synchronous](https://polymer-library.polymer-project.org/3.0/docs/devguide/data-system#data-flow-is-synchronous).

PR #2793, opened 2019-03-24
Add missing alsologtostderr flag to mpi-operator args

terrytangyuan:
mpi-operator doesn't log to std by default, this `kubectl logs` doesn't retrieve the logs. 

See `alsologtostderr` flag is also set in [mpi-operator.yaml](https://github.com/kubeflow/mpi-operator/blob/master/deploy/3-mpi-operator.yaml#L23).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2793)


terrytangyuan:
It seems like there are some flaky Argo tests in presubmit. This is passing now.

PR #2795, opened 2019-03-24
E2E kfctl go test should verify that Kubeflow is deployed.

jlewi:
* Adds the component tests but not the spark operator tests.

* Related to #2610 E2E test for go binary.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2795)

### DIFF
--- a/testing/kfctl/kf_is_ready_test.py
+++ b/testing/kfctl/kf_is_ready_test.py
@@ -1,0 +1,66 @@
+import datetime
+import logging
+import os
+import subprocess
+import tempfile
+import uuid
+from retrying import retry
+
+import pytest
+
+from kubeflow.testing import util
+from testing import deploy_utils
+
+def test_kf_is_ready(namespace):
+  """Test that Kubeflow was successfully deployed.
+
+  Args:
+    namespace: The namespace Kubeflow is deployed to.
+  """
+
+  logging.info("Using namespace %s", namespace)
+
+  # Need to activate account for scopes.
+  if os.getenv("GOOGLE_APPLICATION_CREDENTIALS"):
+    util.run(["gcloud", "auth", "activate-service-account",
+              "--key-file=" + os.environ["GOOGLE_APPLICATION_CREDENTIALS"]])
+
+
+
+  api_client = deploy_utils.create_k8s_client()
+
+  util.load_kube_config()
+
+  # Verify that components are actually deployed.
+  # TODO(jlewi): We need to parameterize this list based on whether
+  # we are using IAP or basic auth.
+  deployment_names = [
+      "argo-ui",
+      "centraldashboard",
+      "cert-manager",
+      "cloud-endpoints-controller",
+      "jupyter-web-app",
+      "ml-pipeline",
+      "ml-pipeline-scheduledworkflow",
+      "ml-pipeline-ui",
+      "notebooks-controller",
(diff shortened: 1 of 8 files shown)


jlewi:
Test passed before rebasing
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/2795/kubeflow-presubmit/6687/

So this should be ready to go.

jlewi:
Most recent failure was due to running out of CPU quota.

testing/workflows/components/kfctl_go_test.jsonnet
```diff
@@ -213,6 +213,27 @@ local dagTemplates = [
...
+      [        
+        "pytest",
+        "kf_is_ready_test.py",
+        // I think -s mean stdout/stderr will print out to aid in debugging.
+        // Failures still appear to be captured and stored in the junit file.
+        "-s",
+        // Increase the log level so that info level log statements show up.
+        "--log-cli-level=info",
+        // Test timeout in seconds.
+        "--timeout=500",
```
gabrielwen: if it's waiting for the deployment to be ready, 500 seconds seems to be too constraint?

testing/kfctl/kf_is_ready_test.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+  api_client = deploy_utils.create_k8s_client()
+
+  util.load_kube_config()
+
+  # Verify that components are actually deployed.
+  # TODO(jlewi): We need to parameterize this list based on whether
+  # we are using IAP or basic auth.
+  deployment_names = [
```
gabrielwen: maybe read config files from bootstrap/config/... ?

testing/workflows/components/kfctl_go_test.jsonnet
jlewi: Intersting. It doesn't seem to work like that so I guess I can remove it.
If we look at the runs the tests are definitely running longer than 5 minutes.
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-postsubmit-kfctl-go-4485a65-1297-efc5?tab=workflow

testing/kfctl/kf_is_ready_test.py
jlewi: I think it might be better to be explicit. This way its clear what the expectations are (e.g. which servicecs could be running).

If we parse the config and get a list of components dynamically then its not clear if we are setting expectations correctly. Also, if we do it that way and someone accidentally deletes some component then our E2E test wouldn't catch that.

jlewi:
/approve

Fingers crossed tests pass.

jlewi:
/test all

Latest failure was on delete; looks like a transitory issue calling Deployment manager delete.

```
Error: couldn't delete KfApp: coordinator Delete failed for gcp: Gcp.Delete is failed for kubeflow-ci-deployment/k
fctl-d64a: Deleting kfctl-d64a error: Get https://www.googleapis.com/deploymentmanager/v2/projects/kubeflow-ci-deployment/global/operations/operation-1553
648831540-585090eb5ac11-86b3ff73-d1b65ebc?alt=json&prettyPrint=false: read tcp 10.36.54.88:58554->74.125.141.95:443: read: connection reset by peer
```

PR #2798, opened 2019-03-25
Stop setting JupyterHub parameters and pull config from PR branch.

### LINKED ISSUE #2796: kfctl generate fails "jupyter" is not a component or a module
I started seeing the following failures in the presubmit to #2795.

My PR is using: 4addac44f235fa89145e586be5dcb7fbd94fe261

```
util.py                     71 INFO     time="2019-03-24T21:53:18Z" level=info msg="Writing component at '/mnt/test-data-volume/jlewi-kfctl-test-2795-0324
-143944/kfctl-3944/ks_app/components/tf-job-operator.jsonnet'" filename="component/create.go:92"
util.py                     71 INFO     Error: couldn't generate KfApp: coordinator Generate failed for ksonnet: Failed to set param jupyter platform gke:
 could not find component: extracting module and component names from jupyter: "jupyter" is not a component or a module
```

My suspicion is this is related to #2764 which removed the jupyter component from one of the config files.

I suspect there are some versioning issues here; e.g. when we call `kfctl generate` in the test we aren't pulling the config from the release branch so we aren't testing config changes in the presubmit.

/cc @zabbasi @gabrielwen @kkasravi

jlewi:
* Fix #2796; stop installing the jupyter component (jupyterhub) and
  trying to set parameters on that component.

* In presubmits we need to pull the config file listing the ksonnet packages,
  and components to configure from the PR branch and not master.

* To support local running of the test, only activate the service account if
GOOGLE_APPLICATION_CREDENTIALS is set.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2798)

### DIFF
--- a/bootstrap/config/kfctl_basic_auth.yaml
+++ b/bootstrap/config/kfctl_basic_auth.yaml
@@ -63,9 +63,6 @@ componentParams:
   cloud-endpoints:
     - name: secretName
       value: admin-gcp-sa
-  jupyter:
-    - name: platform
-      value: gke
   ambassador:
     - name: ambassadorServiceType
       value: NodePort
--- a/bootstrap/config/kfctl_default.yaml
+++ b/bootstrap/config/kfctl_default.yaml
@@ -39,9 +39,6 @@ componentParams:
   application:
     - name: components
       value: <list-of-components>
-  jupyter:
-    - name: platform
-      value: none
   ambassador:
     - name: ambassadorServiceType
       value: NodePort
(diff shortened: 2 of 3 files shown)


jlewi:
@gabrielwen I think kfctl_iap.yaml is already up to date.
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_iap.yaml

I think @zabbasi might have already updated it; it doesn't look like we are still setting parameters on the jupyter component.

PR #2801, opened 2019-03-25
kfctl interface in gcp.go/ksonnet.go for deploy app

kunmingg:
* remove gcloud dependency from gcp.go,
* remove kubectl dependency from ksonnet.go,
* parameterize auth method for gcp.go and ksonnet.go.

so kfctl and deploy app can use same code with their own auth methdo.

New interface example:
Generate() and Apply() need to be called by deploy app, it need to satisfy:
* no dependency on ENV var
* no dependency on local config like gcloud or kube config
* need to be thread safe
* only do disk read/write within it's own app dir.

Now for kfctl we prepare auth config in GetKfApp()

Have to remove application to avoid concurrent label edit.
Will hard code label for v0.5 and use application component once kustomize.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2801)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -85,6 +85,12 @@ type Gcp struct {
 	tokenSource oauth2.TokenSource
 	// When isCLI is false, following code need to be multi-thread safe, and can not access local configs or gcloud cli
 	isCLI bool
+	// requried when choose basic-auth
+	username        string
+	encodedPassword string
+	// requried when choose iap
+	oauthId     string
+	oauthSecret string
 }
 
 // GetKfApp returns the gcp kfapp. It's called by coordinator.GetKfApp
@@ -98,7 +104,7 @@ func GetKfApp(kfdef *kfdefs.KfDef) (kftypes.KfApp, error) {
 	ts, err := google.DefaultTokenSource(ctx, iam.CloudPlatformScope)
 	if err != nil {
 		return nil, &kfapis.KfError{
-			Code: int(kfapis.INVALID_ARGUMENT),
+			Code:    int(kfapis.INVALID_ARGUMENT),
 			Message: fmt.Sprintf("Get token error: %v", err),
 		}
 	}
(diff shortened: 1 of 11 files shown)


bootstrap/pkg/utils/gcputils.go
```diff
@@ -71,64 +59,3 @@ func BuildConfigFromClusterInfo(ctx context.Context, cluster *containerpb.Cluste
...
-	log.Infof("Getting rest.Config with GCP for %v/%v/%v", project, loc, cluster)
-	clusterpb, err := GetClusterInfo(ctx, project, loc, cluster)
-	if err != nil {
-		return nil, err
-	}
-	return BuildConfigFromClusterInfo(ctx, clusterpb)
-}
-
-// Create a config that serves as kubeconfig.
-func CreateKubeconfig(ctx context.Context, project string, loc string, cluster string,
```
gabrielwen: please don't delete this.  I need it to fix #2586

gabrielwen:
could you try to bring up some deployments and make sure both IAP/basic auth runs as expected?

jlewi:
#2795 Adds some logic to the e2e test to verify that Kubeflow is actually deployed successfully. Its currently only doing basic_auth. That should give us confidence though that we are creating Kubeflow correctly though.

bootstrap/pkg/utils/gcputils.go
kunmingg: Done

kunmingg:
@jlewi 
PR description updated.

jlewi:
@Kunming is this ready for another look after our discussions this afternoon? The PR description still mentions ApplyInternal. Are you planning on keeping that?

kunmingg:
@jlewi 
ready for review, PR description updated.

kunmingg:
@jlewi 
Updated.
Saw label conflicts in earlier presubmit test:
```
Component basic-auth apply failed; Error: handle object: patching object from cluster: merging object with existing state: Deployment.apps \"basic-auth\" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{\"app\":\"basic-auth\"}: `selector` does not match template `labels`
```
Seems caused by concurrent label edit (ksonnet apply and application component)

kunmingg:
@jlewi @gabrielwen 
Updated

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -83,14 +81,47 @@ const IstioNamespace = "istio-system"
...
+	}
 	_gcp := &Gcp{
-		KfDef: *kfdef,
+		KfDef:       *kfdef,
+		client:      client,
+		tokenSource: ts,
+		isCLI:       true,
+	}
+	if _gcp.Spec.Email == "" {
+		account, err := GetAccount()
```
gabrielwen: `isCLI` flag to protect this call?

gabrielwen:
one last minor comment, otherwise lgtm.

PR #2802, opened 2019-03-25
Add named context to KUBECONFIG

### LINKED ISSUE #2604: [GCP] kfctl create named context and set namespace
The kfctl go binary when running on GCP should do the equivalent of the following
https://github.com/kubeflow/kubeflow/blob/76230712d80805251f2ac54d66b0f02838de1c79/scripts/gke/util.sh#L156

Specifically it should create a context named after the KfApp and set the namespace to be the namespace where kubeflow is deployed.

gabrielwen:
Fixes #2604

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2802)


bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -430,6 +537,9 @@ func (gcp *Gcp) updateDM(resources kftypes.ResourceEnum) error {
 	if err := cred_cmd.Run(); err != nil {
 		return fmt.Errorf("Error when running gcloud container clusters get-credentials: %v", err)
 	}
+	if _, err = os.Stat(kftypes.KubeConfigPath()); !os.IsNotExist(err) {
+		gcp.AddNamedContext()
+	}
```
kunmingg: Can we call it at the end of apply()?
Since this function is ```updateDM()```?

gabrielwen: in `kfctl.sh` we do it in `updateDM` tho: https://github.com/kubeflow/kubeflow/blob/80baee129657f7aa4e93e3adbb820e0629ddbbfd/scripts/gke/util.sh#L158

my guess is because we want to create KUBECONFIG entries immediately after the deployment is created.  let me know if you think moving to `Apply()` is more appropriate, tho.

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -337,6 +339,112 @@ func (gcp *Gcp) ConfigK8s() error {
 	return nil
 }
 
+// Add a conveniently named context to KUBECONFIG.
+func (gcp *Gcp) AddNamedContext() error {
```
jlewi: Could we add a unittest to verify the config file is modified correctly?

jlewi: The integration test is good. 

It would be good to have a unittest as well; but we can leave that for a follow on PR since I think we have higher priority items for 0.5.

For a unittest we might want to refactor the code so that we have a simple method to modify the file. Then we can write a unittest in which we list test cases with input and expected output and verify that the code does the correct thing.

jlewi:
/lgtm
/hold
So @kunmingg has a chance to review.

jlewi:
Not sure why the test failed; Argo's not showing the logs for the failed kfctl.sh apply step. But if you get the workflow YAMl and identify the pod for the failed step you can see why it failed.

gabrielwen:
seems to be some kind of resource issue?  will run presubmit later.

bootstrap/pkg/kfapp/gcp/gcp.go
gabrielwen: sounds good, adding a new issue for tracking: #2817

bootstrap/pkg/kfapp/gcp/gcp.go
kunmingg: Calling it at the end of apply() won't be conflict with "create KUBECONFIG entries immediately after the deployment is created".
Would be nice to scope down updateDM() to deal with deployment manager logics.

gabrielwen: Done.

jlewi:
/lgtm
/approve

I'll let @kunmingg remove the hold or @gabrielwen you can remove it if you don't think you need @kunmingg 's sign off

kunmingg:
/hold cancel
Thanks!

PR #2807, opened 2019-03-25
Parameterize containerRuntimeExecutor on argo template

Tomcli:
We want to add the `containerRuntimeExecutor` parameter to the Argo template because we want to experiment and develop with different Argo container runtime. The parameter (linked below) also supported in Argo 2.2.0, so it shouldn't break any backward compatibility.

https://github.com/argoproj/argo/blob/master/docs/workflow-controller-configmap.yaml#L79

Thank you in advance and we are welcome any feedback to improve this.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2807)


Tomcli:
@IronPan Thank you very much for the review. I updated the argo test with my corresponding changes and now it passed the kubeflow-presubmit. Let me know if you have any question, thanks.

Tomcli:
Hi @IronPan, want to check back on this. I fixed the recent conflicts with the `master` branch. Does this look good to you? Thanks.

animeshsingh:
@IronPan @vicaire @Ark-kun coming to this - the reason its needed in our (IBM's case) is because we are no longer supporting Docker on IBM Cloud Kubernetes Service (IKS), and this gives us an option to experiment with PNS executor. We have been working with Argo folks to get the PNS executor in, and right now seems the safer choice
https://github.com/argoproj/argo/issues/1256

vicaire:
@IronPan, could you please approve so that this can be checked in?

PR #2810, opened 2019-03-26
fixes 'kfctl delete fails when no platform or minikube is specified'

### LINKED ISSUE #2809: kfctl delete fails when no platform or minikube is specified
Run the following

kfctl init kubeflow
cd kubeflow
kfctl generate all
kfctl apply all

Then after a few minutes run
kfctl delete all

kkasravi:
fixes #2809

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2810)

### DIFF
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -461,7 +461,7 @@ func (kfapp *coordinator) Delete(resources kftypes.ResourceEnum) error {
 		// deleting the PLATFORM means deleting the cluster. We remove k8s first in order free up any cloud vendor
 		// resources. Deleting k8 resources is a best effort and partial delete or failure should not
 		// prevent PLATFORM (cluster) deletion
-		_ := k8s()
+		_ = k8s()
 		return platform()
 	case kftypes.K8S:
 		return k8s()
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -453,11 +453,15 @@ func (kfapp *coordinator) Delete(resources kftypes.ResourceEnum) error {
 
 	switch resources {
 	case kftypes.ALL:
-		fallthrough
-	case kftypes.PLATFORM:
 		if err := k8s(); err != nil {
-			return err
+			return fmt.Errorf("error while deleting k8 resources, aborting deleting the platform. Error %v", err)
 		}
+		return platform()
+	case kftypes.PLATFORM:
+		// deleting the PLATFORM means deleting the cluster. We remove k8s first in order free up any cloud vendor
+		// resources. Deleting k8 resources is a best effort and partial delete or failure should not
+		// prevent PLATFORM (cluster) deletion
+		_ := k8s()
 		return platform()
 	case kftypes.K8S:
 		return k8s()
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"github.com/ghodss/yaml"
 	gogetter "github.com/hashicorp/go-getter"
+	kfapis "github.com/kubeflow/kubeflow/bootstrap/pkg/apis"
 	kftypes "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps"
 	kfdefs "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps/kfdef/v1alpha1"
 	"github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/gcp"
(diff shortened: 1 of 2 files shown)


bootstrap/pkg/kfapp/ksonnet/ksonnet.go
```diff
@@ -288,6 +288,50 @@ func (ksApp *ksApp) deleteGlobalResources(config *rest.Config) error {
 }
 
 func (ksApp *ksApp) Delete(resources kftypes.ResourceEnum) error {
+	if ksApp.Spec.Platform != kftypes.GCP {
```
jlewi: Why aren't you invoking this code on GCP? If you are deleting K8s resources I would expect that to also work on GCP.

On GCP I think we want to delete K8s resources before deleting the cluster because this triggers some cleanup.

e.g. deleting ingress should allow the GCP resources to be cleaned up; whereas if we don't delete the K8s resource and just delete the cluster they won't get cleaned up.

gabrielwen: what kind of GCP resources?  currently we skip deleting resources for the following: https://github.com/kubeflow/kubeflow/blob/master/testing/test_deploy_app.py#L443-L450

currently these codes won't work if user doesn't point KUBECONFIG to the correct context.

kkasravi: Gabriel is right, i think but i’ll verify. I’ll also add an e2e test

jlewi: It looks like this code is deleting all the K8s resources that we created as part of deploying Kubeflow.

So why wouldn't we also delete those resources in the case of GCP platform?

At worst its an unnecessary operation because we will end up deleting the cluster.

However, in some cases if we don't delete the K8s resource and trigger corresponding deletion handlers in K8s we don't properly clean up all resources.

The logic here:
https://github.com/kubeflow/kubeflow/blob/master/testing/test_deploy_app.py#L443-L450

Is indicating that certain resources like http-maps aren't being properly deleted.

We never create those resources explicitly. Those resources are created by the ingress that we create.

When we delete Kubeflow we shouldn't have to explicitly delete those resources (url-map, etc...) The fact that we do is a work around for the fact that those resources aren't being adequately garbage collected when the owning resource (GKE cluster is deleted).

I think if we explicitly delete the K8s resources as opposed to just deleting that cluster helps avoid these types of resource leaks.

Also from a design perspective, its better if we minimize the number of "if ${platform} then ...".

kkasravi: @jlewi +1

I also realized that `kfctl delete k8s` would not delete the cluster, so k8 cleanup needs to occur.

>On GCP I think we want to delete K8s resources before deleting the cluster because this triggers some cleanup.

I'll invert the order of operations - (package managers delete -> platforms delete)

kkasravi: @jlewi 

For `kfctl delete platform` - it seems like this should behave the same as `kfctl delete all`.

jlewi: I'm on the fence; what do you recommend?

What happens if there is a bug that prevents 

`kfctl delete k8s` from succeeding?

In that case we might want to fall back to just running `kfctl delete platform`.

So maybe the behavior could be

`kfctl delete all` - Invoke k8s delete and abort on error

`kfctl delete platform` - try to invoke k8s delete but continue to platform if k8s delete throws an error.

Thoughts?

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -453,11 +453,11 @@ func (kfapp *coordinator) Delete(resources kftypes.ResourceEnum) error {
 
 	switch resources {
 	case kftypes.ALL:
-		if err := platform(); err != nil {
+		fallthrough
+	case kftypes.PLATFORM:
+		if err := k8s(); err != nil {
```
jlewi: nit: Add a comment explaining why we want to call K8s as part of platform deletion.

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -453,11 +453,15 @@ func (kfapp *coordinator) Delete(resources kftypes.ResourceEnum) error {
 
 	switch resources {
 	case kftypes.ALL:
-		if err := platform(); err != nil {
-			return err
+		if err := k8s(); err != nil {
+			return fmt.Errorf("error while deleting k8 resources, aborting deleting the platform. Error %v", err)
```
gabrielwen: could you use new `KfError` interface when adding new error message?

example:
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/gcp/gcp.go#L351

kkasravi:
Sorry - doing it now

bootstrap/pkg/kfapp/coordinator/coordinator.go
kkasravi: added

PR #2813, opened 2019-03-26
fixes 'kfctl apply fails to create k8s resources; tries to deploy to 127.0.0.1'

### LINKED ISSUE #2791: kfctl apply fails to create k8s resources; tries to deploy to 127.0.0.1
When I first run `kfctl apply -v all` it fails with the following error

```
ERRO[0054] (Will retry) Component metacontroller apply failed; Error: Attempting to deploy to environment 'default' at '127.0.0.1', but cannot locate a server at that address  filename="ksonnet/ksonnet.go:191"
```
If I rerun `kfctl apply -v all` it works.

So it looks like there is a problem getting the cluster to talk when we first run kfctl apply and call ksonnet apply.

I'm using version:
v0.4.0-rc.1-264-g20a80512

/cc @gabrielwen @kkasravi

kkasravi:
fixes #2791

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2813)

### DIFF
--- a/bootstrap/pkg/kfapp/ksonnet/ksonnet.go
+++ b/bootstrap/pkg/kfapp/ksonnet/ksonnet.go
@@ -437,10 +437,16 @@ func (ksApp *ksApp) envSet(envName string, host string) error {
 		actions.OptionAppRoot: ksApp.ksRoot(),
 		actions.OptionEnvName: ksApp.KsEnvName,
 		actions.OptionServer:  host,
+		actions.OptionOverride: true,
 	})
 	if err != nil {
 		return fmt.Errorf("There was a problem setting ksonnet env: %v", err)
 	}
+	loadApp, loadErr := app.Load(afero.NewOsFs(), ksApp.KApp.HTTPClient(), ksApp.ksRoot())
+	if loadErr != nil {
+		return fmt.Errorf("could not reload the ksonnet env: %v", err)
+	}
+	ksApp.KApp = loadApp
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


jlewi:
Can you remove the retry logic around` kfctl apply` in the E2E test here:
https://github.com/kubeflow/kubeflow/blob/4485a65079a1b445195cb7cc34035655919aaaea/testing/kfctl/kfctl_go_test.py#L79

Right now that retry logic is masking problems like this.

If there are legitimate retryable failures we can bad more specific retry logic that only retries when those specific errors occur.

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
```diff
@@ -437,10 +437,16 @@ func (ksApp *ksApp) envSet(envName string, host string) error {
 		actions.OptionAppRoot: ksApp.ksRoot(),
 		actions.OptionEnvName: ksApp.KsEnvName,
 		actions.OptionServer:  host,
+		actions.OptionOverride: true,
```
jlewi: What's this doing?

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
```diff
@@ -437,10 +437,16 @@ func (ksApp *ksApp) envSet(envName string, host string) error {
...
 	})
 	if err != nil {
 		return fmt.Errorf("There was a problem setting ksonnet env: %v", err)
 	}
+	loadApp, loadErr := app.Load(afero.NewOsFs(), ksApp.KApp.HTTPClient(), ksApp.ksRoot())
```
jlewi: I see. So I'm guessing RunEnvSet is just modifying the config files on disks and we need to explicitly load the app to pick up the changes.

jlewi:
Looks good. Only comment is to update the test so that we can verify its fixed and catch errors like this in the future.

jlewi:
Looks like a tf-serving flake
/lgtm
/approve

PR #2815, opened 2019-03-26
Use YAML modifications instead of regex replacement

### LINKED ISSUE #2515: kfctl, replace regex edits with yaml modifications of individual fields
kfctl should need to do regex substitution on config files, rather read the structure in via yaml, change the field and write the structure out

gabrielwen:
Fixes #2515

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2815)


gabrielwen:
@kunmingg @lluunn could you review this PR first?  presubmit seems to be flaky for the time being.

gabrielwen:
@kunmingg @lluunn : please take a look when you have chance, thanks!

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -583,16 +583,209 @@ func setNameVal(entries []configtypes.NameValue, name string, val string, requir
...
+		return &kfapis.KfError{
+			Code:    int(kfapis.INTERNAL_ERROR),
+			Message: fmt.Sprintf("Error when writing IAM bindings: %v", err),
+		}
+	}
+	return nil
+}
+
+// Replace placeholders and write to cluster-kubeflow.yaml
+func (gcp *Gcp) writeClusterConfig(src string, dest string) error {
```
lluunn: I think "write" is not the word, because we are copying and replacing.
But not sure what's the better option...

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -614,66 +807,22 @@ func (gcp *Gcp) generateDMConfigs() error {
...
+	if err := gcp.writeIamBindingsFile(from, to); err != nil {
+		return err
 	}
-	iamBindingsData, iamBindingsDataErr := ioutil.ReadFile(to) // just pass the file name
-	if iamBindingsDataErr != nil {
-		return fmt.Errorf("could not read %v Error %v", to, iamBindingsDataErr)
+
+	from = filepath.Join(sourceDir, CONFIG_FILE)
+	to = filepath.Join(gcpConfigDir, CONFIG_FILE)
+	if err := gcp.writeClusterConfig(from, to); err != nil {
```
lluunn: If we can't have a better function name, add a comment here that this is to replace the placeholder

bootstrap/pkg/kfapp/gcp/gcp.go
gabrielwen: we are actually reading and writing with replaced file content.

bootstrap/pkg/kfapp/gcp/gcp.go
gabrielwen: added.

gabrielwen:
/test kubeflow-presubmit
--

PR #2819, opened 2019-03-26
fixes 'kfctl downloading the kubeflow to a local cache should be idempotent'

### LINKED ISSUE #2818: kfctl downloading the kubeflow repo to a local cache should be idempotent
I've seen the download to cache fail half-way through. It `kfctl init` is rerun we shouldn't return an error if the cache exists, rather just replace it.

kkasravi:
fixes #2818

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2819)

### DIFF
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -77,7 +77,10 @@ func downloadToCache(platform string, appDir string, version string, useBasicAut
 	cacheDir := path.Join(appDir, kftypes.DefaultCacheDir)
 	// idempotency
 	if _, err := os.Stat(cacheDir); !os.IsNotExist(err) {
-		os.RemoveAll(cacheDir)
+		rmErr := os.RemoveAll(cacheDir)
+		if rmErr != nil {
+			return nil, fmt.Errorf("couldn't delete directory %v Error %v", cacheDir, rmErr)
+		}
 	}
 	cacheDirErr := os.Mkdir(cacheDir, os.ModePerm)
 	if cacheDirErr != nil {


bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -76,7 +76,10 @@ func downloadToCache(platform string, appDir string, version string, useBasicAut
 	cacheDir := path.Join(appDir, kftypes.DefaultCacheDir)
 	// idempotency
 	if _, err := os.Stat(cacheDir); !os.IsNotExist(err) {
-		os.RemoveAll(cacheDir)
+		rmErr := os.RemoveAll(cacheDir)
```
gabrielwen: is this gonna be used by click to deploy app?  if it is, need to be careful of thread safety.

@kunmingg

jlewi: What's the threading issue here? 

My expectation is each request in the deploy app uses a separate instance of KfApp and uses a different directory to copy down the configs to.

So even if we simultanteously handle two duplicate requests for the same app; both would be modifying a different directory.

gabrielwen: yea, using different directory would be fine.  just wanna make sure it's working that way.

jlewi:
/lgtm
/approve
/hold
@kunmingg Any concerns about use in deploy app?

bootstrap/pkg/kfapp/coordinator/coordinator.go
kunmingg: Shouldn't affect logic of deploy app. thanks

jlewi:
Most recent failures are kubeflow/testing#342

jlewi:
In the recent test it looked like delete failed.

/test all

PR #2823, opened 2019-03-27
Add aws platform option

Jeffwan:
1. Documentation will come to kubeflow/website soon.
2. This PR only includes shell scripts change to provision cluster. Go cmd is in progress. 
3. AWS use ALB ingress controller to manage external traffic. Traffic will route to istio first and then come into Ambassador. 
4. Integrate few aws services like Lustre for DL training and Cloudwatch as log backend.
4. Looks like Kubeflow will move to Istio for routing later and we install [istio](https://www.kubeflow.org/docs/components/istio/) in the progress of kubeflow provision.
5. Authentication and TLS will not be enabled by default. They needs some manual efforts from users now. This will be documented in kubeflow/website.
6. Using eksctl to provision EKS cluster.  This is added in `apply platform` phase

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2823)

### DIFF
--- a/deployment/aws/infra_configs/iam_alb_ingress_policy.json
+++ b/deployment/aws/infra_configs/iam_alb_ingress_policy.json
@@ -1,0 +1,103 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Effect": "Allow",
+        "Action": [
+          "acm:DescribeCertificate",
+          "acm:ListCertificates",
+          "acm:GetCertificate"
+        ],
+        "Resource": "*"
+      },
+      {
+        "Effect": "Allow",
+        "Action": [
+          "ec2:AuthorizeSecurityGroupIngress",
+          "ec2:CreateSecurityGroup",
+          "ec2:CreateTags",
+          "ec2:DeleteTags",
+          "ec2:DeleteSecurityGroup",
+          "ec2:DescribeInstances",
+          "ec2:DescribeInstanceStatus",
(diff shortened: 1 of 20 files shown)


kkasravi:
@Jeffwan - this is a great contribution.

There is a way to add aws to kfctl (golang) so it can be loaded dynamically - we may have 
a strategy along these lines if we want to keep kfctl binary size small.

Let me know if you have any questions on kfctl.

Thanks!

Jeffwan:
@kkasravi Thanks. I am looking at golang kfctl now and this would be next step for it. What's the current status now? For short time, are kfctl.sh and kfctl(golang) both supported at the same time and move to golang step by step?

kkasravi:
@Jeffwan plan is to release kfctl (golang) for 0.5 and drop kfctl.sh. 
You can continue to use kfctl.sh in 0.5 to test aws, but the goal would be to add an aws.go - eg 
- cmd/plugins/aws/aws.go
- pkg/kfapp/aws/aws.go

you can use dockerfordesktop as a template.

Jeffwan:
@kkasravi Our plan is to have kfctl.sh ready first and then move to kfctl (golang). Is there a way to merge this change back to 0.5? we can finish golang kfctl work for 0.6.

jlewi:
/ok-to-test
/lgtm
/approve

While this PR is large most of the files are AWS specific. I only paid attention to the shared bits (kfctl.sh).

PR #2838, opened 2019-03-27
delete endpoints

### LINKED ISSUE #2190: [gcp] kfctl delete should ensure CloudEndpoints service is deleted
When we delete Kubeflow, the Cloud Endpoints resource won't be deleted if we just delete the deployment.

kfctl should explicitly delete it.

Does cloud endpoints support labels? It would be nice if we could use labels to identify the KF deployment that owned a cloud endpoint so we could easily clean up endpoints.

Related to: kubeflow/testing#268

gabrielwen:
Fixes #2190 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2838)


bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -597,6 +598,65 @@ func deleteDeployment(deploymentmanagerService *deploymentmanager.Service, ctx c
 	return nil
 }
 
+// Delete endpoint service from resources.
+func (gcp *Gcp) deleteEndpoints(ctx context.Context) error {
+	client, clientErr := google.DefaultClient(ctx, servicemanagement.ServiceManagementScope)
```
kunmingg: Now you can get client from gcp directly.

gabrielwen: done.

PR #2853, opened 2019-03-28
add labels for notebook so that gcp credentials will be injected by webhook

### LINKED ISSUE #2770: update notebook controller so that notebook pod has proper annotation for admission webhook
see #2641

lluunn:
Fix #2770

/cc @jlewi @zabbasi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2853)

### DIFF
--- a/components/notebook-controller/pkg/controller/notebook/notebook_controller.go
+++ b/components/notebook-controller/pkg/controller/notebook/notebook_controller.go
@@ -286,10 +286,20 @@ func generateStatefulSet(instance *v1alpha1.Notebook) *appsv1.StatefulSet {
 	}
 
 	// Inject GCP credentials
-	if os.Getenv("GCP_CREDENTIALS") != "" {
-		labels := &ss.Spec.Template.ObjectMeta.Labels
-		(*labels)["gcp-cred-secret"] = "user-gcp-sa"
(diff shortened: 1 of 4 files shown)


components/notebook-controller/pkg/controller/notebook/notebook_controller.go
```diff
@@ -283,6 +284,14 @@ func generateStatefulSet(instance *v1alpha1.Notebook) *appsv1.StatefulSet {
 			},
 		},
 	}
+
+	// Inject GCP credentials
+	if os.Getenv("GCP_CREDENTIALS") != "" {
```
jlewi: How about generalizing this?

Take an environment variable that is a comma separated list of labels to apply to the pods.

lluunn: done

abhi-g:
@lluunn is this waiting for another review from @jlewi

lluunn:
yes, thanks

r2d4:
We also need the environment variable `GOOGLE_APPLICATION_CREDENTIALS` to be set and point to the mounted secret to have parity with the previous Kubeflow jupyter deployment.

PR #2869, opened 2019-03-29
Add volumn support and unit tests for mpi-job

### LINKED ISSUE #2858: Add volume support for mpi-job
1. mpi-job by default doesn't have volume support, this is good for synthetic data. Recently, I need to train model using real storage. I'd like to add volume options.

2. mpi-job doesn't have unit test coverage. mpi-job-custom and mpi-job-simple are are jsonnet files which are not easy to test. I'd like to make some adjustment and make them easy to test.

I am working on a PR to address these issues. Prototype won't be changed and they just need to be refactored.

Jeffwan:
Resolve #2858 

1. Did some refactor and make mpi-job testable. Add unit tests for both mpi-job and mpi-operator
2. Add volume support for mpi-job.

I didn't change/add/remove any prototype names.  should be very safe.

This is still for v1alpha1. Notice mpi-job has v1alpha2 released and corresponding prototype should be separate tasks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2869)

### DIFF
--- a/kubeflow/mpi-job/mpi-job.libsonnet
+++ b/kubeflow/mpi-job/mpi-job.libsonnet
@@ -1,0 +1,87 @@
+
+{
+  local k = import "k.libsonnet",
+
+  parts(env, params):: {
+    local containerCommand = if params.command != "null" then
+      {
+        command: std.split(params.command, ","),
+      }
+      else {},
+
+    local containerArgs = if params.args != "null" then
+      {
+        args: std.split(params.args, ","),
+      }
+      else {},
+
+    local storageEnabled =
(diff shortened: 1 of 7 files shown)


terrytangyuan:
BTW, there's a typo in the title of this PR: "volumn" -> "volume".

Jeffwan:
> Thanks for this! v1alpha2 is still in the works so please stay tuned. One question is that whether we should switch to use something else other than ksonnet since it will not be maintained anymore.

Yes, I think we can switch to kustomize in 0.6. 

> BTW, there's a typo in the title of this PR: "volumn" -> "volume".
Thank you!

Issue #2601, opened 2019-03-01
[Golang profile controller] fix permission leak when namespace exists.


kunmingg:
When a namespace ```ns1``` created by admin1,
user1 can get permission by creating profile called ```ns1```.
We should identify this case and fix it.

lluunn:
cc

kunmingg:
Remove from 0.5.0 as we are not likely to ship it in this release.
Todo:
add to 0.6.0 once tag is created.

kunmingg:
@gabrielwen

PR #2877, opened 2019-03-30
remove profile from v0.5

kunmingg:
We should exclude profile from default set till permission leak:
https://github.com/kubeflow/kubeflow/issues/2601 is fixed.


abhi-g:
Looks like a valid test failure @kunmingg

kunmingg:
@abhi-g 
Thanks! Updated

Issue #2383, opened 2019-02-04
[kfctl go] Need a minimal release process


jlewi:
We will need a minimal release process for the kfctl go binary.

I think we can close this issue when we have instructions for how to build and release the kfctl go binary as part of a release.

We will need this for 0.5 since we want to include kfctl.

jlewi:
I think a minimal release process would be 

1. E2E workflow to build and test go binary
2. E2E test should copy the binary to GCS
    * Presubmits should use a bucket in kubeflow-ci
    * Postsubmits should use a public bucket (perhaps in project gcr.io/kubeflow-images-public)

gabrielwen:
1. Seems like we have E2E test ready.
1. What is the connection between presubmit and kubeflow-ci bucket?

jlewi:
It looks like Darwin requires its own build.

jlewi:
We should modify the makefile to determine the architecture and set the name in the file appropriately.
This is what we do in minikube
https://github.com/kubeflow/kubeflow/blob/3fa97bc61e7b4b1eeb0062a8cbc98920e9da43d5/scripts/setup-minikube.sh#L14

But I don't see how we are inferring its OSX.

PR #2894, opened 2019-04-01
Create a minimal release process for kfctl.

jlewi:
* Instructions for building binaries and uploading to GitHub release page.
* We need to figure out if/how we cross-compile for linux.
  * Cross compiling seems to be unnecessary
  * The go binaries built on linux appear to work on macosx.

Fix: #2383

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2894)


gabrielwen:
for my own learning: where will the binary go?

kkasravi:
> for my own learning: where will the binary go?
/usr/local/bin. In the future brew will do this as well

jlewi:
@gabrielwen @kkasravi Should we tar.gz the binary?

gabrielwen:
my last compiled binary is the size of 125MB, so yes, I think we should compress it.

PR #2898, opened 2019-04-01
Cherry-pick #2823 to v0.5 - Add aws platform option

Jeffwan:
* Add aws platform option

* Move ingress and fsx driver to ksonnet libs

* Add istio with ingress support

* Disable Jwt token checking by default

* Add cluster provision and destroy scripts

* Correct eksctl created role name

* Clean up scripts and fix errors

* Add unit tests from aws packages

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2898)


jlewi:
It looks like the argo workflows completed successfully but it took longer than the prow timeout.
/test all

PR #2903, opened 2019-04-02
fix patch string for admission webhook

### LINKED ISSUE #2896: admission webhook overwrites NB_PREFIX env var
The admission webhook successfully inject the volume/env var, but it seems overwrites env var instead of appends to it.
```
env:
    - name: GOOGLE_APPLICATION_CREDENTIALS
      value: /secrets/gcp-service-account-credentials/user-gcp-sa.json
image: ...
```

So the notebook is not working properly 404:
Screen Shot 2019-04-01 at 15 45 15

lluunn:
Fix #2896

Reference: https://github.com/istio/istio/blob/master/pilot/pkg/kube/inject/webhook.go#L377

Tested the notebook is accessible, and the volume/env are there by webhook

/cc @jlewi @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2903)

### DIFF
--- a/components/gcp-admission-webhook/main.go
+++ b/components/gcp-admission-webhook/main.go
@@ -49,9 +49,15 @@ var codecs = serializer.NewCodecFactory(scheme)
 // patchString patches: Add volume, volumeMount, and environment variable
 const patchString = `[
 {"op":"add","path":"/spec/volumes","value":{"name":"gcp-credentials","secret":{"secretName": "%s"}}},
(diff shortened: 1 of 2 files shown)


jlewi:
/lgtm
/approve
Do we have an issue adding unittests for this?

jlewi:
It looks like the generate pods was never started. Pod events shows "
Error: context deadline exceeded"
and there are no pod logs.

Issue #2845, opened 2019-03-28
Jupyter web app should respect namespace in central dashboard


jlewi:
Screenshot
![jupyternamespace](https://user-images.githubusercontent.com/777219/55130237-570a6300-50d7-11e9-9d12-e85acfc37f05.png)

The central dashboard is displaying the jupyter web app in an iframe.

The central dashboard has a namespace selector. This namespace selector is not wired up to the
jupyter web app.

/cc @avdaredevil @kimwnasptd @vkoukis 

@avdaredevil this might be a good opportunity to figure out how the namespace selector should interact with individual web apps.

vkoukis:
@jlewi we could get this in for 0.5, but we need a precise description of how the namespace selector works.

Will it be functional in 0.5? I guess the question is, who is responsible for listing K8s namespaces, and how can iframe-d apps access the contents of this selector?

zabbasi:
Similarly tf-operator and Katib dashboard's default namespace should respect namespace in central dashboard. @richardsliu @avdaredevil

kimwnasptd:
Also as a side note, in our discussion here #2521 people expressed the need to _not_ show k8s concepts such as Namespaces to the end user.

One option that was proposed was to use the term Project instead of Namespace. Should we also take this into account for the dashboard?

jlewi:
> One option that was proposed was to use the term Project instead of Namespace. Should we also take this into account for the dashboard?

One of the principles we try to follow in Kubeflow is not introducing new concepts, terms, patterns when they are already such things in K8s.

So w.r.t to namespace vs. project. If we introduce project, I think we should be creating something that is fundamentally different from a namespace. If a project is 1:1 mapping to namespace then I think we should just call it namespace.

> Will it be functional in 0.5? I guess the question is, who is responsible for listing K8s namespaces, and how can iframe-d apps access the contents of this selector?

I think this is the question for @avdaredevil .

prodonjs:
There are a couple of different options for making the iframed applications aware of a namespace selection from its parent.
1. The parent page injects JS into the iframe child that is invoked when the parent's namespace selector changes. This requires the parent to be aware of and manipulate the DOM state of its children. This is a bit brittle since the child app can change and any injected scripts would fail.
2. The iframe children accept **namespace** as part of their URL structure or as a query-string parameter. The selector is then made to synchronize with the URL in both directions. This allows the parent to set/change the iframe's src to reflect its namespace.

[...]

PR #2913, opened 2019-04-02
Central Dashboard - Add statefulness to generated links

prodonjs:
Ensures that all links on the Central Dashboard are namespace-aware in order to preserve state and allow it to be passed down to child components.

/area front-end
/area centraldashboard
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2913)

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -97,7 +97,6 @@ export class MainPage extends PolymerElement {
         };
     }
 
-
     /**
      * Array of strings describing multi-property observer methods and their
      * dependant properties
@@ -134,8 +133,9 @@ export class MainPage extends PolymerElement {
      * @param {MouseEvent} e
      */
     openInIframe(e) {
-        const url = new URL(e.currentTarget.href);
-        window.history.pushState({}, null, `_${url.pathname}`);
+        // e.currentTarget is an anchor element
+        const url = e.currentTarget.href.substr(e.currentTarget.origin.length);
+        window.history.pushState({}, null, `_${url}`);
         window.dispatchEvent(new CustomEvent('location-changed'));
         e.preventDefault();
     }
@@ -204,6 +204,20 @@ export class MainPage extends PolymerElement {
      */
     _isInsideOfIframe() {
         return window.location !== window.parent.location;
+    }
+
+    /**
+     * Builds and returns an href value preserving the existing query string and
+     * hash.
+     * @param {string} href
+     * @return {string}
+     */
+    _buildHref(href) {
+        const current = new URL(window.location.href);
+        const url = new URL(href, window.location.origin);
+        url.search = current.search;
+        url.hash = current.hash;
+        return url.href;
     }
 }
 
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -210,13 +210,16 @@ export class MainPage extends PolymerElement {
      * Builds and returns an href value preserving the existing query string and
      * hash.
      * @param {string} href
+     * @param {Object} queryParams
      * @return {string}
      */
-    _buildHref(href) {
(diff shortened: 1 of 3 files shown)


prodonjs:
addresses #2845

components/centraldashboard/public/components/main-page.js
```diff
@@ -205,6 +205,20 @@ export class MainPage extends PolymerElement {
...
+
+    /**
+     * Builds and returns an href value preserving the existing query string and
+     * hash.
+     * @param {string} href
+     * @return {string}
+     */
+    _buildHref(href) {
+        const current = new URL(window.location.href);
+        const url = new URL(href, window.location.origin);
```
avdaredevil: Why do we need this, isn't that implicit?
```suggestion
        const url = new URL(href);
```

components/centraldashboard/public/components/main-page.js
```diff
@@ -134,8 +133,9 @@ export class MainPage extends PolymerElement {
      * @param {MouseEvent} e
      */
     openInIframe(e) {
-        const url = new URL(e.currentTarget.href);
-        window.history.pushState({}, null, `_${url.pathname}`);
+        // e.currentTarget is an anchor element
+        const url = e.currentTarget.href.substr(e.currentTarget.origin.length);
```
avdaredevil: I think `slice` is the recommended js-onic way of substringing:
```suggestion
        const url = e.currentTarget.href.slice(e.currentTarget.origin.length);
```

components/centraldashboard/public/components/main-page.pug
```diff
@@ -32,10 +33,10 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
                 paper-tabs.bottom(selected='[[page]]', attr-for-selected='page',
                         hides, hidden$='[[hideTabs]]')
                     paper-tab(page='dashboard', link)
-                        a.link(tabindex='-1', href='/') Dashboard
+                        a.link(tabindex='-1', href$='[[_buildHref("/")]]') Dashboard
                     paper-tab(page='activity', link)
-                        a.link(tabindex='-1', href='/activity') Activity
-                aside#NamespaceSelector(hidden$='[[!equals(page, "activity")]]')
+                        a.link(tabindex='-1', href$='[[_buildHref("/activity")]]') Activity
```
avdaredevil: Won't these be static links?? All the `_buildHref`?

prodonjs: I guess that depends. I was thinking that namespace would behave similar to how projectId behaves in the GCP console. It's propagated along with every link even if the page being displayed doesn't use it to change its behavior.

components/centraldashboard/public/components/main-page.js
prodonjs: According to MDN, *If url is a relative URL, base is required, and will be used as the base URL. If url is an absolute URL, a given base will be ignored.* If we were getting the href from the anchor element's href property it would already be absolute, but since we are passing in a relative string we need to provide the base. The suggested change failed when I tried it.

components/centraldashboard/public/components/main-page.js
prodonjs: Changed.

prodonjs:
@avdaredevil PTAL when you have the chance. I think this addresses things correctly. If we find that other pages are often introducing links that need to be namespace aware, we should consider moving this to a mixin along with some of the other utility methods you created.

components/centraldashboard/public/components/main-page.js
avdaredevil: Ok, perhaps I misunderstood what this was doing. Thanks for following through!

avdaredevil:
I agree I can make a PR next week to refactor a few of these changes.
/hold cancel
/lgtm
/approve

Issue #2915, opened 2019-04-02
Notebook pods need to be attached to jupyter-notebook service account


abhi-g:
The notebook controller is creating Notebook pods with "default" service account. When launching training jobs through fairing on Kubeflow, I get an error that says that the account doesn't have permissions to create pods in the namespace.

ApiException: (403)
Reason: Forbidden
HTTP response headers: HTTPHeaderDict({'Audit-Id': '4224d1da-089b-401d-a661-d96a1533a1f0', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'Date': 'Tue, 02 Apr 2019 18:36:19 GMT', 'Content-Length': '255'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"pods is forbidden: User \"**system:serviceaccount:kubeflow:default**\" cannot create pods in the namespace \"kubeflow\"","reason":"Forbidden","details":{"kind":"pods"},"code":403}

Instead it should be attaching the "jupyter-notebook" service account to the notebook pods. [ At least until we implement the profiles for multi-user isolation]

PR #2917, opened 2019-04-02
Use jupyter service account for notebook CRD

r2d4:
We could also implement this in the notebook controller, since doing it in the notebook CRD itself would allow any service account to be mounted. However, we will most likely want that behavior in the future

cc @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2917)


r2d4:
Fixes https://github.com/kubeflow/kubeflow/issues/2915

PR #2919, opened 2019-04-02
Add jupyter notebook svc account and roles

r2d4:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2919)

### DIFF
--- a/kubeflow/jupyter/jupyter-web-app.libsonnet
+++ b/kubeflow/jupyter/jupyter-web-app.libsonnet
@@ -87,6 +87,97 @@
         apiGroup: "rbac.authorization.k8s.io",
       },
     },
+    
+    notebookRole:: {
+      apiVersion: "rbac.authorization.k8s.io/v1beta1",
+      kind: "Role",
+      metadata: {
+        name: "jupyter-notebook-role",
+        namespace: params.namespace,
+      },
+      rules: [
+        {
+          apiGroups: [
+            "",
+          ],
+          resources: [
+            "pods",
+            "pods/log",
+            "secrets",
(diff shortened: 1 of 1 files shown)


r2d4:
cc @lluunn

kubeflow/jupyter/jupyter-web-app.libsonnet
```diff
@@ -87,6 +87,97 @@
...
+      },
+      roleRef: {
+        apiGroup: "rbac.authorization.k8s.io",
+        kind: "Role",
+        name: "jupyter-role",
+      },
+      subjects: [
+        {
+          kind: "ServiceAccount",
+          name: "jupyter",
```
lluunn: jupyter-notebook?

lluunn:
Only the notebook in this namespace will work. 
Maybe the correct approach is let notebook controller to create the service account and attach to the notebook pod in the same namespace.

r2d4:
Can we currently spawn notebooks outside kubeflow namespace?

kubeflow/jupyter/jupyter-web-app.libsonnet
r2d4: fixed

lluunn:
cc @jlewi

lluunn:
I think for 0.5 we should restrict to kubeflow namespace.
The admission webhook will also mount the secret which only exists in kubeflow namespace.

Later in 0.6, we can provision these things (secret for gcp, service account) using profile.

So
/lgtm
/approve
/hold
In case someone disagrees

PR #2920, opened 2019-04-02
add retry around IAM APIs and ksonnet app init

kunmingg:
* We need to retry around IAM APIs when concurrent deploys happen to same project
* ks init might timeout (quite rare)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2920)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -590,14 +590,35 @@ func (gcp *Gcp) updateDM(resources kftypes.ResourceEnum) error {
 		}
 	}
 
-	policy, policyErr := utils.GetIamPolicy(gcp.Spec.Project, gcpClient)
-	if policyErr != nil {
-		return &kfapis.KfError{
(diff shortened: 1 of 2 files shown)


gabrielwen:
seems like you are retrying on every errors?

kunmingg:
Do you mean IAM one or ksonnet init one?

gabrielwen:
both - if you don't return `backoff.Permanent`, it'll keep retrying.

kunmingg:
@gabrielwen 
It's hard to predict if an error should be Permanent, relying on ```exp.MaxElapsedTime = time.Minute``` to set max retry period to be 1 minute.

@kkasravi 
If we can guarantee we don't need to undo ```ClearIamPolicy()``` under any case, then it might be OK? Try to make ```ClearIamPolicy()``` really conservative by unit tests.

kunmingg:
@kkasravi 
I feel it's really hard to make ```ClearIamPolicy+SetIamPolicy``` transactional.
If those 2 keep failing then basically indicate that we lose write access to IAM policy...

PR #2926, opened 2019-04-03
Enable iframed pages to communicate with Central Dashboard.

prodonjs:
addresses #2845 
child of #2913 

This WIP PR demonstrates a working solution for passing the active namespace to child pages. We bind to the **load** event of the iframe element and pass the active namespace (if selected) down through the message bus. We then pass the namespace to the active iframe anytime it is changed by observing a property change.

See https://screencast.googleplex.com/cast/NjUwNzg4MTA3NTU3MjczNnw5NDI1OGNjNi1jZg for a demo or pull and run it locally. Run `python -m SimpleHTTPServer` from the centraldashboard directory to run serve the test.html page included in this commit which receives the message from the parent page. 

/area front-end
/area centraldashboard
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2926)

### DIFF
--- a/components/centraldashboard/README.md
+++ b/components/centraldashboard/README.md
@@ -82,10 +82,14 @@ code with business logic.
 ---
 
 ## Style-Guide
-Kubeflow central dashboard is a visualization and networking platform that links fellow sub-apps within the Kubeflow Ecosystem together. As a result we have various web-apps that are iframed within the app. To keep this experience uniform, we have a style guide for all Kubeflow Integrators to follow / use.
+Kubeflow central dashboard is a visualization and networking platform that links
+fellow sub-apps within the Kubeflow Ecosystem together. As a result we have
+various web-apps that are iframed within the app. To keep this experience
+ uniform, we have a style guide for all Kubeflow Integrators to follow / use.
 
 ### Approach
-We store our CSS palette in [_public/kubeflow-palette.css_](public/kubeflow-palette.css). This contains variables such as
+The CSS palette is found at [_public/kubeflow-palette.css_](public/kubeflow-palette.css).
+When deployed to a cluster, it can be imported from `cluster-host/kubeflow-palette.css`
 
 Name | Value | Usage
 --- | --- | ---
@@ -124,12 +128,12 @@ Since the Central Dashboard wraps other Kubeflow components in an iframe, a
 standalone Javascript library is exposed to make it possible to communicate
 between the child pages and the dashboard.
 
-The library is available for import from `/lib.bundle.js` within the cluster.
-When imported in this manner, a `centraldashboard` module is created in the
-global scope. To establish a channel for communication between the iframed page
-and the Dashboard, use the following example which would disable a `<select>`
-element and assign the value from the Dashboard's Namespace selector when it
-changes:
+The library is available for import from `/dashboard_lib.bundle.js` within the
+cluster. When imported in this manner, a `centraldashboard` module is created in
+the global scope. To establish a channel for communication between the iframed
+page and the Dashboard, use the following example which would disable a
+`<select>` element and assign the value from the Dashboard's Namespace selector
+when it changes:
 
 ```
 window.addEventListener('DOMContentLoaded', function (event) {
--- a/components/centraldashboard/README.md
+++ b/components/centraldashboard/README.md
@@ -118,3 +118,34 @@ body > .Main-Toolbar {background-color: var(--primary-background-color)}
 .List > .item:not(:first-of-type) {border-top: 1px solid var(--border-color)}
 /* etc */
 ```
+
+## Client-Side Library
+Since the Central Dashboard wraps other Kubeflow components in an iframe, a
+standalone Javascript library is exposed to make it possible to communicate
+between the child pages and the dashboard.
+
+The library is available for import from `/lib.bundle.js` within the cluster.
+When imported in this manner, a `centraldashboard` module is created in the
+global scope. To establish a channel for communication between the iframed page
+and the Dashboard, use the following example which would disable a `<select>`
+element and assign the value from the Dashboard's Namespace selector when it
+changes:
+
+```
+window.addEventListener('DOMContentLoaded', function (event) {
+    if (window.centraldashboard.CentralDashboardEventHandler) {
+        // Init method will invoke the callback with the event handler instance
+        // and a boolean indicating whether the page is iframed or not
+        window.centraldashboard.CentralDashboardEventHandler
+            .init(function (cdeh, isIframed) {
+                var namespaceSelector = document.getElementById('ns-select');
+                namespaceSelector.disabled = isIframed;
+                // Binds a callback that gets invoked anytime the Dashboard's
+                // namespace is changed
+                cdeh.onNamespaceSelected = function (namespace) {
+                    namespaceSelector.value = namespace;
+                };
+            });
+    }
+});
+```
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -274,7 +274,7 @@ export class MainPage extends PolymerElement {
      * @param {Event} responseEvent AJAX-response
      */
     _onPlatformInfoResponse(responseEvent) {
-        const { response } = responseEvent.detail;
+        const {response} = responseEvent.detail;
         this.platformInfo = response;
         if (this.platformInfo.kubeflowVersion) {
             this.buildVersion = this.platformInfo.kubeflowVersion;
@@ -286,12 +286,11 @@ export class MainPage extends PolymerElement {
      * load event as well as when the namespace changes.
      */
     _sendNamespaceMessage() {
-        if (this._iframeConnected) {
-            this.$.PageFrame.contentWindow.postMessage({
-                type: NAMESPACE_SELECTED_EVENT,
-                value: this.namespace,
-            }, this._iframeOrigin);
-        }
+        if (!this._iframeConnected) return;
+        this.$.PageFrame.contentWindow.postMessage({
+            type: NAMESPACE_SELECTED_EVENT,
+            value: this.namespace,
+        }, this._iframeOrigin);
     }
 
     /**
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -33,6 +33,8 @@ import './namespace-selector.js';
 import './dashboard-view.js';
 import './activity-view.js';
 import './not-found-view.js';
+import {MESSAGE, PARENT_CONNECTED_EVENT, IFRAME_CONNECTED_EVENT,
+    NAMESPACE_SELECTED_EVENT} from '../library.js';
 
 const VALID_QUERY_PARAMS = ['ns'];
 
@@ -53,7 +55,7 @@ export class MainPage extends PolymerElement {
             subRouteData: Object,
             queryParams: {
                 type: Object,
-                value: () => {},
+                value: null, // Necessary to preserve queryString from load
             },
             iframeRoute: Object,
             menuLinks: {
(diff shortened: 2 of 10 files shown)


components/centraldashboard/public/components/main-page.js
```diff
@@ -66,7 +68,7 @@ export class MainPage extends PolymerElement {
                         href: '/docs',
                     },
                     {
-                        iframeUrl: '/jupyter/',
+                        iframeUrl: 'http://localhost:8000/test.html',
```
avdaredevil: 😕 ❓

components/centraldashboard/public/components/main-page.js
```diff
@@ -205,6 +215,36 @@ export class MainPage extends PolymerElement {
...
+        return url.href.slice(url.origin.length);
+    }
+
+    /**
+     * Sends a message to the iframe message bus. This is used on the iframe
+     * load event as well as when the namespace changes.
+     */
+    _sendNamespaceMessage() {
+        const iframe = this.$.PageFrame;
+        if (this.page === 'iframe' && this.queryParams.ns && iframe.src) {
```
avdaredevil: _**[nit]**_ Decrease code depth
```suggestion
        if (!(this.page === 'iframe' && this.queryParams.ns && iframe.src)) return;
```

components/centraldashboard/public/components/main-page.pug
```diff
@@ -45,7 +46,8 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
             neon-animatable(page='activity')
                 activity-view(namespace='[[queryParams.ns]]')
             neon-animatable#iframe-page(page='iframe')
-                iframe#PageFrame.flex(src='[[iframeUrl]]')
+                iframe#PageFrame.flex(src='[[iframeUrl]]',
+                    on-load='_sendNamespaceMessage')
```
avdaredevil: Don't need this in adherence to new design

components/centraldashboard/public/components/main-page.js
prodonjs: I added this so you could easily see how it behaves with a sample iframed page. See the latest update.

components/centraldashboard/public/components/main-page.pug
prodonjs: Ack'ed

prodonjs:
@avdaredevil PTAL.

This approach does the following:
- Packages an independent library which will be served from `/lib.bundle.js` on the cluster.
- Clients load the library which creates a global variable `centraldashboard`. There's a lot of nuance into how Webpack can package its output as a library (https://github.com/webpack/webpack/issues/2933) but  this approach is compatible with just about every technique used by the other sites in KF.
- Clients register callbacks for `onParentConnected` and `onNamespaceSelected`. Currently, the former doesn't pass any information back, but we could use it to prime the initial state such as passing the list of namespaces down. The latter is obviously where the iframed-app could set its namespace property.

See this [screencast](https://screencast.googleplex.com/cast/NTAwMzA4OTgzMDQ3NzgyNHw4ZDhkNDQ5OS04Nw) for a demo or pull this PR down and run `npm run build && npm start`. Then navigate to **http://localhost:8082/_/test** and observe how it behaves when selecting namespaces.

If you are in agreement with this approach, I can add tests for everything, cut out the parts used for demo purposes, and get ready to move this forward.

prodonjs:
@avdaredevil I think the code for this is in the final state ready for your review. I am going to work on building a guide and demonstrate how we can add this functionality to other applications using the [jupyter-web-app](https://github.com/kubeflow/kubeflow/tree/master/components/jupyter-web-app) as an example.

Until that's ready, I'm keeping this on hold but please review the code in the meantime.

/hold

prodonjs:
A working demonstration of the iframe communication behavior between the Central Dashboard and the Jupyter notebooks UI can be found at https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/.

Sample integration code is shown in the following branch:
https://github.com/kubeflow/kubeflow/compare/master...prodonjs:iframe-lib-import-test?expand=1

avdaredevil:
Hey Jason, 

Demo looks awesome! 2 things:
- Could we make the UI namespace be read-only in notebooks UI or simply a label instead of a live drop down when in iframe view.
- Could the API be revamped to work something like:

```js
window.addEventListener('DOMContentLoaded', function (event) {
    const {CentralDashboard_SDK} = window
    if (!CentralDashboard_SDK) return
    const sdk = new CentralDashboard_SDK()
    sdk.onNamespaceSelected((n, e) => {
        document.getElementById('ns-select').value = n // e could be the event if you really need it (optional for handler)
    })
});
```

prodonjs:
Excellent suggestions! I've implemented the changes and made the event handler a singleton.

> Hey Jason,
> 
> Demo looks awesome! 2 things:
> 
> * Could we make the UI namespace be read-only in notebooks UI or simply a label instead of a live drop down when in iframe view.

Yes, I definitely think that would be the ideal state once we're ready to roll out. I've implemented that behavior in the demo (https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/_/notebooks?ns=default)

> * Could the API be revamped to work something like:
> 
> ```js
> window.addEventListener('DOMContentLoaded', function (event) {
>     const {CentralDashboard_SDK} = window
>     if (!CentralDashboard_SDK) return
>     const sdk = new CentralDashboard_SDK()
>     sdk.onNamespaceSelected((n, e) => {
>         document.getElementById('ns-select').value = n // e could be the event if you really need it (optional for handler)
>     })
> });
> ```
> 
> ### EDIT:
> Based on further correspondence, I think an SDK model like this might make a better dev-experience, and a Socket.io like feel:
> 
> ```js
> CentralDashboard_SDK.init(sdk => {
>     // At this point, dom is loaded, sdk loaded, and an instance was passed
>     //  (this sdk can also be a singleton internally handled by the lib)
>     sdk.onNamespaceSelected((namespace, event) => /* handler */)
>     sdk.send('get-namespace')
>     sdk.send('recent-notebook', {notebook: 'AP Notebook', id: '2af3'})
>     sdk.telemetry(/*blah, which can be optionally implemented*/)
> })
> ```

I've implemented an API similar to the one you suggested. The handler is a singleton and the init method lets the user supply a callback that will get invoked with the instance and a flag indicating whether they are in an iframe or not. This would be the place to add any DOM-manipulation based on messages from the library. I've also updated the README with a sample of how to use it.

LMK what you think.

prodonjs:
/hold cancel
@avdaredevil PTAL. I think this should be ready to move ahead with. My latest commit exposes the child page styles as a cluster-wide resource and renames the exported library to dashboard_lib.bundle.js.

components/centraldashboard/webpack.config.js
```diff
@@ -48,10 +48,15 @@ const COMPONENT_RULES = [
 
 module.exports = {
     mode: ENV,
-    entry: resolve(SRC, 'index.js'),
+    entry: {
+        app: resolve(SRC, 'index.js'),
+        dashboard_lib: resolve(SRC, 'library.js'),
+    },
     output: {
         filename: '[name].bundle.js',
```
avdaredevil:
```suggestion
         filename: 'centraldashboard.sdk.js',
```

avdaredevil:
There are also conflicts

components/centraldashboard/README.md
```diff
@@ -118,3 +122,34 @@ body > .Main-Toolbar {background-color: var(--primary-background-color)}
...
+                var namespaceSelector = document.getElementById('ns-select');
+                namespaceSelector.disabled = isIframed;
+                // Binds a callback that gets invoked anytime the Dashboard's
+                // namespace is changed
+                cdeh.onNamespaceSelected = function (namespace) {
+                    namespaceSelector.value = namespace;
+                };
+            });
+    }
+});
```
avdaredevil: A few questions:
- Why can't `window.centraldashboard` **be** the `CentralDashboardEventHandler` or simply `window.CentralDashboardEventHandler` (if this name is preferred)
  - Also since the scope for the SDK is to also have the child be able to pass messages back to central-dashboard, wouldn't it make more sense to have a different name, since EventHandling isn't the only task it does?
- The approach of `cdeh.onNamespaceSelected = fn` only allows users to link a single handler, instead of a linked list approach. Is this an intended use case for all sub-apps?
  - What I mean is that `cdeh.onNamespaceSelected(fn)` would allow users to bind multiple handlers
- Since we're calling the lib `SDK`, wouldn't it make sense to name all the objects accordingly?
  - `cedh` -> `sdk`
  - `CentralDashboardEventHandler` -> `CentralDashboard_SDK`
- Are we using `var` because of retroactive support?

These are just notes/questions, I am not trying to suggest we do things one way or another. I am just making sure it makes sense from a user perspective. With that here is what I was envisioning:
```js
window.addEventListener('DOMContentLoaded', e => {
    const CD_SDK = window.CentralDashboard_SDK
    if (!CD_SDK) return
    CD_SDK.init((sdk, isIframed) => {
        const namespaceSelector = document.getElementById('ns-select')
        namespaceSelector.disabled = isIframed
        // Binds a callback that gets invoked anytime the Dashboard's
        // namespace is changed
        sdk.onNamespaceSelected(namespace => {
            namespaceSelector.value = namespace
        })
    })
})
```

components/centraldashboard/public/components/main-page.js
```diff
@@ -220,13 +256,47 @@ export class MainPage extends PolymerElement {
      */
     _buildHref(href, queryParams) {
         const url = new URL(href, window.location.origin);
-        VALID_QUERY_PARAMS.forEach((qp) => {
-            if (queryParams[qp]) {
-                url.searchParams.set(qp, queryParams[qp]);
-            }
-        });
+        if (queryParams) {
```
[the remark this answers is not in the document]
prodonjs: Normally I would agree, but in this case the return value is built using the same method regardless of whether or not the query parameters exist. The conditional test and subsequent iterator just enhances the output by adding the valid query parameters to the URL. The alternative seems less obvious from a readability standpoint.

```
if (!queryParams) {
    return url.href.slice(url.origin.length);
}
VALID_QUERY_PARAMS.forEach((qp) => {
    if (queryParams[qp]) {
        url.searchParams.set(qp, queryParams[qp]);
    }
});
return url.href.slice(url.origin.length);

File: components/centraldashboard/webpack.config.js (hunk 4 above)
Comment (prodonjs) [reply components/centraldashboard/webpack.config.js]: That will not work because the [name] placeholder gets replaced by the item from the `entry` property (ie. app for the application code, dashboard_lib for the library). That's why I renamed the entry to dashboard_lib so that the emitted output is `dashboard_lib.bundle.js`. `centraldashboard_lib.bundle.js` felt too long.

File: components/centraldashboard/public/components/main-page.js [hunk 7]
```diff
@@ -220,13 +256,47 @@ export class MainPage extends PolymerElement {
...
+        }
         return url.href.slice(url.origin.length);
     }
+
+    /**
+     * Sends a message to the iframe message bus. This is used on the iframe
+     * load event as well as when the namespace changes.
+     */
+    _sendNamespaceMessage() {
+        if (this._iframeConnected) {
```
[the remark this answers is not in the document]
prodonjs: I don't see any specific guidelines of code depth in the Google JS Style Guide. I tend to agree that getting beyond 2 levels decreases readability, but if there's a guideline that you think we should enforce, perhaps it's something that can be added to the eslint configuration to provide explicit direction.

components/centraldashboard/README.md
prodonjs: > Why can't window.centraldashboard be the CentralDashboardEventHandler or simply window.CentralDashboardEventHandler (if this name is preferred)
Also since the scope for the SDK is to also have the child be able to pass messages back to central-dashboard, wouldn't it make more sense to have a different name, since EventHandling isn't the only task it does?

Webpack exports the library as a [UMD](https://github.com/umdjs/umd) and since there are multiple exports rather than a single default export, we have to access the class by its module-qualified name. For the time being, I didn't see a scope beyond sending and responding to events  in the near to intermediate term future. I'm also not great at naming so if you think there's a better name I'm open to it.

> The approach of cdeh.onNamespaceSelected = fn only allows users to link a single handler, instead of a linked list approach. Is this an intended use case for all sub-apps?
What I mean is that cdeh.onNamespaceSelected(fn) would allow users to bind multiple handlers

For simplicity's sake I just allowed for a single handler rather than multiple observers. I couldn't think of a scenario where the latter would be necessary especially since we're building this to fit a fairly narrow requirement at the moment.

> Since we're calling the lib SDK, wouldn't it make sense to name all the objects accordingly?
cedh -> sdk
CentralDashboardEventHandler -> CentralDashboard_SDK

If you feel strongly about including SDK in the name we can obviously make that change. To me, SDK implies a set of tools to facilitate development on a particular platform which I don't necessarily feel that we're offering. Rather, I think we're offering a lightweight coupling between disparate applications. In the future, perhaps we would need other types of library utilities that might be better served from their own class structure and this naming facilitates that.

> Are we using var because of retroactive support?
Yes, since the Jupyter application does not go through any type of transpilation, the example snippet in the README is plain old ES3 Javascript.

If you feel we're missing the mark and would like to pull back into a deeper design phase, I'm happy to oblige and perhaps follow a more directed approach for integrating with the namespace selector on the Jupyter notebook server property.

prodonjs:
@avdaredevil Where do you think we stand on this? I'm working on the changes to generalize the iframe-link support and some of the namespace related changes inside of main-page need to be incorporated. I can either patch them in, or depending on where you think we are with respect to getting this merged, I can rebase after this is in.

PR #2930, opened 2019-04-03
update UI to include port-forward; update instructions

kunmingg:
Will deploy to staging so we can verify user experience.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2930)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -29,8 +29,8 @@ import appConfigPath from './user_config/app-config.yaml';
 
 const IngressType  = {
   BasicAuth: 'Login with Username Password',
-  Iap: 'Login with GCP Iap',
-  Skip: 'Skip Endpoint'
+  DeferIap: 'Setup Endpoint later',
+  Iap: 'Login with GCP Iap'
 };
 
 interface DeployFormState {
@@ -190,7 +190,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
                      value={this.state.ingress} onChange={this._handleChange('ingress')}>
             <MenuItem key={1} value={IngressType.Iap}>{IngressType.Iap}</MenuItem>
             <MenuItem key={2} value={IngressType.BasicAuth}>{IngressType.BasicAuth}</MenuItem>
-            <MenuItem key={3} value={IngressType.Skip}>{IngressType.Skip}</MenuItem>
+            <MenuItem key={3} value={IngressType.DeferIap}>{IngressType.DeferIap}</MenuItem>
           </TextField>
         </div>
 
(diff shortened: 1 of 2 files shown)


kunmingg:
/hold
need verify user experience on staging.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -29,6 +27,12 @@ import appConfigPath from './user_config/app-config.yaml';
 // selects auto domain then we should automatically supply the suffix
 // <hostname>.endpoints.<Project>.cloud.goog
 
+const IngressType  = {
+  BasicAuth: 'Login with Username Password',
+  Iap: 'Login with GCP Iap',
+  Skip: 'Skip Endpoint'
```
jlewi: What is skip doing?

In the case of Skip endpoint can we still setup IAP but with fake OAuth client credentials?

Should we call it something like

'Iap : partial setup'
or

'Iap:  Finish OAuth setup later'

I know its not great but I'd prefer a name that is accurate and consistent with long term plans.

jlewi: We can change this to something like shared OAuth client later when that's available.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -214,6 +233,12 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
           </div>
         </Collapse>
 
+        <Collapse in={this.state.ingress === IngressType.Skip}>
+          <div style={styles.row}>
+            <li>Skip creating service endpoint, use port-forward to connect to kubeflow service</li>
```
jlewi: Can we say something like access IAP via UI but complete OAuth setup later?

I don't want to call out port-forwarding because then we are setting an expectation that this is an option we will support in the future.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -389,6 +421,14 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
         value: 'NodePort'
       });
     }
+    // Remove ingress when user choose to use port-forward
+    if (this.state.ingress === IngressType.Skip) {
```
jlewi: I thought we were just going to supply fake client id, client_secret

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -406,6 +446,20 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
     }
   }
 
+  private async _toPortForward() {
```
jlewi: Do we need to support port-forwarding via Cloud Shell Anymore?

I guess we can keep it for now and remove post-next.

jlewi:
Thanks for the quick turn around.

kunmingg:
@jlewi 
Updated.
Now user can insert oauth later to enable IAP.
or later our backend can insert when detect request contains empty client id.

Issue #2716, opened 2019-03-15
Ship v0.5.0 release


kunmingg:
Track todos here and ship release v0.5.0
Todos:
- [x] https://github.com/kubeflow/kubeflow/issues/1859
- [x] https://github.com/kubeflow/kubeflow/issues/2588
- [x] https://github.com/kubeflow/kubeflow/issues/2723
- [x] https://github.com/kubeflow/kubeflow/issues/2641
- [x] https://github.com/kubeflow/kubeflow/pull/2733

TFJob blockers:
- [x] https://github.com/kubeflow/tf-operator/issues/550
- [x] https://github.com/kubeflow/tf-operator/issues/916
- [x] https://github.com/kubeflow/tf-operator/issues/929

Pytorch blockers:
- [x] https://github.com/kubeflow/pytorch-operator/issues/134

Update release images:

- [x] tf-operator
- [x] pytorch-operator
- [x] katib
- [x] jupyter images
- [x] dashboard

Release branch:
- [x] Cut release branch
- [ ] e2e release demo test on release branch 

Update Docs

- [ ] Update the changelog by running [update_changelog](https://github.com/kubeflow/kubeflow/blob/76107ff3aa7f9fde6fd7f15f2ec0a69a35874efa/scripts/update-changelog.sh)

Cherrypick:
- [x] #2778 
- [x] #2853 
- [x] #2875

Other
- [ ] kfctl pin to 0.5 version when it's ready.
- [ ] fix `kfctl version`

lluunn:
I will cut the first RC early next week

richardsliu:
Need to update release images for:
- tf-operator
- pytorch-operator
- katib
- jupyter images
- dashboard

kunmingg:
One more pending:
* Switch default to new notebook UI + controller before release cut
@zabbasi

lluunn:
notebook CR status:
https://github.com/kubeflow/kubeflow/issues/2588
https://github.com/kubeflow/kubeflow/issues/2723

@vkoukis @kimwnasptd

avdaredevil:
This PR:
#2733

lluunn:
admission webhook: https://github.com/kubeflow/kubeflow/issues/2641

mvartakAtVerta:
ModelDB V2 integration: #2748

IronPan:
One todo for Pipeline: https://github.com/kubeflow/kubeflow/issues/2754

gabrielwen:
[...]

PR #2934, opened 2019-04-04
kfctl will fail for --platform gcp if /Users/kdkasrav/.kube/config doesn't exist

### LINKED ISSUE #2933: kfctl will fail for --platform gcp if $HOME/.kube/config doesn't exist
Gcp will create /root/.kube/config but not until `kfctl apply ...` is called. Initializing Ksonnet.restConfig and Ksonnet.apiConfig in ksonnet GetKfApp is an error since they aren’t used until `kfctl apply...` . Since ksonnet is called after gcp in `kfctl apply...` initializing these values should be done inside of ksonnet Apply.

kkasravi:
fixes #2933

- Ordering of the components under bootstrap/config/{kfctl_default.yaml, kfctl_basic_auth.yaml, kfctl_iap.yaml} is important. Metacontroller needs to go first.
- Remove references to the "application" component in ksonnet.Delete
- Call ksonnet Delete to delete global resources

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2934)

### DIFF
--- a/bootstrap/Makefile
+++ b/bootstrap/Makefile
@@ -20,6 +20,7 @@ GOPATH ?= $(HOME)/go
 BOOTSTRAPPER_IMG ?= gcr.io/$(GCLOUD_PROJECT)/bootstrapper
 KFCTL_IMG ?= gcr.io/$(GCLOUD_PROJECT)/kfctl
 TAG ?= $(eval TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)-$(shell git diff | shasum -a256 | cut -c -6))$(TAG)
+KFCTL_TARGET ?= kfctl
 # set to -V
 VERBOSE ?= 
 PLUGINS_ENVIRONMENT ?= $(GOPATH)/src/github.com/kubeflow/kubeflow/bootstrap/bin
@@ -109,7 +110,7 @@ push-to-github-release: build-kfctl-tgz
 build-kfctl-container:
 	docker build \
 		--build-arg GOLANG_VERSION=$(GOLANG_VERSION) \
-		--target=kfctl \
+		--target=$(KFCTL_TARGET) \
 		--tag $(KFCTL_IMG):$(TAG) .
 	@echo Built $(KFCTL_IMG):$(TAG)
 
@@ -154,7 +155,7 @@ install: build-kfctl dockerfordesktop.so
 	@cp bin/kfctl /usr/local/bin
 
 run-kfctl-container: build-kfctl-container
-	@docker run -d -v $HOME/.kube:/root/.kube -t $(KFCTL_IMG):$(TAG)
+	@docker run -v $$HOME/.kube:/root/.kube -v $$HOME/.config:/root/.config --entrypoint /bin/sh -it $(KFCTL_IMG):$(TAG)
 
 run-local-docker:
 	docker run -d -it --name bootstrapper \
--- a/bootstrap/Makefile
+++ b/bootstrap/Makefile
@@ -21,6 +21,8 @@ BOOTSTRAPPER_IMG ?= gcr.io/$(GCLOUD_PROJECT)/bootstrapper
 KFCTL_IMG ?= gcr.io/$(GCLOUD_PROJECT)/kfctl
 TAG ?= $(eval TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)-$(shell git diff | shasum -a256 | cut -c -6))$(TAG)
 KFCTL_TARGET ?= kfctl
+MOUNT_KUBE ?= "-v $$HOME/.kube:/root/.kube"
+MOUNT_GCP ?= "-v $$HOME/.config:/root/.config"
 # set to -V
 VERBOSE ?= 
 PLUGINS_ENVIRONMENT ?= $(GOPATH)/src/github.com/kubeflow/kubeflow/bootstrap/bin
@@ -155,7 +157,7 @@ install: build-kfctl dockerfordesktop.so
 	@cp bin/kfctl /usr/local/bin
 
 run-kfctl-container: build-kfctl-container
-	@docker run -v $$HOME/.kube:/root/.kube -v $$HOME/.config:/root/.config --entrypoint /bin/sh -it $(KFCTL_IMG):$(TAG)
+	@docker run $(MOUNT_KUBE) $(MOUNT_GCP) --entrypoint /bin/sh -it $(KFCTL_IMG):$(TAG)
 
 run-local-docker:
 	docker run -d -it --name bootstrapper \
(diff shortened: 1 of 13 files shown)


bootstrap/config/kfctl_basic_auth.yaml
```diff
@@ -20,6 +20,7 @@ packages:
   - tf-serving
   - tf-training
 components:
+  - metacontroller
```
gabrielwen: can we keep it alphabetically ordered?

bootstrap/Makefile
```diff
@@ -20,6 +20,9 @@ GOPATH ?= $(HOME)/go
 BOOTSTRAPPER_IMG ?= gcr.io/$(GCLOUD_PROJECT)/bootstrapper
 KFCTL_IMG ?= gcr.io/$(GCLOUD_PROJECT)/kfctl
 TAG ?= $(eval TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)-$(shell git diff | shasum -a256 | cut -c -6))$(TAG)
+KFCTL_TARGET ?= kfctl
+MOUNT_KUBE ?=  -v $(HOME)/.kube:/root/.kube
```
jlewi: Why do you need to reference the Kubeconfig to build kfctl?

jlewi: NM.

bootstrap/config/kfctl_default.yaml
```diff
@@ -20,12 +20,12 @@ packages:
   - tf-serving
   - tf-training
 components:
+  - metacontroller
```
jlewi: Please call this change out in the PR description.

jlewi:
Please describe the fix and other changes in the PR description.

gabrielwen:
this PR needs to be part v0.5 release #2716

bootstrap/config/kfctl_basic_auth.yaml
gabrielwen: ok, I see some error messages that might be related to this change - could you add the reason to description as Jeremy suggested?  thanks.

gabrielwen:
/lgtm

will let @jlewi approve it :)

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
```diff
@@ -80,24 +78,24 @@ func GetKfApp(kfdef *kfdefs.KfDef) kftypes.KfApp {
...
-	_kfapp.apiConfig = kftypes.GetKubeConfig()
 	return _kfapp
 }
 
 // Apply applies the ksonnet components to target k8s cluster.
 // Remind: Need to be thread-safe: this entry is share among kfctl and deploy app
 func (ksApp *ksApp) Apply(resources kftypes.ResourceEnum) error {
-	if ksApp.restConfig == nil || ksApp.apiConfig == nil {
+	// build restConfig and apiConfig using $HOME/.kube/config if the file exist
+	restConfig := kftypes.GetConfig()
```
kunmingg: We cannot have apply() calling GetConfig():
When apply() is used by deploy app, there's no kube config

kunmingg:
@kkasravi
Can you reproduce?
Initializing Ksonnet.restConfig and Ksonnet.apiConfig in ksonnet GetKfApp will print warning message instead of throw error if .kube/config doesn't exist.

kunmingg:
@kkasravi 
It's working on master head without ```.kube/config```.
we shouldn't assume ```.kube/config``` exists during ksonnet.generate() since the cluster doesn't exist yet. [Gcp.Apply() runs after ksonnet.generate()]

Can we instead change logic at: https://github.com/kubeflow/kubeflow/blob/03bb4df7795264d890be5b4082284ef8bee6426f/bootstrap/pkg/kfapp/ksonnet/ksonnet.go#L466 ?
Instead of fetch from k8s master, we can assign a default value to it like 
```
k8sSpec = "version:v1.11.7"
```

kkasravi:
@kunmingg I believe this is what you were looking for. Ksonnet.initKs no longes assumes .kube/config.

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
```diff
@@ -51,8 +51,6 @@ type ksApp struct {
 	// ksonnet env name
 	KsEnvName  string
 	KApp       app.App
-	restConfig *rest.Config
```
kunmingg: Thanks! Can we keep credential objects in ksApp?

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
kunmingg: Can we update here as well?

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
kunmingg: caller from web app will build ksApp object and call thread-safe functions

jlewi:
/unassign @jlewi 
Looks like @kunmingg Has a better handle on this than me.

kkasravi:
@kunmingg @jlewi looks like there is a regression on delete since global resources no longer have an application label on them. I need to make a call to call the equivalent of `ks delete default`, I thought I would do it in this PR, I'll note it in the description. Should have the fix in an hour or so.

kkasravi:
Now correctly deleting all k8 resources

jlewi:
@kunmingg @kkasravi Anything blocking this ?

bootstrap/pkg/kfapp/ksonnet/ksonnet.go
```diff
@@ -80,19 +80,19 @@ func GetKfApp(kfdef *kfdefs.KfDef) kftypes.KfApp {
...
-	_kfapp.apiConfig = kftypes.GetKubeConfig()
 	return _kfapp
 }
 
 // Apply applies the ksonnet components to target k8s cluster.
 // Remind: Need to be thread-safe: this entry is share among kfctl and deploy app
 func (ksApp *ksApp) Apply(resources kftypes.ResourceEnum) error {
+	// build restConfig and apiConfig using $HOME/.kube/config if the file exist
+	ksApp.restConfig = kftypes.GetConfig()
+	ksApp.apiConfig = kftypes.GetKubeConfig()
```
kunmingg: Can we have
```
ksApp.restConfig = kftypes.GetConfig()
ksApp.apiConfig = kftypes.GetKubeConfig()
```
back to part of ```func GetKfApp(kfdef *kfdefs.KfDef)```?

```Apply()``` need to be thread safe and cannot set credentials.
``` kftypes.GetConfig()``` will print info instead of throw error when cannot get credential during ```generate```

kunmingg:
/lgtm
/approve
/hold cancel
Thanks!

Issue #2885, opened 2019-04-01
Include platform specific links in central dashboard


jlewi:
I think it might be valuable if the central dashboard could include links to platform/cloud resources related to Kubeflow.

For example on GCP it would be useful to have links to

1. the deployment manager page associated with a Kubeflow deployment
1. the GKE cluster page for a deployment
1. stackdriver logs.

I think this raises some design questions

1. How do we make the links pluggable since they will be different on every platform?

/cc @prodonjs @avdaredevil

prodonjs:
It looks like Spartakus is able to ascertain the platform information from the other nodes in the cluster (https://github.com/kubernetes-incubator/spartakus/blob/a593c5e0f13efa415752bf4ebaccdacddf7e979a/pkg/volunteer/kubernetes.go#L64), so that pattern can be followed assuming that the Central Dashboard's server component has the appropriate roles within the cluster.

If we can get a map of what we'd want to show for each platform, this is something we can target for 0.6.0.

jlewi:
@avdaredevil see https://github.com/kubeflow/kubeflow/issues/2976#issuecomment-485407082 for not being blocked on uber Application K8s resource being defined.

prodonjs:
@jlewi https://kubeflow-1.endpoints.prodonjs-kubeflow-dev.cloud.goog/ demonstrates the behavior for GCP-specific platform-links. We extract the project ID and use it to build the links that display below the *GCP Information* header. Currently, we don't have any links defined for other platforms.

We also retrieve the version number from the first Application CR found in the namespace of the dashboard to show it in the bottom right-hand corner. If no CR exists, we display unknown.

Any thoughts?

PR #2935, opened 2019-04-04
Use platform info to display provider-specific links.

prodonjs:
addresses #2885 

Adds new /api/platform-info endpoint to expose provider information to allow front-end to set platform specific links.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2935)

### DIFF
--- a/components/centraldashboard/app/k8s_service.ts
+++ b/components/centraldashboard/app/k8s_service.ts
@@ -3,22 +3,65 @@ import * as k8s from '@kubernetes/client-node';
 /** Information about the Kubernetes hosting platform. */
 interface PlatformInfo {
   provider: string;
+  providerName: string;
+  kubeflowVersion: string;
 }
+
+/**
+ * Relevant fields from the description property of the Application CRD
+ * https://github.com/kubernetes-sigs/application/blob/master/config/crds/app_v1beta1_application.yaml
+ */
+interface V1BetaApplicationDescriptor {
+  version: string;
+}
+
+/**
+ * Relevant fields from the spec property of the Application CRD
+ * https://github.com/kubernetes-sigs/application/blob/master/config/crds/app_v1beta1_application.yaml
+ */
+interface V1BetaApplicationSpec {
+  descriptor: V1BetaApplicationDescriptor;
+}
+
+/** Generic definition of the Kubeflow Application CRD */
+interface V1BetaApplication {
+  apiVersion: string;
+  kind: string;
+  metadata?: k8s.V1ObjectMeta;
+  spec: V1BetaApplicationSpec;
+}
+
+interface V1BetaApplicationList {
+  items: V1BetaApplication[];
+}
+
+const APP_API_GROUP = 'app.k8s.io';
+const APP_API_VERSION = 'v1beta1';
+const APP_API_NAME = 'applications';
 
 /** Wrap Kubernetes API calls in a simpler interface for use in routes. */
 export class KubernetesService {
-  private k8sApi: k8s.Core_v1Api;
+  private namespace = 'kubeflow';
+  private coreAPI: k8s.Core_v1Api;
+  private customObjectsAPI: k8s.Custom_objectsApi;
 
   constructor(private kubeConfig: k8s.KubeConfig) {
     console.info('Initializing Kubernetes configuration');
     this.kubeConfig.loadFromDefault();
-    this.k8sApi = this.kubeConfig.makeApiClient(k8s.Core_v1Api);
+    const context =
+        this.kubeConfig.getContextObject(this.kubeConfig.getCurrentContext());
+    if (context && context.namespace) {
+      this.namespace = context.namespace;
+    }
+    this.coreAPI = this.kubeConfig.makeApiClient(k8s.Core_v1Api);
+    this.customObjectsAPI =
+        this.kubeConfig.makeApiClient(k8s.Custom_objectsApi);
   }
 
   /** Retrieves the list of namespaces from the Cluster. */
   async getNamespaces(): Promise<k8s.V1Namespace[]> {
     try {
-      const {body} = await this.k8sApi.listNamespace();
+      const {body} = await this.coreAPI.listNamespace();
       return body.items;
     } catch (err) {
       console.error('Unable to fetch Namespaces:', err.body || err);
@@ -29,7 +72,7 @@ export class KubernetesService {
   /** Retrieves the list of events for the given Namespace from the Cluster. */
   async getEventsForNamespace(namespace: string): Promise<k8s.V1Event[]> {
     try {
-      const {body} = await this.k8sApi.listNamespacedEvent(namespace);
+      const {body} = await this.coreAPI.listNamespacedEvent(namespace);
       return body.items;
     } catch (err) {
       console.error(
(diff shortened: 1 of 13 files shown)


prodonjs:
/hold
/area front-end
/area centraldashboard
/assign @avdaredevil 
/assign @prodonjs 


@avdaredevil - I've got this in a state that you can take and work on the front-end side of things to at least get a style for the kinds of links you'd like to show. Here are a couple of things we will probably need to consider:
- Since namespaces and platform info will also be necessary for the landing page, we may want to consider consolidating the platform-info and namespaces server endpoints into something that fetches are returns all of the relevant data in a single request.
- We'll probably need to parse other pieces out of the providerID string in order to be able to provide project-relevant links to GCP console pages. The full providerID string for GCP looks like this: **gce://\<project\>/\<zone\>/\<node\>**, but we'll need to figure out what it looks like for the other nodes we want to support (if any).

prodonjs:
@avdaredevil Here is an updated screenshot of the current state. I am going to have this PR ready for review for you tomorrow.

Screenshot from 2019-04-22 16-39-11

avdaredevil:
Looks good, @prodonjs I feel the quick links and GCP links can probably be merged with a divider, thoughts?

prodonjs:
Screenshot from 2019-04-23 12-46-21

There's the screenshot of the look with the platform-specific links merged in with Quick Links. The code should be in a ready-to-review state. I am still going to hold on merging because I need to add additional RBAC permissions.

prodonjs:
/hold cancel

@avdaredevil This should be ready for your final review. You can view it live at https://kubeflow-1.endpoints.prodonjs-kubeflow-dev.cloud.goog.

I will need to make a separate commit for the RBAC changes since new APIs need to be accessed by the Central Dashboard's service account.

prodonjs:
Changes made to reorder links and place Stackdriver first.

components/centraldashboard/public/components/dashboard-view.css
```diff
@@ -52,25 +75,11 @@ paper-card.thin {
...
-}
-
-#Quick-Links .link .button {
-    color: var(--accent-color);
-    background: rgba(0, 125, 252, 0.25);
-    border-radius: 50%
+#Platform-Links header {
+    font-size: 18px;
+    font-weight: 400;
+    font-family: var(--paper-card-header_-_font-family);
```
avdaredevil: This is an internal Polymer variable, we should not use that. Please try to use the family directly, or the raw variable that feeds in this value.

components/centraldashboard/public/components/dashboard-view.css
```diff
@@ -52,25 +75,11 @@ paper-card.thin {
...
-    font-style: italic;
-    pointer-events: none
-}
-
-#Quick-Links .link .button {
-    color: var(--accent-color);
-    background: rgba(0, 125, 252, 0.25);
-    border-radius: 50%
+#Platform-Links header {
+    font-size: 18px;
```
avdaredevil:
```suggestion
    font-size: 1.25em;
```

Try to use `em`s since they scale according to ancestors and is easier to build responsive pages.

_**Note:** `1.25em` may not be the best, please experiment and find out what size is best_

components/centraldashboard/public/components/dashboard-view.css
```diff
@@ -52,25 +75,11 @@ paper-card.thin {
...
+    font-weight: 400;
+    font-family: var(--paper-card-header_-_font-family);
+    padding: 8px 16px;
```
avdaredevil:
```suggestion
    padding: .5em 1em;
```

As suggested above to replace `px` with `em`

components/centraldashboard/public/components/dashboard-view.js
```diff
@@ -55,16 +79,42 @@ export class DashboardView extends PolymerElement {
...
+            let gcpProject = '';
+            if (pieces.length >= 3) {
+                gcpProject = pieces[2];
+            }
+            this.platformLinks = GCP_LINKS.map((l) => {
+                return {
+                    text: l.text,
+                    link: l.link + gcpProject,
+                };
+            });
```
avdaredevil: This chunk above can be written as:
```suggestion
            this.platformLinks = GCP_LINKS.map((l) => ({
                text: l.text,
                link: l.link + gcpProject,
            }));
```

components/centraldashboard/public/components/dashboard-view.pug
```diff
@@ -9,10 +9,17 @@ paper-card#Getting-Started(heading='Getting Started')
...
-    article.link.more-coming
-        paper-item-body More coming soon
-        paper-icon-button.button(icon='arrow-forward', disabled)
+        a.link(href$='[[item.link]]', tabindex='-1', target='_blank')
+            paper-icon-item [[item.text]]
+                paper-icon-button.button(icon='arrow-forward',
+                    slot='item-icon', alt='[[item.text]]')
+    template(is='dom-if', if='[[platformLinks]]')
+        section#Platform-Links
+            header [[ platformName ]] Information
```
avdaredevil:
```suggestion
            header [[platformName]] Information
```

components/centraldashboard/app/k8s_service.ts
```diff
@@ -24,12 +72,61 @@ export class KubernetesService {
...
+    }
+
+    // Gets the first Application entity for the namespace
+    try {
+      const response = await listApplicationsPromise;
+      const body = response.body as V1BetaApplicationList;
+      if (body.items && body.items.length && body.items[0].spec &&
+          body.items[0].spec.descriptor &&
+          body.items[0].spec.descriptor.version) {
+        info.kubeflowVersion = body.items[0].spec.descriptor.version;
```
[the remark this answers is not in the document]
prodonjs: While I agree that this is a more succinct solution, it's also much more difficult to read. There are alternatives that would allow for Java 8 style Optionals (https://github.com/bromne/typescript-optional), and it may make sense to include them in the future but I tend to favor the [Python-zen guidelines](https://www.python.org/dev/peps/pep-0020/#id3) to avoid favoring brevity over readability.

components/centraldashboard/app/k8s_service.ts
```diff
@@ -24,12 +72,61 @@ export class KubernetesService {
...
+    const listApplicationsPromise =
+        this.customObjectsAPI.listNamespacedCustomObject(
+            APP_API_GROUP, APP_API_VERSION, this.namespace, APP_API_NAME);
+    listApplicationsPromise.catch(rejectionHandler);
+
+    // Gets the providerID from the first node
+    try {
+      const {body} = await listNodePromise;
+      const provider = body.items.map((n) => n.spec.providerID).find(Boolean);
+      if (provider) {
```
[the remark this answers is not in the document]
prodonjs: I think this comment was on a prior version. In the current code, the clause is necessary because it's where we set the return value.

components/centraldashboard/public/components/dashboard-view.css
prodonjs: Understood. It looks like this is unnecessary anyways as the font-family is being inherited as Google Sans like we want.

components/centraldashboard/public/components/dashboard-view.css
prodonjs: 1.25 seems to come out pretty well.

components/centraldashboard/public/components/dashboard-view.css
prodonjs: Done.

components/centraldashboard/public/components/dashboard-view.js
prodonjs: Understood but it does require the additional enclosing parentheses to tell the interpreter to treat the expression as an object literal. I wish it worked a little more like single-statement lambdas in Java but for multi-line expressions, I find the explicit return to be a bit more obvious when reading.

components/centraldashboard/public/components/dashboard-view.pug
prodonjs: Done

prodonjs:
@avdaredevil Comments addressed.

components/centraldashboard/public/components/dashboard-view.js
avdaredevil: Yeah, all languages have their nuances. But if you think `return`s are generally more readable, we can stick to that type of syntax.

components/centraldashboard/app/k8s_service.ts
```diff
@@ -24,12 +72,71 @@ export class KubernetesService {
...
+  }
+
+  /**
+   * Returns the provider identifier or 'other://' from the K8s cluster.
+   */
+  private async getProvider(): Promise<string> {
+    let provider = 'other://';
+    try {
+      const {body} = await this.coreAPI.listNode();
+      const foundProvider =
```
avdaredevil: 110 - 114 can be written as:
```js
    provider = body.items.map((n) => n.spec.providerID).find(Boolean) || provider;
```

avdaredevil: Everything else looks good!

prodonjs: Understood. I think this is probably stylistic preference more than anything else which I tend to error on the side of block constructs where possible since to me the intuition here is that if it's not set, we won't make any changes to the variable rather than reassign its value to itself. If you feel strongly about making the change I can do so.

PR #2938, opened 2019-04-04
Extract exact backend service name

### LINKED ISSUE #2550: IAP enabler issue: same NodePort port number
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/gcp/setup_backend.sh#L30

I ran into a situation that there are two backend services with the same port number.
BACKEND_ID=7586477182690125543 9050688055668560594

And then this [line](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/gcp/setup_backend.sh#L56) 
failed because the second number is not a command
```
/var/shared/healthz.env: line 2: 9050688055668560594: command not found
```

cc @jlewi 

Wondering if there is a more robust way to determine the backendID

gabrielwen:
Fixes #2550 

Use `.metadata.annotations` information in ingress service to extract exact backend service name.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2938)

### DIFF
--- a/kubeflow/gcp/iap.libsonnet
+++ b/kubeflow/gcp/iap.libsonnet
@@ -9,6 +9,7 @@
       envoyAdminPort: 8001,
       envoyStatsPort: 8025,
       useIstio: util.toBool(_params.useIstio),
+      ingressService: "envoy-ingress"
     },
     local namespace = if params.useIstio then params.istioNamespace else params.namespace,
 
@@ -384,6 +385,10 @@
                   {
                     name: "SERVICE",
                     value: if params.useIstio then "istio-ingressgateway" else "envoy",
+                  },
+                  {
+                    name: "INGRESS_SERVICE",
+                    value: params.ingressService,
                   },
                   {
                     name: "ENVOY_ADMIN",
--- a/kubeflow/gcp/iap.libsonnet
+++ b/kubeflow/gcp/iap.libsonnet
@@ -9,7 +9,7 @@
       envoyAdminPort: 8001,
       envoyStatsPort: 8025,
       useIstio: util.toBool(_params.useIstio),
-      ingressService: "envoy-ingress"
+      ingressName: "envoy-ingress"
     },
     local namespace = if params.useIstio then params.istioNamespace else params.namespace,
 
@@ -387,8 +387,8 @@
                     value: if params.useIstio then "istio-ingressgateway" else "envoy",
                   },
                   {
-                    name: "INGRESS_SERVICE",
-                    value: params.ingressService,
+                    name: "INGRESS_NAME",
+                    value: params.ingressName,
                   },
                   {
                     name: "ENVOY_ADMIN",
@@ -889,7 +889,7 @@
                   },
                   {
                     name: "INGRESS_NAME",
-                    value: "envoy-ingress",
+                    value: params.ingressName,
                   },
                 ],
                 volumeMounts: [
@@ -920,7 +920,7 @@
       apiVersion: "extensions/v1beta1",
       kind: "Ingress",
       metadata: {
-        name: "envoy-ingress",
+        name: params.ingressName,
         namespace: namespace,
         annotations: {
           "kubernetes.io/tls-acme": "true",
@@ -975,7 +975,7 @@
           config: [
             {
               http01: {
-                ingress: "envoy-ingress",
+                ingress: params.ingressName,
               },
               domains: [
                 params.hostname,
@@ -1005,7 +1005,7 @@
       spec: {
         project: endpointParams.project,
         targetIngress: {
-          name: "envoy-ingress",
+          name: params.ingressName,
           namespace: namespace,
         },
       },
--- a/kubeflow/gcp/setup_backend.sh
+++ b/kubeflow/gcp/setup_backend.sh
@@ -26,7 +26,7 @@ BACKENDS=$(kubectl --namespace=${NAMESPACE} get ingress ${SERVICE}-ingress -o js
 BACKEND_NAME=$(echo $BACKENDS | grep -o "k8s-be-${NODE_PORT}--[0-9a-z]\+")
 while [[ -z ${BACKEND_ID} ]]; do
   BACKEND_ID=$(gcloud compute --project=${PROJECT} backend-services list --filter=${BACKEND_NAME} --format='value(id)')
-  echo "Waiting for backend id PROJECT=${PROJECT} NAMESPACE=${NAMESPACE} SERVICE=${SERVICE} filter=name~k8s-be-${NODE_PORT}- ..."
+  echo "Waiting for backend id PROJECT=${PROJECT} NAMESPACE=${NAMESPACE} SERVICE=${SERVICE} filter=name~${BACKEND_NAME}"
   sleep 2
 done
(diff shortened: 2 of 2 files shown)


kubeflow/gcp/setup_backend.sh
```diff
@@ -22,9 +22,18 @@ gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS
 gcloud config list
 
 NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[0].nodePort}')
+echo "node port is ${NODE_PORT}"
+
+while [[ -z ${BACKENDS} ]]; do
+  BACKENDS=$(kubectl --namespace=${NAMESPACE} get ingress ${SERVICE}-ingress -o jsonpath='{.metadata.annotations.ingress\.kubernetes\.io/backends}')
```
lluunn: using ${SERVICE}-ingress as ingress name is too fragile. Can we pass it in as a environment var?

kubeflow/gcp/setup_backend.sh
```diff
@@ -22,9 +22,18 @@ gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS
...
+  echo "fetching backends info: ${BACKENDS}"
+  sleep 2
+done
+BACKEND_NAME=$(echo $BACKENDS | grep -o "k8s-be-${NODE_PORT}--[0-9a-z]\+")
+echo "backend name is ${BACKEND_NAME}"
```
lluunn: should we check backend_name is not empty ?

kubeflow/gcp/setup_backend.sh
```diff
@@ -22,9 +22,18 @@ gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS
...
 while [[ -z ${BACKEND_ID} ]]; do
-  BACKEND_ID=$(gcloud compute --project=${PROJECT} backend-services list --filter=name~k8s-be-${NODE_PORT}- --format='value(id)')
-  echo "Waiting for backend id PROJECT=${PROJECT} NAMESPACE=${NAMESPACE} SERVICE=${SERVICE} filter=name~k8s-be-${NODE_PORT}- ..."
+  BACKEND_ID=$(gcloud compute --project=${PROJECT} backend-services list --filter=${BACKEND_NAME} --format='value(id)')
```
lluunn: do you need "name~" in filter?

kubeflow/gcp/setup_backend.sh
gabrielwen: seems it will be different if using istio:
https://github.com/kubeflow/kubeflow/blob/48edfe412d0a3dd1dd4efa73a866dc5826c92336/kubeflow/gcp/iap.libsonnet#L386

could you show me the service name if using istio?

kubeflow/gcp/setup_backend.sh
gabrielwen: I've never checked if not using it - it works with it, though.

kubeflow/gcp/setup_backend.sh
gabrielwen: done.

kubeflow/gcp/setup_backend.sh
lluunn: The service after ingress is istio-ingressgateway if using Istio.
 I think we should just make ingress name as a param in iap.libsonnet, and pass it here as environment

gabrielwen: just to make sure: is it `envoy-ingress` for IAP or `istio-ingressgateway` for istio?

gabrielwen: actually, is it possible that you bring up a KF cluster with Istio so that I could update the logic with it?

lluunn: They are different things. You can check the iap.libsonnet:
1. the ingress name is always envoy-ingress (we can change, but doesn't matter)
2. the SERVICE (the service right behind ingress) is envoy for now, and istio-ingressgateway for iap.

gabrielwen: updated, please help me take a look, thanks!

kubeflow/gcp/iap.libsonnet
```diff
@@ -9,6 +9,7 @@
       envoyAdminPort: 8001,
       envoyStatsPort: 8025,
       useIstio: util.toBool(_params.useIstio),
+      ingressService: "envoy-ingress"
```
lluunn: can we call this ingressName? I think it's more clear

kubeflow/gcp/iap.libsonnet
```diff
@@ -385,6 +386,10 @@
                     name: "SERVICE",
                     value: if params.useIstio then "istio-ingressgateway" else "envoy",
                   },
+                  {
+                    name: "INGRESS_SERVICE",
```
lluunn: same here, the env var should be INGRESS or INGRESS_NAME

kubeflow/gcp/iap.libsonnet
lluunn: Please also change the ingress component name to use this variable

kubeflow/gcp/setup_backend.sh
```diff
@@ -22,9 +22,19 @@ gcloud auth activate-service-account --key-file=${GOOGLE_APPLICATION_CREDENTIALS
 gcloud config list
 
 NODE_PORT=$(kubectl --namespace=${NAMESPACE} get svc ${SERVICE} -o jsonpath='{.spec.ports[0].nodePort}')
+echo "node port is ${NODE_PORT}"
+
+while [[ -z ${BACKEND_NAME} ]]; do
+  BACKENDS=$(kubectl --namespace=${NAMESPACE} get ingress ${INGRESS_SERVICE} -o jsonpath='{.metadata.annotations.ingress\.kubernetes\.io/backends}')
```
lluunn: check the new env var not empty at line 6.

kubeflow/gcp/setup_backend.sh
lluunn: but you removed the "name~". Do you want to put it back?

gabrielwen:
quick question: do you think we should include this in 0.5 release?

kubeflow/gcp/iap.libsonnet
gabrielwen: done.

kubeflow/gcp/iap.libsonnet
gabrielwen: done.

lluunn:
I think so. We are going to do some cherrypick, and can include this one.

kubeflow/gcp/setup_backend.sh
gabrielwen: done.

kubeflow/gcp/setup_backend.sh
gabrielwen: done.

lluunn:
Thanks for the fix!
/lgtm
/approve

gabrielwen:
verified with manual deployment.

PR #2939, opened 2019-04-04
quick fix for deploy UI with v0.5

kunmingg:
* Skip password input check when choose "Setup Endpoint later"
* rename error from ```spec not found```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2939)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -118,7 +118,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
       ingress: IngressType.Iap,
       kfversion: 'v0.4.1',
       // Version for local test. Staging and Prod with overwrite with their env vars.
-      kfversionList: ['v0.4.1'],
+      kfversionList: ['v0.5.0-rc.3', 'v0.4.1'],
       password: '',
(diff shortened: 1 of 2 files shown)


lluunn:
which part is "skip endpoint"?

kunmingg:
@lluunn 
One option on UI called ```Setup Endpoint later```:
User won't have an working ingress, and they don't need to create oauth or input username/password

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -490,7 +490,8 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
           return;
         }
       }
-    } else {
+    }
```
lluunn: So skip endpoint will not IngressType.Iap nor IngressType.BasicAuth?
Can we do else if (this.state.ingress === IngressType.BasicAuth) here, and then 
else {
// skip endpoint
log something
}

kunmingg: Log here would be user facing as this part runs at user browser.
Do you mean print some messages for  "Setup Endpoint later" case?
There's a message field for each mode, we can use that to explain behavior of user's choice

PR #2967, opened 2019-04-08
fix python env for common packages

stpabhi:
This PR will address where common packages are not installed for python3 (base) environment.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2967)

### DIFF
--- a/components/tensorflow-notebook-image/install.sh
+++ b/components/tensorflow-notebook-image/install.sh
@@ -46,7 +46,7 @@ jupyter labextension install @jupyter-widgets/jupyterlab-manager
 
 # Install common packages from requirements.txt for both python2 and python
 pip --no-cache-dir install -r /tmp/requirements.txt
-source activate py2
+source activate base
 pip --no-cache-dir install -r /tmp/requirements.txt
 
 # Do chown in this layer for significant size savings


components/tensorflow-notebook-image/install.sh
```diff
@@ -46,7 +46,7 @@ jupyter labextension install @jupyter-widgets/jupyterlab-manager
 
 # Install common packages from requirements.txt for both python2 and python
 pip --no-cache-dir install -r /tmp/requirements.txt
-source activate py2
+source activate base
```
jlewi: So is the problem that source activate py2 has no effect because that's already the default?

What is the base environment?

stpabhi: ```source activate py2``` is done [here](https://github.com/kubeflow/kubeflow/blob/master/components/tensorflow-notebook-image/install.sh#L7) already. ```base``` is python3 environment and the latter activate should be ```base``` to install the packages from requirements.txt

jlewi:
/lgtm
/approve
/hold 
Thanks! @zabbasi could you take a quick look and LGTM if it looks good to you

PR #2968, opened 2019-04-08
set injectGcpCredentials to true by default

kunmingg:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2968)


lluunn:
it will add labels by default, but no real effect if not on gcp. cc @jlewi 

/lgtm
/approve

Issue #2911, opened 2019-04-02
kfctl evaluate kustomize KEP 


kkasravi:
[Kustomize Generators and Transformers](https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/20190318-kustomize-generators-transformers.md#implementation-detailsnotesconstraints-optional) adds 2 new fields transformers and generators. These may provide solutions within current ksonnet components including testing.

jlewi:
@kkasravi What is the remaining work to be done here?

kkasravi:
Generators are available in kustomize and generators can be used in manifests testing to generate tests. No more work is required here

Issue #2629, opened 2019-03-05
kfctl kustomize, ksonnet - package manager info should be integrated into config files


kkasravi:
Currently kfctl will open a file under bootstrap/config (depending on the init arguments) as part of it's initialization. These files are default.yaml, gcp_prototype.yaml, kfctl_basic_auth.yaml, kfctl_iap.yaml.
This information is read into the golang struct ComponentConfig 
```
type ComponentConfig struct {
	// Name of repository.
	Repo string `json:"repo,omitempty"`
	// List of default components.
	Components []string `json:"components,omitempty"`
	// List of default packages.
	Packages []string `json:"packages,omitempty"`
	// Parameters to be set to components using ks param set.
	ComponentParams Parameters `json:"componentParams,omitempty"`
	// Platform type.
	Platform string `json:"platform,omitempty"`
}
```

Currently this doesn't include any package manager information and should. 

The generated file 'app.yaml' that is produced after `kfctl init <name>` 
should also include package manager information so that subsequent 
subcommands `kfctl generate...`, `kfctl apply...` know which package manager to
delegate to.

The package manager stanza's need to include version information since each package manager
will download a repo using version information

ksonnet will download https://github.com/kubeflow/kubeflow/tarball/" + coordinator.Client.Spec.Version + "?archive=tar.gz

kustomize will download https://github.com/kubeflow/manifests/tarball/" + kustomize.Kustomize.Spec.Version + "?archive=tar.gz

jlewi:
@kkasravi Any update on this?

Looks like #2685 may be related.

jlewi:
@kkasravi Is this fixed by #2971?

kkasravi:
yes

Issue #2599, opened 2019-03-01
kfctl kustomize implement Apply and Delete methods


kkasravi:
Kustomize.Apply will need to do something similar to kubectl apply -f output.yaml
Kustomize.Delete will need to delete all resources including CRDs, ClusterRoles, ClusterRoleBindings and PVs. 

Apply can use k8utils.go CreateResourceFromFile
Delete should probably look for labels. Ksonnet.Delete should be referenced.

swiftdiaries:
/assign @swiftdiaries 
I'll pick this one up

swiftdiaries:
This is done. For now, the YAML built using kustomize is piped out to kubectl.
Link: https://github.com/swiftdiaries/kubeflow/commit/a0ef6bb14e879893e80c1af00333ab9d78b63cc8

swiftdiaries:
I'll create a PR as soon as #2548 is merged because this is based off of that

swiftdiaries:
I guess this is already done.
We can re-visit if it's warranted.

/cc @kkasravi 
/close

Issue #2568, opened 2019-02-27
kfctl kustomize implement Init 


kkasravi:
Implement the Init method which will require a Kustomize Struct similar to Ksonnet's
```
// Ksonnet implements the KfApp Interface
type KsApp struct {
	// ksonnet root name
	KsName string
	// ksonnet env name
	KsEnvName string
	KApp      app.App
	// kstypes.Ksonnet is autogenerated similar to k8 types
	// It has a Spec member of type KsonnetSpec that holds common fields like platform, version, repo
	// components, packages, parameters. This information is used to generate a ksonnet app.
	// It is not persisted into kfctl's app.yaml.
	KsApp *kstypes.Ksonnet
}
```

Init will require this definition and implementation of the function GetKfApp.
It will also require downloading the kustomize manifest to be downloaded 
similar to what Ksonnet does in its Init.

jlewi:
@kkasravi Is this fixed by #2971?

kkasravi:
yes

PR #2971, opened 2019-04-08
Kfctl switch packagemanagers

### LINKED ISSUE #2445: Add kustomize support to kfctl

kkasravi:
Implementation of kustomize. See [First Class Support for Kustomize Packages In Kubeflow](https://docs.google.com/document/d/156H79GiQ-gCAputHzcCdmD6gZZaMBLqC-kzwKqc2MIA/edit#heading=h.c4khqss4mjzp)

fixes #2445, #2566, #2568, #2599, #2629, #2911 

paired with [kubeflow/manifests/#13](https://github.com/kubeflow/manifests/pull/13)

The kustomize package manager will read its app.yaml and apply the packages, components and componentParams to kustomize in the following way:

- packages are always top-level directories under the manifests repo
- all components are also directories but may be a subdirectory in a package.
  - components may be a top-level directory if there is a base, overlay in that directory and the component name is equal to the package name. 
  - otherwise a component is a sub-directory whose name matches the component name in app.yaml.
- component parameters are applied to a component's params.env file which must contain an entry whose key matches the component parameter. The params.env file is used to generate a ConfigMap. Entries in params.env are resolved as kustomize vars or referenced in a deployment or statefulset env section.
- components are output as `<component>.yaml` under the kustomize subdirectory after `kfctl generate...`. 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2971)


kkasravi:
A few runtime errors remaining that I'll be able to correct by Monday

- ImagePullBackOff
  Deployment: centraldashboard
  Reason	
  Cannot pull image 'gcr.io/kubeflow-images-public/centraldashboard:0.5.0' from the registry.
- PodUnschedulable
Deployment: minio
Reason	
Cannot schedule pods: persistentvolumeclaim "minio-pvc" not found.
- PodUnschedulable
Deployment: mysql
Reason	
Cannot schedule pods: persistentvolumeclaim "mysql-pv-claim" not found.

kkasravi:
@jlewi I think this is stable enough to go in. I'll open separate issues based on this for subsequent improvements.

jlewi:
/lgtm
/approve

I think we should follow up on the dynamic creation of the Application resource; #3047

jlewi:
PR to keep Argo test results longer should be merged so hopefully on next test run test failures won't be deleted.

/test all

Not sure why the CLA check is failing.

jlewi:
Woo Hoo!
/lgtm
/approve

PR #2985, opened 2019-04-10
update kfctl version for 0.5 branch

lluunn:
for #2716

- kfctl should use v0.5.0 by default
- `kfctl version` will print v0.5.0

Will rebuild the binary after this.

/cc @kunmingg @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2985)

### DIFF
--- a/bootstrap/cmd/kfctl/cmd/version.go
+++ b/bootstrap/cmd/kfctl/cmd/version.go
@@ -43,5 +43,5 @@ func init() {
 }
 
 func versionfunc(cmd *cobra.Command, args []string) {
-	fmt.Println("v20181207-4e7f4ed-198-gaeea303e-dirty-03e65e")
+	fmt.Println("v0.5.0")
 }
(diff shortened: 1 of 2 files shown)


lluunn:
```
 Error: couldn't apply KfApp:  (kubeflow.error): Code 500 with message: coordinator Apply f
ailed for gcp:  (kubeflow.error): Code 400 with message: gcp apply could not update deployment manager Error Set New IamPolicy err
or: googleapi: Error 400: The number of members in the policy (1,502) is larger than the maximum allowed size 1,500., badRequest
```

jlewi:
@kkasravi @lluunn Could we use ldflags (https://blog.polyverse.io/how-to-embed-versioning-information-in-go-applications-f76e2579b572) to allow the version to be easily substituted in at build time?

jlewi:
Should we clearly call out the following?
 
   * Commit - of the source code
   * Tag (e.g. 0.5 vs. master)
   * Arch - Architecture it was built for

jlewi:
Looks like like @kkasravi has #2988 to use a build flag

Issue #368, opened 2018-03-06
Liveness/Readiness checks for TF Serving


jlewi:
See #64.

We should configure liveness and readiness probes for TF Serving. 

Its not clear if TF Serving out of the box has handlers that would be suitable for use in a liveness/readiness probe. I don't know that a gRPC server provides any either. So this could potentially require upstream work to TFServing.

jlewi:
Does anyone know if TFServing provides a suitable handler that we could use for a basic readiness or liveness probe?

What about using the latest version of TFServing which has a REST interface?

jlewi:
From Peng Yu in Slack

They are using the gRPC port

```
    readinessProbe:
          tcpSocket:
            port: 8500
          initialDelaySeconds: 60
          timeoutSeconds: 300
```

lluunn:
https://github.com/tensorflow/serving/blob/master/tensorflow_serving/model_servers/http_rest_api_handler.cc#L188

TF serving's http server supports get model status. We can use that for readiness probe.

jlewi:
I think we need to cherry pick this into the release branch.

jlewi:
Lets cherrypick this onto the 0.3. branch and then close this issue.

We can figure out what to do about a readiness probe in the future.

jlewi:
Looks like it was cherry-picked onto the v0.3-branch 
https://github.com/kubeflow/kubeflow/blob/v0.3-branch/kubeflow/tf-serving/tf-serving-template.libsonnet#L122

So closing this issue.

PR #2987, opened 2019-04-10
Add model status endpoint to http-proxy

mrkm4ntr:
Add model status endpoint to http-proxy. This endpoint will be used to support readiness probe of tf-serving.
BTW, I'm not quite sure where is image of http-proxy built. After new image will be built,  I will create a PR to add readiness probe to ksonnet template of http-proxy container like below.
```
readinessProbe: {
  httpGet: {
    path: "/model/" + $.params.modelName + ":status",
    port: 8000,
  },
  initialDelaySeconds: 30,
  periodSeconds: 30,
},
```

Related issue about readiness probe is here.
https://github.com/kubeflow/kubeflow/issues/368#issuecomment-429670371

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2987)


lluunn:
Hi, thanks for the PR.
However, we are not using that http-proxy anymore because TF Serving has native http support.
https://www.tensorflow.org/tfx/serving/api_rest

mrkm4ntr:
OK, I will send PR for supporting readiness probe using native REST API.

mrkm4ntr:
@lluunn The REST endpoint of tf-serving can not be used for readiness probe because it returns always 200. We have to check response whether model status is available. I completely agree that http-proxy is not needed currently, but some sidecar proxy is needed for readiness probe. How do you think?

mrkm4ntr:
I am talking about sidecar health server in the following article.
https://ahmet.im/blog/advanced-kubernetes-health-checks/

PR #2999, opened 2019-04-11
Adds favicons to Central Dashboard

prodonjs:
Uses the same iconset as the Kubeflow docs website.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2999)


prodonjs:
Hey @avdaredevil , I realized we didn't have any favicons in our header so I copied the ones from the Docs site and added them here.

PR #3019, opened 2019-04-15
Allow multiple namespaces

### LINKED ISSUE #3000: Jupyter UI should allow creating notebooks in all namespaces
in Kubeflow v 0.5, UI allows to launch notebook in only Kubeflow namespace [#2925](https://github.com/kubeflow/kubeflow/issues/2925). 
This is because, Kubeflow 0.5 i) by default GCP and K8S service accounts are only available in kubeflow namespace, and ii) every notebook is expected to apply an admission hook to mount GCP and K8S service accounts.  Therefore, notebooks fails to create in other namespaces, and consequently UI disables launching notebooks in other namespaces.

We need to modify UI so that it allows launching notebooks in all namespaces. This is because we would like  to enable testing various works for multi-user Jupyter. Further, if users manually create gcp and and k8s service accounts in their desired namespaces, [as described here,](https://github.com/zabbasi/website/blob/jupyter2/content/docs/components/jupyter.md) they can safely create notebook.

kimwnasptd:
closes #3000 
(yay we reached 3000)
Allow a Notebook to be launched in any Namespace from the UI

will need to make a small follow up PR to update the image ksonnet param once the new image gets build from GCB.

cc @zabbasi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3019)

### DIFF
--- a/components/jupyter-web-app/kubeflow_jupyter/default/app.py
+++ b/components/jupyter-web-app/kubeflow_jupyter/default/app.py
@@ -162,7 +162,5 @@ def home():
     nmsps = [base_ns]
     logger.warning("Can't  list namespaces: %s" % api.parse_error(e))
 
-  # todo(apverma): To Be removed after multi-user isolation support is implemented
-  nmsps = [base_ns] # Issue #2937
   return render_template(
       'notebooks.html', prefix=prefix(), title='Notebooks', namespaces=nmsps)
--- a/components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
+++ b/components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
@@ -13,7 +13,7 @@
     <select class="mdl-textfield__input" id="ns-select" name="ns">
       <!-- Fill with the namespaces the user can see -->
       {% for ns in namespaces %}
-        <option value="{{ ns }}">{{ ns }}</option>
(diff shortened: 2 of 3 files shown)


jlewi:
/assign @zabbasi 

@zabbasi @lluunn Can you review this PR please? I believe this is blocking progress on multi-user jupyter.

components/jupyter-web-app/kubeflow_jupyter/default/app.py
```diff
@@ -162,7 +162,5 @@ def home():
     nmsps = [base_ns]
     logger.warning("Can't  list namespaces: %s" % api.parse_error(e))
 
-  # todo(apverma): To Be removed after multi-user isolation support is implemented
-  nmsps = [base_ns] # Issue #2937
   return render_template(
-      'notebooks.html', prefix=prefix(), title='Notebooks', namespaces=nmsps)
+    'notebooks.html', prefix=prefix(), title='Notebooks', namespaces=nmsps)
```
avdaredevil: Could we also get a change that defaults `kubeflow` by default? It was being added here: #3074

zabbasi:
@avdaredevil  can we please merge this and set the default in another pr?

components/jupyter-web-app/kubeflow_jupyter/default/app.py
kimwnasptd: Sure!

PR #3025, opened 2019-04-15
Use exact name when updating backend healthz path

### LINKED ISSUE #3014: healthy check URI is / instead of /healthz
Please look at `path`:
image

gabrielwen:
Fixes #3014 

Root cause is because of port conflict.  It seems when installing ISTIO, node port is always assigned as [31380](https://github.com/kubeflow/kubeflow/blob/cbc792bd082f1c3575a02f62faaa643cbbe198d8/dependencies/istio/install/istio-noauth.yaml#L13745).  Therefore when more than one deployments exists, update_backend.sh will not succeed when updating healthz path.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3025)


lluunn:
/lgtm
/approve
thanks!

PR #3038, opened 2019-04-17
change 'iap' case in c2d text

amygdala:
every time I see this label in the webapp, the sans-serif font makes me think it's 'lap' with an L.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3038)


amygdala:
Just a ping on this.  It is an extremely trivial change.

PR #3039, opened 2019-04-17
Update profile readme according to golang profile controller.

stpabhi:

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3039)

### DIFF
--- a/kubeflow/profiles/prototypes/README.md
+++ b/kubeflow/profiles/prototypes/README.md
@@ -54,14 +54,13 @@ For __stan__ the RoleBinding may look the following
 
 ```yaml
 apiVersion: rbac.authorization.k8s.io/v1
-kind: RoleBinding
+kind: ClusterRoleBinding
 metadata:
   name: stan
-  namespace: kubeflow
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: Role
-  name: view
+  kind: ClusterRole
+  name: profiles-view
 subjects:
 - kind: ServiceAccount
   name: stan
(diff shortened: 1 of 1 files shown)


kubeflow/profiles/prototypes/README.md
```diff
@@ -54,14 +54,13 @@ For __stan__ the RoleBinding may look the following
 
 ```yaml
-kind: RoleBinding
+kind: ClusterRoleBinding
```
kkasravi: I think these changes in the README.md should be paired with code changes in the profiler-controller.

I don't think this will work since an individual will need to have cluster level access, the idea is that each individual's RBAC would be scoped to the namespace.

stpabhi: These bindings are not created by user or controller. Cluster admin should create a view clusterrole and cluster rolebindings for profiles resource to even allow the user to create, get profiles. Otherwise user can’t create profiles without rbac. As profile CR is cluster scoped, it has to bind with cluster role and cluster rolebinding.

kkasravi: These 2 fixed resources: ClusterRole and ClusterRoleBinding - would users be added to ClusterRoleBinding as subjects?

The Profile CR doesn't need to be cluster scoped, it could be submitted to the kubeflow namespace and the controller would create the new namespace. This would avoid granting the user any cluster level rights.

stpabhi: > These 2 fixed resources: ClusterRole and ClusterRoleBinding - would users be added to ClusterRoleBinding as subjects?
> Yes, either groups or service account or User
> The Profile CR doesn't need to be cluster scoped, it could be submitted to the kubeflow namespace and the controller would create the new namespace. This would avoid granting the user any cluster level rights.

Recently, Profile CR is changed to cluster scope from Namespaced [https://github.com/kubeflow/kubeflow/blob/master/kubeflow/profiles/profiles.libsonnet#L16](here) and that's the reason we can't do Role and RoleBindings anymore.

kkasravi: thanks for the explanation - This might be a bug or something specific to the golang implementation of the profile_controller. @lluunn do you recall why the Profile CR had to have a scope of cluster?

stpabhi:
@kkasravi is it good to merge?

stpabhi:
@kkasravi thanks. Could you also add the appropriate labels?

stpabhi:
@kkasravi could you please add approve label?

Issue #2712, opened 2019-03-15
click-to-deploy and kfctl should share code for shared logic.


kunmingg:
We have gcp_util / k8s_utils / ksonnet_util / kustomize_util which can be shared between click-to-deploy and kfctl.

click-to-deploy API rewrite and kustomize migration could be 2 chances to help with this effort.

jlewi:
At a minimum to deploy 0.5 do we need to rewrite the handler for "/e2edeploy" to use the kfctl library so that the logic is the same?

Could we add support for this and rollout it out to prod behind a query argument so that we could test it out before making it the default? And then change the default so that users will use the new logic but can fall back to the old logic if necessary.

kunmingg:
TODOs:
- https://github.com/kubeflow/kubeflow/issues/2610
- refactor kfctl interface to fit use case of app backend.
- app backend: add execute path using libs imported from kfctl.
- kfctl:
  1. Monitoring integration
  2. retry wrapper around API calls

jlewi:
@kunmingg Any update? Looks like we are waiting on #3045.

kunmingg:
Will try to merge #3045 early this week.

jlewi:
I believe the updated click to deploy is in staging and we just need to wait for it to be in prod before we can close this issue.

jlewi:
Closing this because the new refactored backend which uses the same logic is now in prod.

Issue #2152, opened 2018-12-21
[gcp-click-to-deploy] isolate each request into it's own container.


kunmingg:
Deploy service backend should have one container per request to process all actions.
This provide container level isolation. 
Each container should exit after all actions successfully done.

jlewi:
/assign @jlewi 

I will try to make some progress on this in the next sprint.

jlewi:
Bump to P0 because this is blocking deploying 0.6 with click to deploy.

jlewi:
A number of PRs have been committed; #3534 has a draft PR with a bunch of the ongoing development work.

The next steps are

1. Refactor gcp.go to fetch the information from KfDef spec

    *  Make KfDef a member of Gcp as opposed to embedding it
    * Add backfill logic for the Gcp plugin
    * Add methods that allow the web app to set the token source

1. Modify kustomize.go

    * Provide a method for injecting the RestConfig
    * Change the methods for building the RestConfig to allow injecting the token source

jlewi:
Making steady progress on this.
I have working code in #3534 ; I'm in the process of splitting that up into individual PRs and getting them merged.

Currently waiting on #3594 after which can send out more PRs.

We don't plan on blocking releasing 0.6.X on this issue.

This issue is still P0 though because people won't be to deploy 0.6 with click to deploy until its fixed.

jlewi:
@kunmingg and @gabrielwen will work on getting  the updated backend deployed this week.

/assign @gabrielwen @kunmingg

gabrielwen:
Now we have deployed new backend to staging and taking traffic.

@kunmingg is this issue considered resolved?

jlewi:
@gabrielwen Lets close this once its in production.

jlewi:
I believe it is in staging and we are just waiting on adding garbage collection for old statefulsets before deploying in production.

jlewi:
This has been deployed in staging and production so marking this as fixed.

PR #3045, opened 2019-04-18
add kfctl to deploy app backend

kunmingg:
For: https://github.com/kubeflow/kubeflow/issues/2712

* Deploy app backend: add execute path using libs imported from kfctl.
  * User can choose kfctl path by setting url params: ```kfctl=true```
* Unify request body type between deploy app and kfctl.
* make deployment manager operations run concurrently.
* deprecate cloud shell utils as user experience is not smooth.
* Unify Istio util functions from deploy app and kfctl.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3045)


bootstrap/cmd/bootstrap/app/kfctlUtil.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return err
+	}
+	ts := oauth2.StaticTokenSource(&oauth2.Token{
+		AccessToken: req.Token,
+	})
+	ctx := context.Background()
+
+	// run gcp generate / apply
+	gcpApp := gcp.BuildKfApp(kfdef, oauth2.NewClient(ctx, ts), ts)
```
kkasravi: it might be easier to use coordinator which wraps the platform and package managers within its Generate and Apply. That way it would also switch package managers based on whats in the [KfDefSpec](https://github.com/kkasravi/kubeflow/blob/kfctl_switch_packagemanagers/bootstrap/pkg/apis/apps/kfdef/v1alpha1/application_types.go#L39)

kunmingg: Agree, the only problem is now in coordinator
```platform.apply()``` and ```pkgManager.apply()``` are executed together and no space to insert credentials between them.

How about change coordinator order to be:
```
platform.generate()
platform.apply()
pkgManager.generate()
pkgManager.apply()
```
?

Comment (kkasravi) [reply bootstrap/cmd/bootstrap/app/kfctlUtil.go]: then coordinator would not implement KfApp, I think that may have a number of impacts.

It sounds like platform.Apply get credentials that PkgManager.Apply needs?
What if
-  Coordinator has a platformPostHook callback so we don't surface GCP in Coordinator?
or
- the platform is passed to the package manager and the package manager can call the platforms applyCallback?

**Comment by kunmingg** [comment]
@jlewi @gabrielwen 
Ready for review
@kkasravi 
Shall we unify Coordinator in future PRs?

File: bootstrap/cmd/bootstrap/app/kfctlUtil.go (hunk 1 above)
Comment (jlewi) [reply bootstrap/cmd/bootstrap/app/kfctlUtil.go]: @kunmingg @kkasravi Don't Apply and Generate take an enum specifying what phase to run e.g. platform and k8s so can't you do

coordinator.generate(platform)
coordinator.apply(platform)
coordinator.generate(k8s)
coordinator.apply(k8s)

Would this hide the details of the package manager? It would be nice if the app didn't have to change when the coordinator logic changed; e.g. to kustomize.

**Comment by kunmingg** [comment]
updated and rebased

File: bootstrap/cmd/bootstrap/app/kfctlUtil.go [hunk 2]
```diff
@@ -0,0 +1,112 @@
...
+
+	var gcpApp kftypes.KfApp
+	// run gcp generate / apply
+	if kfdef.Spec.UseBasicAuth {
+		UsernameData, err := base64.StdEncoding.DecodeString(req.Username)
+		if err != nil {
+			log.Errorf("Failed decoding username: %v", err)
+			return err
+		}
+		gcpApp = gcp.BuildKfApp(kfdef, oauth2.NewClient(ctx, ts), ts, req.StorageOption, req.SAClientID,
```
kkasravi: i'm wondering if we can use coordinator at this point? Especially since it would allow us to go directly to kustomize rather than ksonnet?

bootstrap/cmd/bootstrap/app/kfctlUtil.go
```diff
@@ -0,0 +1,112 @@
...
+			string(UsernameData), req.PasswordHash, "", "")
+	} else {
+		gcpApp = gcp.BuildKfApp(kfdef, oauth2.NewClient(ctx, ts), ts, req.StorageOption, req.SAClientID,
+			"", "", req.ClientID, req.ClientSecret)
+	}
+	
+	if err = gcpApp.Generate(kftypes.ALL); err != nil {
```
kkasravi: is it required to do gcp.Generate, Apply then ksonnet.Generate, Apply?
If coordinator is used it would be coordinator calling gcp.Generate, ksonnet/kustomize.Generate, gcp.Apply, ksonnet/kustomize.Apply

bootstrap/cmd/bootstrap/app/k8sAuth.go
```diff
@@ -41,6 +42,32 @@ func buildClusterConfig(ctx context.Context, token string, project string, zone
 	}, nil
 }
 
+// BuildClientCmdAPI takeks k8s config and access token, build and return clientcmdapi.Config entry
```
kkasravi: kustomize uses kubernetes-v1.13.4, client-go-v10.0.0 - is there anything (other than ksonnet) that prevents bootstrap from using the the later APIs (I know that client-go requires some changes)?

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -40,23 +41,19 @@ import (
 	type_v1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
```
kkasravi: this would be ks8s.io/client-go/v2/tools/clientcmd/api until we can remove the /tmp/v2 code - which would be after ksonnet is totally removed.

bootstrap/cmd/bootstrap/app/ksServer.go
```diff
@@ -411,19 +396,19 @@ func (s *ksServer) InstallIstio(ctx context.Context, req CreateRequest) error {
...
 		log.Errorf("Failed to create istio CRD: %v", err)
 		return err
 	}
-	err = CreateResourceFromFile(config, path.Join(regPath, "../dependencies/istio/install/istio-noauth.yaml"))
+	err = utils.CreateResourceFromFile(config, path.Join(regPath, "../dependencies/istio/install/istio-noauth.yaml"))
 	if err != nil {
 		log.Errorf("Failed to create istio manifest: %v", err)
 		return err
 	}
 	//TODO should be a cli parameter
```
kkasravi: this has been eliminated in a later version of kustomize that would be a follow-on PR after yours. These resources are now in manifests.

bootstrap/pkg/utils/k8utils.go
```diff
@@ -234,57 +235,38 @@ func patchOrCreate(mapping *meta.RESTMapping, config *rest.Config, group string,
...
-		return &kfapis.KfError{
-			Code:    int(kfapis.INVALID_ARGUMENT),
-			Message: err.Error(),
-		}
+		return err
 	}
-	objects := bytes.Split(data, []byte(bootstrap.YamlSeparator))
+	splitter := regexp.MustCompile(YamlSeparator)
+	objects := splitter.Split(string(data), -1)
 	var o map[string]interface{}
```
kkasravi: this needs to move inside the for loop otherwise we're merging multiple objects

jlewi:
@kunmingg Our goal in #2152 is to handle a single deployment per statefulset. If we assume that each deployment will be run in its own statefulset; does that make things easier?

IIRC correctly one of the current issues with invoking kfctl as a library is that it is modifying the app.yaml file stored on disk.

Combined with #2152  I think it might be ok and we will actually want to persist the Kubeflow app on disk. Each statefulset would have its own PVC and ensure isolation.

kunmingg:
Sounds good, will make assumption that during coordinator steps each thread will have isolated env.

jlewi:
Thanks @kunmingg

kunmingg:
update
it now make assumption that each request will be processed in isolated env.

jlewi:
Thanks @kunmingg A couple things

1. Could you update the PR description to call out the changes needed to make ClickToDeploy use kfctl as a go library

   * e.g. it looks like you are defining a new interface GcpPlatform could you explain what that's for in the PR description?

1. I think this is almost ready but I had some questions about what various code is doing PTAL at my comments in reviewable and maybe we can chat in person as well.

1. Just a reminder for the future; splitting unrelated changes into multiple PRs really helps the review process go faster
    * e.g. the cloud shell and DM changes could probably each have been their own PRs.

jlewi:
/lgtm
/approve

I think this is ready to be merged. I think any additional work can be folded into the effort to run each deployment in its own statefulset.

Thanks @kunmingg

jlewi:
@kunmingg Tests are failing because of 

```
{"name":"studyjob-controller","group":"extensions","kind":"deployments"},"code":404}
```

I think you just need to rebase on master.

kunmingg:
updated for tests.

PR #2984, opened 2019-04-10
 e2e test for kfctl istio

lluunn:
/cc @gabrielwen @kunmingg 

fix #2891

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/2984)

PR #3054, opened 2019-04-19
e2e test for kfctl istio

### LINKED ISSUE #2891: kfctl e2e test for istio

### LINKED ISSUE #3049: kfctl install istio fails with GOOGLE_APPLICATION_CREDENTIALS
Just like what we do in [test](https://github.com/kubeflow/kubeflow/blob/master/testing/kfctl/kfctl_go_test.py): 
1. use GOOGLE_APPLICATIOB_CREDENTIALS to point to a service account
1. gcloud activate that
1. set email to that service account

Installing istio failed to create clusterrole.
However, there is already a binding for service account and cluster admin.

Maybe it's because when installing istio we get another identity / wrong way getting credential?

lluunn:
fork from #2984

Fix #3049 
Fix #2891 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3054)


lluunn:
test passed!
/assign @gabrielwen @kunmingg

testing/kfctl/conftest.py
```diff
@@ -21,6 +21,10 @@ def pytest_addoption(parser):
       "--use_basic_auth", action="store", default="False",
       help="Use basic auth.")
 
+  parser.addoption(
+      "--use_istio", action="store", default="False",
```
gabrielwen: shouldn't you use something like this?
https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/get_kf_testing_cluster.py#L229-L237

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -664,23 +664,29 @@ func (gcp *Gcp) updateDM(resources kftypes.ResourceEnum) error {
...
-		return &kfapis.KfError{
-			Code: err.(*kfapis.KfError).Code,
-			Message: fmt.Sprintf("Build ClientConfig error: %v",
-				err.(*kfapis.KfError).Message),
+		if _, err := os.Stat(kftypes.KubeConfigPath()); !os.IsNotExist(err) {
+			gcp.AddNamedContext()
 		}
 	}
+	// Get client for kube config.
+	client := kftypes.GetConfig()
```
gabrielwen: could you see if this method can be used?
https://github.com/kubeflow/kubeflow/blob/de1fc808b40f58e4bbb7199fedb68df29d412c98/bootstrap/pkg/utils/gcputils.go#L80

ideally we should isolate between GCP and k8s.

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -664,23 +664,29 @@ func (gcp *Gcp) updateDM(resources kftypes.ResourceEnum) error {
...
-		gcp.Spec.Zone, gcp.Name, gcp.tokenSource)
-	if err != nil {
-		return &kfapis.KfError{
-			Code: err.(*kfapis.KfError).Code,
-			Message: fmt.Sprintf("Configure K8s is failed: %v",
-				err.(*kfapis.KfError).Message),
+	// kfctl only
+	// Setup kube config
+	if gcp.isCLI {
+		// TODO(#2604): Need to create a named context.
```
gabrielwen: could you help remove this line?

bootstrap/pkg/kfapp/gcp/gcp.go
lluunn: That creates clientcmdapi.Config, but we need rest.Config

testing/kfctl/conftest.py
lluunn: it's a different module called pytest. See above for basic auth

bootstrap/pkg/kfapp/gcp/gcp.go
gabrielwen: then it's this one:
https://github.com/kubeflow/kubeflow/blob/de1fc808b40f58e4bbb7199fedb68df29d412c98/bootstrap/pkg/utils/gcputils.go#L60

lluunn:
failure on kfctl.sh test

/retest

PR #3059, opened 2019-04-19
clusterDomain params adding to katib

### LINKED ISSUE #3046: host of istio route rule uses hardcode cluster-domain
I notice that host of istio route rule for multiple services of kubeflow component is like $servicename + $namespace + ".svc.cluster.local" in multiple PR, such as #3013 #3028 #3033 #3036

However `.cluster.local` can be set as a different value for k8s cluster with `cluster-domain` option.
I think $servicename + $namespace + ".svc" is enough.

@gabrielwen WDYT?

gabrielwen:
Fixes #3046

For other components, I've already added params to them.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3059)

### DIFF
--- a/kubeflow/katib/prototypes/all.jsonnet
+++ b/kubeflow/katib/prototypes/all.jsonnet
@@ -14,6 +14,7 @@
 // @optionalParam studyJobControllerImage string gcr.io/kubeflow-images-public/katib/studyjob-controller:v0.1.2-alpha-156-g4ab3dbd The image for studyjob-controller.
 // @optionalParam metricsCollectorImage string gcr.io/kubeflow-images-public/katib/metrics-collector:v0.1.2-alpha-156-g4ab3dbd The image for metrics-collector.
 // @optionalParam injectIstio string false Whether to inject istio sidecar; should be true or false.
+// @optionalParam clusterDomain string svc.cluster.local DNS config to cluster domain.
 
 local k = import "k.libsonnet";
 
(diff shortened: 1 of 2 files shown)


kubeflow/katib/prototypes/all.jsonnet
```diff
@@ -14,6 +14,7 @@
 // @optionalParam studyJobControllerImage string gcr.io/kubeflow-images-public/katib/studyjob-controller:v0.1.2-alpha-156-g4ab3dbd The image for studyjob-controller.
 // @optionalParam metricsCollectorImage string gcr.io/kubeflow-images-public/katib/metrics-collector:v0.1.2-alpha-156-g4ab3dbd The image for metrics-collector.
 // @optionalParam injectIstio string false Whether to inject istio sidecar; should be true or false.
+// @optionalParam clusterDomain string svc.cluster.local DNS config to cluster domain.
```
hougangliu: In fact, clusterDomain should be "cluster.local"-like without svc.
k8s composes a service full dns name with $servicename.$servicenamespace.svc.$clusterDomain.

gabrielwen: fixed.

Kyrremann:
May not be the correct place, but how am I supposed to use this new `customDomain` variable? I need `ml-pipeline` to use `custom.local` in stead of `cluster.local`.

Issue #2630, opened 2019-03-05
kfctl kustomize test suite (unit and e2e)


kkasravi:
the kustomize components in the repo kubeflow/manifests should have similar unit, e2e tests as ksonnet.

swiftdiaries:
/assign @swiftdiaries 
I'll pick this one up :)

jlewi:
@swiftdiaries How's it going with the testing? Do we have an E2E test that could be used to verify that deploying with kustomize is working?

swiftdiaries:
I'm not sure about the approach for unit testing components. 
I'm working on kfctl kustomize E2E test, I should have something before EOD or early tomorrow

jlewi:
@swiftdiaries How's this coming?  Looks like you have #3060 in progress?

swiftdiaries:
#3060 is for the E2E test. Now that #2971 is merged, I'll go ahead and merge that.
For the unit tests of manifests, I'm currently exploring the kustomize test harness. [Reference](https://github.com/kubernetes-sigs/kustomize/tree/master/pkg/target).

I'm having a bit of trouble because the implementation is split between two packages inside the same directory. I'm not sure how to import them for use inside Kubeflow. Any advice on how to do that?

jlewi:
@swiftdiaries unfortunately you are probably the expert; so I don't have much help to offer. Were you able to make any progress?

kkasravi:
@swiftdiaries I just [copied](https://github.com/kubeflow/manifests/pull/34/files#diff-ba78b29234976b9867e1c7aaaf216b8d) it into manifests/tests including the loadertest .

I also autogen'd the tests with a bash script -

jlewi:
@swiftdiaries What's the current status?

* Do we have an ETA for an E2E test?
* What are the blockers if any?

jlewi:
@swiftdiaries Any update I think @kkasravi committed the fixes to make kustomize +  ISTIO work on master. Can we get #3060 updated and merged.

swiftdiaries:
Cool thanks!
I'll get it updated and push it before tomorrow :)

jlewi:
According to @kkasravi #3375 pulled in @swiftdiaries changes in #3060.

Lets verify that we have a postsubmit test that is running and passing using ISTIO and kustomize and then close this issue.

jlewi:
kfctl_go_test.py is setting the package manager to kustomize here
[...]

PR #3060, opened 2019-04-19
kfctl go binary - kustomize e2e test

swiftdiaries:
Addresses: https://github.com/kubeflow/kubeflow/issues/2630

Adding a WIP label because this needs PR #2971 to go through.
I tried the test set of commands on a baremetal cluster. But it should be fine for the test cluster as well.

/cc @jlewi @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3060)


jlewi:
Thanks @swiftdiaries it looks like there are a lot of changes to the go code that don't look related to testing. Do you need to rebase to pick up  #2971 or are the changes intentional?

swiftdiaries:
Thanks for the review !
I've rebased now that #2971 went through. 
I followed https://github.com/kubeflow/kubeflow/pull/3054/ to implement something similar for package manager. 
PTAL.

prow_config.yaml
```diff
@@ -81,6 +81,25 @@ workflows:
       workflowName: kfctl-go
       useBasicAuth: true  
       useIstio: false
+  # Run kustomize package manager test
+  - app_dir: kubeflow/kubeflow/testing/workflows
+    component: kfctl_go_test
+    name: kfctl-go-basic-auth
```
jlewi: Is the name accurate? Should this be kfctl-go-kustomize?

testing/kfctl/conftest.py
```diff
@@ -24,6 +24,10 @@ def pytest_addoption(parser):
   parser.addoption(
       "--use_istio", action="store", default="False",
       help="Use istio.")
+  
+  parser.addoption(
+      "--package_manager", action="store", default="False",
```
jlewi: Nit: Give this flag a more descriptive name; e.g. use_kustomize

ksonnet and kustomize are both package managers.

testing/kfctl/kfctl_go_test.py
```diff
@@ -78,6 +78,9 @@ def test_build_kfctl_go(app_path, project, use_basic_auth, use_istio):
   else:
     init_args.append("--use_istio=false")
 
+  if package_manager:
+    init_args.append("--package-manager kustomize")
```
jlewi: If this is a string why are you taking a boolean as an argument to the test? Why don't you just pass through the value directly?

swiftdiaries: Agreed. I thought since we had just the two to choose from (ksonnet and kustomize) I'd use a boolean

testing/kfctl/conftest.py
swiftdiaries: Okay, will do

swiftdiaries: But since we're using the same flag for kfctl is better to keep it the same here?

swiftdiaries:
___
*[testing/kfctl/conftest.py, line 29 at r3](https://reviewable.io/reviews/kubeflow/kubeflow/3060#-LdB-388fV-InNZFo8su-r3-29:-LdQF-R89Y37otYZua4u:b-jkz56z) ([raw file](https://github.com/kubeflow/kubeflow/blob/7d2dc2ccbf4439b3b5f48ae5fe83722057895f84/testing/kfctl/conftest.py#L29)):*

If your going to use the same name as the kfctl flag then we should just plumb the value through; so if kfctl takes a string then we should take a string here and then just pass that value along.
</blockquote></details>

Updated to use a string instead of boolean.

swiftdiaries:
___
*[testing/kfctl/kfctl_go_test.py, line 82 at r3](https://reviewable.io/reviews/kubeflow/kubeflow/3060#-LdB-388fV-InNZFo8sw-r3-82:-LdfmysZ715epO5Mmi8U:bb6un1j) ([raw file](https://github.com/kubeflow/kubeflow/blob/7d2dc2ccbf4439b3b5f48ae5fe83722057895f84/testing/kfctl/kfctl_go_test.py#L82)):*

Why do you need the if statement? Can't we just always pass the value of the package_manager through?
</blockquote></details>

Done

swiftdiaries:
___
*[testing/kfctl/kfctl_go_kustomize_test.py, line 30 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/3060#-Ld7C3X6BLis6EthK8ka:-LdfnAlXDTL4KNhUpLZ1:b-896fix) ([raw file](https://github.com/kubeflow/kubeflow/blob/4e0dbb3afb35af9d10c564e9d708f4eb26e6fa91/testing/kfctl/kfctl_go_kustomize_test.py#L30)):*

It looks like there is a lot of code duplication with:
https://github.com/kubeflow/kubeflow/blob/master/testing/kfctl/kfctl_go_test.py

Could we just reuse that file and add a parameter to control whether or not we use kustomize?
</blockquote></details>

Done.

swiftdiaries:
___
*[prow_config.yaml, line 102 at r4](https://reviewable.io/reviews/kubeflow/kubeflow/3060#-LddcAhhEbwMgTeibloP:-LdfnEwXEmhh6wI3gmLC:b-uz1ryg) ([raw file](https://github.com/kubeflow/kubeflow/blob/00e17701e955b707b25c96fbfbd77eb9498baa18/prow_config.yaml#L102)):*

Could we enable ISTIO? For 0.6 we want to begin testing with ISTIO and kustomize as soon as possible.
ISTIO+kustomize is the combo we care about.

Will the test pass if enable ISTIO?
</blockquote></details>

Enabled ISTIO. But the tests seem to fail. We haven't migrated ISTIO yet. I haven't tried out ISTIO with kustomize. Let me try that to see if it works.

jlewi:
/lgtm
/approve

@swiftdiaries Any idea why the test are failing?

jlewi:
@swiftdiaries Do you know why the test is failing? Is this because of #3207

swiftdiaries:
We're currently not installing Istio and basic-auth as part of kustomize. The test needs both to succeed.

jlewi:
@swiftdiaries Do we have issues tracking that?

Don't we have separate tests for basic auth and for IAP. Could we enable kustomize for the IAP E2E test before we enable it on the basic auth test?

jlewi:
@swiftdiaries ping?

kkasravi:
Hi @swiftdiaries 

i might be missing something but it seems like this isn't changing test_deploy.py where we use `ks` to create a project and test deployments. To switch to kustomize we would want to create a project and test deployments using kfctl (preferably) or kustomize.

kkasravi:
doing a retest to see why its failing
/retest

jlewi:
test_deploy.py is outdated. I think we haven't deleted it because the outdated minikube tests are still using it.

kfctl_go_test.py is the newer test which uses the kfctl go binary.

jlewi:
ping?

@swiftdiaries Are you still working on this? @kkasravi Committed the changes to make kustomize + ISTIO work on master so it would be good to get a test committed that verifies that.

kkasravi:
@jlewi @swiftdiaries #3375 included Adhita's changes except for the [typo](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_go_test.jsonnet#L232) in workflows/components/kfctl_go_test.jsonnet

I opened PR #3435 - one character change. Please approve.

kkasravi:
this PR can be closed

jlewi:
Thanks @kkasravi

PR #3022, opened 2019-04-15
WIP: Adds a doc on how to create notebooks in user defined name…

zabbasi:
Related to #3017
Needs #3000 to resolve first.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3022)

PR #3074, opened 2019-04-22
Update jupyter ui to allow creating notebook in all available namespces

### LINKED ISSUE #3000: Jupyter UI should allow creating notebooks in all namespaces

zabbasi:
Undo #2951
Fix #3000 
Needs testing
To unblock #3022 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3074)

//cc @kimwnasptd @avdaredevil

### DIFF
--- a/components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
+++ b/components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
@@ -13,7 +13,7 @@
     <select class="mdl-textfield__input" id="ns-select" name="ns">
       <!-- Fill with the namespaces the user can see -->
       {% for ns in namespaces %}
-        <option value="{{ ns }}">{{ ns }}</option>
+        <option value="{{ ns }}" {{ ns == 'kubeflow'|yesno:"selected" }}>{{ ns }}</option>
       {% endfor %}
     </select>
     <label class="mdl-textfield__label" for="ns">Select Namespace</label>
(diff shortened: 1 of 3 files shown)


components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
```diff
@@ -13,7 +13,7 @@
     <select class="mdl-textfield__input" id="ns-select" name="ns">
       <!-- Fill with the namespaces the user can see -->
       {% for ns in namespaces %}
-        <option value="{{ ns }}">{{ ns }}</option>
+        <option value="{{ ns }}" {{ ns == 'kubeflow'|yesno:"selected" }}>{{ ns }}</option>
```
avdaredevil: This cherry-pick is from a PR that had a syntax error.


You can use:
```suggestion
        <option value="{{ ns }}" {% if ns == 'kubeflow' %}selected{% endif %}>{{ ns }}</option>
```

zabbasi: 👍

zabbasi:
duplicated to #3019

Issue #2086, opened 2018-12-13
Being able to deploy Kubeflow in private GKE cluster


dsdinter:
At the moment it is not possible to deploy Kubeflow in a private cluster as some images are pulled from remote registries (e.g., Dockerhub, quay.io). We should be able to somehow mirror these images in gcr.io to be able to pull them from private GKE clusters.
There is a [solution](https://cloud.google.com/container-registry/docs/using-dockerhub-mirroring) for Docker Hub public images but we should be able to support other public repositories as well.

dsdinter:
In the interim you can follow the below steps to deploy in a private GKE cluster including the ML Pipeline component (Thanks @IronPan!):
```shell
### Set ${KUBEFLOW_SRC} path and GCP ${PROJECT} with your own values

mkdir ${KUBEFLOW_SRC}
cd ${KUBEFLOW_SRC}
export KUBEFLOW_TAG=v0.3.4
curl https://raw.githubusercontent.com/kubeflow/kubeflow/${KUBEFLOW_TAG}/scripts/download.sh | bash

KFAPP=kubeflow
${KUBEFLOW_SRC}/scripts/kfctl.sh init ${KFAPP} --platform gcp --project ${PROJECT}
cd ${KFAPP}
${KUBEFLOW_SRC}/scripts/kfctl.sh generate platform

##### Set private cluster #####

${KUBEFLOW_SRC}/scripts/kfctl.sh apply platform
${KUBEFLOW_SRC}/scripts/kfctl.sh generate k8s
```
[Create](https://www.kubeflow.org/docs/started/getting-started-gke/#create-oauth-client-credentials) an OAuth client ID
```shell
export CLIENT_ID=<CLIENT_ID from OAuth page>
export CLIENT_SECRET=<CLIENT_SECRET from OAuth page>
ks param set ambassador statsdImage gcr.io/ml-pipeline/kubeflow/statsd:0.37.0
ks param set ambassador ambassadorImage gcr.io/ml-pipeline/kubeflow/ambassador:0.37.0
ks param set argo workflowControllerImage gcr.io/ml-pipeline/kubeflow/workflow-controller:v2.2.0
sed -i '' "s,mysql:5.6,gcr.io/ml-pipeline/kubeflow/mysql:5.6," vendor/kubeflow/pipeline/mysql.libsonnet
[...]

PR #3080, opened 2019-04-23
Scripts for replicating Docker images using GCB to support private GKE and VPC service controls.

jlewi:
* With private GKE we can't pull docker images from non-GCR registries (e.g. quoay)

* To support private GKE clusters we want to make it easy for users to
  mirror Kubeflow images to their own registry

* We create a GCB workflow to retag Kubeflow images

  * Using GCB is advantageous because it avoids pulling the networks over
    the user's network.

* Update the script to update the Kubeflow components to use the images in the
  user's registry.

* The kubeflow cluster isn't fully accessible yet.

  looks like pipelines and some other components still have images that
  need to be ported over.

Related to #2086

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3080)


scripts/gke/use_gcr_for_all_images.sh
```diff
@@ -3,37 +3,70 @@
...
+
+  if ks component list | awk '{print $1}' | grep -q "^katib$"; then
+    ks param set katib vizierDbImage ${registry}/mysql:8.0.3
+  fi
+
+  if ks component list | awk '{print $1}' | grep -q "^metacontroller$"; then
+    ks param set metacontroller image ${registry}/metacontroller:v0.3.0
+  fi
+}
+
```
IronPan: do you mind update pipeline's minio and mysql images 
https://github.com/kubeflow/kubeflow/blob/master/kubeflow/pipeline/pipeline.libsonnet#L9

nfs image can be ignored since it's not being used. 
Thanks!

jlewi: Done

PR #3084, opened 2019-04-23
Update aws OWNERS file and make ingress configurable

Jeffwan:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3084)

### DIFF
--- a/kubeflow/aws/OWNERS
+++ b/kubeflow/aws/OWNERS
@@ -1,3 +1,3 @@
 approvers:
   - jeffwan
-  - jlewi
+  - mameshini
--- a/scripts/aws/util.sh
+++ b/scripts/aws/util.sh
@@ -171,6 +171,11 @@ generate_aws_ks_app() {
 
   # Since JupyterHub will be removed evently, we skip authentication for it.
   popd
+
+  # write configuarion in generate phase and user can still configure it.
+  INGRESS_COMPONENT=',\\"aws-alb-ingress-controller\\",\\"istio-ingress\\"'
+  sed -i.bak "/^KUBEFLOW_COMPONENTS/ s/$/$INGRESS_COMPONENT/" ${KUBEFLOW_REPO}/${DEPLOYMENT_NAME}/${ENV_FILE}
+  rm ${KUBEFLOW_REPO}/${DEPLOYMENT_NAME}/${ENV_FILE}.bak
 }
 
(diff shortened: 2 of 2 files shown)


kubeflow/aws/OWNERS
```diff
@@ -1,3 +1,8 @@
 approvers:
   - jeffwan
-  - jlewi
\ No newline at end of file
+  - jlewi
```
jlewi: No-need to add these folks; You can actually go ahead and remove me as well. OWNERS are automatically inherited (unless you explicitly add the option to disable it). 
That ensures all the OWNERs at the lower levels can act as a backstop to approve PRs.

The primary purpose of adding finer grained OWNERs files is to better assign reviews to people who have the most context.

All of the folks other than you are Googlers with little understanding of AWS :) So adding them as approvers doesn't really help because all we can do is pretty much rubber stamp these PRs since we don't have much understanding of AWS.

So a better thing to do would be to try to grow the set of approvers and reviewers by getting more folks in the community using AWS involved. Then you can review and approve each other's PRs.

@mameshini and the folks have been working a lot with Kubeflow on AWS. Maybe you could add them as approvers or reviewres?

jlewi:
You need to join the Kubeflow org;
https://github.com/kubeflow/internal-acls/blob/master/github-orgs/kubeflow/org.yaml

I could have sworn I LGTM"d a PR from you adding you but I'm not seeing it and you aren't a member.

jlewi:
I see now; you did
https://github.com/kubeflow/community/pull/253

Which was the wrong the file. I should have looked closer.

kubeflow/aws/OWNERS
Jeffwan: Sounds great! Thanks for the details. I will clean it up and add @mameshini as approver.

Jeffwan:
Em. I will file another PR to add to member list. Thanks!

kubeflow/aws/OWNERS
Jeffwan: Clean it up and hope to attract more folks.

Jeffwan:
https://github.com/kubeflow/internal-acls/pull/86/ @jlewi  Here's member request PR. Thanks!

Jeffwan:
Sounds like a timeout issue. Let's try one more time.

Issue #2978, opened 2019-04-09
Make kubeflow use Istio instead of ambassador by default


kunmingg:
Include sub issues in https://github.com/orgs/kubeflow/projects/7

gabrielwen:
should we also add routing rules for all components with `getambassador.io/config`?  e.g.:
- tensorbaord
- kubebench-dashboard
- openvino
- echo-server
- centraldashboard
- basic-auth
- seldon
- tf-job-operator

kunmingg:
Definitely.
openvino should be part of pipeline migration effort.

@avdaredevil any concerns for migrating centraldashboard to istio?

avdaredevil:
Will there be any route level / behavioral changes in terms of navigation between embassador vs istio?

I didn't think there would be any issues on our side.

/cc @prodonjs any thoughts?

gabrielwen:
I don't think it'll be any different (and it shouldn't be).  @avdaredevil @prodonjs please help review #3037 when you have chance, thanks.

gabrielwen:
the to-do list is down to 2 items - basic auth and seldon.

jlewi:
Things are working with ksonnet but not kustomize. So we need to get it working with kustomize.

jlewi:
@gabrielwen @kunmingg What is the current status of this?

kunmingg:
kfctl now use istio by default.

PR #3087, opened 2019-04-23
Flip injectIstio param to true

gabrielwen:
Flip default to be true since now we make ISTIO as default.

Relates to #2978

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3087)


lluunn:
The test will fail because not all test are using istio.

gabrielwen:
Now the PR is merged with #3075 I think it should be good now.

kubeflow/katib/prototypes/all.jsonnet
```diff
@@ -13,14 +13,14 @@
 // @optionalParam vizierDbImage string mysql:8.0.3 The image for vizier db.
 // @optionalParam studyJobControllerImage string gcr.io/kubeflow-images-public/katib/studyjob-controller:v0.1.2-alpha-156-g4ab3dbd The image for studyjob-controller.
 // @optionalParam metricsCollectorImage string gcr.io/kubeflow-images-public/katib/metrics-collector:v0.1.2-alpha-156-g4ab3dbd The image for metrics-collector.
-// @optionalParam injectIstio string false Whether to inject istio sidecar; should be true or false.
+// @optionalParam injectIstio string true Whether to inject istio sidecar; should be true or false.
 // @optionalParam clusterDomain string cluster.local DNS config to cluster domain.
 
-local k = import "k.libsonnet";
```
lluunn: revert the quote change

gabrielwen: done.

scripts/util.sh
```diff
@@ -94,17 +94,19 @@ createKsApp() {
   ks generate openvino openvino
   ks generate jupyter jupyter
   ks generate notebook-controller notebook-controller
-  ks generate jupyter-web-app jupyter-web-app
-  ks generate centraldashboard centraldashboard
-  ks generate tf-job-operator tf-job-operator
-  ks generate tensorboard tensorboard
+  ks generate jupyter-web-app jupyter-web-app --injectIstio false
```
lluunn: this is used by kfctl.sh right?
Do you need to update anything for kfctl golang?

gabrielwen: no, because we stopped maintaining kfctl.sh, we don't have an option in kfctl.sh to install Istio.  this is why I disable `injectIstio` on all components.  golang is fine since we default to install Istio now.

PR #3090, opened 2019-04-23
Added autocomple support for Deploy Form

### LINKED ISSUE #2952: [GCP] Make autofill work with click to deploy
We'd like autofill to work with the click to deploy app.

In particular, we'd like web browsers to treat client id and client secret as username/password and store them securely while still allowing the user the convenience of auto-filling them when a user comes back to deploy.kubeflow.cloud

/assign @avdaredevil

avdaredevil:
## About
This change uses the [HTML Autofill Spec](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#autofill) and [TextField's API](https://material-ui.com/api/text-field/) to create an ARIA compliant autocomplete form for OAuth Client ID and Secret fields.

/area front-end
/priority p1
/assign @avdaredevil
/cc @prodonjs @jlewi @Ark-kun

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3090)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -210,11 +210,11 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
                 then enter as IAP Oauth Client ID and Secret</li>
           </div>
           <div style={styles.row}>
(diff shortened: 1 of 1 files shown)


prodonjs:
/assign @kunmingg 

for owners approval

PR #3102, opened 2019-04-24
Adds Option to Enable SHM for notebook containers

hno2:
This commit includes:
- A new Frontend Toggle to enable/disable shm
- A short ReadMe Section
- A new Backend Function enable_shm

I didn't include an option to limit the size of the shared memory, as this [will not work](https://github.com/kubernetes/kubernetes/issues/28272#issuecomment-383290704).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3102)


hno2:
I signed it!

hno2:
@prodonjs  So what exactly is the difference between default and rok/ What does rok stand for? I could not find any info on that?
I can transfer my changes to rok, if needed

prodonjs:
I have the same questions as you have! I am relatively unfamiliar with the application details, but it looks like @kimwnasptd submitted most of the code here. Perhaps he could chime in?

kimwnasptd:
Hi @prodonjs @hno2, Rok UI is the one [miniKF](https://www.kubeflow.org/docs/started/getting-started-minikf/) uses. I think it would be better to rename it from `rok` to `minikf` to make it more clear.

The Rok/MiniKF UIs main differences from the default are
* it won't render the Notebook Table unless a user Rok secret is found
* The submit form has one extra field 
* Adds annotations to the created PVCs

I'm planning to rewrite both frontends in Angular so they can be easier to maintain and also factor out the common code. I will also add a readme that explains the differences between the UIs in detail.

For this and subsequent PRs, since someone needs to have miniKF to test the changes with the rok/minikf UI, its understandable to not update the minikf UI. I am also keeping an eye on changes of the default UI and will make a PR to update the minikf/rok one when necessary. 

So @hno2, since these changes will also work with the minikf ui, if you could also make another commit in this PR to update the `add_notebook.hmtl` and `app.py` of the rok/minikf ui that would be helpful. Else @prodonjs you can lgtm this and I will make a subsequent PR to also update the minikf ui.

Hope I clarified things a little bit.

hno2:
I just added my changes to rok, too! 
I guess,  you can take it from here @prodonjs ?

prodonjs:
Thanks so much for your clarification Kimonas! This looks good now.

/lgtm
/approve

prodonjs:
@kimwnasptd It looks like I can LGTM, but I'm not an owner so I think we still need your /approve comment to get this merged.

hno2:
The Bot ask for the approval of @ioandr or am I wrong. Im not sure about the process here.

prodonjs:
Not you are not wrong at all but both @ioandr and @kimwnasptd are listed in the OWNERS file and since Kimonas had already chimed in here I figured he may oblige with the final approval.

kimwnasptd:
Definitely!
/approve

brtknr:
Any plans to add this to a release?

Issue #2172, opened 2018-12-26
Jupyter CR should create ISTIO Virtual Service


jlewi:
See design doc: http://bit.ly/kf_jupyter_design_doc

We will use ISTIO to route traffic to individual jupyter notebooks.

To support this the Jupyter custom resource should create the ISTIO virtual service and define the appropriate route.

jlewi:
@gabrielwen Are you working on this? IIUC I think getting the Jupyter CR to create the virtual service rules to set up the reverse proxy routes is one of the last blockers to using ISTIO instead of Ambassador (#2978)?

gabrielwen:
/assign @gabrielwen 

yes: #3104

gabrielwen:
/close

#3104 creates ISTIO virtual services.

PR #3104, opened 2019-04-25
ISTIO support for notebook controller

gabrielwen:
Relates to #2978

Tested through manual deployment:
1. Remove ambassador from component list.
1. Bring up KF cluster.
1. Navigate to `Notebooks` and use the UI to create a new notebook server.
1. Verify that `CONNECT` could correctly find Jupyter notebook UI page.
1. Verify that when deleting a notebook server, virtual service is gone with it.

/assign @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3104)

### DIFF
--- a/components/notebook-controller/pkg/controller/notebook/notebook_controller.go
+++ b/components/notebook-controller/pkg/controller/notebook/notebook_controller.go
@@ -224,32 +224,21 @@ func (r *ReconcileNotebook) Reconcile(request reconcile.Request) (reconcile.Resu
 	if err := controllerutil.SetControllerReference(instance, virtualService, r.scheme); err != nil {
 		return reconcile.Result{}, err
 	}
-	// Check if the virtual service already exists
+	// Check if the virtual service already exists. Noop if it already exists.
 	foundVirtual := &unstructured.Unstructured{}
 	foundVirtual.SetAPIVersion("networking.istio.io/v1alpha3")
 	foundVirtual.SetKind("VirtualService")
-	justCreated = false
 	err = r.Get(context.TODO(), types.NamespacedName{Name: virtualServiceName(instance.Name,
 		instance.Namespace), Namespace: instance.Namespace}, foundVirtual)
 	if err != nil && errors.IsNotFound(err) {
 		log.Info("Creating virtual service", "namespace", instance.Namespace, "name",
 			virtualServiceName(instance.Name, instance.Namespace))
 		err = r.Create(context.TODO(), virtualService)
-		justCreated = true
 		if err != nil {
 			return reconcile.Result{}, err
 		}
 	} else if err != nil {
 		return reconcile.Result{}, err
-	}
-	// Override virtual service if needed.
-	if !justCreated && util.DiffUnstructured(foundVirtual, virtualService) {
-		log.Info("Overriding virtual service", "namespace", instance.Namespace, "name",
-			virtualServiceName(instance.Name, instance.Namespace))
-		err = r.Update(context.TODO(), virtualService)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
 	}
 
 	// Update the status if previous condition is not "Ready"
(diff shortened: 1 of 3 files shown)


components/notebook-controller/pkg/controller/notebook/notebook_controller.go
```diff
@@ -204,6 +214,12 @@ func (r *ReconcileNotebook) Reconcile(request reconcile.Request) (reconcile.Resu
 		if err != nil {
 			return reconcile.Result{}, err
 		}
+		log.Info("Creating virtual service", "namespace", service.Namespace, "name",
```
lluunn: shouldn't you do another check for virtualService by itself?
This if block is when you don't see the "service"?

kubeflow/jupyter/notebook_controller.libsonnet
```diff
@@ -1,32 +1,32 @@
 {
-  local util = import "kubeflow/common/util.libsonnet",
```
lluunn: we use double quote else where.
I think maybe your editor/formatter did this?
Could you revert the change here

lluunn:
Looks great.
Have you test this (create/ delete notebook, check controller logs)?
After testing, could you add it (what steps you did, what's the result) to PR description?
Thanks.

components/notebook-controller/pkg/controller/notebook/notebook_controller.go
gabrielwen: done.

kubeflow/jupyter/notebook_controller.libsonnet
gabrielwen: yea... it's `vim-jsonnet`...  reverted.

components/notebook-controller/pkg/controller/notebook/notebook_controller.go
```diff
@@ -216,6 +219,28 @@ func (r *ReconcileNotebook) Reconcile(request reconcile.Request) (reconcile.Resu
...
+	if err != nil && errors.IsNotFound(err) {
+		log.Info("Creating virtual service", "namespace", instance.Namespace, "name",
+			virtualServiceName(instance.Name, instance.Namespace))
+		err = r.Create(context.TODO(), virtualService)
+		if err != nil {
+			return reconcile.Result{}, err
+		}
+	} else if err != nil {
+		return reconcile.Result{}, err
+	}
```
lluunn: you need an update step here, see 213~220 for service, or 185~193 for ss

gabrielwen: doesn't seem to have a proper way of doing it - having this error when trying to run update:
`"error":"virtualservices.networking.istio.io \"notebook-kubeflow-nb-testing\" is invalid: metadata.resourceVersion: Invalid value: 0x0: must be specified for an update`

gabrielwen: added.

components/notebook-controller/pkg/util/util.go
```diff
@@ -79,3 +80,27 @@ func CopyServiceFields(from, to *corev1.Service) bool {
...
+	if !found {
+		return true
+	}
+	if err != nil {
+		return true
+	}
+
+	toSpec, found, err := unstructured.NestedMap(to.Object, "spec")
+	if !found {
+		return true
```
lluunn: we should not return here.
We should still SetNestedMap to copy spec from "from" to "to" right?

gabrielwen: done.

lluunn:
thank you! 
/lgtm
/approve

/hold
Could you test one more thing:
remove/edit the virtualService and make sure the reconcile recover the virtualService.
If that works, please cancel the hold

gabrielwen:
/hold cancel

watch event added and verified.

lluunn:
nice!
/lgtm

gabrielwen:
Fixes #2172

PR #3107, opened 2019-04-25
Python 2/3 agnostic dict iteration in update_ks_app.py

### LINKED ISSUE #3098: upgrade_ks_app.py is not python3 compatible
Since python2 is nearly EOL it would be better to have this script work with python3.

JohnPaton:
Fixes #3098, it turned out to be a tiny change. There is more info on iterating over dicts in a 2/3 agnostic way at [PEP469](https://www.python.org/dev/peps/pep-0469/). 

I wasn't sure whether to open this PR against the 0.5 branch or master so just let me know if I did the wrong thing!

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3107)

### DIFF
--- a/scripts/upgrade_ks_app.py
+++ b/scripts/upgrade_ks_app.py
@@ -57,7 +57,7 @@ def main():
   registries = app['registries']
   libraries = app['libraries']
 
-  for name in registries.iterkeys():
+  for name in registries.keys():
     if name != registry_name:
       logging.info("Skipping registry %s", name)
       continue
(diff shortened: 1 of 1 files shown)


scripts/upgrade_ks_app.py
```diff
@@ -57,7 +57,7 @@ def main():
   registries = app['registries']
   libraries = app['libraries']
 
-  for name in registries.iterkeys():
+  for name in registries.keys():
```
jlewi: nit: can you do 
for name, _ in registries.items()

I think most linters prefer that.

JohnPaton: Done!

PR #3108, opened 2019-04-25
Support for multiple overlays in kustomize

### LINKED ISSUE #3101: combining component manifest overlays
In general we can expect different kustomize targets to combine different types of overlays. 

Combining different overlays in an ad-hoc manner is possible in kustomize but should 
be done on the fly, possibly via a script in the manifests repo and using componentParameter settings in kfctl.

For profiles - the kustomization.yaml that combines base, overlays/debug and overlays/devices looks like the following:
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: yyy
bases:
- base
patchesStrategicMerge:
- overlays/devices/deployment.yaml
- overlays/debug/deployment.yaml
```

In general the algorithm is 
```
  for each overlay
    add the path to the overlay resource files under patchesStrategicMerge
```

Various approaches (and combination):
1. Use a kustomize plugin generator (see https://github.com/kubernetes-sigs/kustomize/issues/1010)
1. Add a script to the manifests repo to create a kustomization.yaml that combines overlays
1. Document how to do create a kustomization.yaml the combines overlays in the README
1. For kfctl change app.yaml to combine overlays as shown below for profiles:
```
...
  componentParams:
    profiles:
    - name: overlays
      value:
      - overlays/devices
      - overlays/debug
```

kfctl would generate a kustomization.yaml under `.cache/manifests/<version>` that includes the strategic merge of the overlays + base.
 
Note: kfctl generates a kustomization.yaml now under `.cache/manifests/<version>` when creating yaml output per component under `<deployment>/kustomize`.

kkasravi:
Paired with kubeflow/manifests/pull/31 and kubeflow/manifests/pull/34

Multiple overlays -

This PR adds the ability for kfctl to combine more than one overlay during `kfctl generate ...`. An example is shown below where the profiles target in [manifests](https://github.com/kkasravi/manifests/tree/multiple_overlays/profiles) can include either debug changes in the Deployment or Device information in the Namespace (the devices overlay is not fully integrated with the Profile-controller at this point in time and is intended as an example) or **both**. 

```
profiles
├── base
│   └── kustomization.yaml
└── overlays
    ├── debug
    │   └── kustomization.yaml
    └── devices
        └── kustomization.yaml
```

#### What are Multiple Overlays?

Normally kustomize provides the ability to overlay a 'base' set of resources with changes that are merged into the base from resources that are located under an overlays subdirectory. For example
if the kustomize [target](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/glossary.md#target) is named foo there will be a foo/base and possibly one or more overlays such as foo/overlays/bar. A kustomization.yaml file is found in both foo/base and foo/overlays/bar. Running `kustomize build` in foo/base will generate resources as defined in kustomization.yaml. Running `kustomize build` in foo/overlays/bar will generate resources - some of which will overlay the resources in foo/base.

Kustomize doesn't provide for an easy way to combine more than one overlay for example foo/overlays/bar, foo/overlays/baz. However this is an open feature request in kustomize [issues](https://github.com/kubernetes-sigs/kustomize/issues/759). The ability to combine more than one overlay is key to handling components like tf-job-operator which has several types of overlays that can 'mix-in' whether a TFJob is submitted to a namespace or cluster-wide and whether the TFJob uses gang-scheduling.

#### Merging multiple overlays

Since each overlay includes '../../base' as its base set of resources - combining several overlays where each includes '../../base' will cause `kustomize build` to abort, complaining that it recursed on base. The approach is to create a kustomization.yaml at the target level that includes base and the contents of each overlay's kustomization file. This requires some path corrections and some awareness of the behavior of configMapGenerator, secretMapGenerator and how they are copied from each overlay. This kustomization.yaml can be constructed manually, but is integrated within kfctl via the app.yaml file. Using tf-job-operator as an example, if its componentParams has the following
```
  componentParams:
    tf-job-operator:
    - name: overlay
       value: cluster
    - name: overlay
    - value: gangscheduled
```

Then the result will be to combine these overlays eg 'mixin' these overlays in the kustomization.yaml file under tf-job-operator.

#### Merging multiple overlays to generate app.yaml

Normally when `kfctl init ...` is called it will download the kubeflow repo under `<deployment>/.cache` and read one of the config files under `.cache/kubeflow/<version>/bootstrap/config`. These config files define packages, components and component parameters (among other things). Each config file is a compatible k8 resource of kind *KfDef*. The various config files are:
- kfctl_default.yaml
- kfctl_basic_auth.yaml
- kfctl_iap.yaml

Both kfctl_basic_auth.yaml and kfctl_iap.yaml contain the contents of kfctl_default.yaml plus some additional changes specific to using basic_auth when the cluster is created and platform specific resources if the platform is **gcp**. This PR corrects this redundancy by using kustomize to combine a **gcp** overlay and/or a **basic_auth** overlay. Additionally, due to pipeline refactoring, the kustomize package manager has split the bundled pipeline component in ksonnet to a set of individual pipeline targets. This results in the following:

```
config
├── base
│   └── kustomization.yaml
└── overlays
    ├── basic_auth
    │   └── kustomization.yaml
    ├── gcp
    │   └── kustomization.yaml
    ├── ksonnet
    │   └── kustomization.yaml
    └── kustomize
        └── kustomization.yaml
```

Based on the cli args to `kfctl init...`, the correct overlays will be merged to produce an app.yaml.
The original files have been left as is until UI integration can be completed in a separate PR (@kunmingg )
Fixes #3101

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3108)


bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -577,18 +578,55 @@ func (kustomize *kustomize) writeConfigFile() error {
...
 	comp := path.Join(kustomize.Spec.ManifestsRepo, compPath)
 	compKustomizationFile := filepath.Join(comp, kftypes.KustomizationFile)
 	base := path.Join(kustomize.Spec.ManifestsRepo, compPath, "base")
 	baseKustomizationFile := filepath.Join(base, kftypes.KustomizationFile)
-	platformOverlay := path.Join(kustomize.Spec.ManifestsRepo, compPath, "overlays", kustomize.Spec.Platform)
-	platformKustomizationFile := filepath.Join(platformOverlay, kftypes.KustomizationFile)
-	if _, err := os.Stat(platformKustomizationFile); err == nil {
-		bases = append(bases, extractSuffix(kustomize.Spec.ManifestsRepo, platformOverlay))
+	overlays := kustomize.overlayResources(compName, compPath)
+	if kustomize.Spec.Platform != "" && len(overlays) == 0 {
```
hougangliu: I think we should not ignore kustomize.Spec.Platform != ""  && len(overlays) >0 case.
for example, for minikube platform, by `overlays`, we can support if enable-gang-scheduling enabled or not, refer to [here](https://github.com/kubeflow/manifests/issues/27)

hougangliu:
@kkasravi for others' better understanding, I suggest to add an use case in this PR about multiple overlays feature :)

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -577,44 +572,126 @@ func (kustomize *kustomize) writeConfigFile() error {
...
+		}
+		if overlayKustomization.PatchesJson6902 != nil {
+			for _, value := range overlayKustomization.PatchesJson6902 {
+				patchJson := new(patch.Json6902)
+				patchJson.Target = value.Target
+				patchAbsolutePath := filepath.Join(overlayPath, value.Path)
+				patchJson.Path = extractSuffix(compDir, patchAbsolutePath)
+				kustomization.PatchesJson6902 = append(kustomization.PatchesJson6902, *patchJson)
+			}
+		}
```
hougangliu: `configMapGenerator`, `secretGenerator`,`crds` in overlay also creates k8s resources.
In addition, overlay resources may refer `vars`.

jlewi:
Can you provide more explanation in the PR description please? 

* Please explain what you mean by multiple overlays
* Please explain what needs to change to support multiple overlay

jlewi:
Thanks @kkasravi the description is very helpful. Is the kustomize community thinking about multiple overlays?

Does support for multiple overlays depend on kfctl? Would a user no longer be able to generate manifests for a specific application just by using kustomize?

kkasravi:
> Is the kustomize community thinking about multiple overlays?

yes, there have been a few requests/issues opened
- [Support for mixins](kubernetes-sigs/kustomize/issues/759)
- [Support kustomize spec mixins from annotation](kubernetes-sigs/kustomize/issues/942)
- [Extend patch from one target to multiple targets](kubernetes-sigs/kustomize/issues/720)

- There is also a KEP [Extending kustomize patches to patch multiple targets](https://github.com/kubernetes/enhancements/blob/4d9b6088d43d75850133f0df7910706aa7272377/keps/sig-cli/kustomize-extend-patch.md)
- Finally transformers and generator plugins have been added but this would require adding a script or .so generator. In lieu of my current approach (see below) I think this would be more complicated than needed.

> Does support for multiple overlays depend on kfctl? Would a user no longer be able to generate manifests for a specific application just by using kustomize?

kfctl will generate a kustomization.yaml that includes multiple overlays, but this could also be done manually. 

Below is tf-job-operator - showing directories and kustomization.yamls. (resources aren't shown).

```
tf-job-operator
├── base
│   └── kustomization.yaml
└── overlays
    ├── cluster
    │   └── kustomization.yaml
    ├── cluster-gangscheduled
    │   └── kustomization.yaml
    ├── namespaced
    │   └── kustomization.yaml
    └── namespaced-gangscheduled
        └── kustomization.yaml
```

kfctl will generate a kustomization.yaml for every component regardless of whether multiple overlays are requested or not. This kustomization.yaml is shown below

```
tf-job-operator
├── base
│   └── kustomization.yaml
├── kustomization.yaml
└── overlays
    ├── cluster
    │   └── kustomization.yaml
    ├── cluster-gangscheduled
    │   └── kustomization.yaml
    ├── namespaced
    │   └── kustomization.yaml
    └── namespaced-gangscheduled
        └── kustomization.yaml
```

This kustomization.yaml will have a top level namespace set and some other things that might be specified in app.yaml componentParams. If either base or a specific overlay is specified, that information is copied into the generated kustomization.yaml.

The approach is to use kustomize's API - `MakeEmptyAccumulator()`.
This will take child ResMap's (kustomization.yamls) and allow different types of merges (replace, override) to absorb the other kustomization.yamls while adjusting path references. I'm pretty far along using this approach.

jlewi:
Could a user use kustomize directly with our manifests without using kfctl?

For example if I just want to deploy TFJob operator could I 

1. Clone kubeflow/manifests
1. cd to tf-job-operator
1. Run kustomize build

kkasravi:
yes, though they would need to pick an overlay. eg
```
1. Clone kubeflow/manifests
1. cd manifests/tf-training/tf-job-operator/overlays/cluster
1. kustomize build
```

```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  labels:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  name: tfjobs.kubeflow.org
spec:
  additionalPrinterColumns:
  - JSONPath: .status.conditions[-1:].type
    name: State
    type: string
  - JSONPath: .metadata.creationTimestamp
    name: Age
    type: date
  group: kubeflow.org
  names:
    kind: TFJob
    plural: tfjobs
    singular: tfjob
  scope: Namespaced
  subresources:
    status: {}
  validation:
    openAPIV3Schema:
      properties:
        spec:
          properties:
            tfReplicaSpecs:
              properties:
                Chief:
                  properties:
                    replicas:
                      maximum: 1
                      minimum: 1
                      type: integer
                PS:
                  properties:
                    replicas:
                      minimum: 1
                      type: integer
                Worker:
                  properties:
                    replicas:
                      minimum: 1
                      type: integer
  version: v1beta1
  versions:
  - name: v1beta1
    served: true
    storage: true
  - name: v1beta2
    served: true
    storage: false
---
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  name: tf-job-operator-dashboard-service-account
  namespace: kubeflow
---
apiVersion: v1
kind: ServiceAccount
metadata:
  labels:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  name: tf-job-operator-service-account
  namespace: kubeflow
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  name: cluster-role
rules:
- apiGroups:
  - tensorflow.org
  - kubeflow.org
  resources:
  - tfjobs
  - tfjobs/status
  verbs:
  - '*'
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - '*'
- apiGroups:
  - storage.k8s.io
  resources:
  - storageclasses
  verbs:
  - '*'
- apiGroups:
  - batch
  resources:
  - jobs
  verbs:
  - '*'
- apiGroups:
  - ""
  resources:
  - configmaps
  - pods
  - services
  - endpoints
  - persistentvolumeclaims
  - events
  verbs:
  - '*'
- apiGroups:
  - apps
  - extensions
  resources:
  - deployments
  verbs:
  - '*'
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRole
metadata:
  name: dashboard-cluster-role
rules:
- apiGroups:
  - tensorflow.org
  - kubeflow.org
  resources:
  - tfjobs
  - tfjobs/status
  verbs:
  - '*'
- apiGroups:
  - apiextensions.k8s.io
  resources:
  - customresourcedefinitions
  verbs:
  - '*'
- apiGroups:
  - storage.k8s.io
  resources:
  - storageclasses
  verbs:
  - '*'
- apiGroups:
  - batch
  resources:
  - jobs
  verbs:
  - '*'
- apiGroups:
  - ""
  resources:
  - configmaps
  - pods
  - services
  - endpoints
  - persistentvolumeclaims
  - events
  - pods/log
  - namespaces
  verbs:
  - '*'
- apiGroups:
  - apps
  - extensions
  resources:
  - deployments
  verbs:
  - '*'
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: cluster-role-binding
roleRef:
  kind: ClusterRole
  name: cluster-role
subjects:
- kind: ServiceAccount
  name: service
---
apiVersion: rbac.authorization.k8s.io/v1beta1
kind: ClusterRoleBinding
metadata:
  name: dashboard-cluster-role-binding
roleRef:
  kind: ClusterRole
  name: dashboard-cluster-role
subjects:
- kind: ServiceAccount
  name: dashboard-service
---
apiVersion: v1
data:
  controller_config_file.yaml: |-
    {
        "grpcServerFilePath": "/opt/mlkube/grpc_tensorflow_server/grpc_tensorflow_server.py"
    }
kind: ConfigMap
metadata:
  labels:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  name: tf-job-operator-config-map
  namespace: kubeflow
---
apiVersion: v1
kind: Service
metadata:
  annotations:
    getambassador.io/config: |-
      ---
      apiVersion: ambassador/v0
      kind:  Mapping
      name: tfjobs-ui-mapping
      prefix: /tfjobs/
      rewrite: /tfjobs/
      service: tf-job-operator-dashboard.kubeflow
  labels:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  name: tf-job-operator-dashboard-service
  namespace: kubeflow
spec:
  ports:
  - port: 80
    targetPort: 8080
  selector:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  type: ClusterIP
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  name: tf-job-operator-dashboard-deployment
  namespace: kubeflow
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: tf-job-operator-application
      kustomize.component: tf-job-operator
  template:
    spec:
      containers:
      - command:
        - /opt/tensorflow_k8s/dashboard/backend
        env:
        - name: KUBEFLOW_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        image: gcr.io/kubeflow-images-public/tf_operator:v0.5.0
        name: dashboard-container
        ports:
        - containerPort: 8080
      serviceAccountName: dashboard-service
---
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: tf-job-operator-application
    kustomize.component: tf-job-operator
  name: tf-job-operator-deployment
  namespace: kubeflow
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: tf-job-operator-application
      kustomize.component: tf-job-operator
  template:
    metadata:
      labels:
        app.kubernetes.io/name: tf-job-operator-application
        kustomize.component: tf-job-operator
    spec:
      containers:
      - args:
        - --alsologtostderr
        - -v=1
        command:
        - /opt/kubeflow/tf-operator.v1beta2
        env:
        - name: KUBEFLOW_NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        image: gcr.io/kubeflow-images-public/tf_operator:v0.5.0
        name: container
        volumeMounts:
        - mountPath: /etc/config
          name: config-volume
      serviceAccountName: tf-job-operator-service-account
      volumes:
      - configMap:
          name: tf-job-operator-config-map
        name: config-volume
```

kkasravi:
We could make an argument that the base should always be deployable, though in this case if we defaulted to cluster for example in tf-job-operator/base and added cluster-role.yaml, cluster-role-binding.yaml to base, then if the user picked tf-job-operator/overlays/namespaced we would need to remove cluster-role - which isn't well supported right now in kustomize

jlewi:
Thanks @kkasravi. I'm satisfied. LGTM on my end.

jlewi:
/lgtm
/approve
/hold for @hougangliu 

@hougangliu can you take another look please and LGTM if it all looks good?

bootstrap/config/overlays/gcp/gcp_prototype.yaml
```diff
@@ -0,0 +1,59 @@
+# Sample config for kubeflow bootstrapper for v0.4
```
gabrielwen: do we have a type definition like this?
https://github.com/kubeflow/kubeflow/blob/3d56531ab8a35dc2663a6e764284051ed020ba34/bootstrap/config/types.go#L28

If we are gonna parse it in apps such as kfctl or bootstraper, could we define types as well?

bootstrap/config/overlays/basic_auth/kfctl_default-patch.yaml
```diff
@@ -0,0 +1,8 @@
+- op: add
```
gabrielwen: same thing here, is this used by apps?  if it is, could we also add types?

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
kkasravi: Updated

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -577,84 +572,392 @@ func (kustomize *kustomize) writeConfigFile() error {
...
+			}
+		case types.BehaviorMerge, types.BehaviorReplace, types.BehaviorUnspecified:
+			fallthrough
+		default:
+			parent.ConfigMapGenerator = append(parent.ConfigMapGenerator, *parentConfigMapArgs)
+			kustomizationMaps[configMapGeneratorMap][childConfigMapArgs.Name] = true
+		}
+	}
+
+	updateParamFiles()
```
hougangliu: if err := updateParamFiles(); err != nil {
    return err
}

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -577,84 +572,392 @@ func (kustomize *kustomize) writeConfigFile() error {
...
+		resourceAbsoluteFile := filepath.Join(targetDir, string(value))
+		resourceFile := extractSuffix(compDir, resourceAbsoluteFile)
+		if _, ok := kustomizationMaps[resourcesMap][resourceFile]; !ok {
+			parent.Resources = append(parent.Resources, resourceFile)
+			kustomizationMaps[resourcesMap][resourceFile] = true
+		}
+	}
+	for _, value := range child.Images {
+		imageName := value.Name
+		if _, ok := kustomizationMaps[imagesMap][imageName]; !ok {
```
hougangliu: if multiple overlays define same imageName, always use the first overlay. Should we log a msg about ignoring later images? so do other fields, such as crd, Vars

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -577,84 +572,392 @@ func (kustomize *kustomize) writeConfigFile() error {
...
+		log.Warnf("cannot merge nameprefix %v ", child.NamePrefix)
+	}
+	if child.GeneratorOptions != nil && parent.GeneratorOptions == nil {
+		parent.GeneratorOptions = child.GeneratorOptions
+	}
+	if child.NameSuffix != "" {
+		log.Warnf("cannot merge namesuffix %v ", child.NamePrefix)
+
+	}
+	if len(child.CommonLabels) > 0 {
```
hougangliu: ignore all CommonLabels and others? saying I have no overlay, only `base` as line 826, should we still drop these?
maybe for these, we don't try to merge, just use the first one, that is 
```
if len(child.CommonAnnotations) > 0 &&  len(parent.CommonAnnotations) == 0 {
   parent.CommonAnnotations = child.CommonAnnotations  // maybe push them one by one
}
```

hougangliu: sorry, I just didn't find the code to include `base`

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -577,84 +572,392 @@ func (kustomize *kustomize) writeConfigFile() error {
...
+			params, paramFileErr := readLines(paramFile)
+			if paramFileErr != nil {
+				return &kfapis.KfError{
+					Code:    int(kfapis.INVALID_ARGUMENT),
+					Message: fmt.Sprintf("could not open %v. Error: %v", paramFile, paramFileErr),
+				}
+			}
+			for i, param := range params {
+				paramName := strings.Split(param, "=")[0]
+				if val, ok := paramMap[paramName]; ok && val != "" {
```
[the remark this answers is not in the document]
hougangliu: if `componentParams` has defined namespace and project, paramMap will have the two key, right?
if so, once params.env also defines  namespace and project, it will always be written by paramMap counterpart in priority, which may be different with kfDef.Namespace and kfDef.Spec.Project

jlewi:
@kkasravi What's the status of this PR. Is that just addressing the comments from @hougangliu and @gabrielwen ? Are we waiting on @hougangliu and @gabrielwen to review and possibly approve latest changes?

kkasravi:
@jlewi @gabrielwen This fixes #3207 but requires kubeflow/manifests/pull/59 to be merged as well.

PR #3110, opened 2019-04-25
Renames Documentation link and removes it from iframe.

### LINKED ISSUE #2546: Don't iframe the docs into the new Kubeflow UI

prodonjs:
closes #2546 

/area front-end
/area centraldashboard
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3110)

### DIFF
--- a/components/centraldashboard/public/components/main-page.pug
+++ b/components/centraldashboard/public/components/main-page.pug
@@ -12,8 +12,11 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
             a(href$='[[_buildHref("/", queryParams)]]', tabindex='-1')
                 paper-item.menu-item Home
                 aside.divider
+            a(href  ='https://www.kubeflow.org/docs/about/kubeflow/',
+                tabindex='-1', target="_new")
+                paper-item.menu-item Documentation
             template(is='dom-repeat', items='[[menuLinks]]')
-                a(href$='[[_buildHref(item.href, queryParams)]]',
+                a.iframe-link(href$='[[_buildHref(item.href, queryParams)]]',
                     on-click='openInIframe', tabindex='-1')
                     paper-item.menu-item [[item.text]]
         aside.flex
--- a/components/centraldashboard/public/components/main-page.pug
+++ b/components/centraldashboard/public/components/main-page.pug
@@ -11,10 +11,11 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
         iron-selector(selected='{{sidebarItemIndex}}')
             a(href$='[[_buildHref("/", queryParams)]]', tabindex='-1')
                 paper-item.menu-item Home
-                aside.divider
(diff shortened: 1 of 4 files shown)


components/centraldashboard/public/components/main-page.pug
```diff
@@ -12,8 +12,11 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
             a(href$='[[_buildHref("/", queryParams)]]', tabindex='-1')
                 paper-item.menu-item Home
                 aside.divider
```
avdaredevil: Move divider under the Docs link

components/centraldashboard/public/components/main-page.pug
```diff
@@ -12,8 +12,11 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
...
+            a(href  ='https://www.kubeflow.org/docs/about/kubeflow/',
+                tabindex='-1', target="_new")
+                paper-item.menu-item Documentation
```
avdaredevil: Can we add an external link looking icon on this? The way pipelines used to do this.

Perhaps:
image

components/centraldashboard/public/components/main-page.pug
```diff
@@ -12,8 +12,11 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
             a(href$='[[_buildHref("/", queryParams)]]', tabindex='-1')
                 paper-item.menu-item Home
                 aside.divider
+            a(href  ='https://www.kubeflow.org/docs/about/kubeflow/',
```
avdaredevil:
```suggestion
            a(href='https://www.kubeflow.org/docs/about/kubeflow/',
```

components/centraldashboard/public/components/main-page.pug
prodonjs: Sounds good. I also moved it out of the anchor tag since we really don't want that to be clickable. The CSS has been changed slightly to account for that.

components/centraldashboard/public/components/main-page.pug
prodonjs: Not sure how those extra spaces got there. I went with the launch icon since that's what we used for the other doc links on the landing page. I think we should probably try to be consistent but I'm open to your thoughts as well.

prodonjs:
Screenshot from 2019-04-25 16-45-24

Good suggestions. See the above screenshot for the new look.

components/centraldashboard/public/components/main-page.pug
avdaredevil: Yeah, I was looking for it, forgot we already had that icon in our app. This is good!
Mind if I make a quick change in your PR?

avdaredevil:
Here is the design I got:
image

prodonjs:
Sgtm. I like your design better!

PR #3115, opened 2019-04-25
Color Scheme Adjusted to incorporate a better set of colors

avdaredevil:
## About
Uniform color scheme from all pages thus far. Also, Primary colors updated from notebooks UI which is a good example of complementary color palettes

### Tags:
/assign @avdaredevil 
/cc @prodonjs 
/area front-end
/area centraldashboard
/priority p2

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3115)


avdaredevil:
The latter, since this is for sub-apps to follow and no longer conforms to our local style palette.
Bundling would be super useful. @prodonjs would you mind implementing the Babel css bundle?

prodonjs:
/lgtm
/approve

I'll work on making sure this gets exposed in a manner consistent with our library.

PR #3116, opened 2019-04-26
CSS Added to improve #3110

### LINKED ISSUE #2546: Don't iframe the docs into the new Kubeflow UI

avdaredevil:
## About:
Before | After
--- | ---
image | image

/area front-end
/area centraldashboard
/assign @avdaredevil 
/cc @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3116)


components/centraldashboard/public/components/main-page.pug
```diff
@@ -12,9 +12,9 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
             a(href$='[[_buildHref("/", queryParams)]]', tabindex='-1')
                 paper-item.menu-item Home
             a(href='https://www.kubeflow.org/docs/about/kubeflow/',
-                tabindex='-1', target="_new")
-                paper-icon-item.menu-item Documentation
-                    iron-icon(icon="launch", slot="item-icon")
+                tabindex='-1', target="_blank")
```
prodonjs: Thanks for catching this. I don't know why I get that mixed up.

PR #3117, opened 2019-04-26
fixed based images of tf gpu notebooks

zabbasi:
updated base image of jupyter notebooks of tf 1.13.1gpu and tf 2.0.0a0gpu from the following:
- [tf build configuration ](https://www.tensorflow.org/install/source#tested_build_configurations)
- [nvidia docker image tags](https://hub.docker.com/r/nvidia/cuda/tags?page=5),
- [nvidia dockerfile for nvidia/cuda:10.0-cudnn7-runtime-ubuntu16.04 ](https://gitlab.com/nvidia/cuda/blob/ubuntu16.04/10.0/runtime/cudnn7/Dockerfile)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3117)

### DIFF
--- a/components/tensorflow-notebook-image/versions/1.13.1gpu/version-config.json
+++ b/components/tensorflow-notebook-image/versions/1.13.1gpu/version-config.json
@@ -1,5 +1,5 @@
 {
-  "BASE_IMAGE": "nvidia/cuda:9.0-cudnn7-runtime-ubuntu16.04",
+  "BASE_IMAGE": "nvidia/cuda:10.0-cudnn7-runtime-ubuntu16.04",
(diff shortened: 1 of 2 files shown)


stpabhi:
@zabbasi Do you want to test and confirm if tf 2.0.0 alpha works with CUDA 10? I couldn't find the docs either.

/lgtm
/hold

zabbasi:
Screen Shot 2019-04-26 at 10 45 48 AM

stpabhi:
@zabbasi Should we use 18.04 runtime instead of 16.04 such that both CPU and GPU uses same os base?

/cc @jlewi

Issue #2982, opened 2019-04-09
Default IAM role and k8s rbac role for new user under multi-user case


kunmingg:
* IAM role should give access to k8s cluster and IAP app. (Create a custom role? kubeflow-user)
* k8s rbac role should block access to any k8s resource by default. (Assume user self-serve to start through UI)

jlewi:
@kunmingg How's this coming? Can you explain the issue again? Will we just use RBAC as opposed to GKE IAM roles to grant users access to the KF cluster so that they only have access to specific namespaces?

kunmingg:
DevOps / Admin need to grant each user with IAM role:
```
1. Kubernetes Engine Cluster Viewer
```
This role allow user to access clusters, but cannot get/create/delete k8s resources in any namespace by default.

2. user self-serve to create profile, profile will patch rbac role.
3. user can access their namespace via kubectl or Istio.

Will send PR for missing pieces.

jlewi:
I spoke with @kunmingg and I think the issue here is we need to figure out what IAM role to grant users so that at a minimum they can run

```
gcloud container clusters get-credentials ${CLUSTER}
```

in order to configure .kubeconfig.

Once they've run that command we can rely on RBAC for fine grained permissions.

jlewi:
Can close this out once #3123 is merged and we also need to update kubeflow website to tell folks how to set this.

jlewi:
@kunmingg What is the current status of this?

kunmingg:
Will add a README explain how admin manage access via profile, and close this one.

jlewi:
@kunmingg Any progress on the README?

jlewi:
@kunmingg ping?

kunmingg:
Should be covered by:
https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/README.md

Issue #2980, opened 2019-04-09
Users should  be able to create notebooks without service account


zabbasi:
Currently if a given namespace does not have  gcp secret and k8s service account namely user-gcp-sa and jupyter-notebook respectively, then user cannot create any notebook (ui's notebook status shows creating forever). We need to resolve this by giving the option to user to choose his desired service accounts to be installed in the notebook's pod.

zabbasi:
/priority/p0

kimwnasptd:
@jlewi will we need this feature for 0.6? 

IIUC in @avdaredevil 's [WIP PR](https://github.com/kubeflow/kubeflow/pull/3503) the Centraldashboard will only show the Namespaces in which a user has `edit` permissions. This means that the Namespaces available to a user will be the ones that were created by a profile and the user is either the owner or was added with a Binding.

In this case the ServiceAccount `default-editor` will be available in the listed Namespaces and the Notebook Pods will be created.

But if this PR will not be shiped with 0.6, then which SA should the webapp give to the Notebook Pods?

Will this PR be part of 0.6? If not, should the jupyter webapp let the user choose the SA for the Notebooks, or should we just create the `default-editor` for the `kubeflow` namespace? AFAIK we don't create it right now, so the Notebook Pods will fail to be created in that Namespace.

jlewi:
I don't think this is an issue. notebooks are working in 0.6.2

PR #3123, opened 2019-04-26
fix permission leak; give default admin permission for namespace owner

kunmingg:
* fix https://github.com/kubeflow/kubeflow/issues/2601
* Create owner's default k8s rbac rolebinding for https://github.com/kubeflow/kubeflow/issues/2982
* generate service accounts for user pods
  * Can be used by https://github.com/kubeflow/kubeflow/issues/2980
* fix duplicate update for newly created binding
* use k8s default role when possible
* fix go generate

Manually tested, works as expected.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3123)

### DIFF
--- a/components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
+++ b/components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
@@ -39,8 +39,8 @@ const (
 type ProfileStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-	Status  ProfileState `json:"status" protobuf:"bytes,2,opt,name=status,casttype=k8s.io/api/core/v1.ConditionStatus"`
-	Message string       `json:"message,omitempty" protobuf:"bytes,5,opt,name=message"`
+	Status  ProfileState `json:"status,omitempty"`
+	Message string       `json:"message,omitempty"`
 }
 
 // +genclient
--- a/components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
+++ b/components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
@@ -27,10 +27,20 @@ type ProfileSpec struct {
 	Owner rbacv1.Subject `json:"owner,omitempty"`
 }
 
+type ProfileState string
+
+const (
+	ProfileSucceed ProfileState = "Succeed"
+	ProfileFailed  ProfileState = "Failed"
+	ProfileUnknown ProfileState = "Unknown"
+)
+
 // ProfileStatus defines the observed state of Profile
 type ProfileStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
+	Status  ProfileState `json:"status" protobuf:"bytes,2,opt,name=status,casttype=k8s.io/api/core/v1.ConditionStatus"`
+	Message string       `json:"message,omitempty" protobuf:"bytes,5,opt,name=message"`
 }
 
 // +genclient
--- a/components/profile-controller/pkg/controller/profile/profile_controller.go
+++ b/components/profile-controller/pkg/controller/profile/profile_controller.go
@@ -123,7 +123,7 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
 	ns := &corev1.Namespace{
 		ObjectMeta: metav1.ObjectMeta{
 			Annotations: map[string]string{"owner": instance.Spec.Owner.Name},
-			Name: instance.Name,
+			Name:        instance.Name,
 		},
 	}
 	if err := controllerutil.SetControllerReference(instance, ns, r.scheme); err != nil {
@@ -143,8 +143,15 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
 	// No need to update namespace
 	val, ok := foundNs.Annotations["owner"]
 	if (!ok) || val != instance.Spec.Owner.Name {
-		return reconcile.Result{}, fmt.Errorf("namespace already exist, but not owned by profile creator %v",
-			instance.Spec.Owner.Name)
+		log.Info(fmt.Sprintf("namespace already exist, but not owned by profile creator %v",
+			instance.Spec.Owner.Name))
+		instance.Status = kubeflowv1alpha1.ProfileStatus{
+			Status: kubeflowv1alpha1.ProfileFailed,
+			Message: fmt.Sprintf("namespace already exist, but not owned by profile creator %v",
+				instance.Spec.Owner.Name),
+		}
+		r.Update(context.TODO(), instance)
+		return reconcile.Result{}, nil
 	}
 
 	// TODO: add role for impersonate permission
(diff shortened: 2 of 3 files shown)


components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -166,13 +174,14 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
 	roleBinding := &rbacv1.RoleBinding{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      "default",
+			Name:      "namespaceAdmin",
 			Namespace: instance.Name,
 		},
+		// Use default ClusterRole 'admin' for profile/namespace owner
 		RoleRef: rbacv1.RoleRef{
 			APIGroup: "rbac.authorization.k8s.io",
-			Kind:     "Role",
```
lluunn: Why do we bind clusterrole admin to the user?

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -139,7 +141,13 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
 		return reconcile.Result{}, err
 	}
 	// No need to update namespace
+	val, ok := foundNs.Annotations["owner"]
+	if (!ok) || val != instance.Spec.Owner.Name {
```
lluunn: I think this if condition is confusing. When will ok to false?
Should we do a else from above so we know the namespace exist?

Also, I am not sure it's the right behavior to return a reconcile error.
I think it will just come back again later forever.
Should we instead set the status of the profile to be invalid or something? @jlewi WDYT?

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: Creating a roleBinding in target namespace will grant user namespace-scoped permission.

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: ```ok == false``` if user try to create a profile pointing to an existing namespace like default.
Sure we can save status to be failed with the error message.

lluunn: If the namespace is not found, foundNs will be empty and ok also == false right?

I think it's more clear to do a else from line 142:

```
} else { // line 142
  // err != nil, namespace exist
  if foundNs.Annotations["owner"] != instance.Spec.Owner.Name {
    ....
  }
}
```

components/profile-controller/pkg/controller/profile/profile_controller.go
lluunn: could you add a comment explaining that clusterrole can be to namespaced roleref and only has namespaced access?

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: updated

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -119,346 +120,159 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			return reconcile.Result{}, nil
+		}
 	}
-	// No need to update namespace
 
-	role := generateRole(instance)
-	if err := controllerutil.SetControllerReference(instance, role, r.scheme); err != nil {
-		return reconcile.Result{}, err
+	// Update service accounts
+	// "default-editor" would be default "edit" permission for pods in user namespace
```
lluunn: Could you clarify the comment, what's default-editor, what's edit, what does this function do

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -119,346 +120,159 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	if err = r.updateServiceAccount(instance, "default-editor", "edit"); err != nil {
+		log.Info("Failed Updating ServiceAccount", "namespace", instance.Name, "name",
+			"defaultEdittor", "error", err)
+		return reconcile.Result{}, nil
```
lluunn: do you not need to return the error?

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -119,346 +120,159 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
-	}
-	if !(reflect.DeepEqual(roleBinding.RoleRef, found.RoleRef) && reflect.DeepEqual(roleBinding.Subjects, found.Subjects)) {
-		found.RoleRef = roleBinding.RoleRef
-		found.Subjects = roleBinding.Subjects
-		log.Info("Updating RoleBinding", "namespace", roleBinding.Namespace, "name", roleBinding.Name)
-		err = r.Update(context.TODO(), found)
-		if err != nil {
-			return reconcile.Result{}, err
-		}
+	if err = r.updateRoleBingding(instance, roleBinding); err != nil {
```
lluunn: typo updateRoleBingding --> updateRoleBinding. here and elsewhere

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -119,346 +120,159 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		log.Info("Failed Updating Owner Rolebinding", "namespace", instance.Name, "name",
+			"defaultEdittor", "error", err)
+		return reconcile.Result{}, nil
 	}
 	return reconcile.Result{}, nil
 }
 
-func generateRole(instance *kubeflowv1alpha1.Profile) *rbacv1.Role {
-	role := &rbacv1.Role{
+func (r *ReconcileProfile) updateServiceAccount(profileIns *kubeflowv1alpha1.Profile, saName string, ClusterRoleName string) error {
```
lluunn: comment please: what kind of update

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -119,346 +120,159 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+				Namespace: profileIns.Name,
 			},
 		},
 	}
-	return role
+	r.updateRoleBingding(profileIns, roleBinding)
+	return nil
+}
+
+func (r *ReconcileProfile) updateRoleBingding(profileIns *kubeflowv1alpha1.Profile,
```
lluunn: typo

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: good catch

lluunn:
We need e2e test for it, but we can do it in a followup PR.
In that case, could you manually test it and add the result to PR description?

kunmingg:
@lluunn 
description updated

components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
```diff
@@ -27,10 +27,20 @@ type ProfileSpec struct {
...
+	ProfileSucceed ProfileState = "Succeed"
+	ProfileFailed  ProfileState = "Failed"
+	ProfileUnknown ProfileState = "Unknown"
+)
+
 // ProfileStatus defines the observed state of Profile
 type ProfileStatus struct {
 	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
+	Status  ProfileState `json:"status" protobuf:"bytes,2,opt,name=status,casttype=k8s.io/api/core/v1.ConditionStatus"`
```
lluunn: what does the  protobuf:"..." part mean?

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -119,346 +120,163 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+		val, ok := foundNs.Annotations["owner"]
+		if (!ok) || val != instance.Spec.Owner.Name {
+			log.Info(fmt.Sprintf("namespace already exist, but not owned by profile creator %v",
+				instance.Spec.Owner.Name))
+			instance.Status = kubeflowv1alpha1.ProfileStatus{
+				Status: kubeflowv1alpha1.ProfileFailed,
+				Message: fmt.Sprintf("namespace already exist, but not owned by profile creator %v",
+					instance.Spec.Owner.Name),
+			}
+			r.Update(context.TODO(), instance)
```
lluunn: have you manually tested this?

components/profile-controller/pkg/apis/kubeflow/v1alpha1/profile_types.go
kunmingg: legacy from my previous version

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -119,346 +120,163 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			log.Info("Creating ServiceAccount", "namespace", serviceAccount.Namespace,
+				"name", serviceAccount.Name)
+			err = r.Create(context.TODO(), serviceAccount)
+			if err != nil {
+				return err
+			}
+		} else {
+			return err
+		}
+	}
```
lluunn: I guess you don't need to update service account as it doesn't have any spec?

lluunn:
two minor questions. LGTM

Also, how do you fix go generate?

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: yes

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: Yes

kunmingg:
I removed unused generate target, some legacy from dep as kube builder assume their user will use dep.

PR #3125, opened 2019-04-26
Version Related Changes

### LINKED ISSUE #3100: [Central Dashboard] Add commit hash to version at build time
To ease debugging, ensure that the commit hash of the Central Dashboard is passed at build time and included in the UI for ease of version identification.

avdaredevil:
## Updates
- [Windows] make-win updated with all the relevant updates in Makefile
- Dev build is now labeled dev-build
- Dashboard version is now versioned with a commit as well

image

### Tags:
/area front-end
/area centraldashboard
/assign @avdaredevil 
/cc @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3125)

### DIFF
--- a/components/centraldashboard/Dockerfile
+++ b/components/centraldashboard/Dockerfile
@@ -7,7 +7,9 @@ COPY public/ ./public
 COPY webpack.config.js *.json .nycrc ./
 
 ARG kubeflowversion
+ARG commit
 ENV BUILD_VERSION=$kubeflowversion
+ENV BUILD_COMMIT=$commit
 
 RUN npm install && npm run build && npm prune --production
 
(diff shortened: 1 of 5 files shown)


prodonjs:
Sorry for jumping the gun. I realized that this isn't going to work because the Dockerfile doesn't checkout the package from Git. The make process initiates the docker build by copying the source files to the Docker environment and building the package. Hence the presubmit failures since the Docker container doesn't have any awareness of the Git environment.

It seems like the better approach will be to follow the pattern whereby the BUILD_VERSION env variable is set by the process initiated from the checked-out source and is then referenced in the Dockerfile and the webpack config.

avdaredevil:
No worries! Resolved the issue. I didn't realize that was the behavior either; slight oversight on my part. I've left the old behavior as a fallback for dev-mode. But for prod mode, it will get the information from `makefile` or `make-win.ps1`

Issue #3032, opened 2019-04-16
StatefulSet needs ServiceName for Argocd validation


mcminis1:
Stateful sets need a service name to sync in Argocd.

There are 2 missing StatefulSet.serviceNames:
iap.libsonnet : backend-updater
webhook.libsonnet: webhook-bootstrap

jlewi:
@mcminis1 Any interest in submitting a PR?

mcminis1:
@jlewi PR has been created. Thanks.

PR #3134, opened 2019-04-28
3032 statefulset service name

mcminis1:
This PR addresses issue #3032 

ArgoCD expects statefulSets to have a serviceName. In this PR we add a service name to the two stateful sets that have this error.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3134)


kubeflow/gcp/iap.libsonnet
```diff
@@ -276,6 +276,7 @@
           service: "backend-updater",
         },
       },
+      serviceName: "backend-updater",
```
lluunn: we don't have such service, will it work?
Same below

mcminis1: I have both of these in production. Seems to work fine. As far as I can tell this is the workaround.

It's an open (rotten) issue on the k8s repo.
https://github.com/kubernetes/kubernetes/issues/69608

jlewi:
@mcminis1 Any idea why the tests are failing?

PR #3137, opened 2019-04-29
[KUBEFLOW-3136] Fix OAUTH requirement in gke util [WIP]

holdenk:
I need to test this fixes the issue still but here's a proposed solution.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3137)


holdenk:
We're not using these shell scripts anymore.

Issue #3021, opened 2019-04-15
kfctl use ISTIO by default by 04/30


jlewi:
Filing this issue to track the time by when we switch to using ISTIO as opposed to Ambassador by default.

I'd like to target 04/30. This would give us 8 weeks to then work out any kinks on master before cutting 0.6.

@gabrielwen @lluunn Does it seem reasonable to flip over to using ISTIO by default by 04/30?

lluunn:
sgtm

gabrielwen:
sounds reasonable to me.

hougangliu:
@jlewi  in 0.6, will Ambassador be dropped or co-exist with Istio to proxy kubeflow UIs?

jlewi:
@hougangliu the goal is to be migrated completely over to ISTIO

gabrielwen:
@jlewi should we close this one?  ISTIO is by default now and this issue doesn't look like blocked by kustomize.

PR #3139, opened 2019-04-29
Remove ambassador from IAP default config

gabrielwen:
Relates #3021 #2978 #2261

- Remove ambassador from kfctl go binary default config (IAP).
- Remove health check routing rule for default path `/`
- Add `injectIstio` param to basic auth default config.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3139)


bootstrap/config/kfctl_basic_auth.yaml
```diff
@@ -40,6 +40,14 @@ components:
   - tensorboard
   - tf-job-operator
 componentParams:
+  argo:
+  - initRequired: true
```
lluunn: what's initRequired for? Why is it true?

gabrielwen:
```
if initRequired:
  ks generate ...
else:
  ks generate
  ks param set
```

lluunn: I think this is optional param, so should be false?

gabrielwen: because it's optional param in prototypes, this needs to be true.  the only thing that can be false is like this:
https://github.com/kubeflow/kubeflow/blob/8e6c6cd38b55b179e169a67e59b0d6028b501875/kubeflow/pipeline/pipeline.libsonnet#L18

lluunn: IIRC optional params are not required to set at generate.
can set it by "ks param set" later

gabrielwen: there was a conversation such that these two are not equivalent, that is why we designed it this way.

PR #3142, opened 2019-04-29
change max interval for gcp Retry to 30 seconds

gabrielwen:
Currently by using default ExponentialBackoff, it's 1 minute, which is kinda too long.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3142)


bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -77,6 +77,14 @@ const (
 // The namespace for Istio
 const IstioNamespace = "istio-system"
 
+var DEFAULT_EXP_BACKOFF = backoff.ExponentialBackOff{
```
kunmingg: Can we have a function like NewKubeflowBackOff() which returns a new backoff object each time?
Otherwise we need to reset it each time before we use it.
Also if we use backoff in multi-thread case, they might overwrite each other (on same backoff object).

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -77,6 +77,14 @@ const (
...
+	InitialInterval:     backoff.DefaultInitialInterval,
```
kunmingg: Do you want to make InitialInterval bigger?

bootstrap/pkg/kfapp/gcp/gcp.go
gabrielwen: done.

bootstrap/pkg/kfapp/gcp/gcp.go
gabrielwen: done.

PR #3161, opened 2019-04-30
Grant user-created notebook namespaced default edit permission

kunmingg:
We can migrate all user created pods to service accounts.
```default-editor```
```default-viewer```
Those accounts will be managed by profile: https://github.com/kubeflow/kubeflow/pull/3123
We can start with notebook.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3161)

### DIFF
--- a/kubeflow/jupyter/jupyter-web-app.libsonnet
+++ b/kubeflow/jupyter/jupyter-web-app.libsonnet
@@ -88,6 +88,7 @@
       },
     },
 
+    // TODO: "default-editor" will be shared by multiple components; update here once other components switched to new auth-model
     defaultEditorServiceAccount:: {
       apiVersion: "v1",
       kind: "ServiceAccount",
--- a/kubeflow/jupyter/jupyter-web-app.libsonnet
+++ b/kubeflow/jupyter/jupyter-web-app.libsonnet
@@ -87,93 +87,32 @@
         apiGroup: "rbac.authorization.k8s.io",
       },
     },
-    
-    notebookRole:: {
-      apiVersion: "rbac.authorization.k8s.io/v1beta1",
-      kind: "Role",
-      metadata: {
-        name: "jupyter-notebook-role",
-        namespace: params.namespace,
-      },
-      rules: [
-        {
-          apiGroups: [
-            "",
-          ],
-          resources: [
-            "pods",
-            "pods/log",
-            "secrets",
-            "services",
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jlewi** [comment]
Will the default rules be the same across all K8s distribution?

Is there a downside to using our own service accounts? 

It seems like one advantage of using our own service accounts is that we are very explicit about what we need and we don't have to worry about the default roles varying across k8s distributions and causing problems for us.

**Comment by kunmingg** [comment]
According to k8s docs seems default ```edit``` role allows edit any namespaced resource except rbac.
In long term, will we maintain our per-service roles?
One advantage about using standard roles would be when user customize their pods, they can expect their pods have ```standard``` roles, which should being widely used.

File: kubeflow/jupyter/jupyter-web-app.libsonnet [hunk 1]
```diff
@@ -87,93 +87,32 @@
...
       metadata: {
-        name: "jupyter-notebook-role-binding",
+        name: "default-editor-role-binding",
         namespace: params.namespace,
       },
       roleRef: {
         apiGroup: "rbac.authorization.k8s.io",
-        kind: "Role",
-        name: "jupyter-notebook-role",
+        kind: "ClusterRole",
```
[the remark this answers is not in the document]
kunmingg: ClusterRole becomes namespace-scoped when used by RoleBinding.

kubeflow/jupyter/jupyter-web-app.libsonnet
```diff
@@ -87,93 +87,32 @@
...
-    },
-    
-    notebookServiceAccount:: {
+
+    defaultEditorServiceAccount:: {
       apiVersion: "v1",
       kind: "ServiceAccount",
       metadata: {
-        name: "jupyter-notebook",
+        name: "default-editor",
```
lluunn: Is this service account going to be shared between different components?
If so, it should not be part of jupyter-webapp right?

kunmingg: Agree, that would be the final state.
As of today since we start by moving only notebooks to the new auth model, shall we take a shortcut here?

lluunn: I see.
Can you file an issue and add TODO here to move this sa and rolebinding ?

PR #3165, opened 2019-04-30
Pipelines - Added the artifact repository configMap

Ark-kun:
/assign @IronPan

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3165)


Ark-kun:
> Hi @Ark-kun with ml-metadata server coming up, I assume we will we be splitting the artifacts between minio +metadata server where they belong?

I think there needs to be a solid default data storage solution.

Ark-kun:
It's already been added.

PR #3171, opened 2019-05-02
define kubeflow Access Management API

kunmingg:
Define kubeflow access management API which serves following need:
* fulfill user-facing CUJs
  * self-register
  * access sharing
  * user / namespace listing
* provide endpoint for central dashboard and other components to fulfill ```list``` verb
  * query users that can access an namespace
  * query namespaces that accessible from a user

API doc:
https://rebilly.github.io/ReDoc/?url=https://raw.githubusercontent.com/kubeflow/kubeflow/89be54122464e5b033e7b1d192dc8f0e057f1a42/components/access-management/api/swagger.yaml

Note:
* this API will only accept request from Istio and require user identity in coming request
 
Once we locked down the APIs we can implement frontend & backend components.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3171)

### DIFF
--- a/components/access-management/README.md
+++ b/components/access-management/README.md
@@ -30,8 +30,11 @@ The goal is to support multi-tenancy kubeflow cluster / services.
 #####`/v1/bindings`
 * `create`: create new binding; 
   * called when admin or owner share namespace access with other users.
-* `get`: get binding; 
+* `get`: get bindings; 
   * called when query for binding status.
+  * filter by user=... | namespace=... | role=owner/editor to query permissions related to a user / namespace.
+  * For example filter by "user=abc@def.com" when need to list namespaces for user "abc@def.com".
+  * For example filter by "namespace=abc" when need to list users of a namespace "abc"
 * `delete`: delete binding; 
   * called when admin or owner revoke namespace access.
 
--- a/components/access-management/README.md
+++ b/components/access-management/README.md
@@ -1,0 +1,44 @@
+# Kubeflow Access Management API
+
+Kubeflow Access Management API provides fine-grain user-namespace level access control.
+The goal is to support multi-tenancy kubeflow cluster / services.
+
+## Resources
+
+### Profile
+- Profile contains owner which refers to a k8s user.
+- Profile will spawn a namespace with same name and make profile owner the namespace owner.
+- Profile will fulfill necessary k8s resources to enable owner access to kubeflow servuces under multi-tenancy mode.
+- Delete profile will delete namespace and other k8s resources owned by this profile.
+
+### Binding
+- Binding contains a user-namespace pair.
+- Binding will give user edit access to referred namespace.
+- Delete binding will revoke user's access in binding.
+
+
+## APIs
+
+#####`/v1/profiles`
+* `create`: create new profile; 
+  * called when new user self-register.
+
+#####`/v1/profiles/{profile}`
+* `delete`: delete profile; 
+  * called when admin or owner delete profile / namespace.
+
+#####`/v1/bindings`
+* `create`: create new binding; 
+  * called when admin or owner share namespace access with other users.
+* `get`: get binding; 
+  * called when query for binding status.
+* `delete`: delete binding; 
+  * called when admin or owner revoke namespace access.
+
+#####`/v1/users/{user}/namespaces`
+* `get`: return namespaces that queried user can access as owner or editor.
+  * called when need to list namespaces for an user: list namespaces on notebook spawner for current user
+
+#####`/v1/profiles/{profile}/users`
+* `get`: return users with owner or editor permission to queried profile / namespace.
+  * called when need to list users of a namespace
--- a/components/access-management/README.md
+++ b/components/access-management/README.md
@@ -7,14 +7,23 @@ The goal is to support multi-tenancy kubeflow cluster / services.
 
 ### Profile
 - Profile contains owner which refers to a k8s user.
-- Profile will spawn a namespace with same name and make profile owner the namespace owner.
-- Profile will fulfill necessary k8s resources to enable owner access to kubeflow servuces under multi-tenancy mode.
+- Profile will create a namespace with same name and make profile owner the namespace owner.
+- Profile will create necessary k8s resources to enable owner access to kubeflow servuces under multi-tenancy mode.
+- After the namespace is created, the owner can grant access to additional users or groups using the API or by creating RBAC roles & bindings directly.
 - Delete profile will delete namespace and other k8s resources owned by this profile.
 
 ### Binding
 - Binding contains a user-namespace pair.
 - Binding will give user edit access to referred namespace.
 - Delete binding will revoke user's access in binding.
+
+
+## Use Cases
+#### Case 1: Self serve Kubeflow
+
+A Kubeflow admin would like to grant a set of users the ability to create/delete a namespace in which they can consume Kubeflow. 
+The resulting namespace should only be accessible to the user who created it; or people they grant access to.
+RBAC doesn't directly support this. So we provide a service on top of RBAC which will enforce this based on the user's identity
 
 
 ## APIs
--- a/components/access-management/README.md
+++ b/components/access-management/README.md
@@ -3,7 +3,7 @@
 Kubeflow Access Management API provides fine-grain user-namespace level access control.
 The goal is to support multi-tenancy kubeflow cluster / services.
 
-## Resources
+## Resources under management
 
 ### Profile
 - Profile contains owner which refers to a k8s user.
@@ -25,6 +25,10 @@ A Kubeflow admin would like to grant a set of users the ability to create/delete
 The resulting namespace should only be accessible to the user who created it; or people they grant access to.
 RBAC doesn't directly support this. So we provide a service on top of RBAC which will enforce this based on the user's identity
 
+#### Case 2: Allow user to update / share resources they own.
+
+We will not grant user permission to edit cluster-scoped CRD resources. Then to enable user editting their own resources,
+we provide update / share / delete APIs and perform permission check on every coming request.  
 
 ## APIs
 
@@ -42,15 +46,7 @@ RBAC doesn't directly support this. So we provide a service on top of RBAC which
 * `get`: get bindings; 
   * called when query for binding status.
   * filter by user=... | namespace=... | role=owner/editor to query permissions related to a user / namespace.
-  * For example filter by "user=abc@def.com" when need to list namespaces for user "abc@def.com".
-  * For example filter by "namespace=abc" when need to list users of a namespace "abc"
+  * For example filter by "user=abc@def.com" when need to list namespaces for user "abc@def.com"; called when need to list namespaces for an user (list namespaces on notebook spawner for current user)
+  * For example filter by "namespace=abc" when need to list users of a namespace "abc"; called when need to list users of namespace "abc".
 * `delete`: delete binding; 
   * called when admin or owner revoke namespace access.
-
-#####`/v1/users/{user}/namespaces`
-* `get`: return namespaces that queried user can access as owner or editor.
-  * called when need to list namespaces for an user: list namespaces on notebook spawner for current user
-
-#####`/v1/profiles/{profile}/users`
-* `get`: return users with owner or editor permission to queried profile / namespace.
-  * called when need to list users of a namespace
--- a/components/access-management/api/swagger.yaml
+++ b/components/access-management/api/swagger.yaml
@@ -85,8 +85,8 @@ paths:
         500:
           description: "Internal Server Error"
     get:
-      summary: "Get binding specified in request body"
-      description: "If succeeded, returned binding will contain it's current status"
+      summary: "Get bindings filtered by query param"
+      description: "bindings will contain it's current status"
       operationId: "read binding"
       consumes:
         - "application/json"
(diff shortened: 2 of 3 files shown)


jlewi:
@kunmingg How does the profile spawner API compare to the profile CR?

/assign @kkasravi

kunmingg:
@jlewi 
profile spawner and profile CR is like notebook spawner and notebook CR.
One difference is profile spawner backend could be queried by other UIs like central dashboard and notebook spawner UI.

Only admin or profile spawner backend will have permission to edit profile CR.

jlewi:
@kunmingg If we need to build a web app like notebook manager to manage profiles why wouldn't that just  use the K8s APIServer to create/list/get/delete profile resources?

kunmingg:
@jlewi 
Compared to k8s API, there are 2 benefits from it I can think of: 
1. enable self-serve registry + namespace owner's ability for access management (share/revoke) at the same time.
2. common interface for other components like central dashboard to determine user permission
   * query users that can access an namespace
   * query namespaces that accessible from a user

jlewi:
Can you create a README.md and add a link to the redoc rendering?

We should also add explanation of what use cases the API is targeting.

For example, it looks like one of the motivating use cases is to allow the jupyter web app to list the namespaces that should be shown to a user.

jlewi:
Chatted with @kunmingg Here's what we are thinking

* We should follow K8s style 

   * Stick with CRUD style APIs and avoid custom verbs if we can
   * To grant users permission to a Kubeflow namespace; i.e. to share it we should do that by creating
      RoleBindings of users/groups to appropriate roles

* The reason we need an API and corresponding layer of abstraction on top of RBAC is because just relying on RBAC is insufficient

     * We want users to be able to delete profile resources; but only the profile resources they own
        * with RBAC we can only scope a users permission by namespace
        * But the profile resources all live in the same namespace so there's no way to say user Alice can only delete profile resources associated with Alice
        * We can handle this by creating a server which implements delete and implements the appropriate check 

      * We run into similar issues with ISTIO
         * Kunming can you clarify what those are?

Following a crud style API
  
   * Introduce a Profile resource
      * Delete is used to allow a user to delete the profile resources they own
      * List/Get is used to allow clients to determine which profiles (namespaces) are associated with a user
          * Query args will allow filtering by user
          * So Jupyter Web App for example would issue a list with a query arg for user Alice and then iterate over the resulting list of profiles to get all the namespaces Alice has access to

   * Introduce a Binding resource
       * To  share a profile with a user, you create a Binding granting the user access some role like "kubeflow-user"


/cc @kkasravi

kkasravi:
> We want users to be able to delete profile resources; but only the profile resources they own
with RBAC we can only scope a users permission by namespace
> But the profile resources all live in the same namespace so there's no way to say user Alice can only delete profile resources associated with Alice
> We can handle this by creating a server which implements delete and implements the appropriate check

I think there are some good reasons for having this API in front of k8s. But I thought it would be instructive to validate the current profiles flow based on master (or pending PR's against master). For example, I used the profiles manifest in kubeflow/manifests/pull/31 since it has a corrected crd (from profile-controller) and a corrected Role that is created in the 'kubeflow' namespace. 

1. `$ kfctl init ~/experiments --project constant-cubist-173123 --package-manager kustomize@pull/31 --version pull/3108 --namespace=kubeflow`
2. `cd experiments`
3. `$ kfctl generate all`
4. `$ kfctl apply all`
The Role created in the 'kubeflow' namespace is used to allow user's to submit a Profile. The rules are only the profiles resource with verbs 'create', 'list', 'get'. Removing 'get' results in failure so all 3 are needed.
```
$ ☞  kubectl get roles profiles-role -oyaml -n kubeflow
kind: Role
metadata:
  creationTimestamp: "2019-05-05T15:25:10Z"
  labels:
    app.kubernetes.io/name: experiments
    kustomize.component: profiles
  name: profiles-role
  namespace: kubeflow
  resourceVersion: "3057421"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/experiments/roles/profiles-role
  uid: f8a2dd93-6f49-11e9-9bef-42010a8a01ef
rules:
- apiGroups:
  - kubeflow.org
  resources:
  - profiles
  verbs:
  - create
  - list
  - get
```
Unfortunately this allows a user to see other user's profiles, something that the profile-spawner API could filter but if the user were to use kubectl, we would probably want an Admission Controller. This wouldn't obviate the need for the profile-spawner, but the filtering would happen server-side.

5. `kubectl apply -f alice-service-account.yaml` where 

alice-service-account.yaml
```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: alice
  namespace: kubeflow
```
 
6. Patch the profile-role-binding in kubeflow to add the 'alice' ServiceAccount
```
$ ☞  kubectl get rolebindings.rbac.authorization.k8s.io profiles-role-binding -oyaml
kind: RoleBinding
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"rbac.authorization.k8s.io/v1","kind":"RoleBinding","metadata":{"annotations":{},"name":"profiles-role-binding","namespace":"experiments"},"roleRef":{"apiGroup":"rbac.authorization.k8s.io","kind":"Role","name":"profiles-role"},"subjects":[{"kind":"ServiceAccount","name":"profiles-service-account","namespace":"experiments"},{"kind":"ServiceAccount","name":"alice","namespace":"experiments"}]}
  creationTimestamp: "2019-05-05T15:25:10Z"
  labels:
    app.kubernetes.io/name: experiments
    kustomize.component: profiles
  name: profiles-role-binding
  namespace: experiments
  resourceVersion: "3057444"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/experiments/rolebindings/profiles-role-binding
  uid: f8a721ff-6f49-11e9-9bef-42010a8a01ef
roleRef:
  kind: Role
  name: profiles-role
subjects:
- kind: ServiceAccount
  name: profiles-service-account
  namespace: experiments
- kind: ServiceAccount
  name: alice
  namespace: experiments
```

7. `kubectl config set-context $(kubectl config current-context) --user=alice --namespace=kubeflow`
8. `kubectl apply -f profile.yaml`

profile.yaml
```
kind: Profile
metadata:
  name: resnet50
  namespace: experiments
spec:
  owner:
    kind: ServiceAccount
    name: alice
    namespace: experiments
```

These are the rolebindings the profile-controller is creating in the profile namespace 'resnet50'
```
$ ☞  kubectl get rolebindings.rbac.authorization.k8s.io
NAME             AGE
default-editor   61m
default-viewer   61m
namespaceAdmin   61m
```

9. `kubectl get rolebindings.rbac.authorization.k8s.io namespaceAdmin -oyaml`
```
kind: RoleBinding
metadata:
  creationTimestamp: "2019-05-05T15:25:33Z"
  name: namespaceAdmin
  namespace: resnet50
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: Profile
    name: resnet50
    uid: 063bea79-6f4a-11e9-9bef-42010a8a01ef
  resourceVersion: "3057546"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/resnet50/rolebindings/namespaceAdmin
  uid: 0647123f-6f4a-11e9-9bef-42010a8a01ef
roleRef:
  kind: ClusterRole
  name: admin
subjects:
- kind: ServiceAccount
  name: alice
  namespace: experiments
```

10. `kubectl get rolebindings.rbac.authorization.k8s.io default-editor -oyaml`
```
kind: RoleBinding
metadata:
  creationTimestamp: "2019-05-05T15:25:33Z"
  name: default-editor
  namespace: resnet50
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: Profile
    name: resnet50
    uid: 063bea79-6f4a-11e9-9bef-42010a8a01ef
  resourceVersion: "3057537"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/resnet50/rolebindings/default-editor
  uid: 0640a834-6f4a-11e9-9bef-42010a8a01ef
roleRef:
  kind: ClusterRole
  name: edit
subjects:
- kind: ServiceAccount
  name: default-editor
  namespace: resnet50
```

11. `kubectl get rolebindings.rbac.authorization.k8s.io default-viewer -oyaml`
```
kind: RoleBinding
metadata:
  creationTimestamp: "2019-05-05T15:25:33Z"
  name: default-viewer
  namespace: resnet50
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    kind: Profile
    name: resnet50
    uid: 063bea79-6f4a-11e9-9bef-42010a8a01ef
  resourceVersion: "3057542"
  selfLink: /apis/rbac.authorization.k8s.io/v1/namespaces/resnet50/rolebindings/default-viewer
  uid: 0644702a-6f4a-11e9-9bef-42010a8a01ef
roleRef:
  kind: ClusterRole
  name: view
subjects:
- kind: ServiceAccount
  name: default-viewer
  namespace: resnet50
```

It's possible to grant other users the 'edit' role and again have an admission controller that prevents deletion of resources that user's don't own. I'm also wondering if this would be something that the Application Controller could monitor and control since it owns all resources.

kunmingg:
@kkasravi 
Any reason to block ```get profile``` from individual users?
From my understand ```get profile of  current cluster``` is like ```view IAM page of current project```, it should not cause permission leak.

```edit``` role + ```admission controller``` is ```allow by default``` + ```block on condition```:
* It requires ```admission controller``` is up and healthy, if it's down then permission leak happens: access becomes wide open. 

```profile``` + ```API``` model is  ```block by default``` + ```allow on condition```:
Even API server is down or profile / binding got deleted accidentally, we won't have permission leak.

jlewi:
@kunmingg is it possible in the API spec to list query args so that for the get method we could list the query arg used to filter by user?

For the create binding request should the payload more closely resemble an rbac binding?
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#rolebinding-v1-rbac-authorization-k8s-io

e.g. right now it takes user and kind; but an rbac role has roleRef and subjects.  I think we want to mirror that so we can create bindings for account types other than users (e.g. service account, groups).

The API should largely be a a shim over the K8s API so I think it makes sense to take the same resources the K8s API takes and just pass them through. 

The user can just specify roleRef and reference an actual RBAC role created by the profile. This way if we have more than 1 role nothing needs to change.

kkasravi:
@kunmingg 

See my UPDATE - where it's possible to now list, watch and create Profiles but not get them.

> Any reason to block get profile from individual users?
From my understand get profile of current cluster is like view IAM page of current project, it should not cause permission leak.

If we allow **get** then user 'Joe' can **list** profiles and then **get** 'resnet50' and discover that its owner is Alice. This would be a permission leak. If we use secrets or something opaque in the Profile owner's filed then we could lookup that opaque value to get the actual user.

kkasravi:
@jlewi @kunmingg 

Would we want to add `<cluster>-user@<project>.iam.gserviceaccount.com` and `<cluster>-vm@<project>.iam.gserviceaccount.com` as subjects to the 'resnet's role-binding or would we only inject their credentials on pods?

kunmingg:
@jlewi 
Now `get` methods are merged with filter options.

@kkasravi 
It could be more clear if `binding` here only deal with k8s resource/service access permission. 
For cloud service accounts we can let user setup via pod preset.

jlewi:
> Would we want to add <cluster>-user@<project>.iam.gserviceaccount.com and <cluster>-vm@<project>.iam.gserviceaccount.com as subjects to the 'resnet's role-binding or would we only inject their credentials on pods?

Not sure I understand this question in the context of this PR? Do you mean would we want the profile controller to automatically add that binding?

I think the answer is no.  The SA
<cluster>-user@<project>.iam.gserviceaccount.com

Only makes sense it a single tenant situation. For example it only makes sense if user alice creates a new GCP project for herself and deploys Kubeflow in that project. In that case Alice is basically agreeing to the fact that what our setup of a new project i.e. create <cluster>-user@ makes sense for her use case.

A multi-tenant solution would be like the following.

Acme Co. has an internal ML Infra team that is administering a shared GKE cluster. Each team will get their own isolated namespace to run Kubeflow.

The shared GKE cluster is running in project "ml-infra". 

The recommendations team needs to access GCP resources in their project "recs". They already have SA  "recs-notebooks"@ that has access to the requisite GCP resources.

There's no way that the profile controller could know to use "recs-notebooks@" for the recs team.

Someone on the recs team would have to 

* Create an RBAC binding to give that SA any RBAC permissions necessary in their namespace
* Create a K8s secret in the namespace storing a GCP service account key
* Create a pod presets resource to inject that particular service account into pods.

kunmingg:
@kkasravi 
Thanks! It's accurate except we only allow one owner per profile.
For layer based controllers, I feel it's non trivial to manage resource arrays.

lluunn:
I think we should put the discussion here into a doc.
1. to have a better picture of the security model
1. sort out the roleBinding and CUJ for each persona 
1. why we need a server on top of k8s + istio rbac
1. compare and align with [istio](https://istio.io/blog/2018/soft-multitenancy/) and [k8s](https://github.com/kubernetes-sigs/multi-tenancy/blob/master/docs/profiles/profile-soft-multitenancy-s1.md) multi-tenancy model

kkasravi:
+1, borking the security model would not be good. A day or so spent on @lluunn suggestions may save lots of time later

kkasravi:
/lgtm 

thanks @kunmingg for your patience!

PR #3185, opened 2019-05-03
Extend the grace period for ksonnet.go to delete namespace

zhenghuiwang:
https://github.com/kubeflow/testing/issues/361

We have testing infra resources leaked due to undeleted ingress resources. A possible cause is the ingress controller doesn't have enough time to delete the resource completely. https://github.com/kubernetes/ingress-gce/issues/136#issuecomment-417480155

[It usually takes at least 90s to delete a compute instance.](https://cloud.google.com/compute/docs/instances/deleting-instance) So this PR extends the grace period from 100s to 300s.

/cc @gabrielwen @kkasravi @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3185)

### DIFF
--- a/bootstrap/pkg/kfapp/ksonnet/ksonnet.go
+++ b/bootstrap/pkg/kfapp/ksonnet/ksonnet.go
@@ -18,6 +18,13 @@ package ksonnet
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"regexp"
+	"time"
+
 	"github.com/cenkalti/backoff"
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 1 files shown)


gabrielwen:
test is failed?

jlewi:
Do we need to make the same change for kustomize?

A longer term solution might be to wait until there are no more ingresses in the namespace or the namespace is actually deleted and then exit.

zhenghuiwang:
> A longer term solution might be to wait until there are no more ingresses in the namespace or the namespace is actually deleted and then exit.

https://github.com/kubeflow/kubeflow/issues/3203 for tracking.

Issue #3154, opened 2019-04-30
[CentralDashboard] Match UX Specifications [Tracker]


prodonjs:
Implement changes to match [UX specifications](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZYbjDGnaKqEC/files/MCEJu8Y2hyDScZFDtRjUQxaOQcsK82q9efU)

- [x] 1. Correct header behavior to show as white when viewing an iframed application 
- [x] 2. Move and restyle namespace selector
- [x] 3. Update sidebar link and card content
- [x] 4. Update font and spacing for cards
- [x] 5. Update hover/active styles for sidebar
- [x] 6. Move horizontal view tabs from header to content section.

/area front-end
/area centraldashboard
/priority p1
/cc @ajayalfred 
/assign @avdaredevil 
/assign @prodonjs

avdaredevil:
Could you please add `[Tracker]` To the title? And thanks for making this, I'll tag respective PRs to this

prodonjs:
Sure I noticed that in the past but I didn't know it's significance. You
may be able to change the issue title but I'm not sure.

PR #3190, opened 2019-05-03
Centralize iframe link building and replace quick shortcuts links

prodonjs:
Also factor out utility functions into a Mixin that can be used across components.

Addresses point 3 from #3154 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3190)

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -33,15 +33,15 @@ import './namespace-selector.js';
 import './dashboard-view.js';
 import './activity-view.js';
 import './not-found-view.js';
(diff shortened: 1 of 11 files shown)


components/centraldashboard/public/components/main-page.js
```diff
@@ -33,15 +33,15 @@ import './namespace-selector.js';
 import './dashboard-view.js';
 import './activity-view.js';
 import './not-found-view.js';
+import utilitiesMixin from './utilities-mixin.js';
```
avdaredevil: Awesome!

PR #3157, opened 2019-04-30
Remove blue specification for toolbar

prodonjs:
Addresses item 1 from #3154 by removing the **blue** specification for the app-toolbar. Also adds, the shadow elevation indicated from the specifications.

Dashboard:
Screenshot from 2019-04-30 15-15-58

Notebook Servers:
Screenshot from 2019-04-30 15-16-13

Pipelines:
Screenshot from 2019-04-30 15-16-36

/area front-end
/area centraldashboard
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3157)

PR #3191, opened 2019-05-03
 Restyles dashboard namespace selector to match UX specifications

prodonjs:
Addresses item 2 from #3154
Child of #3157 

Empty State:
Screenshot from 2019-05-03 16-19-20

Selected:
Screenshot from 2019-05-03 16-19-42

When viewing iframed-application:
Screenshot from 2019-05-03 16-20-04

/area front-end
/area centraldashboard

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3191)

### DIFF
--- a/components/centraldashboard/public/components/main-page.css
+++ b/components/centraldashboard/public/components/main-page.css
@@ -121,6 +121,10 @@ app-toolbar>header {
     fill: currentColor;
 }
 
+#NamespaceSelector {
+    flex-basis: 33%;
+}
+
 #NamespaceSelector>paper-dropdown-menu {
     color: var(--accent-color);
     --paper-input-container-focus-color: var(--accent-color);
--- a/components/centraldashboard/public/components/main-page.css
+++ b/components/centraldashboard/public/components/main-page.css
@@ -87,13 +87,6 @@ app-drawer .footer>.build {
     color: rgba(255, 255, 255, 0.64)
 }
 
-#NamespaceSelector {
-    position: absolute;
-    right: 1em;
-    top: 50%;
-    transform: translateY(-50%)
-}
-
 app-toolbar {
     @apply --layout-horizontal;
     @apply --layout-center;
@@ -108,6 +101,10 @@ app-toolbar>header {
 #Narrow-Slider {
     @apply --layout-horizontal;
     @apply --layout-center;
+}
+
+#Narrow-Slider[hidden] {
+    display: none;
 }
 
 #Narrow-Slider>.Logo {
--- a/components/centraldashboard/public/components/main-page.pug
+++ b/components/centraldashboard/public/components/main-page.pug
@@ -40,9 +40,9 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
                     paper-icon-button#Menu(icon='menu', drawer-toggle)
                     figure.Logo
                         |!{logo}
-                aside#NamespaceSelector(hides, hidden$='[[hideNamespaces]]')
-                    namespace-selector(query-params='{{queryParams}}',
-                        selected="{{namespace}}")
+                namespace-selector#NamespaceSelector(
+                    query-params='{{queryParams}}', selected="{{namespace}}",
+                    hides, hidden$='[[hideNamespaces]]')
                 paper-tabs.bottom(selected='[[page]]', attr-for-selected='page',
                         hides, hidden$='[[hideTabs]]')
                     paper-tab(page='dashboard', link)
--- a/components/centraldashboard/public/components/main-page.pug
+++ b/components/centraldashboard/public/components/main-page.pug
@@ -40,15 +40,15 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
                     paper-icon-button#Menu(icon='menu', drawer-toggle)
                     figure.Logo
                         |!{logo}
+                aside#NamespaceSelector(hides, hidden$='[[hideNamespaces]]')
+                    namespace-selector(query-params='{{queryParams}}',
+                        selected="{{namespace}}")
                 paper-tabs.bottom(selected='[[page]]', attr-for-selected='page',
                         hides, hidden$='[[hideTabs]]')
                     paper-tab(page='dashboard', link)
                         a.link(tabindex='-1', href$='[[buildHref("/", queryParams)]]') Dashboard
                     paper-tab(page='activity', link)
                         a.link(tabindex='-1', href$='[[buildHref("/activity", queryParams)]]') Activity
-                aside#NamespaceSelector(hides, hidden$='[[hideNamespaces]]')
-                    namespace-selector(query-params='{{queryParams}}',
-                        selected="{{namespace}}")
         neon-animated-pages(selected='[[page]]', attr-for-selected='page',
                             entry-animation='fade-in-animation',
                             exit-animation='fade-out-animation')
--- a/components/centraldashboard/public/components/namespace-selector.js
+++ b/components/centraldashboard/public/components/namespace-selector.js
@@ -1,4 +1,8 @@
 import '@polymer/iron-ajax/iron-ajax.js';
+import '@polymer/iron-icon/iron-icon.js';
+import '@polymer/iron-icons/iron-icons.js';
+import '@polymer/paper-button/paper-button.js';
+import '@polymer/paper-menu-button/paper-menu-button.js';
 import '@polymer/paper-dropdown-menu/paper-dropdown-menu-light';
 import '@polymer/paper-listbox/paper-listbox.js';
 import '@polymer/paper-item/paper-item.js';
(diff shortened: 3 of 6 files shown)


prodonjs:
/assign @avdaredevil 
/assign @ajayalfred 

Here's the PR for the namespace selector restyling. This PR has the commits from #3157 in it, so I'd like to have that approved and merged first before this one goes, but I wanted to at least get it out for review.

https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog has the latest state.

ajayalfred:
Nice! Review comments:

1. Is it possible to default to 'default' namespace instead of showing 'Select namespace' in the dropdown? If the data in the dashboard is being pulled from the default namespace on first time page load, then the dropdown can reflect this.

2. The dropdown arrow appears separated from the dropdown text and this looks awkward because there is no outline border. Any way possible to auto-resize the dropdown based on text length so that the arrow and text can appear closer together? If that's not possible, we may have to add an outline to the control to create a combined look.

3. Dropdown menu appears above the control. Can it be made to appear top-right aligned with the control or below the control?

components/centraldashboard/public/components/main-page.css
```diff
@@ -78,60 +82,38 @@ app-drawer .footer>.build {
...
-    right: 1em;
-    top: 50%;
-    transform: translateY(-50%)
-}
-
 app-toolbar {
-    display: flex;
-    justify-content: center;
+    @apply --layout-horizontal;
+    @apply --layout-center;
```
avdaredevil: Nice!

components/centraldashboard/public/components/main-page.css
```diff
@@ -78,60 +82,38 @@ app-drawer .footer>.build {
...
-    position: absolute;
-    left: 0;
-    bottom: 0;
-    top: 0;
-    width: 256px;
+    @apply --layout-horizontal;
+    @apply --layout-center;
+}
+
+#Narrow-Slider[hidden] {
```
avdaredevil: `[hidden]` should be a global style by itself, not linked to any one element

components/centraldashboard/public/components/main-page.css
```diff
@@ -78,60 +82,38 @@ app-drawer .footer>.build {
...
 app-toolbar>header {
     margin-top: auto;
 }
 
 #Narrow-Slider {
-    position: absolute;
-    left: 0;
-    bottom: 0;
-    top: 0;
-    width: 256px;
```
avdaredevil: Why did we change the CSS to this? The two changes here are not Mutually Exclusive

components/centraldashboard/public/components/main-page.css
```diff
@@ -168,6 +150,10 @@ paper-tab[link] a:hover {
     text-decoration: none;
 }
 
+app-header {
+    box-shadow: 0 2px 6px 1px rgba(232, 234, 237, 0.5), 0 1px 3px 1px rgba(60, 64, 67, 0.15);
```
avdaredevil: Could we try to match this to:
```css
var(--paper-elevation-<some number>)
```

components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -13,6 +17,36 @@ export class NamespaceSelector extends PolymerElement {
...
         return html`
             <style>
+                paper-menu-button {
+                    --paper-menu-button: {
+                        font-size: 14px;
+                        color: #3c4043
+                    }
+                }
+
+                paper-button {
```
avdaredevil: Polymer tries to promote using id's since DOM level scoping and optimizations are already applied, and makes the CSS more specific while increasing readabilty (ie. what does this paper-button do?)

components/centraldashboard/public/components/main-page.css
prodonjs: Understood. However, the specificity of applying the flex layout to `#NarrowSlider` overrides the `hidden` style.

Screenshot from 2019-05-06 09-27-04

Also, using opacity: 0 still claims the normal amount of space so it doesn't result in the other elements sliding over to occupy the space the logo would have been in.

components/centraldashboard/public/components/namespace-selector.js
prodonjs: Understood, I've changed it to dropdown-trigger and used the ID for styling purposes. I think I've gotten into the habit of relying on the fact that the template only has a single element of the type in the shadow DOM but it makes sense to be explicit where possible.

components/centraldashboard/public/components/main-page.css
prodonjs: This is not longer applicable after rebasing.

components/centraldashboard/public/components/main-page.css
prodonjs: None of the paper-style [shadow-elevations](https://github.com/PolymerElements/paper-styles/blob/master/shadow.js) match @ajayalfred 's specification. 

Here is a screenshot of applying `--shadow-elevation-2dp, the least aggressive elevation available in paper-styles.
Screenshot from 2019-05-06 09-41-40

prodonjs:
> Nice! Review comments:
> 
> 1. Is it possible to default to 'default' namespace instead of showing 'Select namespace' in the dropdown? If the data in the dashboard is being pulled from the default namespace on first time page load, then the dropdown can reflect this.

This is something I would like to do as well, but it's not entirely clear what the right default choice would be. The *default* namespace is actually not used for anything in Kubeflow. Unless overridden at setup time, *kubeflow* is the namespace that the application's components get installed to. That is likely to change in the upcoming release though as features are being implemented to allow for user-separation. This will result in more namespaces, with users having permissions to view/manage resources in a subset of them. I think we definitely should look to select a reasonable option upon load, but perhaps table that until we get further along with the multi-user support features.

> 2. The dropdown arrow appears separated from the dropdown text and this looks awkward because there is no outline border. Any way possible to auto-resize the dropdown based on text length so that the arrow and text can appear closer together? If that's not possible, we may have to add an outline to the control to create a combined look.

Done. The mocks had the text width as 170px so I set it to that. I've since reduced it to min-width of 120px. I didn't style it so that the arrow moves in to the width of the text, but if you prefer that look, I can change it accordingly.
Screenshot from 2019-05-06 10-34-53
Screenshot from 2019-05-06 10-35-23

> 3. Dropdown menu appears above the control. Can it be made to appear top-right aligned with the control or below the control?

Changed to ensure that the top right of the menu is fixed to the button right of the selector trigger.
Screenshot from 2019-05-06 10-36-45

ajayalfred:
All these changes look great! Just to clarify, the position of the dropdown arrow is relative to the length of the text? If that' the case, it's perfect.

prodonjs:
Yes, you are right. The drop-down arrow is fixed in place as the length of the text is explicitly specified. I think your previous comment mentioned that you wanted the arrow to float alongside the length of the text, correct?

Here's the behavior in this PR currently:
[Fixed width text](https://screencast.googleplex.com/cast/NjM4ODk0MDE0MzU5MTQyNHxiZGJlNzE1NC00MQ)

Here's the floating arrow behavior with the width of the namespace-selector fixed to 200px;
[Floating arrow](https://screencast.googleplex.com/cast/NTY1OTAxNDQxMTA1OTIwMHxiOWI4ZTc3NC05ZQ)

Let me know if you prefer the latter and I'll make the change.

ajayalfred:
Floating arrow is the way to go. The moving tabs is a little awkward right now, but that's a short term problem since tabs will be moved below the global app bar like so..

image

prodonjs:
Sounds good. Latest change removes the min-width but keeps the max-width to apply the ellipsis for a really-long namespace. Also sets the flex-basis to avoid the moving tabs which as you said, will be going away.

ajayalfred:
Sounds great! Thanks Jason.

components/centraldashboard/public/components/main-page.css
avdaredevil: I see, yes the opacity: 0 thing was only for fade in and out, which only should happen on `[hides][hidden]` not on a default `[hidden]`

components/centraldashboard/public/components/main-page.pug
```diff
@@ -40,15 +40,15 @@ app-drawer-layout.flex(narrow='{{narrowMode}}',
                     paper-icon-button#Menu(icon='menu', drawer-toggle)
                     figure.Logo
                         |!{logo}
+                aside#NamespaceSelector(hides, hidden$='[[hideNamespaces]]')
```
avdaredevil: This and the element under could be merged
```suggestion
                namespace-selector#NamespaceSelector(query-params='{{queryParams}}',hides, hidden$='[[hideNamespaces]]',selected="{{namespace}}")
```

avdaredevil:
Looks good, just one change. It's a nit, so:

/lgtm
/approve
/hold

_**Note:** Feel free to remove the hold_

prodonjs:
I agree. Change made to remove the parent aside tag. Needs LGTM again.

PR #3197, opened 2019-05-04
[CentralDashboard] Column View for Dashboard, Platform links card, user image added to spec

avdaredevil:
## About
- Updated the primary view for Dashboard to be 3 columnar
- Platforms card added and styled to spec

## Preview
image

### Labels:
/assign @avdaredevil 
/cc @prodonjs @ajayalfred 
/area front-end
/area centraldashboard
/priority p1

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3197)


components/centraldashboard/public/components/dashboard-view.js
```diff
@@ -119,20 +119,41 @@ export class DashboardView extends PolymerElement {
...
-            if (pieces.length >= 3) {
-                gcpProject = pieces[2];
-            }
-            this.platformLinks = GCP_LINKS.map((l) => {
-                return {
-                    text: l.text,
-                    link: l.link + gcpProject,
-                };
-            });
+        if (!this.platformInfo || this.platformInfo.providerName != 'gce') {
```
prodonjs: The return statement needs to be uncommented if you were to keep this as-is. I used the positive condition in anticipation of adding additional links for other platforms (AWS, Azure, etc.) where we could first check platformInfo and then have another if block or a switch statement for the platform.

components/centraldashboard/public/components/dashboard-view.js
```diff
@@ -119,20 +119,41 @@ export class DashboardView extends PolymerElement {
...
+            };
+        });
+    }
+    /**
+     * Takes a platform abbrev and converts it to a fully qualified name
+     * @param {string} platform
+     * @return {string}
+     */
+    headerForPlatform(platform) {
+        return {
```
prodonjs: I think both this and the `logoForPlatform` method could be eliminated in favor of a statically defined configuration object for each platform. They could be stored in a map keyed by the platform name and then assigned as a component property when the response comes back from the server.

This would let us eliminate all one but one of the platform properties and then just refer to the `platform` object's properties in the template. Then, adding a new platform is just a matter of defining the logo, links, header, etc. What do you think?

components/centraldashboard/public/components/main-page.js
```diff
@@ -227,8 +227,8 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
     _onPlatformInfoResponse(responseEvent) {
         const {response} = responseEvent.detail;
         this.platformInfo = response;
-        if (this.platformInfo.kubeflowVersion) {
-            this.buildVersion = this.platformInfo.kubeflowVersion;
+        if (response.kubeflowVersion) {
```
prodonjs: Any reason for the change here?

components/centraldashboard/public/components/dashboard-view.js
avdaredevil: Yes, I will actually revert it before merging since we'll probably have links relevant to different platforms.

components/centraldashboard/public/components/dashboard-view.js
avdaredevil: Fair enough, I'll go ahead and add it in there! That seems to be a better approach, so all information is centralized. And it scales better as we add more fields.

Unrelated: do you know why the platform info api locally takes 10+ seconds to respond?

components/centraldashboard/public/components/main-page.js
avdaredevil: Brevity; can revert if preferred.

components/centraldashboard/public/components/dashboard-view.js
avdaredevil: Figured out it's because of network timeouts since I cannot connect to the cluster.

components/centraldashboard/public/components/dashboard-view.js
```diff
@@ -119,7 +119,7 @@ export class DashboardView extends PolymerElement {
      * Observer for platformInfo property
      */
     _platformInfoChanged() {
-        if (this.platformInfo && this.platformInfo.providerName === 'gce') {
+        if (this.platformInfo && this.platformInfo.providerName != 'gce') {
```
prodonjs: I think you still want `===` here right?

components/centraldashboard/app/api.ts
```diff
@@ -3,24 +3,47 @@ import express from 'express';
...
+interface PlatformStatics {
+    [key: string]: PlatformMeta;
+}
+interface PlatformMeta {
+    logo: string;
+    title: string;
+}
+
+/** API routes exposing information from Kubernetes */
+function _platformStatics(platform: string): {meta: PlatformMeta|undefined} {
```
prodonjs: I was originally thinking this would live in the client-side code but this works well also. This would be one advantage to using TS in the front-end as well is that we should share type definitions.

I would suggest expanding the interface definition of PlatformInfo inside of the `k8s_service` module. Then, rather than use the _platformStatics function, I would just define an instance of the platform objects that we intend to support. That seems like a more reader-friendly way of defining the configuration since this is effectively acting like an enum.

components/centraldashboard/app/api.ts
```diff
@@ -3,24 +3,47 @@ import express from 'express';
...
+    console.log("Loading meta for", {platform});
+    const meta = ({
+        gce: {
+            logo: 'https://cloud.google.com/_static/images/cloud/icons/favicons/onecloud/super_cloud.png',
```
prodonjs: One reason I had thought this would live in the client-side code was that we would package this image with our code and inline it like we do for the Kubeflow logo itself.

components/centraldashboard/app/api.ts
```diff
@@ -3,24 +3,47 @@ import express from 'express';
...
-          async (req: express.Request, res: express.Response) => {
-            res.json(
-                await k8sService.getEventsForNamespace(req.params.namespace));
-          });
+    const k8sService = new KubernetesService(kubeConfig);
+    return express.Router()
+        .get(
+            '/platform-info',
+            async (_: express.Request, res: express.Response) => {
+                const response = await k8sService.getPlatformInfo();
```
prodonjs: If we're going to keep the platform-specific stuff on the server-side rather than the client, I would suggest moving it to the `k8s_service` module rather than modifying the response here. That will make it easier to add unit tests since we already have tests for that module and we currently don't have tests for this module since there's no real business logic here.

components/centraldashboard/public/components/main-page.js
prodonjs: I don't know of a guideline that applies but it seems like we should try to avoid non-semantic changes unless part of an intentional refactoring effort.

avdaredevil:
@ajayalfred `Google Sans medium` is not a font available via Google Fonts, so for now I am still using `Google Sans` for titles

components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,5 +1,7 @@
 iron-ajax(auto, url='/api/platform-info', handle-as='json',
     on-response='_onPlatformInfoResponse')
+iron-ajax(auto, url='/api/user-info', handle-as='json',
```
prodonjs: I think when we decide to implement real support for obtaining user identity, we should consider providing a new server endpoint that returns the platform information, list of namespaces, and user identity in a single message. With this change, we'd be issuing 3 AJAX calls after the page loads (Chrome has a limit of 6 concurrent per hostname) to initialize the state of the page so it would probably be a good idea to consolidate. See my previous comment in the server code about tabling this until we have a working identity attainment mechanism.

components/centraldashboard/public/components/resources/cloud-platform-data.js
```diff
@@ -0,0 +1,22 @@
+export const getGCPData = (project) => ({
```
prodonjs: I think this is perfect for future contributors to add functions for other platforms. It might be worth adding a top-level module comment block explaining the purpose of the module as well as a function level doc comment here.

components/centraldashboard/public/components/dashboard-view.js
```diff
@@ -119,20 +102,20 @@ export class DashboardView extends PolymerElement {
...
             }
-            this.platformLinks = GCP_LINKS.map((l) => {
-                return {
-                    text: l.text,
-                    link: l.link + gcpProject,
-                };
-            });
+            platformProps = getGCPData(gcpProject);
         }
+        Object.entries(platformProps).forEach(([key, value]) => {
```
prodonjs: I think this can be simplified for readability even more by dropping the individual platformLinks, platformName, etc. properties and having the object that gets returned by the helper function in the `cloud-platform-data` module meet a certain specification. Then, the template should reference that object's properties (ie. platformDetails.links, platformDetails.name, etc.).

See the attached suggestion.
[suggested_change.patch.txt](https://github.com/kubeflow/kubeflow/files/3162541/suggested_change.patch.txt)

components/centraldashboard/public/components/resources/cloud-platform-data.js
```diff
@@ -0,0 +1,22 @@
...
+            text: 'Deployment Manager',
+            link: `https://console.cloud.google.com/dm/deployments?project=${project}`,
+        },
+        {
+            text: 'Kubernetes Engine',
+            link: `https://console.cloud.google.com/kubernetes/list?project=${project}`,
+        },
+    ],
+    title: 'Google Cloud Platform',
+    logo: '/public/assets/gcp-logo.png',
```
prodonjs: Make sure you import the asset here so webpack adds it to the bundle (similar to how we grab the Kubeflow logo in main-page.js).

components/centraldashboard/public/components/main-page.css
```diff
@@ -177,6 +177,16 @@ neon-animatable {
     height: 100%;
 }
 
+#User-Badge {
```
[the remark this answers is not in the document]
avdaredevil: We can chat about this tomorrow?

components/centraldashboard/public/components/dashboard-view.pug
```diff
@@ -1,25 +1,26 @@
...
+                paper-icon-item
+                    iron-icon(icon='launch', slot='item-icon')
+                    paper-item-body(two-line)
+                        .header [[item.text]]
+                        aside(secondary) [[item.desc]]
+.column
     template(is='dom-if', if='[[platformLinks]]')
-        section#Platform-Links
-            header [[platformName]] Information
+        paper-card#Platform-Links(class$='[[platformName]]', heading='[[platformTitle]]', image='[[platformLogo]]')
```
prodonjs: It looks like this DOM structural change breaks the unit test [here](https://github.com/kubeflow/kubeflow/blob/0d85f7c7a2e22abffad66849a77b6d7242dcd36f/components/centraldashboard/public/components/dashboard-view_test.js#L49). Presubmit tests pass because we don't do client-side tests yet in the presubmit environment because there's a bunch of roundabout stuff necessary to get headless Chrome to work. 

For now, I've just been issuing `npm run coverage` before pushing to make sure everything passes and coverage is high.

components/centraldashboard/public/components/main-page.pug
avdaredevil: Makes sense, stashed this change

components/centraldashboard/public/components/dashboard-view.pug
avdaredevil: image

Fixed

components/centraldashboard/app/api.ts
```diff
@@ -5,22 +5,22 @@ import {KubernetesService} from './k8s_service';
 
 /** API routes exposing information from Kubernetes */
 export function api(kubeConfig: KubeConfig): express.Router {
```
prodonjs: It looks like we have different spacing styles for Typescript and it's not being enforced in the project by tslint. I'm using the Google Typescript standard which is a two-space indent.

 Do you want to revert this file since there are no semantic changes. I'll make sure we enforce the proper spacing on save in a future commit, otherwise we'll just keep flip-flopping back and forth each time one of us touches a file.

components/centraldashboard/public/components/dashboard-view.css
```diff
@@ -81,6 +88,40 @@ paper-card.thin {
...
+.link {
+    display: block;
+    margin: .5em 1em;
+}
+
+.link paper-icon-item {
+    @apply --layout-flex;
+    @apply --layout-justified;
+    @apply --layout-horizontal-reverse;
+    border: 1px solid var(--border-color);
```
prodonjs: Ajay's mocks remove the borders from the dashboard card links.

components/centraldashboard/public/components/dashboard-view.css
```diff
@@ -81,6 +88,40 @@ paper-card.thin {
...
+    padding: .5em 1em;
+    border-radius: 5px;
+    overflow:hidden;
+    --paper-item-icon: {
+        color: var(--accent-color);
+        background: rgba(0, 125, 252, 0.25);
```
prodonjs: The mocks remove the background from the icon button.

components/centraldashboard/app/api.ts
avdaredevil: Sure done!

components/centraldashboard/public/components/dashboard-view.css
avdaredevil: Fixed and updated

components/centraldashboard/app/api.ts
```diff
@@ -5,22 +5,22 @@ import {KubernetesService} from './k8s_service';
...
-          async (_: express.Request, res: express.Response) => {
-            res.json(await k8sService.getNamespaces());
-          })
-      .get(
-          '/activities/:namespace',
-          async (req: express.Request, res: express.Response) => {
-            res.json(
-                await k8sService.getEventsForNamespace(req.params.namespace));
-          });
+	const k8sService = new KubernetesService(kubeConfig);
```
prodonjs: Can we just revert this file out for now? It looks like the tabs are even more out of whack.

components/centraldashboard/public/components/dashboard-view.js
```diff
@@ -1,6 +1,7 @@
 import '@polymer/iron-icon/iron-icon.js';
 import '@polymer/iron-icons/iron-icons.js';
 import '@polymer/paper-card/paper-card.js';
+import '@polymer/paper-ripple/paper-ripple.js';
```
prodonjs: Can we make sure `@polymer/paper-ripple` is included as a dependency in package.json? It's obviously being included by one of our other dependencies but since we are importing it specifically here, we should declare it as such.

components/centraldashboard/public/components/resources/cloud-platform-data.js
```diff
@@ -0,0 +1,37 @@
...
```
prodonjs: Ahh sorry I missed this. The FQP when the application is deployed is /assets. The `public` path isn't propagated by the build process so it's just like the paths for the favicons in index.html.

ajayalfred:
Sorry just catching up with all this! :(
Let's use our meeting to catch up on open action items and the font-size issue.

components/centraldashboard/public/components/dashboard-view.js
avdaredevil: That's a good point!

prodonjs:
@avdaredevil I think this is looking really sharp. I pushed the changes to my cluster (https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/).

I still see a few things from the [mock](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZYbjDGnaKqEC/files/MCEJu8Y2hyDScWR0uwkSmYmqTA1fRs4iJFc) that need addressed.

1. Change Getting Started to Documentation on card with links to KF documentation.
2. Move launch icons to right-hand side of items in Documentation card.
3. Change color of launch icons to material-grey-600 (#80868b)

We should also consider adding a second line of text for the Quick shortcuts and GCP links, but I think we should get this merged and then add that as a follow-on change.

ajayalfred:
All these changes are looking REALLY good!! Thanks for the hard work matching everything to spec.

In regards to width, I hear your point. I made some tweaks to the layout, including min and max-widths, changes to padding between cards and width of the left nav. That should give us help the cards fit in the available horizontal space, hopefully. Can you try this out?

image

avdaredevil:
New Design:
image

ajayalfred:
Can you deploy these changes to a server so I can review it?

avdaredevil:
Updated with labels for Quick Shortcuts:
image

@ajayalfred

prodonjs:
Looks great. The only thing I see is that the minimum column width still seems to be 360px vs. 320px in the redlines.

p2wpBfOXzuU

Also, did we get any kind of consensus on the em vs. px guidance?

Finally, @avdaredevil I think we're very close to merging so you may want to squash all these commits and rebase on top of master. I *think* the automated job tries to do that anyways, but I would hate to see this change introduce 20+ commits into master if it doesn't.

avdaredevil:
@prodonjs Addressed all comments, and rebased

PR #3201, opened 2019-05-06
Update packages and fix TS tests to avoid compiler issues

prodonjs:
This changes bumps the minor versions of most packages, including a fix for **kubernetes/client-node** that was resulting in expired token warnings after about 15m when developing locally. The changes to the `k8s_service_test.ts` were made to indicate to the TS compiler that the mock objects were consistent with the expected types returned from the service.

/area centraldashboard
/area front-end
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3201)


avdaredevil:
Looks great!

/lgtm
/approve

PR #3202, opened 2019-05-06
Adds generalized webhook

### LINKED ISSUE #2844: Generalize the admission webhook to set environment variables and secrets based on config file
The admission webhook 
https://github.com/kubeflow/kubeflow/blob/master/components/gcp-admission-webhook/main.go

Right now hardcodes the environment variables and secrets to mount.
https://github.com/kubeflow/kubeflow/blob/c5977793d5615e4ff867d5f0853e26a1201deb13/components/gcp-admission-webhook/main.go#L51

Can we make this configurable? e.g. make it easy for users to provide a spec describing the environment variables and secrets to mount?

Can we do something like PodPreset?
https://kubernetes.io/docs/tasks/inject-data-application/podpreset/

A simple thing would be to make the web-hook configurable via a YAML file provided via config map.

/cc @eLco @lluunn

zabbasi:
- integrated the existing gcp-admission-webhook with an implementation of [podpreset](https://github.com/jpeeler/podpreset-crd)
- resolves #2844
- related to https://github.com/kubeflow/manifests/pull/33
- please [read here](https://github.com/zabbasi/kubeflow/blob/webhook11/components/admission-webhook/README.md) to see how it works.
- tested manually on a kubeflow cluster
   *  create podpreset in different namespaces with various specs
   * create pods in different namespace and check that the expected podpresets is applied to the pods

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3202)

/assign @lluunn


components/admission-webhook/README.md
```diff
@@ -0,0 +1,89 @@
+## Goal
+We need a way to inject common data (env vars, volumes) to pods (e.g. notebooks).
+See [issue](https://github.com/kubeflow/kubeflow/issues/2641).
+K8s has [PodPreset](https://kubernetes.io/docs/concepts/workloads/pods/podpreset/) resource with similar use-case, however it is in alpha. 
+K8s admission-controller](https://godoc.org/k8s.io/api/admissionregistration/v1beta1#MutatingWebhookConfiguration) and CRD can be used to implement PodPreset as done in [here](https://github.com/jpeeler/podpreset-crd).
```
lluunn: missing "[" before admission-controller

components/admission-webhook/README.md
```diff
@@ -0,0 +1,89 @@
...
+Here is the workflow on how this is used in Kubeflow:
+
+1. Users create  PodPreset manifests which describe additional runtime requirements (i.e., volume, volumeMounts, environment variables) to be injected  into a Pod at creation time.
+PodPresets use label selectors to specify the Pods to which a given PodPreset applies.
+As an example, the following manifest declares a PodPrest to add the secret ```gcp-secret``` in to pods. 
+
+```
+apiVersion: "kubeflow.org/v1alpha1"
+kind: PodPreset
+metadata:
```
lluunn: is podpreset namespaced or cluster scope?

components/admission-webhook/README.md
```diff
@@ -0,0 +1,89 @@
...
+      caBundle: "..."
+    rules:
+      - operations: [ "CREATE" ]
+        apiGroups: [""]
+        apiVersions: ["v1"]
+        resources: ["pods"]
+```
+
+This specifies
+1. When there is a pod being created (see `rules`) in the namespace,
```
lluunn: remove "in the namespace". This applies to all namespaces right?

kubeflow/admission-webhook/README.md
```diff
@@ -0,0 +1,12 @@
+# gcp
```
lluunn: Admission Webhook

Add a link to point to the README in components

kubeflow/admission-webhook/podpreset_controller.libsonnet
```diff
@@ -0,0 +1,98 @@
+{
```
lluunn: file name should not be controller

kubeflow/admission-webhook/podpreset_controller.libsonnet
```diff
@@ -0,0 +1,98 @@
...
+    local podpresetCRD = {
+      apiVersion: "apiextensions.k8s.io/v1beta1",
+      kind: "CustomResourceDefinition",
+      metadata: {
+        name: "podpresets.kubeflow.org",
+      },
+      spec: {
+        group: "kubeflow.org",
+        version: "v1alpha1",
+        scope: "Namespaced",
```
lluunn: So it's namespaced. Could you mention this in README.

kubeflow/admission-webhook/webhook.libsonnet
```diff
@@ -227,6 +229,97 @@
     },  // webhookConfigmap
     webhookConfigmap:: webhookConfigmap,
 
+local webhookRole = {
```
lluunn: why do you need these? the role/sa/rolebinding

components/admission-webhook/main.go
```diff
@@ -0,0 +1,502 @@
...
+	if err != nil {
+		return toAdmissionResponse(err)
+	}
+	jsonPatchBytes, _ := json.Marshal(jsonPatch)
+
+	reviewResponse.Patch = jsonPatchBytes
+	pt := v1beta1.PatchTypeJSONPatch
+	reviewResponse.PatchType = &pt
+
+	// if !reviewResponse.Allowed {
```
lluunn: remove? or uncomment?

components/admission-webhook/main.go
```diff
@@ -0,0 +1,502 @@
...
+		return toAdmissionResponse(err)
+	} else if err != nil {
+		klog.Errorf("error fetching podpresets: %v", err)
+		return toAdmissionResponse(err)
+	}
+
+	klog.Info(fmt.Sprintf("fetched %d podpreset(s) in namespace %s", len(list.Items), pod.Namespace))
+	if len(list.Items) == 0 {
+		klog.V(5).Infof("No pod presets created, so skipping pod %v", pod.Name)
+		fmt.Printf("No pod presets created, so skipping pod %s", pod.Name)
```
lluunn: remove?

components/admission-webhook/main.go
```diff
@@ -0,0 +1,502 @@
...
+		presetNames[i] = pp.GetName()
+	}
+
+	klog.Info(fmt.Sprintf("Matching PP detected of count %v, patching spec", len(matchingPPs)))
+
+	// detect merge conflict
+	err = safeToApplyPodPresetsOnPod(&pod, matchingPPs)
+	if err != nil {
+		// conflict, ignore the error, but raise an event
+		// TODO: investigate why GetGenerateName doesn't work, might be because it's too early to exist yet
```
lluunn: what's this todo?

components/admission-webhook/main.go
```diff
@@ -0,0 +1,502 @@
...
+		msg := fmt.Errorf("conflict occurred while applying podpresets: %s on pod: %v err: %v",
+			strings.Join(presetNames, ","), pod.GetName(), err)
+		//recordConflictEvent(recorder, &pod, msg)
```
lluunn: remove?

lluunn:
Have you tested manually?
Please add what's been tested in PR description

components/admission-webhook/README.md
zabbasi: podpreset is a crd which is cluster scope by default.

kubeflow/admission-webhook/webhook.libsonnet
zabbasi: So that webhook server can access k8s pod/deployments.. and podpresets resources.

components/admission-webhook/README.md
lluunn: It's namespaced right?
see kubeflow/admission-webhook/podpreset-crd.libsonnet 
What do you mean "by default"?

lluunn: Can you put some namespace in the spec, and mention that each user will only view and apply the podpreset in the namespace?

kubeflow/admission-webhook/webhook.libsonnet
lluunn: Does webhook need to access deployments? It only list podpreset right?

components/admission-webhook/README.md
zabbasi: I meant by default, CRD itself is  non-namespaced and is available to all namespaces (https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/)
 Sure will added scope in the spec.

kubeflow/admission-webhook/webhook.libsonnet
zabbasi: It actually needs to read pods, because in order to provide the response it first appends the podpreset' spec (env, volume. volumeMount) to the pod's spec and then computes the differences with respect to the original pod spec.  Have modified rules to give it access to pods instead of deployments.

zabbasi: actually tested the webhook, no need for pods' rbac rule.

lluunn: It doesn't need deployment either right?
It's not reading deployment/pod, but getting the spec from apiserver, and return the change to make.

lluunn:
Thanks, looks good overall, just one minor question on rbac role.

Also, two high level questions:
1. are we going to change the name as discussed in the community meeting?
2. do we need to wait for the PR for fixing cert?

kubeflow/admission-webhook/webhook.libsonnet
zabbasi: Correct. it is updated.

lluunn:
you need to merge master

zabbasi:
> Thanks, looks good overall, just one minor question on rbac role.
> 
> Also, two high level questions:
> 
> 1. are we going to change the name as discussed in the community meeting?
> 2. do we need to wait for the PR for fixing cert?

1. rbac role is already applied
1. I think we can do PodPreset renaming in future PRs when it is integrated with Jupyter-web-app
1. certificate is already fixed in this pr (This PR removes gcp-cred-admission component and adds a new components where cert is already fixed)

Issue #3159, opened 2019-04-30
Release v0.5.1


kunmingg:
* change v0.5 (and future release) kfctl default version from master to v0.5.1

kunmingg:
master head could be broken at any time.
We should use release version as kfctl default version.
@jlewi 
@gabrielwen

lluunn:
what fixes do we need for 0.5.1?

gabrielwen:
current problem is disparity where CLI binary is built with v0-5-branch but using data in master.  We could make the change but in the meantime kustomize should also be considered as it's in different repository now (manifests).  Should we also make branches in manifests?

kunmingg:
We can create release tag(v0.6) on manifests during next release.
For v0.5 we won't support kustomize.

lluunn:
So we just need to change this to 0.5 at 0.5 branch right?

gabrielwen:
Yes: https://github.com/kubeflow/kubeflow/pull/3170

lluunn:
After that, need to update kfctl binary

gabrielwen:
#3170 merged.  should we update the release?

lluunn:
https://github.com/kubeflow/kubeflow/releases/tag/v0.5.1

tyrsius:
[This PR](https://github.com/kubeflow/kubeflow/pull/2983) added some functionality for AWS Cluster deployment, which wasn't in 0.5.0 but was in master. Now that master doesn't point to itself, the docs on using the `--awsClusterName` flag are incorrect.

tyrsius:
Some of the AWS features introduced in [This PR](https://github.com/kubeflow/kubeflow/pull/2983) were broken on master by this change. All of the arguments to `kfctl.sh init` use default values, regardless of what is passed in. This wasn't happening on yesterday's master.

lluunn:
Is it related to 0.5.1?
The change here doesn't touch master.
Do you want to file a separate issue to describe?

tyrsius:
Hmm, perhaps its not. Ill open another issue

Jeffwan:
Let me know if anything I can help for aws related changes.

jlewi:
Closing this issue because 0.5.1 has been cut

PR #3205, opened 2019-05-06
kfctl use 0.5.1

lluunn:
/cc @kunmingg @gabrielwen

For #3159 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3205)


lluunn:
test won't pass because we don't have 0.5.1 target yet.

PR #3216, opened 2019-05-07
Reads from local repo if -r is set

gabrielwen:
Currently the flag repo is not used.  Adding support for using local repository.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3216)


gabrielwen:
@derekhh thanks for fixing it :)

gabrielwen:
Close this for now as this will make 0.5 not compatible with master and future versions.

Issue #3227, opened 2019-05-08
admission webhook certificate problem


lluunn:
The communication between mutatingWebhookConfiguration and the webhook pod needs a certificate, which we provision a self-signed one in this [script](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/gcp/create_ca.sh).

It's Statefulset, and runs:
1. create a cert
1. create a secret to store the cert (this secret will be mounted by webhook pod)
1. periodically check mutatingWebhookConfiguration has the same cert, if not, patch it

The problems here are, 
1. when this pod restarts, it creates a new cert, and patches it to mutatingWebhookConfiguration. However, the webhook server will not reload the the certificate and we will get tls error
2. When we redeploy everything, the webhook pod will read the old secret (since the secret exists), the script will patch new cert to mutatingWebhookConfiguration,  and thus also get tls error.
3. when we delete the component, the secret is not deleted.

cc @zabbasi @jlewi

lluunn:
/assign  @zabbasi
Zahra is fixing this.

jlewi:
@zabbasi is this fixed by #3229 or is there more to do?

zabbasi:
This is fixed  by #3229 and is cherry picked to v0.5.0. 
/close

PR #3229, opened 2019-05-09
Fix webhook to read  updated secret 

zabbasi:
#3227 
Webhook pod when created  loads the secret generated by webhook-bootstarp to communicate with mutatingWebhookConfiguration and then consumes it forever. However if webhook-bootstarp (a statefulset) is restarted, it updates the secret. Therefore we need to force webhook to upload the updated secret. To this end,we chose to restart the webhook pod upon updating the secret. 
/assign @lluunn 

tested:
 -  delete webhook-bootstrap pod
 -  create a new notebook
 -  make sure webhook pod mutates the notebook pod correctly 

This needs to be merged to v0.5 release.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3229)

### DIFF
--- a/kubeflow/gcp/create_ca.sh
+++ b/kubeflow/gcp/create_ca.sh
@@ -91,6 +91,15 @@ kubectl create secret generic ${secret} \
         --dry-run -o yaml |
     kubectl -n ${namespace} apply -f -
 
+# restart the webhook server if it already exists, ignore the error otherwise
+# this is for https://github.com/kubeflow/kubeflow/issues/3227. 
+# Webhook pod once created loads the secret in the begining and starts serving. 
+# Therefore, if secret is updated, then Webhook pod needs to be restarted  
+webhookPod=$(kubectl get pods -n ${namespace} | grep gcp-cred-webhook- |awk '{print $1;}')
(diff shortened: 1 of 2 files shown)


kubeflow/gcp/create_ca.sh
```diff
@@ -91,6 +91,15 @@ kubectl create secret generic ${secret} \
         --dry-run -o yaml |
     kubectl -n ${namespace} apply -f -
 
+# restart the webhook server if it already exists, ignore the error otherwise
+# this is for https://github.com/kubeflow/kubeflow/issues/3227. 
+# Webhook pod once created loads the secret in the begining and starts serving. 
+# Therefore, if secret is updated, then Webhook pod needs to be restarted  
+webhookPod=$(kubectl get pods -n ${namespace} | grep gcp-cred-webhook- |awk '{print $1;}')
```
lluunn: what happens if grep matches nothing (pod not exist)?
Will that be error or webhookPod just equals empty string?
Do you need to check webhookpod empty before kubectl delete pod?

zabbasi: The following line just ignores the error.

lluunn:
thanks!
/lgtm
/approve

PR #3231, opened 2019-05-09
Prune root OWNERs file to better assign reviews.

jlewi:
* @ellis-biegelow and @texasmichelle  are being assigned a lot of reviews because we are hitting
  the root OWNERs file; these reviews shouldn't be assigned to them. 

* Removing them from root OWNERs file is a partial fix; the other part is to create finer grained OWNERs files with better a set of users.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3231)

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -6,5 +6,4 @@ approvers:
   - kunmingg
   - lluunn
   - pdmack
-  - r2d4
   - richardsliu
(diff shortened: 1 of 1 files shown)


richardsliu:
Should Matt still be one of the approvers?

abhi-g:
We can move him to reviewers. I am doubtful he'll have much time after the
summer :)

jlewi:
Since the intent is to prune OWNERs in an effort to better auto-assign reviewers and to put more emphasis on fine grained OWNERs files I think it makes sense to remove him.

/cc @r2d4

Issue #2979, opened 2019-04-09
Make Profile controller support multi-user notebook 


kunmingg:
For each profile CR (created through registry UI), controller need to manage following k8s resources:

Due end of Q2:
* Namespace
* K8s rbac Role & Rolebinding to grant k8s user and service account necessary access to above namespace.
  * enable user’s kubectl access
  * enable service-account’s in-cluster access
* Grant impersonate permission of k8s user account to notebook spawner backend
* Istio rbac rule

Nice to have:
* Grant impersonate permission of namespace service-accounts to notebook-controller

kunmingg:
This effort should be followup of ```multi-user notebook with manual rules```

jlewi:
So next step is to look at the demo script: #3022  and have the profile resource create the appropriate RBAC roles and bindings for Jupyter.

Follow on step is for profile resource to create namespace level access rules for ISTIO to restrict which users can access services in that namespace.

* Before we automate it with the profile resource we should add to the demo script (#3022) what the required commands would be.

jlewi:
@kunmingg What is the current status of this? 

/assign @kunmingg

kunmingg:
Can close once https://github.com/kubeflow/kubeflow/pull/3234 merge

Issue #3206, opened 2019-05-06
cannot connect to notebooks which are created in any namespace other than kubeflow


zabbasi:
Created a kubeflow cluster from master which is deployed with IAP on GCP. 
When notebooks are created in kubeflow namespace, we can successfully connect to the notebooks form ui.
using [this instruction](https://docs.google.com/document/d/1zT1yhDgmS59_uMWJ76hr7FCyXTDQ1Ju1MVADnlCfNEQ/edit) we can also successfully create notebooks in any other namespace. However, while we can connect to the created notebooks using port-forwarding, we cannot connect to the  notebooks from ui (virtual service is created and its configuration looks fine ).

kunmingg:
The gateway is missing in non-kubeflow namespaces.

Kyrremann:
@zabbasi is the problem that it dosen't show up in the UI? Then you can use the search function, and search for your specific namespace.

As our problem is that if we use a different namespace then `kubeflow`, we need to search for our namespace each time we want to connect to a notebook.

jlewi:
Obsolete.

I'm trying out 0.6 and I was able to spawn a notebook in the non Kubeflow namespace and connect to it.

PR #3234, opened 2019-05-09
profile and Istio integration

kunmingg:
Make profile controller manage necessary istio resources to manage namespace access via Istio.
Istio types are generated from Istio 1.1 CRD schema.

fix https://github.com/kubeflow/kubeflow/issues/2979

Each profile will create `IstioServiceRole` and `IstioServiceRoleBinding` to allow profile owner access services in the namespace. This is necessary as we use `whitelist` access control mode for user namespaces: all access are denied without support from Istio Rbac

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3234)


components/profile-controller/pkg/apis/istio/v1alpha1/servicerolebinding_types.go
```diff
@@ -0,0 +1,82 @@
...
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ServiceRoleBindingSpec defines the desired state of ServiceRoleBinding
```
lluunn: I don't think we should maintain our istio type.
Istio should provide it, there is an open PR

lluunn: We can use unstructure to manage istio resource in profile controller

kunmingg: Which open PR? 
Do they have estimate when to support?

kunmingg: We don't need to maintain our istio type, it's one time effort by generating types from Istio CRD.
If we use `unstructured`, for each CRD call we'll need to maintain a struct like https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/pkg/controller/notebook/notebook_controller.go#L448
Considering we might need to manage several Istio CRDs, that's non-trivial efforts to make sure all `[]interface{}` match their true CRD types.

jlewi:
/assign @lluunn 
/assign @gabrielwen 

I'll let @lluunn and @gabrielwen review this PR.

jlewi:
Please update 
https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/OWNERS

Add @kunmingg and @gabrielwen as reviewers

components/profile-controller/pkg/apis/istio/v1alpha1/doc.go
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
gabrielwen: quick question: for the header, should we use kubeflow version or kubernetes version?  I've seen them here and there...

kunmingg: For Istio types, since they are generated from Istio CRD, `kubernetes ` or `istio` might be more accurate than `kubeflow`.
@jlewi wdyt? any concerns?

components/profile-controller/pkg/apis/istio/v1alpha1/servicerolebinding_types.go
lluunn: https://github.com/istio/api/pull/764

lluunn: How did you generate?

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -201,6 +222,145 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+			log.Info("Updating Istio ServiceRoleBinding", "namespace", istioServiceRoleBinding.Namespace,
+				"name", istioServiceRoleBinding.Name)
+			err = r.Update(context.TODO(), foundSrb)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	istioGateway := &istionetworking.Gateway{
```
lluunn: Why do you need a gateway per namespace?

components/profile-controller/pkg/apis/istio/v1alpha1/servicerolebinding_types.go
kunmingg: I generate by
* copy Istio CRD difnition
* code generate via kube builder, like our other CRDs

Seems the PR has been there since Jan. 
We can migrate to it once it's merged and included in istio release.

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: Because gateway is namespace scoped.
IIUC each gateway only responsible for one service mesh, in our case it's roughly services in one namespace.

lluunn: https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/#VirtualService

You can define virtualservices bind to gateway in another namespace.

So I thought we only need one istio-ingressgatway service + one gateway resource per cluster?

kunmingg: Seems not working on our current Istio version.
Maybe we can resolve it in future by updating Istio to latest release.

yanniszark: @kunmingg can you please share the manifest you are trying?
I found istio/istio#11131 and it seems you also need to specify the gateway's namespace. 
The syntax in new releases is <namespace>/<name>, but in older releases it was shortname or fqdn (<name>.<namespace>)
We should be able to make it work!

kunmingg: @yanniszark 
I tried both:
- `<namespace>/<name>`: got `istio "gateway" configuration is invalid: domain name invalid`
- `<name>.<namespace>`: cannot access notebook from UI after change gateway.

The Istio I use is in https://github.com/kubeflow/kubeflow/tree/master/dependencies/istio

yanniszark: @kunmingg I read the source code from the version we use and it seems that `/` notation is not introduced yet.
After trying a couple of different ways, I managed to make it work with `gateway: <name>.<namespace>.svc.cluster.local`.

However, we are using a pre-release version of Istio, which is by-definition not stable.
It seems to me that we should upgrade to a stable version.

kunmingg: Great! Let me have a try.
Will update Istio to stable release.

lluunn: I updated the istio manifest to 1.1.6 in https://github.com/kubeflow/kubeflow/pull/3262

yanniszark: Since we updated, the `<ns>/<name>` form will work and is the recommended way.

lluunn:
the namespace/gateway might also fix https://github.com/kubeflow/kubeflow/issues/3206

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -201,6 +221,97 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
...
+	}
+	istioServiceRoleBinding := &istiov1alpha1.ServiceRoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      ServiceRoleBindingIstio,
+			Namespace: profileIns.Name,
+		},
+		Spec: istiov1alpha1.ServiceRoleBindingSpec{
+			Subjects: []*istiov1alpha1.Subject{
+				{
+					Properties: map[string]string{"request.headers[x-goog-authenticated-user-email]": "accounts.google.com:" + profileIns.Spec.Owner.Name},
```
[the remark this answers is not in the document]
kunmingg: Good point, now offer overwrite option via env var

components/notebook-controller/pkg/controller/notebook/notebook_controller.go
```diff
@@ -438,7 +438,8 @@ func generateVirtualService(instance *v1alpha1.Notebook) (*unstructured.Unstruct
 	if err := unstructured.SetNestedStringSlice(vsvc.Object, []string{"*"}, "spec", "hosts"); err != nil {
 		return nil, fmt.Errorf("Set .spec.hosts error: %v", err)
 	}
-	if err := unstructured.SetNestedStringSlice(vsvc.Object, []string{"kubeflow-gateway"},
+	// TODO(kunming): change to <namespace>/<name> format once Istio update to 1.1 stable release.
```
lluunn: we have 1.1 now, is namespace/name working?

kunmingg: Nice

jlewi:
@kunmingg Can you please update the PR description to explain what ISTIO resources the profile controller is creating and why?

Its not clear to me why the profile controller is creating ISTIO rules. If I understand @yanniszark and @lluunn we should be able to use the ISTIO gateway in the kubeflow shared namespace to route to services in other namespaces.

So my expectation is it would be the individual CR's creating actual services that would need to create ISTIO rules.

For example, I would expect the Jupyter controller to create an ISTIO virtual service to route traffic to the notebook.

kunmingg:
@jlewi 
description updated.
Jupyter controller will create an ISTIO virtual service to route traffic to the notebook. But Jupyter controller cannot control who can access which notebook.
So profile will grant owner the Istio access permission to all services in ns.

jlewi:
@kunmingg Can you explain why are you defining IstioRbacRole and IstioServiceRoleBinding? Is pkg/apis/istiobarc just a way to use kubebuilder to generate go client libraries for the ISTIO CRDs?

Are the files in pkg/apis/istiorbac  autogenerated? If so how?

components/profile-controller/README.md
```diff
@@ -0,0 +1,72 @@
+# Profile CRD
+
+Kubeflow Profile CRD is designed to solve access management within multi-user kubernetes cluster.
+
+Profile access management provide namespace level isolation based on:
```
kkasravi: provide -> provides

components/profile-controller/README.md
```diff
@@ -0,0 +1,72 @@
...
+- Istio rbac access control
+
+**Resources managed by profile CRD:**
+
+Each profile CRD will manage one namespace (with same name as profile CRD), and will have one owner.
+Specifically, each profile CRD will manage following resources:
+
+- Namespace
+- K8s rbac Role & Rolebinding to grant profile owner access to above namespace via k8s API (kubectl)
+- Istio namespace-scoped gateway & rbac permission to grant profile owner access to above namespace via Istio (browser)
```
kkasravi: Is the user still able to submit a Profile CR and have the namespace and resources created?
You may want to name all the kinds within the Namespace
- ServiceRole
- ServiceRoleBinding
- Role
- RoleBinding
- etc

components/profile-controller/README.md
```diff
@@ -0,0 +1,72 @@
...
+  - This is needed in order to get cluster access by `gcloud container clusters get-credentials`
+- kubeflow cluster with version v0.6
+
+## Manage access control and resources
+
+### manual access management by admin
+
+Cluster admin will manage access management for cluster users:
+
+**To create and reserve namespace `ns1` for user `abc@def.com`**
```
kkasravi: Is this done as user `abc@def.com` or as admin user?

components/profile-controller/README.md
```diff
@@ -0,0 +1,72 @@
...
+    name: abc@def.com
+```
+
+**To revoke access to namespace `ns1` from user `abc@def.com` and delete namespace `ns1`**
+- delete profile ns1:
+```
+kubectl delete profile ns1
+```
+
+You should have kubeflow components deployed inside your k8s cluster. Generated ksonnet application is store in [app-dir](./bootstrapper.yaml#L65).
```
kkasravi: Are these instructions for the admin?

- Should probably be updated for kustomize.
- Would each user have a ConfigMap that would define what components they would like to launch in their profile namespace? 
- Would the ConfigMap be created by them within their profile namespace?
- Have we figured out how this works - I thought Jobs, Notebook Pods would be launched into the profile namespace from the controller running in the kubeflow namespace?
- May want to point to a CUJ or similar
-

components/profile-controller/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+##### How to generate Istio rbac CRD types
+
+- We use kube builder https://book.kubebuilder.io/quick_start.html
+```
+kubebuilder init --domain istio.io --license apache2 --owner "The Kubernetes Authors"
+kubebuilder create api --group rbac --version v1alpha1 --kind ServiceRole
+kubebuilder create api --group rbac --version v1alpha1 --kind ServiceRoleBinding
+```
+- Then copy ServiceRole / ServiceRoleBinding schema from Istio release version https://github.com/istio/istio/releases/tag/1.1.6 to `pkg/apis/istiorbac/v1alpha1/*_types.go`
```
kkasravi: Was this a one time activity?

components/profile-controller/cmd/manager/main.go
```diff
@@ -33,7 +34,12 @@ import (
 func main() {
 	var metricsAddr string
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	var userIdHeader string
+	var userIdPrefix string
```
kkasravi: Does the profile-controller run as a user?

components/profile-controller/pkg/apis/istiorbac/v1alpha1/servicerole_types.go
```diff
@@ -0,0 +1,73 @@
+/*
```
kkasravi: The README could have a short overview on ServiceRole, ServiceRoleBinding, Role, RoleBinding and interactions with the user.

components/profile-controller/README.md
kunmingg: updated

components/profile-controller/README.md
kunmingg: admin, updated.

components/profile-controller/README.md
kunmingg: Oh, this seems copied from bootstrap part, removed.

components/profile-controller/README.md
kunmingg: This instruction indicate how the Istio code are generated.
Should be updated no more than twice a year. (usually when Istio has breaking API change)

components/profile-controller/cmd/manager/main.go
kunmingg: It has admin permission.
Here header and prefix are consistent across each platform.
Like service on GCP or on prem ldap will have same userIdHeader and prefix.

components/profile-controller/pkg/apis/istiorbac/v1alpha1/servicerole_types.go
kunmingg: updated.

kkasravi:
my biggest concerns were that 
1. we no longer provided a 'self-serve' environment for the user 
2. the user needed a clusterrolebinding

these both do not appear to be required.
/lgtm

kunmingg:
Thanks! I feel maybe we should rename `profile` to `workspace`, which feels more accurate to what it does.

kkasravi:
I originally thought about workspace, but thought 'workspace' might be too overused and that there may be use cases like allocating protected namespaces in production to run inference models where this type of namespace isn't really a workspace. But it's probably more like a workspace than a 'profile' if it means an area to do work - whether it's a user or a deployment in production. 

Renaming is fine assuming we haven't gotten the kubeflow community too used to this term.
The multi-tenancy group seems to be calling Profile a 'TenantNamespace' and the group of users which can create TenantNamespaces a Tenant. https://docs.google.com/document/d/1hpJX5O_siMmNGMvIHvz8Pm7XOjJLz5g57XWrgwWarFw/edit#heading=h.vwq4rntm30ig
Their model seems a bit more complicated in some ways.

PR #3244, opened 2019-05-10
Jupyter Web App Integration with Central-Dashboard SDK

### LINKED ISSUE #2845: Jupyter web app should respect namespace in central dashboard

avdaredevil:
## About
This PR integrates the Central Dashboard Event Handler SDK into the Jupyter Web App:
- Ability to know if the page is running in iframe-view or standalone
- Hide the namespace selector on index.html when in iframe
- Event-driven value acquisition directly from central-dashboard's selector
- Disabled input box on add_notebooks.html (read-only)

## Preview
Main Page | Create Notebook Page
--- | ---
image | image

### Labels
/area front-end
/area centraldashboard
/area jupyter
/cc @prodonjs @kimwnasptd 
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3244)


components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
```diff
@@ -1,80 +1,98 @@
 {% extends "layout.html" %}
 {% block javascript %}
-  {{ super() }}
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.11/lodash.min.js"></script>
-  <script src="{{ prefix + url_for('static', filename='js/notebooks.js') }}"></script>
+	{{ super() }}
+    <script defer>
+        // Establish communication with Central Dashboard
+        window.addEventListener('DOMContentLoaded', function (event) {
```
prodonjs: Since we aren't running this JS code through any type of transpilation, do you want to keep it as vanilla as possible and avoid the ES6 destructuring? Whatever we decide, I think we should be consistent in this snippet and right now the rest is mostly ES3-style.

components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
```diff
@@ -1,80 +1,98 @@
...
+            const {CentralDashboardEventHandler: CD_EvHandler} = window.centraldashboard || {}
+            if (!CD_EvHandler) return
+            CD_EvHandler.init(function(handler, isIframed) {
+                document.body.classList[isIframed?'add':'remove']('iframe-view')
+                if (!isIframed) return
```
prodonjs: Let's try to adhere to our normal JS style guidelines from the Google JS style guide with respect to semi-colons and one variable declaration per line.

https://google.github.io/styleguide/jsguide.html#features-one-variable-per-declaration.

components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
avdaredevil: Yessir

components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
```diff
@@ -1,80 +1,99 @@
 {% extends "layout.html" %}
 {% block javascript %}
-  {{ super() }}
-  <script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.11/lodash.min.js"></script>
-  <script src="{{ prefix + url_for('static', filename='js/notebooks.js') }}"></script>
+	{{ super() }}
+    <script defer>
+        // Establish communication with Central Dashboard
+        window.addEventListener('DOMContentLoaded', function (event) {
+            const CD_EvHandler = (window.centraldashboard || {}).CentralDashboardEventHandler;
```
prodonjs: *Nit:* `const` is still an ES6 feature if we're worried about behavior in browsers without native support.

components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
```diff
@@ -1,80 +1,99 @@
...
+            if (!CD_EvHandler) return;
+            CD_EvHandler.init(function(handler, isIframed) {
+                document.body.classList[isIframed?'add':'remove']('iframe-view');
```
prodonjs: This feels a bit terse compared to the approach using an `if` block and explicitly calling the `add` or `remove` method on classList. Maybe I'm just slow haha but is there a guide you've consulted in the past as to when it's preferable to use this style over if blocks?

Also, is there some style or trigger being applied to `body.iframe-view`?

components/jupyter-web-app/kubeflow_jupyter/default/templates/layout.html
```diff
@@ -1,45 +1,47 @@
 <!DOCTYPE html>
 <html>
 <head>
-  <!-- Required meta tags -->
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <!-- Required meta tags -->
```
prodonjs: The indentation settings seem to be different here and they really seem to be out of whack in the templates below. You may need to disable auto-formatting on save temporarily to avoid changing more than just the relevant sections as it's somewhat tough to see what changed.

components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html
avdaredevil: I've found the js-esque method to be condensed and readable. Simple articles like [this one](https://medium.com/@jraleman/ternary-operators-vs-if-else-statements-6c26f7d034f7) kind of articulate what I mean. A person needs to end up writing a lot more code (therefore a lot more re-reading later) when using if:

```js
document.body.classList[isIframed?'add':'remove']('iframe-view');
```
VS
```js
if (isIframed) {
    document.body.classList.add('iframe-view');
} else {
    document.body.classList.remove('iframe-view');
}
```
I see the former being more readable and doesn't increase code depth. But maybe we can actually come up with our own addendum to the style guide and establish common run-ins such as these?

---

To your secode question, it was, but I removed it in favor of `.disabled`, though maybe it makes sense to scope all SDK behavior under the iframe-view, by prepending all CSS rules with:
```css
body.iframe-view /* normal selector */ {/* css here */}

File: components/jupyter-web-app/kubeflow_jupyter/default/templates/layout.html (hunk 5 above)
Comment (avdaredevil) [reply components/jupyter-web-app/kubeflow_jupyter/default/templates/layout.html]: I see, will do

**Comment by avdaredevil** [comment]
Image deployed is: gcr.io/kubeflow-dev/jupyter-web-app:5babf2df

File: components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html (hunk 4 above)
Comment (prodonjs) [reply components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html]: It might be worthwhile to discuss style decisions with the team. Jon has JS readability so he would be a good resource to weigh in on gray areas. I definitely think we should adhere to Google style guide rules as much as possible because the collective wisdom of Google engineering over the years is far better than anything I could come up with.

Comment (avdaredevil) [reply components/jupyter-web-app/kubeflow_jupyter/default/templates/notebooks.html]: /cc @jharbuck

**Comment by kimwnasptd** [comment]
@avdaredevil really nice job, it looks great! 

Just a question, I was trying to test the changes with the image you mentioned above `gcr.io/kubeflow-dev/jupyter-web-app:5babf2df`, but the namepace wasn't changing. 

I am using the `:latest` tag of `centraldashboard` image. From the DevTools' Source tab I can see that `dashboard_lib.bundle.js` is being exposed. But, when I check the `iframe`'s contentWindow for the jupyter-web-app, the `onmessage` key is set to null. Also, the contentWindow doesn't have a `centraldashboard`. 

Do you see the same behavior or am I missing something?

**Comment by avdaredevil** [comment]
Thanks, @kimwnasptd. So the version of centraldashboard isn't pushed yet, I'd use this one instead: 

```
gcr.io/kubeflow-images-public/centraldashboard:v20190517-v0.4.0-rc.1-455-gee78c737
```

I see:
image

**Comment by avdaredevil** [comment]
New Image is: 
```
gcr.io/kubeflow-dev/jupyter-web-app:622b13b6
```

avdaredevil:
@kimwnasptd, did it work for you? If so, would you mind `/lgtm`'ing this change?

kimwnasptd:
@avdaredevil yes, it works really nice! 
/lgtm
/approve

Issue #3225, opened 2019-05-08
[CentralDashboard] Display resource utilization for GKE deployments from Stackdriver


prodonjs:
In accordance with the [UX design](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZYbjDGnaKqEC/files/MCEJu8Y2hyDScWR0uwkSmYmqTA1fRs4iJFc) for the KF dashboard, we should provide users with a high-level view of resource consumption. For starters, CPU and memory utilization should be shown in a time-series view for a window that spans backwards at least 1 hour.

After consideration, the first implementation for obtaining and displaying these metrics will target GKE deployments and utilize the Stackdriver Monitoring API to obtain Node and Pod level metrics.

/area centraldashboard
/area front-end
/assign @prodonjs
/cc @apverma
/priority p1

PR #3251, opened 2019-05-13
Dashboard metrics backend with Stackdriver.

prodonjs:
- Uses Stackdriver Monitoring API as a implementation of a
  MetricsService interface that could be extended for other platforms
  that can be used to obtain time-series metrics for resource utilization.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3251)


prodonjs:
addresses #3225 

/area front-end
/area centraldashboard
/hold

@avdaredevil This is still WIP, but this is what I've started on for the TS backend to support obtaining metrics from Stackdriver. I had to do a bit of refactoring to allow for better dependency injection that makes for much easier testing. Take a look at let me know if you have any thoughts.

components/centraldashboard/app/stackdriver_metrics_service.ts
```diff
@@ -0,0 +1,147 @@
...
+  constructor(
+      private client: monitoring.MetricServiceClient, private projectId: string,
+      private nodeNames: string[]) {}
+
+  async getNodeCpuUtilization(interval: Interval): Promise<TimeSeriesPoint[]> {
+    const response: TimeSeriesPoint[] = [];
+    let filter = `metric.type="${
+        INSTANCE_CPU_UTILIZATION_METRIC_TYPE}" AND metric.label.instance_name = `;
+    if (this.nodeNames.length > 1) {
+      filter += `one_of("${this.nodeNames.join('", "')}")`;
```
avdaredevil: An alternative approach for readability, if you prefer:

```suggestion
      filter += `one_of(${JSON.stringify(this.nodenames)})`;
```

components/centraldashboard/app/stackdriver_metrics_service.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+  }
+
+  private getTimeSeriesInterval(interval: Interval):
+      {startTime: monitoring.Timestamp, endTime: monitoring.Timestamp} {
+    const now = Math.floor(Date.now() / 1000);
+    let secondsToSubtract = 0;
+    switch (interval) {
+      case Interval.Last5m:
+        secondsToSubtract = 60 * 5;
```
avdaredevil: You could refactor this, to simply hold minutes to Substract
```suggestion
        minsToSubtract = 5;
```
And then multiply the `60` in the bottom

prodonjs: I agree. Changed.

components/centraldashboard/app/stackdriver_metrics_service.ts
prodonjs: This actually doesn't have the same effect. The latter would return `["node-1","node-2"]` instead of the `node-1", "node-2` that I need inside of the quotes.

components/centraldashboard/app/api_test.ts
```diff
@@ -0,0 +1,109 @@
...
+      mockK8sService = jasmine.createSpyObj<KubernetesService>(['']);
+
+      testApp = express();
+      testApp.use(express.json());
+      testApp.use('/api', api(mockK8sService));
+      const addressInfo = testApp.listen(0).address();
+      if (typeof addressInfo === 'string') {
+        throw new Error(
+            'Unable to determine system-assigned port for test API server');
+      } else {
```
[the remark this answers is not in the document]
prodonjs: Good call. Changed.

components/centraldashboard/app/stackdriver_metrics_service.ts
avdaredevil: I see, my bad

avdaredevil:
Since WIP
/hold

/lgtm

components/centraldashboard/app/api.ts
```diff
@@ -17,6 +18,31 @@ export function api(kubeConfig: KubeConfig): express.Router {
...
             res.json(await k8sService.getNamespaces());
           })
+      .get(
+          '/metrics/:type((node|podcpu|podmem))',
+          async (req: express.Request, res: express.Response) => {
+            let interval = Interval.Last15m;
+            if (Interval[req.query.interval] !== undefined) {
+              interval = Number(Interval[req.query.interval]);
+            }
+            if (metricsService) {
```
avdaredevil: Just a suggestion, to code-reuse:
```js
if (!metricsService) return res.sendStatus(405);
const method = {
    'node': 'getNodeCpuUtilization',
    'podcpu': 'getPodCpuUtilization',
    'podmem': 'getPodMemoryUsage',
}[req.params.type]
return res.json(await metricsService[method](interval));
```

prodonjs: Typescript won't permit this sort of syntax since it makes it impossible for the type checker to know the types of the methods involved. While I would agree that this is more succinct in terms of lines of codes, I would contend that it's less obvious what the code is doing especially to a first time reader. I've always tried to apply the axiom from the Zen of Python that *explicit is better than implicit*. Practically speaking, I'd rather type a few more lines of code if it saves me the trouble of inspecting what that code is doing 6 or 12 months from now and most of the Google language style rules advise the same.

I'm definitely open to suggestions though, particularly if you have a set of style principles that you think should apply for scenarios like this.

avdaredevil: Makes sense, I wasn't sure. I've grown the style after writing a tons of js, and probably why I am really biased to not knowing when ES2016+ esque code gets less readable.

On the other hand, early return and handling is known generally to be a good design pattern:
- [Software Engineering StackExchange](https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement)
- [Reddit](https://www.reddit.com/r/programming/comments/878zi2/avoid_else_return_early/)
- [HackerNews](https://news.ycombinator.com/item?id=16678209)
- Google tott

prodonjs:
I created #3326 for the front-end work, but I think this is ready for final review and I'll rebase that PR on top of whatever changes get committed here.

/hold cancel

prodonjs:
@avdaredevil PTAL, this is ready for final review after rebasing following your change yesterday.

components/centraldashboard/app/metrics_service_factory.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Factory function to return an instance of the MetricsService based on
+ * platform information.
+ * @returns MetricServiceClient or null if the platform is not currently
+ *  supported.
+ */
+export async function getMetricsService(k8sService: KubernetesService):
+    Promise<MetricsService>|null {
+  try {
+    const [platformInfo, nodes] = await Promise.all(
```
avdaredevil:
```suggestion
    const [platformInfo, nodes] = await Promise.all([
        k8sService.getPlatformInfo(),
        k8sService.getNodes()
    ]);
```

 I think that's a more standard way of writing that

components/centraldashboard/app/stackdriver_metrics_service_test.ts
```diff
@@ -0,0 +1,378 @@
...
+                    startTime: {seconds: 1557705300},
+                    endTime: {seconds: 1557705300},
+                  },
+                  value: {
+                    doubleValue: 0.12,
+                  },
+                },
+              ]
+            }
+          ] as unknown) as gapi.client.monitoring.TimeSeries[];
```
avdaredevil: Why cast it as unknown first?

components/centraldashboard/app/metrics_service_factory.ts
prodonjs: Done.

components/centraldashboard/app/stackdriver_metrics_service_test.ts
prodonjs: Since I am not building a mock with all of the indicated type's properties, Typescript's compiler complains that the object given does not sufficiently overlap with the specified type. By casting it to unknown first, you're able to work around the compiler's warning and say "I know what I am doing in this case". It would generally not be a good idea to do that in production code but for mocking test responses where we only use a subset of the properties in the object, it's a reasonable pattern.

avdaredevil: Gotcha, thanks for the explanation!

PR #3255, opened 2019-05-14
Add permssion for ml-pipeline-ui account to access pod resources

jingzhang36:
I copied the "crd" section to pipeline-ui.libsonnet from pipeline-viewercrd.libsonnet, in order for ml-pipeline-ui service in our pipeline deployment to list/create tensorboard crd.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3255)

### DIFF
--- a/kubeflow/pipeline/pipeline-ui.libsonnet
+++ b/kubeflow/pipeline/pipeline-ui.libsonnet
@@ -144,6 +144,23 @@
             "list",
           ],
         },
+        {
+          apiGroups: [
+            "kubeflow.org",
+          ],
+          resources: [
+            "viewers",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "list",
+            "watch",
+            "update",
+            "patch",
+            "delete",
+          ],
+        },
       ],
     },  // role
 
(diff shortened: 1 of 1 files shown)


jingzhang36:
This PR is related to issue https://github.com/kubeflow/pipelines/issues/1100

kubeflow/pipeline/pipeline-ui.libsonnet
```diff
@@ -185,5 +185,34 @@
         },
       },
     },  // deployUi
+
+    crd: {
```
neuromage: You don't need to add the crd definition here. Instead, you want to give ml-pipeline-ui service account CRUD permission on the `viewer` resource. To do this, modify the service account role definition above, which currently allows `ml-pipeline-ui` permissions to modify pods and pods/logs. Add the following rules as well:

```
 {
          apiGroups: [
            "kubeflow.org",
          ],
          resources: [
            "viewers",
          ],
          verbs: [
            "create",
            "get",
            "list",
            "watch",
            "update",
            "patch",
            "delete",
          ],
        },
```

These are the rules the viewer controller has for modifying `viewer` resources here:
https://github.com/kubeflow/kubeflow/blob/4b1ae00a997cd83b114c5b00d53a0928cdbcb81c/kubeflow/pipeline/pipeline-viewercrd.libsonnet#L54

jingzhang36: Done. Thank you!

kubeflow/pipeline/pipeline-ui.libsonnet
```diff
@@ -144,6 +144,23 @@
...
+            "kubeflow.org",
+          ],
+          resources: [
+            "viewers",
+          ],
+          verbs: [
+            "create",
+            "get",
+            "list",
+            "watch",
```
IronPan: does it need watch, patch, update?

neuromage: We can be pedantic and remove watch/patch, but I imagine update is useful. I don't think the difference is material once you give permissions to create and delete.

IronPan: It is generally good practice to be pedantic with permission to start with minimum. this way we know exactly what UI does by going through this list.

2 month later i need to scratch my head to figure out why ui need to patch/update an immutable crd instance. wdyt?

jingzhang36: Missed the comments here. I'll have a separate PR to clean up the patch and watch permissions, both in this file and ml-pipeline-ui-role

PR #3278, opened 2019-05-15
basic-auth: make kflogin work with NodePort

### LINKED ISSUE #3277: basic-auth: kflogin doesn't work with NodePort service
I'm trying to make basic-auth work with a NodePort Service.
The reason is because LoadBalancer is not always available in on-prem scenarios.
However, when `kflogin` posts the credentials to the gatekeeper, it doesn't specify the correct port, which makes NodePort not work.

yanniszark:
Fix the way the post request is formed so that it uses the correct
port.
Helpful for using basic-auth with a NodePort Service.

Fixes #3277

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3278)

### DIFF
--- a/components/kflogin/src/login.js
+++ b/components/kflogin/src/login.js
@@ -63,8 +63,8 @@ class Login extends Component {
 
   handleClick(){
     try {
-      const hostname = window.location.hostname;
-      axios.post("https://" + hostname + "/apikflogin", {"req-num": "1"}, {
+      const origin = window.location.origin;
+      axios.post(origin + "/apikflogin", {"req-num": "1"}, {
         auth: {
           username: this.state.username,
           password: this.state.password


components/kflogin/src/login.js
```diff
@@ -63,8 +63,8 @@ class Login extends Component {
 
   handleClick(){
     try {
-      const hostname = window.location.hostname;
-      axios.post("https://" + hostname + "/apikflogin", {"req-num": "1"}, {
+      const origin = window.location.origin;
```
chupman: It might be better to replace `window.location.hostname` with `window.location.host`.
You'll still get the port added, but I believe origin will null if you access it without going cross domain.

Issue #3282, opened 2019-05-15
[kfctl] kfctl init with ISTIO still deploys ambassador


gabrielwen:
We already removed ambassador from components when using ISTIO.  Somehow if we deploy from HEAD ambassador is deployed despite of using ISTIO.

gabrielwen:
when running `kfctl init`, with params `--platform gcp --project <project> -V --skip-init-gcp-project`, app.yaml looks like this:
```yaml
apiVersion: kfdef.apps.kubeflow.org/v1alpha1                                                                                                                                                                        
kind: KfDef
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/name: presubmit-testing
  name: presubmit-testing
  namespace: kubeflow
spec:
  appdir: /usr/local/google/home/gabrielwen/kubeflow/presubmit-testing
  componentParams:
    ambassador:
    - name: ambassadorServiceType
      value: NodePort
    argo:
    - initRequired: true
      name: injectIstio
      value: "true"
    centraldashboard:
    - initRequired: true
      name: injectIstio
      value: "true"
    cert-manager:
    - initRequired: true
      name: acmeEmail
      value: johnDoe@acme.com
    cloud-endpoints:
    - name: secretName
      value: admin-gcp-sa
    iap-ingress:
    - initRequired: true
      name: ipName
      value: ipName
    - initRequired: true
      name: hostname
      value: <deployName>.endpoints.<project>.cloud.goog
    jupyter-web-app:
    - initRequired: true
      name: injectIstio
      value: "true"
    katib:
    - initRequired: true
      name: injectIstio
      value: "true"
    notebook-controller:
    - name: injectGcpCredentials
      value: "true"
    spartakus:
    - initRequired: true
      name: usageId
      value: <randomly-generated-id>
    - initRequired: true
    - initRequired: true
      name: reportUsage
      value: "true"
    tensorboard:
    - initRequired: true
      name: injectIstio
      value: "true"
    tf-job-operator:
    - initRequired: true
      name: injectIstio
      value: "true"
  components:
  - metacontroller
  - ambassador
  - argo
  - centraldashboard
  - jupyter-web-app
  - katib
  - notebook-controller
  - pytorch-operator
  - tensorboard
  - tf-job-operator
  - pipeline
  - cert-manager
  - cloud-endpoints
  - gpu-driver
  - iap-ingress
  - spartakus
  email: gabrielwen@google.com
  packageManager: ksonnet
  packages:
[...]

PR #3283, opened 2019-05-15
Fix overlay for app.yaml

### LINKED ISSUE #3273: test_is_kf_ready fails for iap-enabler with istio
Post submit run
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kubeflow_kubeflow/kubeflow-postsubmit/1128463690513780736

The test failed waiting for iap-enabler.

Here's the exception

```
E           kubernetes.client.rest.ApiException: (404)
E           Reason: Not Found
E           HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'Audit-Id': '58ef169e-f59f-4df7-a8d5-74562f29d334', 'Content-Length': '236', 'Date': 'Wed, 15 May 2019 01:27:28 GMT'})
E           HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"deployments.extensions \"iap-enabler\" not found","reason":"NotFound","details":{"name":"iap-enabler","group":"extensions","kind":"deployments"},"code":404}
```

The calling code is

```
      ingress_namespace = "istio-system" if use_istio else namespace
      for deployment_name in ingress_related_deployments:
        logging.info("Verifying that deployment %s started...", deployment_name)
>       util.wait_for_deployment(api_client, ingress_namespace, deployment_name)

kf_is_ready_test.py:69: 
```

I suspect the problem is we are checking the wrong namespace.

It looks like it just started failing.

I suspect this is related to 
https://github.com/kubeflow/kubeflow/pull/3253

It looks like that PR changed the ksonnet parameter used for ISTIO.

So I suspect we are no longer actually enabling ISTIO.

1. kfctl_is_ready should print out the namespace we are checking as well as whether ISTIO is enabled
1. Why didn't this get caught by presubmit for #3253

gabrielwen:
- [x] ambassador overlays.
- [x] `injectIstio` in iap-ingress.
- [x] all params for pipeline.
- [x] spartakus is not GCP specific.

Fixes #3273 
Relates to #3282 

Root cause of failure is because #3283 changed parameter from `useIstio` to `injectIstio` so that istio related flags can be unified and flipped by this line:
https://github.com/kubeflow/kubeflow/blob/a70c866ee7811ec907c63368a43703447348d99f/bootstrap/pkg/kfapp/coordinator/coordinator.go#L292

However, #3108 changed kubeflow app config to use kustomize but overlay didn't include this parameter.  And thus for all ISTIO deployments `iap-ingress` will always be deployed as if it's not for ISTIO.  This PR adds back this flag.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3283)

### DIFF
--- a/bootstrap/config/base/kfctl_default.yaml
+++ b/bootstrap/config/base/kfctl_default.yaml
@@ -29,3 +29,11 @@ spec:
   - spartakus
   - tensorboard
   - tf-job-operator
+  componentParams:
+    spartakus:
+      - name: usageId
+        value: <randomly-generated-id>
+        initRequired: true
+      - name: reportUsage
+        value: "true"
+        initRequired: true
--- a/bootstrap/config/base/kfctl_default.yaml
+++ b/bootstrap/config/base/kfctl_default.yaml
@@ -27,6 +27,7 @@ spec:
   - katib
   - notebook-controller
   - pytorch-operator
+  - spartakus
   - tensorboard
   - tf-job-operator
   componentParams:
--- a/bootstrap/config/base/kfctl_default.yaml
+++ b/bootstrap/config/base/kfctl_default.yaml
@@ -20,7 +20,6 @@ spec:
   - tf-training
   components:
   - metacontroller
-  - ambassador
   - argo
   - centraldashboard
   - jupyter-web-app
@@ -30,7 +29,3 @@ spec:
   - spartakus
   - tensorboard
   - tf-job-operator
-  componentParams:
-    ambassador:
-    - name: ambassadorServiceType
-      value: NodePort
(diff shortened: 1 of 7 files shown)


jlewi:
@gabrielwen can you please update the PR description to describe the problem and how this fixes it.

bootstrap/config/base/kfctl_default.yaml
```diff
@@ -20,16 +20,20 @@ spec:
...
   components:
   - metacontroller
-  - ambassador
   - argo
   - centraldashboard
   - jupyter-web-app
   - katib
   - notebook-controller
   - pytorch-operator
+  - spartakus
```
kkasravi: if we do 
`kfctl init ~/kubeflow` we'll get spartakus - is that what we want? (it's not in kfctl_default.yaml)

kkasravi: otherwise
/lgtm

PR #3292, opened 2019-05-16
Allow defining VPC id and region for AWS alb-ingress-controller

chris922:
In case ec2metadata is not available for the alb-ingress-controller pod it is required to configure the VPC id and region.

We had this need and thus I enhanced the `aws-alb-ingress-controller` component to support defining these values.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3292)

### DIFF
--- a/kubeflow/aws/prototypes/aws-alb-ingress-controller.jsonnet
+++ b/kubeflow/aws/prototypes/aws-alb-ingress-controller.jsonnet
@@ -5,6 +5,8 @@
 // @param name string Name for the component
 // @param clusterName string Kubernetes Cluster Name
 // @optionalParam albIngressControllerImage string docker.io/amazon/aws-alb-ingress-controller:v1.1.0 ALB Ingress Controller Image.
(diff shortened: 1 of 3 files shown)


kubeflow/aws/prototypes/aws-alb-ingress-controller.jsonnet
```diff
@@ -5,6 +5,8 @@
 // @param name string Name for the component
 // @param clusterName string Kubernetes Cluster Name
 // @optionalParam albIngressControllerImage string docker.io/amazon/aws-alb-ingress-controller:v1.1.0 ALB Ingress Controller Image.
+// @optionalParam awsVpcId string null VPC ID of the cluster in case ec2metadata is not available from the controller pod
```
Jeffwan: Nice catch. Some users may use kube2iam and ec2metadata may not be available.

chris922:
Hmm, I don't know why the test still fails. Has anybody an idea what is wrong here? My branch is up-to-date with the kubeflow master

chris922:
I rebased my changes onto the master and it looks like the tests were now passed successfully 😃

PR #3294, opened 2019-05-16
Add verb "create" for Secrets to ClusterRole "pipeline-runner"

elikatsis:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3294)

### DIFF
--- a/kubeflow/pipeline/pipeline-apiserver.libsonnet
+++ b/kubeflow/pipeline/pipeline-apiserver.libsonnet
@@ -206,6 +206,8 @@
             "secrets",
           ],
           verbs: [
+            "create",
+            "delete",
             "get",
           ],
         },
--- a/kubeflow/pipeline/pipeline-apiserver.libsonnet
+++ b/kubeflow/pipeline/pipeline-apiserver.libsonnet
@@ -207,7 +207,6 @@
           ],
           verbs: [
             "create",
-            "delete",
             "get",
           ],
         },


IronPan:
Thanks @elikatsis for the change. 
I would like to understand better the use case for this change. 
Having a clusterrole with create/delete secret seems very broad permission. Especially delete.

elikatsis:
@hongye-sun, Hongye, I'm moving [the conversation](https://github.com/kubeflow/pipelines/pull/1342#issuecomment-493176421) here since it's more relevant.
@IronPan, thank you for your time, Yang.

You are totally right about the `delete` permission and sorry for that. It was a mechanistic move, I will remove it ASAP.

The causation of this PR is to run [this sample](https://github.com/kubeflow/pipelines/blob/master/samples/resourceops/resourceop_basic.py) as part of the E2E tests, because I mistakenly thought there was a `create` permission for `Secret`s.

However, I think it does make sense to be able to create a `Secret` during a pipeline and pass it to other Ops, as discussed after [Hongye's comment](https://github.com/kubeflow/pipelines/issues/801#issuecomment-463315216).
And, why not, use it for customers when serving models.

hongye-sun:
@elikatsis, thanks for your explanation. Honestly, I am still not convinced. My assumption is that user will never create secret in the pipeline, because the secret will be stored as clear text in argo's resource spec. It's not secure and should not be encouraged.

PR #3297, opened 2019-05-16
User Info badge and API Added

avdaredevil:
## About
This PR:
- Adds a user badge to the toolbar
- Mocks an API that can provide user image and email

## Preview
image

### Labels
/area front-end
/area centraldashboard
/cc @prodonjs
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3297)


components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,5 +1,5 @@
-iron-ajax(auto, url='/api/platform-info', handle-as='json',
-    on-response='_onPlatformInfoResponse')
+iron-ajax(auto, url='/api/env-info', handle-as='json',
```
prodonjs: You'll need to update the unit test that mocks this API response.

components/centraldashboard/public/components/main-page.js
```diff
@@ -90,6 +92,11 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
             hideNamespaces: {type: Boolean, value: false, readOnly: true},
             notFoundInIframe: {type: Boolean, value: false, readOnly: true},
             namespace: {type: String, observer: '_namespaceChanged'},
+            __placeholderImage: {
```
prodonjs: It doesn't look like this is being used anymore, right? If so, drop the **/public** segment from the path.

components/centraldashboard/app/api.ts
```diff
@@ -1,26 +1,48 @@
...
+
+interface EnvironmentInfo {
+    user: UserInfo;
+    platform: PlatformInfo;
+}
+
+/** Function that provides user information back */
+function getUserInfo() {
+    return {
+        email: 'blah@blah.com',
```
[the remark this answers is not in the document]
avdaredevil: Makes total sense

components/centraldashboard/public/components/main-page.pug
avdaredevil: Done!

components/centraldashboard/public/components/main-page.js
avdaredevil: Just fixed it, and used this value instead.

PR #3308, opened 2019-05-18
Fix notebook example

andreyvelich:
I think, here: https://github.com/kubeflow/kubeflow/tree/master/kubeflow/jupyter/prototypes#design yaml structure is incorrect.
According to this: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L3004:

1. `securityContext` is not a list.
2. `PodSpec` doesn't have `ttlSecondsAfterFinished` field.

/assign @kkasravi @lluunn 
/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3308)

### DIFF
--- a/kubeflow/jupyter/prototypes/README.md
+++ b/kubeflow/jupyter/prototypes/README.md
@@ -38,9 +38,8 @@ spec:
             cpu: 500m
             memory: 1Gi
         workingDir: "/home/jovyan"
-      ttlSecondsAfterFinished: 300
       securityContext:
-      - fsGroup: 100
+        fsGroup: 100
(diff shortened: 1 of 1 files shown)


kkasravi:
@andreyvelich the only approver is @lluunn so my /lgtm won't do anything.

I'll add myself to the approvers list in the OWNERS file since I had done the original implementation based on metacontroller. I'll ask @lluunn to approve the OWNERS PR.

In the meantime, @lluunn could you /lgtm? Also I opened a related [issue](https://github.com/kubeflow/manifests/issues/86) in manifests

PR #3319, opened 2019-05-20
Add details to "conditions" in notebook status

### LINKED ISSUE #2821: Using "conditions" for notebook status
In #2723 , we added a field, containerStatus to notebook CR's status. 
We can move it to conditions, and the Webui will display the latest condition.

/assign @zabbasi 

cc @jlewi @vkoukis

zabbasi:
resolves #2821
tested:
- create notebook from ui
- make sure notebook state transition is shown up under notebook.status.conditions

Do we need to remove notebook.status.containerStatus? (if yes, we need to update jupyter UI)

/cc @lluunn  @jlewi 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3319)


lluunn:
Yeah I think we should remove notebook.status.containerStatus.
We can check this in first, and then make change to the UI to read from condition, and then remove it.

PR #3325, opened 2019-05-21
allow commonlabels, commonannotations to be copied to parent kustomization.yaml

kkasravi:
if the parent kustomization does not define conflicting commonLabels or commonAnnotations, its ok to copy the overlay ones to the parent when mixing in overlays

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3325)


gabrielwen:
/lgtm

optional: could we add some logging?  DEBUG level should be fine.

PR #3326, opened 2019-05-21
Resource charts for CPU Utilization

### LINKED ISSUE #3225: [CentralDashboard] Display resource utilization for GKE deployments from Stackdriver

prodonjs:
Adds charts to display the past hour of CPU resource utilization at the Node level and the top 5 Pods.

Backend implementation is in #3251 
Closes #3225

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3326)

### DIFF
--- a/components/centraldashboard/public/components/dashboard-view.pug
+++ b/components/centraldashboard/public/components/dashboard-view.pug
@@ -10,12 +10,14 @@
                     paper-ripple
 template(is='dom-if', if='[[platformDetails.resourceChartsLink]]')
     .column
-        resource-chart(heading='Cluster CPU Utilization', metric='cpu',
+        resource-chart(header-text='Cluster CPU Utilization', metric='cpu',
             interval='Last60m',
-            external-link='[[platformDetails.resourceChartsLink]]')
-        resource-chart(heading='Pod CPU Utilization', metric='podcpu',
+            external-link='[[platformDetails.resourceChartsLink]]',
+            external-link-text='[[platformDetails.resourceChartsLinkText]]')
+        resource-chart(header-text='Pod CPU Utilization', metric='podcpu',
             interval='Last60m',
-            external-link='[[platformDetails.resourceChartsLink]]')
+            external-link='[[platformDetails.resourceChartsLink]]'
+            external-link-text='[[platformDetails.resourceChartsLinkText]]')
 .column
     template(is='dom-if', if='[[platformDetails.links]]')
         paper-card#Platform-Links(class$='[[platformDetails.name]]',
--- a/components/centraldashboard/public/components/dashboard-view.pug
+++ b/components/centraldashboard/public/components/dashboard-view.pug
@@ -8,7 +8,25 @@
                         .header [[item.text]]
                         aside(secondary) [[item.desc]]
                     paper-ripple
+template(is='dom-if', if='[[platformDetails.resourceChartsLink]]')
+    .column
+        resource-chart(heading='Cluster CPU Utilization', metric='cpu',
+            interval='Last60m',
+            external-link='[[platformDetails.resourceChartsLink]]')
+        resource-chart(heading='Pod CPU Utilization', metric='podcpu',
+            interval='Last60m',
+            external-link='[[platformDetails.resourceChartsLink]]')
 .column
+    template(is='dom-if', if='[[platformDetails.links]]')
+        paper-card#Platform-Links(class$='[[platformDetails.name]]',
+            heading='[[platformDetails.title]]', image='[[platformDetails.logo]]')
+            template(is='dom-repeat', items='[[platformDetails.links]]')
+                a.link(href$='[[item.link]]', tabindex='-1', target='_blank')
+                    paper-icon-item.external
+                        paper-ripple
+                        aside [[item.text]]
+                        paper-icon-button.button(icon='open-in-new',
+                            slot='item-icon', alt='[[item.text]]', tabindex=-1)
     paper-card#Documentation(heading='Documentation')
         template(is='dom-repeat', items='[[documentationItems]]')
             a.link(href$='[[item.link]]', tabindex='-1',
@@ -20,16 +38,3 @@
                         aside(secondary) [[item.desc]]
                     paper-icon-button.button(icon='open-in-new',
                         slot='item-icon', alt='[[item.text]]', tabindex=-1)
-.column
-    template(is='dom-if', if='[[platformDetails.links]]')
-        paper-card#Platform-Links(class$='[[platformDetails.name]]',
-            heading='[[platformDetails.title]]', image='[[platformDetails.logo]]')
-            template(is='dom-repeat', items='[[platformDetails.links]]')
-                a.link(href$='[[item.link]]', tabindex='-1', target='_blank')
-                    paper-icon-item.external
-                        paper-ripple
-                        paper-item-body(two-line)
-                            .header [[item.headerText]]
-                            aside(secondary) [[item.secondaryText]]
-                        paper-icon-button.button(icon='open-in-new',
-                            slot='item-icon', alt='[[item.text]]', tabindex=-1)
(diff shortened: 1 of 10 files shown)


prodonjs:
/hold
/cc @avdaredevil 
/cc @ajayalfred 
/area front-end
/area centraldashboard

Putting this on hold as I still need to do a bit of work, but wanted to get the ball rolling for review. You can see it live @ https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/

prodonjs:
@avdaredevil @ajayalfred 

I incorporated most of the feedback from our meeting yesterday into the latest version which is available at https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/.

I went with the filter-list menu icon for the interval selector which I think looks pretty clean as opposed to a full drop-down selector with text. The text for the external link is now dynamic based on the hosting platform, so future integrates with other Cloud providers can be plugged in.
Screenshot from 2019-05-24 10-30-53

@avdaredevil This should be in the final state ready for review. I've left the hold in place since I want to make sure the backend PR (#3251) goes in first. So you can limit your review to the front-end code.

components/centraldashboard/public/components/dashboard-view.pug
```diff
@@ -8,7 +8,27 @@
                         .header [[item.text]]
                         aside(secondary) [[item.desc]]
                     paper-ripple
+template(is='dom-if', if='[[platformDetails.resourceChartsLink]]')
+    .column
```
avdaredevil: I don't think the column should be optional, only perhaps the chart cards within. What do you think @prodonjs @ajayalfred?

components/centraldashboard/public/components/resource-chart.js
```diff
@@ -0,0 +1,352 @@
...
+        return html`
+        <style include="iron-flex iron-flex-alignment paper-material-styles">
+            :host {
+                @apply --paper-material-elevation-1;
+                background-color: var(--paper-card-background-color,
+                    var(--primary-background-color));
+                border-radius: 5px;
+                display: inline-block;
+                font-size: 14px;
+                max-width: 400px;
```
avdaredevil: Can we create a mixin called `--dashboard-card` up at dashboard level, which has these shared styles. Then each card implementation can simply do:

```css
:host {
    /* Foo Bar */
    @apply --dashboard-card;
}
```

components/centraldashboard/public/components/resource-chart.js
```diff
@@ -0,0 +1,352 @@
...
+                        <paper-item value="Last60m">1h</paper-item>
+                        <paper-item value="Last180m">3h</paper-item>
+                    </paper-listbox>
+                </paper-menu-button>
+                <paper-icon-button id="refresh" icon="refresh"
+                    title="Refresh chart" alt="Refresh chart"
+                    on-click="_refresh"></paper-button>
+            </header>
+            <section id="chart-container">
+                <canvas id="chart" height="400" width="400"></canvas></section>
```
avdaredevil:
```suggestion
                <canvas id="chart" height="400" width="400"></canvas>
            </section>
```

components/centraldashboard/public/components/resource-chart.js
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    /**
+     * Handles the Metrics response to build datasets and create the Chart.
+     * @param {Event} responseEvent
+     */
+    _onResponse(responseEvent) {
+        let i = 0;
+        this._chartOptions.data.datasets = [];
+        for (const ds of this._buildDatasets(responseEvent.detail.response)) {
```
avdaredevil: #### suggestion:
Remove lines 230-231
```suggestion
        const dataSets = this._buildDatasets(responseEvent.detail.response)
        const chartSets = dataSets.map((ds, i) => ({
            /* direct object creation and return */
        }))
        this._chartOptions.data.datasets = chartSets
```

components/centraldashboard/public/components/resource-chart_test.js
```diff
@@ -0,0 +1,215 @@
...
+            .selectedItem.innerText).toBe('30m');
+        // Validate chart properties directly
+        expect(resourceChart._chart.data.datasets.length).toBe(5);
+        expect(resourceChart._chart.options.scales.yAxes[0].ticks
+            .stepSize).toBeUndefined();
+        expect(resourceChart._chart.options.legend.display).toBe(false);
+
+        const [ds1, ds2, ds3, ds4, ds5] = resourceChart._chart.data.datasets;
+        expect(ds1.label).toBe('node-10');
+        expect(ds1.backgroundColor).toBe('#1e88e5');
```
avdaredevil: Should we be testing background colors? Assuming we want to stick to the addage:

> Test Behavior, Not Implementation **([tott](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html))**

components/centraldashboard/public/components/dashboard-view.pug
prodonjs: My thought here was that if there was no metrics provider defined for the platform, the third column would become to the second column. It does create some unused space, but unless we were to abandon the grid layout in favor of a flex layout with wrapping, I don't know that there's a simple way to resolve it.

In a non-GCP hosted scenario, here's what a user would see.
Screenshot from 2019-05-29 09-01-18

components/centraldashboard/public/components/resource-chart.js
prodonjs: That's a fantastic idea. Done.

components/centraldashboard/public/components/resource-chart.js
prodonjs: Ack'ed.

components/centraldashboard/public/components/resource-chart.js
prodonjs: Great suggestion. I sometimes forgot that the functional methods for JS arrays are more powerful and accept additional callback parameters besides the element. I also adopted your style for the implicit return for a single object in a lambda.

components/centraldashboard/public/components/resource-chart_test.js
prodonjs: Normally I wouldn't want to do so, but in this case since we are implementing the charts specification and we don't have an easy way to verify the content of the chart in test, it seemed like a good way to validate that each series has a unique color that matches expectations.

prodonjs:
@avdaredevil see my responses to your comments.

Latest commit pushed to https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/.

avdaredevil:
Merge conflicts, but otherwise:
/hold
/lgtm
/approve

ajayalfred:
Agree we want the cards to left align in the absence of cards in the middle. So anytime there's content missing to the left, move the card left.. and any time there's content missing below, move the card up.

components/centraldashboard/public/components/resource-chart_test.js
ajayalfred: Agree we want the cards to left align in the absence of cards in the middle. So anytime there's content missing to the left, move the card left.. and any time there's content missing below, move the card up.

prodonjs:
Changes are now rebased on top of master which includes the merged tweaks from our discussion last week.

Latest image is now pushed to https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/

prodonjs:
/hold cancel
@avdaredevil can you give this the lgtm again? I added the left padding to the external link icon.

PR #3334, opened 2019-05-22
ISTIO related deployments are deployed to wrong namespace

kkasravi:
- [ ] moves dependencies/istio in manifests under /manifests/gcp/{istio,istio-install}
- [ ] if useIstio == true, adds bootstrap/config/overlays/istio which will 
  - add components {istio,istio-install} to app.yaml
  - add istio as an overlay in componentParams for all components which have a virtual-service.yaml
- [ ] sets namespace to istio-system in iap-ingress, istio-install components
- [ ] removes useIstio: false componentParams
- [ ] allows a component to override --namespace in componentParams

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3334)


kkasravi:
replaced by #3375

PR #3338, opened 2019-05-22
Merge deprecated kfctl-test with kfctl-go-test

### LINKED ISSUE #3333: Install istio for the exiting kfctl e2e test
The presubmit test workflow [kfctl-test](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_test.jsonnet) uses [scripts/kfctl.sh](https://github.com/kubeflow/kubeflow/blob/master/scripts/kfctl.sh) to deploy Kubeflow and does not install istio. However currently notebook-controller have a dependency on VirtualServices. That's why the test fails for any update on the notebooks-controller image update. 
Kfctl-test performs some component specific tests which is missing in [kfctl-go-test](https://github.com/kubeflow/kubeflow/blob/master/testing/workflows/components/kfctl_go_test.jsonnet). Therefore, we need to either merge the two tests or install instio on kfctl-test.

related to #3306 
blocked #3318

/cc @gabrielwen 
/cc @kunmingg

zabbasi:
resolves #3333
kfctl-test uses scripts/kfctl.sh which is no longer maintained. However it contains components tests which is missing in the new kfctl-go-test. This pr deprecates kfctl-test  and adds component tests to  kfctl-go-test.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3338)


testing/workflows/components/workflows.libsonnet
```diff
@@ -224,7 +208,7 @@
             "--skip_tests=test_simple_tfjob_gpu",
           ],
         },  // run tests
-        dependencies: ["wait-for-kubeflow"],
+        dependencies: null,
```
kunmingg: Do we need to wait kubeflow to be ready?

zabbasi: I've just modified kfTests which is used as a sub workflow in kfctl-go-test (initially it was used at kfctl-test workflow). kfctl-go-test does not have the "wait-for"kubeflow" step. instead it has kfctl-is-ready step (see http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-kfctl-go-iap-istio-3338-7797fce-8546-8147?tab=workflow)

kunmingg: Should we depend on " kfctl-is-ready"?

PR #3345, opened 2019-05-23
SVG Iconset Dynamic Generation from Icons Assets

avdaredevil:
## About:
- Dynamic icon generated svg_set now created
- Reads `./public/assets/icons` and dynamically imports all as icons (`icon='kubeflow:<icon_name>'`)

### meta:
/assign @avdaredevil 
/cc @prodonjs 
/area front-end
/area centraldashboard
/priority p1

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3345)


components/centraldashboard/public/components/resources/kubeflow-icons.js
```diff
@@ -6,7 +6,20 @@ import '@polymer/iron-icon/iron-icon.js';
...
 import {html, htmlLiteral} from '@polymer/polymer/lib/utils/html-tag.js';
-import bolt from '../../assets/icons/ic_bolt.svg';
+
+/**
+ * Converts an imported folder into a Object.entries() type of
+ *  [[key, val], ...] list.
+ * @param {object} context
+ * @return {[...[string, string]]}
+ */
+function createFileMap(context) {
```
prodonjs: What do you think about the following? Drops the use of the lambdas and returns an object rather than a tuple for more clarity. I'd actually go a step further though and have this helper function actually build the `<g>` tag elements and rename it to something like `getIconDefinitions`.

```suggestion
function createFileMap(context) {
return context.keys().map((k) => {
        const name = k.split(/\/|\\/).slice(-1)[0].replace('.svg', '');
        const data = stripSVG(context(k).default || context(k));
        return {name, data};
    });
}
}
```

components/centraldashboard/public/components/resources/kubeflow-icons.js
```diff
@@ -16,14 +29,23 @@ import bolt from '../../assets/icons/ic_bolt.svg';
...
 }
 
+// Uses Webpack specific syntax to require all favicons
+// eslint-disable-next-line no-undef
+const importData = require.context('../../assets/icons', true, /\.svg$/);
+const fileData = createFileMap(importData).map(([name, data]) =>
+    `<g id="${name}">${data}</g>`
+).join('');
+// eslint-disable-next-line no-console
+console.log(createFileMap(importData));
```
prodonjs: Remove.

components/centraldashboard/public/components/resources/kubeflow-icons.js
```diff
@@ -6,7 +6,20 @@ import '@polymer/iron-icon/iron-icon.js';
...
+function createFileMap(context) {
+    const keys = context.keys();
+    const values = keys.map(context);
+    const _v = (v) => stripSVG(v.default || v);
+    const _name = (n) => n.split(/\/|\\/).slice(-1)[0].replace('.svg', '');
+    return keys.map((k, i) => [_name(k), _v(values[i])]);
+}
 
 /**
  * Takes SVG source returns its contents.
```
prodonjs: Can you clarify the intent of the regexes in the comment block? It seems like it they strip any excess whitespace and then trim anything but the content inside the SVG tag, but it's nice to not have to read them haha.

components/centraldashboard/public/components/resources/kubeflow-icons.js
```diff
@@ -16,14 +29,23 @@ import bolt from '../../assets/icons/ic_bolt.svg';
 function stripSVG(svg) {
     const content = svg
         .replace(/\r?\n|(?=>)\s+(?=<)/g, ' ')
-        .replace(/.*?<svg.*?>(.+?)<\/svg>/, '$1');
+        .replace(/^.*?<svg.*?>(.+?)<\/svg>.*$/, '$1');
     if (content == svg) throw Error('Invalid SVG data');
```
prodonjs: How does this behave if someone mistakenly puts a bad file in the icons folder? If it results in a page that doesn't render  on development and fails unit tests, that would be perfect. The main concern is just that we don't end up with someone being able to mistakenly put something there that gets past tests and would result in a faulty image.

components/centraldashboard/public/components/resources/kubeflow-icons.js
avdaredevil: Done, in favor of the suggestion!

prodonjs:
Did you mean to push a new commit with the comments resolved?

prodonjs:
/lgtm
Need to rebase on top of master to resolve the conflict.

components/centraldashboard/public/components/resources/kubeflow-icons.js
avdaredevil: Now comes with a verbose error in console. Should be enough to help the dev debug the issue, but not break the page itself.

components/centraldashboard/public/components/resources/kubeflow-icons.js
avdaredevil: Just did! Sorry about that :)

avdaredevil:
Added comments, and rebased! PTAL @prodonjs 

Thanks!

PR #3351, opened 2019-05-23
Improve admission-webhook to use Job (instead of Statefulset) to exchange TLS certificate

### LINKED ISSUE #3305: improve admission-webhook bootstrap server certificate creation
Current bootstrap server runs a [script](https://github.com/kubeflow/kubeflow/blob/master/kubeflow/admission-webhook/create_ca.sh) whcih on every launch does the following steps:
 1. create a CA (CA-cert, CA-key)
 1. create a server private key (server-key)
 1. create a CSR using server-key and sign it with CA-key (server-cert)
1. create a secret from server-key and server-cert
1. restart webhook server so that it uses the newly created secret
1. periodically update MutatingWebhooksConfigurations.webhooks.clientConfig.caBundle using CA-cert

 bootstrap server for the above solution requires to be in running state for the entire life-cycle of Kubeflow which results  in unnecessary overhead (computation and efficiency). 
 It can be improved using either of the following approaches:
1.  instead of periodic update, we just need to update  MutatingWebhooksConfigurations.webhooks.clientConfig.caBundle once and finish bootsrap server's job. This is because a successful run  of the script guarantees that both webhook server and MutatingWebhooksConfigurations have access to a matching secret/certificate. (i.e., webhook server will be able to do a healthy TLS communications with apiserver)

1. modify the script to create the  CA, and secret only once (no need to force restart the webhook server) and use the stored CA-cert to update MutatingWebhooksConfigurations.webhooks.clientConfig.caBundle for the first time . The script should do the steps 1-4 if the secret does not already exist. Otherwise it just need to confirm if MutatingWebhooksConfigurations.webhooks.clientConfig.caBundle  has the correct value and then exit (i.e., bootstrap server's job is complete). 

/cc @lluunn  @jlewi

zabbasi:
resolves #3305

Solution 1 in #3305 is implemented  (solution 2 is not used since could not find a way to create stateful job in k8s).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3351)


lluunn:
have you tested that if re-apply still works?

zabbasi:
> have you tested that if re-apply still works?

have deleted webhook and reapplied it and it worked.

PR #3354, opened 2019-05-24
Jupyter web app custom command

michael-balint:
Many container images have jupyter, but do not launch it in the way specified in [the Kubeflow docs](https://www.kubeflow.org/docs/notebooks/custom-notebook/), making these images incompatible as Kubeflow notebooks without modification. This PR adds a `Custom Command` field in the `New Notebook` pop-up, with the suggested default specified in the docs. Note that this will override whatever `CMD` is specified in the container.

Screen Shot 2019-05-10 at 10 15 46 AM

With this, as long as a container has jupyter, it should work with Kubeflow OOB.

This PR also contains a bug fix for where `create_notebook` was failing when `shm_enabled` wasn't checked.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3354)


prodonjs:
/cc @kimwnasptd 

@kimwnasptd can you give this a look? I don't have OWNERS approval.

Issue #3182, opened 2019-05-02
Verify Istio gateway/virtual services for different components on premise


elviraux:

yanniszark:
I've been looking into how the Istio Gateway and VirtualServices work and I would love to help with this issue.

From my understanding, the goal of this issue is to have Kubeflow working with Istio without caring for any form of auth.
In other words, make sure the `User->Gateway->KF Components` routes are functional.

My plan is the following:
1. Start from a clean master config.
2. Deploy with the default `kfctl` commands and start checking components.
3. For each component, note what is missing from `kfctl`.
4. Patch `kfctl` so that it deploys correctly with Istio.

swiftdiaries:
Awesome ! Thanks for picking this up :) 
I'll assign the issue to you 
/assign @yanniszark

swiftdiaries:
From my initial look into this, the pipelines and katib UI routes work as expected but the other virtual services don't. 
Also, I had to install Istio and Kubeflow Gateway separately. And set the istio-ingressgateway service in the istio namespace to NodePort to make it adapt for on-premise clusters.
Hope this helps :)

yanniszark:
@swiftdiaries thanks for the help!
I have made a few fixes and now it seems to work using the default installation guide.
This means it will also work for an on-prem installation.
The work is tracked in #3376

yanniszark:
I believe this can close now, these services were confirmed as part of the existing_arrikto PR (https://github.com/kubeflow/kubeflow/pull/3538)
Deploying with [existing_arrikto](https://www.kubeflow.org/docs/started/getting-started-k8s/#Kubeflow-for-Existing-Clusters---by-Arrikto) confirms that these services work in an on-prem environment.

/close

PR #3375, opened 2019-05-30
ISTIO related deployments are deployed to wrong namespace

kkasravi:
needs to be paired with https://github.com/kubeflow/manifests/pull/102

- [x] allows a component to override --namespace in componentParams
- [x] if useIstio == true, adds bootstrap/config/overlays/istio which will 
  - add istio components to app.yaml
  - add istio as an overlay in componentParams for all components which have a virtual-service.yaml
- [x] removes useIstio: false componentParams
- [x] adds cert-manager-crds (manifests#102) due to [kustomize #821](https://github.com/kubernetes-sigs/kustomize/issues/821) and [kustomize KEP #865](https://github.com/kubernetes/enhancements/pull/865)

*** ___This PR also includes breaking changes with ksonnet that require a switch to kustomize.
Moving the yaml files under dependencies/istio to manifests has no ksonnet equivalent.___ ***

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3375)

Issue #2685, opened 2019-03-12
kfctl config files should be moved to kustomization


kkasravi:
Right now under bootstrap/config we have 
gcp_prototype.yaml
kfctl_basic_auth.yaml
kfctl_default.yaml
kfctl_iap.yaml

Each file repeats a lot of information and should be an overlay on a base file. 
This can be a kustomize component.

One possibility would be to have a kustomize transformer config that references the base config but applies variants to the client spec fields as described [here](https://github.com/kubernetes-sigs/kustomize/blob/master/examples/transformerconfigs/crd/README.md)

jlewi:
Right now these files are listing which ksonnet packages and how they should be configured in different scenarios (e.g. using IAP vs. basic auth).

So I agree we'll need to come up with a corresponding solution for this problem when we migrate off ksonnet to kustomize.

jlewi:
I recall discussing this somewhere; maybe in slack? 

Is using overlays the right direction?

Overlays add a lot of complexity. In particular, overlays hurt readability because in order to understand what the actual config file would be, one has to go through the mental gymnastics of combining the overlays.

In contrast, if each YAML file provides a self-contained description of how kubeflow should be deployed then its much easier to reason about.

If we are worried about duplication in the YAML files, can we use tooling (kustomize, jsonnet, jinja, etc...) to maintain templates from which we generate the YAML files? But check in the YAML files and let a single YAML file be the input to kfctl.

This avoids pushing the complexity of generating those templates into kfctl.

The other way we can deal with duplication is to write tests that verify expectations which should be true across a set of YAML files. For example, if there a bunch of YAML files for which a bunch of ISTIO related settings should all be the same we can write a test to verify that this is true across all the configs.

kkasravi:
> The other way we can deal with duplication is to write tests that verify expectations which should be true across a set of YAML files. For example, if there a bunch of YAML files for which a bunch of ISTIO related settings should all be the same we can write a test to verify that this is true across all the configs.

we have something like this under /manifests/tests - where expected is embedded in the golang code and it's compared with generated.

> Overlays add a lot of complexity. In particular, overlays hurt readability because in order to understand what the actual config file would be, one has to go through the mental gymnastics of combining the overlays.

This has been greatly simplified to merging one or more kfctl_default.yaml
```
config
├── base
│   └── kfctl_default.yaml
└── overlays
    ├── application
    │   └── kfctl_default.yaml
    ├── basic_auth
[...]

PR #3376, opened 2019-05-30
on-prem: verify gateway/virtualservices for various 

yanniszark:
**Fixes:**
#3182

**Description:**
This PR makes changes to ensure that Istio VirtualServices are created for the various 
Kubeflow components.
In other words, verify that all components work with Istio enabled, for an on-prem scenario.

In order to test the routes, we port-forward the Istio gateway ingress service to localhost.
```bash
kubectl port-forward service/istio-ingressgateway -n istio-system 8080:80
```
I also tested with going directly to NodePort and it seems to work normally.

First, we edit kfctl to ensure that Istio is installed in case --use_istio is set.
This is a common behaviour, so it is done in the Apply function of the coordinator,
along with creating the namespace.

In short, this PR pulls common code out of the GCP platform and into every run
of kfctl apply.
After this PR, we should clean up platform GCP so it doesn't take unnecessary actions.
/cc @kkasravi 

Then, we notice that the default installation uses the ksonnet overlay in 
`config/overlays/ksonnet/kfctl_default.yaml`, which doesn't include all the necessary
componentParams for using Istio.
We add the necessary componentParams and also fix the invalid names for the minio and mysql PVs.

**Steps to test:**
1. [Checkout PR](https://help.github.com/en/articles/checking-out-pull-requests-locally)
2. Build the kfctl binary
```bash
KF_REPO=$(pwd)/kubeflow
cd $KF_REPO/bootstrap
make build-kfctl-container
```
3. Deploy according to the default guide:
```bash
cd $KF_REPO/bootstrap/bin
KFAPP=myapp
mkdir $KFAPP
export PATH=$(pwd):$PATH

kfctl init $KFAPP --verbose --repo $KF_REPO  
cd $KFAPP
kfctl generate all --verbose
kfctl apply all --verbose
```
4. Port forward the gateway ingress service:
```bash
kubectl port-forward service/istio-ingressgateway -n istio-system 8080:80
```
5. Open `http://localhost:8080/` in browser and check components.

**Components tested:**

* Notebooks
    - [x] Create a new notebook server
    - [x] Connect to notebook server
    - [x] Create a new notebook
    - [x] Run [an example](https://www.kubeflow.org/docs/notebooks/setup/) in python
* Pipelines
    - [x] Access pipelines UI
    - [x] Submit and run an example pipeline (parallel execution)
    - [x] Access Tensorboard
* TFJob
    - [x] Access dashboard
    - [x] Run a TFJob
* Katib
    - [x] Access dashboard
    - [x] Run a StudyJob
* Argo
    - [x] Access dashboard
* Grafana with Istio Dashboards
    - [x] Access grafana
* Tensorboard
    - [x] Access Tensorboard

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3376)

### DIFF
--- a/bootstrap/config/overlays/ksonnet/kfctl_default.yaml
+++ b/bootstrap/config/overlays/ksonnet/kfctl_default.yaml
@@ -4,10 +4,37 @@ metadata:
   name: config
 spec:
   componentParams:
+    argo:
+    - initRequired: true
+      name: injectIstio
+      value: "false"
+    centraldashboard:
+    - initRequired: true
+      name: injectIstio
+      value: "false"
+    jupyter-web-app:
+    - initRequired: true
+      name: injectIstio
+      value: "false"
+    katib:
+    - initRequired: true
+      name: injectIstio
+      value: "false"
     pipeline:
-    - name: mysqlPd
-      value: <deployName>-storage-metadata-store
-    - name: minioPd
-      value: <deployName>-storage-artifact-store
     - name: injectIstio
+      value: "false"
+    spartakus:
+    - initRequired: true
+      name: usageId
+      value: myrandomstring
+    - initRequired: true
+      name: reportUsage
+      value: "false"
+    tensorboard:
+    - initRequired: true
+      name: injectIstio
+      value: "false"
+    tf-job-operator:
+    - initRequired: true
+      name: injectIstio
       value: "false"
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -19,7 +19,9 @@ package coordinator
 import (
 	"fmt"
 	"github.com/ghodss/yaml"
+	bootstrap "github.com/kubeflow/kubeflow/bootstrap/cmd/bootstrap/app"
 	"github.com/kubeflow/kubeflow/bootstrap/config"
+	configtypes "github.com/kubeflow/kubeflow/bootstrap/config"
 	kfapis "github.com/kubeflow/kubeflow/bootstrap/pkg/apis"
 	kftypes "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps"
 	kfdefs "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps/kfdef/v1alpha1"
@@ -30,6 +32,8 @@ import (
 	"github.com/mitchellh/go-homedir"
 	log "github.com/sirupsen/logrus"
 	"io/ioutil"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	valid "k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"os"
@@ -97,7 +101,7 @@ func getConfigFromCache(pathDir string, kfDef *kfdefs.KfDef) ([]byte, error) {
 	if dataErr != nil {
 		return nil, &kfapis.KfError{
 			Code:    int(kfapis.INTERNAL_ERROR),
-			Message: fmt.Sprintf("can not encode as yaml Error %v", configPath, resMapErr),
+			Message: fmt.Sprintf("can not encode as yaml Error %v", dataErr),
 		}
 	}
 	return data, nil
@@ -258,17 +262,17 @@ func NewKfApp(options map[string]interface{}) (kftypes.KfApp, error) {
 			APIVersion: "kfdef.apps.kubeflow.org/v1alpha1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: appName,
+			Name:      appName,
 			Namespace: namespace,
 		},
 		Spec: kfdefs.KfDefSpec{
 			ComponentConfig: config.ComponentConfig{
-				Platform:platform,
+				Platform: platform,
 			},
-			Project: project,
+			Project:        project,
 			PackageManager: packageManager,
-			UseBasicAuth: useBasicAuth,
-			UseIstio: useIstio,
+			UseBasicAuth:   useBasicAuth,
+			UseIstio:       useIstio,
 		},
 	}
 	configFileBuffer, configFileErr := getConfigFromCache(cacheDir, kfDef)
@@ -405,6 +409,7 @@ type coordinator struct {
 }
 
 func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
+
 	platform := func() error {
 		if kfapp.KfDef.Spec.Platform != "" {
 			platform := kfapp.Platforms[kfapp.KfDef.Spec.Platform]
@@ -429,6 +434,15 @@ func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
 	}
 
 	k8s := func() error {
+
+		// Apply common functionality first
+		if err := kfapp.applyCommon(); err != nil {
+			return &kfapis.KfError{
+				Code:    int(kfapis.INTERNAL_ERROR),
+				Message: fmt.Sprintf("failed to apply common resources: %v", err),
+			}
+		}
+
 		kfapp.PackageManagers = *getPackageManagers(kfapp.KfDef)
 		for packageManagerName, packageManager := range kfapp.PackageManagers {
 			packageManagerErr := packageManager.Apply(kftypes.K8S)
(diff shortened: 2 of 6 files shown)


bootstrap/config/overlays/ksonnet/kfctl_default.yaml
```diff
@@ -4,10 +4,37 @@ metadata:
   name: config
 spec:
   componentParams:
+    argo:
```
kkasravi: we'll no longer use ksonnet in about a week. We're waiting on a few blocking PRs. 
injectIstio will also be replaced by an istio overlay and the istio resources under dependencies/istio will be deployed from manifests. So most of the changes in here will be reverted. 

If these changes stabilize main then it's a good reason to merge them, even though most of the changes won't be used after PRs 3045, 3375, and manifests 102. Let me know if that's the goal with this PR.

yanniszark: @kkasravi to my understanding, `kfctl` can support multiple package managers.
Since that seems to be the case, isn't it ok if we stabilize `ksonnet` and just leave it as is?

kkasravi: @yanniszark i think this would be difficult, having seen what happens (or doesnt' happen) between kustomize and ksonnet thus far. It wouldn't really be 'deprecated' since there would be differences in what kustomize installs and what ksonnet installs. My feeling is just getting the kustomize testing to parity with ksonnet and standardizing best practices within manifests is a higher priority than making sure ksonnet is working correctly, IMO it ends up diverting efforts towards switching to kustomize.

vkoukis: @kkasravi Hey Kam, I understand your concern about getting everything working with kustomize, and I agree. We would have made the changes directly on the kustomize package manager, if we had a solid place to start from.

But given that there is currently no support for Istio on-prem, and it is a stated goal for 0.6 that we want to support Istio uniformly and not have it as a GCP-only feature, we need to have a way to iterate now, to make it for 0.6. We need to have a solid way to use the kfctl tool not only on GCP but also on-prem, and I understand kfctl + kustomize are not yet stable in this area.

Can we get this PR merged, so we have solid ground to work on now, during the 0.6 cycle? Even it means using kfctl + ksonnet. This will allow us to move in parallel, while kfctl + kustomize stabilizes. We are happy to contribute fixes and ensure everything works with kustomize as well, but I understand the current evel of kfctl + kustomize support does not allow us to do so.

bootstrap/config/overlays/ksonnet/kfctl_default.yaml
```diff
@@ -4,10 +4,37 @@ metadata:
...
+    - initRequired: true
```
kkasravi: these componentParams were merged to master under [basic-auth](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/overlays/basic_auth/kfctl_default.yaml#L10) and [gcp](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/overlays/gcp/kfctl_default.yaml#L7)

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -457,6 +469,65 @@ func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
...
+	config := kftypes.GetConfig()
+	kubeclient := kftypes.GetClientset(config)
+
+	// Create KFApp's namespace
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: kfapp.KfDef.Namespace,
+		},
+	}
+	log.Infof("Creating namespace: %v", ns.Name)
```
kkasravi: i understand why this has been moved into coordinator, but it still remains in gcp.
For the use case where --platform gcp is provided we would attempt install istio twice

Although this fixes the on-prem use-case, won't it break others?

bootstrap/config/overlays/ksonnet/kfctl_default.yaml
kkasravi: thanks @vkoukis i believe this may address istio on-prem, but may also break when '--platform gcp'.
I've added comments in specific areas.

bootstrap/config/overlays/ksonnet/kfctl_default.yaml
yanniszark: The `gcp` overlay is, well, GCP-specific.
The `basic-auth` overlay is also GCP-specific at the moment.

In general, I don't think it makes sense for those componentParams to be elsewhere.
They are the absolutely necessary componentParams that should accompany the included components.
They should be in the default ksonnet overlay.

This also shouldn't break `basic-auth` or `gcp`.

bootstrap/pkg/kfapp/coordinator/coordinator.go
yanniszark: I didn't want to touch any GCP code in this PR, to make it as small as possible and thus easy to merge.
It also doesn't break anything.
`createNamespace` returns nil error if the Namespace is found and same with Istio.

yanniszark:
@kkasravi the reason that the tests are failing is the following:

```
util.py                     71 INFO     time="2019-06-04T22:37:51Z" level=warning msg="could not open /mnt/test-data-volume/kubeflow-presubmit-kfctl-go-basic-auth-3376-f97db0e
-7456-55a5/kfctl_test/.kube/kubeconfig Error stat /mnt/test-data-volume/kubeflow-presubmit-kfctl-go-basic-auth-3376-f97db0e-7456-55a5/kfctl_test/.kube/kubeconfig: no such file
 or directory" filename="apps/group.go:272"
util.py                     71 INFO     panic: runtime error: invalid memory address or nil pointer dereference
util.py                     71 INFO     [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x26016a1]
```

When `GetConfig()` can't find the kubeconfig file, it doesn't throw an error, but just logs the error with level=warning (should be error) and returns a nil pointer.
So the problems here are:

1. Why are we failing to read the kubeconfig file from the CI environment? This is what's blocking the PR.
2. `GetConfig()` should be fixed to either return an error or panic. Tracked in #3409

kkasravi:
@yanniszark 

kfctl will use gcloud credentials or kube/.config credentials depending on whether `--platform gcp` is called. Moving `common code` up into coordinator results in an attempt to create a k8 client before the the gcloud credentials are created. 

the other impact is that we break the encapsulation of coordinator which now begins to create k8 resources normally delegated to platform or package-manager

jlewi:
I have some questions about this PR

> In short, this PR pulls common code out of the GCP platform and into every run
of kfctl apply.
After this PR, we should clean up platform GCP so it doesn't take unnecessary actions.

What is the GCP code you are pulling out ? Is it just the ISTIO install code.

As mentioned in a couple of comments on slack and other places; after thinking about it I'm leaning towards the idea that installing ISTIO should be platform dependent and not done when "--platform=None."

What does your test platform actually look like? Are you running on MiniKF? IIUC minikf is already doing a bunch of setup outside of kfctl; e.g. creating a K8s cluster doing various config etc... So did you consider having your setup code for minikf install ISTIO before calling 

kfctl init --use_istio=true --platform=None

Is installing ISTIO the only issue with kustomize --master or were there other problems you encountered?

If you are running with ksonnet is this really fixing #3182?  Won't we have to repeat this process to verify its working with kustomize to ensure we have it fixed for 0.6?

See discussion in kubeflow/kubeflow#2685. Does my point there about not using overlays to generate the YAML files change anything?

bootstrap/config/overlays/ksonnet/kfctl_default.yaml
jlewi: @vkoukis Can you clarify what you mean by 

>  Istio uniformly and not have it as a GCP-only feature, 

As I mentioned in slack after thinking about it following the community meeting I do think it might make sense for ISTIO installation to be tied to the platform and the responsibility of whoever owns the corresponding platform.

ISTIO is heavily tied to the platform; IMO.

The only other support needed for ISTIO should be the definition of virtual services and those are all platform agnostic.

bootstrap/config/overlays/ksonnet/kfctl_default.yaml
jlewi: @yanniszark See my comment in https://github.com/kubeflow/kubeflow/issues/2685 about not using the overlays to generate the KfDef YAML files but instead using a self contained YAML file for each platform/config combo we want to support with a clear set of owners.

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -457,6 +471,65 @@ func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
...
+	if err != nil && !errors.IsAlreadyExists(err) {
+		log.Errorf("Error creating namespace %v", ns.Name)
+		return &kfapis.KfError{
+			Code:    int(kfapis.INTERNAL_ERROR),
+			Message: err.Error(),
+		}
+	}
+
+	// Apply Istio if use_istio option is set
+	if kfapp.KfDef.Spec.UseIstio {
```
jlewi: I think installing istio when --platform=None will cause problems.

I think lots of people will want to install KF on their existing cluster which already has ISTIO installed and don't want us to mess with it; see for example
https://github.com/kubeflow/kubeflow/issues/3396

yanniszark:
@jlewi  thanks for all the thoughts on how to install Istio!
It's a tricky issue and I have been talking with @kkasravi  to better understand
how kfctl is supposed to work.
From that discussion, here are some agreed-upon points:

* The platform step shouldn't be applying yaml to Kubernetes.
In other words, Istio should be installed as a Kubeflow component.
This thought is also reflected by the #3375 PR, which moves the Istio installation
from applying manifests under dependencies/istio, to the streamlined component installation
process we use for everything else.

Regarding leaving installation of Istio up to the platform:

* I agree with you that a platform might want to handle the Istio installation
differently (eg managed Istio on GKE). As Istio installation, I define the Istio CRDs + accompanying controllers
(Istio Control Plane).
For example, GKE might want to use managed Istio.
To accomodate that, we could have a cli flag `--install-default-istio`,
which will dictate if `kfctl` should install the default Istio control plane.
This will be translated to adding extra components (istio-crds, istio-install) to the KfDef definition.
What do you think of a solution like this ?

* For applying Istio objects dependent on the platform (eg ServiceEntries for google APIs),
I consider that as using Istio vs installing Istio.
I believe these should be bundled in a different component which will be included by the respective
platform.


Why I began with ksonnet:

* I am running with ksonnet because that's what is stable and currently working.
However, I agree that it should work for kustomize and I have also began testing
#3375 to ensure that it works for Istio on-prem

About `--platform=None`:

* There is currently no platform None in the Golang kfctl.
Specifying `--platform=None` will give an error.
Instead, we don't specify a platform and only the k8s step is done.
Is this what you meant?

Regarding not using overlays to generate the YAML files:

* If we go back to using just YAML files for the config of each platform,
we just need to make sure that they contain the right components.
That should be enough.

jlewi:
@yanniszark @vkoukis Thanks for the good discussion in slack this morning. Would you mind summarizing here what you think the next steps for the PR is.

> The platform step shouldn't be applying yaml to Kubernetes.
In other words, Istio should be installed as a Kubeflow component.
This thought is also reflected by the #3375 PR, which moves the Istio installation
from applying manifests under dependencies/istio, to the streamlined component installation
process we use for everything else.

This is overly complicated so let me try to shed some light. The meaning of "platform" is overloaded. IT is really two things.

1. It is a "phase" during the generate and apply process
    * Certain resources like the K8s cluster need to be created before other resources like the K8s resources themselves
    * So we divided each process into two phases platform and k8s

1. A specific platform e.g "gcp" is an oppionated way of configuring Kubeflow.
  
    * Every phase of deployment (e.g. generate/apply) can invoke different logic based on the platform
    * The intent is to give platform owners the ability to customize things
   
The two definitions are in conflict here.

* We (Google) really only felt qualified to have an oppinion about the installation/configuration of ISTIO on the configs we own (i.e. --platform=None). 
* The reason we installed it during the platform phase was because the original k8s phase was using ksonnet and converting the ISTIO yaml specs to ksonnet was unruly; so we decided to just do it as part of the platform phase
* With the switch to kustomize its pretty easy to turn it into a kustomize package so that's what @kkasravi  did and now its installed during the k8s phase (when it is)

* This makes it very easy to treat ISTIO as an application that platform owners can choose to install or not
   * Per discussion in slack; I think ISTIO should still be tied to a platform because whether or not to install it is an oppinion about how kubeflow should be installed.

This should make it really easy for platform owners to configure ISTIO or other compatible service meshes as we see fit
 
e.g on GCP when we are ready to switch to managed istio we will

* remove the ISTIO app from kfctl_gcp_istio.yaml
* Turn on managed ISTIO in our GKE deployment manager configs

yanniszark:
The services were confirmed as part of the existing_arrikto PR (https://github.com/kubeflow/kubeflow/pull/3538).

Issue #3236, opened 2019-05-09
Add Pipelines card to provide one-click access to KF Pipeline runs


prodonjs:
As part of the UX enhancements [specifications](https://gallery.googleplex.com/projects/MCHbtQVoQ2HCZYbjDGnaKqEC/files/MCEJu8Y2hyDScWR0uwkSmYmqTA1fRs4iJFc), the KF dashboard should have a card that shows recently executed KF Pipeline runs allowing the user to access current or recently active runs from the KF landing page.with single click. This mirrors to the functionality described in #3120 for Jupyter Notebooks. 

/area front-end
/area centraldashboard
/assign @prodonjs
/priority p2

prodonjs:
/cc @avdaredevil 
/cc @ajayalfred 

Wanted to run a quick set of questions by you. Kubeflow Pipelines has two major artifact types; Pipeline definitions that define a workflow graph to be executed in the Kubernetes engine, and Pipeline runs that are instantiations of those definitions run at a particular time. In my mind, it seems like it would be useful to have quick access to both.

This screenshot below shows a local change I'm working on where the Pipeline Definitions card shows the 5 most recently created Pipeline definitions, and the Pipeline Runs card shows the runs. 
![Dhx9kD179hQ](https://user-images.githubusercontent.com/6835846/58655264-bc511f00-82e7-11e9-99c6-db0998139ad8.png)

- Do you think this approach makes sense or should there only be one Pipelines card?
- If there's only one, should it show definitions, runs, or provide a toggle to switch between them?
- For runs, we can obtain the status, so how should we represent that?
  - My first thought would be to color the icon green for success, red for failure, and some form of yellow/orange if pending.

ajayalfred:
Great suggestion! I think we should show them as 2 distinct cards.

For the pipeline definition card, the title can read "Pipelines" and show the 5 most recently used/modified/uploaded pipelines.

For pipeline runs card, the title can read "Pipeline Runs" and show the 5 most recent runs. Agree showing status will be useful. Instead of the pipeline icon to the left of each run, show the status in that space. We use a different icon to represent runs anyway, but status will definitely be more useful. Status icons for KFP available here –https://drive.google.com/corp/drive/folders/1dDrRWSFSM4UQNZLzWJy2Df8Uez4bGj30

Finally, can you stack the pipeline and pipeline runs card vertically in one column. They provide contextual information for each other that works best in conjunction with being placed together.

prodonjs:
[...]

PR #3377, opened 2019-05-30
Adds recent Jupyter Notebooks card to Central Dashboard

### LINKED ISSUE #3120: Add Notebooks card to provide one-click access to Juypter Notebooks
Currently, the user journey to access a Jupyter Notebook in KF requires a user to connect to the appropriate Jupyter server from the Notebook Servers UI, and then click on the notebook they wish to open. This requires at least three clicks from the Kubeflow central dashboard and ends up opening up two additional tabs.

An optimized experience would allow the user to see their most recently modified Notebooks and begin working in them from the context of the KF dashboard with a single click.

/area front-end
/area centraldashboard
/assign @prodonjs

prodonjs:
closes #3120 

/area front-end
/area centraldashboard
/cc @avdaredevil 
/cc @ajayalfred

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3377)


prodonjs:
It's starting to look like we have a real dashboard on our hands here! 

The image with these changes is pushed to https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/.

A couple of things to note:
- Jupyter servers are namespaced, so the card is not populated until a namespace is chosen.
- I use a generic material horizontal loading indicator when the requests to get Notebooks are pending. I don't think we have a spec for that so feel free to provide a better suggestion for giving some kind of activity indicator.
- Results are limited to the 5 most recently modified notebooks. That seems to keep the card height where we'd want it. But, since our API requests to the Jupyter servers return all results, another option would be to add a pagination control to the card's footer.

ajayalfred:
Looks hot!!!

image
Change the empty state message to "No Notebooks in namespace (name of namespace)"

The horizontal loader looks great. I noticed it's a few pixels below the pin line though. Can you make it appear just under the pinline? See spec below
image

Regarding no. of results – I think we can start with 5 most recent to begin with. I wouldn't want to add pagination in the dashboard since it'll feel heavy handed and the idea of a dashboard is to be a quick glance-and-bounce page. We could consider increasing the no. of results to 10 since the dashboard is pretty empty at the moment. That said, it's likely that the user will find their most recently used notebook in the first 5 results :)

prodonjs:
@ajayalfred Changes to the loading indicator and the column spacing are in place and should be live in https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/.

@avdaredevil The latest commit (bf1ac3b) fixes an issue I noticed whereby the profile image was trying to load from /undefined. It took a little digging but apparently the `iron-image` component explicitly checks for an [empty string](https://github.com/PolymerElements/iron-image/blob/master/iron-image.js#L318) rather than a falsy value so because the user.image property was undefined, it was trying to fetch it. There are other ways of addressing it but this seemed to be the simplest.

LMK if there's anything else.

prodonjs:
/hold

There's a CSS issue with the dashboard-view now that I removed width: 100%. The scrollbar won't be at the far-right once we have more vertical content. However, please review the other JS changes.

prodonjs:
/hold cancel
This should be ready for final review. @ajayalfred I addressed some of the points you made on issue #3236 with respect to moving the charts to the first column and the notebooks card to the second. I also forced the x-axis labels to remain horizontal and shrunk the secondary text on each card item so that it's consistent across the board.

Changes are pushed the cluster I have linked to previously. The middle column looks a little bare when you first load the page before selecting a namespace, but that will improve once we add the two Pipelines cards which I should have a PR out today for.

components/centraldashboard/public/components/dashboard-view.css
```diff
@@ -31,13 +22,32 @@
...
+    display: grid;
+    grid-gap: .75em; /* 12px */
+    grid-template-columns: repeat(3, 1fr);
+    margin: 0 auto;
+}
+
+.column {
+    grid-row: 1;
+    min-width: 320px;
+    max-width: 400px;
```
avdaredevil: If we have this logic defined here, all cards should no longer have this logic, and instead should be:
```css
width: 100%
```

components/centraldashboard/public/components/dashboard-view.pug
```diff
@@ -1,44 +1,47 @@
...
-            external-link-text='[[platformDetails.resourceChartsLinkText]]')
-.column
-    template(is='dom-if', if='[[platformDetails.links]]')
-        paper-card#Platform-Links(class$='[[platformDetails.name]]',
-            heading='[[platformDetails.title]]', image='[[platformDetails.logo]]')
-            template(is='dom-repeat', items='[[platformDetails.links]]')
-                a.link(href$='[[item.link]]', tabindex='-1', target='_blank')
+        paper-card#Quick-Links(heading='Quick shortcuts')
+            template(is='dom-repeat', items='[[quickLinks]]')
+                iframe-link.link(role='link', href$="[[item.link]]")
```
avdaredevil: Instead of providing role here, can we have the `role` defined from within the component, defaulted to `link`?

prodonjs: Good call. For some reason the idea of setting the attribute in the component's setup eluded me. Thanks for the suggestion!

components/centraldashboard/public/components/dashboard-view.css
prodonjs: Agreed. I think I took care of this already by removing the widths from the `--dashboard-card` mix-in.

avdaredevil:
Hey @prodonjs made a quick change, lemme know your thoughts?

components/centraldashboard/public/components/dashboard-view.css
avdaredevil: Ah, I may have missed that. Feel free to resolve

avdaredevil:
/hold
For review from @prodonjs 

/lgtm
/approve

prodonjs:
I like it!

/hold cancel

ajayalfred:
Loader position looks perfect! 
The blue and gray color don't appear to match the spec. Can you change the blue to #669DF6 and the gray container to #F1F3F4? Thanks!

prodonjs:
Yes, I will get those colors fixed on #3379 .

PR #3381, opened 2019-05-31
WIP store junits in subdirectories 

zabbasi:
related to https://github.com/kubeflow/testing/issues/372

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3381)


jlewi:
@zabbasi Any update on this? Based on our discussion, I think the suggestion was to repeat this using a simpler repository like kubeflow/testing and moving all tests into subdirectories so that we could then verify that in gubernator all junit files were properly processed.

zabbasi:
> @zabbasi Any update on this? Based on our discussion, I think the suggestion was to repeat this using a simpler repository like kubeflow/testing and moving all tests into subdirectories so that we could then verify that in gubernator all junit files were properly processed.

Verified it and realizes gubernator wont pul out the junits which are stored in subdirectories (this also has been already confirmed by an gubernator engineer).

zabbasi:
Here are some notes:
- Prow supports having junit xml files in subdirectories and it will upload everything in the $ARTIFACTS directory. Prow does not write junit, that is the responsibility of the tests. Junit (and all other files) in the $ARTIFACTS directory are uploaded by the sidecar pod utility: https://github.com/kubernetes/test-infra/tree/master/prow/sidecar

- TestGrid reads any nested folders in $Artifacts 

- gubernator has the ability to read nested folders in $Artifact, however the running instance is currently configured with flag `recursive=false` which means that it does not read junits if they are stored in subfolders. 
  -- https://github.com/kubernetes/test-infra/blob/master/gubernator/view_build.py#L150
  -- https://github.com/kubernetes/test-infra/blob/master/gubernator/config.yaml#L42

/cc @jlewi

PR #3394, opened 2019-06-03
Add Notebook CRD validation

zabbasi:
related to #2417

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3394)


zabbasi:
@lluunn  looks like there is already [notebook CRD validation manifest] (https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/config/crds/notebook_v1alpha1_notebook.yaml) Not sure if it is used when deploying notebook resources though? Would you recommend to close this PR?

zabbasi:
No need to add validation to ksonnet component. Will use the [auto generated validation manifest](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/config/crds/notebook_v1alpha1_notebook.yaml) for Ksutomize component.

PR #3395, opened 2019-06-04
Fix make failing to 'build' in bootstrap/

LucidNeko:
```
# github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/coordinator
pkg/kfapp/coordinator/coordinator.go:100:13: Sprintf call needs 1 arg but has 2 args
...
Makefile:141: recipe for target 'build' failed
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3395)


yanniszark:
Hi @LucidNeko, thanks for your PR!
It seems like a dupe of #3365.
Can you confirm?

LucidNeko:
Hi @yanniszark, ah had missed that one.
Yes, it looks like a dupe - and on closer reading of the surrounding code #3365 is the correct solution.
I'll close this off - have a good day!

PR #3402, opened 2019-06-04
Fix multiple iframe-link behaviors.

prodonjs:
- Removes redundant forwarder for iframe-links. This was resulting in pushing
  the same state to history twice, which would require the user to issue two
  back button clicks in their browser if they wished to return to the previous URL.
- In addition, we now set the iframe's location programmatically rather than by
  changing the src attribute, which could also result in history states being
  pushed inadvertantly.
- We also avoid passing the ns query parameter down to avoid unncessary reloads
  of iframed-content in favor of using our message passing library.

/area centraldashboard
/area front-end
/cc @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3402)

### DIFF
--- a/components/centraldashboard/.eslintrc.json
+++ b/components/centraldashboard/.eslintrc.json
@@ -14,6 +14,7 @@
         "SharedArrayBuffer": "readonly"
     },
     "parserOptions": {
+        "ecmaVersion": 2017,
         "sourceType": "module",
         "ecmaFeatures": {
             "impliedStrict": true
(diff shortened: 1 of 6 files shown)


prodonjs:
These changes are available to test in https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog/

prodonjs:
@avdaredevil Commit a8f0fc1 enables the Crtl-click behavior Jon pointed out during our bug bash. I think our original implementation relied on some assumptions that are no longer applicable, so we can let the anchor tag exhibit it's normal behavior and our app-location component handles the client-side routing without the need for us to fire a custom event.

I also looked into the issue on the Jupyter Notebook Servers UI. It does not look like the `change` event gets fired when we set the namespace value through the SDK. That explains why the list of servers does not change. This seems consistent with how the change event behaves vs. the input event.

avdaredevil:
/lgtm
/approve

Works great, the back button seems to behave predictably as well!

Issue #2766, opened 2019-03-21
make basic auth works with Istio


lluunn:
Currently only iap part has changes for using istio

gabrielwen:
@kunmingg should we close this?  currently basic auth could work with ISTIO.  Should we close it or should we wait until basic auth doesn't need ambassador?

kunmingg:
@gabrielwen 
Does basic auth work with ISTIO on master head now?
Can you share the command?

gabrielwen:
I think what I did is:
`kfctl init basic-auth-testing --platform gcp --project gabrielwen-learning -V --skip-init-gcp-project --use_istio --use_basic_auth`
`kfctl generate all -V`
`kfctl apply all -V`

jlewi:
@gabrielwen @kunmingg  Can we update the E2E test that is using basic auth to use ISTIO? If that test passes then we know basic auth is working with ISTIO right?

jlewi:
Here's a passing postsubmit for basic auth;
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kubeflow_kubeflow/kubeflow-postsubmit/1135679782407639040

Is this not using ISTIO?

gabrielwen:
No, it's not currently:
https://github.com/kubeflow/kubeflow/blob/4fda196a8a6cb59ec6ada75952b4762e51fcd1af/prow_config.yaml#L65

gabrielwen:
basic auth is broken for this:
https://github.com/kubeflow/kubeflow/blob/b989f84457ec88aaa543a21e864060ea74c71536/kubeflow/gcp/update_backend.sh#L18

port name for ISTIO is indeed `http2`, but for ambassador is `ambassador`.  fixing it.

gabrielwen:
After #3404 we also need to migrate the changes to kustomize.

@lluunn

jlewi:
Do we have an E2E test that we can use for basic-auth + ISTIO + kustomize?

lluunn:
With https://github.com/kubeflow/kubeflow/pull/3404, we have basic auth + istio. But it's not using kustomize yet i think.

jlewi:
Thanks @lluunn . Lets update the test to use auth+istio+kustomize and close this issue when it passes.

lluunn:
I think we need to port the change of update_backend.sh, setup_backend.sh in https://github.com/kubeflow/kubeflow/pull/3404 to kustomize manifest
[...]

PR #3404, opened 2019-06-04
Make basic auth works with ISTIO - JSONNET

gabrielwen:
Relates to #2766
- [x] Fixes ambassador mapping for basic auth.
- [x] Update test_jsonnet for ambassador to test ISTIO.
- [x] Change ambassador namespace to istio namespace if using istio.
- [x] Change basic auth ingress namespace to istio namespace if using istio.
- [x] Fixes backend updater when using ISTIO with basic auth.
- [x] Adds mapping service from ambassador to ISTIO mesh for all traffic except health check (/whoami) and login service (/kflogin).
- [x] Flip the flag to use ISTIO with basic auth in E2E test.

With this PR ingress traffic with basic auth looks like:
```
Client  --->  ambassador  --->  ISTIO service mesh (if having valid token)
                     |    \
                     |      \
                     |        \
                basic-auth       login-ui
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3404)


kubeflow/common/prototypes/ambassador.jsonnet
```diff
@@ -8,6 +8,8 @@
 // @optionalParam ambassadorNodePort number 0 Optional nodePort to use when ambassadorServiceType is NodePort {30000-32767}.
 // @optionalParam ambassadorImage string quay.io/datawire/ambassador:0.37.0 The image for the API Gateway.
 // @optionalParam replicas number 3 The number of replicas.
+// @optionalParam injectIstio string false The namespace where Istio is installed
```
lluunn: wrong comment

kubeflow/gcp/basic-auth-ingress.libsonnet
```diff
@@ -430,6 +436,45 @@
     ),
     cloudEndpoint:: cloudEndpoint,
 
+    local istioMappingSvc = {
+      apiVersion: "v1",
```
lluunn: This service seems not matching to any deployment. Is it only for the annotation?

kubeflow/gcp/basic-auth-ingress.libsonnet
```diff
@@ -430,6 +436,45 @@
...
+        namespace: namespace,
+        annotations: {
+          "getambassador.io/config":
+            std.join("\n", [
+              "---",
+              "apiVersion: ambassador/v0",
+              "kind:  Mapping",
+              "name: istio-mapping",
+              "prefix_regex: true",
+              "prefix: /(?!whoami|kflogin).*",
```
lluunn: So this means whoami or kflogin will both be directed to istio-ingressgateway?
Why do we care about whoami here?
Isn't why kflogin goes to istio-ingressgateway?

kubeflow/gcp/basic-auth-ingress.libsonnet
```diff
@@ -430,6 +436,45 @@
...
+              "rewrite: \"\"",
```
lluunn: this is "" ?
Why should rewrite be this?

gabrielwen: rewrite being this so that it's not rewritten.  this is a hack to redirect all traffic to istio ingress gateway and the actual rewrite will be there, according to virtual service configs.

kubeflow/gcp/basic-auth-ingress.libsonnet
gabrielwen: this means anything except /whoami and /kflogin will match.

kubeflow/gcp/basic-auth-ingress.libsonnet
gabrielwen: yes.

kubeflow/common/prototypes/ambassador.jsonnet
gabrielwen: updated.

kubeflow/gcp/basic-auth-ingress.libsonnet
lluunn: ah I see, thanks

kubeflow/gcp/basic-auth-ingress.libsonnet
lluunn: If we don't want rewrite, can we remove this line?
Or it won't work?

gabrielwen: If we don't put this line, default will rewrite prefix to / and it will be wrong when passing to ISTIO.

PR #3408, opened 2019-06-04
Adds new service config for tf-operator

krishnadurai:
Moves port config to tf-operator service instead of deployment
Moves Prometheus config to service as well

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3408)

### DIFF
--- a/kubeflow/tf-training/prototypes/tf-job-operator.jsonnet
+++ b/kubeflow/tf-training/prototypes/tf-job-operator.jsonnet
@@ -13,7 +13,6 @@
 // @optionalParam injectIstio string false Whether to inject istio sidecar; should be true or false.
 // @optionalParam clusterDomain string cluster.local DNS config to cluster domain.
 // @optionalParam monitoringPort string 8443 Port for monitoring agent to scrape metrics from.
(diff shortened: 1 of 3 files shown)


kubeflow/tf-training/prototypes/tf-job-operator.jsonnet
```diff
@@ -13,6 +13,7 @@
 // @optionalParam injectIstio string false Whether to inject istio sidecar; should be true or false.
 // @optionalParam clusterDomain string cluster.local DNS config to cluster domain.
 // @optionalParam monitoringPort string 8443 Port for monitoring agent to scrape metrics from.
+// @optionalParam enablePrometheus string false If set true, enable prometheus monitoring metrics reporting.
```
johnugeorge: Since monitoring is enabled by default, does this flag add any extra meaning?

richardsliu: Agree - I think we should just have this enabled. Also kustomize doesn't have the notion of boolean flags.

Issue #3348, opened 2019-05-23
Add namespace retrieval to environment info API route


prodonjs:
Namespace selector is a zero-state component so rather than load it with a separate AJAX request, it should be returned alongside the user and platform information that is returned on first-load.

/area front-end
/area centraldashboard
/assign @prodonjs 
/cc @avdaredevil

avdaredevil:
Should we consider an interval refresh, in case namespaces, get added or removed, or are those pretty major events (that won't happen too frequently) thus we should expect a reload in that case?

prodonjs:
I'm assuming that once the UI integration with the Profile Controller is in place, namespace creation should be an event that we are managing. Even then, I wouldn't imagine it's something that a particular user would do all that often where we'd want to grab new namespaces repeatedly.

The other piece that would be interesting to know is whether we only want to show namespaces with Kubeflow items deployed in them. Many of the other namespaces are either for Kubernetes internal components (kube-system) or are not used by the application (default, kube-public). It would be nice if we only showed namespaces that could be used for Kubeflow artifacts.

PR #3420, opened 2019-06-06
Consolidate initial state retrieval and retrieve user email from IAP headers

prodonjs:
- Eliminates separate call to fetch namespaces.
- Retrieves user email address from deployments secured by IAP.

/area front-end
/area centraldashboard
/cc @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3420)

### DIFF
--- a/components/centraldashboard/app/api.ts
+++ b/components/centraldashboard/app/api.ts
@@ -3,72 +3,100 @@ import express from 'express';
 import {KubernetesService, PlatformInfo} from './k8s_service';
 import {Interval, MetricsService} from './metrics_service';
 
-interface UserInfo {
-  email: string;
-  image: string;
-}
-
 interface EnvironmentInfo {
-  user: UserInfo;
+  namespaces: string[];
   platform: PlatformInfo;
+  user: string;
 }
 
-/** Function that provides user information back */
-function getUserInfo() {
-  return {
-    email: 'user@kubeflow.org',
-  };
-}
+const IAP_HEADER = 'X-Goog-Authenticated-User-Email';
+const IAP_PREFIX = 'accounts.google.com:';
 
-/** API routes exposing information from Kubernetes */
-export function api(
-    k8sService: KubernetesService,
-    metricsService?: MetricsService): express.Router {
-  return express.Router()
-      .get(
-          '/env-info',
-          async (_: express.Request, res: express.Response) => {
-            const [platform, user] = await Promise.all([
-              k8sService.getPlatformInfo(),
-              getUserInfo(),
-            ]);
-            const returnPayoad = {platform, user} as EnvironmentInfo;
-            res.json(returnPayoad);
-          })
-      .get(
-          '/metrics/:type((node|podcpu|podmem))',
-          async (req: express.Request, res: express.Response) => {
-            let interval = Interval.Last15m;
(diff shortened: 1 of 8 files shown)


prodonjs:
closes #3348 

Validate behavior in: https://kf-dev.endpoints.prodonjs-kubeflow-dev.cloud.goog

components/centraldashboard/app/api.ts
```diff
@@ -3,72 +3,100 @@ import express from 'express';
...
+    return this.platformInfo;
+  }
+
+  /**
+   * Retrieves user information from headers.
+   * Supports:
+   *  GCP IAP (https://cloud.google.com/iap/docs/identity-howto)
+   */
+  private getUser(req: express.Request): string {
+    let email = 'anonymous@kubeflow.org';
```
avdaredevil: _**nit**_ Code depth
```suggestion
    const iap = req.header(IAP_HEADER)
    return iap? iap.slice(IAP_PREFIX.length) : 'anonymous@kubeflow.org';
```

components/centraldashboard/app/api.ts
```diff
@@ -3,72 +3,100 @@ import express from 'express';
...
+              console.log('getting env-info');
+              const [platform, user, namespaces] = await Promise.all([
+                this.getPlatformInfo(),
+                this.getUser(req),
+                this.k8sService.getNamespaces(),
+              ]);
+              res.json({
+                platform,
+                user,
+                namespaces: namespaces.map((n) => n.metadata.name),
```
avdaredevil: Keep in mind this needs to be fresh as of 30 seconds, as per @kunmingg's suggestion. So we may need another API to keep this fresh, or use sockets

components/centraldashboard/app/api.ts
```diff
@@ -3,72 +3,100 @@ import express from 'express';
...
+                    res.json(await this.metricsService.getPodCpuUtilization(
+                        interval));
+                    break;
+                  case 'podmem':
+                    res.json(
+                        await this.metricsService.getPodMemoryUsage(interval));
+                    break;
+                  default:
+                }
+              } else {
```
avdaredevil: **_code depth--_** Remove `else`, put a return after `switch`

components/centraldashboard/app/api.ts
prodonjs: What specifically needs to be fresh, the list of namespaces? In that case we still have the dedicated namespaces endpoint to get the latest content back.

components/centraldashboard/app/api.ts
prodonjs: Good call. Refactored to early return.

components/centraldashboard/app/api.ts
prodonjs: I was expecting that eventually there might be additional headers that would be interrogated to support other authentication providers besides IAP.

avdaredevil: I see, sure, we can leave it as it is.

components/centraldashboard/app/api.ts
avdaredevil: Yup, it was more an FYI, than an ask for change. I thought your commit message said we removed that endpoint.

PR #3429, opened 2019-06-09
Enable applications

kkasravi:
- [x] add application overlay to bootstrap/config

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3429)


jlewi:
@kkasravi Why are we doing this as an overlay and adding a flag? Why not just always install the application controller?

jlewi:
Here's an image I copied over from sigapps for the application controller.
gcr.io/kubeflow-images-public/kubernetes-sigs/application:1.0-beta 

Assuming we use that lets make a note somewhere that it was obtained manually from the sig apps team @ Google.

bootstrap/config/overlays/application/kfctl_default.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: kfdef.apps.kubeflow.org/v1alpha1
+kind: KfDef
+metadata:
+  name: config
+spec:
+  components:
+  - application
+  packages:
+  - application
+  componentParams:
```
jlewi: Why make this an overlay? Why not just add it to the default config file?

bootstrap/cmd/kfctl/cmd/init.go
```diff
@@ -62,6 +62,7 @@ or a <name>. If just <name>, a directory <name> will be created in the current d
 
 		useIstio := initCfg.GetBool(string(kftypes.USE_ISTIO))
 		disableUsageReport := initCfg.GetBool(string(kftypes.DISABLE_USAGE_REPORT))
+		enableApplications := initCfg.GetBool(string(kftypes.ENABLE_APPLICATIONS))
```
jlewi: Do we need to add another flag? Can we just make it required for now?

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -299,6 +303,9 @@ func NewKfApp(options map[string]interface{}) (kftypes.KfApp, error) {
 	kfDef.Spec.UseBasicAuth = useBasicAuth
 	kfDef.Spec.UseIstio = useIstio
 	kfDef.Spec.PackageManager = packageManager
+	kfDef.Spec.EnableApplications = true
```
hougangliu: dup with line 278
BTW, do this mean `kfDef.Spec.EnableApplications` is always true? if yes, I think we may not need this flag and just always include `application` into overlays

jlewi:
> my thoughts are application CRs and application-controller will vary and expose differents ways to group applications, different ways to surface INFO items, different relationships among applications. For example, the default application-controller doesn't interpret the various recommended labels attached to each resource. As documented, the default application-controller is minimal.

Could you explain what you mean here? 

IIUC there are two different issues here

1. The Application CRs and whether these should be an overlay in the kustomize manifests
1. Whether deployment of application controller should be optional

One of my primary goals of introducing application CRs is to provide a common and straightforward way for users to answer basic questions like

* Which applications are installed?
* Are the applications healthy?

So I lean towards making the application controller and CRs installed by default and thus not overlays.

Per discussion elsewhere,  I think we are planning on moving to using a single static YAML file containing the complete KfDef as input to kfctl.

So if a user wants to customize the application controller they can do so by customizing their KFDef file or by modifying the kustomize manifest for application controller.

Given the above

1. Should we not add flag --enable-applications to kfctl? 
1. Should we not add an overlay "application" to the kustomize manifests and instead add them to the base of each manifest?

Here are the potential problems I could possibly forsee and their mitigation.

1. User doesn't want to or can't install application controller
   
    * Our kustomize manifests technically only have a dependency on the Application CRD resource
    * So it would be sufficient for users to create the Application CRD without installing any controller
    * Users could hack the kustomize manifests and remove the Application CRs.

I think it might be better to start off by requiring Application CRs and the controller. If we get feedback that this is causing problems then we can relax it later.

I think this is better than not requiring Application CRs and then trying to impose it later.

> my thoughts are application CRs and application-controller will vary and expose differents ways to group applications, different ways to surface INFO items, different relationships among applications.

I think we should try to standardize this at least in the context of Kubeflow. Ensuring that certain information (e.g. version etc...) is present and conforms to some standard is going to be important to enable people to build tooling that depends on this info (e.g. tools for monitoring/upgrading Kubeflow). 

So should we push in the direction of standardizing and limiting the ability to customize things.

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -171,7 +171,7 @@ func GetKfApp(kfdef *kfdefsv2.KfDef) kftypes.KfApp {
 		},
 	}
 
-	// build restConfig and apiConfig using $HOME/.kube/config if the file exist
+	// build restConfig and apiConfig using $HOME/.kube/config if the file exists
```
hougangliu: Please also update `Delete` method about deleteGlobalResources (I think we don't need this method any more since delete Application CR will delete GlobalResources, too)

jlewi:
> The default is installing the application-crd, application-controller and application CRs (via overlays). I agree with your comment for the paired manifest issue.

What does this mean in terms of adding the flag `--enable-applications`? Does it mean we don't need that flag?

bootstrap/config/overlays/application/kfctl_default.yaml
```diff
@@ -0,0 +1,27 @@
...
+    jupyter-web-app:
+    - name: overlay
+      value: application
+    notebook-controller:
+    - name: overlay
+      value: application
+    pytorch-operator:
+    - name: overlay
+      value: application
+  enableApplications: true
```
hougangliu: when enableApplications flag is true,  application overlay is redundant?

kkasravi:
@jlewi please /lgtm again - build passes

PR #3430, opened 2019-06-10
Scaffolding for click to deploy to handle each deployment in its own statefulset

jlewi:
* Create a skeleton impelmentation of a router that will handle requests
  by spinning up a statefulset for each deployment (if one doesn't exist)
 and send the request to it.

* This PR creates a skeleton router.

  * The router interface now handles a single create deployment request.

  * The router provides some sample code for creating a K8s service.

  * In the next PR we will actually configure the service based on the request
    and create a suitable statefulset to act as the backend.

  * A full implementation of the backend is blocked by #3045 which is making
    the changes to use kfctl as a library in click to deploy.

Related to: #2152

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3430)


bootstrap/cmd/bootstrap/app/router.go
```diff
@@ -0,0 +1,128 @@
...
+	http.Handle("/kfctl/apps/v1alpha2/create", optionsHandler(createHandler))
+}
+
+
+// Apply runs apply on a ksonnet application.
+func (r *kfctlRouter) CreateDeployment(ctx context.Context, req CreateRequest) error {
+	// TODO(jlewi):
+	// 1. Do IAM check
+	// 2. Check if service exists by sending request
+	// 3. If service doesn't exist create a service and statefulset
```
kunmingg: Shall we connect to stateful set directly?
And by using stateful set we assume pod will auto restart?

jlewi: What does connecting to stateful set directly mean?
The statefulset ensures that a new pod will be created in the event the existing pod is terminated e.g. because its evicted because of resource issues.

We connect via the associated service because that gives us a stable way to address the backend across pod restarts.

kunmingg: I'm wondering how much extra cost to send request with services.
Any alternative is let router create/delete pods directly, and use pod dns address to communicate.

jlewi: Using a service is pretty standard. Is there a reason you're worried about latency?
Latency should be pretty small O(1)s and given creating a deployment is on the order of 20 minutes its insignificant.

Currently I think resilience is a much bigger problem.

Issue #3433, opened 2019-06-10
0.5.1 not deployable due to recent master change


gokuatkai:
Hello, 
v0.5.1 uses master in aws utils.sh on the istio dependencies.
Thus it was removed throu: 
https://github.com/kubeflow/kubeflow/commit/f10f4fb954ad8617d0d6d35ba01356e4b19f3815
There is also a patch here:
https://github.com/kubeflow/kubeflow/commit/d38b6e7c02cf487d86f001269a971bd13b48d08c
But this last patch, I think should either:
- use the `KUBEFLOW_TAG` instead of `KUBEFLOW_VERSION`
- or add the letter `v` in front of `${KUBEFLOW_VERSION}`
Otherwise it wont find anything as the url is:
https://raw.githubusercontent.com/kubeflow/kubeflow/v0.5.1/dependencies/istio/install/crds.yaml
(containing the `v`)

Is there any chance to have v0.5.1 tag patched correctly?

jlewi:
The code on the release branch should not depend on master; so that seems like a bug that we should fix.

/cc @Jeffwan

Jeffwan:
@gokuatkai  I will ask my teammate to cherry-pick this change to 0.5-branch and then I will file a PR to update documentation to use stable branch.

gokuatkai:
I tried to cherry pick this commit (https://github.com/kubeflow/kubeflow/commit/d38b6e7c02cf487d86f001269a971bd13b48d08c) but it fails, missing the letter `v`

Jeffwan:
> I tried to cherry pick this commit ([d38b6e7](https://github.com/kubeflow/kubeflow/commit/d38b6e7c02cf487d86f001269a971bd13b48d08c)) but it fails, missing the letter `v`

Double check this issue. We need to use `KUBEFLOW_TAG` rather than `KUBEFLOW_VERSION` in the istio url.

onur-sam-gtn-ai:
`scripts/aws/util.sh` is trying to reach the files in `dependencies/` but they no longer exist.
This applies both to the v0.5.1 (tries to download from master) and the master (tries to download from ${KUBEFLOW_TAG} which is again master).
There seems to be a problem with kiali (the required secret doesn't exist?) in v0.5.1 as well so even after hardcoding into util.sh, v0.5.1 is still not deployable.
I've only been able to deploy using commit 877a1a71bf317951bd0527db15f68d73c2e55c73.
[...]

Issue #3332, opened 2019-05-22
Kubeflow on EKS not starting jupyter notebooks


amarrella:
Hello, 
I'm playing around with a Kubeflow installation on EKS following the instructions in https://www.kubeflow.org/docs/aws/deploy/install-kubeflow/ 

Everything seems to work fine, but when I try to start a jupyter notebook server and i click "SPAWN" the request returns a 500.

Checking the logs in the jupyter-web-app pod I see the following:
```
[2019-05-22 15:31:20,934] ERROR in app: Exception on /api/namespaces/kubeflow/notebooks [POST]
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 2292, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1815, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1718, in handle_user_exception
    reraise(exc_type, exc_value, tb)
  File "/usr/local/lib/python3.6/dist-packages/flask/_compat.py", line 35, in reraise
    raise value
  File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.6/dist-packages/flask/app.py", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/app/kubeflow_jupyter/default/app.py", line 40, in post_notebook
    poddefaultLabels = api.get_poddefaults_labels(namespace)
  File "/app/kubeflow_jupyter/common/api.py", line 123, in get_poddefaults_labels
    ns, "poddefaults")['items']
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/apis/custom_objects_api.py", line 1432, in list_namespaced_custom_object
    (data) = self.list_namespaced_custom_object_with_http_info(group, version, namespace, plural, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/apis/custom_objects_api.py", line 1538, in list_namespaced_custom_object_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/api_client.py", line 321, in call_api
[...]

PR #3443, opened 2019-06-11
fix aws provider istio version

M00nF1sh:
Fix ISTIO version.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3443)


Jeffwan:
This should fix 
https://github.com/kubeflow/kubeflow/issues/3433
https://github.com/kubeflow/kubeflow/issues/3332

PR #3448, opened 2019-06-11
Rewrite the jupyter-web-app with Angular

### LINKED ISSUE #3256: Make Jupyter web app ship with its dependencies
Currently the Jupyter-web-app image pulls from js and css CDNs that reside on the public internet.

If a user is accessing the jupyter web app UI from a browser in an environment with no access to the internet, the UI is unusable.

### LINKED ISSUE #3439: Rewrite the jupyter-web-app using a web framework
The webapp is currently written in plain HTML/JS. The reason behind that was to leverage existing code from the previous JupyterHub implementation. But using plain HTML/JS makes it more difficult to add more advanced features to the UI, such as a multi select for #2992. Also, it is making it more challenging to modularize the code in order to reduce code duplication.

Rewritting the webapp in a framework could tackle most of these problems and make to webapp easier to maintain and further develop.

/cc @jlewi @avdaredevil @prodonjs

kimwnasptd:
/closes #3439
/closes #3256

In this PR we replace the current HTML/JS implementation with Angular. We also made some changes in the backend to accomodate the changes in the new frontend. Functionality wise the new UI is up to date with the existing one.

The new image produced from this code will work out the box with the existing deployment/service, but needs the updated `config.yaml`. The new config has some new fields for setting a default value for shm, as well as selecting a specific StorageClass for the Workspace and Data Volumes. The new `config.yaml` will be the only change we must make to the manifests.

#### How to try out:
1. Set the jupyter-web-app image to `<public-new-image>`
2. Edit the `jupyter-web-app-config`:
	* In `workspaceVolume.value.name.value` set it to `'workspace-{notebook-name}'` (the quotes are important). `{notebook-name}` will be replaced in the UI with the Notebook's name as the user types it.
    * Add a `class` key in `workspaceVolume.value` with `value: '{none}'`
    * If you have default values for `dataVolumes`, then add the quotes to their names too.
    * Make the value of `extraResources` to be `{}` and not `{{}}`.
	* In the end of the file, after `extraResources`, add an entry `shm` with a boolean as default value.
```yaml
...
    accessModes:
      ## The Access Mode of the Workspace Volume
      ## Supported values: 'ReadWriteOnce', 'ReadWriteMany', 'ReadOnlyMany'
      value: ReadWriteOnce
    class:
      # The StrageClass the PVC will use if type New. Special values are:
      # {none}: default StorageClass
      # {empty}: empty string ""
      value: '{none}'
...
extraResources:
  # Extra Resource Limits for user's Notebook
  value: "{}"
shm:
  value: true
  # readOnly: true
```

/assign prodonjs
/assign avdaredevil
/assign jlewi
/assign kimwnasptd

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3448)


prodonjs:
Wow this is excellent work Kimonas! I built and deployed this image in a cluster and it seems to be working as intended. The UI seems to have parity and now that it's in a modern framework, it should be much easier to iterate and add functionality going forward.

I'll give the LGTM on the client-side code itself, but before moving forward I'd like to make sure someone with more experience in spawning new Notebook servers with different options can take a look and do some UAT.

/lgtm

jlewi:
@kimwnasptd Are you deleting the old web app in this PR as well?

Would it make sense to leave the old code there and install both applications just until we've had a chance to make sure the new one is WAI. If its non-trivial then I think we should just move forward with this PR as is and if necessary we can bring back the old version later. I'll approve this PR and you can just remove the hold.

Do we have an issue tracking adding/updating the kustomize manifest for the jupyter web app?

/approve
/hold

kimwnasptd:
Thank you @prodonjs!

@jlewi yes, the PR deletes the files of the old webapp. It would make sence to keep the old code and install both of them, but it would be indeed non trivial. Would also vote to move forward with this PR, if its ok with you, and make any necessary future change to ensure that this one is WAI.

For the manifests, I created an [issue](https://github.com/kubeflow/manifests/issues/143) to keep track of the files/changes that need to be made. I'll leave the `hold` on this PR until I add the necessary changes to the manifests and make sure that the new UI can be deployed with kustomize.

jlewi:
@kimwnasptd How's this coming?

kimwnasptd:
@jlewi the PR for the config is [up](https://github.com/kubeflow/manifests/pull/163). Once its merged i'll remove the hold and update the manifests with the new tag.

kimwnasptd:
The needed config is merged with master in manifests. Proceeding with merging this PR
/hold cancel

kimwnasptd:
@jlewi @prodonjs I updated the commits to resolve a conflict from #3507, could you lgtm this again in order for it to merge?

cc @abhi-g

Issue #3453, opened 2019-06-12
kfctl: Dockerfile duplicates Makefile logic


yanniszark:
The Dockerfile we use for bootstrap/kfctl uses logic that already exists inside the Makefile, eg:
https://github.com/kubeflow/kubeflow/blob/aeac4312c103caf901bd48acbae3c5c6f5d3fb4b/bootstrap/Dockerfile#L64-L72

This results in code duplication and mismatch between what the normal and containerized build does.

Steps to fix: Dockerfile should call make commands and not replicate their functionality.

yanniszark:
Fixed by #3455

Issue #3452, opened 2019-06-12
kfctl: Makefile vet, fmt targets don't cover the whole codebase


yanniszark:
The `vet` and `fmt` targets in the `kfctl` Makefile don't cover the whole codebase.
https://github.com/kubeflow/kubeflow/blob/aeac4312c103caf901bd48acbae3c5c6f5d3fb4b/bootstrap/Makefile#L60-L65

This causes issues like #3365  to not be caught in the unit tests.

Steps to fix:
* Fix fmt, vet Makefile targets

yanniszark:
Fixed by #3455

PR #3455, opened 2019-06-12
kfctl: fix Makefile fmt,vet targets and refactor Dockerfile to use make

yanniszark:
* #3452 Fix the Makefile's fmt, vet targets. These targets didn't include all necessary files. 
* #3453 Refactor the Dockerfile and eliminate duplicate logic with make commands.

/assign @yanniszark 
/assign @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3455)


yanniszark:
> the problem with running make from the dockerfile is the steps docker logs as layers are now hidden and it can't do cache management when pushing a new docker image

Indeed.
However, the big layers are 2 in this case:
1. The layer containing the dependencies (created by go mod download)
2. The layer containing the binary (created by go build - inside make build-{kfctl,bootstrap})

These two layers are separate in the Dockerfile so we gain the major benefits of the Dockerfile caching.
Splitting the make build-{kfctl, bootstrap} targets in separate commands wouldn't save us much time/network bandwidth and would invite problems because of the code duplication which might go out-of-sync.

> what does DOCKER_BUILDKIT do?
The reason we use it here is because we target a specific stage of the Dockerfile.
If we don't use the DOCKER_BUILDKIT, docker builds all the stages up to the target one, even unnecessary ones.
DOCKER_BUILDKIT analyzes the stage dependency tree and doesn't build unnecessary stages.

PR #3463, opened 2019-06-13
Add katib v1alpha2

richardsliu:
Fixes https://github.com/kubeflow/katib/issues/513

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3463)


hougangliu:
@richardsliu In fact, I cannot understand why we need https://github.com/kubeflow/manifests/pull/128 to reorganize katib package, all these components of katib should be installed together to make Katib workable.
I think a package should be made of multiple components only when parts of components may be installed for a package (but katib cannot work well if there are some components non-installed) @kkasravi  please correct me if I misunderstand.

kkasravi:
Using some of kustomize’s features like nameprefix, commonLabels works best across a logical set of resources (service, deployment, configmap). Also, kustomize won’t order resources like CRDs, so these should be split out

kkasravi:
This is similar to how pipeline is structured, different groupings of microservices that work together

richardsliu:
@kkasravi Does this mean the CRDs should be in their own component? Currently in https://github.com/kubeflow/manifests/pull/131, the Experiment and Trial CRDs are placed in the same component as katib-controller. Should they be separated out?

bootstrap/config/base/kfctl_default.yaml
```diff
@@ -25,7 +25,12 @@ spec:
   - centraldashboard
   - admission-webhook  
   - jupyter-web-app
-  - katib
+  - katib-db
+  - katib-manager
+  - katib-controller
+  - katib-ui
+  - metrics-collector
+  - suggestion
```
kunmingg: What is "suggestion"?

richardsliu: Katib algorithm suggestions (grid, random, bayesian, etc)

bootstrap/config/kfctl_default.yaml
```diff
@@ -28,7 +28,12 @@ components:
   - centraldashboard
   - admission-webhook  
   - jupyter-web-app
-  - katib
+  - katib-db
```
kunmingg: What's the benefit to break katib into:
  - katib-db
  - katib-manager
  - katib-controller
  - katib-ui
?

richardsliu: @kkasravi mentioned this in https://github.com/kubeflow/kubeflow/pull/3463#issuecomment-501846718.

This allows groupings of related microservices and the use of kustomize features like prefixes and labels.

Issue #2997, opened 2019-04-11
[Discussion] Master branch tagging


jlewi:
We commonly use [git describe](https://git-scm.com/docs/git-describe) to describe version information; e.g. for image tags.

```
git describe --tags --long
```

This will output the most recent tag, number of commits since the tag and then the actual commit.

On release branches this provides nice output e.g. on v0.5-branch

```
git describe --tags --long
v0.5.0-0-ge7d0b5a7
```

Since the branch hasn't diverged yet from the v0.5.0 tag.

If we check out a more commit before that e.g. 

```
git checkout 20205e4
v0.5.0-rc.3-12-g20205e4f
```

So its obvious we are 12 commits past the v0.5.0-rc.3 tag.

On master though we don't consistently tag it.

```
git describe --tags --long
v0-29-g9625cde8
```

Not sure where the v0 tag came from.

Should we add a v0.Y-dev tag at the fork point with the previous release branch? e.g. 

```
git merge-base upstream/master upstream/v0.5-branch
4485a65079a1b445195cb7cc34035655919aaaea
```

avdaredevil:
A thing to add for a few folk which I ran into as well, is to make sure you have the latest copy of tags by running:

```sh
git fetch prod
```

_**Note:** In my case `prod` points to this Kubeflow repo, while origin points to my personal fork_

jlewi:
I think I could delete the v0 tag by doing something like

```
git push upstream :refs/tags/v0
```

jlewi:
Created a release 

```
v0.6.start
```

At the start of the release branch.

PR #3466, opened 2019-06-14
Updated master with the latest image for central-dashboard for v0.6

avdaredevil:
^^

---
/cc @jlewi @prodonjs 
/assign @avdaredevil 
/area front-end
/area centraldashboard
/priority p0

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3466)

### DIFF
--- a/kubeflow/common/prototypes/centraldashboard.jsonnet
+++ b/kubeflow/common/prototypes/centraldashboard.jsonnet
@@ -3,7 +3,7 @@
 // @description centraldashboard Component
 // @shortDescription centraldashboard
 // @param name string Name
-// @optionalParam image string gcr.io/kubeflow-images-public/centraldashboard:v0.5.0 Image for the central dashboard
+// @optionalParam image string gcr.io/kubeflow-images-public/centraldashboard:v20190614-v0-208-g17ef0ecd Image for the central dashboard
 // @optionalParam injectIstio string false Whether to inject istio sidecar; should be true or false.
 // @optionalParam clusterDomain string cluster.local DNS config to cluster domain.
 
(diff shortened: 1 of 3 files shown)


kubeflow/common/prototypes/centraldashboard.jsonnet
```diff
@@ -3,7 +3,7 @@
 // @description centraldashboard Component
 // @shortDescription centraldashboard
 // @param name string Name
-// @optionalParam image string gcr.io/kubeflow-images-public/centraldashboard:v0.5.0 Image for the central dashboard
+// @optionalParam image string gcr.io/kubeflow-images-public/centraldashboard:v20190613-v0.4.0-rc.1-527-g7ca3cfb3 Image for the central dashboard
```
prodonjs: Can we retag the image that's being pointed to here so it doesn't appear that it's a stepping backwards?

jlewi: See https://github.com/kubeflow/kubeflow/issues/2997

There are two issues here

1. The tags of the master branch
1. The expression used in the Makefile to get the descriptor; I think we want to switch to using

    ```
   git describe --tags --long
    ```

avdaredevil: From the command above I am getting: `v0.4.0-rc.1-528-g65c0b34f`. Is this correct?

avdaredevil:
Presubmit failure fix seems to be available @ #3474.

jlewi:
@avdaredevil Can you rebase off master to pull in the fix to the failing tests (#3475). That should fix the presubmit.

prodonjs:
I would imagine this should be updated in the manifests repo as well, right?

https://github.com/kubeflow/manifests/blob/master/common/centraldashboard/base/kustomization.yaml

Issue #3419, opened 2019-06-06
provide a fixed set of yaml files under bootstrap/config that reflect kfctl cli options  


kkasravi:
overlays under bootstrap/config (basic_auth,  gcp,  istio,  ksonnet, kustomize) are used to create a kfctl_default.yaml in Coordinator.Init by applying StrategicMergePatch to the set of overlays that match the kfctl cli options (--platform, --basic_auth, --use_istio).

Rather than do this within Coordinator, it should be done within separate tooling that generates different kfctl_default.yaml files across the permutation of cli options. Coordinator would then just derive the kfctl_default* filename and use that.

The location and name of the config file will incorporate the cli-options:

config/{platform}/kf{-basic_auth}{-istio}.yaml

permutations:
- platform=gcp
  - config/gcp/kf.yaml
  - config/gcp/kf-istio.yaml
  - config/gcp/kf-basic_auth.yaml
  - config/gcp/kf-basic_auth-istio.yaml
- platform=none
  - config/kf.yaml
  - config/kf-istio.yaml

invalid option combinations result in the absence of the file and a usage message from kfctl

jlewi:
Should we aim to get rid of the different flags and just take the name of the YAML config file e.g.

```
kfctl init --config=kfctl_gcp_istio.yaml
```

This would allow us to eventually support URIs e.g.

```
kfctl init --config=https://github.com/kubeflow/kubeflow/myconfig.yaml
```
So people could easily create and share new configs.

kkasravi:
I think this is a good idea, it would solve a few problems.

kkasravi:
the existing bootstrap/config/{base,overlays} should be moved into manifests so that tooling in manifests can generate new config files if utilized.

Issue #2831, opened 2019-03-27
kfctl should support pulling config from a local directory and more general git syntax


jlewi:
kfctl downloads the configuration here:
https://github.com/kubeflow/kubeflow/blob/cee0f6e3f84e01009abef3b81db65c6e3f36d91a/bootstrap/pkg/kfapp/coordinator/coordinator.go#L89

It currently assumes we are pulling the config from a git repo. 

We should support copying the repo from a directory. This way its easy to test local changes.

My suggestion would be to allow the version flag to be a uri. We could then do something like

```
--version=file://
```

The code can then determine how to download the files correctly based on the URI

/cc @kkasravi @gabrielwen

jlewi:
Related the git syntax makes assumptions about the repository

```
  -v, --version string          desired version of Kubeflow or master if not specified. Version can be master (eg --version master) or a git tag (eg --version=v0.5.0), or a PR (eg --version pull/<id>). (default "master")
```

So when you do "pull/<id>" its assuming repository is "kubeflow".

Is there a convenient way to support URIs?

For reference here's the syntax github recommends for checking out PRs

```
git checkout -b kubeflow-bot-update_jupyter_625e24e master
git pull https://github.com/kubeflow-bot/kubeflow.git update_jupyter_625e24e
```

We also have [checkout_repos.sh](https://github.com/kubeflow/testing/blob/master/images/checkout_repos.sh)

Which uses

```
{REPO_ORG}/{REPO_NAME}@{SHA}:{PULL_NUMBER}
```

So it makes an assumption that it is github.

jlewi:
@kkasravi Do we support local repos yet? Do our presubmit tests just rely on pulling from git via the PR?

Should we punt to 0.7?

jlewi:
#3468 add --config file added some support for pulling from a local file. Repo paths are using hashi-corps go-getter which also supports local file.
https://github.com/hashicorp/go-getter

We should probably be consistent.

GoGetter's getAny method supports a ref keyword.
https://github.com/hashicorp/go-getter#git-git

We can use this with GitHub refs
https://help.github.com/en/articles/checking-out-pull-requests-locally

jlewi:
You can test it out with go-getter's command line string.
https://github.com/hashicorp/go-getter/blob/master/README.md#installation-and-usage

yanniszark:
@jlewi I like this a lot!
Anything that pulls things from URIs should use the same tool.
[...]

PR #3468, opened 2019-06-14
kfctl: add --config flag for specifying a static config file

yanniszark:
As per the discussion in #3419, it seems that we want to move to an approach where:
* Each platform specifies its own set of static yamls files for the KfDef config.
* A user can specify a KfDef config to use, which should give us reproducibility, easy sharing of configs and is in-line with the GitOps philosophy.

This PR adds the `--config` flag to `kfctl`.
When set, `kfctl` will read the KfDef from the specified file instead of generating it using kustomize.
Platforms can also override this option and specify their own static file.

Partially related to: #2831 support pulling from a local repository
   * This will pull the config from a local repository but we still have to support adding and pulling
      all of the repos/manifests on which we depend from a local repo.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3468)

### DIFF
--- a/bootstrap/cmd/kfctl/cmd/init.go
+++ b/bootstrap/cmd/kfctl/cmd/init.go
@@ -62,6 +62,7 @@ or a <name>. If just <name>, a directory <name> will be created in the current d
 
 		useIstio := initCfg.GetBool(string(kftypes.USE_ISTIO))
 		disableUsageReport := initCfg.GetBool(string(kftypes.DISABLE_USAGE_REPORT))
+		config := initCfg.GetString(string(kftypes.CONFIG))
 
 		options := map[string]interface{}{
 			string(kftypes.PLATFORM):              platform,
@@ -75,6 +76,7 @@ or a <name>. If just <name>, a directory <name> will be created in the current d
 			string(kftypes.USE_ISTIO):             useIstio,
 			string(kftypes.DISABLE_USAGE_REPORT):  disableUsageReport,
 			string(kftypes.PACKAGE_MANAGER):       packageManager,
+			string(kftypes.CONFIG):                config,
 		}
 		kfApp, kfAppErr := coordinator.NewKfApp(options)
 		if kfAppErr != nil || kfApp == nil {
@@ -188,6 +190,18 @@ func init() {
 		initCmd.Flags().Lookup(string(kftypes.DISABLE_USAGE_REPORT)))
 	if bindErr != nil {
 		log.Errorf("couldn't set flag --%v: %v", string(kftypes.DISABLE_USAGE_REPORT), bindErr)
+		return
+	}
+
+	// Config file option
+	initCmd.Flags().String(string(kftypes.CONFIG), "",
+		`Static config file to use. Can be either a local path or a URL.
+For example:
+--config=https://github.com/kubeflow/kubeflow/tree/master/bootstrap/config/platform_existing.yaml
+--config=platform_gcp.yaml`)
+	bindErr = initCfg.BindPFlag(string(kftypes.CONFIG), initCmd.Flags().Lookup(string(kftypes.CONFIG)))
+	if bindErr != nil {
+		log.Errorf("couldn't set flag --%v: %v", string(kftypes.CONFIG), bindErr)
 		return
 	}
 }
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -30,6 +30,7 @@ import (
 	"github.com/kubeflow/kubeflow/bootstrap/v2/pkg/kfapp/kustomize"
 	"github.com/mitchellh/go-homedir"
 	log "github.com/sirupsen/logrus"
+	goopenuri "github.com/utahta/go-openuri"
 	"io/ioutil"
 	valid "k8s.io/apimachinery/v2/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/v2/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -63,9 +63,28 @@ func GetKfApp(kfdef *kfdefsv2.KfDef) kftypes.KfApp {
 	return _coordinator
 }
 
-func getConfigFromCache(pathDir string, kfDef *kfdefsv2.KfDef) ([]byte, error) {
+func getStaticConfigForPlatform(kfDef *kfdefsv2.KfDef) string {
```
jlewi: What's the purpose of this function? Shouldn't any config  go into the static file?

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -63,9 +63,28 @@ func GetKfApp(kfdef *kfdefsv2.KfDef) kftypes.KfApp {
...
+		if err != nil {
+			return nil, &kfapis.KfError{
+				Code:    int(kfapis.INTERNAL_ERROR),
+				Message: fmt.Sprintf("error reading specified config file %s: %v", staticConfigFile, err),
+			}
+		}
+		return configFileBytes, nil
+	}
+
+	configPath := filepath.Join(kfDef.Spec.Repo, kftypes.DefaultConfigDir)
```
jlewi: It looks like we are looking at a KfDef spec in order to locate a YAML file from which we get the KfDef spec.

Can we simplify this and not use a KfDef spec to provide whatever info (e.g. repo information to locate the yaml file).

For example suppose the syntax was 

```
kfctl init --config=file://myrepo/path/kfdef.yaml
```

Then the argument fully specifies the location so there's no reason to reference some other kfdef spec to figure out where to locate it from.

jlewi:
I was expecting to change the code here
https://github.com/kubeflow/kubeflow/blob/7ca3cfb39368eb49042992f4e16215e41e1953c6/bootstrap/pkg/kfapp/coordinator/coordinator.go#L258

The current code is a bit roundabout. 

```
	kfDef := &kfdefsv2.KfDef{
...
	}
	configFileBuffer, configFileErr := getConfigFromCache(cacheDir, kfDef)
```

Its roundabout because we are creating a partially constructed `KfDef` and then using that to then read a `KfDef` spec from a yaml file. This leads to confusion about what happens if the original kfDef spec and the one in the yaml file have conflicts. 

Why not just do something like the following

````
pathToConfigFile := cacheDir + "/" + args.config
kfDef, err := loadKfDef(pathToConfigFile)
````

File: bootstrap/pkg/kfapp/coordinator/coordinator.go (hunk 1 above)
Comment (yanniszark) [reply bootstrap/pkg/kfapp/coordinator/coordinator.go]: The idea is:
* If the user has specified a config file, that file is used.
* If the user hasn't specified a config file, the config value can be seeded with a static config file for the specific platform.

File: bootstrap/pkg/kfapp/coordinator/coordinator.go (hunk 2 above)
Comment (yanniszark) [reply bootstrap/pkg/kfapp/coordinator/coordinator.go]: I agree that it's a convoluted way to do it and that we should refactor this part of kfctl to be more simple/clear in what it's doing.
However, I  believe it is a bit out of scope for this PR and would be a pretty big PR on its own.

**Comment by yanniszark** [comment]
@jlewi indeed the way we construct KfDef is convoluted and we need to make it more clear and simple.
Right now, after the KfDef is constructed with kustomize, the following values are overridden:
```go
	kfDef.Name = appName
	kfDef.Spec.AppDir = appDir
	kfDef.Spec.Platform = platform
	kfDef.Namespace = namespace
	kfDef.Spec.Version = version
	kfDef.Spec.Repo = path.Join(cacheDir, kftypes.KubeflowRepo)
	kfDef.Spec.Project = options[string(kftypes.PROJECT)].(string)
	kfDef.Spec.SkipInitProject = options[string(kftypes.SKIP_INIT_GCP_PROJECT)].(bool)
	kfDef.Spec.UseBasicAuth = useBasicAuth
	kfDef.Spec.UseIstio = useIstio
	kfDef.Spec.PackageManager = packageManager
```

In the case of `--config` should these be also overriden?
Should they be left as they are in the config file?
I believe this is something that merits further discussion.

For this PR, I made --config act as a replacement to the kustomize process of creating the config file.
I think that is a good place to start.

PS. I also added a small library (go-openuri) so that --config can easily accept either local files or http(s) schemes.

bootstrap/pkg/kfapp/coordinator/coordinator.go
jlewi: I'm not sure I follow. Isn't the idea to get rid of the paltform flag and replace it with a static config file?

i.e. instead of specifying --platform=gcp, I'd specify 
```
--config=kfctl_gcp_istio.yaml
```

To put it another way. What if we made it an error to specify both "--platform" and "--config"? Would we still need this function?

jlewi:
You raise a good point about keeping the delta in this PR small and therefore continuing to override certain values in KfDefSpec after loading KfDefSpec.

What does that have to though with constructing a partially defined KfDefSpec in order to locate the KfDefSpec.yaml we want to load?

Can't the steps be

1. Look at --config
1. If --config isn't a full URI; append defaults as needed (e.g. assume its in kubeflow/kubeflow on a particular branch)
1. Load KfDefSpec.yaml - just call your go-get-URI function.
1. At this point processing KfDef should be the same regardless of whether it was produced via static config file or via kustomize.
1. Override KFDefSpec with any values as needed

Am I missing something?

yanniszark:
Thanks for the review @jlewi! I agree with many concerns that you raise. Let me elaborate:

> What does that have to though with constructing a partially defined KfDefSpec in order to locate the KfDefSpec.yaml we want to load?

The `getConfigFromCache` function, which exists in kfctl before this PR, uses a partially constructed KfDef to find the correct paths in order to apply kustomization and generate the final KfDef.
The `getStaticConfigForPlatform` function that this PR introduces just follows the example of that function.
However, I agree it would be more correct to just pass whatever arguments are necessary.
I will change the function arguments.

The steps right now are:
1. Load --config value. If a value isn't given by the user, check if the platform wants to specify a static config file. In other words, `--platform=existing` becomes synonymous with `--config=.cache/kubeflow/kubeflow/master/bootstrap/config/platform_existing.yaml`.
2. If --config is set, load it. Otherwise generate it with kustomization and load it.
3. Override certain values.

jlewi:
Thanks @yanniszark Let me know when this is ready for another look.

jlewi:
Thinking longer term does this confusion go away if  [repo](https://github.com/kubeflow/kubeflow/blob/d491a71e14284cb052f84004fc36975d59beced2/bootstrap/config/base/kfctl_default.yaml#L6) property that's in KFDef points to the proper git location for that file?

For example, we will have a static config `kfctl_gcp_istio.yaml` on the v0.6-branch. The value of repo in that config file should point to the repos on that branch e.g. https://github.com/kubeflow/kubeflow/blob/v0.6-branch/

So kfctl takes a single parameter pointing at the static file containing KfDef and then reads the repos from that file.

I think we should remove from kfctl the overriding of the repos field in KfDef and probably other fields. I don't think we need it anymore and its causing confusion.

In other words, I think the desired semantics should be similar to how kubectl works;

e.g with kubectl you can create a YAML file

```
kubectl create -f myresource.yaml
```

If you want to customize the resource you do it by editing the yaml file before you run create. You don't generally supply additional command line arguments to the create function to override certain parts of  the spec e.g. you don't do

```
kubectl create -f myresource.yaml --name=myresource
```

yanniszark:
@jlewi I really like this approach and the similarities to kubectl.
This also makes init similar to generate and apply, in that they also load a KfDef file, app.yaml.

How do you suggest we go about this for 0.6?
Keep both or aim for having --config and deprecating --platform and others?

jlewi:
@yanniszark Lets try to get this PR in ASAP because once we have config taking in a complete YAML file we can avoid using overlays which I think will improve velocity a lot.

For this PR I think we just want to make the minimal changes to

1. Take in --config
1. If --config is not a complete URI make it a complete URI based on some sensible default; i.e. kubeflow/kubeflow@master
1. Load that via gooopenURI

yanniszark:
@jlewi got it!
I'll have the first iteration asap.

jlewi:
Thanks @yanniszark

bootstrap/cmd/kfctl/cmd/init.go
```diff
@@ -61,6 +61,7 @@ or a <name>. If just <name>, a directory <name> will be created in the current d
 		}
 
 		useIstio := initCfg.GetBool(string(kftypes.USE_ISTIO))
```
kkasravi: wouldn't we want to remove all other cli options in Init like --project, --platform, --namespace, --repo, ...

bootstrap/pkg/kfapp/coordinator/coordinator.go
kkasravi: i think command completion should provide a list of static config files available to help the user
eg
--config kfctl_<tab><tab> kfctl_default.yaml kfctl_default_istio.yaml kfctl_gcp.yaml kfctl_gcp_istio.yaml, ...

bootstrap/cmd/kfctl/cmd/init.go
yanniszark: I agree that the plan is to remove all of those flags.
However, as this is PR should not break anything, it will just create a different code path.
Once all platforms migrate, we should be able to delete the old code path.

yanniszark:
@jlewi the first iteration is ready.

* The `--config` flag accepts either a path or a URL. If the protocol is missing in front, it is assumed to be a file.
* We still need to override some fields to maintain the same functionality. We need to override these because their values are mainly determined at runtime. Namely:
    * `repo`: contains the information of where the kubeflow cached repo resides. 
    * `appDir`: container the information of where the app.yaml resides.

After deciding on a better practice for `repo` in #3471, we should be able to make this procedure cleaner.

jlewi:
/lgtm
/approve

/hold 

I only had 2 nitpicky comments otherwise this looks great.

PR #3470, opened 2019-06-14
Add admission-webhook components to kustomize default components 

### LINKED ISSUE #3469: Add/fix admission-webhook to  kfctl kustomize default components
admission-webhook is implemented using two components (bootstrap and webhook) in kustomize ([see here](https://github.com/kubeflow/manifests/tree/master/admission-webhook)) . However its components are not added to kustomize kfctl components.  
/cc @kkasravi

zabbasi:
fixes #3469 
/cc @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3470)


bootstrap/config/base/kfctl_default.yaml
```diff
@@ -40,3 +41,6 @@ spec:
       - name: reportUsage
         value: "true"
         initRequired: true
+    admission-webhook-bootstrap:
```
kkasravi: you can do the same thing if you add nameprefix admission-webfix- in kustomization.yaml. Though it adds it to all resource names.

zabbasi:
blocked by [this bug](https://github.com/kubeflow/manifests/issues/133)

yanniszark:
Just a side note:
At first I was going to say this doesn't fix #3469 but now I'm not really sure what the issue is about.

Is #3469 about not adding webhook and bootstrap? 
Or is it about the kfctl component discovery logic being broken? 
If it's the first one, I will open another issue for the second one.

abhi-g:
@yanniszark It seems like kfctl relies on the changes in this file to discover the webhook component. Also see jlewi's comments on the issue.

yanniszark:
@abhi-g it does and that's why the discovery logic of kustomize components is wrong.
I opened an issue for this: https://github.com/kubeflow/kubeflow/issues/3491

abhi-g:
ok. thanks

abhi-g:
@kkasravi @yanniszark The tests have passed now. Please take a look and lgtm if this needs to be merged.

Issue #3471, opened 2019-06-14
kfctl: repo for KfDef is not calculated correctly


yanniszark:
The repo field in KfDef should be the folder under which the kubeflow repository code lives.
In other words, it must hold that:
```console
$ ls <repo>
bootstrap
build
...
```

Currently, we are passing the wrong value for repo.
The value we are passing is `<repo>/kubeflow`:
https://github.com/kubeflow/kubeflow/blob/7ca3cfb39368eb49042992f4e16215e41e1953c6/bootstrap/pkg/kfapp/coordinator/coordinator.go#L296

While the repo value is wrong, the way that GCP handles it makes it work.
More specifically, the path.Dir function is used which cuts out the wrong part.
https://github.com/kubeflow/kubeflow/blob/7ca3cfb39368eb49042992f4e16215e41e1953c6/bootstrap/pkg/kfapp/gcp/gcp.go#L1211

We should pass the correct value and use it as is.

lluunn:
I think it was intended.
Because under github repo kubeflow/kubeflow, the dir kubeflow is the repo for ksonnet registry.
So it was defined that way I believe.

What error does this cause?

yanniszark:
@lluunn I see what you mean.
Does it make sense to keep it that way since:

1. We are moving away from kustomize but other code, unrelated to kustomize, depends on this option.
2. The name of the field (repo) and its description (Name of repository.) suggests it is the Kubeflow repo and not a ksonnet registry. Couple that with the fact that there is an actual RegistryConfig type inside `applications_type.go` and it gets somewhat confusing. If the repo was actually representing a ksonnet registry, shouldn't it use the RegistryConfig type?

lluunn:
Yeah, I think we should probably change as we don't need the kubeflow dir as the ks registry anymore.

cc @kunmingg @kkasravi

jlewi:
Should KfDef.repo be referring to cachedir at all?

Cachedir should be an internal property of the KfApp. The value in the KfDef should refer to the source of truth e.g. a specific commit in the git repo. This way if the user checks in the app into source control and checks out on a different machine we can reconstruct the cachedir from the source of truth. 

See for example #2959 we should redownload files to cachedir if they don't exist.

A second issue I forsee is that with the move to kustomize we now have multiple repos

* GCP/Azure configs are stored in kubeflow/kubeflow
* kustomize manifests are stored in kubeflow/manifests

[...]

PR #3472, opened 2019-06-14
kfctl: pass correct value to repo field of KfDef

yanniszark:
Fixes: #3471

This commit fixes the repo value of KfDef, which is wrong.
Other code that depends on it being wrong is also corrected.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3472)


jlewi:
@yanniszark Is this PR still relevant?

As noted in #3471 I believe this is fixed on master by the fact that we now specify repos in KfDef.Spec and store locations of the cache in KfDef.Status.

yanniszark:
@jlewi this is fixed by the repos field addition.
Thanks for the pr gc

PR #3479, opened 2019-06-17
Click to deploy; Launch a statefulset for each deployment

jlewi:
* Per #2152 we want to start spinning up separate K8s statefulsets to
  handle each Kubeflow deployment.

* Define a new interface KfctlService that defines the methods for creating and managing deployments

* Create a new struct KfctlServer to handle the requests using the Kfctl go libraries

  * A unique KfctlServer is instantiated in its own statefulset for each deployment.

* Crete a new struct KfctlClient to talk to KfctlServer
  * This is used by the router to send requests to the KfctlServer instantiated for a deployment

* kfctl_test provides a unitest for KfctlServer and KfctlClient
  * The test starts up KfctlServer in a go func and then tries sending an http request to it using
    KfctlClient

  * This required modifying KsServer so that could return the port its listening on.

* Modify router.go to handle requests by spinning up a stateful set and forwarding requests to it

  * We give each statefulset and service a unique name
  * The name is a hash of the project, deployment name and zone
  * This ensures requests for the same deployment are always routed to the same instance.

  * router_test provides a unittest for the hashing

* Add flags to bootstrap to control what mode it starts in

  * router - bootstrap starts up as a router; requests are handled by spinning up separate statefulsets and forwarding the requests to them.

    * In router mode we use the downward API to get the pod name and namespce
    * The router then fetches the pod information so we can get the current docker image.
      * This is the docker image used when spawning statefulsets
    * The image isn't available via the Downward API and there is no easy way with kustomize
      to keep the image in sync with a command line argument or environment variable
      * vars doesn't appear to work with the image name
  * kfctl - bootstrap starts up as a server to manage a single deployment

 * kustomize/ contains kustomize manifests for deploying the router

   * This are intended for development not production
   * They can be used with skaffold

* Status:

  * router can successfully spawn a statefulset but on the first attempt forwarding the request to
    the backend it will timeout. It succeeds on retry.

Setup skaffold for deploying the bootstrapper for iterative deployment

* To test the router we want to deploy it on Kubernetes so that it can
  spin up statefulsets and then forward requests to them.

* We setup skaffold to support iterative deployment.

* Create Dockerfile.bootstrap.dev to support iterative builds of just the
  bootstrap binary

  * To allow for fast builds the base container will be an image with the
    go modules already cached.

* Create a .dockerignore file

  * Skaffold uses this to exclude files when deciding when to trigger builds
    as well as files to exclude for upload to GCB.

  * Create test_deploy_kfctl_router.yaml to support deploying the kfctl
    webapp for development with skaffold.

* Related to: #2152

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3479)


bootstrap/cmd/bootstrap/app/kfctlClient.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// We construct a single ratelimiter middleware, to limit the total outgoing
+	// QPS from this client to all methods on the remote instance. We also
+	// construct per-endpoint circuitbreaker middlewares to demonstrate how
+	// that's done, although they could easily be combined into a single breaker
+	// for the entire remote instance, too.
+	limiter := ratelimit.NewErroringLimiter(rate.NewLimiter(rate.Every(time.Second), 100))
```
kunmingg: Should we have the limiter on router side that rate limit total request of each router?

bootstrap/cmd/bootstrap/app/kfctlClient.go
```diff
@@ -0,0 +1,74 @@
...
+
+	// Each individual endpoint is an http/transport.Client (which implements
+	// endpoint.Endpoint) that gets wrapped with various middlewares. If you
+	// made your own client library, you'd do this work there, so your server
+	// could rely on a consistent set of client behavior.
+	var createEndpoint endpoint.Endpoint
+	{
+		createEndpoint = httptransport.NewClient(
```
kunmingg: We might need to check service availability before make post:
statefulSet might need minutes to have underlying pod ready.

bootstrap/cmd/bootstrap/app/router.go
```diff
@@ -81,48 +100,223 @@ func (r *kfctlRouter) RegisterEndpoints() {
...
+
+	if name == "" {
+		return "", fmt.Errorf("name is required")
+	}
+
+	if zone == "" {
+		return "", fmt.Errorf("zone is required")
+	}
+	h := sha256.New()
+	h.Write([]byte(fmt.Sprintf("name=%s,project=%s,zone=%s", name, project, zone)))
```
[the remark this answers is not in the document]
jlewi: If only the version changes we'd still want the request to be processed by the same statefulset so that we can figure out what to do about the version mismatch. 

The deployments that kfctl creates (e.g. the GCP resources only depend on project, zone, name) so it won't be unique if the only thing we change is the version.

bootstrap/cmd/bootstrap/app/kfctlClient.go
jlewi: Can you explain. This is rate limiting requests from router instances to the statefulsets. Are you taking about rate limiting the incoming requests to the router? How would we do that without using something like ISTIO/knative?

bootstrap/cmd/bootstrap/app/kfctlClient.go
jlewi: Yup. I plan on fixing this in a follow on PR. As noted in the PR description this fails on the first request but succeeds on retry.

bootstrap/cmd/bootstrap/app/router.go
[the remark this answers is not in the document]
kunmingg: Got it, then zone should not be part of hash input as well since name itself is unique per project.
We might want same stateful set to handle requests with different zones. So we can dedup if they have same name.

kunmingg:
Only 1 comment otherwise
/lgtm

bootstrap/cmd/bootstrap/app/router.go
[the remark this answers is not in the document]
jlewi: Good point.

jlewi:
test failure was some flake

```
til.py                     71 INFO     Error: couldn't delete KfApp:  (kubeflow.error): Code 500 with message: coordinator Delete failed for gcp:  (kubef
low.error): Code 400 with message: Error when cleaning IAM policy: googleapi: Error 409: There were concurrent policy changes. Please retry the whole read
-modify-write with exponential backoff., aborted
util.py                     71 INFO     Usage:
util.py                     71 INFO     kfctl delete [all(=default)|k8s|platform] [flags]
```

jlewi:
Test failure looks unrelated to this pr: #/3523; problem with running git merge-base HEAD and computing the diff.

/test all

jlewi:
Test seem stuck in pending state. I think it might be a prow issue.

/test all

Issue #1668, opened 2018-09-29
[gcp] E2E test to verify IAP and certmanager works


jlewi:
We need an E2E test that verifies IAP and certmanager works; so that we catch bugs like
#1666 

The Click To Deploy E2E test (#1578) might be sufficient.

/area testing
/priority p1
/area 0.4.0

jlewi:
So #1578 #1732 

I think the issue with those tests is they might not be fully covering certmanager because we need to reuse certs to avoid hitting Lets Encrypt Quota Limits.

I don't think there's any obvious way to address this. So I'm going to mark this as fixed for now and we can create issues for specific  needs as they arise.

jlewi:
Looks like this is fixed by #3482

PR #3482, opened 2019-06-18
E2e test to verify iap endpoint

lluunn:
Manually launched the workflow and the test passed

Related to kubeflow/kubeflow#1668

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3482)


lluunn:
/assign @jlewi 
/assign @kunmingg 

ready for review.

prow_config.yaml
```diff
@@ -26,6 +26,25 @@ workflows:
     params:
       installIstio: true
       workflowName: deployapp-istio
+  - app_dir: kubeflow/kubeflow/testing/workflows
+    component: kfctl_go_test
+    name: kfctl-go-iap-endpoint
+    job_types:
+      - postsubmit
```
jlewi: Lets also modify the existing kfctl_go_test to only run on presubmit. otherwise we are duplicating the test on postsubmit and perioidc runs.

jlewi:
This is great thanks for doing this.

prow_config.yaml
lluunn: good point, fixed.

PR #3499, opened 2019-06-19
Create a token refresher and IAM checker.

jlewi:
* Related to #2152 we want to manage each deployment using its own statefulset
  and get shot-lived credentials from the request.

* To suppport this we need to define a token source that provides a thread
  safe way to swap out a static token.

* We also add an IAM check function to test that the access token provides
  sufficient access to the project. This will be used to reject requests
  if the user dosn't present a token indicating they have access to the
  requested project.

* Furthermore, we run the check as part of the credential refresher to ensure
  we don't swap out a credential that doesn't have access to the project
  the refresher is tied to.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3499)


bootstrap/cmd/bootstrap/app/gcpUtils.go
```diff
@@ -308,3 +310,55 @@ func (s *ksServer) ApplyIamPolicy(ctx context.Context, req ApplyIamRequest) erro
...
+	// might be better to use cluster Admin or similar permission.
+	req := &crm.TestIamPermissionsRequest{
+		Permissions: []string{"resourcemanager.projects.setIamPolicy"},
+	}
+
+	exp := backoff.NewExponentialBackOff()
+	exp.InitialInterval = 2 * time.Second
+	exp.MaxInterval = 5 * time.Second
+	exp.MaxElapsedTime = time.Minute
+	exp.Reset()
```
kunmingg: Should we decrease retry numbers considering cloudresourcemanager has API quota limit? (100 per project per minute?)

PR #3503, opened 2019-06-19
[Central Dashboard v2] Multi-User Isolation PR

### LINKED ISSUE #3437: multi-user UI for access management
MVP:
each user can create / delete their reserved profile (managed namespace).

Ideally we should support 
1. namespace owner share / revoke access with other users 
2. provide admin page which can manage all namespaces access permissions.

### LINKED ISSUE #3555: Access Management UI: add userid-header and userid-prefix cli flags
Sub-issue for #3553.
Add cli flags to configure:
* Which header we read for the user info (now it's only `X-Goog-Authenticated-User-Email`)
* What prefix is expected (now it's expected to be in the form of `accounts.google.com:<user>`, so the prefix is `accounts.google.com:`)

/priority p0

cc @jlewi @avdaredevil

avdaredevil:
**resolves #3437
resolves #3555
resolves: #3554**

## Features
- Updated base lib with readBinding -> readBindings, proper return types
- Integrated payload info in env-info
- Cluster Admin detection

---
/hold
/area front-end
/area centraldashboard
/cc @prodonjs 
/assign @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3503)


components/centraldashboard/app/profile_controller.ts
```diff
@@ -0,0 +1,827 @@
+/**
```
prodonjs: Let's name this file profile_controller_client to make it clear it's a client library and not part of our application server. It also might be useful to put it into a separate folder in case the generated Typescript doesn't comply with our tslint rules. When I tried building this, it failed because the linter was picking up things in the generated code that don't comply with the guidelines.

You can add a linterOptions section to the tslint.json file:
```
"linterOptions": {
        "exclude": [
            "app/clients"
        ]
    }
```

components/centraldashboard/app/api.ts
```diff
@@ -27,6 +32,16 @@ export class Api {
     return this.platformInfo;
   }
 
+  private getProfileAndAuth(req: express.Request): [DefaultApi, AuthObject] {
```
prodonjs: Do you think maybe this makes sense to separate into two methods? Instantiating the controller from the host name of the request seems like one logical function and resolving the authentication header seems like another. I could imagine that the latter will eventually need to expand as additional authentication methods are supported. In addition, the return of two unrelated objects in a tuple seems awkward.

components/centraldashboard/app/api.ts
```diff
@@ -27,6 +32,16 @@ export class Api {
...
+    const controller = this._controller || (
+      this._controller = new DefaultApi(`${req.get('host')}/kfam`)
```
prodonjs: Resolving the user-facing hostname and cluster-level path would work if the request were coming from the browser, but within the cluster it seems like we would unnecessarily be routing the request back out to the public internet only to have it arrive back at the external LB.

I think we can determine the internal endpoint from the K8s service name from within the cluster so the traffic can be handled internally.

components/centraldashboard/app/api.ts
```diff
@@ -40,6 +55,20 @@ export class Api {
     return email;
   }
 
+  /**
+   * Retrieves workgroup info from Profile Controller.
+   */
+  private async getWorkgroup(req: express.Request, user: string): Promise<WorkgroupInfo> {
+    const [profileController, auth] = this.getProfileAndAuth(req);
+    const adminResponse = await profileController.v1RoleClusteradminGet(user, auth);
+    const bindings = await profileController.readBindings(user, auth);
```
prodonjs: It seems like the readBindings method expects a string that is only the user's email address here and thus fails to compile.

components/centraldashboard/app/profile_controller.ts
avdaredevil: Done! Doesn't make sense to rename now that it's in a folder called clients?

components/centraldashboard/app/api.ts
avdaredevil: Fixed, and updated, no longer need to do this

components/centraldashboard/app/api.ts
avdaredevil: Resolved, with an in-cluster host.

components/centraldashboard/app/api.ts
avdaredevil: Fixed, thanks!

components/centraldashboard/app/api.ts
```diff
@@ -40,6 +51,21 @@ export class Api {
...
 
+  /**
+   * Retrieves workgroup info from Profile Controller.
+   */
+  private async getWorkgroup(req: express.Request, user: string): Promise<WorkgroupInfo> {
+    const {profileController} = this;
+    const auth = this.getAuthOption(req);
+    const adminResponse = await profileController.v1RoleClusteradminGet(user, auth);
+    const bindings = await profileController.readBindings(user, undefined, undefined, auth);
+    const namespaces = bindings.body.bindings;
```
prodonjs: This seems like it will change the shape of the response given back to the client. It looks like it might be the `referredNamespace` property that needs to be referenced perhaps?

components/centraldashboard/app/server.ts
```diff
@@ -14,10 +15,11 @@ async function main() {
   const app: express.Application = express();
   const k8sService = new KubernetesService(new KubeConfig());
   const metricsService = await getMetricsService(k8sService);
+  const profileController = new DefaultApi('http://kfam.kubeflow.svc.cluster.local/kfam/');
```
prodonjs: Do we have to think about the situation where the user chooses to use a different namespace to deploy the core KF components to since it is possible with a CLI-based deployment? In that case, I think both the dashboard and kfam services will be deployed in the same namespace so in theory it should be possible to determine the namespace and use it.

prodonjs: Got some clarity from chatting with Zhenghui on Metadata concerning the default namespace. It seems many assumptions are made about the default namespace being **kubeflow** so we may be OK with this for now but it might be worth adding a TODO or handling it dynamically if we find it's a trivial change.

avdaredevil: So if my understanding is correct from my convo with @abhi-g we should **always** have a `kubeflow` namespace (_which is almost like `system`_). If that's incorrect, I'll def add it in

components/centraldashboard/app/api.ts
avdaredevil: Yeah I was planning on the front-end being aware of all these details. I'll make front-end changes tomorrow.

components/centraldashboard/app/server.ts
prodonjs: Yeah I think that's the assumption being made across all system components. I do believe it's possible to change the name of the default namespace to something else when deploying using the CLI method, but all bets are off. We're following this approach in the Metadata project so it seems reasonable to do so here as well.

components/centraldashboard/app/api.ts
prodonjs: Whatever you prefer. If it starts to get pretty big, feel free to break the front-end changes into another PR. I think there are still two considerations though that need to be accounted for.

1. Showing namespaces from other profiles that a user might be a *contributor* to.
2. Showing all namespaces for clusters where identity-awareness is not available. This would be all scenarios other than GKE with IAP at this point so we should probably make sure it maintains consistency with today's behavior.

components/centraldashboard/app/server.ts
avdaredevil: Fair enough, will do!

components/centraldashboard/app/api.ts
```diff
@@ -1,23 +1,29 @@
...
 const IAP_HEADER = 'X-Goog-Authenticated-User-Email';
 const IAP_PREFIX = 'accounts.google.com:';
 
+interface WorkgroupInfo {
+  namespaces: WorkgroupBinding[];
+  isClusterAdmin: boolean;
+}
+
+interface AuthObject {
+  [IAP_HEADER]: string;
```
prodonjs: I do think we should given consideration to @yanniszark 's comment regarding the support for non-Google authentication systems in the future. At the very least, please add a TODO to indicate that it will be addressed in the future.

components/centraldashboard/public/components/main-page.js
```diff
@@ -241,11 +241,24 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
      * @param {Event} responseEvent AJAX-response
      */
     _onEnvInfoResponse(responseEvent) {
+        const roleMap = {
```
prodonjs: Perhaps this could be a module level Map with the same entries?

components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -85,6 +114,10 @@ export class NamespaceSelector extends PolymerElement {
                 value: '',
                 notify: true,
             },
+            selectedNamespace: {
+                type: String,
```
prodonjs: The type doesn't seem to match the default value.

components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -113,6 +155,8 @@ export class NamespaceSelector extends PolymerElement {
      * @param {string} selected
      */
     _onSelected(selected) {
```
prodonjs: I think this observer fires before before namespaces is assigned a value as I was seeing console errors when I deployed it. Probably easiest just to add a guard for namespaces being populated.

https://polymer-library.polymer-project.org/3.0/docs/devguide/observers#observers-and-element-initialization

components/centraldashboard/app/api.ts
yanniszark: Thanks @prodonjs!
This was also [discussed in the community meeting](https://docs.google.com/document/d/1Wdxt1xedAj7qF_Rjmxy1R0NRdfv7UWs-r2PItewxHpE/edit) and it was agreed that the header/prefix value should be configurable.

components/centraldashboard/app/api.ts
```diff
@@ -2,18 +2,23 @@ import express from 'express';
...
 import {Interval, MetricsService} from './metrics_service';
-
-interface EnvironmentInfo {
-  namespaces: string[];
-  platform: PlatformInfo;
-  user: string;
-}
+import {DefaultApi, Binding as WorkgroupBinding} from './profile_controller';
 
 const IAP_HEADER = 'X-Goog-Authenticated-User-Email';
```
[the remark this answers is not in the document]
avdaredevil: Hey @yanniszark, sorry I never responded to this comment, I will expose that via env variables, however, I was hoping to solve this in a better way by having the headers rewritten by our auth layer with a re-signed JWT (_or not based on how we implement this_). For now, I'll update this PR with the appropriate changes as a short-term solution.

Thanks!

/cc @jlewi

components/centraldashboard/public/components/namespace-selector.js
avdaredevil: Fixed!

components/centraldashboard/app/api.ts
avdaredevil: Thanks, @prodonjs @yanniszark. While Jeremy and the team review the header based approach, I've included a configurable environment flag in af2039aadcd8776318a4d6833d60640abe66d36f.

components/centraldashboard/app/api.ts
avdaredevil: 1. Already implemented
2. The current behavior is to attempt to interact with Profile Controller API, if the API fails, then fallback action happens to fetch all namespaces

components/centraldashboard/app/api.ts
yanniszark: That's great, thanks a lot @avdaredevil! 
Once this is merged and a manifest is made, let's make sure that we make these configurable. kubeflow/manifests#179

yanniszark: @avdaredevil can you please rename the env vars to USERID_HEADER and USERID_PREFIX?
IAP is a Google-specific service and doesn't convey the correct meaning.
In addition, this is the name used in other places, like the profile controller.

components/centraldashboard/app/profile_route_handlers.ts
```diff
@@ -0,0 +1,74 @@
...
+// tslint:disable-next-line: no-any
+const G = global as any;
+G.USER_CACHE = {} || G.USER_CACHE;
+
+export interface AuthObject {
+    [iapHeader: string]: string;
+}
+
+export class UserData {
+    _req: Request;
```
prodonjs: *nit*: Be consistent with property names. Since Typescript does enforce visibility, I don't believe the underscore prefix convention is followed. Also, for properties assigned in the constructor you can use the shorthand form to assign them.

https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/

components/centraldashboard/app/api.ts
avdaredevil: Done! Sorry about that, wasn't aware.

components/centraldashboard/app/profile_route_handlers.ts
avdaredevil: Done!

avdaredevil:
Hey @prodonjs thanks for the timely reviews, a few things:
- I am aware of tests, and will tackle those after all the features are added
- I can't find your comments regarding the cookie approach, is there something I can take a look at as reference, so I know how you're envisioning this
- All PR comments so far have been resolved with 679660ce

prodonjs:
@avdaredevil I'm not sure what happened to the cookie comment. But, the gist of it was that it seemed like the middleware approach you were using would be a good candidate for [cookie-session](https://github.com/expressjs/cookie-session). It seems like the amount of data being stored per user session is small enough to fit inside the size limits for a cookie (~4Kb) and this feels like a pretty standard use of them to manage user-specific session data.

jlewi:
@avdaredevil @prodonjs how's this PR coming? Is it still a WIP or is it being reviewed?
Do we have an ETA for when this will be merged?

prodonjs:
@avdaredevil is out-of-office this week but AFAIK, this is still WIP. I've been involved only at a high-level for reviews, and am out of the office myself for a week starting Friday. My conservative estimate for when this would be ready to merge would be mid-week of 7/22 given the remaining work of writing tests, potential review-related changes, and end-to-end scenario testing.

prodonjs:
@avdaredevil See my email and corresponding [commit](https://github.com/prodonjs/kubeflow/commit/fe302499cd7cde8fd80285c5ba05f7118872c440) which refactors the backend to support integration with the KFAM service.

avdaredevil:
Looks good @prodonjs I've added your change, let's try to get this merged, and I am currently using your commit as the diff-base for my Manage-Users page.

components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -59,13 +79,22 @@ export class NamespaceSelector extends PolymerElement {
             <paper-menu-button no-overlap horizontal-align="right">
                 <paper-button id="dropdown-trigger" slot="dropdown-trigger">
                     <iron-icon icon="group-work"></iron-icon>
-                    <span>[[selected]]</span>
+                    <article id="SelectedNamespace">
+                        <aside class='owned-badge'
```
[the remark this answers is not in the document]
ajayalfred: What is this badge thingy? This is the first I'm seeing of it. Lol.
I would not recommend a blue dot. A blue dot means active, not ownership. If we want to convey ownership a simple string suffix "(Owner)" at the end of the namespace name will do.

components/centraldashboard/app/api.ts
```diff
@@ -28,16 +50,61 @@ export class Api {
...
+  /**
+   * Converts Kubernetes Namespace types to WorkgroupBindings to ensure
+   * compatibility between identity-aware and non-identity aware clusters
+   */
+  private mapNamespacesToWorkgroupBindings(
+      user: string, namespaces: V1Namespace[]): WorkgroupBinding[] {
+    return namespaces.map((n) => ({
+      user: {kind: 'user', name: user},
+      referredNamespace: n.metadata.name,
+      RoleRef: {
```
prodonjs: I missed this in the earlier review. This should be renamed to `roleRef` in accordance with the WorkgroupBinding definition. Not sure why the TS compiler didn't catch this given the return type of the function.

components/centraldashboard/app/api_test.ts
```diff
@@ -40,84 +67,323 @@ describe('Dashboard API', () => {
...
+      expect(mockK8sService.getNamespaces).not.toHaveBeenCalled();
       expect(mockK8sService.getPlatformInfo).toHaveBeenCalled();
+      expect(mockProfilesService.readBindings)
+          .toHaveBeenCalledWith('test@testdomain.com');
+      expect(mockProfilesService.v1RoleClusteradminGet)
+          .toHaveBeenCalledWith('test@testdomain.com');
+    });
+  });
+
+  describe('Has Workgroup', () => {
```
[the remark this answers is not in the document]
avdaredevil: Done!

components/centraldashboard/public/components/main-page.js
```diff
@@ -130,6 +141,25 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
...
+        this.$.envInfo.generateRequest();
+    }
+
+    /**
+     * Set state for loading registration flow in case no workgroup exists
+     * @param {Event} ev AJAX-response
+     */
+    _workgroupStatus(ev) {
+        const {user, hasWorkgroup, hasAuth} = ev.detail.response;
+        this._setIsolationMode(hasAuth?'multi-user':'single-user');
```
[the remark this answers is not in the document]
avdaredevil: Done

components/centraldashboard/public/components/namespace-selector.js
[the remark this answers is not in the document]
avdaredevil: Fixed:
image

components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -85,6 +114,12 @@ export class NamespaceSelector extends PolymerElement {
                 value: '',
                 notify: true,
             },
+            selectedNamespace: {
```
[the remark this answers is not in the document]
avdaredevil: Done

components/centraldashboard/public/components/registration-page.pug
```diff
@@ -0,0 +1,29 @@
+iron-ajax#MakeNamespace(method='POST', url='/api/create-workgroup', handle-as='json', last-error='{{error}}', last-response='{{resp}}', on-response='_successSetup', content-type='application/json', loading='{{submittingWorkgroup}}', on-input='clearInvalidation')
```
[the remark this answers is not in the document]
avdaredevil: Done

components/centraldashboard/public/components/registration-page.js
```diff
@@ -0,0 +1,83 @@
...
+        API.body = {referredNamespace: this.namespaceName};
+        API.generateRequest();
+    }
+    _successSetup() {
+        this.flowComplete = true;
+        this.sleep(800).then(() => {
+            const el = this.$.ApiMessage;
+            el.style.width = `${el.scrollWidth}px`;
+        });
+        this.sleep(6000).then(async () => {
```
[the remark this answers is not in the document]
avdaredevil: Implemented a bit of both in a seamless animation flow, let's discuss this over VC tomorrow if possible

components/centraldashboard/public/components/main-page.js
```diff
@@ -247,9 +277,15 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
     _onEnvInfoResponse(responseEvent) {
         const {platform, user, namespaces} = responseEvent.detail.response;
         this.user = user;
-        this.namespaces = namespaces;
+        this.namespaces = namespaces.map((n) => ({
+            user: n.user.name,
+            namespace: n.referredNamespace,
+            role: roleMap.tr(n.RoleRef.name),
```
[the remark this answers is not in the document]
avdaredevil: done

components/centraldashboard/webpack.config.js
```diff
@@ -9,8 +9,9 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const ScriptExtHtmlWebpackPlugin = require('script-ext-html-webpack-plugin');
```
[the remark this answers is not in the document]
avdaredevil: done

components/centraldashboard/app/api.ts
avdaredevil: done

components/centraldashboard/public/components/registration-page.js
```diff
@@ -0,0 +1,83 @@
...
+            namespaceInput: {type: Object},
+            userDetails: {type: Object, observer: '_onUserDetails'},
+            page: {type: Number, value: 0},
+            namespaceName: String,
+            error: Object,
+            flowComplete: {type: Boolean, value: false},
+            showAPIText: {type: Boolean, value: false},
+        };
+    }
+    ready() {
```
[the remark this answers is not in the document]
avdaredevil: Done

ajayalfred:
The string is "Owner" not "Owned". Owned has too many colloquial connotations (https://www.urbandictionary.com/define.php?term=Owned)

That said, "Owner" isn't a good match either since it's related to roles/users. Therefore doesn't work great in a list of namespaces.

What other strings can we use to define this relationship in the dropdown?

While we're brainstorming, let's at least use "Owner" in the menu.

components/centraldashboard/public/components/manage-users-view.js
```diff
@@ -0,0 +1,90 @@
...
+            newContribEmail: String,
+            contribError: Object,
+            contributorInputEl: Object,
+        };
+    }
+    ready() {
+        super.ready();
+        this.contributorInputEl = this.$.ContribEmail;
+    }
+    nsBreakdown(ns) {
```
prodonjs: Can you add doc block comments to these methods to describe their behavior? Also, it might be worth throwing some of the namespace transformation code that seems common between a few places into a shared module somewhere.

components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -110,6 +148,20 @@ export class NamespaceSelector extends PolymerElement {
         }
     }
 
+    /**
+     * Sets the query string namespace parameter to the selected value.
```
prodonjs: Update comment to reflect what this is doing.

components/centraldashboard/app/api.ts
```diff
@@ -96,14 +97,24 @@ export class Api {
...
+    }));
+  }
+
+  // tslint:disable-next-line: no-any
+  private surfaceProfileControllerErrors = (res: Response, errorMessage: string) => (err: any) => {
+    const status = (err.response && err.response.statusCode) || 400;
+    const error = err.body || errorMessage;
+    console.log(`${errorMessage} ${error}`);
+    res.status(status).json({error});
+    // tslint:disable-next-line: no-any
```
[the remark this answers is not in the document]
avdaredevil: Fixed!

components/centraldashboard/app/api.ts
```diff
@@ -96,14 +97,24 @@ export class Api {
...
+      roleRef: {
+        apiGroup: '',
+        kind: 'ClusterRole',
+        name: 'editor',
+      },
+    }));
+  }
+
+  // tslint:disable-next-line: no-any
+  private surfaceProfileControllerErrors = (res: Response, errorMessage: string) => (err: any) => {
```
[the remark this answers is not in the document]
avdaredevil: Good call, changed!

components/centraldashboard/app/api.ts
```diff
@@ -181,6 +193,19 @@ export class Api {
...
                   req.params.namespace));
             })
+        .get(
+            '/get-contributors/:namespace',
+            async (req: Request, res: Response) => {
+              const {namespace} = req.params;
+              const {body} = await this.profilesService
+                .readBindings(undefined, namespace)
+                .catch(this.surfaceProfileControllerErrors(res, `Unable to fetch contributors for ${namespace}`));
+              if (!(body instanceof BindingEntries)) return; // A response was already sent
```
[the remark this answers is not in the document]
avdaredevil: Same here, fixed!

components/centraldashboard/public/components/main-page.js
```diff
@@ -94,9 +104,12 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
             hideTabs: {type: Boolean, value: false, readOnly: true},
             hideNamespaces: {type: Boolean, value: false, readOnly: true},
             notFoundInIframe: {type: Boolean, value: false, readOnly: true},
+            registrationFlow: {type: Boolean, value: false, readOnly: true},
             namespaces: Array,
             namespace: {type: String, observer: '_namespaceChanged'},
             user: String,
+            isClusterAdmin: {type: Boolean, value: false},
+            isolationMode: {type: String, value: 'undecided', readOnly: true},
```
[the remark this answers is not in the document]
avdaredevil: Makes sense, however here's my take on this one:
- I prefer the UI being able to hold the exact state as this allows us to control what's seen around the page more seamlessly.
- Somewhat like version: `unknown`, `dev`, or `v1.x.x`.
- It's also a nicer syntax in Polymer when doing comparisons or showing it in the sidebar.

components/centraldashboard/public/components/main-page.js
```diff
@@ -153,6 +196,11 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
             this._setActiveMenuLink(this.subRouteData.path);
             this._setIframeLocation();
             break;
+        case 'manage-users':
+            this.sidebarItemIndex = 0;
```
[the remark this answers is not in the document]
avdaredevil: Right, that was a complete oversight. Thanks!

components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,5 +1,8 @@
-iron-ajax(auto, url='/api/env-info', handle-as='json',
+iron-ajax(auto, url='/api/has-workgroup', handle-as='json',
+    on-response='_workgroupStatus', loading='{{pageLoading}}')
```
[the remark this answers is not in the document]
avdaredevil: Makes perfect sense, done!

components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,5 +1,8 @@
...
+iron-ajax#envInfo(auto, url='/api/env-info', handle-as='json',
```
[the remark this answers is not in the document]
avdaredevil: Fixed with a computed prop: `_shouldFetchEnv`.

components/centraldashboard/public/components/main-page.pug
```diff
@@ -1,5 +1,8 @@
...
     on-response='_onEnvInfoResponse')
+aside#PageLoader(hidden='{{!pageLoading}}')
```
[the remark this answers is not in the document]
avdaredevil: There's no advantage when it comes to semantic HTML in terms of performance or browser rendering. div's were the default HTML 4 syntax to declare block-type elements. With HTML5 this was discouraged in favor of more expressive elements such as:
- `nav` - Navigational elements
- `main` - Main content on page
- `section` / `article` / `header` / `footer` - As the name suggests
- `aside` - Doesn't hold content, or is an additive element (like a helper widget or informational element)

There are some benefits however:
- SEO - where richer HTML allows the Search Engines to better understand the structure of the page thus result in better results in search
- Readability - the tag name suggests the behavior / purpose of the element or its children

components/centraldashboard/public/components/namespace-selector.js
avdaredevil: Done!

components/centraldashboard/public/components/main-page.js
```diff
@@ -245,11 +293,21 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
      * @param {Event} responseEvent AJAX-response
      */
     _onEnvInfoResponse(responseEvent) {
-        const {platform, user, namespaces} = responseEvent.detail.response;
-        this.user = user;
-        this.namespaces = namespaces;
+        const {platform, user,
+            namespaces, isClusterAdmin} = responseEvent.detail.response;
+        Object.assign(this, {user, isClusterAdmin});
+        this.namespaces = namespaces.map((n) => ({
```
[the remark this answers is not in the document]
avdaredevil: Can't because the iron-ajax is not in that component, and the registrationFlow state (in main-app) gets unset when namespace count is 0 (which is what allows the flow page to close and show the central-dashboard.

components/centraldashboard/public/components/manage-users-view.js
avdaredevil: Done. There is no transformation that's being repeated. And the one, in this case, is simply excluding from the list of all namespaces the `ownedNamespace` which was pre-computed in main-app

components/centraldashboard/public/components/main-page.js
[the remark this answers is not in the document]
prodonjs: Understood. If it's purely presentational and not used to drive any logic that seems OK. Might be handy to use an enum though if you like.

components/centraldashboard/public/components/main-page.js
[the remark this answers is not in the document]
prodonjs: I don't mean the ajax request itself but I just meant passing the response that's collected here directly to the child component(s) that deal with the namespace (BindingEntries) data. Just make sure you add/fix the unit tests that cover this.

components/centraldashboard/public/components/main-page.pug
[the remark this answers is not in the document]
prodonjs: Understood.

jlewi:
@avdaredevil How's this coming?

avdaredevil:
This is fixed and deployed at: https://kube1.endpoints.kubeflow-ap.cloud.goog/ under image: gcr.io/kubeflow-images-public/centraldashboard:v20190726-v0.6.0-rc.0-30-g2141a36b

prodonjs:
I wasn't able to access your cluster but I was able to deploy the image to an existing cluster and test it. @ajayalfred if you want to take a look at things try https://kf-master.endpoints.prodonjs-kubeflow-dev.cloud.goog. It should dump you into the registration flow.

avdaredevil:
Fixed @prodonjs @ajayalfred, you guys should be able to access the cluster now

avdaredevil:
@prodonjs Can we merge this?

prodonjs:
I'm OK with this moving forward as long as the project leads are onboard. I would recommend creating a followup issue to add unit tests as soon possible to try to keep coverage our coverage where it has been.

/lgtm
/approve

I think the hold is still in place until you remove it.

avdaredevil:
Removed hold as confirmed with @jlewi, tests will come in an incoming PR. All code has been manually tested in IAP and basic auth settings

jlewi:
@avdaredevil Thanks!

Lets use some of the existing issues or a new issue to track getting an updated image onto master and rolling it out.

PR #3513, opened 2019-06-21
gatekeeper: decode base64 encoded username

x28646510:
Hello, I got login error when I tried to deploy basic auth component on Azure AKS. Then, I found a bug that username is not decoded from base64 to plain string.
I have confirmed that this fix resolves login error.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3513)


x28646510:
Sorry, login error is my fault. I close this PR.

Issue #3497, opened 2019-06-19
upgrade to use kustomize v3+


kkasravi:
kustomize v2.0.3 had a number of limitations including not guaranteeing ordering of resources which caused us to create components for just crds. We should upgrade to v3+ and update manifests to remove workarounds for these limitations as well as take advantage of other new features.
Note: kustomize releases page is [here](https://github.com/kubernetes-sigs/kustomize/releases)

Kustomize 3.0 is being released the end of June 2019. Several features worth noting:
- bases are deprecated, everything is a resource. This makes overlays / mixins much simpler
- env becomes envs and can take a list - which could be the contents of param.env
- resource ordering is preserved
- apply will add an inventory object which can be used as a cache
- user plugins are given equal footing with built-in plugins
- plug-ins can be chained for example a generator plugin producing a list of applications could be the input to individual application generators
- edit can now remove resources 
- subcommands like apply can be implemented server-side using CRDs and an example is in cli-experimental

jlewi:
@kkasravi Do we want to do this before 0.6 or can we wait for 0.7?

kkasravi:
it would be valuable to get it in even if we don't fix the crds workaround.

jlewi:
@kkasravi @jbrette Any update on this?

What sort of risk is involved with upgrading to v3? Is v3 backwards compatible with our existing manifests?

jlewi:
Bump to P0 because we should figure out whether we are upgrading and if so when & how prior to Kubeflow 1.0.

jbrette:
@jlewi This should have been closed a long time ago. At least the upgrade in the kubeflow and the manifest has been merged. I did not follow since but I had in mind something was supposed to be replicated in the kfctl repo.

kkasravi:
@jlewi @jbrette yes we're using kustomize v3.1.0 in kubeflow/kfctl
https://github.com/kubeflow/kfctl/blob/master/go.mod#L89

kkasravi:
@jlewi this can be closed

Issue #3371, opened 2019-05-29
bootstrap: replace ksonnet with kustomize, move to v2


kkasravi:
Bootstrap should use the kustomize API and remove calls to ksonnet

kkasravi:
@kunmingg 
bootstrap/cmd/kfctl is mostly done removing references to ksonnet

should I do the same to bootstrap and have you review or would you like to do it and I can review?

thanks
Kam

kunmingg:
@kkasravi 
Great! Once https://github.com/kubeflow/kubeflow/pull/3045 merged everything should go through kfctl + kustomize.

jlewi:
Duplicate of #3368

Issue #3368, opened 2019-05-28
kfctl: remove ksonnet, move to v2


kkasravi:
We should remove all references to ksonnet and move to v2 for kfctl and bootstrap 

- [x] move kfctl code from under bootstrap to kubeflow/kfctl
- [x] kubeflow/kfctl/go.mod should now have module github.com/kubeflow/kubeflow/kfctl/v2
- [x] bootstrap go.mod should now have module github.com/kubeflow/kubeflow/bootstrap/v2
- [x] replace all references of github/kubeflow/kubeflow/bootstrap with either github/kubeflow/kubeflow/bootstrap/v2 or github/kubeflow/kubeflow/kfctl/v2
- [x] replace alls references to /tmp/v2 with kubernetes v1.13.4, client-go v10.0, kustomize v2.1.0 (#3497), controller-runtime v0.1.12 

background info:
there are  [2 alternatives](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher) for v2 modules. We currently use ‘major subdirectory’ but will move to ‘major branch’. 

In order to load both ksonnet and kustomize in kfctl we implemented v2 modules for the following (v2.zip):
- k8s.io:
  - api@v1.13.4
  - apimachinery@v1.13.4
  - cli-runtime@v1.13.4
  - apiextensions-apiserver@v1.13.4
  - apiserver@v1.13.4
  - client-go@v10.0
- sigs.k8s.io
  - application@sha67ae7f915d4ea994e0805886928d9ec40e9526a2
  - controller-runtime@v0.1.12
  - kustomize@v2.0.3

Additional info is [here](https://research.swtch.com/vgo-module)

jlewi:
Bump to P0 because the v2.zip code is causing problems.

To minimize churn; can we do a multi step process. In particular, can we get rid of the dependency on different versions of K8s libs without refactoring the directory structure in [kubeflow/bootstrap](https://github.com/kubeflow/kubeflow/tree/master/bootstrap)

My suggestion is as follows

1. Open several PRs to remove all imports of ksonnet

    * Open separate PRs for each module/package
    * e.g. one PR would remove the import in 
 https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/ksonnet/ksonnet.go
    * Rather than simply deleting ksonnet.go which would require us to delete the caller code; we can just replace code in ksonnet.go with returning an error.
[...]

Issue #2812, opened 2019-03-26
[Discussion] Where should kfctl go code live? New repository?


jlewi:
The code for kfctl is currently in the bootstrap directory
https://github.com/kubeflow/kubeflow/tree/master/bootstrap

We also have the kfctl code directory
https://github.com/kubeflow/kubeflow/tree/master/kfctl

I believe the intent was to eventually move the code into the kfctl directory or simply rename the bootstrap directory. We've postponed doing that though because this will be a disruptive change.

If we're going to move the code should we also move it to a new repo?

If we move it a new repo what else should go in that repo?

  * Should we move the click to deploy app into that repo as well?

/cc @kunmingg @gabrielwen @kkasravi @ashahba

kkasravi:
We could have clients in this repo, not sure about arena though

jlewi:
What is clients in this case? Do you mean plugins for specific platforms?

PR #3528, opened 2019-06-23
[WIP] Remove ksonnet. Move kfctl, bootstrap to v2

kkasravi:
Resolves the following issues:

- [x] bootstrap: replace ksonnet with kustomize, move to v2 #3371 
- [x] kfctl: remove ksonnet, move to v2 #3368
- [x] upgrade to use kustomize v2.1.0 #3497
- [x] Problem building kfctl - /tmp/v2 not properly created #2790
- [x] Provide a build option that will create a versioned package for k8 libraries #2636
- [x] upgrade to use kustomize v2.1.0 #3497
- [x] [Discussion] Where should kfctl go code live? New repository? #2812

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3528)


jlewi:
@kkasravi I think moving kfctl should be its own PR and I still have some questions about when exactly.

Moving the code is a risk of breaking a lot of things. There are lots of places that build the code and lots of outstanding PRs that will have merge conflicts when we do a massive move.

I actually think it makes sense to put kfctl into its own repository. So I think I lean towards leaving the code where it is now and then eventually moving development into its own repository kubeflow/kfctl.

jlewi:
Do the other changes like removing ksonnet, upgrading kustomize etc..., do these all need to go into one massive PR? 

I think it would really help if we can break these up to individual PRs.

jlewi:
Thanks @kkasravi 

In general, I think its really helpful if we break things down into small, focused PRs. This is less likely to create merge conflicts for other folks working on the code base. It also makes it easier to rollback in the event of a breakage.

Large scale refactorings that touch a lot of files (e.g. moving kfctl) or changing package imports should be handled with care because 1) they are likely to cause merge conflicts and 2) difficult to rollback.

w.r.t to kfctl; I think moving to its own repo makes sense long term but that's more work and I don't think we can do it before 0.6.

If we were going to stick with kfctl in repo kubeflow then I think it might make sense to do the move before 0.6. That would make it easier to cherry-pick fixes on the 0.6 branch. But if we decide to move to its own repo then I think we should just keep it in its own repo.

> It's slightly tricky to remove v2.zip and move bootstrap and kfctl to v2.

What's the connection between the move and removing v2.zip? I thought v2.zip was some hack for go modules because we have two different versions of K8s client that we pull in because of ksonnet. 

Can we remove that hack and always use a single version of our dependencies while still leaving all of our .go files at their current location?

Regardless, I would guess a prerequisite would be a PR just deleting the ksonnet code like
https://github.com/kubeflow/kubeflow/tree/master/bootstrap/pkg/kfapp/ksonnet

Also regarding the kustomize upgrade; should this be split into two PRs

1. Upgrade kustomize dependency to the new version
1. Remove/update any code that was used to work around limitations in the previous versions of kustomize that are now fixed.

kkasravi:
>If we were going to stick with kfctl in repo kubeflow then I think it might make sense to do the move before 0.6. That would make it easier to cherry-pick fixes on the 0.6 branch. But if we decide to move to its own repo then I think we should just keep it in its own repo.

Moving kfctl to kubeflow/kubeflow/kfctl before 0.6 would be a big help. Moving to its own repo after 0.6 would be changing kubeflow/kubeflow/kfctl to kubeflow/kfctl for imports, etc and should be straightforward.

> Can we remove that hack and always use a single version of our dependencies while still leaving all of our .go files at their current location?

we could and that could proceed any move - however bootstrap's ksServer.go is called from '/kfctl/apps/apply'. I don't know if this is still used but in this PR I removed references to ksonnet in bootstrap. @kunmingg should review first. 

but we could then do

1. remove ksonnet app and all ksonnet imports (requires cleanup in ksServer)
2. remove v2.zip - use single version of dependencies
3. move kfctl to kubeflow/kubeflow/kfctl. All refs to kfctl with be kubeflow/kubeflow/kfctl/v2
4. upgrade kustomize to v2.1.0
5. cleanup kustomize workarounds due to limitations in v2.0.3

jlewi:
@kkasravi I was thinking that maybe a good way to proceed would be to start by imports to "github.com/ksonnet/ksonnet/.

Rather than try to update the calling code to work with out ksonnet; I would just replace the code that uses it with a NotImplementedError.

In particular I think 
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/ksonnet/ksonnet.go

Is the main code that is still using ksonnet; so we can just update that module to return errors rather than call ksonnet code.

The advantage of this approach is

1. We can eliminate ksonnet from our dependencies
1. We don't have to update all the caller code in one go
    * e.g. KsServer can be updated in follow on PRs to replace calls to ksonnet with code that will work
      without ksonnet.

kkasravi:
closing this PR since the approach is to break this up into small pieces and have the ksonnet app return [NotImplementedError](https://github.com/kubeflow/kubeflow/pull/3528#issuecomment-511241258)

PR #3534, opened 2019-06-24
[WIP] Preview of a bunch of changes

jlewi:
This PR won't be submitted as is. It will be broken up into multiple PRs. It is meant to provide a preview

* application_types.go is the first PR to be submitted. There are a bunch of changes to support storing all of the parameters as part of KfDef

   * We add a plugins field which can be used to contain arbitrary structs for any plugins
   * The spec of the plugin is a runtime.RawExtension so that each plugin can  define its own struct
   * GCP will now be a plugin and all of its parameter will eventually migrate to the GCPPluginSpec
   * We also define secrets; secrets can be supplied as references (currently to environment variables) to prevent them from being stored in the spec and checked into git

* coordinator.go defines a backfillKfDefFromOptions method

    * This will be the 2nd PR
    * This backfills all the KfDef spec and plugin information from the current command line options
    * After this PR code (e.g. gcp.go) can be rewritten to use the new Plugins as opposed to using the current code.
    * It will also allow us to start deprecating command line flags while preserving existing code

* gcp.go

     * This will be the 3rd PR
     * It will be updated to get all of the information it needs from KfDef as opposed to passing around an extra GCPArgs struct

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3534)


kkasravi:
@jlewi - just starting to look at this.

jlewi:
Closing this I think all relevant changes have been ported over.

Issue #3502, opened 2019-06-19
Jupyter service on master not working


lluunn:
When I click "spawn", nothing happens.
Saw this in the pod
```
HTTP response headers: HTTPHeaderDict({'Audit-Id': '888aaf55-53d0-4086-99ba-b64cb2893c23', 'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosniff', 'Date': 'Wed, 19 Jun 2019 22:10:59 GMT', 'Content-Length': '344'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"podpresets.kubeflow.org is forbidden: User \"system:serviceaccount:kubeflow:jupyter-web-app-service-account\" cannot list podpresets.kubeflow.org in the namespace \"kubeflow\"","reason":"Forbidden","details":{"group":"kubeflow.org","kind":"podpresets"},"code":403}
```

 podpreset has been renamed to poddefaults.kubeflow.org

lluunn:
The jupyter webapp image is from May2, needs update.

lluunn:
Next error: the statefulset cannot be created
```
Warning  FailedCreate  9s (x14 over 30s)  statefulset-controller  create Pod test1-0 in StatefulSet test1 failed error: pods "test1-0" is forbidden: error looking up service account kubeflow/default-editor: serviceaccount "default-editor" not found
```

lluunn:
@kunmingg is there supposed to be a default-editor service account, and we missed it in kustomize?
Or we changed the name?
I can see there is a service account called jupyter-web-app-notebook-service-account

kunmingg:
@lluunn 
You can create a profile and use notebook in profile-namespace.
kubeflow namespace is a system namespace for shared components

kunmingg:
`jupyter-web-app-notebook-service-account` deleted in:
https://github.com/kubeflow/manifests/pull/158
It seems not have a consumer.

lluunn:
So users are not supposed to do stuff in kubeflow namespace?
Seems like an unnecessary constraint if the user is admin and doesn't need multi-user feature.
cc @jlewi

lluunn:
I think jupyter-web-app-jupyter-notebook-role-binding binds the wrong service account name, no prefix.
https://github.com/kubeflow/manifests/blob/master/jupyter/jupyter-web-app/base/role-binding.yaml#L10

jlewi:
[...]

Issue #2981, opened 2019-04-09
Notebook Spawner should check for edit permissions before spawning the notebook


kunmingg:
When talking to k8s API, Notebook Spawner backend should 
* read user id from jwt 
* impersonate as k8s user account
* k8s master then block or proceed based on rbac rule of k8s user account

yanniszark:
Following from [PR#3234](https://github.com/kubeflow/kubeflow/pull/3234), I think we should discuss how the userid will be inferred from the JWT.
The current implementation uses a google-specific header to get the userid, making it not work for other platforms.
I can think of two solutions for this:

1. Add a flag in each operator interacting with those headers (profile controller, notebook spawner doing impersonation etc), denoting which header it should use to read the userid. We should also add an option to infer it from the claims of the JWT.
2. Rewrite the platform-specific header or appropriate claim onto a new pre-agreed-upon header (like `x-kubeflow-authenticated-user-email`) using Istio Rules.

I have tried the 2nd option with this Rule and it works:
```yaml
apiVersion: config.istio.io/v1alpha2
kind: rule
metadata:
  name: google-header-rewrite
  namespace: istio-system
spec:
  match: source.labels["istio"] == "ingressgateway"
  requestHeaderOperations:
  - name: kubeflow-authenticated-user-email
    values:
      - request.headers["x-goog-authenticated-user-email"]
```

In addition, we should also be able to reference jwt claims, [as others have been able to](https://discuss.istio.io/t/question-decode-jwt-and-put-sub-into-a-request-header/1213/8?u=yanniszark).

jlewi:
@yanniszark Any update on the current plan based on our April 12 meeting? Unfortunately, the update @vkoukis posted into slack is no longer accessible.

IIRC correctly I think the plan is

* Each name space will have some service accounts with read or write permissions
* Web app will use that service account to create resources
* We discussed creating some sort of lookup service to map users to namespace.

Does the access management API provide the necessary functionality?
https://github.com/kubeflow/kubeflow/tree/master/components/access-management

jlewi:
Actually  here are @vkoukis notes
https://groups.google.com/forum/#!msg/kubeflow-discuss/Yu9DtqexIi4/ZPEc-NNGDAAJ

kunmingg:
access management API  can answer:
if user A can access namespace B.
So Notebook Spawner backend can behave like:
For each coming request:
1. parse user identity from header (provide by IAP or basic auth or LDAP)
[...]

PR #3537, opened 2019-06-24
jupyter-web-app: Integrate with PodDefaults

### LINKED ISSUE #2992: Jupyter spawner ui - let user select admission hooks to apply to the notebook
We are increasingly using webhooks to inject volumes and environment variables into Jupyter pods.

For example

  * On GCP we inject GCP credentials into Jupyter pods
  * We'd like to use to inject GIT secrets (see #2889 )

To support this we'd like the jupyter spawner UI to let users choose which admission hooks to apply to their notebooks.

Admission hooks are controlled based on labels. So step 1 would be to let users specify which labels to apply to their notebooks.

The next step would allow the web ui to introspect the cluster to see which admission hooks are installed in the cluster. Ideally each admission hook would have a human readable label describing what it does (e.g. as annotation). This would be something like

   * Attach S3 credentials for buckets ...

#2844 we'd like to generalize the admission web hook so users could easily define admission hooks for different things. Possibly by creating a CR similar to PodPresets.

Then the spawner UI could show a set of drop down boxes with these descriptions and a user could just check it. Under the hood the spawner UI could get the labels used by the corresponding admission hook and attach them. This will require some UI work as well as work on generalizing the web admission hook #2844.

/cc @zabbasi @vkoukis @kimwnasptd 

@zabbasi might pick up some of this work. I think a design doc would be a good place to start.

### LINKED ISSUE #3536: jupyter-web-app: Error 200 on load
In the new UI if no namespace is selected from the centraldashboard before opening the Jupyter UI, then the value `""` is passed as namespace. Because of this the GET will hit an invalid endpoint resulting in an error 200 popup.

The webapp should ignore the empty value from the centraldashboard.

kimwnasptd:
closes #2992 
closes #3536 

The user can select which PodDefaults will be applied in the Notebooks Pod. In the UI's `select` element the user sees the `desc` field of the `PodDefaults`. 

Under the hood the UI will attach the labels for the selected PodDefaults to the new Pods, which will trigger the `PodDefault` controller.

Also, the `config.yaml` has a [configurations](https://github.com/kubeflow/manifests/blob/master/jupyter/jupyter-web-app/base/config-map.yaml#L134) section from which an admin can set a default set of labels - PodDefaults that will be selected from the UI on load. The admin could also make the `readOnly` attr in the config `true` to block the users from editing this section.

@jlewi @zabbasi In the UI I've named the section as `Configurations`. Should we instead maintain the 1:1 naming with the k8s objects and rename it to `Pod Defaults`?
configs

/cc zabbasi
/cc prodonjs
/cc jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3537)


jlewi:
/assign @lluunn 
/assign @avdaredevil 

@avdaredevil Could you review the JS/Web changes please?

@lluunn Could you review functionality/admission hooks integration please?

components/jupyter-web-app/backend/kubeflow_jupyter/common/api.py
```diff
@@ -101,6 +101,17 @@ def get_notebooks(ns):
     )
 
 
+def get_poddefaults(ns):
+    return wrap_resp(
+        'poddefaults',
```
lluunn: just curious: the pattern here seems to be using single quote for first param, and double quote for third and later ones. What's the reason?

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
```diff
@@ -334,6 +334,23 @@ def set_notebook_memory(notebook, body, defaults):
...
+        labels = defaults["configurations"]["value"]
+        logger.info("Using default Configurations: {}".format(labels))
+    elif body.get("configurations", None) is not None:
+        labels = body["configurations"]
+        logger.info("Using form's Configurations: {}".format(labels))
+    else:
+        labels = defaults["configurations"]["value"]
+        logger.info("Using default Configurations: {}".format(labels))
+
+    for l in labels:
```
lluunn: do we need to check the labels is iterable?

lluunn:
Just minor questions. Backend LGTM

 I think we call it "configuration" might be confusing, both in FE and BE.

jlewi:
@kimwnasptd Our priority is unblocking the release. So if we have something that is good enough (sounds like we do) lets go with what we have and we can consider future improvements later.

If its an easy fix then someone just make the call about what to go with and lets do it.

jlewi:
@kimwnasptd Do you have an ETA for when we can get this PR merged and the package in kubeflow/manifests updated?

kimwnasptd:
@lluunn @avdaredevil updated the code to align with your comments

@jlewi I think we should go with `configurations` in this PR to unblock the release.

There are still some open issues regarging the webapp and multi user implementation like #2980 and #2981. I'd like to put priority on them first. Once we are done with them I can come up and rename to `Pod Defaults`.

If we won't have enough time before the release to make this change, then we could keep `Configurations` and I'll update the docs to make it clear that this section refers to `Pod Defaults`. Then changing it to `Pod Defaults` will be one of my first PRs for 0.7.

Also, I had already updated the `config.yaml` in the manifests so after we merge this PR we would only need to update the image.

WDYT?

jlewi:
Sounds good. We can create minor releases 0.6.x too if needed. 

Lets try to do the following ASAP

1. Get this PR merged
1. Get the image on master updated
1. Deploy on master and see what's working.

Specifically with regards to #2980. Does spawning notebooks work on non GCP?

e.g. do things work on minikf?

kimwnasptd:
Not in the `kubeflow` Namespace. The Notebooks that get created from the webapp will use the `default-editor` SA. If the creation of a Namespace happens from the Profile Controller then this SA will be created and the Notebooks Pod will be deployed.

But this SA is not present in the `kubeflow` Namespace. As you [said](https://github.com/kubeflow/kubeflow/issues/3502#issuecomment-503828467) we could create a profile for the `kubeflow` Namespace or we could create the `default-editor` SA from the webapp's manifests for `kubeflow`. Either way, lets open an issue to discuss/keep track of it there.

kimwnasptd:
It seems that the `kubeflow-presubmit-tf-notebooks-release` test is failing. IIRC this one builds the images for the Notebooks.

Checked the commits for `components/tensorflow-notebook-image` and found this [commit](https://github.com/kubeflow/kubeflow/pull/3587/files) from two days ago that updated the Dockerfile. 

Also, from the [logs](http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-tf-notebook-release-3537-416def6-7024-25bb?tab=workflow&nodeId=kubeflow-presubmit-tf-notebook-release-3537-416def6-7024-25bb-851366601&sidePanel=logs%3Akubeflow-presubmit-tf-notebook-release-3537-416def6-7024-25bb-851366601%3Amain) I can see that it fails when it tries to 
```bash
/bin/bash -c pip install --upgrade pip==19.0.1 && pip --no-cache-dir install ${TF_PACKAGE}.... ${PIPELINE_SDK_PACKAGE} ....
```

@IronPan could the new version of the Pipeline SDK be the problem or it might something else?

jlewi:
@kimwnasptd Can you try rebasing on the latest master?
I don't think this PR should be triggering the notebook presubmit if it didn't modify that code. 
Its probably doing that because you haven't picked up the latest changes from master so it things this PR is modifying the code in the notebook images.

jlewi:
/lgtm
/approve
/hold

I'm rubber stamping the PR. If the PR is good to go and the tests pass just remove the hold.

Issue #3518, opened 2019-06-21
kfctl semantics replace init & generate with build  and also add apply


jlewi:
Should we change the semantics of kfctl and replace the init and generate commands with

```
kfctl build --config=path to config file
```

and 

```
kfctl build
```
   * Look for app.yaml in the current directory

kfctl build would not create an app directory but instead treat the current directory as the app dir.

This eliminates the need to cd into the app directory after running `kfctl init`. Right now the directory creation that happens in init is akward because the next two commands generate/apply then execute from the created directory.

One time `init` actions (like API enablement). Should happen during apply. 

We can do that by changing the semantics of kfctl to act like a K8s controller and treating the apply method as the reconcile method.

* kfctl will repeatedly call apply until some conditions in KfDef.Status is reached
* `app.yaml` will be updated so conditions are persisted on error
* KfDef will be passed back and forth between the KfApps so that KfApps can modify KfDef.Status

I think this would also allow post apply platform specific logic like @yanniszark suggested e.g. printing out the endpoint to access Kubeflow.

kkasravi:
right now after `kfctl generate` you get a collection of microservices
```
admission-webhook.yaml   cert-manager.yaml        katib-ui.yaml            pytorch-job-crds.yaml
ambassador.yaml          cloud-endpoints.yaml     metacontroller.yaml      pytorch-operator.yaml
api-service.yaml         gpu-driver.yaml          metrics-collector.yaml   scheduledworkflow.yaml
application-crds.yaml    istio-crds.yaml          minio.yaml               spartakus.yaml
application.yaml         istio-install.yaml       mysql.yaml               suggestion.yaml
argo.yaml                istio.yaml               notebook-controller.yaml tensorboard.yaml
basic-auth-ingress.yaml  jupyter-web-app.yaml     persistent-agent.yaml    tf-job-operator.yaml
basic-auth.yaml          katib-controller.yaml    pipelines-runner.yaml
centraldashboard.yaml    katib-db.yaml            pipelines-ui.yaml
cert-manager-crds.yaml   katib-manager.yaml       pipelines-viewer.yaml
```
if we did something like kfctl build && kubectl apply <output-dir> then would we have 
kfapp-controllers (eventually application-controllers) deployed as well as KfApp CRs where the kfapp-controllers would claim ownership of KfApps and update their status?

jlewi:
I guess we might do that but I think my proposal is a little simpler 

The current UX is

```
kfctl init ...
[...]

PR #3515, opened 2019-06-21
Refactor kfctl to begin address a number of issues related to repo management

jlewi:
* kfctl needs to support multiple repositories since the configuration is
  now split across kubeflow/kubeflow and kubeflow/manifests

* We should use KfDef.Status to pass information through the kfctl libraries
  that is temporary such as keeping track of locations where code is cached.

  * Prior to this repo the fields Repo and ManifestsRepo were reset to
    point to the local cachedir. This means those links were no longer valid
    if the app was moved to a different folder or machine

  * With this PR we add Spec.Repos to keep track of source of truth
    and Status.ReposCache to store information about the local cache

    * This way the app can always rebuild the cache if needed.

* Use hashicorp's go-getter to provide a consistent syntax for downloading the
  repos. go-getter supports fetching URIs and specifying lots of variants (e.g pull) no need to invent our own syntax.

* Define a SyncCache method on KfDef that will sync multiple repositories.

* Refactor NewKfApp; split it into 3 well defined methods with clearer semantics

   1. CreatekfDefFromOptions
     
       * This will parse command line options and construct a KfDef from the command line options
       * This will call the newly added backfileFromoptions method to backfill from KfDef from the many 
          command line arguments we currently have (e.g. --email)
       * It will not attempt to write the KfDef to any file

    1. CreateKfAppCfgFile
        
        * This will attempt to write the KFDef to app.yaml
        * It will create the directory if needed
        * It will return an error if the app.yaml exists

    1. Constructs a KfApp from an app.yaml file
        
         * This is intended to be the main entry point once the config file exists.
         * Click To Deploy can use this to load the app after writing the CfgFile to disk.

* Make repos a list and allow specifying URIs that go-getter will fetch

  * Also allow specifying the root path inside the untarred file. We do
    this because github appears to create different subdirectories in the
    tarballs in unpredictable ways.

* kfctl_gcp_iap.0.6.yaml is intended to be a single static config for
  initiating GCP.

  * Relates to #3419 (provide a fixed static set of YAML files for KfDef)

* Define KfDef.SyncCache to provide a systematic way to synchronize all
  repos. This will allow us to get rid of the code in kustomize.go to
  give special treatment to kustomize.

* Define backfill methods to port existing command line flags to KfDef

   * We need to have two different methods because kfctl init && kfctl generate take different command
      line arguments.

   * Update kustomize.go to call SyncCache if needed and and to use new fields.

Related to #3491 (kfctl discovery logic is flawed)
[...]

PR #3538, opened 2019-06-24
kfctl: add platform existing

yanniszark:
This PR adds a new platform implementation, platform=existing.
Platform existing is designed to not depend on any cloud-specific feature and be compatible with any existing Kubernetes cluster.

The PR is organized in commits.

### Clone the repo

```bash
git clone https://github.com/kubeflow/kubeflow.git
KF_REPO="$(pwd)/kubeflow"

cd $KF_REPO 
git fetch upstream pull/<PR_NUMBER>/head:kfctl-platform-existing
git checkout kfctl-platform-existing
```

### Build the kfctl binary

Go into the bootstrap folder and build the kfctl binary.
The resulting binary is available inside the bin folder.

```bash
cd $KF_REPO/bootstrap
mkdir -p bin
make build-kfctl-container
```

### Follow the instructions from docs

Follow the instructions from [the corresponding docs PR](https://github.com/kubeflow/website/blob/4ff9e654c14670a035070898ae049012187e219a/content/docs/started/getting-started-k8s.md).
Change config to: 
```bash
CONFIG=$KF_REPO/bootstrap/config/kfctl_platform_existing.yaml
```

and edit the config repos, changing the kubeflow repo to:
```
- name: kubeflow
  uri: file://<path_to_kubeflow_folder>
``` 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3538)


ryandawsonuk:
I stepped through running this in GKE with @yanniszark was able to get it up and running and could log in to the cluster. We noticed a few small things along the way that @yanniszark is going to follow up on
/lgtm

jtfogarty:
I'd like to see this included in 0.6 
Thanks

yanniszark:
> from the build failure it looks like this is still using ksonnet. Perhaps rebase from master?

I just rebased yesterday so I thought it would be good :sweat_smile: 
Well, I have rebased again on the latest master, let's see how this goes!

yanniszark:
cc @jlewi 

> Why do you need special logic for installing ISTIO? Can't we just handle this as a kustomize package?

We need to first install Istio, get the LoadBalancer IP and then apply Dex and the AuthService with the correct config.
Essentially, we need to get the LoadBalancer IP first.

> What resources are we creating from raw YAML? Any reason we can't handle them as kustomize packages?

Istio, Dex, the AuthService.


In the near future, these will be moved into kustomize packages.
Especially for Istio, this is also relevant: https://github.com/kubeflow/kubeflow/issues/3540

The issue of not knowing the IP beforehand can be solved either with the new design of looping between k8s and platform apply (https://github.com/kubeflow/kubeflow/issues/3518) or by adding an agent inside Kubernetes (as a kustomize package).

In the first iteration, these manifests will live under `deployment/existing` as we do with platform specific configs.

jlewi:
@yanniszark I suggest renaming the value of platform to something like "istio-dex". I think this will create a much better UX when replace platform with plugins.

if we tell a user to add plugin "istio-dex" to their KfDef; I think its natural to interpret that as configuring/installing ISTIO & dex.

In contrast telling a user to add plugin "existing"; is not all intuitive what that means.

The interpretation of "existing" is also confusing.  Its natural for a user to assume that if they want to install Kubeflow on their existing K8s cluster they should set "--platform=existing". But just because a user has an existing cluster doesn't mean installing "istio-dex" is the correct thing to do. e.g. what if that cluster has ISTIO already installed? Or if the user doesn't have sufficient  permissions to install ISTIO.

I think if we treat it as a plugin and name it based on what that plugin does; i.e. install ISTIO & Dex then I think its more intuitive.

In this PR I think we only need to change the user facing bits; i.e. value of the user facing flag. Internal bits like file names could be changed later.

Thoughts?

jlewi:
@kkasravi yes you are correct.

> As discussed in the past, the purpose of a platform is to offer an opinionated installation of Kubeflow.
Yes but the "--platform" flag is ill-named and going away. The way people will publish an oppinionated config is by publishing a YAML file that fully describes everything. 

> Platform existing is opinionated, in the sense that it avoids depending on any cloud-specific feature and will install successfully on any existing Kubernetes cluster.

This isn't quite true. If a user has an existing cluster with ISTIO installed then installing ISTIO on it will likely cause problems.

To @kkasravi 's point about logging. Logging is another example of a pluggable component.

Imagine we define a reference logging solution; e.g. a cluster level  fluentd logger that dumps to a PD.

So now we have 2 plugins. "istio-dex", "pd-cluster-logger". These can now be composed you can use one, both or none. 

We can now construct an opinionated config just by publishing a YAML file with these plugins listed.

Existing platforms will be moving to plugins. "gcp" as a plugin name may no longer be a good fit but for backwards compatibility/migration that's where we'll start.

Since this is the first PR introducing "existing" I think it makes more sense not to persist a pattern that no longer makes sense and instead change the name to something more likely to make sense going forward.

jlewi:
@yanniszark I know this PR/work hasn't been easy and you're running smack into all the churn in kfctl; is there anything I can do to help and unblock you?

@kkasravi Yes that is pretty much what I'm thinking except that Platforms and PackageManagers morph into Plugins. I filed #3566 to track the refactor as part of 0.7.

vkoukis:
@jlewi
Hey Jeremy!

> I agree the way to publish an opinionated config is going to publish a YAML file and use --config.

We agree with this as well. This has been our intention all along, and that's why @yanniszark after the agreement you had to move towards independent YAML files as arguments to `--config`.

> To your point about about it being "service"; I'm using the term "plugin". Kubeflow is intended to be pluggable and one of the pluggable bits of the architecture is the service mesh gateway and AuthN solution.
>
> So I like the term "plugin".
> 
> I agree that plugins are meant to be composable. So we might also have a plugin for logging and wand to compose the logging plugin with the ISTIO+dex plugin.
>

We don't have a problem with moving towards plugins as well. However, this PR is not about adding a new plugin, it's about adding a new platform/config named "existing". We had established consensus on this following our Slack discussion, initially here, after your proposal to define a new platform:
https://kubeflow.slack.com/archives/CF94DURGF/p1559836215017900?thread_ts=1559700461.000900&cid=CF94DURGF
> Instead of using a new flag introduce a new platform. The point of a platform is too have an oppionated way of installing Kubeflow with a clear set of owners who are the deciders of what those opinions should be.

and then here, where we agreed on the name, after your proposal:
https://kubeflow.slack.com/archives/CJUJ79T28/p1559926856031800?thread_ts=1559926825.031700&cid=CJUJ79T28
> 1. Lets pick a suitable name to be used as the platform



> (maybe "existingcluster"?)

Following that, we demo-ed the proposed reference architecture during the community meeting, and it seems to be getting traction within the user community, e.g.,
https://github.com/kubeflow/kubeflow/pull/3538#issuecomment-505416614 by @ryandawsonuk
>I stepped through running this in GKE with @yanniszark was able to get it up and running and could log in to the cluster. We noticed a few small things along the way that @yanniszark is going to follow up on.
> (Would also be cool to support minikube but I think that could be a separate PR.)
> /lgtm

and https://github.com/kubeflow/kubeflow/pull/3538#issuecomment-505504357  by @jtfogarty
>I'd like to see this included in 0.6
>Thanks

So, while we agree with the general notion of plugins as a direction for kfctl, and you have already opened an issue for 0.7:
> I filed #3566 to track the refactor as part of 0.7

This PR is about the *platform* UX. Even when we do have the ISTIO+Dex functionality packaged as a plugin, which can well be named "ISTIO+Dex", this PR is about laying the groundwork for an opinionated reference architecture named "existing". The architecture is something bigger than a single plugin. When we have such a plugin mechanism, we can have an "ISTIO+Dex" plugin and swtich the config to refer to it. But, before that, we need our own, opinionated config file, which we will be augmenting with more plugins; for example, we will include cert-manager functionality, instead of self-signed HTTPS.

This is not unlike what you are currently doing with GCP. Currently, your opinionated approach installs ISTIO. In the future, you plan to switch to Google-managed ISTIO. You also plan to switch to Google-managed SSL, and you are probably unwilling to rename "gcp" to "gke+iap+istio+ssl", and then rename it to "gke+iap+managed_istio+managed_ssl" or something like that.

So, the `--config existing` UX, and having an opionionated reference architecture coded as a distinct config file *is* the UX we are actually targeting, and have validated with the community. So I'd respectfully ask we keep it.

> My only real hangup right now is "existing". I don't think there is a single config that will work for all existing clusters so I think the term "existing" is problematic and will cause confusion.

Well, the same thing could be said about `gcp`; there is no single config that will work for any cluster on GCP, it's just a name for your opinionated reference architecture on GCP.

We do need a name for our reference architecture, and based on the above a name like "ISTIO+Dex" can only refer to *part* of this architecture [a plugin], given that we will augment this architecture with more components, and may as well migrate away from this plugin in the future.

The "existing" name was in fact your proposal, and a good name in our opinion. Our goal is *not* to create a config that will work with any existing cluster. We are opinionated about what our reference architecture expects on this "existing" cluster, but we are not trying to make it work everywhere.
As it currently stands, kfctl cannot even run on an existing on-prem cluster, since all current config files make GCP-specific assumptions, and this was the reason why you proposed we move with an independent config file in the first place.

Our current goal for config "existing" is to assume no-cloud specific functionality. Of course, we will be documenting its assumptions, and its operation in the docs. This should minimize the possibility of any misunderstanding. So, do we guarantee that it will work everywhere? Not at all. But, then, as you said, this would be a matter of creating a new opinionated config, with a different set of requirements.

So, I would propose we keep "existing". We can always rename the config, and adjust the documentation accordingly, if we have any sort of user feedback that this creates confusion.

kkasravi:
Moving to plugins flattens the model - as you’ve noted - to opinionated config files. All plugins need to implement the KfApp API, and all config files can include overlays. So this PR can go in mostly as is other than what we call the config file and what parts are moved to manifests. Since istio is already in manifests, it would make sense to use istio. The other dex/authn could be moved now or in a follow up PR. This would leverage almost all of Yannis’s work IMO and we avoid any discussion of whether this is a platform which no longer applies. Thoughts?

jlewi:
@vkoukis I'm really hung up on the name "existing". I thought I'd raised my concerns about it before but I agreed to it because I couldn't think of something different. 

I think we now have a better name that is more future proof and will cause less user confusion. I believe this is a 1 line change; just change the value of the constant.

I think of "existing" as logically different from "gcp", "aws", "rok", "openshift", "azure", etc... The difference is that a value like "gcp", "aws", "rok", "arrikto" and "openshift" are clearly associated with a group that should be the final arbitrators of what belongs and what doesn't.

From a user's point of view, I don't think a user would expect any of these values to correspond to generic, consensus approved configs.

The challenge of a generic reference architecture is who gets to decide what assumptions are the right one?

> this PR is about laying the groundwork for an opinionated reference architecture named "existing".

Whose opinions are you planning on representing? We as a community haven't established a process for resolving disputes and forming opinions representative of Kubeflow as a whole. 

A name like "existing" and "reference" is unclear about who owns those opinions and gets to arbitrate those decisions.

So I'd like to come up with a name that tries to sidestep that. I think there are two approaches

1. Pick a name e.g. "arrikto" that clearly conveys whose oppinions are represented and who is the arbitrator (and message it as such)

1. Pick a name that provides less room for interpretation

     * A name like "istio-dex" provides less room for interpretation
     * Its more narrowly scoped which means its easier to provide backwards compatibility

jlewi:
> This is not unlike what you are currently doing with GCP. Currently, your opinionated approach installs ISTIO. In the future, you plan to switch to Google-managed ISTIO. You also plan to switch to Google-managed SSL, and you are probably unwilling to rename "gcp" to "gke+iap+istio+ssl", and then rename it to "gke+iap+managed_istio+managed_ssl" or something like that.

Actually that's pretty much what we are planning.

#3515 Added a "--config" file which allows people to specify a URI pointing at their config file. That also adds very flexible support for URIs so config files don't need to be stored in a specific repo.

We are planning on publishing multiple configs corresponding to different common scenarios. In that PR
I started[kfctl_gcp_iap.0.6.yaml](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.0.6.yaml)

We will likely also have a `kfctl_gcp_basic_auth.0.6.yaml`.

We will likely also start to publish manifests for different collections of applications; e.g. some users just want pipelines.

Composability is one of our key goals and something a lot of customers have been asking us for.

So we're trying to move in the direction where there are lots of example configs and you find one close to what you want and then you modify it.

So the goal is composability. This is why I think treating what you have here as an "istio-dex" plugin is more powerful then treating it as a reference architecture and more future proof.

A plugin can now be composed with other plugins. So that plugin can be combined with the gcp or aws plugins so that not only does a user get istio+dex but the cluster & cloud resources are provisioned as well.

Or it could be composed with the Rok applications/plugins.

We are moving towards semantics like

```
kfctl build --config=http//..
kfctl apply
```

If I could I'd get rid of `--platform` and all the other flags now, but we're too close to the RC.

jlewi:
I'd like 0.6 to provide a migration path from

```
kfctl init --platform=....
kfctl generate  ....
kfctl apply ...
```

to 

```
kfctl build --config=http//..
kfctl apply
```

Both will be supported in 0.6 but only something like the latter in 0.7.

To do this we will start updating kfctl  so that if a user does 

```
kfctl init --platform=....
kfctl generate  ....
kfctl apply ...
```

we end up creating a KfDef in app.yaml that is consistent with

```
kfctl build
kfctl apply
```

What this means is that every "--platform" will result in a Plugin being added to KFDef.Plugins.

So it makes sense to pick a value for --platform that will make sense as a plugin.

i.e. at some point we should publish a YAML that looks like

```
apiVersion: kfdef.apps.kubeflow.org/v1alpha1
kind: KfDef
Spec:
   plugins:
       - name: istio-dex
         username: myuser
         password: 
             name: basic-auth-password

   ...
```

A plugin named "existing" is less self-explanatory IMO.

So I hear you loud and clear about wanting to unblock the PR. I also  hear you and understand that we agreed on a plan and now that plan's changing.

I'm hoping that changing the name is minimally disruptive while also aligning us better for where we want to go.

jlewi:
@yanniszark @vkoukis Thoughts? I'd like to see this go in. On my end I think if we can agree on the name this is good to go. I think the remaining refactoring can happen later.

kkasravi:
Do overlays remain in the plugin model? If so, is there a sense of when a plug-in would be required vs an overlay? I think I can describe the differences assuming overlays are still leveraged, but I was wondering if this was changing. 

Also were you thinking that kustomize-v3.0 would go into a 0.6 point release? It has a number of significant new features.

jlewi:
@kkasravi lets take that discussion off this PR since I don't think its directly relevant.

vkoukis:
We understand your concerns and would like to propose the following solution:


@jlewi
Hey Jeremy, thanks for the detailed answer.

We understand your concern about making clear who the arbitrator for each config file is.

> So I'd like to come up with a name that tries to sidestep that. I think there are two approaches
> 
> 1. Pick a name e.g. "arrikto" that clearly conveys whose oppinions are represented and who is the arbitrator 

We like this approach, because it will allow us to iterate more quickly and efficiently in improving our proposal for deploying Kubeflow in existing clusters.
So, let's use the name `kfctl_existing_arrikto.0.6.yaml` for the kfctl config file.

It conveys explicitly that it is meant to run on existing Kubernetes clusters and represents our assumptions/opinions/expectations.

> (and message it as such)

Absolutely. We will modify the documentation to describe exactly what this config file assumes, what it expects, and what it deploys, making explicit that this config file is maintained by Arrikto. This is how we are also going to be messaging it, when replying to any support requests on Slack.

Now, on the (temporary, for 0.6) issue of how to name the `platform` attribute:

> So the goal is composability. This is why I think treating what you have here as an "istio-dex" plugin is more powerful then treating it as a reference architecture and more future proof.

> A plugin can now be composed with other plugins. So that plugin can be combined with the gcp or aws plugins so that not only does a user get istio+dex but the cluster & cloud resources are provisioned as well.

We really like the idea of a plugin mechanism in kfctl.
We agree that it makes sense to name the plugin "istio+dex", for composability.
We will definitely move towards it in v0.7, and @yanniszark will continue to invest effort, similarly to how he helped start support for `--config` in the first place.

> What this means is that every "--platform" will result in a Plugin being added to KFDef.Plugins.
> So it makes sense to pick a value for --platform that will make sense as a plugin.

We understand that you want to name the new platform `istio+dex`, since that's what the name of the plugin in 0.7 is going to be. However, the current notion of platform in 0.6 is something quite different.

Since platforms currently are a part of v0.6, we believe the name of the platform (while platforms still exist), should be representative of its meaning as you had defined it in our Slack conversation:

> an opionated way of installing Kubeflow with a clear set of owners who are the deciders of what those opinions should be.

Because of this, we propose the value of platform also becomes `existing_arrikto` in this PR.
In 0.7, we will convert the istio+dex functionality into its own plugin, all code that currently exists as platform `existing_arrikto` will become part of the reconciliation mechanism for the `istio+dex` plugin, and only the `kfctl_existing_arrikto` config file will remain, so we can continue adding new plugins to it.

WDYT?

jlewi:
"--platform=existing_arrikto"  sounds good to me.

yanniszark:
@jlewi thanks for the detailed and fruitful discussion.
I have made the following changes to the PR:

* Changed `existing` to `existing_arrikto`.
* Added some logic so that Kubeflow master is downloaded by default, using the new repo mechanisms. The `GetRepo()` and `SetRepo()` functions help with that.
* Minor fixes in platform `existing_arrikto`.

All changes are in new commits, to make reviewing easier.
Please take a look whenever you can :smile:

yanniszark:
> What code are you referring to? I'm referring to all the code you added here to set the repo.
>I don't think that code belongs here in GetPlatform.
>
>Regardless I propose moving that to a new PR so we don't keep blocking this PR.

The code was setting the manifests repo even before I added the helper function.
However, I removed everything from the extra logic and instead changed the config file to refer to master and the code in existing_arrikto to use the correct path.

jlewi:
My mistake looks like the changes were showing up in reviewable but not git; probably because reviewable wasn't taking into account the rebase.

jlewi:
@yanniszark Do you know why the test failed?

yanniszark:
> @yanniszark Do you know why the test failed?

Not yet.
I rebased again today and will see if that fixes it.

deployment/existing/auth_oidc/authservice.tmpl
```diff
@@ -0,0 +1,53 @@
...
+  template:
+    metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
+      labels:
+        app: authservice
+    spec:
+      containers:
+      - name: authservice
+        image: gcr.io/arrikto/kubeflow/oidc-authservice:v0.1
```
ryandawsonuk: Where can we find the source code for this image?

ryandawsonuk: Is it https://github.com/yanniszark/ambassador-auth-oidc ?

yanniszark: Exactly, in the `kubeflow` branch.
The upstream repo is https://github.com/ajmyyra/ambassador-auth-oidc.
I have made upstream issues for some of the fixes I had to make and I'll try to merge all the extra code upstream soon.

yanniszark:
@jlewi looks like the CI is ok now.
Can you please approve again ?

ryandawsonuk:
Great to see this merged! Nice work guys. @yanniszark @sarahmaddox  is there going to be a new version of the https://github.com/kubeflow/website/pull/833 PR which got reverted?

deployment/existing/auth_oidc/authservice.tmpl
ryandawsonuk: is the branch 'kubeflow' or 'feature-kubeflow'?

PR #3551, opened 2019-06-26
Fix 403 forbidden response when downloading file from JupyterLab in Google Chrome

gracechristina:
Setting `c.NotebookApp.allow_remote_access = True` in `jupyter_notebook_config.py` to fix the inability to download file to local disk from JupyterLab in kubeflow. Previously, without that configuration, user is not able to download file in Google Chrome and an error message of `Failed-Forbidden` is received.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3551)


gracechristina:
Signed CLA and re-triggering the bot.

PR #3557, opened 2019-06-26
add centraldashboard application CR

kkasravi:
adds a [centraldashboard application CR](https://github.com/kubeflow/manifests/issues/126)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3557)

### DIFF
--- a/bootstrap/config/overlays/application/kfctl_default.yaml
+++ b/bootstrap/config/overlays/application/kfctl_default.yaml
@@ -12,7 +12,7 @@ spec:
     application:
     - name: overlay
       value: application
-    jupyter:
+    centraldashboard:
     - name: overlay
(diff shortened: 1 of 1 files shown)


bootstrap/config/overlays/application/kfctl_default.yaml
```diff
@@ -12,7 +12,7 @@ spec:
     application:
     - name: overlay
       value: application
-    jupyter:
+    centraldashboard:
```
jlewi: Was "jupyter" a bug?

kkasravi: We’re removing this component so yes

PR #3565, opened 2019-06-27
Removing unnecessary rbac authorization for tf operator

johnugeorge:
Configmaps are not used in operator anymore. Unnecessary rules are also removed

/hold

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3565)


jlewi:
@johnugeorge Do we also need to update 
https://github.com/kubeflow/manifests ?

ksonnet is no longer really used.

johnugeorge:
Yes. I just thought of keeping the changes in sync. I will make same changes in manifests also

jlewi:
@johnugeorge I think its fine to commit this.

PR #3570, opened 2019-06-28
Create a fake coordinator to support unittesting.

jlewi:
* To support unittesting of the click to deploy server we create a fake
  implementation of coordinator.KfApp

* We also define the Builder interface so we can inject a constructor
  that will create a fake as opposed to real implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3570)

### DIFF
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -40,6 +40,19 @@ import (
 	"strings"
 )
 
+// Builder defines the methods used to create KfApps.
+// Primary purpose is to allow injecting a fake for use in testing.
+type Builder interface {
+	LoadKfAppCfgFile(cfgFile string) (kftypes.KfApp, error)
+}
+
+type DefaultBuilder struct {
+}
+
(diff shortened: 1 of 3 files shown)


jlewi:
Tests failed because it fails to build because of import

```
util.py                     71 INFO     pkg/kfapp/coordinator/fake/fake.go:6:2: unknown import path "github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/gcp/
fake": cannot find module providing package github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/gcp/fake
```

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -40,6 +40,19 @@ import (
 	"strings"
 )
 
+// Builder defines the methods used to create KfApps.
+// Primary purpose is to allow injecting a fake for use in testing.
+type Builder interface {
+	LoadKfAppCfgFile(cfgFile string) (kftypes.KfApp, error)
```
lluunn: do you need create here? and below?

jlewi: I don't think so. This is sufficient right now for testing kfctlServer (part of click to deploy).
I can; will add more methods in the future if we need it.

Issue #3566, opened 2019-06-27
Refactor  kfctl package managers, platforms and plugins 


jlewi:
Opening this issue to track refactoring of how platforms and package managers into platforms.

I think we want to make the following changes

* Unify Platform and Package managers by treating them as plugins
* A Plugin operates on KfDef and can store state in KfDef.Status
* A plugin implements a simple interface like

   ```
   type Plugin interface {
         Reconcile(*KfDef) error
   }
   ```
   * There's some additional complexity as we need to support injecting plugin specific information like clients and credentials for cloud services

* Plugins have controller/reconcile like semantics

   * They operate on KfDef.Spec and store state in KfDef.status
   * Following K8s conventions Status.conditions should be used to track state

* Plugins store structures specific to their plugin in KfDef.Spec.Plugins (e.g. as runtime.Rawextension)
* kfctl would repeatedly invoke reconcile until some conditions are reached for each plugin (e.g. failed or succeeded)

lluunn:
cc

Issue #3520, opened 2019-06-21
Fix dynamic loading of platform plugins


jlewi:
#3045 disabled dynamic loading in kfctl of KfApp plugins for platforms. Plugins are still supported for the package manager.

The relevant change is here
https://github.com/kubeflow/kubeflow/tree/1c07de4d4825f381062c7d7091398b837113b7f1/bootstrap/pkg/kfapp

Specifically getPlatform no longer calls LoadKfApp to dynamically load the plugin if there is no match.

This change was made because getPlatform's return type was changed to kftypes.Platform which is no longer compatible with LoadKfApp.

This was changed because #3045 also introduced the interface kfTypes.Platform in order to add a GetK8sConfig method to allow plugins to configure how K8s credentials are obtained.

Possible fixes

1. Define a LoadKfPlatform method that would dynamically load platform plugins

    * Code would be similar to LoadKfApp but return type kftypes.Platform

1. Get rid of kftypes.Platform by adding GetK8sConfig to the KfApp interface

    * It could be a null op in the case of package managers.

Also related to #3518 which suggests adopting semantics like reconcile.

jlewi:
This will be fixed in an upcoming PR in which I add plugins to KFDef.Spec and allow each plugin to provide an arbitrary struct as a runtime.RawExtension

jlewi:
#3651 refactored the coordinator code.

PackageManagers and Platforms should now have the same interface; platforms no longer take extra args during construction.

It should not be relatively straightforward to add back dynamic loading of plugins.

gabrielwen:
@jlewi I think this was fixed?

PR #3577, opened 2019-06-29
Refactor gcp KfApp to be more of a plugin and support click to deploy.

jlewi:
* Click to deploy and #2152 require the following

  * We want to be able to inject the Tokensource used by GCP client libraries
  * We want to get GCP plugin specific variables from the KfDef structure and not pass in
    some side channel of bytes passed to GetKfApp

* To support injection of token source

  * We add a GetPlugin method to the coordinator so that callers can get a specific plugin and interact with it

    * Click to deploy will use this to inject the token source for GCP

  * We also modify delay creation of GCP clients in the GCPPlugin to before they are needed
    e.g clients are now constructed in the apply and initGcpProject functions

* We also modify the GCP plugin to get secret information (e.g. BasicAuth and IAP OAuth info) from secrets stored
  in KfDef. Secrets can be stored as references (e.g. to environment variables) so that sensitive info won't
  get checked into source control.

* We add a backfill method to backfill KfDef options from the various command line flags.

  * This is a stepping stone to eventually getting rid of those flags and just relying the spec for everything.

* We refactor GCP plugin so that KfDef is a member variable rather than an embedded type.

  * This makes it easier to invoke KfDef methods on it.

  * We refactor GCP to use the save method defined on KFDef so that sensistive information is stripped out before
    writing to disk.

* This also partially addresses #3520 (Fix dynamic loading of platform plugins)

  * #3520 was broken because the signature of GetPlatform was changed to GetPlatform(*KfDef, [][]byte)
    This PR changes the signature back to GetPlatform(*KfDef); so all plugins should now take a single
    argument (*KfDef)

  * There is an open issue #3566 to futher refactoring to treat package managers and platforms consistently as
    plugins

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3577)


bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -1523,10 +1597,108 @@ func (gcp *Gcp) createSecrets() error {
...
+		pluginSpec.CreatePipelinePersistentStorage = proto.Bool(pluginSpec.GetCreatePipelinePersistentStorage())
+		log.Infof("CreatePipelinePersistentStorage not set defaulting to %v", *pluginSpec.CreatePipelinePersistentStorage)
+	}
+
+	if pluginSpec.Auth == nil {
+		pluginSpec.Auth = &Auth{}
+	}
+
+	if pluginSpec.Auth.BasicAuth == nil && pluginSpec.Auth.IAP == nil {
+		log.Warnf("Backfilling auth; this is deprecated; Auth should be explicitly set in Gcp plugin")
```
lluunn: when should auth be set in the future, init?

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -110,67 +110,63 @@ type dmOperationEntry struct {
...
-		err := json.Unmarshal(platformArgs, &gcpArgs)
-		if err != nil {
-			return nil, err
-		}
+// GetPluginSpec gets the plugin spec.
+func (gcp *Gcp) GetPluginSpec() *GcpPluginSpec {
+	// Not passing a pointer interface is a common cause of deserialization problems
+	pluginSpec := &GcpPluginSpec{}
+
+	// It should be fine to swallow the error
```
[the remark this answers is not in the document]
jlewi: On second thought we shouldn't swallow the error. I was thinking the only way we would fail is if the plugin wasn't defined in which case returning an empty spec was ok.

But we could also have a case where we have marshaling error; e.g. because of a YAML error or because the spec changed.

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -110,67 +110,63 @@ type dmOperationEntry struct {
...
-			}
+
+	if gcp.tokenSource == nil {
+		// Defensive Programming.
+		// If we try to create a DefaultTokenSource when an AccessToken is provided
+		// Something has gone wrong. So we guard against that.
+		// If accessToken is provided gcp.TokenSource should be set and we should use
+		// that.
+		if _, err := gcp.kfDef.GetSecret(GcpAccessTokenName); !kfdefs.IsSecretNotFound(err) {
+			return errors.WithStack(fmt.Errorf("Creating a default token source when AccessToken is provided is disallowed"))
```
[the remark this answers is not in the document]
jlewi: Fixed.

I don't think we need to check anything else. if !IsSecretNotFound then either

1. The secret exists 
2. There was another error (indicating the secret exists but couldn't be parsed).

In either case we want to trigger the warning.

bootstrap/pkg/kfapp/gcp/gcp.go
jlewi: I think it should be set in the YAML file.
e.g.
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.0.6.yaml

So GcpPluginSpec should be specified in that file

jlewi:
Thanks @lluunn PTAL

jlewi:
Deleting the test directory failed.

Logs for argo pod attached.

Main container never starts.

[kubeflow-presubmit-kfctl-go-iap-istio-3577-3203d8a-4848-1d65-2463764818.log.txt](https://github.com/kubeflow/kubeflow/files/3350310/kubeflow-presubmit-kfctl-go-iap-istio-3577-3203d8a-4848-1d65-2463764818.log.txt)

jlewi:
Pod event logs are attached. There is an ImageInspectError
[kubeflow-presubmit-kfctl-go-iap-istio-3577-3203d8a-4848-1d65-2463764818.event.log.txt](https://github.com/kubeflow/kubeflow/files/3350330/kubeflow-presubmit-kfctl-go-iap-istio-3577-3203d8a-4848-1d65-2463764818.event.log.txt)

jlewi:
/test all
see kubeflow/testing#413 for test issues

bootstrap/pkg/kfapp/gcp/gcp.go
lluunn: I see, thanks

Issue #3491, opened 2019-06-18
kfctl: kustomize component discovery logic is flawed


yanniszark:
`kfctl` uses a function called `mapDirs` to discover the components available and map each component's name to its path.
However, the logic it uses for discovery is flawed and can lead to perplexing errors.

https://github.com/kubeflow/kubeflow/blob/24c39e3cd66862b9433f23f7f9cc2937c70a39dd/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go#L522-L523

In the above check, to check if a folder is a component, we check if the folder is a package name and a component name.
This makes the discovery's results _depend_ on what the user has asked.
The component discovery should yield the same results no matter what components the user asks for.

eg.
For a structure like:
```
|-- admission-webhook
    |-- bootstrap
        |-- base
    |-- webhook
        |-- base
```

If the KfDef is wrong like the following, the discovery code will think that the `admission-webhook` folder is a component, because of the check above.
```yaml
packages:
  - admission-webhook
components:
  - admission-webhook
```

If instead the KfDef is like the following, the components `bootstrap` and `webhook` will be discovered correctly.
```yaml
packages:
  - admission-webhook
components:
  - bootstrap
  - webhook
```

jlewi:
Should we refactor the spec and get rid of components, packages, and componentParams?

This terminology no longer really makes sense now that we are using ksonnet.

Should we just refactor this into a list of applications

```
Applications:
   -  name: central-dashboard
      repoName:
      repoPath: 
      overlays:
         - overlay1 
         - overlay2
     parameters:
         - name: 
           value
``` 

For each application we can no specify its relative path within the repo so there's no more magic discovery going on.

jlewi:
@yanniszark Can you help me understand what is release blocking? Unfortunately I think @kkasravi is OOO this rest of the week.

Why do we have packages and components? These concepts don't really make sense to me in the context of kustomize?

Should we pick one maybe packages and treat the value as a path to a kusotmization.yaml file.

So instead of 

```
packages:
  - admission-webhook
components:
  - bootstrap
  - webhook
```

We just do

```
components:
  - admission-webhook/bootstrap
  - admission-webhook/webhook
```

And treat the value as the relative path to the kustomization file?

Alternatively we could add something like application_types 

```
Applications:
   -  name: central-dashboard
      repoName:
      repoPath: 
      overlays:
         - overlay1 
         - overlay2
      parameters:
          - name: 
             value
```

[...]

PR #3581, opened 2019-07-01
Add a list of Applications to KfDef.

jlewi:
* Application is intended as a replacemnt for the fields packages, components,
  componentParams.

* See #3491; discovery logic for kustomize is broken.

  * The new Application type allows us to specify in the YAML spec the
    repository and path within the repository of the kustomize package to use

  * This will allow us to address that issue because we will be able to
    stop relying on the discovery logic and instead just specify the packages
    in the app.yaml that we need.

* This PR just defines the type

* Forthcoming PRs

  * A PR to backfill Application based on packages, components & componentParams
    so that we don't break compatibility with existing specs.

  * Update kustomize.go to use Application

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3581)

### DIFF
--- a/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
+++ b/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
@@ -70,12 +70,32 @@ type KfDefSpec struct {
 	Repos              []Repo   `json:"repos,omitempty"`
 	Secrets            []Secret `json:"secrets,omitempty"`
 	Plugins            []Plugin `json:"plugins,omitempty"`
+
+	// Applications defines a list of applications to install
+	Applications []Application `json:"applications,omitempty"`
 }
 
 var DefaultRegistry = RegistryConfig{
 	Name: "kubeflow",
 	Repo: "https://github.com/kubeflow/kubeflow.git",
 	Path: "kubeflow",
+}
+
+// Application defines an application to install
+type Application struct {
+	Name            string           `json:"name,omitempty"`
+	KustomizeConfig *KustomizeConfig `json:"kustomizeConfig,omitempty"`
+}
+
+type KustomizeConfig struct {
+	RepoRef    *RepoRef           `json:"repoRef,omitempty"`
+	Overlays   []string           `json:"overlays,omitempty"`
+	Parameters []config.NameValue `json:"parameters,omitempty"`
+}
+
+type RepoRef struct {
+	Name string `json:"name,omitempty"`
+	Path string `json:"path,omitempty"`
 }
 
 // Plugin can be used to customize the generation and deployment of Kubeflow
@@ -650,3 +670,4 @@ func IsSecretNotFound(e error) bool {
 	_, ok := e.(*SecretNotFound)
 	return ok
 }
+


jlewi:
@yanniszark  for development if nothing else we need the ability to pull kubeflow/manifests from pull requests for testing as well as local filesystems. So I think we need Applications to refer to a repo in Repos that can be pointed at some arbitrary URL.

We also need the ability to specify the location of the package in the repo; this fixes the discovery issue you mentioned earlier.

So I think we want each Application to have a reference to a repo. We could hardcode the assumption that each application is coming from a single repo but that strikes me as brittle and I think that will cause problems.

I guess the question is where we should put the path information for the package. I threw it in RepoRef.
But I guess we could also move the path information into KustomizeConfig.

e.g.

```
RepoRef    *RepoRef           `json:"repoRef,omitempty"`
Overlays   []string           `json:"overlays,omitempty"`
Parameters []config.NameValue `json:"parameters,omitempty"`
Path           string
```

I don't have a strong preference. I lean towards leaving it as is for the sake of getting something out in 0.6.

We will have to do a v1beta1 of kfDef in 0.7 and that will necessitate significant changes and we can fold any changes into that.

yanniszark:
> So I think we want each Application to have a reference to a repo. We could hardcode the assumption that each application is coming from a single repo but that strikes me as brittle and I think that will cause problems.

When you say a single repo, you don't mean just kubeflow/manifests, but any one repo, correct?
Because we can also define local, external repos in the spec.Repos field.
I don't think it's unreasonable for a user to have all the packages in one place, but I'm ok either way.

In the case we go with multiple repos, can we ensure a sensible default for the RepoRef, which is filled in case it's not specified ? (eg ManifestsRepoName).

> I guess the question is where we should put the path information for the package. I threw it in RepoRef.
> But I guess we could also move the path information into KustomizeConfig.

The truth is it confused me a bit seeing it inside RepoRef.
To me, it makes more sense for RepoRef to be a string, that is the name of the repo.
I'm okay either way though.

jlewi:
> The truth is it confused me a bit seeing it inside RepoRef.
To me, it makes more sense for RepoRef to be a string, that is the name of the repo.
I'm okay either way though.

I think we should keep RepoRef as a struct. I think this makes us more future proof because if we decide
a RepoRef should contain additional fields we can just add them.

> I don't think it's unreasonable for a user to have all the packages in one place, but I'm ok either way.

I think assuming all packages are in one place is limiting. For example, suppose there's a bug in one application manifest e.g. Katib and so one wants to pull that application from a different branch/repo then
all the other applications.

If each Application can specify a different repo then this because trivially to deal with. Add a new repo to Repos and update the ref in the application.

If we assume all applications are in the same repo then the user has to create a mirror pulling in the version of each package they want.

> In the case we go with multiple repos, can we ensure a sensible default for the RepoRef, which is filled in case it's not specified ? (eg ManifestsRepoName).

That's a good idea.  Lets do that in a follow on PR.  By design this PR is only containing the struct fields.

Default setting should happen in coordinator.go.

Specifically I think either in  NewKfApp
https://github.com/kubeflow/kubeflow/blob/b60cb3f3e4722678bc4d2871d499e49abc991818/bootstrap/pkg/kfapp/coordinator/coordinator.go#L408

Or CreateKfFromDefinitions
https://github.com/kubeflow/kubeflow/blob/b60cb3f3e4722678bc4d2871d499e49abc991818/bootstrap/pkg/kfapp/coordinator/coordinator.go#L224

Is where any defaulting should happen.

In other words any defaults should be explicitly set when the KFDef is first created and turned into explicit values set in the KfDef so that subsequent steps can rely on the values being set.

We should also add IsValid routines to validate the resources.

jlewi:
@yanniszark Any additional comments?

yanniszark:
@jlewi thanks for addressing all my comments :)
The current pr looks ok to me.

/lgtm

PR #3582, opened 2019-07-01
Fix SyncCache and define errors to indicate Secret not found.

jlewi:
* SyncCache needs to check the downloaded directory and set ReposCache.LocalPath

  * Previously we were relying on Root being set in KfDef.Spec.Repo
    but that won't work with pull requests.

  * With pull requests e.g. "pull/<PR number>" the directory unpacks
    to <repository>-<branch>-<commit>

  * We don't want users to have to specify the commit in KfDef so we have
    to have SyncCache look at the unpacked directory and get it.

  * SyncCache looks at the URI of the repo to decide whether to apply that
    logic because we also want to support local directories which end up
    getting sym-linked.

* Define a specific error and helper function to detect when a Secret is not
  found.

* Fix a bug in IsPluginNotFound - we should handle the case when err = nil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3582)


jlewi:
presubmit appears to be stuck because merge-base didn't succeed so its picking up changes to the TF dockerfiles and running that workflow.

PR #3588, opened 2019-07-01
kustomize should support injecting the K8s restconfig.

jlewi:
* To do this we define a setter interface that will allow the K8s client to
  be injected.

* We also move the initialization of the K8s clients out of GetPlatform and
  into a function initK8sClients that isn't called until we actually need
  the client (i.e. in apply)

  * This gives callers a chance to inject the RESTConfig

* Some minor cleanup of kustomize

  * Delete some unused code; like kustomize.Application

  * Make KfDef a member variable rather than embedding it; this is cleaner.

    * Not sure what the original impetus of embedding it was.
    * Making it a member variable makes it more obvious that we are passing
      around a pointer to the different KfApps that are all modifying the same
      KfDef; this is important because we want to use the KfDef.Spec/Status
      to pass information around.

* Replace kustomize.go's customize WriteConfigFile method with one in
  application_types

  * This ensures secrets are properly stripped out.

  * It also ensures consistency with respect to the filename used.

* Related to #2152 and GCP click to deploy; we need to inject the K8s client.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3588)

### DIFF
--- a/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
+++ b/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
@@ -27,6 +27,7 @@ import (
 	kfapisv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis"
 	kftypesv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps"
 	kfdefsv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1"
+	"github.com/kubeflow/kubeflow/bootstrap/v2/pkg/utils"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


jlewi:
/assign @kunmingg 
/assign @yanniszark 
/cc @kkasravi 

E2E tests are running; may need to make some minor edits if there are test failures. unittests are passing locally though.

jlewi:
The GCPPlugin isn't setting the RestConfig; injection is happening outside of coordinator.

Here's a preview of how the click to deploy server is doing it.
https://github.com/kubeflow/kubeflow/blob/41b388166a5abddbf337e2ded48e5a543f2d8493/bootstrap/cmd/bootstrap/app/kfctlServer.go#L191

So code that wants to do injection does the following

1. Call KfDef.GetPlugin()
1. Assert that the plugin implements the desired interface
1. Invoke the desired interface on that plugin.

A plugin could do that if needed since it has a reference to KfDef. Right now that's unnecessary
because  in CLI mode, kustomize.go just falls back to instantiating a default K8s Client based on KubeConfig.

The setter interface is defined in kustomize because this particular interface is intended to be specific to this plugin.

We could define generic interfaces that multiple plugins might implement.

But we might also define Getter/Setter interfaces that are plugin specific. 
For example, GCP defines a setter
https://github.com/kubeflow/kubeflow/blob/b9e2f0d11ff893edfbb4585cf40cea8d5379e08e/bootstrap/pkg/kfapp/gcp/gcp.go#L98

That allows injecting the token source. That only makes sense for the GCP plugin.

jlewi:
@yanniszark Could you take another look at this please?

jlewi:
@yanniszark Rebased this one as well.

yanniszark:
Thanks for cleaning it up a bit and using the new functions from application_types.
/lgtm

PR #3594, opened 2019-07-02
Add helper methods to set application parameters and define conditions.

jlewi:
* Add helper methods to start setting application parameters.

  * The helper methods take care of providing backwards compatibility.
  * Newer KFDefs will use KfDef.Spec.Application to store applications
    and their parameters.
  * Older specs use KfDef.Spec.CompParams.

* The new methods will help us transition because code can be updated
  to use the new methods.

* Also start defining some constant corresponding to various conditions.
  kfctl will use these to report status.

Related to #3491

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3594)


jlewi:
Checkout failed. Logs weren't available in Argo.
Rerun the test.
/test all

jlewi:
Test failure is likely: kubeflow/testing#413

jlewi:
Test ran successfully but delete failed.

Will rebase and retry.

bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
```diff
@@ -247,6 +247,31 @@ type RepoCache struct {
 
 type KfDefConditionType string
 
+const (
+	// JobCreated means the job has been accepted by the system,
```
lluunn: Comment mismatches with definition; same below

bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
```diff
@@ -655,3 +681,101 @@ func IsSecretNotFound(e error) bool {
...
+	for _, p := range parameters {
+		if p.Name == paramName {
+			return p.Value, true
+		}
+	}
+
+	return "", false
+}
+
+func setParameter(parameters []config.NameValue, paramName string, value string) []config.NameValue {
```
lluunn: should we do it in place (modify the input directly)? So that we don't have to use the return value
 a = setParameter(a, xx)

yanniszark: @lluunn because of how slices work in Go, it is necessary to return the resulting slice.
[Excerpt from golang.org](https://blog.golang.org/slices):

> It's important to understand that even though a slice contains a pointer, it is itself a value. Under the covers, it is a struct value holding a pointer and a length. It is not a pointer to a struct. 

In other words, if you have appended the slice in a function and then don't return it, the slice of the calling function may contain obsolete data.
[relevant stackoverflow issue](https://stackoverflow.com/questions/20195296/golang-append-an-item-to-a-slice)

bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
```diff
@@ -655,3 +681,101 @@ func IsSecretNotFound(e error) bool {
...
+		pIndex = len(parameters) - 1
+	}
+
+	parameters[pIndex].Name = paramName
+	parameters[pIndex].Value = value
+
+	return parameters
+}
+
+type AppNotFound struct {
```
yanniszark: Should we declare a new type or reuse the KFError struct with a new error code ?

bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
jlewi: Done.

bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
jlewi: @yanniszark is correct. I think  @lluunn might be asking why not take (*KfDef).

This method is specifically intended to be a helper method called from SetApplicationParameter.
So we refactored out the common bits.  I didn't want to make it a func (d*KfDef) function because that makes it less clear what the function is doing; i.e. the function could be reading modifying any (d *kfDef). The current function signature is more narrowly scoped which I think is better.

Also its beneficial in that the function def can have fewer side effects (i.e. modifying *KfDef).

bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go
jlewi: I think it makes more sense to define a new type(s) here; I think that's more consistent with Go style. These errors are specific to this package.

I think if we want to define common error codes then we should consider making KFError an interface and then these errors can implement that interface.

jlewi:
@lluunn @yanniszark sorry for the long delay. Would you mind taking another look? I'd love to get this PR submitted so I can send out the next batch of PRs which is updating kustomize.go and gcp.go to use this code.

jlewi:
Thanks @lluunn 
/approve

Issue #3593, opened 2019-07-02
gcp.go should assert error is of type kfError before accessing its fields


jlewi:
We have lots of code like this

https://github.com/kubeflow/kubeflow/blob/bb4d1d9832b826e491e701bbbe5d52310d96fdda/bootstrap/pkg/kfapp/gcp/gcp.go#L832

```
return  &kfapis.KfError{
			Code: secretsErr.(*kfapis.KfError).Code,
			Message: fmt.Sprintf("gcp apply could not create secrets Error %v",
				secretsErr.(*kfapis.KfError).Message),
		}
```

This is unsafe because this will segfault if the error fails the assertion. 

We should be more defensive.

In a forthcoming PR I will add a method like

```
func NewKfErrorWithMessage(e error, msg string) error {
	kErr, ok := e.(*KfError)

	if !ok {
		log.Infof("Error is not a KfError; %v", e)

		return &KfError{
			Code:    int(UNKNOWN),
			Message: msg + "; " +  e.Error(),
		}
	}
	return &KfError{
		Code:    kErr.Code,
		Message: msg + "; " + kErr.Message,
	}
}
```

which can be used to replace the existing code.

Filing this issue to track actually updating the code.

PR #3595, opened 2019-07-02
Some error and logging utilities.

jlewi:
* Define a PrettyPrint function based on serializing to json.

* Define a helper method for KfErrors to combine an existing error with
  a message.

  * This is related to #3593; there are lots of places in GCP.go where
    we assume an error is of type KfError without asserting whether that's
    true

  * This helper method will help us fix those locations.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3595)

### DIFF
--- a/bootstrap/v2/pkg/apis/kferrors.go
+++ b/bootstrap/v2/pkg/apis/kferrors.go
@@ -16,6 +16,7 @@ package apis
 
 import (
 	"fmt"
+	log "github.com/sirupsen/logrus"
 )
 
 type StatusCode int
@@ -38,4 +39,26 @@ type KfError struct {
 func (e *KfError) Error() string {
 	return fmt.Sprintf(" (kubeflow.error): Code %d with message: %v",
 		e.Code, e.Message)
+}
+
+// NewKfErrorWithMessage will propogate the error with the given message.
+//
+// TODO(jlewi): Not sure this is the best way to propogate the error messages and turn them
+// into KfErrors. There was a lot of code that was doing this but not asserting that the error
+// was a KfError which was causing segmentation faults so I wrote this helper method.
+func NewKfErrorWithMessage(e error, msg string) error {
+	kErr, ok := e.(*KfError)
+
+	if !ok {
+		log.Infof("Error is not a KfError; %v", e)
+
+		return &KfError{
+			Code:    int(UNKNOWN),
+			Message: msg + "; " + e.Error(),
+		}
+	}
+	return &KfError{
+		Code:    kErr.Code,
+		Message: msg + "; " + kErr.Message,
+	}
 }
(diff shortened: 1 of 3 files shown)


bootstrap/v2/pkg/utils/logging.go
```diff
@@ -0,0 +1,20 @@
+package utils
+
+import (
+	"encoding/json"
+	"fmt"
+	log "github.com/sirupsen/logrus"
+)
+
+// PrettyPrint returns a pretty format output of any value.
+func PrettyPrint(value interface{}) string {
```
yanniszark: Should we name this `PrettyPrintJSON` since it only works for json input?
In addition, you might find this library interesting: https://github.com/davecgh/go-spew

jlewi: It should work for pretty much any struct. Can you think of a struct it wouldn't work for?
In which case I'd vote for the shorter name.

jlewi:
/test all 
see kubeflow/testing#413 for test issues

bootstrap/v2/pkg/utils/logging.go
yanniszark: You're right, it seems like it will work for any type.
A side-effect is that json tags will affect the output (eg some fields may be omitted or named differently).
Not sure if we want this or not.
Since the name is PrettyPrint I assume it's ok.
Otherwise, if we want this for debugging, maybe we should use the `go-spew` library.

jlewi: I think its ok. Its only for pretty printing the output not debugging.

jlewi:
/test all

@yanniszark Can you LGTM please if you are happy with this.

bootstrap/v2/pkg/apis/kferrors.go
```diff
@@ -39,3 +40,25 @@ func (e *KfError) Error() string {
...
+// was a KfError which was causing segmentation faults so I wrote this helper method.
+func NewKfErrorWithMessage(e error, msg string) error {
+	kErr, ok := e.(*KfError)
+
+	if !ok {
+		log.Infof("Error is not a KfError; %v", e)
+
+		return &KfError{
+			Code:    int(UNKNOWN),
+			Message: msg + "; " + e.Error(),
```
yanniszark: Final thought: Do you think that maybe we should use errors.Wrap here?
Or we don't want to include a stacktrace ?

yanniszark:
I added a final comment, but this PR is good either way.
/lgtm

bootstrap/v2/pkg/apis/kferrors.go
jlewi: I think we need to rethink how we handle errors and be more consistent and thorough and ideally include stacktrace info.

Including stacktraces is definitely a good idea.

This is mostly a quick fix to deal with the fact that we have a bunch of unsafe code right now.

PR #3598, opened 2019-07-02
kfctl: existing_arrikto: add self-signed https certificates

yanniszark:
This PR adds self-signed HTTPS certificates to secure the Istio gateway.
Currently, these certificates are for the LoadBalancer IP.
In the near future, it will be possible to use a DNS domain and trusted certificates.

Some remarks:
* The AuthService and Istio Gateway Deployment now live in the same namespace (istio-system) in order to share the cert secret.

/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3598)

### DIFF
--- a/bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
+++ b/bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
@@ -303,10 +303,37 @@ func getEndpoints(kubeclient client.Client) (string, string, error) {
 }
 
 func createSelfSignedCerts(kubeclient client.Client, addr string) error {
+
+	cert, key, err := generateCert(addr)
+	if err != nil {
+		return errors.WithStack(err)
+	}
+
+	// Create secret from them
+	secret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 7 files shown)


bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -290,6 +302,72 @@ func getEndpoints(kubeclient client.Client) (string, string, error) {
 	return kfEndpoint, oidcEndpoint, nil
 }
 
+func createSelfSignedCerts(kubeclient client.Client, addr string) error {
```
jlewi: Can we write a unittest? Maybe split up the code into a function to define the K8s v1.Secret object and a function which actaully calls it.

Then its easy to write a unittest to verify the secret is defined as expected.

deployment/existing/auth_oidc/envoy-filter.yaml
```diff
@@ -11,8 +11,8 @@ spec:
   - filterConfig:
```
jlewi: Do we have an issue tracking making these kustomize packages?

PR #3599, opened 2019-07-02
GCP Click-to-Deploy UI changes for v0.6.x

### LINKED ISSUE #3560: [GCP] Click to Deploy UI update to support 0.6
We need to make some updates to the click to deploy UI and frontend to support 0.6.

https://github.com/kubeflow/kubeflow/tree/master/components/gcp-click-to-deploy

Here are the changes needed

1. The endpoint for createDeployment will be changed to `/kfctl/apps/v1alpha2/create`
1. The payload will be a [KfDef](https://github.com/kubeflow/kubeflow/blob/9bb2c0176506950415cee95c602d66dd88cc3f92/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go#L24)
1. The KfDef should be constructed as follows
    * Fetch a YAML file from an http link and load the KfDef from it
    * override fields in the YAML spec based on the fields in the UI
1. As with the current UI; i'd like a drop down that is easy to populate with human readable descriptions e.g "0.6 with IAP" and behind the scenes are mapped to specific URI

    * I'd also like the ability to enter a specific URI link (possibly via query arg)

1. We will need to preserve existing code paths to continue to support v0.5.
    * It would be nice if the drop down box for version listed options for 0.6 and 0.5 and based
      on version we invoked correct code path.

/cc @avdaredevil @prodonjs

prodonjs:
closes #3560 

/area frontend
/platform gcp

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3599)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -18,8 +18,8 @@ import {encryptPassword, flattenDeploymentOperationError, getFileContentsFromGit
 
 /** Relative paths from the root of the repository. */
 enum ConfigPath {
-    V05 = 'components/gcp-click-to-deploy/app-config.yaml',
-    V06 = 'bootstrap/config/kfctl_gcp_iap.0.6.yaml'
+    V05 = 'v0.5-branch/components/gcp-click-to-deploy/app-config.yaml',
+    V06 = 'v0.6-branch/bootstrap/config/kfctl_gcp_iap.0.6.yaml'
 }
 
 /** Versions available for deployment. */
@@ -201,7 +201,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
             zone: 'us-central1-a',
         };
 
-        this.versionChanged = this.versionChanged.bind(this);
+        this._versionChanged = this._versionChanged.bind(this);
     }
 
     public async componentDidMount() {
@@ -226,7 +226,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
             });
         }
         // By default, load the v0.6.0 specification
-        this.loadConfigFile(ConfigPath.V06);
+        this._loadConfigFile(ConfigPath.V06);
     }
 
     public render() {
@@ -309,7 +309,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
 
                 <div style={styles.row}>
                     <TextField select={true} label="Kubeflow version:" required={true} style={styles.input} variant="filled"
-                        value={this.state.kfversion} onChange={this.versionChanged}>
+                        value={this.state.kfversion} onChange={this._versionChanged}>
                         {this.state.kfversionList.map((version, i) => (
                             <MenuItem key={i} value={version}>{version}</MenuItem>
                         ))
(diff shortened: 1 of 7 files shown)


prodonjs:
/uncc @abhi-g 
/assign @jlewi 

I think I'm pretty close to being able to test how this behaves with an 0.6.0 compatible endpoint. Is #3652 a dependency for the front-end changes?

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -14,868 +14,982 @@ import * as React from 'react';
...
-// When I tried using fetch API to do that I ran into errors that I interpreted as chrome blocking
-// downloads from other domains from the one the app is being served at.
-// So for security reasons it might be better to just bundle the configs.
-// When we build a docker image as part of our release process we can just
-// copy in the latest configs.
-import appConfigPath from './user_config/app-config.yaml';
+/** Relative paths from the root of the repository. */
+enum ConfigPath {
+    V05 = 'components/gcp-click-to-deploy/app-config.yaml',
+    V06 = 'bootstrap/config/kfctl_gcp_iap.0.6.yaml'
```
kunmingg: Shall we just load different config template for iap vs basic-auth cases?

components/gcp-click-to-deploy/src/Utils.ts
```diff
@@ -1,5 +1,8 @@
 import * as bcrypt from 'bcryptjs';
 
+const GITHUB_BASE_URL =
+    'https://raw.githubusercontent.com/kubeflow/kubeflow/master/';
+
```
kunmingg: Config file on kubeflow master (or other branch) might not compatible with kfctl backend.
Shall we bake v0.5 & v0.6 config files into image during build?

components/gcp-click-to-deploy/src/DeployForm.tsx
prodonjs: Good point. I think it might be easier to remove the IAP-specific sections from this config if the user chooses not to use IAP rather than have to manually add them if basing the config on the non-IAP file.

components/gcp-click-to-deploy/src/Utils.ts
prodonjs: Currently, that's how it's done for the 0.5.0 compatible deployment but Jeremy had a comment to try to fetch it from GitHub. Pulling from master is probably problematic but we can easily point to a branch for each version to ensure that we get a consistent file that should be compatible with the version in question.

prodonjs:
> Thanks!
> Is there any framework that we can add unit tests against request body of each version?
> We can follow in separate PRs.

Yes, both the Pipelines and Metadata UIs are written in the same front-end framework (React) and have solid unit-test coverage using the preferred framework for testing with React (Jest). Adding tests would likely require some refactoring so it's not something I considered given the timeline for getting this ready but it should definitely be added in the future.

prodonjs:
@jlewi @kunmingg I've made some additional changes to make sure the request format to the kfctl server matches the expected format. I can test and get the request to be successfully received by a locally running instance of the bootstrapper server, but I don't think I am configuring it properly to actually have it start the deployment process.

Is there a deployed instance of the service that I can proxy requests to? I'd like to be able to manually test all of the v0.6.0 permutations to see what other changes need to be made. Supporting v0.5.0 on the same UI deployment may be a bit of a challenge unless we have a CORS-enabled endpoint running on an alternative server that lets us use the old request path, or, ideally, if the new service can support v0.5.0 deployments simply by setting the version argument.

kunmingg:
@gabrielwen 
Can you share the test cluster with new backend with @prodonjs ?
Thanks!

kunmingg:
@prodonjs 
Is the new UI image available for test?

prodonjs:
I don't think I pushed one this morning. I can do it either a little bit
later this evening or you can checkout the PR and build it from the Makefile

kunmingg:
Great! will build from this PR.

kunmingg:
Can we update uri here based on user's choice?
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.yaml#L12
For example for `v0.6.1` uri will be
`https://github.com/kubeflow/kubeflow/archive/v0.6.1.tar.gz`

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -14,868 +14,1120 @@ import * as React from 'react';
...
+            Namespace: 'kubeflow',
+            Project: project,
+            ProjectNumber: projectNumber,
+            SAClientId: saClientId,
+            Token: saToken,
+            UseKfctl: this.state.kfctlLib,
+            Zone: this.state.zone,
+        };
+        if (this.state.ingress === IngressType.Iap) {
+
```
kunmingg: Ideally we should pass pasword hash here instead of plain password.
And password type is SecretRef. For example: https://github.com/kubeflow/kubeflow/pull/3830/files#diff-a7d91e7a1b9afdcbb5b17f72957d5863R122

prodonjs: Not a problem. I can definitely make the change. To this point, I haven't been able to get a request to work regardless of the auth method but I'll get this in the next change.

prodonjs:
> Can we update uri here based on user's choice?
> https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.yaml#L12
> For example for `v0.6.1` uri will be
> `https://github.com/kubeflow/kubeflow/archive/v0.6.1.tar.gz`

Yes, I guess the only question is how/where we maintain the map of "version" to "archive". Right now we show v0.6.0 in the drop-down and we use that to determine the which base YAML file to load and replace with the user's options. I don't think we want to offer all of the patch-level versions to the user so maybe the best option is just to make sure https://github.com/kubeflow/kubeflow/blob/v0.6-branch/bootstrap/config/kfctl_gcp_iap.0.6.yaml gets updated with the right paths since that's the file we load.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -1076,28 +1091,31 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
...
-                createPipelinePersistentStorage: permanentStorage,
-                username: saClientId,
+                createPipelinePersistentStorage: permanentStorage
             }
         };
         if (ingress === IngressType.BasicAuth) {
             gcpPlugin.spec.auth = {
                 basicAuth: {
-                    password: this.state.password,
+                    password: btoa(encryptPassword(this.state.password)),
```
kunmingg: password field actually has type
```
&kfdefsv2.SecretRef
```
@jlewi 
What's your suggestion about passing password hash instead of raw password?

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -14,31 +14,33 @@ import * as React from 'react';
...
+import {
+    encryptPassword, flattenDeploymentOperationError,
+    getFileContentsFromGitHub, log, wait
+} from './Utils';
 
 /** Relative paths from the root of the repository. */
 enum ConfigPath {
     V05 = 'v0.5-branch/components/gcp-click-to-deploy/app-config.yaml',
-    V06 = 'v0.6-branch/bootstrap/config/kfctl_gcp_iap.0.6.yaml'
+    V06 = 'master/bootstrap/config/kfctl_gcp_iap.yaml'
```
kunmingg: `v0.6-branch` should be more stable compared to `master`
`master` branch could be broken at any time.

prodonjs: Yes I agree. But [v0.6-branch/bootstrap/config/kfctl_gcp_iap.0.6.yaml](https://github.com/kubeflow/kubeflow/blob/v0.6-branch/bootstrap/config/kfctl_gcp_iap.0.6.yaml)  doesn't have the `applications` section set. [master/bootstrap/config/kfctl_gcp_iap.yaml](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.yaml) has `applications` but not `components`, `componentParams`, or `packages`.

If possible, we should check a base config file into the v0.6-branch that has the main structure that needs to be in the request body so that the UI doesn't have to build any new elements but simply replace parameter values. Alternatively, we could revert to the mechanism used in the current deployment UI where we provide the config file at deployment time, but pulling from GitHub definitely reduces the deployment complexity.

components/gcp-click-to-deploy/src/DeployForm.tsx
prodonjs: In the latest commit, I switched this to use a SecretRef.

prodonjs:
@kunmingg @jlewi We seem to be getting close. I'm able to get deployments started for both v0.5 and v0.6 requests. The only thing that doesn't seem to make sense is that after the DM jobs finish and the K8s clusters appear in my project, none of the KF workloads are actually created in the kubeflow namespace. All I see are pods in the kube-system namespace.

Interestingly, I'm seeing the same behavior when I generate an app.yaml with kfctl locally and submit it using the kfctlClient.

kunmingg:
/lgtm
Thanks! 
There might be some more API changes coming on backend side but we can adjust in follow up PRs.

PR #3601, opened 2019-07-02
keep manifest in kfctl directory as kustomize pkg

### LINKED ISSUE #3558: kustomize packages for installed apps should be in ${KFAPP}/kustomize
My expectation is that the kustomize packages for installed applications should be stored in
${KFAPP}/kustomize

* This way if user checks it into source control their manifests are actually checked in
* Furthermore, for complex manipulations; user can just edit the kustomize manifests
* I think ${KFAPP}/kustomize should only contain manifests for the applications user installed
  * It shouldn't just be a dump of kubeflow/manifests
* So the behavior of kfctl should be to cache kubeflow/manifests somewhere and then copy applications as they are added to ${KFAPP}/kustomize

* We should write the code in a way that its easy to write a unittest that this is happening as expected

lluunn:
Fix #3558 

Generate now only copies the pkg dir to kustomize/
Apply will call GenerateKustomizationFile to get the manifest.

```
kf0702/kustomize$ ls
api-service         cert-manager-crds   jupyter-web-app     minio               profiles            tf-job-operator
application         cloud-endpoints     katib-controller    mysql               pytorch-job-crds    webhook
application-crds    gpu-driver          katib-db            notebook-controller pytorch-operator
argo                iap-ingress         katib-manager       persistent-agent    scheduledworkflow
bootstrap           istio               katib-ui            pipelines-runner    spartakus
centraldashboard    istio-crds          metacontroller      pipelines-ui        suggestion
cert-manager        istio-install       metrics-collector   pipelines-viewer    tensorboard

kf0702/kustomize$ ls pytorch-operator/
base               kustomization.yaml overlays
```

/cc @jlewi @kkasravi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3601)

### DIFF
--- a/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
+++ b/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
@@ -27,6 +27,7 @@ import (
 	kfapisv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis"
 	kftypesv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps"
 	kfdefsv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1"
+	"github.com/otiai10/copy"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
 	"io/ioutil"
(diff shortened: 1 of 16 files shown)


bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -205,28 +206,46 @@ func (kustomize *kustomize) Apply(resources kftypes.ResourceEnum) error {
 
 	kustomizeDir := path.Join(kustomize.Spec.AppDir, outputDir)
 	for _, compName := range kustomize.Spec.Components {
-		kustomizeFile := filepath.Join(kustomizeDir, compName+".yaml")
-		if _, err := os.Stat(kustomizeFile); err != nil {
+		resMap, err := GenerateKustomizationFile(&kustomize.KfDef, kustomizeDir, compName,
```
jlewi: Is GenerateKustomizationFile creating the kustomizaiton.YAML file? Or is it creating the kustomization.yaml file?

It looks like its creating the kustomizationFile because its taking ComponentParams.

Should that be done in generate?

The idea would be that if a user runs
kfctl generate
edits the kustomize package
kfctl apply

Then any user supplied edits should be picked up.

But right now it looks like those changes would be overwritten if during apply we call GenerateKustomizationFile.

Not sure that's actually possible right now with the way GenerateKustomizationFile works; it looks like its coupling the writing of the kustomization file with evaluating overlays and params.

What if we just ran GenerateKustomizationFile during generate like we do now but the Copying of the
kustomize package so that kustomize/  would contain the list of kustomize packages.

lluunn: makes sense, thanks.

I separated the writing kustomize.yaml and evaluation part in GenerateKustomizationFile

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -28,6 +28,7 @@ import (
 	kftypesv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps"
 	kfdefsv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1"
 	"github.com/kubeflow/kubeflow/bootstrap/v2/pkg/utils"
+	"github.com/otiai10/copy"
```
jlewi: What made you pick this package? Are we using it in any other place?

lluunn: I use this for copy directory; I just google and found it.

jlewi: Yeah I ended up running into the same problem and my Googling turned up the same.

lluunn:
Added unit test, PTAL

PR #3607, opened 2019-07-04
parameterise auth redirects and hmac

ryandawsonuk:
HMAC secret to be added at install time. Otherwise this is generated inside the auth service pod. When generated inside the pod then if the pod goes down all the clients have to reconnect.

Parameterisation of redirectUirs is to make it possible to allow other points of entry but by default there will only be the existing one.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3607)


bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -470,3 +480,24 @@ func genRandomString(length int) string {
...
+
+	if len(envContent) < reqLen {
+		log.Println("WARNING: HMAC secret not provided or secret too short. Generating a random one from nonce characters.")
+		return []byte(createNonce(reqLen))
+	}
+
+	return []byte(envContent)
+}
+
+func createNonce(length int) string {
```
SachinVarghese: You can use the genRandomString function here and the createNonce is not really required.

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -154,13 +155,23 @@ func (existing *Existing) Apply(resources kftypes.ResourceEnum) error {
...
 	}{
 		KubeflowEndpoint:        kfEndpoint,
 		OIDCEndpoint:            oidcEndpoint,
-		AuthServiceClientSecret: genRandomString(32),
+		AuthServiceClientSecret: base64.StdEncoding.EncodeToString([]byte(genRandomString(32))),
+		AuthServiceHmacSecret:   base64.StdEncoding.EncodeToString([]byte(initialiseHMACSecretFromEnvOrGen("JWT_HMAC_SECRET", 64))),
+		OIDCRedirectUris:        getListVariableFromEnv("OIDC_REDIRECTURIS"),
 		KubeflowUser:            kubeflowUser,
-	}
+	} //hmac secret logic following https://github.com/yanniszark/ambassador-auth-oidc/blob/feature-kubeflow/login.go#L311
```
yanniszark: I don't see any of that logic here.
Maybe this is an old comment?

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -154,13 +155,23 @@ func (existing *Existing) Apply(resources kftypes.ResourceEnum) error {
...
 		AuthServiceClientSecret string
+		AuthServiceHmacSecret   string
+		OIDCRedirectUris        []string
 		KubeflowUser            *kfUser
 	}{
 		KubeflowEndpoint:        kfEndpoint,
 		OIDCEndpoint:            oidcEndpoint,
-		AuthServiceClientSecret: genRandomString(32),
+		AuthServiceClientSecret: base64.StdEncoding.EncodeToString([]byte(genRandomString(32))),
+		AuthServiceHmacSecret:   base64.StdEncoding.EncodeToString([]byte(initialiseHMACSecretFromEnvOrGen("JWT_HMAC_SECRET", 64))),
```
yanniszark: I don't believe it's necessary to copy the whole logic.
Let's just call genRandomString immediately.

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -174,6 +185,13 @@ func (existing *Existing) Apply(resources kftypes.ResourceEnum) error {
 		return internalError(errors.WithStack(err))
 	}
 
+	//for dex the AuthServiceClientSecret shouldn't be base64 encoded
+	authServiceClientSecretDecoded, err := base64.StdEncoding.DecodeString(data.AuthServiceClientSecret)
```
yanniszark: It seems that we need this because Dex doesn't support setting the client id with secrets.
https://github.com/dexidp/dex/issues/1099

Because of this, we have to decode the value and set it directly in the Dex ConfigMap.
Would be nice if we could solve this upstream or via an initContainer.
@ryandawsonuk can you add a TODO comment here explaining why we do this ?

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -470,3 +493,14 @@ func genRandomString(length int) string {
 	}
 	return string(b)
 }
+
+func initialiseHMACSecretFromEnvOrGen(secEnv string, reqLen int) []byte {
```
yanniszark: I don't think we need this. Let's just call genRandomString directly.

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -154,13 +155,23 @@ func (existing *Existing) Apply(resources kftypes.ResourceEnum) error {
 		KubeflowEndpoint        string
 		OIDCEndpoint            string
 		AuthServiceClientSecret string
+		AuthServiceHmacSecret   string
+		OIDCRedirectUris        []string
```
yanniszark: Maybe we should call this OIDCExtraRedirectURIs since the default redirect uri is included no matter what.

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -174,6 +184,14 @@ func (existing *Existing) Apply(resources kftypes.ResourceEnum) error {
 		return internalError(errors.WithStack(err))
 	}
 
+	//for dex the AuthServiceClientSecret shouldn't be base64 encoded as in configMap
```
yanniszark: nit: It'd be great to keep the comment formatting homogeneous in the code.

//for -> // For
First letter capital and indented by 1 space.

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -154,14 +155,23 @@ func (existing *Existing) Apply(resources kftypes.ResourceEnum) error {
...
 		KubeflowEndpoint:        kfEndpoint,
 		OIDCEndpoint:            oidcEndpoint,
-		AuthServiceClientSecret: genRandomString(32),
+		AuthServiceClientSecret: base64.StdEncoding.EncodeToString([]byte(genRandomString(32))),
+		AuthServiceHmacSecret:   base64.StdEncoding.EncodeToString([]byte(genRandomString(64))),
+		OIDCExtraRedirectUris:   getListVariableFromEnv("OIDC_EXTRA_REDIRECTURIS"),
 		KubeflowUser:            kubeflowUser,
 	}
 
+	//HmacSecret is generated on install as needs to be the same between restarts of authService
```
yanniszark: Same as the other comment.

bootstrap/v2/pkg/kfapp/existing_arrikto/existing.go
```diff
@@ -154,14 +155,23 @@ func (existing *Existing) Apply(resources kftypes.ResourceEnum) error {
...
+	//OIDCExtraRedirectUris provides option to add entrypoints in addition to default gateway
+	log.Debugf("AuthServiceHmacSecret %s", data.AuthServiceHmacSecret)
```
yanniszark: It feels wrong to log a secret value. I think we shouldn't do this at all.

yanniszark:
/assign @kunmingg 

cc @jlewi @lluunn ptal whenever you can

deployment/existing/auth_oidc/dex.tmpl
```diff
@@ -80,8 +80,13 @@ data:
       # After authenticating and giving consent, dex will redirect to
       # this url for the specific client.
       - {{.KubeflowEndpoint}}/login/oidc
+      {{if .OIDCExtraRedirectUris}}
```
krishnadurai: Instead of having `{{.KubeflowEndpoint}}/login/oidc` and `OIDCExtraRedirectUris` can this be made as `OIDCRedirectUris` instead?
cc: @yanniszark

deployment/existing/auth_oidc/envoy-filter.yaml
```diff
@@ -15,10 +15,11 @@ spec:
           cluster: outbound|8080||authservice.istio-system.svc.cluster.local
           failureModeAllow: false
           timeout: 10s
-        authorizationRequest: 
+        authorizationRequest:
           allowedHeaders:
             patterns:
             - exact: "cookie"
+            - exact: "X-Auth-Token"
```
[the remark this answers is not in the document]
ryandawsonuk: The specific name comes from using the ambassador-auth-oidc project to negotiate the oath flow. It's the name it uses https://github.com/ajmyyra/ambassador-auth-oidc/search?q=X-auth-token&unscoped_q=X-auth-token

The reason for adding this in addition to the cookie is that the cookie is specific to a browser whereas a token header can be used by a curl or postman

deployment/existing/auth_oidc/dex.tmpl
ryandawsonuk: On this note https://github.com/kubeflow/kubeflow/pull/3607#discussion_r300729280

deployment/existing/auth_oidc/authservice.tmpl
```diff
@@ -66,4 +66,22 @@ spec:
...
+            valueFrom:
+              secretKeyRef:
+                name: authservice
+                key: JWT_HMAC_SECRET
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: authservice
+  namespace: istio-system
```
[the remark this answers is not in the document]
ryandawsonuk: I think this is a separate question for @yanniszark as authservice isn't introduced in this PR

deployment/existing/auth_oidc/envoy-filter.yaml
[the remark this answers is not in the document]
ryandawsonuk: I think the why 'X-Auth-Token' rather than 'Authorization' is a question for the ambassador-auth-oidc project

deployment/existing/auth_oidc/dex.tmpl
ryandawsonuk: I think this is best discussed with @yanniszark as he and I discussed previously

deployment/existing/auth_oidc/envoy-filter.yaml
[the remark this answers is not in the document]
krishnadurai: @ryandawsonuk I agree. I was not aware of where this header came from initially.

deployment/existing/auth_oidc/dex.tmpl
krishnadurai: Sure. We can resolve this separately.

ryandawsonuk:
Looks to me like after https://github.com/kubeflow/kubeflow/commit/de8be1d132da08c1dec84f5c17bbddc643ff77f3 and https://github.com/kubeflow/manifests/commit/1bb77f3f28ce901947d80161f8e6ff83cafbfa1d#diff-3e2b156b8489e2bc7bbe25c58ef1f497 the template changes for this are now referencing the wrong repository as the master files have moved to the manifests repo.

We've actually found another way to achieve this (by patching the installation outside of kfctl) so closing this rather than updating it. Can discuss further if anyone else has the same need.

PR #3621, opened 2019-07-08
Create a module for managing google cloud source repos.

jlewi:
* This module will be used to

   i) Create a Cloud Source Repo if it doesn't exist
   ii) commit and push changes to the Cloud Source Repo from click to deploy.

* The intention is to run this in a background thread so that we don't block
  deployment if there are problems or if its slow.

* We add a unit/manual test that depends on GCP. By default the test is skipped
  but setting certain environment variables causes the test to be run.

* Use a git go client to interact with git rather than relying on shelling out
  to the git CLI. (There is still one place where we shell out)

* Update the dependencies of golang net package because the git client we
  are using requires a newer version

Related to #2152 Run click to deploy requests in their own pod

Related to kubeflow/testing#416 writing component/e2e tests that use GCP - this PR pvoides an example.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3621)


bootstrap/cmd/bootstrap/app/sourceRepos.go
```diff
@@ -0,0 +1,236 @@
...
+	RemoteName = "gcpsourcerepo"
+)
+
+// NewSourceRepo initializes a repo object for the specified local directory.
+// If the directory doesn't exist an error is raised.
+//
+//
+// If the remote repo doesn't exist it is created.
+//
+// TODO(jlewi): What if the repo exists but the directory doesn't? In that case should
```
kunmingg: Current behavior is `repo/version/deploy_name/actual_configs`.
So all deployments on v0.6 will go to same source repo.

We can always check if repo exists, and clone if it does.
Then we might need to create sub folder based on deploy name; update configs; and commit & push whole repo.

jlewi: Why not make it one repo per deployment? Doesn't that simplify things at least to start with?

We can always add support later for putting multiple deployments into the same repository.

bootstrap/cmd/bootstrap/app/sourceRepos_test.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	t.Logf("Using GCP project: %v; to override set environment variable %v", project, GcpProjectEnvName)
+	localDir, err := ioutil.TempDir("", "kfctl-test-")
+
+	if err != nil {
+		t.Fatalf("Could not create temporary directory; error %v", err)
+	}
+
+	repoName := path.Base(localDir)
```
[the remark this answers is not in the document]
jlewi: What does that have to do with the test?

bootstrap/cmd/bootstrap/app/sourceRepos.go
jlewi: I think NewSourceRepo() is compatible with having multiple apps stored in the same repo.
i.e. we could still have
repo/version/deploy_name/actual_configs

To do that we would just set the app_dir to

app_dir = path.join(local_dir, version, deploy_name)

I'm leaning towards making it 1 repo per app at least to start with. The main reason is I think that simplifies things like dealing with merge conflicts and force pushes. If we have multiple apps per repo I think there is more likely to be multiple writers to the repo which makes it harder to reason about.

Also, I'd like to do all the source repo clones and pushes out of band with the actual deployment so that we don't block deployment based on cloning and pushing.

I believe we can do this even if we want to store multiple apps in the same repo but I think its harder to reason about.

kunmingg: What if the repo already exists? Might be leftover from kubeflow older version, or even user happens to create one by themselves.

jlewi: Depending on how we use NewSourceRepo in KfctlServer it would be overwritten.

Specifically here's how this would currently work

1. KfctlServer creates  ${KFAPP}
1. NewSourceRepo  runs `git init` inside ${KFAPP}
1. NewSourceRepo creates SourceRepo if it doesn't exist
1. KfctlServer will do a force push to ${KFAPP}/master 

So whatever is in the repo would be overwritten.

kunmingg: I assume we'll keep repo for user's custom change after they delete the cluster?
Currently `number of repo` quota per project is 3.
We'll likely hit quota issue if we create new repo for every deployment.

jlewi: Good point. We should definitely reuse repos in that case. I don't think that changes any of the code in this PR.

I think it will just change how we integrate it into kfctl sever.

I think each app should push to its own branch so that we are less likely to get conflicting changes but we can resolve details when we integrate into kfctl server.

Issue #3626, opened 2019-07-09
kfctl without platform not working


lluunn:
```
kfctl init ${KFAPP}
cd ${KFAPP}
kfctl generate k8s -V
```
This will give error:
`ERRO[0000] Couldn't find component istio-crds filename="kustomize/kustomize.go:454"`

After removed the istio components from app.yaml, the generate succeeds.
But apply failed with:
```
Error: couldn't apply KfApp:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:  (kubeflow.error): Code 500 with message: couldn't create resources from /Users/lunkai/kfdev/kops-test/kf0708/kustomize/minio.yaml Error: persistentvolumes "minio-pv" is forbidden: error querying GCE PD volume dls-kf-storage-artifact-store: disk is not found
```

See #3618 3618

lluunn:
remove minio and mysql from components, the apply succeeded.

yanniszark:
Hi @lluunn, I'd like to say that I bumped into very similar problems while writing the `existing_arrikto` config file, which is tailored towards running on existing kubernetes clusters.
For example, I changed minio and mysql to use a default storage class.

Because of that, I propose using the `existing_arrikto` config file, with the following procedure:

```bash
# Apply Istio
kubectl apply -f deployment/existing/istio
# Deploy for existing K8s
kfctl init --config=bootstrap/config/kfctl_existing_arrikto.0.6.yaml
kfctl generate k8s
kfctl apply k8s
```

jlewi:
@yanniszark see my comments in in #3627 about the use cases I think we want to support.

vjrantal:
I tested the config `kfctl_k8s_istio.yaml` from the master branch (commit d0c11dd67ebd514f4d6e584b74fecf880245a86c) on my Kubernetes cluster with these steps:

```
kfctl init ${KFAPP} --config=bootstrap/config/kfctl_k8s_istio.yaml -V
cd ${KFAPP}
kfctl generate all -V
kfctl apply all -V
```

The deployment succeeded, all pods seemed to be running fine and I was able to access the dashboard. However, creation of new notebook servers got stuck to a state that to the end user looks like this:

<img width="1188" alt="Screenshot 2019-07-17 at 14 35 54" src="https://user-images.githubusercontent.com/207474/61372600-4ce7bc80-a8a0-11e9-8d66-55b9a37b3aac.png">

[...]

Issue #3618, opened 2019-07-08
Test core 0.6 CUJs on a cluster without identity auth


jlewi:
Kubeflow 0.6 should work even if requests don't contain the user's identity. In this case, the experience should be equivalent to the 0.5 experience.

The purpose of this issue is to try out core 0.6 CUJs using a Kubeflow deployment that is not identity aware.

The CUJ should be something like the following

1. Deploy Kubeflow with ISTIO but no auth on a K8s cluster

    * ISTIO can be exposed via a nodeport but not a loadbalancer since that would be insecure

1. Connect to ISTIO via kubectl port-forward

1. Launch a notebook
1. Build/Train/Deploy from notebook

@lluunn has been running some tests with kops and this could be used to simulate a non GCP, non-identity aware cluster.

Related to:  #3553

/assign @lluunn

lluunn:
Without platform part:
```
kfctl init ${KFAPP}
cd ${KFAPP}
kfctl generate k8s -V
```

This will give error:
`ERRO[0000] Couldn't find component istio-crds            filename="kustomize/kustomize.go:454"`

lluunn:
This might be that istio is currently in gcp package, but we don't include gcp here, but still want istio.

Related to https://github.com/kubeflow/manifests/issues/115 (move istio to its own pkg)
, and possibly https://github.com/kubeflow/kubeflow/issues/3491 (get rid of pkg/component)

jlewi:
@lluunn if a user runs the following

```
kfctl init ${KFAPP}
cd ${KFAPP}
kfctl generate k8s -V
```

I would not expect that to install ISTIO. My interpretation of the above is install only KF applications on my existing cluster. 

If we don't install ISTIO in this case would that fix the problem?

lluunn:
I removed the istio components from app.yaml, the generate succeeds.
But apply failed with:
```
Error: couldn't apply KfApp:  (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize:  (kubeflow.error): Code 500 with message: couldn't create resources from /Users/lunkai/kfdev/kops-test/kf0708/kustomize/minio.yaml Error: persistentvolumes "minio-pv" is forbidden: error querying GCE PD volume dls-kf-storage-artifact-store: disk is not found
```

lluunn:
Actually, I will filed another issue to track the deployment of k8s without platform.
Here I will try to bypass the error to get the cluster up and running

lluunn:
[...]

PR #3627, opened 2019-07-09
Separate installing istio and using istio to 2 overlays

lluunn:
When running kfctl on an existing cluster, we might already have istio installed, but we still want the components to use istio features.

Related to https://github.com/kubeflow/kubeflow/issues/3618 and https://github.com/kubeflow/kubeflow/issues/3626

/cc @jlewi @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3627)

### DIFF
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -85,6 +85,8 @@ func GetKfApp(kfdef *kfdefsv2.KfDef) kftypes.KfApp {
 	return _coordinator
 }
 
+// Config itself is a kustomize package, https://github.com/kubeflow/kubeflow/tree/master/bootstrap/config.
+// getConfigFromCache returns the KfDef yaml (in bytes) with overlays applied, and also writes it to app.yaml.
 func getConfigFromCache(pathDir string, kfDef *kfdefsv2.KfDef) ([]byte, error) {
 
 	configPath := filepath.Join(pathDir, kftypes.DefaultConfigDir)
@@ -96,6 +98,9 @@ func getConfigFromCache(pathDir string, kfDef *kfdefsv2.KfDef) ([]byte, error) {
 	}
 	if kfDef.Spec.UseIstio {
 		overlays = append(overlays, config.NameValue{Name: "overlay", Value: "istio"})
+	}
+	if kfDef.Spec.InstallIstio {
+		overlays = append(overlays, config.NameValue{Name: "overlay", Value: "install_istio"})
 	}
 	if kfDef.Spec.UseBasicAuth {
 		overlays = append(overlays, config.NameValue{Name: "overlay", Value: "basic_auth"})
@@ -121,7 +126,6 @@ func getConfigFromCache(pathDir string, kfDef *kfdefsv2.KfDef) ([]byte, error) {
 			Message: fmt.Sprintf("error getting manifest %v Error %v", configPath, resMapErr),
 		}
 	}
-	// TODO: Do we need to write to file here?
 	writeErr := kustomize.WriteKustomizationFile(kfDef.Name, configPath, resMap)
 	if writeErr != nil {
 		return nil, &kfapis.KfError{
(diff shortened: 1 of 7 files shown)


jlewi:
Thanks @lluunn 

Why do we need a new flag "install_istio" ?

Wouldn't that flag be synonmous with "--platform=gcp"?

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -85,6 +85,8 @@ func GetKfApp(kfdef *kfdefsv2.KfDef) kftypes.KfApp {
 	return _coordinator
 }
 
+// Config itself is a kustomize package, https://github.com/kubeflow/kubeflow/tree/master/bootstrap/config.
+// getConfigFromCache returns the KfDef yaml (in bytes) with overlays applied, and also writes it to app.yaml.
```
jlewi: nit reverse the order of the two lines of the comment so that the first line in the comment is "getConfigFromCache..."

lluunn:
When we have an existing cluster (so platform != gcp), we might still want to install istio, no?
For example, when I am using kops / on-prem.

`kfctl init --use_istio=true --instal_istio=true` will install istio.

bootstrap/pkg/kfapp/coordinator/coordinator.go
lluunn: updated, thanks

yanniszark:
This PR seems to be extending the kustomize mechanism of generating config file.
I was under the impression that this method is being deprecated in favor of the `--config` flag.

In addition, the `existing_arrikto` config file is tailored towards running on existing kubernetes clusters.
Many of the problems that @lluunn countered (e.g. minio, mysql) are resolved (e.g. by using a default storage class).

Because of that, I suggest using the `existing_arrikto` config file, with the following procedure:

```bash
# Apply Istio
kubectl apply -f deployment/existing/istio
# Deploy for existing K8s
kfctl init --config=bootstrap/config/kfctl_existing_arrikto.0.6.yaml
kfctl generate k8s
kfctl apply k8s
```

jlewi:
@yanniszark @lluunn I believe this work and the work in existing_arrikto should be converging.

Specifically, existing_arrikto is currently installing ISTIO via go code and YAML files; but this PR and the related PR #3626 are moving in the direction of making ISTIO applications listed in the config that can be optionally installed. 

I think there are two things that aren't currently covered by the existing_arrikto config solution that this PR is (or could Address)

1. In 0.6 we should provide a suitable experience when platform isn't set and config isn't set
1. existing_arrikto currently doesn't provide a way for install Kubeflow on a cluster which already has ISTIO installed.

I think this PR is addressing both of these.

I'd like to avoid introducing new flags since as @yanniszark says we are moving away from all these different flags to having a single config file with a YAML spec.

So if we want to avoid introducing new flags we need to figure out what is the meaning of 

```
kfctl init --use_istio=true 
```

and 

```
kfctl ini
```

Here's a list of use cases I think we want to support and how we might do this with various flags

| Create Cluster | Install Istio | Use Istio |  Flags
|---------|--------|-------------|------------|
| Yes | Yes | Yes | `kfctl --platform=gcp --use_istio=true` |
| No | Yes | Yes | `kfctl --config=existing_arrikto.yaml` |
| No | No | Yes | `kfctl --use_istio=truel` |

yanniszark:
@jlewi 
I have thought about it and I can't make a good case for having the `use_istio` flag.
Kubeflow depends on certain Istio CRDs to express:
* Which Services are exposed and at what path (VirtualService).
* Who is allowed to access a certain path (ServiceRole, ServiceRoleBinding).

I believe those CRDs are the bare minimum for the Kubeflow controllers to function, so they should be present in every installation of Kubeflow.
We shouldn't need a flag to install those.

Because of this, I can't really find a meaning for the `use_istio` flag.
The `install-istio` flag makes much more sense.
If it is set, we also install the Istio Control Plane.


> 1. In 0.6 we should provide a suitable experience when platform isn't set and config isn't 
> 2. existing_arrikto currently doesn't provide a way for install Kubeflow on a cluster which already has ISTIO installed.

If we make sure the necessary Istio CRDs are installed, then existing_arrikto with just k8s apply should work for existing clusters.
Then the matrix becomes:

Create Cluster | Install Istio | Use Istio | Flags
-- | -- | -- | --
Yes | Yes | Yes | kfctl --platform=gcp --use_istio=true
No | Yes | Yes | kfctl --config=existing_arrikto.yaml (all)
No | No | Yes | kfctl --config=existing_arrikto.yaml (k8s)

lluunn:
close for #3663 .  

/close

PR #3630, opened 2019-07-10
jupyter-web-app: Use kfam to check if user is authorized

### LINKED ISSUE #3556: Jupyter Web App: add userid-header and userid-prefix cli flags
Sub-issue for #3553.
Add cli flags to configure:
* Which header we read for the user info (now it's only `X-Goog-Authenticated-User-Email`)
* What prefix is expected (now it's expected to be in the form of `accounts.google.com:<user>`, so the prefix is `accounts.google.com:`)

/priority p0

cc @jlewi  @kimwnasptd

kimwnasptd:
closes #3556 

This PR will
* Make use of the USERID_{HEADER, PREFIX} ENV Vars
* Query kfam before making a request

If a request has set the userid headers, then the webapp will ask kfam whether this user is authorized for the specific namespace.

If the request doesn't have this header, then the webapp will just make
the needed API calls.

In both cases the webapp will make the k8s requests with its own
privileged SA.

/assign kimwnasptd
/cc lluunn 
/cc kunmingg 
/cc jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3630)

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -1,15 +1,22 @@
 import logging
+import os
 import sys
 import yaml
 import json
+from flask import request
 from kubernetes import client
 import datetime as dt
 from . import api
 
+# The backend will send the first config it will successfully load
 CONFIGS = [
     "/etc/config/spawner_ui_config.yaml",
     "./kubeflow_jupyter/common/yaml/spawner_ui_config.yaml"
 ]
+
+# The values of the headers to look for the User info
+USER_HEADER = os.getenv("USERID_HEADER", "X-Goog-Authenticated-User-Email")
+USER_PREFIX = os.getenv("USERID_PREFIX", "")
 
 
 # Logging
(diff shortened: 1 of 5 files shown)


components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
```diff
@@ -1,16 +1,23 @@
...
 
+# The backend will send the first config it will successfully load
 CONFIGS = [
     "/etc/config/spawner_ui_config.yaml",
     "./kubeflow_jupyter/common/yaml/spawner_ui_config.yaml"
 ]
 
+# The values of the headers to look for the User info
+USER_HEADER = os.getenv("USERID_HEADER", "X-Goog-Authenticated-User-Email")
+USER_PREFIX = os.getenv("USERID_PREFIX", "")
```
kunmingg: Default prefix for IAP is `accounts.google.com:`

kimwnasptd: done.

PR #3632, opened 2019-07-10
Remove the Pipeline SDK from Notebook image

IronPan:
Pipeline release cadence is different form notebook. Prebake the SDK will cause the version outdated quickly.
KF doc should suggest user install SDK manually.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3632)


jlewi:
@IronPan Do you know what test is failing and why?

PR #3645, opened 2019-07-12
create a default profile for the gcp platform

### LINKED ISSUE #3639: Kfctl create profile for user by default
With 0.6.0 a user can no longer create notebooks in the "kubeflow" namespace. They need to create a profile (see https://bit.ly/2UOzzxB) which will create the namespace with appropriate service accounts.

```
kind: Profile
metadata:
  name: <name of the name space to be used>
spec:
  owner:
    kind: User
    name: <ID of the user to use it>
```

We are planning a UI based flow to do that but the UI hasn't landed yet.

In the meantime, we could have `kfctl` automatically create profile for ${USER} running kfctl.
This would mean out of the box the user has a namespace they can run Kubeflow workloads in.

@yanniszark @kkasravi WDYT? Are either of you able to pick this up?

kkasravi:
fixes #3639, dependencies: kubeflow/manifests#207 

- [x] add a --dry-run to apply so that output can be saved into an output directory
- [x] reorder profiles in the component list
- [x] add email parameter to cloud-endpoints
- [x] add profile.yaml to cloud-endpoints

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3645)

### DIFF
--- a/bootstrap/config/kfctl_gcp_iap.0.6.yaml
+++ b/bootstrap/config/kfctl_gcp_iap.0.6.yaml
@@ -36,6 +36,9 @@ spec:
     cloud-endpoints:
     - name: secretName
       value: admin-gcp-sa
+    - initRequired: true
+      name: email
+      value: email
     iap-ingress:
     - name: namespace
       value: istio-system
--- a/bootstrap/config/overlays/basic_auth/kfctl_default.yaml
+++ b/bootstrap/config/overlays/basic_auth/kfctl_default.yaml
@@ -11,9 +11,11 @@ spec:
   - gpu-driver
   - basic-auth
   - basic-auth-ingress
+  - profiles
   packages:
   - cert-manager
   - gcp
+  - profiles
   componentParams:
     ambassador:
     - name: ambassadorServiceType
@@ -28,6 +30,9 @@ spec:
     cloud-endpoints:
     - name: secretName
       value: admin-gcp-sa
+    - name: email
+      value: email
+      initRequired: true
     basic-auth-ingress:
     - name: ipName
       # TODO change value on the fly: value of ipName need to match resource name in deployment entry.
--- a/bootstrap/config/overlays/gcp/kfctl_default.yaml
+++ b/bootstrap/config/overlays/gcp/kfctl_default.yaml
@@ -23,6 +23,9 @@ spec:
     cloud-endpoints:
     - name: secretName
       value: admin-gcp-sa
+    - name: email
+      value: email
+      initRequired: true
     iap-ingress:
     - name: ipName
       # TODO change value on the fly: value of ipName need to match resource name in deployment entry.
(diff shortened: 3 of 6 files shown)


jlewi:
Thanks for the quick fix?

What's the point of dry-run? What is the difference of dry run vs. just running kfctl generate but not kfctl apply?

I thought email is already in KfDef
https://github.com/kubeflow/kubeflow/blob/7f319fd578c25409fcf93bb5008e84bde4b43ba6/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1/application_types.go#L58

So I thought in KfDef we were just going to check if that field was set and if it was create a profile based on it?

jlewi:
@kkasravi Let me know when this is ready for another look?

kkasravi:
@jlewi ready for another look

kkasravi:
> What's the point of dry-run? What is the difference of dry run vs. just running kfctl generate but not kfctl apply?

kfctl generate won't resolve params, it's done in Apply but the data isn't persisted anywhere

kkasravi:
> So I thought in KfDef we were just going to check if that field was set and if it was create a profile based on it?

We could create a global function that builds a Profile and calls kustomize.deployResources. 
Having it in manifests is imo a better option
1. Declarative config
2. Closer to doing everything with `kustomize build` - less "magic" being done by kfctl
3. Similar to how we do other vars like hostname, ipName,

jlewi:
@yanniszark @kkasravi  Creating a kustomize package for a profile feels a bit like overkill. What I mean is if you are defining a profile copying and editing a YAML file specifying a profile seems simpler (i.e. less lines of code modified) then use a kustomize package and writing a kustomization.yaml and trying to substitute in email via config references.

i.e. if I was to tell someone how to go about creating profiles outside of kfctl; I would suggest to them that they just create a YAML file directly; I don't know that I would recommend using a kustomize package.

So if we want to be declarative should we just spit out the YAML file to ${KFAPP}/yaml  during generate and then apply those files?

That said; the whole point of creating the profile when we do kfctl is to provide a bit of syntactic sugar and provide a stop-gap for 0.6; so at least in this PR I don't think its hugely important that we spit out the YAML to a file in ${KFAPP}/somewhere.

I was expecting something very simple; e.g. at the end of kustomize.go
https://github.com/kubeflow/kubeflow/blob/e8a8a6f35ea627997ab6b160c48b43e421cc1d55/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go#L230

we check if kfdef.Spec.email != "" and if it isn't we create a profile.

That is 

1. Platform independent
1. Avoids any ordering issues because we know the profile CR already exists at that point.

I would punt until 0.7 to figure out what is the right way if any way for users to specify one or more profiles that should be setup automatically.

jlewi:
> kfctl generate won't resolve params, it's done in Apply but the data isn't persisted anywhere

What does resolving params mean? Does this mean writing kustomization.yaml files that create config maps based on the param values?

If that's the case; rather than adding "dryrun"; why not move that up to the generate phase?

Either way should we split this into 2 separate PRs; 1 for dry-run and 1 for creating the profile so that these don't block each other?

jlewi:
Thanks @kkasravi

jlewi:
@kkasravi making a copy of the Profile struct resource doesn't seem too bad. I'd be ok with just copying and pasting that struct into kustomize.go or similar. 

Do we have an open issue on removing the v2 dependencies. 

I left a comment [here](https://github.com/kubeflow/kubeflow/pull/3528#issuecomment-511241258) about beginning to strip out ksonnet as a way to get rid of that hokiness.

kkasravi:
closing this since the approach of creating a default profile from within the code has been chosen over the option of using manifests.

PR #3647, opened 2019-07-12
jupyter-web-app: backend: fix default storage class detection

yanniszark:
The default storage class detection code has a bug.
The bug causes it to not check for the `storageclass.kubernetes.io/is-default-class` annotation.

/assign @kimwnasptd 
cc @ioandr @jlewi 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3647)


kimwnasptd:
@yanniszark good catch!
/lgtm
/approve

PR #3651, opened 2019-07-15
coordinator should load packagemanagers and platform KfApps once at construction time

jlewi:
* coordinator go was dynamically recreating the KfApp for the package manager whenever the
actions init/generate/apply/delete were called.

* This prevents storing state in the KfApp's representing the platform and package manager.

* Coordinator also stores maps Coordinator.platforms and Coordinator.packageManagers so why not
  just get them from there?

* I suspect this was legacy code and might have been required when we were embedding the KfDef inside
  the Kustomize/Ksonnet/Gcp structs rather than just storing a pointer.
  * As a result I think we had to dynamically recreate them in order to pick up the changes to KfDef.

* Change the code so KfApps are constructed once in LoadCfgAppFile and after that we get them
  from the appropriate fields in Coordinator.

* Move the code in GetKfApp to  LoadCfgAppFile.

* Remove import ksonnet from coordinator.go
  
  * ksonnet no longer works on master so rather than try to update code to continue to support ksonnet
    we can just return NotSupported errors for any code branches that are ksonnet only

Related to #2152

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3651)

### DIFF
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -22,7 +22,6 @@ import (
 	"github.com/kubeflow/kubeflow/bootstrap/config"
 	kftypes "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps"
 	"github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/gcp"
-	"github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/ksonnet"
 	"github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/minikube"
 	kfapis "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis"
 	kftypesv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps"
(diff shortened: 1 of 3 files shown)


jlewi:
/assign @yanniszark @kkasravi  
I need to wait to see if the tests pass but otherwise this PR should be ready to go.

jlewi:
Test failed with

```
util.py                     71 INFO     Error: couldn't generate KfApp:  (kubeflow.error): Code 500 with message: coordinator Generate failed for gcp:  (k
ubeflow.error): Code 400 with message: could not generate deployment manager configs under gcp_config Error: could not copy /mnt/test-data-volume/kubeflow
-presubmit-kfctl-go-iap-istio-3651-588c216-3712-7dc3/kfctl-7dc3/.cache/kubeflow/pull/3651/head/deployment/gke/deployment_manager_configs/cluster.jinja to
/mnt/test-data-volume/kubeflow-presubmit-kfctl-go-iap-istio-3651-588c216-3712-7dc3/kfctl-7dc3/gcp_config/cluster.jinja Error cannot create directory: open
 /mnt/test-data-volume/kubeflow-presubmit-kfctl-go-iap-istio-3651-588c216-3712-7dc3/kfctl-7dc3/.cache/kubeflow/pull/3651/head/deployment/gke/deployment_ma
nager_configs/cluster.jinja: no such file or directory
```

jlewi:
@yanniszark @kkasravi tests are passing so this is ready for review.

PR #3652, opened 2019-07-15
KfctlServer should take KfDef as the request payload.

jlewi:
* Delete the method DeployWithKfctl
* kfctlServer will handle the create requests
* CreateDeployment should take a KfDef as the payload and return a kfDef
* kfctlServer - handleDeployment - add the logic necessary to handle deployment
* router - add an IAM access check to see if the request is authorized

Related to: #2152

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3652)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/kfctlServer.go
+++ b/bootstrap/cmd/bootstrap/app/kfctlServer.go
@@ -1,12 +1,35 @@
 package app
 
+// TODO(jlewi): How could we create a unittest? I think one of the things we'd like to test
+// is the interaction between the createDeployment request handler and the background processing.
+// We'd like to verify that if we issue subsequent requests we will eventually get an updated status.
+//
+// 1 test we can right is to inject the channel.
+
 import (
+	"cloud.google.com/go/container/apiv1"
 	"context"
+	"encoding/base64"
 	"encoding/json"
-	"github.com/go-kit/kit/endpoint"
+	"fmt"
 	httptransport "github.com/go-kit/kit/transport/http"
+	kftypes "github.com/kubeflow/kubeflow/bootstrap/pkg/apis/apps"
+	"github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/coordinator"
+	"github.com/kubeflow/kubeflow/bootstrap/pkg/kfapp/gcp"
+	kfdefsv2 "github.com/kubeflow/kubeflow/bootstrap/v2/pkg/apis/apps/kfdef/v1alpha1"
+	"github.com/kubeflow/kubeflow/bootstrap/v2/pkg/kfapp/kustomize"
+	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
+	"golang.org/x/oauth2"
+	"google.golang.org/api/option"
+	containerpb "google.golang.org/genproto/googleapis/container/v1"
+	"k8s.io/api/v2/core/v1"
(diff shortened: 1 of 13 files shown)


gabrielwen:
shouldn't we add KfDef as a field of CreateResponse?

jlewi:
@gabrielwen Do you mean as a header field of the HTTP request? Right now KFDef is the body o fthe request.

This follow the K8s convention
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#-strong-write-operations-deployment-v1-apps-strong-

bootstrap/cmd/bootstrap/app/kfctlServer.go
```diff
@@ -60,15 +301,214 @@ func (r *kfctlServer) RegisterEndpoints() {
...
+		}
+	}
+
+	checkIsMatch := func() bool {
+		s.kfDefMux.Lock()
+		defer s.kfDefMux.Unlock()
+		return isMatch(&s.latestKfDef, &req)
+	}
+
+	if !checkIsMatch() {
```
kunmingg: If `checkIsMatch == true`, we don't need to push same request to channel?

jlewi: We do down down below; line 466. If the request is a match then we keep processing.
We check that it is a valid request and if it is we keep going which means we add it to the channel to actually get processed.

bootstrap/cmd/bootstrap/app/kfctlServer_test.go
```diff
@@ -0,0 +1,327 @@
...
+	// TODO(jlewi): DeepEqual is returning false even though when a pretty print them the results
+	// look the same. Need to figure out how to validate the test properly.
+	//if !reflect.DeepEqual(expectReqStripped, v)
+	if !reflect.DeepEqual(expectReqStripped.Spec.PackageManager, v.Spec.PackageManager) {
+		pWant, _ := Pformat(expectReqStripped)
+		pActual, _ := Pformat(v)
+		t.Errorf("Incorrect CreateDeployment on channel :got\n:%v\nwant:%v", pActual, pWant)
+	}
+
+	expectedToken := "access1234"
```
gabrielwen: nit: should we also refer to this var instead of hardcoding in field `Value` above?

bootstrap/cmd/bootstrap/app/kfctlServer.go
```diff
@@ -15,30 +38,248 @@ const KfctlCreatePath = "/kfctl/apps/v1alpha2/create"
...
+	// Start a background thread to process requests
+	go s.process()
+
+	return s, nil
+}
+
+// handleDeployment handles a single deployment.
+// Returns a pointer to an updated KfDef
+//
+// Not thread safe.
```
gabrielwen: add more context as which part(s) is not thread safe?

bootstrap/cmd/bootstrap/app/kfctlServer.go
```diff
@@ -60,15 +301,214 @@ func (r *kfctlServer) RegisterEndpoints() {
...
+
+		secrets = append(secrets, s)
+	}
+
+	d.Spec.Secrets = secrets
+}
+
+// CreateDeployment creates the deployment.
+//
+// Not thread safe
```
gabrielwen: same here.  I think my concern is that is it possible that we have multiple invocations?  should we make server codes idempotent?

PR #3663, opened 2019-07-15
Add kfctl config for k8s only

### LINKED ISSUE #3626: kfctl without platform not working

lluunn:
Related to #3618 
fix #3626 . 

Create Cluster | Install Istio | Use Istio | Flags
-- | -- | -- | --
Yes | Yes | Yes | `kfctl --platform=gcp --use_istio=true`
No | Yes | Yes | `kfctl --config=kfctl_k8s_istio.yaml`
No | No | Yes | `kfctl --config=kfctl_k8s_istio.yaml` (modify config first, see comment in the file)

/cc @jlewi @kunmingg @yanniszark 

I didn't use existing_arrikto because that also installs some dex thing.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3663)


jlewi:
@lluunn Overally looks good. Instead of having separate config files. Should we just have a single config file and maybe comment out the ISTIO package or maybe leave it in but add a comment telling people to comment it out if they don't want to install ISTIO?

This will get a little easier once I submit the changes to kfctl to use KfDef.Applications.

In the meantime, I think we should try to minimize the number of configs that we maintain.

Also could you please open up an issue about adding an E2E test?

Thanks

lluunn:
Merged the config, and added comment.
Thanks

PR #3668, opened 2019-07-16
kustomize.go should use KfDef.Applications

jlewi:
* kustomize module should use KfDef.Applications to get information about each
  application to install

  * Each Application refers to a kustomize package to build and install

* To provide backwards compatibility, kustomize.go will convert the old style
  kfDef.componentParams, kfDef.components to kfDef.Applications

Related to #2152
Related to #3491

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3668)

### DIFF
--- a/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
+++ b/bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
@@ -166,6 +166,76 @@ func (kustomize *kustomize) initComponentMaps() error {
 	return nil
 }
 
+// backfillApplications backfills the applications from the components componentParams
+func (kustomize *kustomize) backfillApplications() error {
+	currentApplications := map[string]bool{}
+
+	if kustomize.kfDef.Spec.Applications == nil {
+		kustomize.kfDef.Spec.Applications = []kfdefsv2.Application{}
+	}
+
+	// Build a set of applications currently defined.
+	for _, a := range kustomize.kfDef.Spec.Applications {
+		currentApplications[a.Name] = true
+	}
+
+	// We need repoCache to know the local path strip from componentPathMap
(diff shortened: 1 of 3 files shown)


jlewi:
/assign @yanniszark 
/assign @lluunn 

Need to wait for tests to pass but otherwise should be ready for review.

bootstrap/v2/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -166,6 +166,76 @@ func (kustomize *kustomize) initComponentMaps() error {
...
+
+	// We need repoCache to know the local path strip from componentPathMap
+	repo, ok := kustomize.kfDef.Status.ReposCache[kftypes.ManifestsRepoName]
+
+	if !ok {
+		err := fmt.Errorf("Could not backfillApplications; missing repo cache for repo %v", kftypes.ManifestsRepoName)
+		return errors.WithStack(err)
+	}
+
+	// Loop over all the components
```
lluunn: IIUC, each component will become one application?
Is that by design, or just for backward compatible?

For example, for katib https://github.com/kubeflow/manifests/tree/master/katib-v1alpha2
I think all components as one application makes more sense.

jlewi: I think that's initially for backwards compatibility.

Longer term its 1 application for kustomize package. So if there's a way for us to combine all of those resources into a single kustomize package then I think we could potentially consider it a single application.

jlewi: I think this is the open issue tracking composition
https://github.com/kubernetes-sigs/kustomize/issues/1251

PR #3669, opened 2019-07-16
gcp should support KfDef.Application

jlewi:
* We want to migrate to using KfDef.Application to store application parameters
* The GCP module needs to set some of the application parametes
  we use the KfDef helper routines to set application parameters

  * The helper routines are compatible with both KfDef.Applications and the
    older KfDef.component parameters.

* Add KfDef.labels to the deployments that get created

* Fix some unsafe code in gcp.go; some of the code was assuming errors were of
  type KfError without testing the assertion

* Add an IsValid method to test the GCP config

* Gcp should use the KfDef WriteConfigFile method and not define its own method.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3669)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -1704,6 +1704,9 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
 		if err := gcp.kfDef.SetApplicationParameter("basic-auth-ingress", "hostname", gcp.kfDef.Spec.Hostname); err != nil {
 			return errors.WithStack(err)
 		}
+		if err := gcp.kfDef.SetApplicationParameter("basic-auth-ingress", "project", gcp.kfDef.Spec.Project); err != nil {
+			return errors.WithStack(err)
+		}
 	} else {
 		if err := gcp.kfDef.SetApplicationParameter("iap-ingress", "ipName", gcp.kfDef.Spec.IpName); err != nil {
 			return errors.WithStack(err)
@@ -1744,7 +1747,7 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
 		return errors.WithStack(err)
 	}
 
-	// TODO(jlewi): Why are we setting this here?
+	// TODO(jlewi): Why are we setting usage Id here (gcp.go) and not in kustomize.go so we do it for all platforms?
 	rand.Seed(time.Now().UnixNano())
 	if err := gcp.kfDef.SetApplicationParameter("spartakus", "usageId", strconv.Itoa(rand.Int())); err != nil {
 		if kfdefs.IsAppNotFound(err) {
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -1723,17 +1723,6 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
 		minioPdName := gcp.kfDef.Name + "-storage-artifact-store"
 		mysqlPdName := gcp.kfDef.Name + "-storage-metadata-store"
 
-		// TODO(jlewi): I think we can delete this. It looks like we still had componentParams pipeline
-		// but pipeline wasn't listed in components which explains why when we migrated to Application
-		// there wasn't any Application named pipeline.
-		// Looking at https://github.com/kubeflow/manifests/tree/master/pipeline
-		// there is no base application named pipeline.
-		//if err := gcp.kfDef.SetApplicationParameter("pipeline", "mysqlPd", mysqlPdName); err != nil {
-		//	return errors.WithStack(err)
-		//}
-		//if err := gcp.kfDef.SetApplicationParameter("pipeline", "minioPd", minioPdName); err != nil {
-		//	return errors.WithStack(err)
-		//}
 		if err := gcp.kfDef.SetApplicationParameter("minio", "minioPd", minioPdName); err != nil {
 			return errors.WithStack(err)
 		}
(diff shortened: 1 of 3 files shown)


jlewi:
/assign @yanniszark 
/assign @lluunn 

Still waiting on the tests to pass but this should be good to go otherwise.

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -765,15 +743,7 @@ func (gcp *Gcp) updateDM(resources kftypes.ResourceEnum) error {
 		}
 	}
 
-	// Setup kube config
-	// TODO(#3061): figure out how to properly build config without kubeconfig.
-	// TODO(jlewi): Not sure about #3061. I think we only want to run gcloud get-credentials if running
-	// as a CLI and not when running on click to deploy. Should we check TokenSource? What's the best
-	// way to tell whether we are running in cluster?
-
-	accessToken, _ := gcp.kfDef.GetSecret(GcpAccessTokenName)
```
lluunn: Are we still using this token secret?

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -1732,68 +1659,104 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
...
-		gcp.kfDef.Spec.Hostname = gcp.kfDef.Name + ".endpoints." + gcp.kfDef.Spec.Project + ".cloud.goog"
+
+	if err := gcp.kfDef.SetApplicationParameter("profiles", "admin", gcp.kfDef.Spec.Email); err != nil {
+		return errors.WithStack(err)
 	}
+
 	if gcp.kfDef.Spec.UseBasicAuth {
-		gcp.kfDef.Spec.ComponentParams["basic-auth-ingress"] = setNameVal(gcp.kfDef.Spec.ComponentParams["basic-auth-ingress"], "ipName", gcp.kfDef.Spec.IpName, true)
-		gcp.kfDef.Spec.ComponentParams["basic-auth-ingress"] = setNameVal(gcp.kfDef.Spec.ComponentParams["basic-auth-ingress"], "hostname", gcp.kfDef.Spec.Hostname, true)
-		gcp.kfDef.Spec.ComponentParams["basic-auth-ingress"] = setNameVal(gcp.kfDef.Spec.ComponentParams["basic-auth-ingress"], "project", gcp.kfDef.Spec.Project, true)
```
lluunn: do you need to set "project" param

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -1732,68 +1659,104 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
...
+		log.Infof("Configuring pipeline, minio, and mysql applications")
 		minioPdName := gcp.kfDef.Name + "-storage-artifact-store"
 		mysqlPdName := gcp.kfDef.Name + "-storage-metadata-store"
-		gcp.kfDef.Spec.ComponentParams["pipeline"] = setNameVal(gcp.kfDef.Spec.ComponentParams["pipeline"], "mysqlPd", mysqlPdName, false)
-		gcp.kfDef.Spec.ComponentParams["pipeline"] = setNameVal(gcp.kfDef.Spec.ComponentParams["pipeline"], "minioPd", minioPdName, false)
-		if strings.HasPrefix(gcp.kfDef.Spec.PackageManager, "kustomize") {
-			gcp.kfDef.Spec.ComponentParams["minio"] = setNameVal(gcp.kfDef.Spec.ComponentParams["minio"], "minioPd", minioPdName, false)
-			gcp.kfDef.Spec.ComponentParams["mysql"] = setNameVal(gcp.kfDef.Spec.ComponentParams["mysql"], "mysqlPd", mysqlPdName, false)
+
+		// TODO(jlewi): I think we can delete this. It looks like we still had componentParams pipeline
```
lluunn: I think it's true, looking at https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.0.6.yaml,
pipeline seems to be the package name. There is no "pipeline" component

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -1732,68 +1659,104 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
...
-		if comp == "spartakus" {
-			rand.Seed(time.Now().UnixNano())
-			gcp.kfDef.Spec.ComponentParams["spartakus"] = setNameVal(gcp.kfDef.Spec.ComponentParams["spartakus"],
-				"usageId", strconv.Itoa(rand.Int()), true)
+	// TODO(jlewi): Why is this hard coded here?
+	if err := gcp.kfDef.SetApplicationParameter("notebook-controller", "injectGcpCredentials", "true"); err != nil {
+		return errors.WithStack(err)
+	}
+
+	// TODO(jlewi): Why are we setting this here?
```
lluunn: so that rand.Int() in usageId will be random?

bootstrap/pkg/kfapp/gcp/gcp.go
jlewi: Yes; good catch thanks.

bootstrap/pkg/kfapp/gcp/gcp.go
jlewi: Updated the comment; I meant here in gcp.go as opposed to kustomize.go

bootstrap/pkg/kfapp/gcp/gcp.go
jlewi: Done.

bootstrap/pkg/kfapp/gcp/gcp.go
jlewi: Not anymore. This was originally used as a way for click to deploy to inject the OAuth credential.
But we changed the code so Gcp.go now has a token source and code just calls that to get the token.

PR #3675, opened 2019-07-16
turn off Istio isolation for basic auth

kunmingg:
As of today basic auth will be for single user or demo purpose.
Will turn off isolation.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3675)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -1707,6 +1707,9 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
 		if err := gcp.kfDef.SetApplicationParameter("basic-auth-ingress", "project", gcp.kfDef.Spec.Project); err != nil {
 			return errors.WithStack(err)
 		}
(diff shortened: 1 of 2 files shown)


kunmingg:
Got `runtime error: invalid memory address or nil pointer dereference`,
seems some bug around merging overlays

kunmingg:
Seems formatting issue

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -1707,6 +1707,9 @@ func (gcp *Gcp) Generate(resources kftypes.ResourceEnum) error {
 		if err := gcp.kfDef.SetApplicationParameter("basic-auth-ingress", "project", gcp.kfDef.Spec.Project); err != nil {
 			return errors.WithStack(err)
 		}
+		if err := gcp.kfDef.SetApplicationParameter("istio", "clusterRbacConfig", "OFF"); err != nil {
```
jlewi: Why are we setting this in code? Shouldn't this be set in the KfDef.yaml?

jlewi: NM;  I get it this is for backwards compatibility.

PR #3693, opened 2019-07-18
Create PodDefault if platform is GCP

### LINKED ISSUE #3640: [GCP] Create a PodDefault and GCP secret in namespace ${KUBEFLOW}-user by default
With 0.6 notebooks can no longer be spawned in namespace  kubeflow.

instead users should (see https://bit.ly/2UOzzxB)

1. create their own namespace by defining a profile
1. Download and install a GCP secret in that namespce
1. Create a PodDefault in that namespace

Currently users have to do this manually https://bit.ly/2UOzzxB.

kfctl should do this by default for the user running kfctl. This creates an experience equivalent to what we have in 0.5.

I think we can reuse the existing kubeflow-user GCP service account.

#3639 will define the profile by default for the user.

One potential complication is that we will need to get the order of operations correct e.g.
we need the profile to be created before we can install the service account in the namespace.

gabrielwen:
- [x] Remove random suffix from profile name and corresponding namespace.
- [x] Add a blocking wait for the namespace to be created.
- [x] Add `ConfigPodDefault` to deploy `PodDefault` to the default namespace.
- [x] Invoke `ConfigPodDefault` in coordinator.  This is because platform runs before k8s, but we need both to be run so that the deployment state is ready for `ConfigPodDefault` (secret and profile are both created).  Therefore invoking this in gcp is not an option.  I decided not to do this in kustomize to keep it cleaner and independent of GCP pkg.  coordinator is already depending on GCP pkg so I think it's ok to put it there.

Fixes #3640 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3693)

### DIFF
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -783,6 +783,8 @@ func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
 		if err := platform(); err != nil {
 			return err
 		}
+		// TODO(gabrielwen): Need to find a more proper way of injecting plugings.
+		// https://github.com/kubeflow/kubeflow/issues/3708
 		return gcpAddedConfig()
 	case kftypes.K8S:
 		return k8s()
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -755,14 +755,35 @@ func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
 		return nil
 	}
 
+	gcpAddedConfig := func() error {
+		if kfapp.KfDef.Spec.Email == "" || kfapp.KfDef.Spec.Platform != "gcp" {
+			return nil
+		}
+		p := kfapp.Platforms[kfapp.KfDef.Spec.Platform]
+		if p == nil {
+			return &kfapis.KfError{
+				Code:    int(kfapis.INTERNAL_ERROR),
+				Message: "gcp not in Platforms",
+			}
+		}
+		gcp := p.(*gcp.Gcp)
+		return gcp.ConfigPodDefault()
+	}
+
 	switch resources {
 	case kftypes.ALL:
 		if err := platform(); err != nil {
 			return err
 		}
-		return k8s()
+		if err := k8s(); err != nil {
+			return err
+		}
+		return gcpAddedConfig()
 	case kftypes.PLATFORM:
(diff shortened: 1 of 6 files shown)


gabrielwen:
@jlewi currently the name of profile/namespace is `kubeflow-<email with characters substituted>`, e.g. `kubeflow-gabrielwen-at-google-com` in my case, is this ok for you or do you want it to be `kubeflow-user`?

gabrielwen:
this is created PodDefault:
```
32073 (24)> kubectl get poddefaults -n kubeflow-gabrielwen-at-google-com -o yaml
apiVersion: v1
items:
  desc: add gcp credential
  env:
  - name: GOOGLE_APPLICATION_CREDENTIALS
    value: /secret/gcp/user-gcp-sa.json
  kind: PodDefault
  metadata:
    creationTimestamp: 2019-07-18T22:18:48Z
    generation: 1
    name: add-gcp-secret
    namespace: kubeflow-gabrielwen-at-google-com
    resourceVersion: "4446"
    selfLink: /apis/kubeflow.org/v1alpha1/namespaces/kubeflow-gabrielwen-at-google-com/poddefaults/add-gcp-secret
    uid: 03c4d829-a9aa-11e9-b5c9-42010af0000b
  spec:
    selector:
      matchLabels:
        add-gcp-secret: "true"
  volumeMounts:
  - mountPath: /secret/gcp
    name: secret-volume
  volumes:
  - name: secret-volume
    secret:
      secretName: user-gcp-sa
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

gabrielwen:
it may be because the name is too long, let's trump everything after `@` in the email address, then.

jlewi:
/lgtm
/approve

I had one minor comment though about making an error message more informative.

/hold for another look by @yanniszark 

@yanniszark Could you take another look please? Trying to wrap this up in preparation for 0.6.1 later this week

gabrielwen:
@jlewi hold didn't work.

@yanniszark let me know what you think.  I can make a followup PR for it.

yanniszark:
@jlewi looks like hold didn't work. Do you know why that happened?

> I think we want to move in the direction of having K8s controller like semantics.
> i.e. we define a reconcile function that is just called repeatedly until some terminal conditions are reached.

I agree with that approach.
However, since we are adding code to achieve this effect now, I feel it would be better to go with a more extensible way that could be leveraged by platforms in v0.6.
It would be just a few code lines more and we would avoid calling GCP-specific code in the common coordinator code, as well as give other platforms the ability to do things after Apply.
I know that existing_arrikto would like to use it, possibly others.

Issue #3151, opened 2019-04-30
Update AWS deploy to use ISTIO and Kustomize for 0.6


jlewi:
0.6 will introduce ISTIO and Kustomize.

Filing this issue to track updating the AWS deployment instructions

/cc @Jeffwan

jlewi:
Bumping to P0. P0's are release blocking. I think we want to have a story for continuing to deploy on AWS before we finalize the first 0.6 release.

jlewi:
ping?

Jeffwan:
@jlewi Just come back from vacation.. manifest changes are almost done. I will send out PR this week.

jlewi:
@Jeffwan Sounds great. We are coordinating the release on #release if you can try it out on AWS and report any issues that would be great.

jlewi:
It looks like #3695 was cherry-picked onto the v0.6-branch.

@Jeffwan Should we close this issue now or wait for an RC and verify it is working?

Jeffwan:
@jlewi I think we can close it for now. All basic support has been added and cherry-picked back to 0.6. For improvement and bug fix. I will submit separate PRs and we don't need issue to track it.

PR #3695, opened 2019-07-19
Add AWS kfctl(golang) support and 0.6 support

Jeffwan:
Part of the work of https://github.com/kubeflow/kubeflow/issues/3151

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3695)


jlewi:
/assign @jlewi 

HI @Jeffwan moving forward I'd like to try to get away from adding more command line flags. The desired experience will be closer to something like

```
kfctl init --config=<path to config file>
```

So rather than define new config flags; you should define an AWSPlugin
https://github.com/kubeflow/kubeflow/blob/e6944f35149f7c362d32fb9d57fc9f3842a46347/bootstrap/pkg/kfapp/gcp/gcp_plugin.go#L8

And then any AWS fields (e.g. region) should be provided as part of that plugin and set in the KFDef file.
In kfctl you could add logic to try to intelligently set those defaults based on the user's locally configured preferences.

bootstrap/config/kfctl_aws.yaml
```diff
@@ -0,0 +1,164 @@
+apiVersion: kfdef.apps.kubeflow.org/v1alpha1
+kind: KfDef
+metadata:
+  name: kubeflow-aws
+  namespace: kubeflow
+spec:
+  platform: aws
+  componentParams:
```
jlewi: The fields packages, components, componentParams are the deprecated fields that were there for ksonnet.

If you want you can still use these fields but since its a new config I might suggest using the new field

Spec.applications

Each application is just a pointer to a kustomize package.

kfctl will automatically update kfdef.spec.applications based on packages, components,componentParams

so if you run kfctl and then look at app.yaml you can copy the value of KfDef.Spec.Applications from there.

Jeffwan: Sure. I will make some clean ups as you suggest. 
1. Define a plugin to avoid bring in more options
2. Use Applications instead.

jlewi: Thanks

Jeffwan:
Changes are not ready yet. Still working on it. I will address above comments.

Jeffwan:
@jlewi @kkasravi  Changes are ready and I address the feedbacks you left and please have a check.  Only confusion part is what's the best practice if we have different config patterns? Should we create different kfctl_aws_**.yaml template for users? 

1. Use kustomize manner in config file.  User should install kubeflow like this `kfctl init ${clusterName} --config=<path to config file>. 
2. extract aws options to aws plugin. 
3. Revert some of unrelated changes.

kkasravi:
Thanks @Jeffwan 

> @jlewi @kkasravi Changes are ready and I address the feedbacks you left and please have a check. Only confusion part is what's the best practice if we have different config patterns? Should we create different kfctl_aws_**.yaml template for users?


Could you explain a bit more on the different config patterns or point me to the relevanant part(s), or paste several kfctl_aws_*.yaml's you had in mind?

jlewi:
/lgtm
/approve

@Jeffwan I left a couple comments but in the interest of trying to get this into the next RC I'm LGTM.

Jeffwan:
> Thanks @Jeffwan
> 
> > @jlewi @kkasravi Changes are ready and I address the feedbacks you left and please have a check. Only confusion part is what's the best practice if we have different config patterns? Should we create different kfctl_aws_**.yaml template for users?
> 
> Could you explain a bit more on the different config patterns or point me to the relevanant part(s), or paste several kfctl_aws_*.yaml's you had in mind?

I image that user want simple setups can use `bootstrap/config/kfctl_aws.yaml` and users want entire aws service integrations like auth, file storage should use `bootstrap/config/kfctl_aws_coginito.yaml`.  Should I keep two config there? Or just use one `kfctl_aws.yaml` with basic setups and ask users to follow documentation to change yaml file like add different fields by themselves?

deployment/aws/infra_configs/cluster_features.sh
```diff
@@ -1,14 +0,0 @@
-# privateAccess enable private access for your Amazon EKS cluster's Kubernetes API server endpoint
```
fabboe: Hi @Jeffwan - thanks for your work on this. Did you delete this intentionally? For v0.6.1 this broke https://github.com/kubeflow/kubeflow/blob/master/scripts/aws/util.sh - which relies on cluster_features.sh

Following https://www.kubeflow.org/docs/aws/deploy/install-kubeflow/ with export KUBEFLOW_TAG=v0.6.1

Jeffwan: Oh. I see. Thanks for letting me know. This will be only used in 0.5.1 and 0.6 has few changes for this. I will update docs this week to give latest guidance.

fabboe: Cool - thanks!

Issue #3517, opened 2019-06-21
[Bug]GCR read authentication in auto-provisioned nodes


aakashbajaj:
I am using ```kfctl``` binary to generate cluster configuration and deploy kubeflow on GKE. On top of that I am running kubeflow pipelines. Some of the components in my pipeline require high CPU count and thus kubernetes auto-provisions a new node to run the mentioned pod.
My components pull images from private GCR repo in same project.
But when pod is being deployed on the node, it fails with `ErrImagePull` error stating authentication error (doesn't have permissions to pull from GCR).
The same image when pulled on the GPU pool works fine.
I have to give read permissions explicitly to the Node Auto Provisioning Pool:
```
gcloud beta container clusters update ${CLUSTER} \
  --project ${PROJECT} \
  --zone ${ZONE} \
  --enable-autoprovisioning \
  --max-cpu 20 \
  --max-memory 200 \
  --max-accelerator type=nvidia-tesla-p100,count=10 \
  --verbosity error \
  --autoprovisioning-scopes=https://www.googleapis.com/auth/devstorage.read_only
```
After this I am able to deploy pods on auto-provisioned nodes. This should be enabled during the ```kfclt``` deployment.

aakashbajaj:
@jlewi can you point out relevant files.
I will look into it if possible.

jlewi:
https://github.com/kubeflow/kubeflow/blob/5c4fc0b0afab30b7a43dae534db8d3a633ccf214/deployment/gke/deployment_manager_configs/cluster.jinja#L98

Thanks @aakashbajaj 

Do we need to set

1. the service account to be used by auto-provisioned node pools?
1. The scopes to attach to auto-provisioned node pools?

jbottum:
@aakashbajaj did you get this resolved ?

jlewi:
Duplicate of https://github.com/kubeflow/kubeflow/issues/4259

PR #3696, opened 2019-07-19
Set default scopes to be used by node auto-provisioning

daikeshi:
Default OAuth scopes should be set to `VM_OAUTH_SCOPES` if node auto-provisioning is enabled, ow the node created by the auto provisioning will not have gcp storage access. I think this PR will fix issue https://github.com/kubeflow/kubeflow/issues/3517.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3696)


daikeshi:
By default, nodes created by the auto provisioning only have the following oath scopes:
```
autoscaling:
  autoprovisioningNodePoolDefaults:
    oauthScopes:
    - https://www.googleapis.com/auth/logging.write
    - https://www.googleapis.com/auth/monitoring
```

Issue #3638, opened 2019-07-11
[GCP] Integrate with workload identity


jlewi:
We should integrate with GKE Workload Identity so users should no longer have to download and
manage service accounts directly.
https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity

lluunn:
I will look into this.
/assign

lluunn:
It's pretty straightforward to use the feature, basically:
1. create gcp service account
1. create k8s service account
1. add IAM binding

I think there are two tasks:
First, integrate with KF deployment and core components in kubeflow namespaces.
We can get rid of `admin-gcp-sa` in [manifest](https://github.com/kubeflow/manifests/search?q=admin-gcp-sa&unscoped_q=admin-gcp-sa) and kfctl [gcp.go](https://github.com/kubeflow/kubeflow/blob/e6944f35149f7c362d32fb9d57fc9f3842a46347/bootstrap/pkg/kfapp/gcp/gcp.go#L68)

Second, integrate with profile. I am not sure what's the ideal user experience.
Previously, we ask user to create gcp service account + download key + create secret.
Now, we don't need "download key + create secret", but we need a IAM binding, which I think users (in the multi-user scenario) might not have permission to do? 
Or should profile controller take care of creating IAM binding as well? I think there might be a problem that how do we know if the user is allowed to use this service account.

WDYT @jlewi @kunmingg

lluunn:
Actually, the IAM binding is different here:
`gcloud iam service-accounts add-iam-policy-binding`

Which is not the project level iam binding:
`gcloud projects add-iam-policy-binding`

So users might be able to do it, still investigating.

lluunn:
Tested with a pod with workload identity setup:

```
gcloud auth list
ACTIVE  ACCOUNT
*       demo1-admin@kai-test2.iam.gserviceaccount.com
```

gsutil works.

python library also works:
```
client = storage.Client()
bucket = client.get_bucket("kai-test2-models")
bucket.list_blobs()
...
```

jlewi:
Thanks @lluunn so the remaining work here is to figure out how to out to integrate with profiles so that by default any new profile created will automatically bind the edit service account to the appropriate GCP service account?

Any thoughts on how we might do that?

jlewi:
Looks like we have #3917 tracking creating default credentials when creating a new profile.

jlewi:
@lluunn Are we enabling workload identity by default on master?
[...]

PR #3697, opened 2019-07-19
Support gcp workload identity

lluunn:
For #3638
- Pair with https://github.com/kubeflow/manifests/pull/238
- To enable the feature, add the following to the KfDef config:
  ```
  plugins:
  - name: gcp
    spec:
      enableWorkloadIdentity: true
  ```
 

- When workload identity is enabled:
  1) We don't create key for GCP service account
  2) So no secret for GCP service account, we don't mount secretes and set `GOOGLE_APPLICATION_CREDENTIALS`
  3) We create a corresponding k8s service account for each GCP service account, and set things up (see [doc](https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity))
  

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3697)


jlewi:
Thanks @lluunn 
nit: can you link to the relevant issue?

jlewi:
> We create a corresponding k8s service account for each GCP service account, and set things up (see doc)

Do we create a new K8s SA or do we use one of the existing service accounts?

Can we link the GCP service account to the editor service account that profiles creates?

jlewi:
Looking at the instructions:
https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#migrate_workloads_to_workload_identity

It looks like we just need to add an IAM policy binding to the GSA (Google Service Account).

So in principle we could allow the profile controller to setup the binding; it might look something like this

* Profile controller runs with a KSA linked to a GSA
* Profile's GSA has privileges sufficient to
   * Add policy bindings to a GSA
* In the profile; user specifies the GSA and profile controller binds the KSA to it.

This still doesn't address the question of what resources the KF-${USER} GSA granting access to? i.e. someone would still have to set up a different a project with resources like BQ for each time.

So perhaps for now until we better understand what users want we just give users instructions for adding the appropriate KSA to GSA binding in KF-${USER}.

We should do that by default though for the KF-${USER} namespace that kfctl creates.

lluunn:
I updated the manifest change: https://github.com/kubeflow/manifests/pull/238

Will change this PR accordingly.

lluunn:
Sorry just pushed the new commits

lluunn:
typo fixed, thanks

PR #3714, opened 2019-07-22
Add metadata istio overlay into bootstrap configs

zhenghuiwang:
part of fixing https://github.com/kubeflow/metadata/issues/101

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3714)

### DIFF
--- a/bootstrap/config/kfctl_gcp_iap.0.6.yaml
+++ b/bootstrap/config/kfctl_gcp_iap.0.6.yaml
@@ -62,6 +62,9 @@ spec:
     katib-ui:
     - name: overlay
       value: istio
+    metadata:
+    - name: overlay
+      value: application
     minio:
     - name: overlay
       value: minioPd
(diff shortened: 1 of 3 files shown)


bootstrap/config/kfctl_gcp_iap.0.6.yaml
```diff
@@ -62,6 +62,9 @@ spec:
     katib-ui:
     - name: overlay
       value: istio
+    metadata:
+    - name: overlay
+      value: application
```
lluunn: I thought you need overlay istio?

zhenghuiwang: Here is the manifest dir: https://github.com/kubeflow/manifests/tree/master/metadata/overlays/istio

Should this value be `istio`?

zhenghuiwang: QQ: I don' need to change any file with comments like this, right?
```
# TODO(jlewi): This file is legacy config for bootstrapper.
# we can delete it once bootstrapper is using kfctl and KfDef.
......
```

PR #3718, opened 2019-07-23
use --config for testing kfctl

lluunn:
- add config for gcp iap master and gcp basic auth master
- for https://github.com/kubeflow/kubeflow/issues/3744

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3718)


lluunn:
/assign @jlewi 
Ready for review

bootstrap/config/kfctl_gcp_basic_auth_master.yaml
```diff
@@ -0,0 +1,228 @@
...
+  - name: kubeflow
+    root: kubeflow-master
+    uri: https://github.com/kubeflow/kubeflow/archive/master.tar.gz
+  - name: manifests
+    root: master/
+    uri: https://github.com/kubeflow/manifests/archive/master.tar.gz
+    # To get manifest at a PR:
+    #uri: https://github.com/kubeflow/manifests/archive/pull/235/head.tar.gz
+  appdir: /tmp/myapp2
+  componentParams:
```
jlewi: We don't want to use componentParams, packages, and components anymore.

We should be using KfDef.spec.Applications

The easiest way to convert them is to run kfctl and that will populate the app.yaml with the requisite KfDef.Spec.Applications

YOu could then copy and paste it here.

It might be sufficient to run kfctl generate

bootstrap/config/kfctl_gcp_iap_master.yaml
```diff
@@ -0,0 +1,205 @@
+# Please set project and email!
```
jlewi: How is this different from?
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_gcp_iap.0.6.yaml

Should we organize these in a GCP directory?

Per comment above can we use KfDef.Spec.Applications and not KfDef.Spec.componentParams/components/packages

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -528,14 +528,18 @@ func backfillKfDefFromInitOptions(kfdef *kfdefsv2.KfDef, options map[string]inte
...
 	// not so we always override the value with the command line flag.
-	if options[string(kftypes.USE_BASIC_AUTH)] != nil {
-		kfdef.Spec.UseBasicAuth = options[string(kftypes.USE_BASIC_AUTH)].(bool)
-	}
-	if options[string(kftypes.SKIP_INIT_GCP_PROJECT)] != nil {
-		kfdef.Spec.SkipInitProject = options[string(kftypes.SKIP_INIT_GCP_PROJECT)].(bool)
-	}
-	if options[string(kftypes.DELETE_STORAGE)] != nil && kfdef.Spec.Platform == kftypes.GCP {
-		kfdef.Spec.DeleteStorage = options[string(kftypes.DELETE_STORAGE)].(bool)
+	// TODO(lunkai): I think we shouldn't backfill bool flags when using --config
```
jlewi: I think you are right. If CONFIG is specified everything should be in the config and we no longer need to backfill the flags.

I noted in the issue that we should probably raise an error if additional flags are specified with --config.
We could do that in a follow on PR.

testing/kfctl/kfctl_go_test.py
```diff
@@ -91,6 +92,15 @@ def test_build_kfctl_go(app_path, project, use_basic_auth, use_istio):
     if os.getenv("PULL_PULL_SHA"):
       pull_manifests = "@" + os.getenv("PULL_PULL_SHA")
 
+  # We need to specify a valid email because
+  #  1. We need to create appropriate RBAC rules to allow the current user
+  #     to create the required K8s resources.
+  #  2. Setting the IAM policy will fail if the email is invalid.
+  email = util.run(["gcloud", "config", "get-value", "account"])
```
jlewi: Does this work with service accounts?

testing/kfctl/kfctl_go_test.py
```diff
@@ -99,21 +109,22 @@ def test_build_kfctl_go(app_path, project, use_basic_auth, use_istio):
   # We don't run with retries because if kfctl init exits with an error
   # but creates app.yaml then rerunning init will fail because app.yaml
   # already exists. So retrying ends up masking the original error message
+  config_file = "kfctl_gcp_iap_master.yaml"
+  if use_basic_auth:
+    config_file = "kfctl_gcp_basic_auth_master.yaml"
+  with open(os.path.join(src_dir, "bootstrap/config", config_file), 'r') as f:
```
jlewi: Why not just take the full URL of the config file as an argument of kfctl_go_test.py?

kfctl supports https://github.com/hashicorp/go-getter syntax.
So we can specify a full absolute local path or http location.

testing/kfctl/kfctl_go_test.py
```diff
@@ -128,6 +139,12 @@ def test_build_kfctl_go(app_path, project, use_basic_auth, use_istio):
 
   verify_kubeconfig(project, zone, app_path)
 
+def filterSpartakus(spec):
+  for i, comp in enumerate(spec["components"]):
```
jlewi: Lets use KfDef.Spec.Applications

testing/workflows/components/kfctl_go_test.jsonnet
```diff
@@ -225,6 +225,7 @@ local dagTemplates = [
         "-s",
         "--use_basic_auth=" + params.useBasicAuth,
         "--use_istio=" + params.useIstio,
+        "--src_dir=" + srcDir,
```
jlewi: Specify the full path of the config file?

jlewi:
Thanks for picking this up!

testing/kfctl/kfctl_go_test.py
lluunn: done

testing/workflows/components/kfctl_go_test.jsonnet
lluunn: done

testing/kfctl/kfctl_go_test.py
lluunn: done

testing/kfctl/kfctl_go_test.py
lluunn: I think so. This is original code, I just move it.

bootstrap/config/kfctl_gcp_iap_master.yaml
lluunn: I thought we want one for 0.6?
And future change will be made to the master one.

bootstrap/config/kfctl_gcp_basic_auth_master.yaml
lluunn: done

bootstrap/config/kfctl_gcp_iap_master.yaml
jlewi: The config for 0.6 should now be on the v0.6-branch.

The reason I named it kfctl_gcp_iap.0.6.yaml was to distinguish it from some of the other configs in this directory which are for 0.5 and click to deploy.

I think on master we only want one KfDef YAML for gcp + IAP. So I think we should pick kfctl_gcp_iap_master.yaml or kfctl_gcp_iap.0.6.yaml and delete the other one.

Probably keep kfctl_gcp_iap_master.yaml since its been updated to use KfDef.Spec.Applications.

We could probably change the name and remove "master". The version can be conveyed by the branch its on; we don't need it in the name.

testing/kfctl/kfctl_go_test.py
jlewi: It still doesn't look like your taking the full URL.
The repos will be checked out into /mnt/src/....
So my expectation was that you kfctl_go_test.py would just take the full path of the config_file
e.g.
kfctl_go_test.py --config=/mnt/src/kubeflow/kubeflow/bootstrap/config/kfctl_gcp_basic_auth_master.yaml

testing/kfctl/kfctl_go_test.py
jlewi: Did you forget to push changes or does this code need to be updated?

It looks like you updated the specs to use KfDef.Spec.Applications but here you are still looking at spec[components].

It might be good to have filterSpartakus look at both spec["components"] and spec["applications"] this way we filter out spartakus regardless of what format the spec is using.

testing/workflows/components/kfctl_go_test.jsonnet
jlewi: Did you forget to push changes?
I was expecting something like
--config=/mnt/src/.../kfctl_gcp_iap.yaml

lluunn:
@jlewi I think you are looking at the older snapshot.
I can see my new changes

jlewi:
@lluunn Thanks; one last minor commit.
Since you are creating
kfctl_gcp_iap.yaml 

Should we delete kfctl_gcp_iap.0.6.yaml so we avoid confusion?

lluunn:
removed kfctl_gcp_iap.0.6.yaml

PR #3732, opened 2019-07-23
Fix metadata component name

zhenghuiwang:
https://github.com/kubeflow/metadata/issues/101

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3732)


zhenghuiwang:
I've deployed this PR and the overlay is applied correctly.
Can I get it approved? Thanks @lluunn

PR #3742, opened 2019-07-24
Paper-menu-dropdown is aligned left

### LINKED ISSUE #3698: Namespace selection UI is half hidden by navigation bar
KF 0.6, deployment on 7/19 from Master

KF main dashboard, namespace selector
When clicking on the namespace selector at the top of the page, the drop down is half way hidden by the left nav pane, making it impossible to see the names of the namespaces.

namespaceselect

avdaredevil:
## About
- Paper-Menu-Button had the dropdown aligned `right`. This would allow it to flow into the sidebar, the new config should not let this happen

### Meta:
/area centraldashboard
/area front-end
/priority p0
/assign @avdaredevil 
/cc @prodonjs @paveldournov

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3742)

### DIFF
--- a/components/centraldashboard/public/components/namespace-selector.js
+++ b/components/centraldashboard/public/components/namespace-selector.js
@@ -56,7 +56,7 @@ export class NamespaceSelector extends PolymerElement {
                     --paper-button-ink-color: var(--accent-color);
                 }
             </style>
-            <paper-menu-button no-overlap horizontal-align="right">
+            <paper-menu-button no-overlap horizontal-align="left">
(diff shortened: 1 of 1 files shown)


prodonjs:
Interesting. It only seems to be a problem when the number of items grows on a small display size but this does the trick.

/lgtm
/approve

I'm not sure of the process but we should make sure this gets cherry-picked into the v0.6-branch.

avdaredevil:
Awesome, yeah will do @prodonjs!

PR #3758, opened 2019-07-26
Multi-User Isolation 2.0 - Manage Users

avdaredevil:
## Features (some exist in #3503):
- Vaadin charts and paper-toast added
- Profile Controller promise handler created
- `/get-contributors` api route created
- Namespace icon added
- Manage Users page added
- Dashboard made aware of cluster admin status and `isolationMode`
- `Manage Users` sidebar entry only when isolationMode is multi
- New `empty` macro added to utilities mixin
- Paper-chip element added
- TSLint remove any warnings
- Updated failing api_tests.ts
- Corrected the fields being used to/from server for registration
- Corrected sidebar entries for Manage-Users
- Manage Users contrib action color now toggles based on status (error support)
- Rollback errors when successfully added contributor
- Paper-chip border-color updated
- Paper-chip ripple updated to accent color

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3758)

### DIFF
--- a/components/centraldashboard/app/api.ts
+++ b/components/centraldashboard/app/api.ts
@@ -47,7 +47,7 @@ export const ERRORS = {
   operation_not_supported: 'Operation not supported'
 };
 
-export const apiError = (a: {res: Response, error: string, code?: number}) => {
+export function apiError (a: {res: Response, error: string, code?: number}) {
   const {res, error} = a;
   const code = a.code || 400;
   return res.status(code).json({
@@ -58,7 +58,7 @@ export const apiError = (a: {res: Response, error: string, code?: number}) => {
 /**
  * Converts Workgroup Binding from Profile Controller to SimpleBinding
  */
-export const mapWorkgroupBindingToSimpleBinding = (bindings: WorkgroupBinding[]): SimpleBinding[] => {
+export function mapWorkgroupBindingToSimpleBinding (bindings: WorkgroupBinding[]): SimpleBinding[] {
   return bindings.map((n) => ({
     user: n.user.name,
     namespace: n.referredNamespace,
@@ -70,7 +70,7 @@ export const mapWorkgroupBindingToSimpleBinding = (bindings: WorkgroupBinding[])
  * Converts Kubernetes Namespace types to SimpleBinding to ensure
  * compatibility between identity-aware and non-identity aware clusters
  */
-export const mapNamespacesToSimpleBinding = (user: string, namespaces: V1Namespace[]): SimpleBinding[] => {
+export function mapNamespacesToSimpleBinding (user: string, namespaces: V1Namespace[]): SimpleBinding[] {
   return namespaces.map((n) => ({
     user,
     namespace: n.metadata.name,
@@ -81,7 +81,7 @@ export const mapNamespacesToSimpleBinding = (user: string, namespaces: V1Namespa
 /**
  * Converts SimpleBinding to Workgroup Binding from Profile Controller
  */
-export const mapSimpleBindingToWorkgroupBinding = (binding: SimpleBinding): WorkgroupBinding => {
+export function mapSimpleBindingToWorkgroupBinding (binding: SimpleBinding): WorkgroupBinding {
   const {user, namespace, role} = binding;
   return {
     user: {
--- a/components/centraldashboard/app/api.ts
+++ b/components/centraldashboard/app/api.ts
@@ -1,21 +1,13 @@
-import {V1Namespace} from '@kubernetes/client-node';
-import {Router, Request, Response} from 'express';
+import {Router, Request, Response, NextFunction} from 'express';
 
-import {Binding as WorkgroupBinding, DefaultApi} from './clients/profile_controller';
+import {DefaultApi} from './clients/profile_controller';
 import {KubernetesService, PlatformInfo} from './k8s_service';
 import {Interval, MetricsService} from './metrics_service';
-import {ContributorAPI} from './api_contributors';
-
-interface SimpleBinding {
-  namespace: string;
-  role: 'owner' | 'contributor';
-  user: string;
-}
-
-interface WorkgroupInfo {
-  namespaces: SimpleBinding[];
-  isClusterAdmin: boolean;
-}
+import {
+  ContributorAPI,
+  mapNamespacesToSimpleBinding,
+  SimpleBinding,
+} from './api_contributors';
 
 interface EnvironmentInfo {
   namespaces: SimpleBinding[];
@@ -24,28 +16,11 @@ interface EnvironmentInfo {
   isClusterAdmin: boolean;
 }
 
-interface HasWorkgroupResponse {
-  user: string;
-  hasAuth: boolean;
-  hasWorkgroup: boolean;
-}
-
-
-export type SimpleRole = 'owner'| 'contributor';
-export type WorkgroupRole = 'admin' | 'edit';
-export type Role = SimpleRole | WorkgroupRole;
-export const roleMap: ReadonlyMap<Role, Role> = new Map([
-  ['admin', 'owner'],
-  ['owner', 'admin'],
-  ['edit', 'contributor'],
-  ['contributor', 'edit'],
-]);
-
 export const ERRORS = {
   operation_not_supported: 'Operation not supported'
 };
 
-export function apiError (a: {res: Response, error: string, code?: number}) {
+export function apiError(a: {res: Response, error: string, code?: number}) {
   const {res, error} = a;
   const code = a.code || 400;
   return res.status(code).json({
@@ -53,57 +28,19 @@ export function apiError (a: {res: Response, error: string, code?: number}) {
   });
 }
 
-/**
- * Converts Workgroup Binding from Profile Controller to SimpleBinding
- */
-export function mapWorkgroupBindingToSimpleBinding (bindings: WorkgroupBinding[]): SimpleBinding[] {
-  return bindings.map((n) => ({
-    user: n.user.name,
-    namespace: n.referredNamespace,
-    role: roleMap.get(n.roleRef.name as Role) as SimpleRole,
-  }));
-}
-
-/**
- * Converts Kubernetes Namespace types to SimpleBinding to ensure
- * compatibility between identity-aware and non-identity aware clusters
- */
-export function mapNamespacesToSimpleBinding (user: string, namespaces: V1Namespace[]): SimpleBinding[] {
-  return namespaces.map((n) => ({
-    user,
-    namespace: n.metadata.name,
-    role: roleMap.get('edit') as SimpleRole,
-  }));
-}
-
-/**
- * Converts SimpleBinding to Workgroup Binding from Profile Controller
- */
-export function mapSimpleBindingToWorkgroupBinding (binding: SimpleBinding): WorkgroupBinding {
-  const {user, namespace, role} = binding;
-  return {
-    user: {
-      kind: 'User',
-      name: user,
-    },
-    referredNamespace: namespace,
-    roleRef: {
-      kind: 'ClusterRole',
-      name: roleMap.get(role) as WorkgroupRole,
-    }
-  };
+export function NO_ROUTES(req: Request, res: Response, next: NextFunction) {
+  next();
 }
 
 export class Api {
   private platformInfo: PlatformInfo;
-  private contribApi: ContributorAPI;
 
   constructor(
-    private k8sService: KubernetesService,
-    private profilesService: DefaultApi,
-    private metricsService?: MetricsService) {
-      this.contribApi = new ContributorAPI(this.profilesService);
-    }
+      private k8sService: KubernetesService,
+      private profilesService: DefaultApi,
+      private contribApi?: ContributorAPI,
+      private metricsService?: MetricsService,
+    ) {}
 
   /** Retrieves and memoizes the PlatformInfo. */
   private async getPlatformInfo(): Promise<PlatformInfo> {
@@ -114,29 +51,15 @@ export class Api {
   }
 
   /**
-   * Retrieves WorkgroupInfo from Profile Controller for the given user.
-   */
-  private async getWorkgroupInfo(user: User.User): Promise<WorkgroupInfo> {
-    const [adminResponse, bindings] = await Promise.all([
-      this.profilesService.v1RoleClusteradminGet(user.email),
-      this.profilesService.readBindings(user.email),
-    ]);
-    const namespaces = mapWorkgroupBindingToSimpleBinding(
-      bindings.body.bindings || []
-    );
-    return {
-      isClusterAdmin: adminResponse.body,
-      namespaces,
-    };
-  }
-
-  /**
    * Builds EnvironmentInfo for the case with identity awareness
    */
   private async getProfileAwareEnv(user: User.User): Promise<EnvironmentInfo> {
     const [platform, {namespaces, isClusterAdmin}] = await Promise.all([
       this.getPlatformInfo(),
-      this.getWorkgroupInfo(user),
+      ContributorAPI.getWorkgroupInfo(
+        this.profilesService,
+        user,
+      ),
     ]);
     return {user: user.email, platform, namespaces, isClusterAdmin};
   }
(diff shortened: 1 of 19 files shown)


components/centraldashboard/public/components/manage-users-view.pug
```diff
@@ -0,0 +1,35 @@
+h1 Manage Contributors
+iron-ajax#AddContribAjax(method='POST', url='/api/add-contributor', on-response='handleContribCreate', on-error='handleContribCreate')
```
prodonjs: This method doesn't exist on the server.

avdaredevil:
@prodonjs Can we merge this?

prodonjs:
I haven't looked at this one in as much detail. I'll try to take a look at the diffbase from #3503 tomorrow and see if if there's any feedback. Again, I'm OK approving as long as the project leads are on-board given the limited test coverage and time spent in UAT.

components/centraldashboard/app/api.ts
```diff
@@ -23,13 +29,42 @@ interface CreateProfileRequest {
...
+
 interface HasWorkgroupResponse {
+  user: string;
   hasAuth: boolean;
   hasWorkgroup: boolean;
 }
 
-const OPERATION_NOT_SUPPORTED = {
-  error: 'Operation not supported'
+export const roleMap = {
```
prodonjs: I think you might be able to use an Enum here since it provides the reverse mapping automatically.

https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings

components/centraldashboard/app/api.ts
```diff
@@ -84,26 +122,73 @@ export class Api {
...
+    apiError({res, code, error});
+  }
+
+  async getContributors(namespace: string) {
+    const {body} = await this.profilesService
+      .readBindings(undefined, namespace);
+    const users = this.mapWorkgroupBindingToSimpleBinding(body.bindings)
+      .filter((b) => b.role === 'contributor')
+      .map((b) =>
+        b.user
```
prodonjs: Can this be moved up to be on the same line as `.map()` call?

components/centraldashboard/app/api.ts
```diff
@@ -181,33 +268,170 @@ export class Api {
...
+                  res,
+                  msg: 'Unexpected error creating profile',
+                  err,
+                });
+              }
+            })
+        .post(
+              '/add-contributor/:namespace',
+              async (req: Request, res: Response) => {
+                const {namespace} = req.params;
```
prodonjs: It looks like the logic for add and remove contributor starts out very similarly. It would probably be good to move that to a helper method. Also, maybe it would make sense to have use the `route` method to group `get`, `post`, and `delete` methods under a common path for contributors.

components/centraldashboard/app/api.ts
```diff
@@ -181,33 +268,170 @@ export class Api {
...
+                }
+                let errIndex = 0;
+                try {
+                  const binding = this.mapSimpleBindingToWorkgroupBinding({
+                    user: contributor,
+                    namespace,
+                    role: 'contributor',
+                  });
+                  const {headers} = req;
+                  delete headers['content-length'];
```
prodonjs: I see what you meant about the content-length if you were taking the full headers from the incoming request. I assume you don't need anything other than the authentication headers. Couldn't you just take the `req.user.auth` property which extracted the original headers?

components/centraldashboard/app/api.ts
```diff
@@ -181,33 +268,170 @@ export class Api {
...
+                  await this.profilesService
+                    .createBinding(binding, {headers});
+                  errIndex++;
+                  const users = await this.getContributors(namespace);
```
prodonjs: Is it necessary to reload the contributors? If the action succeeds, the UI would be responsible for adding/removing the item from view right?

components/centraldashboard/public/components/manage-users-view.pug
avdaredevil: Fixed, outdated

components/centraldashboard/app/api.ts
avdaredevil: So I completed the entire refactor, just to realize reverse translations are not supported for `string` valued `enums` leading to a larger implementation to get reverse translations by leveraging separately defined functions.

components/centraldashboard/app/api.ts
avdaredevil: Done

components/centraldashboard/app/api.ts
avdaredevil: Done!

components/centraldashboard/app/api.ts
avdaredevil: This is great because instead of the UI updating just the one contributor, it also syncs the rest of the changes in case other changes were made. Every operation to your contributor list makes sure you're in sync with the absolute truth. It also leads to a simpler handler logic as the response for all the APIs are directly property bound to the view.

components/centraldashboard/app/api.ts
avdaredevil: Sure, I'll go ahead and do that.

components/centraldashboard/app/api.ts
avdaredevil: I tried a few approaches since this comment but I keep running into type issues. I've reverted again, feel free to share a snippet that works for you

components/centraldashboard/app/api_contributors.ts
```diff
@@ -0,0 +1,184 @@
...
+
+// From: https://www.w3resource.com/javascript/form/email-validation.php
+const EMAIL_RGX = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/;
+
+/**
+ * Handles an exception in an async block and converts it to a JSON
+ * response sent back to client
+ */
+// tslint:disable-next-line: no-any
+const surfaceProfileControllerErrors = (info: {res: Response, msg: string, err: any}) => {
```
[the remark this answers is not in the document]
avdaredevil: Done!

components/centraldashboard/app/api.ts
prodonjs: Looking at the kfam code, it definitely seems like they only use the [email header](https://github.com/kubeflow/kubeflow/blob/f3b8d2337e8b29ac213fbf94e29a3783c76b5a90/components/access-management/kfam/api_default.go#L227). I'd be reluctant to pass other headers in unnecessarily. Obviously I think you saw already the difficulty in debugging an issue with just `Content-Length`. I'd hate to see a scenario like that appear in a production environment due to some unforeseen interaction with the Kfam server's treatment of the headers.

components/centraldashboard/app/api.ts
prodonjs: You could try something like this:
```
type Role = 'admin'|'owner'|'edit'|'contributor';
const ROLES: ReadonlyMap<Role, Role> = new Map([
  ['admin', 'owner'],
  ['owner', 'admin'],
  ['edit', 'contributor'],
  ['contributor', 'edit'],
]);
```
Then, you can get just use something like `ROLES.get(n.roleRef.name as Role)`. To me, using the built-in types is always easier and lessens the test footprint.

components/centraldashboard/app/api_contributors.ts
```diff
@@ -0,0 +1,184 @@
...
+        console.log('Recieved a call');
+        if (!req.user.hasAuth) {
+            return apiError({
+                res,
+                code: 405,
+                error: ERRORS.operation_not_supported,
+            });
+        }
+        
+        console.log('No Error');
```
prodonjs: These seem like they were added for testing. I'd probably drop them in the final version.

avdaredevil: The only debug statements I forgot to add todo's for, thanks!

avdaredevil: Removed!

components/centraldashboard/app/api_contributors.ts
```diff
@@ -33,152 +33,147 @@ const surfaceProfileControllerErrors = (info: {res: Response, msg: string, err:
     const {res, msg, err} = info;
     const code = (err.response && err.response.statusCode) || 400;
     const error = err.body || msg;
-    console.log(`${msg} ${error}`);
+    console.error(`${msg} ${error}`);
```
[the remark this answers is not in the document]
avdaredevil: I've updated it to make more sense, and commented it as well

components/centraldashboard/app/api.ts
```diff
@@ -233,8 +234,8 @@ export class Api {
               res.json(await this.k8sService.getEventsForNamespace(
                   req.params.namespace));
             })
-        .use('/workgroup', contributorAPI(this.profilesService))
-        .use((req: Request, res: Response, next: NextFunction) => {
+        .use('/workgroup', this.contribApi.routes())
+        .use((req: Request, res: Response) => {
           res.status(404).json(`Could not find the route you're looking for`);
```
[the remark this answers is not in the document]
avdaredevil: That was a mistake fixed! All errors should be standard.

components/centraldashboard/app/api.ts
```diff
@@ -24,86 +16,31 @@ interface EnvironmentInfo {
...
-      kind: 'User',
-      name: user,
-    },
-    referredNamespace: namespace,
-    roleRef: {
-      kind: 'ClusterRole',
-      name: roleMap.get(role) as WorkgroupRole,
-    }
-  };
+export function NO_ROUTES(req: Request, res: Response, next: NextFunction) {
```
prodonjs: Any reason for not using camel-casing for the function here? Also, given that it's only used in one place and is such a trivial function I don't see a need to have it separated and exported.

components/centraldashboard/app/api_contributors.ts
```diff
@@ -32,13 +98,30 @@ const EMAIL_RGX = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/;
...
+    console.error(msg+(devError?` ${devError}`:''));
+    apiError({res, code, error: devError || msg});
 };
 
 export class ContributorAPI {
     constructor(private profilesService: DefaultApi) {}
+    /**
+     * Retrieves WorkgroupInfo from Profile Controller for the given user.
+     */
+    static async getWorkgroupInfo(profilesService: DefaultApi, user: User.User): Promise<WorkgroupInfo> {
```
prodonjs: Why make this a static method that needs to have the `profilesService` passed into it? If it's a normal instance method `profilesService` is already a private member. It also looks like the signature could just take the user's email as the second parameter.

components/centraldashboard/app/api_contributors.ts
```diff
@@ -94,6 +177,24 @@ export class ContributorAPI {
...
+        .get(
+            '/exists',
+            async (req: Request, res: Response) => {
+                const response: HasWorkgroupResponse = {
+                    hasAuth: req.user.hasAuth,
+                    user: req.user.username,
+                    hasWorkgroup: false,
+                };
+                if (req.user.hasAuth) {
+                    const workgroup = await ContributorAPI.getWorkgroupInfo(
```
prodonjs: See my previous comment about the static method. Unless there is something that I am missing, it seems like it could easily be converted to an instance method.

components/centraldashboard/app/api_contributors.ts
```diff
@@ -1,28 +1,94 @@
...
-    mapWorkgroupBindingToSimpleBinding,
-    mapSimpleBindingToWorkgroupBinding,
     ERRORS,
 } from './api';
 
+// From: https://www.w3resource.com/javascript/form/email-validation.php
+const EMAIL_RGX = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/;
+
 // Valid actions for handling a contributor
 type CONTRIBUTOR_ACTIONS = 'create' | 'remove';
```
prodonjs: Type names should follow the normal naming rules and use UpperCamelCase.

http://go/tsstyle#identifiers

components/centraldashboard/app/api_test.ts
```diff
@@ -371,7 +379,7 @@ describe('Dashboard API', () => {
         [header]: `${prefix}test@testdomain.com`,
       };
       const response = await sendTestRequest(url, headers, 405, 'post');
-      expect(response).toEqual({error: 'Unexpected error creating profile'});
+      expect(JSON.stringify(response)).toEqual(JSON.stringify({error: 'Unexpected error creating profile'}));
```
prodonjs: The stringify calls are unnecessary as `.toEqual` does a deep equality test.

components/centraldashboard/app/server.ts
```diff
@@ -34,13 +35,15 @@ async function main() {
   const k8sService = new KubernetesService(new KubeConfig());
   const metricsService = await getMetricsService(k8sService);
   console.info(`Using Profiles service at ${profilesServiceUrl}`);
-  const profilesServices = new DefaultApi(profilesServiceUrl);
+  const profilesService = new DefaultApi(profilesServiceUrl);
+  const contribApi = new ContributorAPI(profilesService);
+
```
prodonjs: Remove extra space.

components/centraldashboard/app/api_contributors.ts
```diff
@@ -32,13 +98,30 @@ const EMAIL_RGX = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/;
...
-    const error = err.body || msg;
-    console.error(`${msg} ${error}`);
-    apiError({res, code, error});
+    const devError = err.body || '';
+    // Msg is the developer reason of what happened, devError is the technical details as to why
+    console.error(msg+(devError?` ${devError}`:''));
+    apiError({res, code, error: devError || msg});
 };
 
 export class ContributorAPI {
```
prodonjs: *nit* This class should probably be named `ContributorApi` to be consistent with Google TS naming guidelines.

components/centraldashboard/public/components/registration-page.css
```diff
@@ -15,7 +15,7 @@
 }
 h1,h2,h3 {margin: 0;font-family: Google Sans;color: var(--heading-color);font-weight: 400}
```
prodonjs: This is tough to read/maintain all being crammed onto the same lines. Can you run beautifier over this file?

components/centraldashboard/public/components/registration-page.js
```diff
@@ -84,7 +85,7 @@ export class RegistrationPage extends utilitiesMixin(PolymerElement) {
             el.style.width = `${el.scrollWidth}px`;
         });
         this.sleep(2000).then(async () => {
-            this.dispatchEvent(new CustomEvent('flowcomplete'));
+            this.fire('flowcomplete');
```
prodonjs: It looks like `fire` was maintained for backwards compatibility with older versions of Polymer, but you override it with a custom definition. https://polymer-library.polymer-project.org/3.0/docs/devguide/events#custom-events.

Should the name differ just to avoid any confusion?

components/centraldashboard/app/api.ts
avdaredevil: I was leaving this exposed for future conditional routing, but yeah, for now, we can just keep it scoped to the code that exists.

components/centraldashboard/app/api_contributors.ts
avdaredevil: Please refer to [api.ts:55](https://github.com/kubeflow/kubeflow/pull/3758/files#diff-d5346240aaa21ec8e07782abf283707aR59)

components/centraldashboard/app/api_contributors.ts
avdaredevil: It is also used in api.ts hence is a static method that can be used outside of an instantiation

components/centraldashboard/app/api_contributors.ts
avdaredevil: Gotcha

components/centraldashboard/app/api_test.ts
avdaredevil: Fixed, that was from a test I was doing to see why a test was failing.

components/centraldashboard/app/server.ts
avdaredevil: Done

components/centraldashboard/public/components/registration-page.css
avdaredevil: Done!

components/centraldashboard/public/components/registration-page.js
avdaredevil: `.fire` -> `.fireEvent`

components/centraldashboard/app/api_contributors.ts
avdaredevil: Done

prodonjs:
I'm good with things to move this ahead.

/lgtm
/approve

You'll need to have this cherry-picked into the v0.6-branch, push a new image, and update the repository. My recommendation would be to tag the image with a semantic number (v0.6.x) that matches the release level similar to the images from previous versions.

avdaredevil:
Amazing, thanks @prodonjs. I'll update the make-win.ps1 script to allow for custom tagging via argument, so that should be more seamless.

PR #3775, opened 2019-07-29
update pipeline version to v0.1.25

timothysu:
this brings it up to the latest version tagged and released per https://github.com/kubeflow/pipelines/releases

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3775)

### DIFF
--- a/kubeflow/pipeline/pipeline.libsonnet
+++ b/kubeflow/pipeline/pipeline.libsonnet
@@ -1,11 +1,11 @@
 {
   // Parameters are intended to be late bound.
   params:: {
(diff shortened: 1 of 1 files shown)


IronPan:
@timothysu Thanks for the PR. Ksonnet is sunsetting. Is this something you are still using?

timothysu:
thank you @IronPan - we just upgraded to 0.6 so i believe it is not. with the upgrade to 0.6 though, the `upgrade_kfp.sh` script is broken. is there way to pull the latest release of pipelines in 0.6?

PR #3777, opened 2019-07-30
integrate new backend with source repo module

kunmingg:
1. Make source repo module to always try to reuse existing repo.
2. Integrate source repo module with handler logic.

Should we keep random folder name for git clone as extra layer protection?

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3777)

### DIFF
--- a/bootstrap/cmd/bootstrap/app/kfctlServer.go
+++ b/bootstrap/cmd/bootstrap/app/kfctlServer.go
@@ -83,12 +83,18 @@ func NewKfctlServer(appsDir string) (*kfctlServer, error) {
 // to the KfDef.
 func (s *kfctlServer) handleDeployment(r kfdefsv2.KfDef) (*kfdefsv2.KfDef, error) {
 	ctx := context.Background()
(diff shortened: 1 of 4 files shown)


jlewi:
What branch do we use? Does each app get its own branch?

I don't think we need a random folder name any more.

kunmingg:
@jlewi 
Now updated to use separate branch for every deployment

kunmingg:
One advantage to clone before init is we can pick up user's existing config if exists?

jlewi:
> One advantage to clone before init is we can pick up user's existing config if exists?
@kunmingg I think that makes sense. 

Can we do that though based on how we use SourceRepo  in kfctlserver rather than making that assumption in SourceRepo?

So can we make SourceRepo support the following cases

1. Dir exists but is not a git repo
    * SourceRepo runs git init and adds the repo as a remote
1. Dir exists and is a git repo
     * SourceRepo adds the repo as a remote
1. Dir doesn't exist
      * SourceRepo clones the repo

Issue #3709, opened 2019-07-22
Kfctl Design Doc for v1beta1


jlewi:
We need an updated design doc for kfctl.

Here are some key features we should capture

1. Adopting K8s controller reconcile semantics (see #3566)
1. Extensibility model i.e. plugin support
1. Proposal for KfDef v1beta1
1. What the desired CLI semantics are (see #3518)
1. Some discussion about code location and organization
    * There had been some discussion about moving it to its own repo

One of the key motivations is to focus on what we need to get kfctl and KfDef 1.0 in the Q4 release.

That means we'd like to have a v1beta1 KfDef in 0.7

jlewi:
@kkasravi @yanniszark either of you interested?

kkasravi:
I would be more than happy to contribute to the design doc, including moving to kustomize 3.X+, removing v2.zip, and any other issues I've been involved in.

yanniszark:
I'd love to contribute to this.
I'm very excited by the work done so far in kfctl. :smile:

jlewi:
@yanniszark Do you want to start a doc?

yanniszark:
@jlewi sure, I will start one in the Kubeflow Drive.

jlewi:
@yanniszark Any update on this?

yanniszark:
I am coming back from PTO this week.
A draft should be out very soon, most likely by the end of this week.

kkasravi:
>copied over from https://github.com/kubeflow/kubeflow/pull/3783#discussion_r312686204
>from @jlewi
>Thanks for raising this @kkasravi ; I didn't realize we were exposing this as a new top level >command.

>@yanniszark @gabrielwen I'd really prefer not to introduce any more user visible changes (e.g. new command); so if we do need to go with a PostApply hook lets implement it so as not be user visible.

@jlewi - Clarification from me: PostApply is not a new command (there is no kfctl/cmd/postapply.go - but it's called from the apply command (kfctl/cmd/apply.go). An alternative would be to call it lower down the stack like - like from coordinator's apply.

yanniszark:
The initial draft of the design doc is here:
https://docs.google.com/document/d/1kviNOdjoBPeoJ56KBvMKWq7SzrSN5rFUQoULK8muD-M/edit?usp=sharing

cc @jlewi @kkasravi

kkasravi:
Thanks @yanniszark !

Did you want comments within the document?

yanniszark:
[...]

PR #3783, opened 2019-07-30
Add PostApply to KfApp interface

gabrielwen:
Adds new PostApply function to the interface for necessary steps after `Apply`.

Also moves PodDefault configurations to this new interface.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3783)

### DIFF
--- a/bootstrap/cmd/kfctl/cmd/apply.go
+++ b/bootstrap/cmd/kfctl/cmd/apply.go
@@ -47,6 +47,10 @@ var applyCmd = &cobra.Command{
 		if applyErr != nil {
 			return fmt.Errorf("couldn't apply KfApp: %v", applyErr)
 		}
+		postApplyErr := kfApp.PostApply(resource)
+		if postApplyErr != nil {
+			return fmt.Errorf("couldn't run post apply on KfApp: %v", postApplyErr)
+		}
 		return nil
 	},
 	ValidArgs: []string{"all", "platform", "k8s"},
--- a/bootstrap/pkg/apis/apps/group.go
+++ b/bootstrap/pkg/apis/apps/group.go
@@ -101,6 +101,7 @@ const (
 // They all implement the API below
 //
 type KfApp interface {
+	PostApply(resources ResourceEnum) error
 	Apply(resources ResourceEnum) error
 	Delete(resources ResourceEnum) error
 	Generate(resources ResourceEnum) error
--- a/bootstrap/pkg/kfapp/aws/aws.go
+++ b/bootstrap/pkg/kfapp/aws/aws.go
@@ -653,6 +653,10 @@ func (aws *Aws) setAwsPluginDefaults() error {
 	return nil
 }
 
+func (aws *Aws) PostApply(resources kftypes.ResourceEnum) error {
+	return nil
+}
+
 // Apply create eks cluster if needed, bind IAM policy to node group roles and enable cluster level configs.
 // Remind: Need to be thread-safe: this entry is share among kfctl and deploy app
 func (aws *Aws) Apply(resources kftypes.ResourceEnum) error {
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -724,6 +724,10 @@ func (kfapp *coordinator) GetPlugin(name string) (kftypes.KfApp, bool) {
 	return r, ok
 }
 
+func (kfapp *coordinator) PostApply(resources kftypes.ResourceEnum) error {
+	return nil
+}
+
 func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
 	platform := func() error {
 		if kfapp.KfDef.Spec.Platform != "" {
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -815,46 +815,14 @@ func (kfapp *coordinator) Apply(resources kftypes.ResourceEnum) error {
 		return nil
 	}
 
-	gcpAddedConfig := func() error {
-		if kfapp.KfDef.Spec.Email == "" || kfapp.KfDef.Spec.Platform != kftypes.GCP {
-			return nil
-		}
-
-		if p, ok := kfapp.Platforms[kfapp.KfDef.Spec.Platform]; !ok {
-			return &kfapis.KfError{
-				Code:    int(kfapis.INTERNAL_ERROR),
-				Message: "Platform GCP specified but not loaded.",
-			}
-		} else {
-			gcp := p.(*gcp.Gcp)
-			p, err := gcp.GetPluginSpec()
-			if err != nil {
-				return err
-			}
-			if *p.EnableWorkloadIdentity {
-				return gcp.SetupDefaultNamespaceWorkloadIdentity()
-			} else {
-				return gcp.ConfigPodDefault()
-			}
-		}
-	}
-
 	switch resources {
 	case kftypes.ALL:
 		if err := platform(); err != nil {
 			return err
 		}
-		if err := k8s(); err != nil {
-			return err
-		}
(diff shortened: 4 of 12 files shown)


gabrielwen:
@yanniszark This should serve as short-term solution before Reconcile pattern kfctl is implemented.  Let me know what you think.

@jlewi do you think we can keep this interface before #3709 is implemented?

jlewi:
@gabrielwen What problem is this supposed to solve before #3709? Do we have a pressing need for a short term solution?

If we need a cheaper short term solution; how much effort is it saving us to implement a postapply hook as opposed to just implementing reconcile?

gabrielwen:
@jlewi as discussed, this is a faster way of providing a hook after Apply.  We only need this PR for a short-term solution while kfctl 1.0 might need way more time to be finished.

gabrielwen:
@yanniszark could you take a look since this is something you asked?  thanks,

jlewi:
I made one suggestion about how to implement the post-apply hook. So if you want to go that route we can. 

> We only need this PR for a short-term solution while kfctl 1.0 might need way more time to be finished.

Do we really need to wait for 1.0? 

Suppose we did something like the following

1. Apply runs a loops which repeatedly calls apply on the platform and the k8s kfApp
1. The apply loop exits based on some conditions set in KFDef.Status

If the loop terminates if KFDef.Status.Conditions is empty then that would preserve the existing behavior.

We could then consider a number of different options to allow the Apply loop to keep running; e.g. we could define Finished and Running conditions for both the platform and K8s and we could keep iterating until we see Finished conditions for both the platform and K8s

bootstrap/cmd/kfctl/cmd/apply.go
```diff
@@ -47,6 +47,10 @@ var applyCmd = &cobra.Command{
 		if applyErr != nil {
 			return fmt.Errorf("couldn't apply KfApp: %v", applyErr)
 		}
+		postApplyErr := kfApp.PostApply(resource)
```
kkasravi: Thinking about this approach, adding a new interface method 'PostApply' and hoisting its call all the way up to the cobra cmd level seems perhaps heavy-handed. PostApply is driven by the need to call the platform to create a DefaultPod after the package manager is done? I say heavy-handed because it seems like the only implementation of PostApply is Gcp but the cost is having all implementations stub-out the method.

One suggestion would be to return a continuation from Apply shown below
```
type Afterwards func() error
type KfApp interface {
	Apply(resources ResourceEnum) (Afterwards, error)
	Delete(resources ResourceEnum) error
	Generate(resources ResourceEnum) error
	Init(resources ResourceEnum) error
}
```

and just have coordinator call this if it's not nil at the end of its Apply.
Gcp would return a continuation that looks similar to [gcpAddedConfig](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/coordinator/coordinator.go#L764)
but much simpler and defined in Gcp.

The gcpAddedConfig in coordinator.Apply
```
	gcpAddedConfig := func() error {
		if kfapp.KfDef.Spec.Email == "" || kfapp.KfDef.Spec.Platform != kftypesv3.GCP {
			return nil
		}

		if p, ok := kfapp.Platforms[kfapp.KfDef.Spec.Platform]; !ok {
			return &kfapis.KfError{
				Code:    int(kfapis.INTERNAL_ERROR),
				Message: "Platform GCP specified but not loaded.",
			}
		} else {
			gcp := p.(*gcp.Gcp)
			p, err := gcp.GetPluginSpec()
			if err != nil {
				return err
			}
			if *p.EnableWorkloadIdentity {
				return gcp.SetupDefaultNamespaceWorkloadIdentity()
			} else {
				return gcp.ConfigPodDefault()
			}
		}
	}
```
Is moved to Gcp in its Apply method
```
	continuation := func() error {
		p, err := gcp.GetPluginSpec()
		if err != nil {
			return err
		}
		if *p.EnableWorkloadIdentity {
			return gcp.SetupDefaultNamespaceWorkloadIdentity()
		} else {
			return gcp.ConfigPodDefault()
		}
	}
        return continuation
```

So the coordinator could just call this at this end of its Apply or could maintain a stack like so:
```
  cb, err := platform.Apply()
  if err != nil ...
  if cb != nil coordinator.push(cb)

  cb, err := packagemanager.Apply()
  if err != nil ...
  if cb != nil coordinator.push(cb)

  while cb = coordinator.pop(); cb != nil {
    cb(resource)
  }
```

yanniszark:
@gabrielwen thanks for following up on this! :)

I also agree with @kkasravi and @jlewi that enforcing implementation of PostApply() across all platforms is a bit excessive.
I also support @jlewi's suggestion:

> Then you can test whether a KfApp implements that interface and then only call the PostApply function if it does?

1. Declare the PostApplyer() interface.
2. In the Apply() function of coordinator, check if the platform implements the PostApplyer() interface.
```go
_, ok := p.(PostApplyer)
	if ok {
		err := p.PostApply()
	}
```
3. Implement the interface for GCP.

Any other platform that needs this functionality can just define a PostApply() function.
This way, the changes became much more localized.

gabrielwen:
> @gabrielwen thanks for following up on this! :)
> 
> I also agree with @kkasravi and @jlewi that enforcing implementation of PostApply() across all platforms is a bit excessive.
> I also support @jlewi's suggestion:
> 
> > Then you can test whether a KfApp implements that interface and then only call the PostApply function if it does?
> 
> 1. Declare the PostApplyer() interface.
> 2. In the Apply() function of coordinator, check if the platform implements the PostApplyer() interface.
> 
> ```go
> _, ok := p.(PostApplyer)
> 	if ok {
> 		err := p.PostApply()
> 	}
> ```
> 
> 1. Implement the interface for GCP.
> 
> Any other platform that needs this functionality can just define a PostApply() function.
> This way, the changes became much more localized.

@yanniszark seems like you have started designing kfctl v1beta1.  If this new interface is going to live long, I can start with a primitive Reconcile-like interface and replace this PR.

@jlewi @kkasravi thoughts?

We can move the discussion to #3709 if you want to.

```go
// Use this to identify which runner is not recoverable.
type KfPermanentError struct {
  Message string
}
func (e *KfPermanentError) Error() string {
  return e.Message
}

// Use this interface to separate individual processes.
type StateRunner interface {
  func Run() error
}

// This is handler each cmd uses.
type CmdHandler struct {
  runners map[string]*StateRunner
}

func (ch CmdHandler) RegisterRunner(name string, runner *StateRunner) {
  ch.runners[name] = runner
}

func (ch *CmdHandler) runLoop() error {
  var results map[string]chan error

  for name, runner := range runners {
    results[name] = make(chan error)
    go func() {
      results[name] <- runner.Run()
    }
  }

  for name, ch := range results {
    e := <-ch
    if e != nil {
      return e
    }
  }
  return nil
}

func (ch *CmdHandler) Run() error {
  for {
    err := ch.runLoop()
    if err == nil {
      return nil
    }
    if _, ok := err.(*KfPermanentError); ok {
      logging.error("Some message for permanent error.")
      return err
    }
  }
  return nil  // or some error message.
}
```

bootstrap/cmd/kfctl/cmd/apply.go
jlewi: Thanks for raising this @kkasravi ; I didn't realize we were exposing this as a new top level command.

@yanniszark @gabrielwen I'd really prefer not to introduce any more user visible changes (e.g. new command); so if we do need to go with a PostApply hook lets implement it so as not be user visible.

jlewi:
@gabrielwen Discussing on #3709 sounds good to me.

@gabrielwen What are the immediate use cases where we need a short term solution? Is the immediate need a way to generalize how GCP is creating a default Profile so that other platforms can implement it?

gabrielwen:
@jlewi the only immediate need for the short term solution is because @yanniszark mentioned they might need it as well.

jlewi:
Thanks @gabrielwen.

It looks like @yanniszark shared his initial design doc. Let's take a look at that and then go from there.

PR #3801, opened 2019-08-01
[KUBEFLOW-3797] When copying files fails provide correct error message

holdenk:
Fix the error messages provided to clarify when the error is not being able to
read the input file.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3801)


holdenk:
Hey @richardsliu would you be comfortable giving this an ok-to-test?

PR #3837, opened 2019-08-07
To use the latest fairing version.

xauthulei:
In order to use new feature of the `fairing`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3837)


xauthulei:
It seems the test env have some issue which not caused by my changes. Thanks

PR #3856, opened 2019-08-09
Culling of Idle Jupyter Notebooks

### LINKED ISSUE #1394: Jupyter - Culling of Idle Pods
JupyterHub has a demo on culling idle pods using a script.
This is present as [an example on JupyterHub service](https://github.com/jupyterhub/jupyterhub/tree/master/examples/cull-idle)
I have a somewhat ghetto solution to implementing this within Kubeflow using configmap

We add the following configuration into the Config Map `jupyterhub-config`:
```
c.JupyterHub.services = [
    {
        'name': 'wget-cull-idle',
        'admin': True,
        'command': ['wget', 'https://raw.githubusercontent.com/jupyterhub/jupyterhub/master/examples/cull-idle/cull_idle_servers.py', '-N']
    },

    {
        'name': 'cull-idle',
        'admin': True,
        'command': ['python', 'cull_idle_servers.py', '--timeout=3600']
    }
]
```

The solution works by having JupyterHub start two managed services: one to fetch the `cull_idle_servers.py` file and one to start it. Hope this is useful!

I'm actually wondering if we should start a specific section in the documentation that deals with configuration easily done through configmap (like this one).
We could talk about things like form customisation, adding more field (like node selection via labels) etc. I feel there's a lot of potential there.
I will be glad to contribute to that documentation.

kimwnasptd:
This PR provides a barebones implementation for Notebooks culling.
A culling library is added, which contains logic for checking if a Notebook is idle for long period based on the `/api/status` endpoint of each Notebook Server.

The Controller will use this library to periodically check if any Notebook needs to be culled. If it does, then a `kubeflow-resource-stopped: $time.Now()` annotation will be added to the Notebook's CR. The controller will set the Replicas count to 0 for any Notebook that has this annotation.

After this PR is merged, culling will be disabled by default. This means that the controller will not stop any Notebook by himself, but will still stop one if the annotation is set. We can enable it by setting the `ENABLE_CULLING="true"` ENV var in the Notebooks Controller.

This culling behavior can be configured with the following ENV Vars:
```bash
ENABLE_CULLING="false"
IDLE_TIME="1440"  #  minutes, One day
CULLING_CHECK_PERIOD="1"  # minutes
CLUSTER_DOMAIN="cluster.local"
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3856)


kimwnasptd:
It is ready for review
/cc @lluunn 
/cc @jlewi 

Also, from the presubmit tests, I see the following error on `checkout`:
```
fatal: could not create leading directories of '/mnt/test-data-volume/kubeflow-presubmit-unittests-3856-c622c09-6320-73fa/src/kubeflow/kubeflow': No space left on device
```

jlewi:
Thanks @kimwnasptd this is very exciting

A couple quick questions/comments

1. Can we add unittests?
1. How do users control the culling behavior? Do we need to update the notebook resource to add options?

jlewi:
@lluunn Do you want to review this?

/assign @lluunn

jlewi:
@kimwnasptd Do we want to use requests? Don't notebooks use websockets? Does that get counted as multiple requests?

Why use the requests rather than the Jupyter's last_activity signal
https://github.com/kubeflow/kubeflow/issues/1394#issuecomment-425215219

Does the Jupyter signal provide a higher fidelity signal for whether a notebook is idle?

kimwnasptd:
@jlewi 
> 1. Can we add unittests?

Are there any existing unittests for the controller? I couldn't find them with a quick glance over the repo

> 2. How do users control the culling behavior? Do we need to update the notebook resource to add options?
> 

The culling behavior can be parameterized with the following ENV variables in the Notebook Controller's Pod. The values with = are the ones that will be used if the ENV vars are not present.
```bash
DEFAULT_PROMETHEUS_SVC = "prometheus.istio-system.svc.cluster.local:9090"
DEFAULT_IDLE_TIME = "1440" 
DEFAULT_CULLING_CHECK_PERIOD = "1"
DEFAULT_ENABLE_CULLING = "false"
```
[ Any time unit is measured in minutes. I set the DEFAULT_IDLE_TIME to 1 day. ]

Would we need to fine grain the culling more? Would we need to have different culling times for different Notebooks?

> @kimwnasptd Do we want to use requests? Don't notebooks use websockets? Does that get counted as multiple requests?
> Why use the requests rather than the Jupyter's last_activity signal
#1394 (comment)
> Does the Jupyter signal provide a higher fidelity signal for whether a notebook is idle?

[ To be honest, I forgot about your comment for `last_activity` and got caught up thinking how to make the culler reusable for the Viewer CR :slightly_smiling_face: ]

From the spec I see that`last_activity` tracks the requests to the REST API (`<notebook-url>/api/*`) and the kernel's activity. It will most probably track the HTTP handshake request to initialize the WebSocket, but I couldn't find if it would also keep track of the traffic send via the WebSocket protocol afterwards.

The `last_activity` endpoint would definitely be more accurate for culling Notebooks, since it also keeps track of the kernel's state. But, I would also like to preserve the generalized `ResourceNeedsCuling` function which works for arbitrary resources. 

So I'm going to:
1. Add a more specialized `NotebookNeedsCulling` function that will use the Jupyter Notebook's API for determining if the Notebook needs to be culled
2.  Keep the `ResourceNeedsCulling` as a default/generalized utility for other controllers that might need to have culling

@jlewi WDYT?

jlewi:
My apologies as well I'd forgotten about it as well.

Having a separate request based culler makes sense to me. One question I have is whether we could potentially use knative for that and configuring it to do 0-1 scaling? Might be worth chatting with @ellis-bigelow and the knative folks about that.

> Would we need to fine grain the culling more? Would we need to have different culling times for different Notebooks?

I think so. I think users may have different preferences and it may differ a lot depending on what the resource foot print is.

If you have a notebook just using 2 CPU then being able to come back to it everyday without losing state might be worth more than saving 2 CPU or maybe require a longer timeout. 

Whereas if your using an expensive GPU you probably want more aggressive culling.

Also when we roll this out it probably makes sense to let users turn it on individually until we're certain it works as expected.

For unittests if there isn't existing unittests can we still add some for new code? For example the culling library? If we take a test driven approach maybe we can write the code in a way that it is easy to unittest? e.g. if you have a function 

```
func(num_requests) bool {
....
}
```

Then its easy to write a unittest for this without having to mock out prometheus or the K8s APIServer.

kimwnasptd:
lets scope the context of this PR and set up the next steps. 

For the Notebooks, the culling should be based on the `last_activity` in `/api/status` of the servers. I'll also add unit tests for the culling library with follow up commits in this PR.

The culling will be disabled by default on the controller so we could have this basic implementation for now and add more features with small follow up PRs (like fine grained culling).

Next topics we need to address would be:
1. For a more general requests based culler should we use knative or rely on ISTIO/prometheus to get the metrics?
2. For having more fine-grained culling time of Notebooks should we add this info on the Notebook's spec or should we use annotations on the CRs?

Lets discuss them on existing/new issues and not block this PR with them.
@jlewi does this sound good?

jlewi:
@kimwnasptd that plan sounds good to me.

Is this PR ready for review per the plan above?

kimwnasptd:
Not yet, working on the commits. Will give a heads up once it's ready for review

kimwnasptd:
@jlewi @lluunn the PR is ready for review

components/notebook-controller/pkg/culler/culler.go
```diff
@@ -0,0 +1,194 @@
...
+
+type NotebookStatus struct {
+	Started      string `json:"started"`
+	LastActivity string `json:"last_activity"`
+	Connections  int    `json:"connections"`
+	Kernels      int    `json:"kernels"`
+}
+
+// Some Utility Fuctions
+func getEnvDef(variable string, default_val string) string {
```
lluunn: def means default? seems a little weird

components/notebook-controller/pkg/culler/culler.go
```diff
@@ -0,0 +1,194 @@
...
+	Timeout: time.Second * 10,
+}
+
+// The constants with name 'DEFAULT_{ENV_Var}' are the default values to be
+// used, if the respective ENV vars are not present.
+// All the time numbers correspond to minutes.
+const DEFAULT_IDLE_TIME = "1440" // One day
+const DEFAULT_CULLING_CHECK_PERIOD = "1"
+const DEFAULT_ENABLE_CULLING = "false"
+const STOP_ANNOTATION = "kubeflow-resource-stopped"
```
lluunn: Could you add some comment for the annotation (what does it mean, when will we add it)

components/notebook-controller/pkg/culler/culler.go
```diff
@@ -0,0 +1,194 @@
...
+
+// Stop Annotation handling functions
+func SetStopAnnotation(meta *metav1.ObjectMeta) {
+	if meta == nil {
+		log.Info("Error: Metadata is Nil. Can't set Annotations")
+		return
+	}
+
+	if meta.GetAnnotations() != nil {
+		annotations := meta.GetAnnotations()
```
lluunn: So this gets a copy of the map, and you need to call set again?
Then do you need to call set in the RemoveStopAnnotation below?

components/notebook-controller/pkg/culler/culler.go
kimwnasptd: Yes, on a second look it seems like I'm requesting the definition of the environment, which doesn't make a lot of sense.
Renamed it to `getEnvDefault` to make it a little more clear.

components/notebook-controller/pkg/culler/culler.go
kimwnasptd: Done, might have overdone it a little bit with the verbosity but the description should be clear

components/notebook-controller/pkg/culler/culler.go
kimwnasptd: Simplified it to edit the `meta.Annotations` directly and not use a Get and then a Set method.
I think it's ok in the `RemoveStopAnnotation`.

kimwnasptd:
@lluunn 
> If the culler is going to be general, should it live in somewhere else? Like being its own component.

I'd like to check knative first and examine if it would make sense to use that as a general culler. 

If in the end we decide that it would be best to have the general culler as a go library, which could be used directly from the controllers, then I'll move the code to a separate component. Since it's used only in Notebooks, I thought of leaving it there for now.

components/notebook-controller/pkg/culler/culler.go
lluunn: I am a little confused. If the controller is setting the annotation, and is also checking the annotation, why do we need the annotation? Doesn't controller just know whether it should be culled?

kimwnasptd: My apologies, I didn't provide enough details regarding that.
Yes, the controller knows from the start if the Notebook should be culled and could do it without the annotation. 

I decided to set an annotation to make it easier for the UI later to check that a Notebook is (being) stopped. Also, if the UI would want to start/stop a Notebook then it would only need to modify the annotation on the CR.

lluunn: I see

lluunn:
/lgtm
/approve
/hold
for @jlewi

jlewi:
@kimwnasptd Is the PR description still up to date? I thought based on 
https://github.com/kubeflow/kubeflow/pull/3856#issuecomment-522493345 
the notebook controller was going to use last activity not a generic culler?

Also lets change the comment so we don't close #1394 automatically when this is merged.
We can use #1394 to track making culling behavior settable per resource. Which might also entail making some UI changes to surface it as well.

jlewi:
One minor comment but otherwise this looks great.
I'm super excited about this feature.

Also I think a generic culler would be useful for managing other apps like TensorBoard.

jshelman:
should this work in a dex enabled cluster?

seeing RBAC: access denied.  when trying to access the api/status from pods in the cluster.

and seeing log entries in the notebook controller:
020-04-11T01:50:31.165Z INFO culler Warning: GET to http://testnb.namespace.svc.cluster.local/notebook/namespace/testnb/api/status: 403

PR #3864, opened 2019-08-10
kfctl: pin configs to v0.6.1

### LINKED ISSUE #3862: v0.6 kfctl_k8s_istio.yaml config should point to 0.6 manifests repos
/kind bug

When I use kfctl_k8s_istio.yaml config for kfctl --config option, the yaml spec points to manifests from master.

I expect the config to point to the manifests from the v0.6-branch

Possibly the documenation https://www.kubeflow.org/docs/started/k8s/kfctl-k8s-istio/ should also point the config file from the v0.6 branch rather than master?

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): v0.6.1
- kfctl version: (use `kfctl version`): v0.6.1
- Kubernetes platform: (e.g. `minikube`) Existing k8s  or Minikube
- Kubernetes version: (use `kubectl version`): 1.15.x
- OS (e.g. from `/etc/os-release`): Linux

yanniszark:
Fixes #3862 

This PR pins the kubeflow and manifests repo versions to v0.6.1 (which is the latest release) for all configs.

/assign @abhi-g @lluunn 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3864)

### DIFF
--- a/bootstrap/config/kfctl_aws.yaml
+++ b/bootstrap/config/kfctl_aws.yaml
@@ -252,11 +252,12 @@ spec:
   enableApplications: true
   packageManager: kustomize
   repos:
-    - name: kubeflow
-      uri: https://github.com/kubeflow/kubeflow/archive/master.tar.gz
-    - name: manifests
-      root: manifests-master
-      uri: https://github.com/kubeflow/manifests/archive/master.tar.gz
+  - name: manifests
+    root: manifests-v0.6.1
+    uri: https://github.com/kubeflow/manifests/archive/v0.6.1.tar.gz
+  - name: kubeflow
+    root: kubeflow-v0.6.1
+    uri: https://github.com/kubeflow/kubeflow/archive/v0.6.1.tar.gz
   useBasicAuth: false
   useIstio: true
   version: master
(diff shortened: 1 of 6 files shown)


lluunn:
I think we want to pin only at 0.6, so PR should make to v0.6-branch?

yanniszark:
@lluunn why is that? The latest release is v0.6.1. Why not use that?

cc @abhi-g

lluunn:
Why not master points to master, 0.6 points to 0.6?

Also test is using the config file, and master should be tested.

abhi-g:
I think Lun-kai's point is to pin to the v0.6-branch rather than a tag such
as v0.6.0, v0.6.1 etc. That way if we have patch releases, we won't need to
update these.

Abhishek

jlewi:
It looks like this PR would be pinning the version on master.
Should we instead check this change in to the v0.6-branch of kubeflow/kubeflow?

I think it makes sense for kubeflow/kubeflow master to point to master of kubeflow/manifests

yanniszark:
@jlewi sounds good to me.
I'll submit a pr for the v0.6-branch.

yanniszark:
I'm closing this one for #3867

Issue #3878, opened 2019-08-13
ENV reference error in Dockerfile


xauthulei:
There is ENV `TF_SERVING_VERSION` defined in [Dockerfile](https://github.com/xauthulei/kubeflow/blob/master/components/tensorflow-notebook-image/Dockerfile) liking this :
`ARG TF_PACKAGE_PY_27=https://storage.googleapis.com/tensorflow/linux/cpu/tensorflow-1.7.0-cp27-none-linux_x86_64.whl`
 `......`
`ENV TF_SERVING_VERSION $TF_PACKAGE_PY_27`

In fact, we want to  get a version from ARG: `TF_SERVING_VERSION`

PR #3879, opened 2019-08-13
Fix the ENV reference error.

xauthulei:
Fixes: #3878


jinchihe:
@xauthulei Seems the presubmit failed due to testing env problem, could you please log a issue in the kubeflow/testing repo to trace, thanks.

PR #3880, opened 2019-08-13
[Central-Dashboard v2] MUI Tests - Backend:

avdaredevil:
## About:
- Divided tests between API-Tests, Workgroup-API-Tests
- Refactored shared test code (modularized)
- /env-info now moved into WorkgroupAPI, routes updated everywhere
- 404 API handler moved outside
- User detection tests added for GCP / Other-IAP / (Basic | Unsecured)
- More to come...

### Meta
/area centraldashboard
/area front-end
/priority p1
/assign @avdaredevil
/cc @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3880)


components/centraldashboard/app/api_workgroup.ts
```diff
@@ -209,6 +258,21 @@ export class WorkgroupApi {
                 }
                 res.json(response);
             })
+        .get(
+            '/env-info',
```
prodonjs: When we spoke last week, I think this route still has the odd blending of concerns between core Kubernetes information (ie. the platformInfo stuff) and Profiles stuff. I would say that it might be more beneficial to split `/env-info` into different routes so that the non-profiles related information can be kept independent of the profiles information.  It will result in an extra client-side request on page load, but I think that will end up keeping the implementations more narrowly scoped and easier to test and change if need be.

components/centraldashboard/app/server.ts
```diff
@@ -19,7 +19,7 @@ import {getMetricsService} from './metrics_service_factory';
 const {
   PORT_1 = 8082,
   PROFILES_KFAM_SERVICE_HOST = 'localhost',
```
prodonjs: We should address the Kfam service routing issue in this PR since we know it's possible for the centraldashboard deployment to come up before profiles does, thereby starting the server without these environment variables being set. Since the dashboard and profiles-kfam services will be deployed in the same namespace, I think it should be safe to use `profiles-kfam` as the default `PROFILES_KFAM_SERVICE_HOST` value so that the full hostname ends up being `http://profiles-kfam:8081`.

avdaredevil: Thus has been fixed in #3899

prodonjs:
@avdaredevil - I think there were still some pending questions here to tie up before this can be merged, so just a gentle ping so we can address it over the next few days.

components/centraldashboard/app/api_workgroup_test.ts
```diff
@@ -0,0 +1,500 @@
...
+    let mockK8sService: jasmine.SpyObj<KubernetesService>;
+    let mockProfilesService: jasmine.SpyObj<DefaultApi>;
+    let testApp: express.Application;
+    let port: number;
+    const newAPI = () => new WorkgroupApi(
+        mockProfilesService,
+        mockK8sService,
+    );
+
+    describe('User Detection', () => {
```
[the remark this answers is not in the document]
avdaredevil: Makes sense, removing.

prodonjs:
Getting a failure from the `Workgroup API Add / Remove Contributor : Should successfully remove a contributor` test. Also, it looks like somehow our linting and formatting styles aren't being enforced so I'm going to look at correcting that.

avdaredevil:
Seems like that was because of: 
- https://stackoverflow.com/questions/299628/is-an-entity-body-allowed-for-an-http-delete-request
- https://github.com/nodejs/node/issues/19179

I just modded the headers:
```js
{...headers, 'Transfer-Encoding': 'chunked'}
```

And this seemed to fix the issue with the server not reading the body. I've also updated the error message to be more precise

prodonjs:
/lgtm
/approve
/hold

Try rebasing on top of master to get presubmit tests to pass.

avdaredevil:
Rebased @prodonjs!

Issue #3859, opened 2019-08-09
Kubeflow main page namespace selection


blublinsky:
/kind bug

When trying to change namespace on Kubeflow main page, the drop down of namespaces is showing behind the left blue pane, which completely obstructs the list

- Kubeflow version: 0.6
- kfctl version: kfctl v0.6.0-0-g71aea0a9
- Kubernetes platform: OpenShift
- Kubernetes version: kubernetes v1.13.4+c62ce01

bbhuston:
Experiencing same issue when installing Kubeflow into an existing GKE cluster

Kubeflow version: 0.6.1-rc.2
kfctl version: kfctl v0.6.1-rc.2-1-g3a37cbc6
Kubernetes platform: GCP/GKE (install into existing cluster)
Kubernetes version: kubernetes v1.13.4

bbhuston:
> Experiencing same issue when installing Kubeflow into an existing GKE cluster
> 
> Kubeflow version: 0.6.1-rc.2
> kfctl version: kfctl v0.6.1-rc.2-1-g3a37cbc6
> Kubernetes platform: GCP/GKE (install into existing cluster)
> Kubernetes version: kubernetes v1.13.4

Just tried installing with the same setup as above but using version 0.6.2-rc.2.   Same issue with namespace selection

shawnho1018:
I encountered similar issue. However, I could not judge if the drop down of namespaces is showing behind the left blue pane. For me, the namespace selection is just not working. When I click "Select namespace", there is a only a button animation (blue color appeared on white button) but no drop-down selection at all. 

my kfctl version:  v0.6.1-rc.2-1-g3a37cbc6
kubernetes platform: Vanilla kubernetes installed from kubeadm
Kubernetes version: kubernetes v.1.15.1

avdaredevil:
Yes, it's the same and has been resolved with the latest release. 

@shawnho1018 seems to have a different issue, and that seems to be based on no namespace list being returned to the dashboard (might be a networking / permissioning issue)

shawnho1018:
> Yes, it's the same and has been resolved with the latest release.
> 
> @shawnho1018 seems to have a different issue, and that seems to be based on no namespace list being returned to the dashboard (might be a networking / permissioning issue)

I also suspected it was the serviceaccount issue but I don't see any error log in 
kubectl logs -n kubeflow -f centraldashboard-79df4778f7-glhfj

[...]

PR #3894, opened 2019-08-14
Pin kubeflow/manifests and kubeflow/kubeflow commits on the v0.6-branch

jlewi:
Related to #3859.

We don't want people deploying with kfctl to pick up changes to the manifests
until they've been tested.

kfctl should be pointing at the KFDef specs on the v0.6-branch.
So we need to pin the commits it points.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3894)

### DIFF
--- a/bootstrap/config/kfctl_existing_arrikto.0.6.yaml
+++ b/bootstrap/config/kfctl_existing_arrikto.0.6.yaml
@@ -101,9 +101,8 @@ spec:
   - profiles
   useIstio: true
   repos:
-  - name: manifests
-    root: manifests-0.6-branch
-    uri: https://github.com/kubeflow/manifests/archive/v0.6-branch.tar.gz
   - name: kubeflow
-    root: kubeflow-0.6-branch
-    uri: https://github.com/kubeflow/kubeflow/archive/v0.6-branch.tar.gz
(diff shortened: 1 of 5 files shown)


yanniszark:
@jlewi thanks for the PR.
It will be a huge help for users.

The repos for every config should be:

```yaml
- name: kubeflow
  root: kubeflow-0.6.1
  uri: https://github.com/kubeflow/kubeflow/archive/v0.6.1.tar.gz
- name: manifests
  root: manifests-0.6.1
  uri: https://github.com/kubeflow/manifests/archive/v0.6.1.tar.gz
```

That is because the folder inside the zip is `<repo>-0.6.1` instead of `<repo>-0.6.1`.

jlewi:
Good catch thanks

jlewi:
We should check I think we might have added some logic to handle things like the fact that commits get appended to the commit.

jlewi:
https://github.com/kubeflow/kubeflow/blob/337fdd0dbcde92d20eb270d4c123bec98ffd2ebc/bootstrap/pkg/apis/apps/kfdef/v1alpha1/application_types.go#L460

I think we might be able to get rid of root.

PR #3899, opened 2019-08-14
[CentralDashboard v2] Registration flow for basic auth

avdaredevil:
## Feature Enhancements | Bug Fixes | Basic Auth Registration:
- Registration View logic added for non-identity networks
- Custom note added for basic auth networks
- `/workgroup/create` is now a non-privileged route
- Privileged API error changed to make more sense, like `access denied`
- Namespace selector now has a no-namespaces state, when Σ(Name) = 0
- [BUG] Fixed character input regex to actually allow correct chars
- [BUG] Removed outdated animation for profile creation

### Meta
/area centraldashboard
/area front-end
/priority p0
/assign @avdaredevil
/cc @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3899)

### DIFF
--- a/components/centraldashboard/app/api_workgroup.ts
+++ b/components/centraldashboard/app/api_workgroup.ts
@@ -209,16 +209,6 @@ export class WorkgroupApi {
                 }
                 res.json(response);
             })
-        .use((req: Request, res: Response, next: NextFunction) => {
-            if (!req.user.hasAuth) {
-                return apiError({
-                    res,
-                    code: 405,
-                    error: ERRORS.operation_not_supported,
-                });
-            }
-            next();
-        })
         .post('/create', async (req: Request, res: Response) => {
             const profile = req.body as CreateProfileRequest;
             try {
@@ -243,6 +233,16 @@ export class WorkgroupApi {
                     err,
                 });
             }
+        })
+        .use((req: Request, res: Response, next: NextFunction) => {
+            if (!req.user.hasAuth) {
+                return apiError({
+                    res,
+                    code: 405,
+                    error: 'Access to this api is denied',
+                });
+            }
+            next();
         })
         .get('/get-all-namespaces', async (req: Request, res: Response) => {
             try {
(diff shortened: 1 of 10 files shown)


components/centraldashboard/app/api_workgroup.ts
```diff
@@ -244,6 +234,16 @@ export class WorkgroupApi {
                 });
             }
         })
+        .use((req: Request, res: Response, next: NextFunction) => {
+            if (!req.user.hasAuth) {
+                return apiError({
+                    res,
+                    code: 405,
+                    error: 'Access to this api is denied',
```
prodonjs: You may want to make the error message more descriptive if you intend on surfacing it to the user. Perhaps something that suggests that the user's identity could not be determined from the request.

avdaredevil: Makes sense, done!

avdaredevil:
So it is actually entirely possible to have namespaces without workgroup bindings, hence even though `kubeflow` the namespace exists, there are no bindings for it (thus is not shown in the list of namespaces). We need to have at least one binding for your namespace list to be non-empty (beyond that for now, customers can use the CLI to create more)

prodonjs:
I guess my question is more from a UX perspective. Should there be an implied binding (or should one be created automatically) in the case where there is no ability to determine a user's identity. In that case, what should the behavior be? Does the first user who tries to access the UI have to go through the registration flow but afterwards any subsequent user bypasses it since all users are given the same access? If so, is it worth showing them the screen or could it just be created silently?

avdaredevil:
I asked the same question, @abhi-g could elaborate more on this. But the general understanding is that this allows people who skip the docs to not be stuck with a network with no bindings (thus unable to use any of the subapps

abhi-g:
@prodonjs to answer your questions... please see inline
> I guess my question is more from a UX perspective. Should there be an implied binding (or should one be created automatically) in the case where there is no ability to determine a user's identity. In that case, what should the behavior be? 
Agreed. the plan is that @kunmingg will make the change such that when deploying Kubeflow without identity, the deployment process will create a default/anonymous user profile with the initial binding, that way when the user lands on the dashboard,  there should already be a namespace available.


> Does the first user who tries to access the UI have to go through the registration flow but afterwards any subsequent user bypasses it since all users are given the same access? If so, is it worth showing them the screen or could it just be created silently?
With the above default profile, there really shouldn't be a case where a user visits the dashboard and there is no binding available. In case we want to be sure that for whenever it was missed, there is a still a profile creation path triggered by a first visit, let's make it implicit, because we are trying to keep the behavior in this case similar to v0.5.

prodonjs:
@avdaredevil This has failing server-side tests which is why the presubmit job is failing (http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-dashboard-release-3899-4bb9d15-9920-41c2?tab=workflow&nodeId=kubeflow-presubmit-dashboard-release-3899-4bb9d15-9920-41c2-378483673&sidePanel=logs%3Akubeflow-presubmit-dashboard-release-3899-4bb9d15-9920-41c2-378483673%3A).

PR #3904, opened 2019-08-14
kfctl: existing_arrikto: update config and enable multi-user identity awareness

yanniszark:
Depends on https://github.com/kubeflow/manifests/pull/274

* Update `kfctl_existing_arrikto.yaml` to use the applications spec. 
* The AuthService adds the header `kubeflow-userid` in every request.
* Parameters are added to each application so that it looks for the `kubeflow-userid` header.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3904)

### DIFF
--- a/deployment/existing/auth_oidc/authservice.tmpl
+++ b/deployment/existing/auth_oidc/authservice.tmpl
@@ -41,7 +41,7 @@ spec:
                 path: tls.crt
       containers:
       - name: authservice
-        image: gcr.io/arrikto/kubeflow/oidc-authservice:v0.2
+        image: gcr.io/arrikto/kubeflow/oidc-authservice:v0.3
         imagePullPolicy: Always
         ports:
         - name: http-api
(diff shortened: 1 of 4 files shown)


deployment/existing/auth_oidc/envoy-filter.yaml
```diff
@@ -19,6 +19,10 @@ spec:
           allowedHeaders:
             patterns:
             - exact: "cookie"
+        authorizationResponse:
+          allowedUpstreamHeaders:
+            patterns:
+            - exact: "kubeflow-userid"
```
Jeffwan: This will be a fix header and as I think it will give access to all requests with this header? 
But for userA who like to visit notebooks of userB, this filter will still authorize request?

yanniszark: @Jeffwan the AuthService will add this header to every request.
If a malicious userA tries to specify a `kubeflow-userid` header to impersonate userB, the AuthService will prevent that, since it overrides the header with the correct value.

yanniszark:
@Jeffwan @gabrielwen can you please lgtm if everything is ok?

deployment/existing/auth_oidc/authservice.tmpl
```diff
@@ -41,7 +41,7 @@ spec:
                 path: tls.crt
       containers:
       - name: authservice
-        image: gcr.io/arrikto/kubeflow/oidc-authservice:v0.2
+        image: gcr.io/arrikto/kubeflow/oidc-authservice:v0.3
```
Jeffwan: @yanniszark BTW, is this application open sourced somewhere?

PR #3912, opened 2019-08-15
Don't set dummy values for project, email, or hostname in kfctl_gcp iap.yaml

jlewi:
* Comment out lines setting default values and provide comments explaining how
  they should be set.

* We don't want to set values because the behavior in kfctl init/generate is
  to try to automatically set intelligent default values but only if no
  value is set. So explicitly setting default values inhibits that behavior.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3912)


lluunn:
do you want to do the same for basic_auth config?

jlewi:
@lluunn good point done;

PR #3914, opened 2019-08-15
WIP statefulset GC and deployment status query endpoint

kunmingg:
- have `--mode=gc`, which will keep looping over statefulsets and try to freeze them. If freeze succeeded (no ongoing request), will delete them.
- kfctlserver: add `FreezeSvc` endpoint
- kfctlserver: add `QueryStatus` endpoint

### DIFF
--- a/bootstrap/cmd/bootstrap/app/kfctlServer.go
+++ b/bootstrap/cmd/bootstrap/app/kfctlServer.go
@@ -12,6 +12,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"github.com/go-kit/kit/endpoint"
 	httptransport "github.com/go-kit/kit/transport/http"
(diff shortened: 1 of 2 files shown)


jlewi:
Thanks @kunmingg I think to summarize our discussion yesterday

* expiration time for the statefulset will be stored in an annotation for the statefulset
* the kfctlServer would update its annotation

Is that what you are thinking?

Rather than add "--mode" would it be simpler just to create a separate main file for the GC?

kunmingg:
@jlewi updated

kunmingg:
@jlewi 
Updated.
Now check if kfctl service exist by DNS lookup before proceeding to requests.

jlewi:
The DNS check is really nice.

I think my main comment is if we can add a bit more test coverage as I suggested in my detailed comment in reviewable.

Issue #3921, opened 2019-08-16
Using non standard `df` on macOS causing error


butuzov:
/kind bug

**What steps did you take and what happened:**

I am using `coreutils` package in order to be close to Unix ecosystem, which provides almost native `df` tool in order to replace macOS's `df` version. This situation causing an issue with determining correct available disk space. 

```bash
> cd scripts/setup-minikube.sh
> chmod +x setup-minikube.sh
> ./setup-minikube.sh
Assign CPUs between 3..22 [12]: 
Assign memory (in GB) between 8..91 [48]: 
df: invalid option -- 'g'
Try 'df --help' for more information.
./setup-minikube.sh: line 161: ((: < 40: syntax error: operand expected (error token is "< 40")
Assign disk space (in GB) between 40..0 [50]: 
```

**What did you expect to happen:**
```bash
Assign CPUs between 3..22 [12]: 
Assign memory (in GB) between 8..91 [48]: 
Assign disk space (in GB) between 40..701 [50]: 50
Choose a virtualizer/hypervisor installed on your system (virtualbox, vmwarefusion, kvm2, hyperkit):
```
  

**Environment:**
* `coreutils` installed with brew package manager.

jlewi:
Duplicate of #3610 
Please see: https://github.com/kubeflow/kubeflow/issues/3610#issuecomment-522019187

kfctl.sh is no longer supported please see that comment for recommended way to install Kubeflow on minikube with 0.6

jlewi:
Reopening this issue; because its not about kfctl.sh but setup-minikube.sh which might be worth updating and keeping.

PR #3922, opened 2019-08-16
macOS's custom `df` path.

butuzov:
This bugfix resolves error client can get if use non-standard `df` (provided by `coreutils` package) on macOS paltform.

See #3921

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3922)


butuzov:
Hm. Should I resubmit PR using 0.6 branch as a base for the fix?

butuzov:
@googlebot I consent.

jlewi:
@butuzov Yes you should probably rebase on either the v0.6-branch or master and then resubmit.

Its usually best to submit to master and then cherrypick on to the release branch.

butuzov:
got it.

PR #3933, opened 2019-08-16
[CentralDashboard v2] Error surfacing for profile controller errors

### LINKED ISSUE #3524: Multi-user CUJ with notebooks and central dashboard test with basic auth
Opening this issue to track verifying that the multi-user features @avdaredevil is adding to central dashboard works in the case of basic auth.

The central dashboard and other multi-user features depend on a header containing the user id.

The header in the JWT will be different depending on the setup and there's some question on how to handle this.

avdaredevil:
## Feature Updates:
- Promise rejection from `/api/workgroup/exists` now caught
- All unhandled promises are printed
- Error handling logic added to main-page
- Error toast added, available for all components to use
- [intuititve] added a close button to toast since it's persistent

### Meta
/area centraldashboard
/area front-end
/priority p1
/assign @avdaredevil
/cc @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3933)

### DIFF
--- a/components/centraldashboard/app/server.ts
+++ b/components/centraldashboard/app/server.ts
@@ -57,7 +57,7 @@ async function main() {
 
 // This will allow us to inspect uncaught exceptions around the app
 process.on('unhandledRejection', error => {
-  console.log('[SEVERE] unhandledRejection', error);
+  console.error('[SEVERE] unhandledRejection', error);
 });
 
 main();
--- a/components/centraldashboard/app/server.ts
+++ b/components/centraldashboard/app/server.ts
@@ -55,4 +55,9 @@ async function main() {
       () => console.info(`Server listening on port http://localhost:${port}`));
 }
 
+// This will allow us to inspect uncaught exceptions around the app
+process.on('unhandledRejection', error => {
+  console.log('[SEVERE] unhandledRejection', error);
+});
+
 main();
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -167,6 +167,34 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
     }
 
     /**
+     * Show a toast error on main-page
+     * @param {string} err Error message to show
+     */
+    showError(err) {
+        const el = this.$.ErrorToast;
+        el.text = err;
+        el.show();
+    }
+
+    closeToast(ev) {
+        const t = ev.target;
+        const el = [t, t.parentNode, t.parentNode.parentNode]
+            .find((e) => e.tagName == 'PAPER-TOAST');
+        el.close();
+    }
+
+    /**
+     * Set state for loading registration flow in case no workgroup exists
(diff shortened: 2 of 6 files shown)


components/centraldashboard/app/server.ts
```diff
@@ -55,4 +55,9 @@ async function main() {
       () => console.info(`Server listening on port http://localhost:${port}`));
 }
 
+// This will allow us to inspect uncaught exceptions around the app
+process.on('unhandledRejection', error => {
+  console.log('[SEVERE] unhandledRejection', error);
```
prodonjs: I'd log this as [error](https://nodejs.org/docs/latest-v10.x/api/console.html#console_console_error_data_args) just so that it can be surfaced more easily in tools like Stackdriver Logging.

components/centraldashboard/public/components/main-page.js
```diff
@@ -166,6 +166,34 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         this.$.welcomeUser.show();
     }
 
+    /**
```
prodonjs: Why not use the declarative `text` and `opened` properties for paper-toast as opposed to imperatively opening/closing it?

https://www.webcomponents.org/element/@polymer/paper-toast/elements/paper-toast

components/centraldashboard/app/server.ts
avdaredevil: Makes sense, will do!

components/centraldashboard/public/components/main-page.js
avdaredevil: This one shows it for the delayed amount (which is 5s), opened I believe will keep it persisted.

prodonjs: https://www.webcomponents.org/element/@polymer/paper-toast/elements/paper-toast#property-duration looks like it can be used to set the duration. Either way is appropriate but it's nice to use the property surface as it helps reduces the mock interactions that need to be tested.

avdaredevil: Fair enough, will change!

avdaredevil: So Jason, for this element, the text prop is already set, so is the duration. My point is that we need this to show after 500ms, and we never need to close this element.

`.show()` opens the element, waits `duration` long, and closes the element. This behavior may not make sense via props, since we set a state that then either get's changed or outdated in `duration` ms.

prodonjs: My suggestion was to have properties on the main-page component that are passed down to the paper toast so that changing them affects the display behavior for paper toast.

Then you could just have this in the markup:
```
paper-toast#ErrorToast(opened='[[errorText]]', text='[[errorText]]', duration=0)
```
That at least handles the display part of it. Then all `closeToast` needs to do is clear the errorText property. You can even drop down to having a single paper-toast element with other props to control the style (error vs. information).

PR #3941, opened 2019-08-19
Fix the error link description when using kfctl install.

xauthulei:
Review comments from @jlewi :
    Right now we periodically copy all the changes from kubeflow/kubeflow to kubeflow/kfctl. So your changes will get overwritten unless you commit this on kubeflow/kubeflow.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3941)


xauthulei:
/assign @jlewi 
/cc @jinchihe 
 Could you review this.

Issue #3662, opened 2019-07-15
setup kfctl nightly builds


swiftdiaries:
It would be great if we could setup nightly builds for kfctl. It would go well along with the E2E tests and help in development. 

/cc @kkasravi @jlewi @yanniszark

jlewi:
@swiftdiaries What are you thinking in terms of how nightly builds are published?

jlewi:
@swiftdiaries We could piggy back on our existing E2E tests to automatically publish the binary somewhere. It might be easiest to start with GCS just because our CI system already has access to GCS; accessing GitHub release pages is a bit more work but doable.

For GitHub I think we'd want to create a GitHub APP and use that to authenticate to our repos.

It might also make sense to try to use TektonCD/pipeline as that would allow us to break down our workflows into reusable tasks so we could have

* 1 Task to build kfctl
* 1 Task to push to GCS
* 1 Task to push to GitHub
* 1 Task to deploy Kubeflow
* N tasks testing that Kubeflow is valid

We currently have a makefile which builds kfctl and pushes it to GitHub using hub-cli.
https://github.com/kubeflow/kubeflow/blob/8da1e4709dfb53867cce9156ccd52945bd82af21/bootstrap/Makefile#L101

A think a good step might just be creating a 1 step Tekton pipeline to build kfctl.

swiftdiaries:
@jlewi Are we looking to move from Argo to TektonCD for E2E tests?

swiftdiaries:
+1 for pushing to GCS
I'll get to the GitHub app in a subsequent PR, the initial one would just be a piggyback on the existing E2E workflow to build and push the kfctl binary to GCS.
Although, for GCS should we create a separate bucket that can be accessed by devs or publish the nightly builds on GitHub?

jlewi:
@swiftdiaries I think TektonCD is worth investigating but I wouldn't block any critical work on that.

>Although, for GCS should we create a separate bucket that can be accessed by devs or publish the nightly builds on GitHub?

I would start with GCS just because its simpler. We can create a public bucket if necessary.

jlewi:
@swiftdiaries Any update on this?

swiftdiaries:
I've pushed a PR that takes builds off of the E2E test to create a bucket and push the binary. 
https://github.com/kubeflow/kubeflow/pull/3949
PTAL

swiftdiaries:
I've added you as a reviewer on that PR

[...]

PR #3949, opened 2019-08-20
builds, pushes kfctl to GCS as part of pre and post submits

swiftdiaries:
Things investigated:

- [x] TektonCD to build kfctl
- [x] Kaniko to build kfctl 
- [x] Argo Cron Jobs
- [x] Prow periodic jobs

TektonCD and Kaniko failed with similar errors while building kfctl. The way the files are mapped either from git checkout to build step in Tekton and local filesystem to container volume for kaniko is buggy (or likely I made a mistake). 
go.mod file is not found in the context in Tekton and the hack/v2.zip is not found in the kaniko build. 
Argo Cron Jobs need to be run as Kubernetes Cron Jobs with the workflow YAML pulled in and `kubectl create`'d on the fly. In our current workflow, running that would be to do a `ks show <app_name> -c <workflow_name> | kubectl create -f -`.
Prow periodic jobs are run like [this 
](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubeflow/kubeflow-periodics.yaml#L3) but I wasn't sure how to add this field from the run_e2e_workflow.py in kubeflow/testing.

I ran these on latest master.

In the interim, I've added an extra workflow to build and push to GCS as part of pre-submits.
I'll continue to work towards making it a nightly build and push. 

Any ideas / thoughts would be good. Please add to discussion here: https://github.com/kubeflow/kubeflow/issues/3662.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3949)


jlewi:
Why presubmits why not postsubmits?

jlewi:
Maybe we should do it on pre and postsubmit builds just to ensure the logic gets tested on presubmits?

swiftdiaries:
We can do both pre and post submits

prow_config.yaml
```diff
@@ -63,6 +63,25 @@ workflows:
       workflowName: kfctl-go
       useBasicAuth: false
       useIstio: true
+  # Run a nightly build of kfctl and push the binary to GCS    
+  - app_dir: kubeflow/kubeflow/testing/workflows
```
jlewi: Why do we need a new prow job? It looks like we are copying the binary in kfctl_go_test.py.
That gets executed on every pre/post and periodic run.
Isn't that sufficient?

testing/kfctl/kfctl_go_test.py
```diff
@@ -65,6 +65,18 @@ def test_build_kfctl_go(app_path, project, use_basic_auth, use_istio, config_pat
   run_with_retries(["make", "build-kfctl"], cwd=build_dir)
   kfctl_path = os.path.join(build_dir, "bin", "kfctl")
 
+  # Push built kfctl binary to GCS bucket
+  import google.cloud.storage as storage
```
jlewi: Why is this import here and not at the top of the file?

testing/kfctl/kfctl_go_test.py
```diff
@@ -65,6 +65,18 @@ def test_build_kfctl_go(app_path, project, use_basic_auth, use_istio, config_pat
...
+  gcs_client = storage.Client()
+  bucket = gcs_client.lookup_bucket("kfctl-nightly")
```
jlewi: Should we just use a GCS bucket and then do different subpaths?
We could also just copy kfctl to the bucket that is used by prow.
Then the binaries would show up in the artifacts tab of spyglass e.g.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/3856/kubeflow-presubmit/1165454527155933184/

And people could get to them just by navigating to the appropriate pre/post/periodic  run.

jlewi:
@swiftdiaries It looks like the commit history is messed up. It looks like you may have done a merge and pulled in changes that are not your own. Can you rebase on master and drop the merge commits?

jlewi:
@swiftdiaries Any update on this? Would be great to wrap it up.

jlewi:
@swiftdiaries ping?

PR #3952, opened 2019-08-20
macOS's custom `df` path

butuzov:
This bugfix resolves the error - a client can get if uses non-standard df
(provided by coreutils package and installed via `brew`) on macOS platform.

See #3921
--
this is repeat pr, based on kubeflow/0.6 and to kubeflow/master

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3952)


butuzov:
@googlebot I consent.

jlewi:
@butuzov it looks like there is a ton of conflicts on this branch; I think you need to rebase.
Plase see kubeflow/website#1094

jlewi:
@butuzov you need to fix the commit history. It looks like you merged in master so you pulled in a bunch of changes that are yours and that's causing conflicts.

You need to rebase on the latest HEAD of master. That will try to replay all of the commits in this PR onto the tip of master. When you do an interactive merge you should up all the commits that are not yours since they will already be on the master branch.

jlewi:
The PR is stale so closing it.

PR #3953, opened 2019-08-20
Adding kfserving to Kfctl config

johnugeorge:
KFServing and Knative(dependency) is added to config files

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3953)


jlewi:
How about adding an E2E test to verify that knative and kfserving is actually deployed?

See:
https://github.com/kubeflow/kubeflow/blob/master/testing/kfctl/kf_is_ready_test.py

Do we also have issues tracking adding documentation on kubeflow.org about kfserving?

johnugeorge:
Created issue for tracking docs on KFServing https://github.com/kubeflow/website/issues/1109

Issue #3893, opened 2019-08-13
Multi-Tenancy Compatibility Reset on v0.6


kunmingg:
/kind feature
We need to guarantee kubeflow Multi-Tenancy Compatibility with different environments:
- [ ] Always create default profile 
- [ ] Istio rbac is only `ON` when coming requests contains identity which is accepted by both Istio and k8s.
- [ ] Notebook spawner should not block user creating notebooks when Istio rbac is `OFF`
- [ ] Test for on prem environment; document how to setup Multi-Tenancy on prem.

jlewi:
See also this comment for list of scenarios we want to test
https://github.com/kubeflow/kubeflow/issues/3437#issuecomment-520674809

kimwnasptd:
> Notebook spawner should not block user creating notebooks when Istio rbac is OFF
@kunmingg could you elaborate a little bit more on this one? What's the use case that the jupyter webapp blocks the user from creating a Notebook?

yanniszark:
@jlewi the existing_arrikto config using Dex is capable of identity awareness.
In order to do that, this issue needs to be resolved first: https://github.com/kubeflow/kubeflow/issues/3553

The packages of some applications in manifests have to be extended to make the userid header customizable. 
Right now, it is set to `x-goog-authenticated-user-email` by default.

I will try to tackle this issue today, as it should be in this release.

jlewi:
@krishnadurai @yanniszark @avdaredevil I created a doc based on the 0.6.0 release demo script.
http://bit.ly/2ZgOE1v

Lets get it updated for 0.6.2 and use to make sure we are testing all the scenarios we care about.

krishnadurai:
@jlewi looks like you would need to grant access to this document.

jlewi:
@krishnadurai Apologies; I thought I'd put it in the kubeflow shared drive but I guess not. Please try again.

kunmingg:
@yanniszark 
Seems https://github.com/kubeflow/kubeflow/blob/v0.6-branch/bootstrap/config/kfctl_existing_arrikto.yaml still have Istio rbac on.
Will we have user identify setup in proxy under this condition?

jlewi:
0.6.2 has been released so I'm going to go ahead and close this.

PR #3877, opened 2019-08-13
Disable istio authorization in kfctl_k8s_istio

Jeffwan:
https://github.com/kubeflow/kubeflow/issues/3876

kfctl_k8s_istio doesn't have authentication, disable authorization to make user visit notebook they created.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3877)

PR #3959, opened 2019-08-20
always create default profile; turn off istio rbac for kfctl_k8s_istio

kunmingg:
* Always create a default profile by setting a default "anonymous" value for user id so that a profile would be created for user "anonymous"

* Turn off ISTIO RBAC in kfctl_k8s_istio because this config won't have the user's identity or JWTs so  ISTIO can't do service Authz.

related: https://github.com/kubeflow/kubeflow/issues/3893

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3959)

### DIFF
--- a/bootstrap/config/kfctl_k8s_istio.yaml
+++ b/bootstrap/config/kfctl_k8s_istio.yaml
@@ -34,6 +34,9 @@ spec:
     name: istio-install
   # This component is the istio resources for Kubeflow (e.g. gateway), not about installing istio.
   - kustomizeConfig:
+      parameters:
+        - name: clusterRbacConfig
+          value: "OFF"
       repoRef:
         name: manifests
         path: istio/istio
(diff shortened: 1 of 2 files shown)


kunmingg:
Verified working on `deploy to existing k8s cluster`

jlewi:
Thanks @kunmingg I'd suggest changing the default user id to "anonymous" as opposed to default or "anonymous-user".

This way if we display it e.g. in the UI I think its a little more intelligible then just "default".

Hope you don't mind but I took a stab at updating the PR description; PTAL

yanniszark:
@kunmingg can you also rebase to pick up the change from #3877 ?

yanniszark:
@jlewi do we always want to create a default profile?
In case multi-user is enabled, I believe @avdaredevil 's new UI guides users through the creation of a new profile.
Maybe we should make this opt-in?

I was thinking:
* Make a new package in manifests which creates a profile.
* In case of GCP, set a parameter for the userid.
* Configs can use this package if they want to.

kunmingg:
@yanniszark 
Creating default profile is to smooth cluster owner experience: they'll always have a profile available once cluster is up. We try to resolve some confusions during v0.6.1 adoption.
Going forward we can have manifests pkg preparing default profiles.

PR #3969, opened 2019-08-21
Update repo commit.

gabrielwen:
Using v0.6.2-rc.1 for both kubeflow and manifests.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3969)

### DIFF
--- a/bootstrap/config/kfctl_gcp_iap.0.6.2rc.yaml
+++ b/bootstrap/config/kfctl_gcp_iap.0.6.2rc.yaml
@@ -297,9 +297,10 @@ spec:
   repos:
   - name: kubeflow
     root: kubeflow-0.6.2
-    uri: https://github.com/kubeflow/kubeflow/archive/v0.6.2.tar.gz
+    # Commit for v0.6.2-rc.1
+    uri: https://github.com/kubeflow/kubeflow/archive/0dbd2550372c003ba69069aeee283bd59fb1341f.tar.gz
   - name: manifests
-    uri: https://github.com/kubeflow/manifests/archive/v0.6.2.tar.gz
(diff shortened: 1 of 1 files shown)


bootstrap/config/kfctl_gcp_iap.0.6.2rc.yaml
```diff
@@ -296,11 +296,11 @@ spec:
...
   - name: kubeflow
-    root: kubeflow-0.6.1
-    uri: https://github.com/kubeflow/kubeflow/archive/v0.6.1.tar.gz
+    root: kubeflow-0.6.2
+    # Commit for v0.6.2-rc.1
+    uri: https://github.com/kubeflow/kubeflow/archive/0dbd2550372c003ba69069aeee283bd59fb1341f.tar.gz
   - name: manifests
-    # TODO(jlewi): This is already outdated; need to pick up kubeflow/manifests#280
-    uri: https://github.com/kubeflow/manifests/archive/a5c7fe3ce73d423baa2a271b19928d6a1ad0f587.tar.gz
+    uri: https://github.com/kubeflow/manifests/archive/6dc0c0d10a8060ff748a74da650776da1ba4e3b7.tar.gz
```
jlewi: Why update the manifests? Is there particular changes?

jlewi: Looks like this is picking up the fix to the version.
https://github.com/kubeflow/manifests/commits/v0.6-branch

jlewi:
As discussed in slack; we are pinning to commits and not tags; to avoid having to to update things when  we promote an RC to a final candidate.

/lgtm
/approve

bootstrap/config/kfctl_gcp_iap.0.6.2rc.yaml
gabrielwen: yap, also updating image.

PR #3971, opened 2019-08-22
[Centraldashboard v2] Env vars added to dockerfile

avdaredevil:
## About
Added production env var in docker file, and statement print in server init.

### Meta
/area centraldashboard
/area front-end
/priority p1
/assign @avdaredevil
/cc @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3971)

### DIFF
--- a/components/centraldashboard/app/server.ts
+++ b/components/centraldashboard/app/server.ts
@@ -11,7 +11,7 @@ import {getMetricsService} from './metrics_service_factory';
 
 const isProduction = process.env.NODE_ENV === 'production';
 const defaultKfam = isProduction
-? 'profiles-kfam'
+? 'profiles-kfam.kubeflow'
 : 'localhost';
 
 /* PROFILES_KFAM env vars will be set by Kubernetes if the Kfam service is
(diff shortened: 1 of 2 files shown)


components/centraldashboard/app/server.ts
```diff
@@ -9,9 +9,10 @@ import {WorkgroupApi} from './api_workgroup';
 import {KubernetesService} from './k8s_service';
 import {getMetricsService} from './metrics_service_factory';
 
-const defaultKfam = process.env.NODE_ENV !== 'production'
-  ? 'localhost'
-  : 'profiles-kfam';
+const isProduction = process.env.NODE_ENV === 'production';
+const defaultKfam = isProduction
+? 'profiles-kfam.kubeflow'
```
prodonjs: I don't know that you *need* to do this since the two services will be deployed together in the same namespace. @kunmingg suggested that on https://github.com/kubeflow/manifests/pull/298/files so if he's good with that, let's go for it. The only problem this could have is if somehow a different namespace was used for both services.

avdaredevil: I tried suggesting the same, but seems @jlewi also wants this change, so I think for now this is fine

kunmingg:
@avdaredevil 
Can we have `profiles-kfam.kubeflow` as default value and only point to local when user provide a env var?
And `ENV` in Dockerfile maybe unavailable when container starts.

PR #3980, opened 2019-08-22
[Centraldashboard v2] Added a /debug route to dashboard for python e2e from Kubeflow

avdaredevil:
## About:
Added a `/debug` route to the dashboard for python e2e from Kubeflow

### Meta
/area centraldashboard
/area front-end
/priority p0
/assign @avdaredevil
/cc @prodonjs @gabrielwen @jlewi @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3980)

### DIFF
--- a/components/centraldashboard/app/server.ts
+++ b/components/centraldashboard/app/server.ts
@@ -52,6 +52,10 @@ async function main() {
       user: req.user,
       profilesServiceUrl,
       codeEnvironment,
+      headersForIdentity: {
+        USERID_HEADER,
+        USERID_PREFIX,
+      },
     });
   });
   app.use('/api', new Api(k8sService, workgroupApi, metricsService).routes());
--- a/components/centraldashboard/app/server.ts
+++ b/components/centraldashboard/app/server.ts
@@ -1,5 +1,5 @@
 import {KubeConfig} from '@kubernetes/client-node';
-import express from 'express';
+import express, {Request, Response} from 'express';
 import {resolve} from 'path';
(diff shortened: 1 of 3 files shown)


components/centraldashboard/app/server.ts
```diff
@@ -46,14 +47,20 @@ async function main() {
...
   app.use(express.static(frontEnd));
   app.use(attachUser(USERID_HEADER, USERID_PREFIX));
-  app.use(
-      '/api', new Api(k8sService, workgroupApi, metricsService).routes());
+  app.get('/debug', (req: Request, res: Response) => {
+    res.json({
+      user: req.user,
+      profilesServiceUrl,
+      codeEnvironment,
+    });
```
abhi-g: should we also add the inferred Kubeflow version in this reponse?

components/centraldashboard/app/server.ts
```diff
@@ -46,14 +47,20 @@ async function main() {
...
+  });
+  app.use('/api', new Api(k8sService, workgroupApi, metricsService).routes());
```
abhi-g: I think this line is actually doing what I asked for above. In that case ignore my previous comment.

components/centraldashboard/app/server.ts
abhi-g: /api/env-info request can be used to query the version of Kubeflow as parsed by the CentralDashboard (to be added to a test)

Issue #3623, opened 2019-07-09
Restrict notebook controller (and other controllers) to a subset of namespaces


jlewi:
We'd like to be able to restrict the notebook controller to a subset of namespaces.

Consider the following scenario; a user has a shared cluster in which some namespaces are used with Kubeflow and some are not.

The notebook controller should only have permission to create and deploy resources in the Kubeflow namespaces.

Ideally the list of namespaces should be dynamic; i.e. a label or annotation should be used to indicate whether a namespace is a KF namespace and the notebook controller should be able to access it.

RBAC should be used to enforce access to a subset of namespaces. This might require dynamically creating new RBAC roles; i.e. when a new namespace is created we might need to create a role binding granting the service account of the notebook controller access to the new namespace.

An open question is how restricting a controller to a subset of namespaces interacts with [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/controller)

   * Does the implementation of controller require at least viewer permissions on all namespaces so that the controller can detect events for resources?
      
See kubeflow/tf-operator#759

The goal should be to define a pattern that all of our operators can follow.

/assign @richardsliu 
/assign @johnugeorge 

@richardsliu @johnugeorge Since you are leading our operator efforts do you want to take a look at this?

johnugeorge:
@jlewi 
Currently, operators can be cluster scoped (which then listens to all namespaces) or namespace scoped(restricted to a single namespace).   We can extend to a list of namespaces instead of one.  Namespace to be watched is currently set as a flag which is loaded during startup. 

Do you propose to have labels similar to istio's `istio-injection: enabled`  labels on a namespace?    Shouldn't this be done outside the operator logic?

jlewi:
@johnugeorge We are hearing the following request from customers. They have a shared K8s cluster. Kubeflow should be restricted to a subset of namespaces enforced via RBAC. So we'd like a good story for how to restrict Kubeflow to a subset of namespaces. I think requirements are

* Kubeflow should only have write permissions on a subset of namespaces in which Kubeflow is being used
* It should be easy to add new namespaces in which Kubeflow is being used.

@IronPan What is pipelines currently doing?

richardsliu:
Pipelines implementation:
1. Scope informers to specific namespaces ([PR](https://github.com/kubeflow/pipelines/pull/851/files))
2. Introduce viewer CRD ([PR](https://github.com/kubeflow/pipelines/pull/1562/files))
[...]

PR #3984, opened 2019-08-22
Restricting controller access with Gatekeeper

richardsliu:
Fixes https://github.com/kubeflow/kubeflow/issues/3623

Design: https://bit.ly/2yJeU5u

Note - this doesn't change any Kubeflow behavior by default. The cluster admin needs to follow the instructions in README.md to configure Gatekeeper in the cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3984)

### DIFF
--- a/components/opa-gatekeeper/README.md
+++ b/components/opa-gatekeeper/README.md
@@ -1,8 +1,58 @@
 # Gatekeeper and Kubeflow
 
-[Gatekeeper](https://github.com/open-policy-agent/gatekeeper) is a policy controller for Kubernetes that enforces CRD-based access control policies.
+[Gatekeeper](https://github.com/open-policy-agent/gatekeeper) is a validating webhook for Kubernetes that enforces CRD-based access control policies.
 In Kubeflow, we use Gatekeeper to restrict controllers to their own namespaces. The details can be found [here](https://bit.ly/2yJeU5u).
 
 ## Installation
 
+1. Follow the instructions [here](https://github.com/open-policy-agent/gatekeeper#deploying-a-release-using-prebuilt-image) to install Gatekeeper controller.
+
+1. Apply the constraint template in this directory:
+```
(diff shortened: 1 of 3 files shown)


jlewi:
Thanks @richardsliu should we make this a kustomize package in kubeflw/manifests rather than a component in kubeflow/components?

I don't have a strong oppinion.

If we make it a kustomize manifest that might make it easier to customize; and  potentially starting to auto-deploy it by including it in KFDef.Spec.Applications

johnugeorge:
@richardsliu  
Isn't deploying via kustomize manifest the better option ?

richardsliu:
I wasn't sure if automatically deploying this as a kustomize package is the right way. There are a couple of problems here:
- Gatekeeper dependencies are in https://raw.githubusercontent.com/open-policy-agent/gatekeeper/master/deploy/gatekeeper.yaml. If we want to build a kustomize package for it, we'll have to maintain a copy of this file in the manifests repo. Can we point a kustomize resource to a Github URL?
- The cluster admin would have to manually configure the constraint anyway (e.g. setting the service account names and labels, etc). We can expose these as parameters, but that adds extra maintenance for users who don't need this feature.

Suggestions?

johnugeorge:
@richardsliu Is this planned to be an optional component for a cluster?

jlewi:
@richardsliu I agree with you we shouldn't be installing it by default. That's a good question about whether kustomize can use a GitHub URL.

Since turning it into a kustomize package might require some effort it might make sense to leave it for the future.

Nonetheless would it be better to check the YAML files and directions in this repo into kubeflow/manifests? I think we might want to centralize all manifests (e.g. DM configs etc...) in kubeflow/manifests. Even if it isn't a kustomize package; is that a better location?

richardsliu:
I will close this PR and open a separate one for kubeflow/manifests.

PR #3978, opened 2019-08-22
Update cofig file versions

gabrielwen:
- Update default version in kfctl go binary to v0.6.2 tag.
- Update config files to pin to commit for both kubeflow and manifest repos.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3978)

PR #3991, opened 2019-08-23
Create 0.6.2 versions of all config files.

jlewi:
This way we can start updating these configs without worrying about
breaking any body that is pulling in the existing configs from the tip of the branch.

This is partially rolling forward #3978

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3991)


kunmingg:
@jlewi 
Does kfctl_existing_arrikto.0.6.2.yaml have proxy that provide user identity like IAP?
Otherwise it might not work?

Issue #3986, opened 2019-08-23
[Central Dashboard] Shows user as anyonmous@kubeflow.org when logged in via IAP


jlewi:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

I'm logged into Kubeflow 0.6.2  (RC 0.2) using this [kfctl_gcp_iap.0.6.rc.2.yaml](https://raw.githubusercontent.com/kubeflow/kubeflow/c3199d931b9b82c8717057c7aca883d76bb5e1d5/bootstrap/config/kfctl_gcp_iap.0.6.2rc.yaml). When I log into the central dashboard via IAP it is showing I am `anonymous@kubeflow.org`.

This looks like a bug in the kustomize manifest. It doesn't look like the environment variables USERID_HEADER is set correctly.

```
  - env:
    - name: USERID_HEADER
      value: $(userid-header)
    - name: USERID_PREFIX
      value: $(userid-prefix)
    - name: PROFILES_KFAM_SERVICE_HOST
      value: profiles-kfam.kubeflow
``` 

See: https://github.com/kubeflow/kubeflow/issues/3900#issuecomment-524343425

Full pod spec is attached
[centraldashboard.spec.yaml.txt](https://github.com/kubeflow/kubeflow/files/3535036/centraldashboard.spec.yaml.txt)

@avdaredevil What is impacted by the failure to set USERID_HEADER and PREFIX correctly? Is it just which user is displayed in the centraldashboard? Does it have any impact on the profiles users will see.

jlewi:
It looks like this breaks the automatic profile creation login when I login as a different user.
http://bit.ly/2ZgOE1v

When I login as a different user it thinks I'm anonymous@kubeflow.org so doesn't redirect me through the login flow.

jlewi:
It looks like the configuration file on the branch is different from the one on master
https://github.com/kubeflow/manifests/blob/v0.6-branch/common/centraldashboard/base/params.yaml

https://github.com/kubeflow/manifests/blob/master/common/centraldashboard/base/params.yaml

We are missing the section

```
- path: spec/template/spec/containers/0/env/0/value
  kind: Deployment
- path: spec/template/spec/containers/0/env/1/value
  kind: Deployment
```

I think that means we are overriding the default targets for variable substitution since we are specifying a configuraiton file. But since we didn't include all of the defaults we are no longer substituting into environment variables for deployments.

jlewi:
[...]

PR #3994, opened 2019-08-24
Set userid prefix and header for IAP and pull in the latest fixes to the applications repo for GCP configs

jlewi:
* This should pull in the fix to #3986 (centraldashboard not set with userid correctly).

* We are only changing the GCP configs because that is the only thing I have manually tested and verified.

* Now that we cherrypicked #302 onto the v0.6-branch of manifests; we need to
set userid-prefix and userid-header parameters for the applications
jupyter-webapp, profiles, and centraldashboard in the IAP config.

Otherwise we end up setting those to the empty string and it thinks we
are an anonymous user.

Related to: kubeflow/kubeflow#3986

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3994)


jlewi:
/assign @abhi-g 
/assign @kunmingg 

This is ready for review.

PR #3996, opened 2019-08-24
Set userid prefix and header for IAP.

jlewi:
* When using IAP, we need to configure various applications to look at
the correct header and prefix so that they get the identity correctly.

Related to #3986

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3996)

### DIFF
--- a/bootstrap/config/kfctl_gcp_iap.yaml
+++ b/bootstrap/config/kfctl_gcp_iap.yaml
@@ -1,9 +1,7 @@
-# Please set project and email!
 apiVersion: kfdef.apps.kubeflow.org/v1alpha1
 kind: KfDef
 metadata:
-  creationTimestamp: null
-  name: myapp2
+  name: kf101
   namespace: kubeflow
 spec:
   repos:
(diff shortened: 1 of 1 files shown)


jlewi:
Test failure is 

```
INFO Error: couldn't apply KfApp: (kubeflow.error): Code 500 with message: kfApp Apply failed for kustomize: (kubeflow.error): Code 500 with message: couldn't create resources from seldon-core-operator Error: StatefulSet in version "v1" cannot be handled as a StatefulSet: v1.StatefulSet.Spec: v1.StatefulSetSpec.VolumeClaimTemplates: []v1.PersistentVolumeClaim: decode slice: expect [ or n, but found {, error found in #10 byte of ...|mplates":{"metadata"|..., bigger context ...|bhook-server-secret"}}]}},"volumeClaimTemplates":{"metadata":{"labels":{"app.kubernetes.io/component|...
```

jlewi:
Test failure is #4001; kubeflow/manifests#312 is merged so hopefully master is fixed.

PR #3998, opened 2019-08-24
Update kfctl_existing_arrikto config to use applications.

jlewi:
* Get rid of the deprecated components and componentparams.
* We need to do this because we will need to configure userid header and
  prefix to be identity aware.

* Update the manifests repo to pull in the latest commits.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/3998)


yanniszark:
Hi @jlewi.
I have already made a PR for adding application and multi user support.
Can you take a look here: https://github.com/kubeflow/kubeflow/pull/3904

jlewi:
Thanks @yanniszark

jlewi:
@yanniszark I'll go ahead and close this but we'll need to pull  #3904 into the v0.6-branch

PR #4004, opened 2019-08-26
Automated cherry pick of #3904: kfctl: existing_arrikto: update config file to use Cherry pick of #3904 on v0.6-branch. #3904: kfctl: existing_arrikto: update config file to use

yanniszark:


yanniszark:
Besides cherry picking the commit, I also changed the v0.6.2 existing_arrikto config file to the appropriate form.

bootstrap/config/kfctl_existing_arrikto.yaml
```diff
@@ -1,105 +1,221 @@
+# This is the config to install Kubeflow on an existing K8s cluster, with support
```
jlewi: Do you want to modify kfctl_existing_arrikto.yaml? My suggestion would probably be to only edit the 0.6.2 files so that you don't risk breaking anyone using the kfctl_existing_arrikto.yaml config from the tip of the branch

deployment/existing/istio/istio-noauth.yaml
```diff
@@ -18981,3 +18981,9 @@ spec:
         maxRequestsPerConnection: 10000
 ---
 
+apiVersion: rbac.istio.io/v1alpha1
+kind: ClusterRbacConfig
+metadata:
+  name: default
+spec:
+  mode: 'ON'
```
jlewi: Shouldn't RBACConfig be off if the mode is "noauth"?

yanniszark: The name of the file has remained from before auth was enabled.
Since this is a cherry pick and it wasn't caught in the original PR, I would suggest leaving it (as it doesn't affect behaviour) and fixing it later in master.

bootstrap/config/kfctl_existing_arrikto.yaml
yanniszark: @jlewi the docs point to the latest tag (v0.6.1 in this case).
As such, I wouldn't expect anything to break in docs.

Excerpt from docs: 
```export CONFIG="https://raw.githubusercontent.com/kubeflow/kubeflow/v0.6.1/bootstrap/config/kfctl_existing_arrikto.0.6.yaml"```

yanniszark: I also edited the v0.6.2 file.
I am not sure what the suggestion is.
Won't the config come from a tagged commit (eg v0.6.2)?
If yes, what is the need for having tagged config files?
Better UX?

PR #4040, opened 2019-08-30
adds apply -f to kfctl

swiftdiaries:
Implements case 1 and case 2 in the new kfctl semantics
[Design doc reference](https://docs.google.com/document/d/1kviNOdjoBPeoJ56KBvMKWq7SzrSN5rFUQoULK8muD-M/edit?pli=1#)

- [x] make platform args optional
- [x] implement new semantics for build and apply

Current workflow for kfctl - 

New sematics single command install:
```
# move into empty directory
kfctl apply -f <path to config file / URL>
```
New semantics build, edit, apply:
```
# move into empty directory
kfctl build -f <path to config file / URL>
kfctl apply
```
Old semantics:
```
kfctl init kf-app --config <path to config file / URL> 
cd kf-app
kfctl generate <platform>
kfctl apply <platform>
```

/cc @yanniszark @jlewi @kkasravi

Related to: #3518

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4040)


bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+		log.Printf("Error getting current working directory: %v", err)
+	}
+	fmt.Println("Present working directory is: %v", cwd)
+	// Check if current directory is empty and if it is error out
+	isEmpty, err := IsCwdEmpty(cwd)
+	if !isEmpty && err != nil {
+		return fmt.Errorf("Current directory is not empty, please try again in an empty directory: %v", err)
+	}
+	kfDef.Spec.AppDir = cwd
+	if kfDef.Spec.AppDir == "" {
```
nrchakradhar: Will this be true in any scenario?

swiftdiaries: Thanks for the review! Let me clean it up a little bit. I copied over code from over another PR, so there's remnant pieces around

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+	kfdefsv3 "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis/apps/kfdef/v1alpha1"
+	"github.com/kubeflow/kubeflow/bootstrap/v3/pkg/kfapp/coordinator"
+	log "github.com/sirupsen/logrus"
+)
+
+// BootstrapKubeflow is used by the kfctl apply sub-command to take in a configfile
+// as a flag and boostrap a KfApp and deploy it
+func BootstrapKubeflow(configFilePath string, kfResource kftypes.ResourceEnum) error {
+	// Set default app name to kf-app
+	appName := "kf-app"
```
yanniszark: Nit: can you change this to a constant?

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+	isValid, msg := kfDef.IsValid()
+	if !isValid {
+		log.Printf("Invalid kfdef: %v", isValid)
+		log.Printf("Error validating generated KfDef, please check config file validity: %v", msg)
+	}
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Printf("Error getting current working directory: %v", err)
+	}
+	fmt.Println("Present working directory is: %v", cwd)
```
yanniszark: Can you use log instead of fmt and change all instances of Println to a log-level function? (eg Info, Warn).

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+		log.Printf("Unable to create KfDef from config file: %v", err)
+	}
+	if kfDef.Name != "" {
+		log.Warnf("Overriding KfDef.Spec.Name; old value %v; new value %v", kfDef.Name, appName)
+	}
+	kfDef.Name = appName
+	isValid, msg := kfDef.IsValid()
+	if !isValid {
+		log.Printf("Invalid kfdef: %v", isValid)
+		log.Printf("Error validating generated KfDef, please check config file validity: %v", msg)
```
yanniszark: This doesn't return an error but continues.

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+// BootstrapKubeflow is used by the kfctl apply sub-command to take in a configfile
+// as a flag and boostrap a KfApp and deploy it
+func BootstrapKubeflow(configFilePath string, kfResource kftypes.ResourceEnum) error {
+	// Set default app name to kf-app
+	appName := "kf-app"
+
+	// Construct KfDef from the configFilePath provided
+	kfDef := &kfdefsv3.KfDef{}
```
yanniszark: This line is unnecessary, the kfDef is re-initialized in the next line.

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+	kfDef, err := kfdefsv3.LoadKFDefFromURI(configFilePath)
+	if err != nil {
+		log.Printf("Unable to create KfDef from config file: %v", err)
+	}
+	if kfDef.Name != "" {
+		log.Warnf("Overriding KfDef.Spec.Name; old value %v; new value %v", kfDef.Name, appName)
+	}
+	kfDef.Name = appName
```
yanniszark: This line should be inside the if statement.
Now it always overrides.

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+	isValid, msg := kfDef.IsValid()
+	if !isValid {
+		log.Printf("Invalid kfdef: %v", isValid)
+		log.Printf("Error validating generated KfDef, please check config file validity: %v", msg)
+	}
+	cwd, err := os.Getwd()
+	if err != nil {
+		log.Printf("Error getting current working directory: %v", err)
```
yanniszark: Likewise, this error condition doesn't throw an error.

bootstrap/cmd/kfctl/cmd/apply.go
```diff
@@ -39,7 +45,19 @@ var applyCmd = &cobra.Command{
 		if resourceErr != nil {
 			return fmt.Errorf("invalid resource: %v", resourceErr)
```
yanniszark: I believe we want to return a KFError everywhere.
Check out how the coordinator does it: https://github.com/kubeflow/kubeflow/blob/7d976050933d642069591b72138eb3c152224d08/bootstrap/pkg/kfapp/coordinator/coordinator.go#L79

Ideally, we could make a helper function `NewInternalKFError(e error, msg string)` in `pkg/apis/kferrors.go` to reduce boilerplate.

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+	// Check if current directory is empty and if it is error out
+	isEmpty, err := IsCwdEmpty(cwd)
+	if !isEmpty && err != nil {
+		return fmt.Errorf("Current directory is not empty, please try again in an empty directory: %v", err)
+	}
+	kfDef.Spec.AppDir = cwd
+	if kfDef.Spec.AppDir == "" {
+		return errors.New("kfDef App Dir not set")
+	}
+	log.Warnf("App directory name: %v", kfDef.Spec.AppDir)
```
yanniszark: This shouldn't be Warn, since it doesn't indicate that something is out of place.

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+	// Save app.yaml because we need to preserve information about the cache.
+	if err := kfDef.WriteToFile(cfgFilePath); err != nil {
+		log.Errorf("Failed to save KfDef to %v; error %v", cfgFilePath, err)
+		return err
+	}
+	log.Infof("Saved configfile as kfdef in path: %v", cfgFilePath)
+
+	// Load KfApp for Generate and Apply
+	KfApp, KfErr := coordinator.LoadKfAppCfgFile(cfgFilePath)
+	if KfErr != nil {
```
yanniszark: Can you change all of these to just err? (kfErr, generateErr, applyErr, ...)

bootstrap/pkg/kfapp/apply/apply.go
```diff
@@ -0,0 +1,103 @@
...
+	log.Println("Kubeflow Apply...")
+	applyErr := KfApp.Apply(kfResource)
+	if applyErr != nil {
+		log.Println("Unable to apply resources for KfApp", applyErr)
+	}
+	return nil
+}
+
+// IsCwdEmpty - quick check to determine if the working directory is empty
+func IsCwdEmpty(name string) (bool, error) {
```
yanniszark: Can you make this a non-exported function? (non-capital first letter: isCwdEmpty)

swiftdiaries:
Thanks for the review @yanniszark !
Addressed the comments, could you PTAL?

swiftdiaries:
Pre-submits fails for unit tests because I haven't rebased to pickup changes to the go.mod file.

jlewi:
Thanks @swiftdiaries !

How about updating the E2E test to use the new semantics?
https://github.com/kubeflow/kubeflow/blob/489e916f280d8e4fde21daab4e1fb49ee2064d80/testing/kfctl/kfctl_go_test.py#L140

bootstrap/cmd/kfctl/cmd/apply.go
```diff
@@ -35,21 +45,27 @@ var applyCmd = &cobra.Command{
...
-		kfApp, kfAppErr := coordinator.LoadKfApp(map[string]interface{}{})
-		if kfAppErr != nil {
-			return fmt.Errorf("couldn't load KfApp: %v", kfAppErr)
+		if err != nil {
+			return fmt.Errorf("couldn't load KfApp: %v", err)
 		}
-		applyErr := kfApp.Apply(resource)
-		if applyErr != nil {
-			return fmt.Errorf("couldn't apply KfApp: %v", applyErr)
+		err = kfApp.Apply(resource)
```
nrchakradhar: Here kfApp may still be "nil" right?

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,112 @@
+package apply
```
nrchakradhar: At cmd/kfctl/cmd/apply.go will be present and then a new directory cmd/kfctl/cmd/apply/ will contain the bootstrapping. Would this be the format followed for the "new" kfctl? for generate as well?

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,112 @@
...
+		return kfErr
+	}
+	// Once init is done, we generate and apply subsequently
+	log.Println("Kubeflow Generate...")
+	err = KfApp.Generate(kfResource)
+	if err != nil {
+		kfErr.Message = fmt.Sprintf("error generating config for kfapp: %v", err)
+		return kfErr
+	}
+	log.Println("Kubeflow Apply...")
```
nrchakradhar: Does it make sense to have apply not done here but in the caller. The difference in the caller would be how the kfApp is obtained right?

jlewi:
@gabrielwen Could you please review this PR?

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,112 @@
...
+	err = KfApp.Apply(kfResource)
+	if err != nil {
+		kfErr.Message = fmt.Sprintf("error apply config on cluster: %v", err)
+		return kfErr
+	}
+	return nil
+}
+
+// isCwdEmpty - quick check to determine if the working directory is empty
```
gabrielwen: how about [ReadDir](https://golang.org/pkg/io/ioutil/#ReadDir) with `os.Getenv("CWD")`?  also, it's confusing that you return both of boolean and error, could you either just return one of them or make it explicit as what stands for what?

gabrielwen:
Relates to kubeflow/kfctl#20

gabrielwen:
Also, have you considered that `-f` might be a remote address, e.g. an URL?
kubeflow/kfctl#21

swiftdiaries:
Thanks for the review ! @gabrielwen 
Right now, it follows the same pattern for the config flag in kfctl. That is, it can take in local files and raw github URLs.

swiftdiaries:
Also, just asking, should this be moved to the `kubeflow/kfctl` once the comments are addressed?

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,117 @@
...
+	err = KfApp.Apply(kfResource)
+	if err != nil {
+		kfInternalErr.Message = fmt.Sprintf("error apply config on cluster: %v", err)
+		return kfInternalErr
+	}
+	return nil
+}
+
+// isCwdEmpty - quick check to determine if the working directory is empty
+func isCwdEmpty() error {
```
gabrielwen: in the case of returning errors, please use KfError:
https://github.com/kubeflow/kubeflow/blob/5b652b18892c7b49af48ecc365b0af1080e0cb06/bootstrap/pkg/apis/kferrors.go#L49

swiftdiaries: Ok. I can change it out.
But, I'm returning a KfError from where isCwdEmpty is invoked. Is that sufficient?

bootstrap/cmd/kfctl/cmd/apply/apply.go
swiftdiaries: I've done this as suggested

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,112 @@
...
+		return kfErr
+	}
+	if kfDef.Name != "" {
+		log.Infof("Overriding KfDef.Spec.Name; old value %v; new value %v", kfDef.Name, appName)
+	}
+	kfDef.Name = appName
+	isValid, msg := kfDef.IsValid()
+	if !isValid {
+		log.Warnf("Invalid kfdef: %v", isValid)
+		log.Errorf("Error validating generated KfDef, please check config file validity: %v", msg)
```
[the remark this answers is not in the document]
swiftdiaries: Done.

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,112 @@
...
+		kfErr.Message = fmt.Sprintf("Error loading KfDef: %v", err)
+		return kfErr
+	}
+	if kfDef.Name != "" {
+		log.Infof("Overriding KfDef.Spec.Name; old value %v; new value %v", kfDef.Name, appName)
+	}
+	kfDef.Name = appName
+	isValid, msg := kfDef.IsValid()
+	if !isValid {
+		log.Warnf("Invalid kfdef: %v", isValid)
```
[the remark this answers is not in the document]
swiftdiaries: Was following the pattern set in the coordinator, but changed it out to a single KfError

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,112 @@
...
+		log.Errorf("Error validating generated KfDef, please check config file validity: %v", msg)
+	}
+	cwd, err := os.Getwd()
+	if err != nil {
+		kfErr.Message = fmt.Sprintf("Error getting current working directory: %v", err)
+		return kfErr
+	}
+	log.Info("Present working directory is: %v", cwd)
```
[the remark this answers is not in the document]
swiftdiaries: Done.

bootstrap/cmd/kfctl/cmd/apply/apply.go
```diff
@@ -0,0 +1,112 @@
...
+	kfdefsv3 "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis/apps/kfdef/v1alpha1"
+	"github.com/kubeflow/kubeflow/bootstrap/v3/pkg/kfapp/coordinator"
+	log "github.com/sirupsen/logrus"
+)
+
+// Set default app name to kf-app
+const appName = "kf-app"
+
+var kfErr = &kfapis.KfError{
+	Code:    int(kfapis.INTERNAL_ERROR),
```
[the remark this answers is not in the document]
swiftdiaries: Added a bunch of Invalid argument errors for when appropriate.

bootstrap/cmd/kfctl/cmd/apply.go
```diff
@@ -25,8 +27,16 @@ import (
 
 var applyCfg = viper.New()
 
+// ConfigFilePath is a flag for the apply sub-command to take in a config file (i.e KfDef)
+// and bootstrap a KfApp by filling in the necessary fields
+var ConfigFilePath string
+
+// Resource is set to k8s by default
+const resource = kftypes.K8S
```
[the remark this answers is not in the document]
swiftdiaries: I've added back the ktypes.Resources to be supplied from the flags, also for `platform/all` I'm deriving the platform from the config file and adding the apply logic specific to the platform by applying the kftypes.ALL to the apply function.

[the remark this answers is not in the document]
jlewi: We should get rid of platform and k8s.

The new semantics should be

```
kfctl apply -f file
```

swiftdiaries:
___
*[bootstrap/cmd/kfctl/cmd/apply.go, line 62 at r4](https://reviewable.io/reviews/kubeflow/kubeflow/4040#-LoTIhL99TwvJ9ztKX6Y:-LoTR1NlCUBqjb6bH2eE:b-oavha8) ([raw file](https://github.com/kubeflow/kubeflow/blob/4ea2b85b930cb42772230ae755208473e7ba11fd/bootstrap/cmd/kfctl/cmd/apply.go#L62)):*

The whole point of the new semantics is not to have resource arguments any more.

The apply command should always call kfApp.Apply(kftypes.All)

Eventually once @gabrielwen implements the reconcile semantics we will change the function signature to just bye

Apply() 
</blockquote></details>

I was thinking about backwards compatibility when I added the resource arguments back in, how are we going to handle that moving forward? 
This was a compromise to have both the old arguments and support the new semantics.

swiftdiaries:
___
*[bootstrap/cmd/kfctl/cmd/apply.go, line 52 at r4](https://reviewable.io/reviews/kubeflow/kubeflow/4040#-LoTIH695lZT6TNElU14:-LoTRkPg0KezCPY2TIQb:b-8kcz0v) ([raw file](https://github.com/kubeflow/kubeflow/blob/4ea2b85b930cb42772230ae755208473e7ba11fd/bootstrap/cmd/kfctl/cmd/apply.go#L52)):*

Can we just call LoadKfAppCfgFile(cfgFile)?

LoadKfApp is deprecated. That method takes a map of command line options but command line options should no longer be needed because all of the options should be defined in the KfDef file.


</blockquote></details>

I initially tried doing that however it seems there are some fields in the KfDef that are needed before calling LoadKfAppCfgFile(cfgFile) and executing the Generate and Apply functions.
So this fills in the required fields in KfDef (I believe this is what the Init steps does).

swiftdiaries:
___
*[bootstrap/cmd/kfctl/cmd/apply.go, line 35 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/4040#-LnVyMS26_d7AzK8ap7H-r2-35:-LoTSgI2Cnv04OA7CsdG:bec5kiq) ([raw file](https://github.com/kubeflow/kubeflow/blob/fe88587ad31eb763d19f735116d4020d2c1e370d/bootstrap/cmd/kfctl/cmd/apply.go#L35)):*

We should get rid of platform and k8s.

The new semantics should be

```
kfctl apply -f file
```
</blockquote></details>

I was thinking about backwards compatibility when implementing this, maybe we should add a Deprecation Notice and sunset this while maintaining support for both semantics?

gabrielwen:
> Also, just asking, should this be moved to the `kubeflow/kfctl` once the comments are addressed?

I think for now we still develop on kubeflow/kubeflow, couldn't the tracking issue for migrations, though..

yanniszark:
@swiftdiaries thanks for taking the time to incorporate all feedback :smile: 
Let me know when this is ready for a final pass.

jlewi:
Thanks @swiftdiaries it looks like there are still some pending comment threads. Do you think you will be able to take a stab at addressing them early next week?

jlewi:
@swiftdiaries Any update on this? Would be great if we could get this merged this week to keep kfctl on track for 0.7.

swiftdiaries:
Yeah, I'm working on reconciling with the comments and the design doc.
I'll update the PR with something by tomorrow.

swiftdiaries:
Note: Needs rebase to fix tests

swiftdiaries:
This PR is ready for review, PTAL
@gabrielwen @jlewi @yanniszark @kkasravi

swiftdiaries:
Pre-submit seems to be failing for jsonnet tests
Haven't debugged further

jlewi:
Thanks @swiftdiaries I saw the jsonnet presubmit failure in another PR as well. 

Filed #4159 to track

swiftdiaries:
___
*[bootstrap/cmd/kfctl/cmd/apply.go, line 35 at r2](https://reviewable.io/reviews/kubeflow/kubeflow/4040#-LnVyMS26_d7AzK8ap7H-r2-35:-LpV8QmG9oPw6jmeC5yJ:bqu8km7) ([raw file](https://github.com/kubeflow/kubeflow/blob/fe88587ad31eb763d19f735116d4020d2c1e370d/bootstrap/cmd/kfctl/cmd/apply.go#L35)):*
<details><summary><i>Previously, yanniszark (Yannis Zarkadas) wrote…</i></summary><blockquote>

Since right now build and apply reuse the coordinator's init generate and apply, how about passing `all` to those functions and removing then switch to new Build/Apply functions inside the coordinator?

In summary:
* Use kftypes.ALL in the new Build/Apply functions for now (change it once plugins are in).
* Build calls `kfApp.Generate(kftypes.ALL)`.
* Apply calls `kfApp.Generate(kftypes.ALL)`  (if required) and `coordinator.Apply(kftypes.ALL)`.

@jlewi how do we know when to call Generate and when not to call it, to enable the scenario of advanced customization?
Maybe we should be explicit and add a skip-build flag.

Later we can add a dedicated Build func in the coordinator.
</blockquote></details>

We are checking if the directory that we are using is empty or not. Shouldn't that suffice?

jlewi:
@swiftdiaries any chance you'll be able to do another pass on this tomorrow? I'd like to get this merged ASAP to keep 0.7 on track.

swiftdiaries:
Ok, I'll commit changes suggested by yannis sometime tonight. 
So that it can be reviewed tomorrow

jlewi:
@swiftdiaries Please let us know when this is ready for another look.

swiftdiaries:
Could you PTAL?
I'll add the new semantics E2E in a separate PR.

yanniszark:
> so can we just move kfApp.Generate into kfApp.Apply

@jlewi that would change the functionality of the current code path so I'm not sure we should do that yet.

swiftdiaries:
@jlewi  Yeah moving that into the Apply function might affect the current semantics path

swiftdiaries:
___
*[bootstrap/pkg/kfapp/coordinator/coordinator.go, line 475 at r7](https://reviewable.io/reviews/kubeflow/kubeflow/4040#-Lpdbgxm2DEvJsu8MHW8:-LpdgZ5_19SIjjpdF1uy:b-60ud2h) ([raw file](https://github.com/kubeflow/kubeflow/blob/c9707555f89042fa1e75bbc2d965a7b28f9a4099/bootstrap/pkg/kfapp/coordinator/coordinator.go#L475)):*
<details><summary><i>Previously, yanniszark (Yannis Zarkadas) wrote…</i></summary><blockquote>

I believe the code starting here up until the KfDef unmarshaling is the same as the LoadKFDefFromURI function in `bootstrap/pkg/apis/apps/kfdef/v1alpha1/application_types.go`.
We should reuse existing functions as much as possible to avoid code duplication and increase robustness.
</blockquote></details>

There was a TODO on that to move to gogetter.GetAny instead of the GetFile so I implemented that here

swiftdiaries:
___
*[bootstrap/pkg/kfapp/coordinator/coordinator.go, line 447 at r7](https://reviewable.io/reviews/kubeflow/kubeflow/4040#-Lpd_B-v8EPr2s8pl2Ba:-Lpdgk4I3bKNXaA9Cq84:b-41kull) ([raw file](https://github.com/kubeflow/kubeflow/blob/c9707555f89042fa1e75bbc2d965a7b28f9a4099/bootstrap/pkg/kfapp/coordinator/coordinator.go#L447)):*
<details><summary><i>Previously, yanniszark (Yannis Zarkadas) wrote…</i></summary><blockquote>

I would expect this function to return a true/false value, based on its naming.
If we want it to do return the cwd if it's not empty, let's name it appropriately so it reveals what it does (eg getCWD with 3 return values).
</blockquote></details>

yeah that sounds reasonable

yanniszark:
> There was a TODO on that to move to gogetter.GetAny instead of the GetFile so I implemented that here

`GetAny` uses a Client without setting the Pwd, so it won't work for relative paths:

https://github.com/hashicorp/go-getter/blob/3f60ec5cfbb2a39731571b9ddae54b303bb0a969/client.go#L28-L40

https://github.com/hashicorp/go-getter/blob/3f60ec5cfbb2a39731571b9ddae54b303bb0a969/get.go#L92-L99

The LoadKFDefFromURI has implemented this functionality (relative imports).
In any case, we should improve the existing function instead of introducing another one.
Finally, I believe that LoadKFDefFromURI already works for relative paths.

jlewi:
> @jlewi Yeah moving that into the Apply function might affect the current semantics path

@yanniszark and @swiftdiaries can you elaborate on the risk here? The existing semantic path would be

```
kfctl generate ...
kfctl apply ...
```

So kfctl apply handles the case where kfctl generate has already been run; e.g. checks whether that directory already exists then doesn't that provide compatibility with the existing semantics?

If we don't fix it in this PR when would we fix it? Would we continue to track it and make it a blocker of closing #3518 (new semantics for kfctl)?

gabrielwen:
> > There was a TODO on that to move to gogetter.GetAny instead of the GetFile so I implemented that here
> 
> `GetAny` uses a Client without setting the Pwd, so it won't work for relative paths:
> 
> https://github.com/hashicorp/go-getter/blob/3f60ec5cfbb2a39731571b9ddae54b303bb0a969/client.go#L28-L40
> 
> https://github.com/hashicorp/go-getter/blob/3f60ec5cfbb2a39731571b9ddae54b303bb0a969/get.go#L92-L99
> 
> The LoadKFDefFromURI has implemented this functionality (relative imports).
> In any case, we should improve the existing function instead of introducing another one.
> Finally, I believe that LoadKFDefFromURI already works for relative paths.

Supplementing this comment, please use this function:
https://github.com/kubeflow/kubeflow/blob/6a0fd60a8bfec01ebff6ff4d1a2760320b399374/bootstrap/pkg/apis/apps/kfdef/kfloader.go#L197

gabrielwen:
also please be aware that #4166 makes some changes.

swiftdiaries:
Reused existing code and moved Generate from cmd/apply.go into the coordinator. 
Changed E2E to new semantics.
I'll wait for the pre-submits to pass and update.
But the code should be ready for review

swiftdiaries:
Ohh got it. My bad. I'll fix that. 
Thank y'all for the very patient code reviews !

jlewi:
@swiftdiaries please let me know when this is ready for another look.

jlewi:
@swiftdiaries Any chance you can address the last comment so we can get merged?

swiftdiaries:
Presubmit error:
Since the configs are all downloaded from GitHub URLs, the new semantics requires kfctl to be run from a directory that is empty.
Need to change `kfctl_go_test.py` to do that

swiftdiaries:
___
*[testing/kfctl/kfctl_go_test.py, line 204 at r8](https://reviewable.io/reviews/kubeflow/kubeflow/4040#-Lphmjk4-cEHPLdW8Ae3:-Lpq3Gc35y9c-uzVebaH:b8r9mvn) ([raw file](https://github.com/kubeflow/kubeflow/blob/60b92e2b076c97c365b2f4bd7a2e3fba7cb6c9c8/testing/kfctl/kfctl_go_test.py#L204)):*

Why are you calling kfctl apply on parent_dir/tmp.yaml but then dumping  app_path/app.yaml

My expectation is that if you do
kfctl apply -f /some/path/tmp.yaml

Then /some/path should be the directory where KFApp is installed. So you should dump the same file on the next line.
</blockquote></details>

Multiple parts of the code are using the app.yaml as a common entrypoint. The coordinator, kustomize.. They all assume the existence of an app.yaml and write to it. So when I tried writing to the same file in the coordinator I found that it was being created anyway by the inner packages. I'm not really sure what packages are creating/writing to the app.yaml.
That might be a separate refactor in itself. I'll file an issue and follow up with that.

swiftdiaries:
Ok, I was able to reproduce
`gcp.Apply()` fails because `gcp.GetPluginSpec()` returns nil, [source](https://github.com/kubeflow/kubeflow/blob/master/bootstrap/pkg/kfapp/gcp/gcp.go#L794).

jlewi:
Thanks @swiftdiaries on my end I think once the tests are passing this is good to go.

swiftdiaries:
I'm having trouble with the plugins part. 
The PluginSpec is not being retrieved properly from the KfDef: https://github.com/kubeflow/kubeflow/pull/4040#issuecomment-536157691

yanniszark:
@jlewi @swiftdiaries it seems that the scope of this PR has expanded since I last saw it.
I need some more time to review it tomorrow.
From first glance, I don't understand what we need the KfAppContext for.

swiftdiaries:
I needed a way to pass the kfapp information from build to apply. Previously, we assumed the existence of an app.yaml in the directory where generate and apply were called. In the new semantics this is no longer the case. 
So I used a KfAppContext to pass it around.

swiftdiaries:
This also enables the GitOps usecase where we need the config file can be built and applied from any directory using relative paths.

yanniszark:
> I needed a way to pass the kfapp information from build to apply. Previously, we assumed the existence of an app.yaml in the directory where generate and apply were called. In the new semantics this is no longer the case.

I think this is still the case.

| command | config file |
| - | - |
| kfctl build -f <cfg> | app.yaml in cwd |
| kfctl apply -f <cfg> (empty cwd or only app.yaml in cwd) | app.yaml in cwd |
| kfctl apply -f <cfg> (app.yaml in cwd plus other files) | app.yaml in cwd (build has run before) |

This is what the design doc defines.
In what case do you need to know the location between calls?

PS. For the last case, it seems better to me to skip the -f flag (since the config is already defined as app.yaml in the cwd) and use that to differentiate whether to skip build or not.
/cc @jlewi

swiftdiaries:
@yanniszark We have a different of interpretation in what the semantics are. 
I initially assumed that the cwd would be the default for all implementations.

Then from @jlewi 's review comments:
> Why are you calling kfctl apply on parent_dir/tmp.yaml but then dumping app_path/app.yaml

> My expectation is that if you do
> kfctl apply -f /some/path/tmp.yaml

> Then /some/path should be the directory where KFApp is installed. So you should dump the same file on the next line.

I changed it so that the `cwd` is NOT the directory where KFApp is installed. 

The previous implementation followed [this](https://github.com/kubeflow/kubeflow/pull/4040#issuecomment-537157398).

yanniszark:
@jlewi this is different [from the examples in the design doc](https://docs.google.com/document/d/1kviNOdjoBPeoJ56KBvMKWq7SzrSN5rFUQoULK8muD-M/edit?pli=1#).
Am I missing something?

swiftdiaries:
Or I might've misunderstood the comments.

jlewi:
@yanniszark and @swiftdiaries I'm not sure I understand the confusion. @yanniszark I don't think the commands you listed in your comment The assumption is user can use pipeline with only default namespace edit role + the custom role. reflect the doc. An argument to -f is always required. This matches the 4 use cases listed in the doc.

The only difference is how we infer the ${KFAPP} dir. 

```
kfctl apply -f /some/local/file/path/kfdef.yaml   
```

* In this case since we have a local file path we should infer ${KFAPP} from the the -f argument
* So ${KFAPP} should be /some/local/file/path

In the case of a remote URI


```
kfctl apply -f https://some/path/kfdefl.ayml
```

* We should use ${CWD} as ${KFAPP}

jlewi:
@swiftdiaries Ok I think I understand what the issue is.

If we do 

```
kfctl apply -f /my/kfapp/mykfdef.yaml
```

The code will use KFAPP=/my/kfapp as the directory.

The original code would also write the updated YAML spec  to /my/kfapp/app.yaml  (and not /my/kfapp/mykfdef.yaml).

This is why the test was dumping /my/kfapp/app.yaml.

I think that would have been fine for now but the desired long term semantics would be that kfctl updates /my/kfapp/mykfdef.yaml  instead.

IIUC that is now fixed by KfAppContext which allows the location of the KFAPP file to be passed around so that any code which needs to update the file knows the location.

jlewi:
In the interest of getting this PR merged and moving forward I'd like to propose the following

1. Revert the KfAppContext changes
    * go back to the commit I commented on three days ago where I said if the tests are passing then we can fix edge cases in the future
    * I think in this PR its ok if user calls kfctl apply -f /some/kfapp/myspec.yaml and the code updates /some/kfapp/app.yaml
    * We can fix this in the future
    * @gabrielwen is introducing a KfConfig to be used as internal datastructure. We can actually use that class to keep track and pass along the appropriate location of the KFDef file so that we update the user supplied location rather than the app.yaml file

jlewi:
@swiftdiaries WDYT? Could we get that done today? I think its urgent we get this PR merged.

swiftdiaries:
Yeah, that seems reasonable. I'll look at using the KfConfig to update the right file.

jlewi:
/assign @yanniszark 
/assign @jlewi 

Thanks @swiftdiaries for the quick fix

/lgtm
/approve
/hold

LGTM assuming tests pass on latest revision. 

@yanniszark Can you please prioritize this if you want to have another look?

yanniszark:
@jlewi in the design doc use cases, we always do `cd ${MYAPP}`.
Because of this, my understanding was that we always use the current directory and not infer anything from the URI given by the user.

As with other things, I believe it's better to be explicit here as well.
Inferring the KFApp directory based on the URI of the config file is confusing (and more complex/fragile in the code) imo.

In addition, I think it's logical to store KfDef in app.yaml, even if we make a copy of the user supplied file. The user supplied file is like the file in your computer when you do `kubectl apply -f`. The app.yaml file is like the object stored in etcd.

To wrap it up, I believe once we are clear about the semantics, this PR will become a lot easier to review:
* Will each ${KFAPP} directory have the config in `app.yaml`?
* Will ${KFAPP} be inferred from the location of the config file? Or will we be explicit and introduce an argument/flag in case the user doesn't want to use the current directory?

jlewi:
@yanniszark Using the directory of app.yaml as the directory of KFAPP matches whats in the design doc examples. Specifically under Case 2 

> kfctl uses the directory of app.yaml to store any files it needs

The reason we do a cd to $KFAPP in that example is because we are curling the YAML def into app.yaml and the way that command is written it depended on being in the directory ${KFAPP}.

> As with other things, I believe it's better to be explicit here as well.

Isn't using the current working directory as ${KFAPP} non explicit?

The problem with using the current working directory as ${KFAPP} is what if the current working directory doesn't match the directory that app.yaml is installed. e.g.

```
cd /some/dir
kfctl apply -f /some/other/dir/app.yaml
```

So if kfctl uses the current working directory then it will treat directory /some/dir as the directory but this is likely wrong. Since app.yaml lives in /some/other/dir, the config files exist there.

Now suppose the user does the following

```
mkdir /apps/myapp
cd /apps/myapp
curl https://.../ > app.yaml
kfctl build -f app.yaml
edit configs
kfctl apply -f app.yaml
```

* So at this point they've deployed KF using the customized configs in /apps/myapp

Now suppose they do

```
cd /home/some/dir
kfctl apply -f /apps/myapp/app.yaml
```

* The meaning of this should be reapply app.yaml with the exact settings
* If we infer ${KFAPP} from app.yaml then we set ${KFAPP} to /apps/myapp/app.yaml 
* kfctl will then pick up the custom configs and this is an idempotent operation.

On the other hand if we use the current working directory then kfctl would now treat KFAPP as /home/some/dir. Since the configs don't exist in that directory we would end up generating new configs and applying them. We would then override the custom configurations. 

> In addition, I think it's logical to store KfDef in app.yaml, even if we make a copy of the user supplied file. The user supplied file is like the file in your computer when you do kubectl apply -f. The app.yaml file is like the object stored in etcd.

I think having two config files will lead to confusion and other problems which is why we should eventually get rid of that.

Suppose we have

```
${KFAPP}/original.yaml
${KFAPP}/app.yaml
```

If a user wants to reapply their deployment (e.g. suppose some resource got deleted and they need to recreate) which yaml file should they use?

original.yaml is out of date so using that file could be problematic.

For example when a person first calls apply on original.yaml we will attempt to fill in the spec with appropriate context dependent values. These values are then persisted in the saved copy of the spec; i.e. app.yaml. 

For example when apply is first called on kfctl_gcp_iap.yaml if project isn't set we will set the default based on the gcloud default. This is then persisted in app.yaml. So at this point calling apply on app.yaml is idempotent because the auto-determined values have been persisted there.

However, suppose the user were to reinvoke kfctl apply but this time using original.yaml. At this point we would try to fill in the default values again and we might not get the same values as when we first ran apply which could cause problems.

yanniszark:
@jlewi I see what you mean.
In my mind, `-f` was always a remote source, so now it makes more sense.
What will the logic be if the config file is not a filesystem path?
Will it fallback to the current directory or will we disable non-filesystem paths and instruct users to use curl?
What should I expect to see in this PR?

jlewi:
@yanniszark I think of 

```
kfctl apply -f http://host/kfdef.yaml
```

as short hand for the following

```
curl http://host/kfdef.yaml > ./app.yaml
kfctl apply -f ${PWD}/app.yaml
```

So under the hood we are creating app.yaml in the current working directory and then using that path. 

The expectation is that your current working directory is empty and I believe @swiftdiaries already added a check for that. This guards against people accidentally overwriting any files in their current path.

jlewi:
> What should I expect to see in this PR?

I think this is covered pretty well by the E2E test which is doing the following

```
util.run([kfctl_path, "apply", "-V", "-f=", + os.path.join(parent_dir, "tmp.yaml")], cwd=app_path)
```

The goal of this PR is precisely to enable the ability to do 

```
kfctl apply -f /some/path/kfdef.yaml
```

And the E2E test is verifying that's working as expected.

 The E2E test could be more comprehensive in terms of testing all the various edge cases. But this PR has already been open for a while and its blocking other stuff (e.g. upgrades) so I'd prefer to iterate in follow on PRs.

swiftdiaries:
Right now, it defaults to the current working directory for URLs and local filepaths.
We check to ensure that the current directory is not empty. 
The E2E doesn't seems to run the 
`util.run([kfctl_path, "apply", "-V", "-f=", + os.path.join(parent_dir, "tmp.yaml")], cwd=app_path)`
line.

yanniszark:
@jlewi what about the case of build?
kfctl build -f / kfctl apply -f with a remote url?
Build will use the remote url, but does it make sense to call apply with the remote url again?
Calling apply with the remote URL will fail, as apply will think that the directory is not empty.
So you end up having different UX for remote vs local.

Local:
kfctl build -f <local>
kfctl apply -f <local>

Remote:
kfctl build -f <remote>
kfctl apply -f <local>

jlewi:
@swiftdiaries pytest is weird in how it formats the errors and logs. You have to scroll up in the logs to where pytest prints out information to see the error that caused the test to abort

```
     logging.info("Running kfctl with config:\n%s", yaml.safe_dump(config_spec))
>     util.run([kfctl_path, "apply", "-V", "-f=", + os.path.join(parent_dir, "tmp.yaml")], cwd=app_path)
E     TypeError: bad operand type for unary +: 'str'
kfctl_go_test.py:217: TypeError
```

The problem is the "+" before os.path.join. You don't need the "+" because it should just be elements in the list.

swiftdiaries:
Oh yeah, that's why I used the KfAppContext
So, you can do 
`kfctl build -f`
It'll store the AppDir and the KubeContext to the KfAppContext
Then you'll do 
`kfctl apply`
Which will apply based on the current context
This also addresses apply-delete skew, there was an open issue for that

swiftdiaries:
Not sure if that's the right way to do it
But I thought it'd be something like the KubeContext

jlewi:
@swiftdiaries I haven't looked at your changes yet but I'm confused about your comment that the bug needs KFContext or KFContextHere. The error I posted above is due to a bug in kfctl_go_test.py; its not an issue in the implementation of kfctl

you just need to change

```
util.run([kfctl_path, "apply", "-V", "-f=", + os.path.join(parent_dir, "tmp.yaml")
```

to

```
util.run([kfctl_path, "apply", "-V", "-f=", os.path.join(parent_dir, "tmp.yaml")
```

We may not need to introduce a KFContext or KFAppContext here.
Because in #4210 @gabrielwen introduced a new structure KFConfig that will act as an internal datastructure for KFDef to deal with version change.

So the plan is to change references in coordinator.go and the plugins to *kfdefs.KFDef to *kfdefs.KFConfig.

So KFConfig can carry any contextual information that needs to be passed along to the plugins. So we could store the app dir and file location in that and use that to pass it along.

jlewi:
@swiftdiaries oh I'm guessing your comments about the context were a reply to the discussion between me and @yanniszark about the semantics and not the test failure.

jlewi:
/lgtm
/approve

CI infrastructure is having some trouble; more discussin #buildcop

jlewi:
/test all

@swiftdiaries Can you rebase off master to pull in changes/fixes to the ci system?

swiftdiaries:
Thank you @jlewi

richardsliu:
@swiftdiaries From the failure logs it looks like the zone (required for GCP) is not passed:
```
util.py                     71 INFO     time="2019-10-03T02:25:24Z" level=error msg="Updating myapp2-storage error: &{Code:CONDITION_NOT_MET Location:/deployments/myapp2-storage/resource
s/myapp2-storage-metadata-store->$.properties Message:error: instance type (null) does not match any allowed primitive type (allowed: [\"string\"])\n    level: \"error\"\n    schema: {\"
loadingURI\":\"#\",\"pointer\":\"/insert/properties/zone\"}\n    instance: {\"pointer\":\"/zone\"}\n    domain: \"validation\"\n    keyword: \"type\"\n    found: \"null\"\n    expected:
[\"string\"]\n ForceSendFields:[] NullFields:[]}" filename="gcp/gcp.go:352"
util.py                     71 INFO     time="2019-10-03T02:25:24Z" level=error msg="Updating myapp2-storage error: &{Code:CONDITION_NOT_MET Location:/deployments/myapp2-storage/resource
s/myapp2-storage-artifact-store->$.properties Message:error: instance type (null) does not match any allowed primitive type (allowed: [\"string\"])\n    level: \"error\"\n    schema: {\"
loadingURI\":\"#\",\"pointer\":\"/insert/properties/zone\"}\n    instance: {\"pointer\":\"/zone\"}\n    domain: \"validation\"\n    keyword: \"type\"\n    found: \"null\"\n    expected:
[\"string\"]\n ForceSendFields:[] NullFields:[]}" filename="gcp/gcp.go:352"
util.py                     71 INFO     Error: couldn't apply KfApp:  (kubeflow.error): Code 500 with message: coordinator Apply failed for gcp:  (kubeflow.error): Code 400 with message:
 gcp apply could not update deployment manager Error could not update deployment manager entries; Updating myapp2-storage error(412): PRECONDITION FAILED
```
How do you process the zone setting?

jlewi:
Most recent failure is 

```
util.py                     71 INFO     time="2019-10-03T02:26:01Z" level=error msg="Updating myapp2-storage error: &{Code:CONDITION_NOT_MET Location:/de
ployments/myapp2-storage/resources/myapp2-storage-metadata-store->$.properties Message:error: instance type (null) does not match any allowed primitive t
ype (allowed: [\"string\"])\n    level: \"error\"\n    schema: {\"loadingURI\":\"#\",\"pointer\":\"/insert/properties/zone\"}\n    instance: {\"pointer\"
:\"/zone\"}\n    domain: \"validation\"\n    keyword: \"type\"\n    found: \"null\"\n    expected: [\"string\"]\n ForceSendFields:[] NullFields:[]}" file
name="gcp/gcp.go:352"
util.py                     71 INFO     time="2019-10-03T02:26:01Z" level=error msg="Updating myapp2-storage error: &{Code:CONDITION_NOT_MET Location:/de
ployments/myapp2-storage/resources/myapp2-storage-artifact-store->$.properties Message:error: instance type (null) does not match any allowed primitive t
ype (allowed: [\"string\"])\n    level: \"error\"\n    schema: {\"loadingURI\":\"#\",\"pointer\":\"/insert/properties/zone\"}\n    instance: {\"pointer\"
:\"/zone\"}\n    domain: \"validation\"\n    keyword: \"type\"\n    found: \"null\"\n    expected: [\"string\"]\n ForceSendFields:[] NullFields:[]}" file
name="gcp/gcp.go:352"
util.py                     71 INFO     Error: couldn't apply KfApp:  (kubeflow.error): Code 500 with message: coordinator Apply failed for gcp:  (kubefl
ow.error): Code 400 with message: gcp apply could not update deployment manager Error could not update deployment manager entries; Updating myapp2-storag
e error(412): PRECONDITION FAILED
```

I think the problem is kfctl_go_test.py isn't setting the zone in the KFDef config file.

The old version of the test was setting email and zone via command line flags

```
  # We need to use retries because if we don't we see random failures	
  # where kfctl just appears to die.	
  run_with_retries([	
      kfctl_path, "generate", "-V", "all", "--email=" + email, "--zone=" + zone	
  ],	
                   cwd=app_path)	
```

I think with the new code you may need to open up the KFDef file and set email and zone.

jlewi:
Latest deployment failure is

```
util.py                     71 INFO     The StatefulSet "application-controller-stateful-set" is invalid: spec: Forbidden: updates to statefulset spec fo
r fields other than 'replicas', 'template', and 'updateStrategy' are forbidden
```

Possibly related to kubeflow/manifests#410 ?

jlewi:
It looks like our periodic job ran successfully not long ago
https://k8s-testgrid.appspot.com/sig-big-data#kubeflow-periodic-master
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kubeflow-periodic-master/1179622741322502146

jlewi:
@swiftdiaries here are the logs
[kubeflow-presubmit-kfctl-go-iap-istio-4040-58f9c38-4880-a17a-1153927365.log.txt](https://github.com/kubeflow/kubeflow/files/3685045/kubeflow-presubmit-kfctl-go-iap-istio-4040-58f9c38-4880-a17a-1153927365.log.txt)

I don't see any evidence that the application controller gets deployed twice.

Are you able to manually deploy?

swiftdiaries:
I can't manually deploy using both the manifests. I'm trying an older version to see if that goes thorugh

swiftdiaries:
I was able to manually deploy with an older KfDef and Workload Identity disabled.

swiftdiaries:
For some reason everything is applied twice. Checking again.

swiftdiaries:
Before we used to set the name from the kfapp directory name that we create and that's unique.
But now since we're using the current working directory we skip that part and we end up deploying to the same cluster repeatedly ! 🤦‍♂

swiftdiaries:
Hopefully this passes 🤞

richardsliu:
/lgtm
/approve
@swiftdiaries will follow up with a fix for https://github.com/kubeflow/kubeflow/issues/4227.

swiftdiaries:
Thanks for the patient reviews all. 
I'll fix edge cases in followups.

PR #4050, opened 2019-09-03
Use info message when waiting for the namespace

vjrantal:
During a normal installation flow, the default profile creation
waits for the target namespace to be created and only after that
continues with the creation. Before this commit, user would see
a warning message about a namespace not found, but not anything
indicating whether the system was able to recover.

With this commit, there will be informational logging until the
namespace is available and a final informational message once
the namespace is found which means that in a normal installation
(without the -V flag) there won't be warnings presented to the user.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4050)


yanniszark:
@vjrantal I agree that an Info message is more suitable.
/lgtm
/approve

vjrantal:
/retest

The error I see in the logs are like:

```
util.py                     71 INFO     fatal: unable to access 'https://git.apache.org/thrift.git/': Failed to connect to git.apache.org port 443: Connection timed out
```

Let's see if the same occurs when retesting.

yanniszark:
@vjrantal Apache Git seems down atm (https://status.apache.org/incidents/63030p4241xj)

vjrantal:
@yanniszark Thanks for the info! I will trigger a retest tomorrow.

PR #4082, opened 2019-09-05
update auto redirect; add name regex check

kunmingg:
- update auto redirect target image
- check deploy name input format

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4082)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -148,6 +148,7 @@ const MYSQL = 'mysql';
 const PASSWORD = 'password';
 const PROFILES = 'profiles';
 const SPARTAKUS = 'spartakus';
+const nameformat = '[a-z]([-a-z0-9]*[a-z0-9])?';
 
 const styles: {[key: string]: React.CSSProperties} = {
     btn: {
@@ -770,7 +771,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
             this._appendLine('Validating if IAP is up and running...');
             const startTime = new Date().getTime() / 1000;
             const img = document.createElement('img');
-            img.src = dashboardUri + 'assets/kf-logo_64px.svg' + '?rand=' + Math.random();
+            img.src = dashboardUri + 'assets/favicon-32x32.png' + '?rand=' + Math.random();
             img.id = 'ready_test';
             img.onload = () => {
                 window.location.href = dashboardUri;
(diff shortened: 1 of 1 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -784,10 +785,10 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
                     if (ready_test != null) {
                         setTimeout(() => {
                             // We rotate on image addresses of v0.4 and v0.5+ t to support both v0.4 and v0.5+
-                            if (ready_test.src.includes('hub/logo')) {
+                            if (ready_test.src.includes('favicon')) {
```
abhi-g: do we still need the if else test? Looks like with the line 774 update, favicon is always going to be in the src?

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -784,10 +785,10 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
                     if (ready_test != null) {
                         setTimeout(() => {
                             // We rotate on image addresses of v0.4 and v0.5+ t to support both v0.4 and v0.5+
```
abhi-g: may be remove the comment if we end up getting rid of the if/else block

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -884,6 +885,15 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
             });
             throw err;
         }
+        const filtered = this.state[deploymentNameKey].match(nameformat);
+        if (!(filtered && this.state[deploymentNameKey] === filtered[0])) {
+            this.setState({
+                dialogAsCode: false,
+                dialogBody: 'Deployment name need to match rgular expression: [a-z]([-a-z0-9]*[a-z0-9])?',
```
abhi-g: s/rgular/regular/

more importantly, should we make this more human readble. ie., name needs to be 25 chars or less of lowercase alphanumeric characters and hyphens.

components/gcp-click-to-deploy/src/DeployForm.tsx
kunmingg: We might have to iterate for v0.5 compatibility, v0.6 and v0.5 have different image location

components/gcp-click-to-deploy/src/DeployForm.tsx
kunmingg: updated

PR #4084, opened 2019-09-06
Remove kfctl.sh

holdenk:
Remove the broken `kfctl.sh` script, folks should use the go program now instead.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4084)


jlewi:
Thanks @holdenk for picking this up.

For scripts/gke I think we should be a bit conservative and only delete scripts/gke/util.sh. The other scripts aren't directly tied to ksonnet I don't think and might still be used in a couple places.

holdenk:
Sounds reasonable, I didn't see any refs to them so I figured it was OK to clean up, I'll revert that tomorrow and just cleanup the util side of the gke stuff.

jlewi:
Thanks @holdenk;

BTW I generally skip over PRs with WIP in the title so when its ready to review please remove WIP from the title.

jlewi:
Thanks @holdenk 
/lgtm
/approve

PR #4089, opened 2019-09-06
set email explicitly for v0.5

kunmingg:
For v0.5 request we need to get email from signed in identity.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4089)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -693,10 +693,11 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
 
         let buildEndpoint: string;
         let requestBody: string;
+        const email = await Gapi.getSignedInEmail() || '';
         if (this.state.kfversion === Version.V05) {
             // buildEndpoint = `${this._configSpec.appAddress}/kfctl/e2eDeploy`;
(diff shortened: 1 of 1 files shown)


prodonjs:
Since we already try to retrieve the `email` value from the component's state, what if we just call `setState({email})` after line [443](https://github.com/kubeflow/kubeflow/blob/1198c95614ceeb1324bc1688281b1647a8dd90df/components/gcp-click-to-deploy/src/DeployForm.tsx#L443) since it's always going to be called before the request is built and sent.

abhi-g:
Please address Jason's comment above

kunmingg:
@prodonjs 
updated, thanks

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -441,6 +441,9 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
         }
 
         const email = await Gapi.getSignedInEmail() || '';
+        this.setState({
+            ['email']: email,
```
avdaredevil: Kuming, you could write this whole blurb as:

```suggestion
            this.setState({email})
```

avdaredevil: Reference: https://alligator.io/js/object-property-shorthand-es6/

PR #4102, opened 2019-09-09
Fixes regex used in namespace registration

### LINKED ISSUE #4101: Namespace registration fails when userDetails include a dot
/kind bug

During Namespace registration the name of the namespace is being retrieved from the email address and the dot is not being replaced with a hyphen causing.  The profile is created but the namespace isn't due to it not being a valid.

I expected my email name joel.cooklin to be replaced with joel-cooklin as the namespace it would create.

The regex at https://github.com/kubeflow/kubeflow/blob/5c4fc0b0afab30b7a43dae534db8d3a633ccf214/components/centraldashboard/public/components/registration-page.js#L59 needs to be changed.  I used `/[^\w]|_|\./g` to fix it.  The `^` operator was previously excluding `\.`.

**Environment:**

  - KF .62 
  - centraldashboard (v20190909-v0.6.0-rc.0-89-g5b652b18)

jcooklin:
Fixes #4101 

During namespace registration users with dots in their email name get an error due to the profile being created but the namespace failing to create since the dot fails validation.


swiftdiaries:
Hey @jcooklin !
Thank you for your PR! And welcome to the Kubeflow community :) 
/ok-to-test

components/centraldashboard/public/components/registration-page.js
```diff
@@ -56,7 +56,7 @@ export class RegistrationPage extends utilitiesMixin(PolymerElement) {
     _onUserDetails(d) {
         this.namespaceName = this.userDetails
             // eslint-disable-next-line no-useless-escape
-            .replace(/[^\w\-\.]|_/g, '-')
+            .replace(/[^\w]|_|\./g, '-')
```
avdaredevil:
```suggestion
            .replace(/[^\w\-]/g, '-')
```

Issue #4056, opened 2019-09-04
kfctl implements reconcile semantics


jlewi:
/kind feature

**Why you need this feature:**
See design doc for kfctl v1: #3709

One of the proposal is for kfctl to have reconcile like implementation to deal with complex ordering of operations.

Filing this issue to track implementation.

jlewi:
Bump to P0 because this must land in 0.7

jlewi:
@gabrielwen Any update on this do you have an ETA?

Based on some of our discussions

* #4106 uses errors to determine whether apply is complete or needs to be reinvoked
* I think we want to change that so we start using conditions so that we don't just swallow errors and keep retrying for ever
* I think we need to do that as part of #4141 moving pod default creation into the gcp plugin
  * We don't want the pod default creation logic to just return an error if we haven't created the kubeflow namespace yet; instead we should set conditions such that the coordinator.apply knows that its not complete and needs to be reinvoked
* We will likely want to add conditions to track that GCP deployment is done so that we can skip it the next time we invoke gcp.go otherwise it will slow things down
  * We might want to keep track of the time we last checked GCP deployment so that if it exceeds some amount of time on reapply it would end up get reinvoked
   * We can do that in a follow on PR

gabrielwen:
@jlewi 
- #4166 provides functionality of using conditions with KfDef.
- after #4166 I'm going to issue another PR to use conditions during Reconcile.
- after the PR mentioned above, #4141 will be blocked as well as other PRs need reconcile semantics.
- level based executions not started but I think it's not 0.7 release blocking, is it?

gabrielwen:
Discussed offline and this is implementation detail to KFCTL.  Since this is not affecting semantics we have to roll out in v0.7, this is not release blocking.  making it p1 instead of p0.

jlewi:
reconcile semantics will be needed for the kfctl operator; see kubeflow/kubeflow#4570.

jlewi:
/lifecycle frozen
Related to kubeflow/kfctl#193 operator roadmap

PR #4106, opened 2019-09-10
Apply with strawman reconcile

gabrielwen:
Relates to #4056
Add a simple loop to run Apply in coordinator that keeps retry until the op succeeds.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4106)

### DIFF
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -24,6 +24,7 @@ import (
 
 	"os"
 
+	"github.com/cenkalti/backoff"
 	"github.com/ghodss/yaml"
 	"github.com/kubeflow/kubeflow/bootstrap/v3/config"
 	kfapis "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis"
@@ -38,6 +39,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	valid "k8s.io/apimachinery/pkg/api/validation"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"time"
 )
 
 // Builder defines the methods used to create KfApps.
@@ -206,6 +208,13 @@ func repoVersionToUri(repo string, version string) string {
 	tarballUrl := "https://github.com/kubeflow/" + repo + "/tarball/" + version + "?archive=tar.gz"
 
 	return tarballUrl
+}
+
+// A strawman approach for reconcile semantics. We keep retrying until the fn returns nil.
+func simpleReconcile(fn func() error) error {
+	return backoff.Retry(func() error {
+		return fn()
+	}, backoff.NewConstantBackOff(10*time.Second))
 }
 
 // CreateKfDefFromOptions creates a KfDef from the supplied options.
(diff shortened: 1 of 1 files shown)


bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -774,24 +783,32 @@ func (kfapp *coordinator) Apply(resources kftypesv3.ResourceEnum) error {
...
 	switch resources {
 	case kftypesv3.ALL:
-		if err := platform(); err != nil {
-			return err
-		}
-		if err := k8s(); err != nil {
-			return err
-		}
-		return gcpAddedConfig()
+		return simpleReconcile(func() error {
```
jlewi: Should this loop be implemented in Apply?
https://github.com/kubeflow/kubeflow/blob/7789ce18a468b255739b62ada53ecf14924b37a0/bootstrap/pkg/kfapp/coordinator/coordinator.go#L777

Also I think you need to think through the exit conditions and continuation behavior.

e.g. how does Platform () indicate that it has reached a terminal state and doesn't need to be invoked again?

Consider the following use case. The GCP plugin wants to have logic that can only be executed once the K8s namespace exists.

So the first pass through the loop the GCP plugin/app would deploy a GKE cluster. But since the namespace and Kubeflow haven't been deployed yet; we can't initialize a default profile in a GCP specific way.

If it returns an error then simpleRecocile would return and we would just retry calling platform(). But that isn't what we want because the namespace still won't exist.

So we need to call k8s() which should succeed and then we want to retry.

gabrielwen: sounds good, let me modify this.

gabrielwen: @jlewi could you take another look?  I tried to make it simple unless you want me to implement the full version of Reconcile in v1beta1.

gabrielwen:
> Why do we need `requests[idx].Requeue` ?
> 
> This seems unnecessary with a level-based design of a controller. As a result, on each iteration apply will not be called for items that have been completed.
> 
> On each iteration it should be perfectly fine to call apply for each kfapp/plugin. A level based design for kfapp/plugin should mean that invoking apply again should be a null op.
> 
> Furthermore, any status information should be captured in KfDef.Status using conditions. So if the GCP plugin wants to track status to prevent expensive operations (e.g. calling DM again) then it should store that info in status.

yea I agree.  `Requeue` is just a small optimization.  I can remove it if you think so.

jlewi:
@gabrielwen I think your right and I'm being overly prescriptive; apologies for that.

This looks good to me.

/lgtm
/approve

/hold Did you want to submit this as is or were there more changes to make before merging this?

What are the next steps to enable a plugin to take of the reconcile pattern? For example, what would it take to move gcpAddedConfig into the the GCP plugin?

gabrielwen:
> @gabrielwen I think your right and I'm being overly prescriptive; apologies for that.
> 
> This looks good to me.
> 
> /lgtm
> /approve
> 
> /hold Did you want to submit this as is or were there more changes to make before merging this?
> 
> What are the next steps to enable a plugin to take of the reconcile pattern? For example, what would it take to move gcpAddedConfig into the the GCP plugin?

`/hold` will not work if you put anything behind it.
The idea is to move the function calls in `gcpAddedConfig` back to `gcp.Apply` and I'll make another PR for that.  I think after we move them back to `gcp.Apply` it'll make `gcp.Apply` fail the first round and succeed next round.

PR #4115, opened 2019-09-12
Support Kubeflow upgrades in kfctl (pt 1)

richardsliu:
Design: https://bit.ly/2ZWmHJ9 

Usage:
```
kfctl apply -f <update config file path> -V
```

Or as two separate steps:
```
kfctl build -f <update config file path> -V
kfctl apply -f <update config file path> -V
```

The update config file looks like:
```
apiVersion: kfupgrade.apps.kubeflow.org/v1alpha1
kind: KfUpgrade
metadata:
  name: update-kf-upgrade
spec:
  currentKfDef:
    name: kf101
  newKfDef:
    name: kf101
    version: master
  baseConfigPath: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_gcp_iap.yaml
```

The tool infers which KfDef to use based on the name and version provided. The logic is:
1. If `newKfDef` is found, use it as the upgrade app;
2. If not, use the base config path to build it. Merge in existing customizations from `currentKfDef`.

 

TODO:
- Delete expired/outdated resources using labels

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4115)


jlewi:
@richardsliu Why do we need the "update" sub command? Isn't the update redundant with the spec file. The semantics should mirror kubectl semantics; e.g.

```
kubectl apply -f upgrade.yaml
```

or

```
kubectl build -f upgrade.yaml
kubectl apply -f upgrade.yaml
```

jlewi:
Thanks @richardsliu

kunmingg:
We might need to check k8s master version:
https://github.com/kubeflow/kubeflow/issues/4150

jlewi:
@richardsliu Is this ready for another look?

richardsliu:
@jlewi Not yet

richardsliu:
@jlewi @swiftdiaries I think this PR needs to follow https://github.com/kubeflow/kubeflow/pull/4040 in order to not conflict with the new kfctl semantics.

richardsliu:
@jlewi This is still pending a few in-flight PRs, but I've added:
- A kfupgrade app for handling upgrade logic;
- New semantics for the KfUpgrade spec (finding and resolving KfDef references)
- Combined the upgrade logic into build and apply (temporary changes made to generate and apply; these will most likely be updated again after #4040 merges)
- Unit tests

jlewi:
Thanks @richardsliu For future reference remove WIP from the title when the PR is ready for another look otherwise I will continue to ignore it.

jlewi:
@richardsliu Any update on this? Is this blocked by #4040?

richardsliu:
@jlewi This is ready for another pass. 

I've integrated the changes after #4040 is merged. Now the semantics are consistent. But a few things are still missing:
1. Deletion of obsolete resources is not implemented. It was pending on the manifests changes for app instance (now reverted).
2. There are some changes to components that broke upgradability, for example:
```
The StatefulSet "application-controller-stateful-set" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden

The Deployment "centraldashboard" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"centraldashboard", "kustomize.component":"centraldashboard"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable

The Deployment "metadata-deployment" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"kustomize.component":"metadata", "component":"server"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

I think updates to label selectors are disallowed. We'll have to figure out some way to fix this.

krishnadurai:
@richardsliu Seems like the old Katib v1alpha2 CRDs are incompatible with v1alpha3; i.e. in case we try to apply v1alpha3 CRs to existing clusters with v1alpha2 CRs - we get this error:
```
CustomResourceDefinition.apiextensions.k8s.io "experiments.kubeflow.org" is invalid: spec.version: Invalid value: "v1alpha3": must match the first version in spec.versions
```

In the design doc for upgrades we have mentioned that CRs should be backwards compatible:

```
Controllers such as tf-job-operator needs to ensure that multiple versions of the custom resource can be served, and the APIs are backward compatible. See the appendix for more details.
```

@johnugeorge @hougangliu @YujiOshima

richardsliu:
@johnugeorge Can katib-v1alpha3 support v1alpha2 resources?

johnugeorge:
@richardsliu, if users have configured metric collectors in v1alpha2, it will break. It is non compatible

jlewi:
@richardsliu nag - please remove WIP from the title when a PR is ready to review.

To deal with immutable fields do we need to delete the old application before we create the new one?

That's not ideal but maybe that's good enough to start with.

jlewi:
@richardsliu is the PR description still up to date or does it need to be updated to reflect the new CLI semantics?

No major comments on my end. The code will need to be updated to handle conversion from v1alpha1 to v1beta1 but I'd say leave that to later. 

Perhaps we should start with v1beta1 (e.g. only support upgrading from a v1beta1 spec to another v1beta1 spec). We could then maybe extend that later to support v1alpha1 to v1beta. 

The code that would change would be the generating of the new KFDef spec from the old KFDef spec. Supporting v1alpha1 to v1beta1 shouldn't be too difficult to add later because we will have converters from v1alpha1 to v1beta1 so we can just run that conversion before invoking the upgrade code.

richardsliu:
@jlewi This should be ready (synced to the latest semantics version). 

Pending work:
- Resource cleanup (pending on app instance changes and manifests versioning)
- E2E test
- Documenting the requirements and process

jlewi:
Thanks @richardsliu  Is the PR description still up to date? 

```
kfctl update build --config=<new config file path> -V
kfctl update apply update.yaml -V
```

Is there really an update command ?

I thought we changed it to 

```
kfctl build -V -f kfupgrade.yaml
kfctl apply -f kfupgrade.yaml
```

jlewi:
Thanks two quick changes and then I think its ready to go

1. Semantics for setting appDir in NewLoadKFDefFromURI I think need to be tweaked; see inline comment
1. PR description needs to be updated to reflect correct semantics.

jlewi:
p.s kubeflow/kubeflow#4191 is my preliminary work on an E2E test for kfctl_upgrade.

richardsliu:
@jlewi Fixed both.

PR #4120, opened 2019-09-13
[CentralDashboard v2] MUI Front-end tests

avdaredevil:
## About
- Added all tests for Manage Users View:
  - Error State
  - Cluster Admin
  - No data
  - Normal View
  - Add Contrib
  - Remove Contrib

### Meta
/area centraldashboard
/area front-end
/priority p1
/assign @avdaredevil
/cc @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4120)

### DIFF
--- a/components/centraldashboard/public/ajax_test_helper.js
+++ b/components/centraldashboard/public/ajax_test_helper.js
@@ -28,12 +28,20 @@ export function mockRequest(component, response, respondWithError = false,
 }
 
 /**
- * Simply yeilds to the next tick, so that all Promised requests can do their
+ * Async sleep function which can pause execution flow or yield for next
+ * executions in the event-loop.
+ * @param {number} t Time in ms.
+ * @return {Promise<undefined>}
+ */
+export const sleep = (t) => new Promise((res) => setTimeout(res, t));
+
+
+/**
+ * Simply yields to the next tick, so that all Promised requests can do their
  * thing.
  */
-export async function yeildForRequests() {
-    const sleep = (t) => new Promise((res) => setTimeout(res, t));
-    await sleep(1);
+export async function yieldForRequests() {
+    await sleep(1); // Waits till next tick
 }
 
 /**
--- a/components/centraldashboard/public/ajax_test_helper.js
+++ b/components/centraldashboard/public/ajax_test_helper.js
@@ -51,18 +51,21 @@ export function mockIronAjax(component, response, respondWithError = false) {
     // eslint-disable-next-line no-console
     const finalEvent = `${respondWithError?'error':'response'}`;
     component.generateRequest = async () => {
+        const eventPayload = createRespPayload(
+            response,
+            respondWithError,
+        );
         if (respondWithError) {
-            component._setLastError({error: response});
+            component._setLastError(eventPayload.detail);
         } else {
             component._setLastResponse(response);
         }
-        const eventPayload = createRespPayload(
-            response,
-            respondWithError
-        );
         component.dispatchEvent(
             new CustomEvent(finalEvent, eventPayload)
         );
+        // eslint-disable-next-line no-console
+        console.log(`[mockIronAjax] Dispatched ${finalEvent}`, eventPayload);
+        return component.lastResponse || component.lastError;
     };
 }
 
--- a/components/centraldashboard/public/ajax_test_helper.js
+++ b/components/centraldashboard/public/ajax_test_helper.js
@@ -44,7 +44,7 @@ export async function yeildForRequests() {
  * @param {boolean=} respondWithError
  */
 export function mockIronAjax(component, response, respondWithError = false) {
-    const {tagName} = component;
+    const {tagName, url} = component;
     expect(tagName).toBe(
         'IRON-AJAX', '[mockRequest] Element passed in was not iron-ajax'
     );
@@ -63,8 +63,8 @@ export function mockIronAjax(component, response, respondWithError = false) {
         component.dispatchEvent(
             new CustomEvent(finalEvent, eventPayload)
         );
-        // eslint-disable-next-line no-console
-        console.log(`[mockIronAjax] Dispatched ${finalEvent}`, eventPayload);
+        // eslint-disable-next-line no-console,max-len
+        console.log(`[mockIronAjax] ${url?url+' ':''}Dispatched ${finalEvent}`, eventPayload);
         return component.lastResponse || component.lastError;
     };
 }
--- a/components/centraldashboard/public/ajax_test_helper.js
+++ b/components/centraldashboard/public/ajax_test_helper.js
@@ -33,8 +33,9 @@ export function mockRequest(component, response, respondWithError = false,
  * @param {number} t Time in ms.
  * @return {Promise<undefined>}
  */
-export const sleep = (t) => new Promise((res) => setTimeout(res, t));
-
+export function sleep(t) {
+    return new Promise((res) => setTimeout(res, t));
+}
 
 /**
  * Simply yields to the next tick, so that all Promised requests can do their
@@ -52,7 +53,7 @@ export async function yieldForRequests() {
  * @param {boolean=} respondWithError
  */
 export function mockIronAjax(component, response, respondWithError = false) {
-    const {tagName, url} = component;
+    const {tagName} = component;
     expect(tagName).toBe(
         'IRON-AJAX', '[mockRequest] Element passed in was not iron-ajax'
     );
@@ -71,8 +72,6 @@ export function mockIronAjax(component, response, respondWithError = false) {
         component.dispatchEvent(
             new CustomEvent(finalEvent, eventPayload)
         );
-        // eslint-disable-next-line no-console,max-len
-        console.log(`[mockIronAjax] ${url?url+' ':''}Dispatched ${finalEvent}`, eventPayload);
         return component.lastResponse || component.lastError;
     };
 }
--- a/components/centraldashboard/public/ajax_test_helper.js
+++ b/components/centraldashboard/public/ajax_test_helper.js
@@ -7,7 +7,7 @@ import 'jasmine-ajax';
  * Intercepts the iron-ajax-request event to be able to respond with a mock
  * response. Returns a Promise that resolves whenever the mockRequest is
  * handled.
- * @param {HTMLElement} component
+ * @param {HTMLElement} component Any component that can access the event
  * @param {object} response
  * @param {boolean=} respondWithError
  * @param {string=} matchUrl
@@ -25,4 +25,54 @@ export function mockRequest(component, response, respondWithError = false,
         });
         component.addEventListener(responseEvent, () => resolve());
     });
+}
+
+/**
+ * Simply yeilds to the next tick, so that all Promised requests can do their
+ * thing.
+ */
+export async function yeildForRequests() {
+    const sleep = (t) => new Promise((res) => setTimeout(res, t));
+    await sleep(1);
+}
+
+/**
+ * Helper method that will mock a request directly on the iron-ajax element
+ * by overriding its `generateRequest()` method
+ * @param {HTMLElement} component Iron Ajax elements only
+ * @param {object} response
+ * @param {boolean=} respondWithError
+ */
+export function mockIronAjax(component, response, respondWithError = false) {
+    const {tagName} = component;
+    expect(tagName).toBe(
+        'IRON-AJAX', '[mockRequest] Element passed in was not iron-ajax'
+    );
+    // eslint-disable-next-line no-console
+    const finalEvent = `${respondWithError?'error':'response'}`;
+    component.generateRequest = async () => {
+        if (respondWithError) {
+            component._setLastError({error: response});
+        } else {
+            component._setLastResponse(response);
+        }
+        const eventPayload = createRespPayload(
+            response,
+            respondWithError
+        );
+        component.dispatchEvent(
+            new CustomEvent(finalEvent, eventPayload)
+        );
+    };
+}
+
+function createRespPayload(response, isError) {
+    const detail = {
+        response,
+    };
+    isError && Object.assign(detail, {
+        error: response,
+        request: {response: {error: response}},
+    });
+    return {detail};
 }
--- a/components/centraldashboard/public/components/manage-users-view_test.js
+++ b/components/centraldashboard/public/components/manage-users-view_test.js
@@ -15,10 +15,9 @@ const TEMPLATE = `
   </template>
 </test-fixture>
 `;
-const TEST_USER = 'test@kubeflow.org';
-const newNS = (namespace, role, user='test@kubeflow.org') => ({role, namespace, user});
-const oNs = newNS('ns1', 'owner');
-const generalNs = [oNs, newNS('ns2', 'contributor'), newNS('ns3', 'contributor')];
+const user = 'test@kubeflow.org';
+const oNs = {namespace: 'ns1', role: 'owner'};
+const generalNs = [oNs, {namespace: 'ns2', role: 'contributor', user}, {namespace: 'ns3', role: 'contributor', user}];
 
 describe('Manage Users View', () => {
     let manageUsersView;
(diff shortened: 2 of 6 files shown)


components/centraldashboard/public/ajax_test_helper.js
```diff
@@ -26,3 +26,53 @@ export function mockRequest(component, response, respondWithError = false,
         component.addEventListener(responseEvent, () => resolve());
     });
 }
+
+/**
+ * Simply yeilds to the next tick, so that all Promised requests can do their
+ * thing.
+ */
+export async function yeildForRequests() {
+    const sleep = (t) => new Promise((res) => setTimeout(res, t));
```
prodonjs: Why not just do await new Promise(..) and avoid defining the inner function?

components/centraldashboard/public/ajax_test_helper.js
```diff
@@ -26,3 +26,53 @@ export function mockRequest(component, response, respondWithError = false,
         component.addEventListener(responseEvent, () => resolve());
     });
 }
+
+/**
+ * Simply yeilds to the next tick, so that all Promised requests can do their
+ * thing.
+ */
+export async function yeildForRequests() {
```
prodonjs: s/yeild/yield/

components/centraldashboard/public/components/manage-users-view_test.js
```diff
@@ -0,0 +1,213 @@
...
+const MU_VIEW_SELECTOR_ID = 'test-manage-users-view';
+const TEMPLATE = `
+<test-fixture id="${FIXTURE_ID}">
+  <template>
+    <manage-users-view id="${MU_VIEW_SELECTOR_ID}"></manage-users-view>
+  </template>
+</test-fixture>
+`;
+const TEST_USER = 'test@kubeflow.org';
+const newNS = (namespace, role, user='test@kubeflow.org') => ({role, namespace, user});
```
prodonjs: This function feels a bit unnecessary since it's just shaping the arguments into a different object. For test-comprehension purposes I would just explicitly define the objects in the constants below.

components/centraldashboard/public/components/manage-users-view_test.js
```diff
@@ -0,0 +1,213 @@
...
+        manageUsersView.ownedNamespace = oNs;
+        manageUsersView.namespaces = generalNs;
+
+        flush();
+        await yeildForRequests();
+
+        expect(manageUsersView.shadowRoot.querySelector('.Acct-Info > .content').innerText)
+            .toBe('test@kubeflow.org');
+        expect(manageUsersView.shadowRoot.querySelector('.Namespaces')
+            .hasAttribute('hidden')).toBe(false, 'Namespaces was still hidden');
```
prodonjs: It looks like you're using the second argument as a failure message. I couldn't find any place in the [Docs](https://jasmine.github.io/api/3.4/matchers) where the second argument is described, but I do see that it will show the message as part of the output if the assertion fails.

components/centraldashboard/public/components/registration-page_test.js
```diff
@@ -0,0 +1,94 @@
...
+const MU_VIEW_SELECTOR_ID = 'test-registration-page';
+const TEMPLATE = `
+<test-fixture id="${FIXTURE_ID}">
+  <template>
+    <registration-page id="${MU_VIEW_SELECTOR_ID}"></registration-page>
+  </template>
+</test-fixture>
+`;
+const TEST_USER = 'test@kubeflow.org';
+const sleep = (t) => new Promise((res) => setTimeout(res, t));
```
prodonjs: Do you want to use the helper functions you defined in the ajax_test_helper module?

components/centraldashboard/public/components/registration-page_test.js
```diff
@@ -0,0 +1,94 @@
...
+
+    it('Should work in a Normal Flow', async () => {
+        mockIronAjax(
+            registrationPage.$.MakeNamespace,
+            {message: 'Success!'},
+        );
+
+        const $e = (selector) => registrationPage.shadowRoot.querySelector(selector);
+        flush();
+        await yeildForAsync(); // So the view can render
```
prodonjs: I'm curious why these have been necessary. I found that the flush() function was all I needed to do to ensure that any rendering operations would finish before validating the DOM state. I just want to make sure there's a definitive reason why we're doing it so that tests don't become flaky at some later point.

components/centraldashboard/public/components/utilities-mixin_test.js
```diff
@@ -1,37 +1,117 @@
...
 
+const PASS_FUNCTION = (_) => _;
+
 describe('Utilities Mixin', () => {
     let testObject;
 
     beforeEach(() => {
         testObject = new TestClass();
+        // Since base class is not HTML
+        testObject.dispatchEvent = PASS_FUNCTION;
```
prodonjs: I would suggest using a mock here that way you can verify interactions easily with it.

components/centraldashboard/public/components/utilities-mixin_test.js
```diff
@@ -1,37 +1,117 @@
...
             .toBe('/foo?other_param=value');
-        expect(testObject.buildHref('/foo?other_param=value', {ns: 'bar'}))
+        expect(buildHref('/foo?other_param=value', {ns: 'bar'}))
             .toBe('/foo?other_param=value&ns=bar');
     });
+
+    it('FireEvent correctly fires Event-Obj or custom events', () => {
+        const fireEvent = testObject.fireEvent.bind(testObject);
+        let lastEvent;
+        spyOn(testObject, 'dispatchEvent').and.callFake((ev) => {
```
prodonjs: There's no need to provide the fake implementation of `dispatchEvent`. The `fireEvent` method's logic builds the argument for `dispatchEvent` based on its parameters so all you really need to do is call `fireEvent` and then validate that the mock was called with the arguments you expect.

components/centraldashboard/public/components/utilities-mixin_test.js
```diff
@@ -1,37 +1,117 @@
...
+                </span>
+            </paper-toast>
+        `;
+        const toast = dom.content.children[0];
+        const [s1, s2, s3] = [1, 2, 3]
+            .map((i) => toast.querySelector(`#s${i}`));
+        [toast, s1, s2, s3].forEach((el) => {
+            el.onclick = closeToast;
+        });
+        toast.close = () => {
```
prodonjs: I'm not intimately familiar with the `paper-toast` object but it feels a bit unusual that you are overriding and adding new methods to its implementation. Your tests are then validating your own overridden logic and not the interface you are testing.

Is there a reason you couldn't rely on the `opened` property?

https://www.webcomponents.org/element/@polymer/paper-toast/elements/paper-toast

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: It, unfortunately, does not allow you to mock a field that never was defined. Since our test class is not really an HTML element, `dispatchEvent` is not a defined value on it. I could define this method within the class, and mock it here, however?

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: That's because this test uses a non-dom base class w/o a test-fixture. So we actually never modify or touch the dom with any of these tests. Thus I never even load paper-toast (though this may work in this test). I just need to make sure that the test does call the `close` method on the correct parent in the `document-fragment` (or fake dom)

components/centraldashboard/public/ajax_test_helper.js
avdaredevil: Changed it, since `sleep` is used in a few places.

components/centraldashboard/public/components/manage-users-view_test.js
avdaredevil: I realized I needed to add these as I was stepping through the tests. Seems like the update from the Ajax response to the view is not synchronous and the `flush()` call does not yield a tick for the async backlog to finish.

components/centraldashboard/public/components/manage-users-view_test.js
avdaredevil: Yeah I found these really useful for test debuggability and resolving issues with new features.

components/centraldashboard/public/components/registration-page_test.js
avdaredevil: I realized I needed to add these as I was stepping through the tests. Seems like the update from the Ajax response to the view is not synchronous and the `flush()` call does not yield a tick for the async backlog to finish.

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: Dispatch event gets called with an event object, how would it compare to event objects?
- The event I construct to compare with
- The event generated by `fireEvent` that is fired?

components/centraldashboard/public/components/registration-page_test.js
avdaredevil: Yup, removed these lines

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: To add to this, the goal is **not to test `paper-toast`** but rather to verify that a `.close()` call was made on said element.

components/centraldashboard/public/ajax_test_helper.js
```diff
@@ -26,3 +26,64 @@ export function mockRequest(component, response, respondWithError = false,
...
     });
 }
+
+/**
+ * Async sleep function which can pause execution flow or yield for next
+ * executions in the event-loop.
+ * @param {number} t Time in ms.
+ * @return {Promise<undefined>}
+ */
+export const sleep = (t) => new Promise((res) => setTimeout(res, t));
```
prodonjs: For consistency's sake, use the `function` syntax rather than exporting a const. That's the preferred Google style.

components/centraldashboard/public/ajax_test_helper.js
```diff
@@ -26,3 +26,64 @@ export function mockRequest(component, response, respondWithError = false,
...
+        if (respondWithError) {
+            component._setLastError(eventPayload.detail);
+        } else {
+            component._setLastResponse(response);
+        }
+        component.dispatchEvent(
+            new CustomEvent(finalEvent, eventPayload)
+        );
+        // eslint-disable-next-line no-console,max-len
+        console.log(`[mockIronAjax] ${url?url+' ':''}Dispatched ${finalEvent}`, eventPayload);
```
prodonjs: Remove before submitting.

components/centraldashboard/public/components/manage-users-view_test.js
```diff
@@ -0,0 +1,212 @@
...
+            manageUsersView.$.GetContribsAjax,
+            contribList,
+        );
+
+        manageUsersView.user = user;
+        manageUsersView.ownedNamespace = oNs;
+        manageUsersView.namespaces = generalNs;
+
+        flush();
+        await yeildForRequests();
```
prodonjs: I think you need to update all occurrences to yieldForRequests.

components/centraldashboard/public/components/registration-page_test.js
prodonjs: I see. I think the difference here is that the `mockRequest` method I was previously using for tests returned a Promise that could be resolved in tests. Now that I take a closer look, is there a way that the two functions `mockRequest` and `mockIronAjax` could be merged into a single function. I notice they have the same signature but different behaviors and the latter doesn't return anything.

components/centraldashboard/public/components/utilities-mixin_test.js
```diff
@@ -1,37 +1,117 @@
+/* eslint-disable max-len */
```
prodonjs: Any particular reason for this? I'd suggest removing it and trimming the lengthy lines to be under the 80 character limit.

components/centraldashboard/public/components/utilities-mixin_test.js
prodonjs: That's not a problem. Just use [jasmine.createSpy()](https://jasmine.github.io/api/3.4/jasmine.html#.createSpy). Then you have a bare spy object that you can later. See my comment in the method where `dispatchEvent` is used below.

components/centraldashboard/public/components/utilities-mixin_test.js
prodonjs: Here's roughly the implementation I would use. The purpose of this test is to verify the interaction with `dispatchEvent`, which is not part of your implementation but is provided by the CustomElement class utilizing the mixin. So you create the spy function for `dispatchEvent` and verify that it gets called with the type of object you expect. The end result is similar to what you started with, but you don't need to have any kind of fake implementation since you're only verifying how dispatchEvent was called. Normally, you could just use the `toHaveBeenCalledWith` matcher, but it doesn't work well when comparing Event objects.

```
let lastEvent;
testObject.dispatchEvent = jasmine.createSpy();
testObject.fireEvent('foo');

lastEvent = testObject.dispatchEvent.calls.mostRecent().args[0];
expect(lastEvent.type).toBe('foo');
expect(lastEvent.detail).toEqual(null);

testObject.fireEvent('bar', {ap: 1});
lastEvent = testObject.dispatchEvent.calls.mostRecent().args[0];
expect(lastEvent.type).toBe('bar');
expect(lastEvent.detail).toEqual({ap: 1});

const ev = new Event('click');
testObject.fireEvent(ev);
lastEvent = testObject.dispatchEvent.calls.mostRecent().args[0];
expect(lastEvent).toBe(ev);
```

components/centraldashboard/public/components/utilities-mixin_test.js
prodonjs: OK this makes sense now. I had just assumed this was using the text fixture structure..

You can then simplify it to:
```
toast.close = jasmine.createSpy();
toast.click();
s1.click();
s2.click();
s3.click();
expect(toast.close).toHaveBeenCalledTimes(3);
```

prodonjs:
@avdaredevil - I made some additional comments and suggestions. I didn't realize the nuance of testing the mixin class without having the HTMLElement methods. It's similar to testing an abstract class by creating a simplified concrete class that would supply the necessary implementation details, but in this case it's fairly easy to provide spies using `jasmine.createSpy` to verify interactions.

FYI, in its current state, this PR fails because of the `yeild` to `yield` change.

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: I initially planned on doing it that way, but this test is slightly ambiguous since it does not ascertain which clicks actually caused the `.close()` calls. Do you think that's fair to ignore in favor of brevity?

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: Removed, I think I copied this over from other tests.

components/centraldashboard/public/components/manage-users-view_test.js
avdaredevil: Done, surprised the IDE didn't catch that.

components/centraldashboard/public/ajax_test_helper.js
avdaredevil: I thought this would make sense to keep in for debugging. Though I can see how it would get annoying with a bigger test. Removing...

components/centraldashboard/public/ajax_test_helper.js
avdaredevil: Done!

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: Yup done!

components/centraldashboard/public/components/utilities-mixin_test.js
avdaredevil: Done!

prodonjs:
/lgtm
/approve
/hold

Fix the formatting in the utililies-mixin.js and this should be good.

avdaredevil:
@prodonjs Fixed those changes, PTAL. Thanks!

PR #4130, opened 2019-09-17
pick up manifest config fix on v0.6 branch

kunmingg:
Part 2 of https://github.com/kubeflow/kubeflow/pull/4122
So config version is consistent on GCP

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4130)


jlewi:
@kunmingg So #4122 doesn't affect the IAP configs but nontheless we want the GCP configs for IAP and basic auth to use the same commit of kubeflow/manifests?

/hold
/lgtm
/approve

kunmingg:
@jlewi 
Try to give consistent version here: https://github.com/kubeflow/website/pull/975/files.
So would be better to use consistent version for all options on GCP.

PR #4141, opened 2019-09-17
move pod default back to gcp

gabrielwen:
With `simpleReconcile` we don't need the hack to generate PodDefault anymore.

Relates to #4056 

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4141)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -837,7 +837,7 @@ func (gcp *Gcp) Apply(resources kftypesv3.ResourceEnum) error {
 			}
 		}
 	}
-	if gcp.KfDef.Email == "" {
+	if gcp.kfDef.Spec.Email == "" {
 		return nil
 	}
 	if *p.EnableWorkloadIdentity {
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -837,8 +837,14 @@ func (gcp *Gcp) Apply(resources kftypesv3.ResourceEnum) error {
 			}
 		}
 	}
-
-	return nil
+	if gcp.KfDef.Email == "" {
+		return nil
+	}
+	if *p.EnableWorkloadIdentity {
+		return gcp.SetupDefaultNamespaceWorkloadIdentity()
+	} else {
+		return gcp.ConfigPodDefault()
+	}
 }
 
 // Try to get information for the deployment. If returned, delete it.
--- a/bootstrap/pkg/kfapp/coordinator/coordinator.go
+++ b/bootstrap/pkg/kfapp/coordinator/coordinator.go
@@ -791,45 +791,16 @@ func (kfapp *coordinator) Apply(resources kftypesv3.ResourceEnum) error {
 		return nil
 	}
 
-	gcpAddedConfig := func() error {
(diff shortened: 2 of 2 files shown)


bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -837,8 +837,14 @@ func (gcp *Gcp) Apply(resources kftypesv3.ResourceEnum) error {
 			}
 		}
 	}
-
-	return nil
+	if gcp.kfDef.Spec.Email == "" {
+		return nil
+	}
+	if *p.EnableWorkloadIdentity {
+		return gcp.SetupDefaultNamespaceWorkloadIdentity()
```
jlewi: Do you need to modify these functions so that they check if the namespace kubeflow exists and only run when it does?
https://github.com/kubeflow/kubeflow/blob/ee8fe0e4280e4d729cc488442bdba797bf9ad995/bootstrap/pkg/kfapp/gcp/gcp.go#L1733

gabrielwen: so we check the namespace creation here:
https://github.com/kubeflow/kubeflow/blob/64f4f79ef7d2f1eed1113dce6955437717b94e1b/bootstrap/pkg/kfapp/kustomize/kustomize.go#L324

when I implemented `ConfigPodDefault` I was thinking that we already check the namespace in kustomize.go, so I skipped the check in gcp.go.  plus if the namespace doesn't exist the function will fail and reconcile will run it again later.  do you think we still need to check it?

jlewi: Sorry I'm a little confused how this works.

You are calling this at the end of Gcp.Apply. But the first time you run Gcp.Apply the namespace kubeflow won't exist. So any attempt to create items inside the kubeflow namespace will fail.

So I would expect there to be a check somewhere to see whether the namespace exists which would be an indication that kustomize.go had successfully run.

It looks like the tests are passing. Is that just because your reconcile loop is based on errors?

gabrielwen: yes, when there has an error, I log a warning and retry:
https://github.com/kubeflow/kubeflow/blob/a4286d524a18e86bda97b5fcf05fa4086063513a/bootstrap/pkg/kfapp/coordinator/coordinator.go#L237

currently the straw man version reconcile simply keeps retrying until retry loop reaches max retrying time.  so if kustomize.Apply fails the depending op gcp.Apply will fail.  we will keep retry both until they both succeeds.

I think what I need to know is whether we need to check namespace creation.  in this PR I only check whether email is set, which follows the logic in kustomize.Apply:
https://github.com/kubeflow/kubeflow/blob/a4286d524a18e86bda97b5fcf05fa4086063513a/bootstrap/pkg/kfapp/kustomize/kustomize.go#L287

if we check namespace in gcp.Apply, what do we do if the namespace doesn't exist?  skip pod default and terminate or return error and retry?

jlewi: I think based on our discussion today in the office. I'm waiting on you to take another pass at this and using conditions to determine whether or not to reapply.

gabrielwen: ok, let me create another PR to add condition/status to v1beta1 KfDef and loop back to this PR.

Issue #4002, opened 2019-08-26
kfctl supports v1beta1 of KfDef (and v1alpha1)


jlewi:
/kind feature

In 0.7 we will introduce a v1beta1 spec for KFDef.
See #3478

This issue tracks adding support for that new spec to kfctl.

With 0.7 we should aim to provide backwards compatibility with v1alpha1 specs. To support that; we should define a conversion function which converts v1alpha1 specs to v1beta1 specs.

jlewi:
@gabrielwen Could you provide an update and ETA please?

richardsliu:
@gabrielwen Can this be closed?

PR #4149, opened 2019-09-18
Kfdef converter

gabrielwen:
Relates to #4002 
Adds a new function to read from a file path and returns KfDef in v1beta1 format.  If the file path points to v1alpha1 KfDef, will converts to v1beta1.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4149)


gabrielwen:
@jlewi @yanniszark @swiftdiaries @kkasravi this PR is ready.

bootstrap/pkg/apis/apps/group.go
```diff
@@ -155,6 +155,14 @@ const (
 	EXISTING_ARRIKTO = "existing_arrikto"
 )
 
+// Plugin kind used starting from v1beta1
+const (
+	AWS_PLUGIN_KIND              = "KfAwsPlugin"
+	GCP_PLUGIN_KIND              = "KfGcpPlugin"
+	MINIKUBE_PLUGIN_KIND         = "KfMinikubePlugin"
+	EXISTING_ARRIKTO_PLUGIN_KIND = "KfExistingArriktoPlugin"
+)
```
swiftdiaries: Hey Gabriel
Thanks for the this!
What about KfDefs for the kfctl_k8s? How would they be converted?

bootstrap/pkg/apis/apps/kfdef/kfloader.go
```diff
@@ -0,0 +1,292 @@
...
+	if err := yaml.Unmarshal(configs, kfdef); err != nil {
+		return nil, &kfapis.KfError{
+			Code:    int(kfapis.INVALID_ARGUMENT),
+			Message: fmt.Sprintf("invalid config file format: %v", err),
+		}
+	}
+	return kfdef, nil
+}
+
+// Loads KfDef fild and returns KfDef in v1beta1. v1alpha1 will be converted to the new version.
```
swiftdiaries: nit: fild -> file

bootstrap/pkg/apis/apps/kfdef/kfloader.go
```diff
@@ -0,0 +1,292 @@
...
+func LoadKfDefFromURI(configFile string) (*kfdefv1beta1.KfDef, error) {
+	if configFile == "" {
```
swiftdiaries: So any KfDef config would be converted to the v1beta1 and then used?

gabrielwen: yes, after this call all KfDef should be in v1beta1.

bootstrap/pkg/apis/apps/group.go
gabrielwen: this is only used for plugins.  from this file it doesn't have any plugins?
https://github.com/kubeflow/manifests/blob/master/kfdef/kfctl_k8s_istio.yaml

bootstrap/pkg/apis/apps/kfdef/kfloader.go
gabrielwen: fixed.

jlewi:
/lgtm
/approve
/hold

Added a hold in case anyone else needs to LGTM this; on my end its ready to go.

gabrielwen:
/hold cancel

sorry I need this for reconcile semantics.  let me know if anyone thinks this PR needs to be updated.  I can make follow up changes.

PR #4151, opened 2019-09-18
anthos config WIP

lluunn:
It's a copy of kfctl_k8s_istio, but remove istio-crd and istio-install.
Keep istio though, they are the resources needed by KF.

/cc @richardsliu @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4151)


lluunn:
Seems the config should go to https://github.com/kubeflow/manifests/tree/master/kfdef

PR #4148, opened 2019-09-18
Use python to define the E2E test workflow for kfctl.

jlewi:
* We need to write an E2E test for kfctl upgrade (kubeflow/kfctl#35).

  Before we do that we want to remove ksonnet from our existing E2E test.

  The E2E test for upgrades will be very similar to the kfctl E2E test
  so it makes sense to convert that test to python before writing
  the upgrade test.

* Related to kubeflow/kubeflow#3035 migrate e2e tests off ksonnet

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4148)

PR #4154, opened 2019-09-19
e2e: existing_arrikto: initial e2e test

yanniszark:
Initial E2E test for existing_arrikto.
This should ensure that nothing breaks with new PRs.
In the near future, we should move this to a pyfunc.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4154)


nrchakradhar:
@yanniszark 
For my understanding can you please let me know what is the reason behind having cluster_creation as a script inside kfctl_go_test.py instead of a separate step in the "dag" which is added only if cluster_creation param is True? This would make it similar to the delete_cluster_step addition.

There is hard coding for cluster name, zone, in 
https://github.com/kubeflow/kubeflow/blob/0c244a34f416305cb3ca8b0648a08c0000b76116/testing/kfctl/delete_existing_cluster.py#L20-L22
that should match with 
https://github.com/kubeflow/kubeflow/blob/0c244a34f416305cb3ca8b0648a08c0000b76116/testing/kfctl/create_existing_cluster.sh#L16-L18

By making creation also a step, it would be possible to remove the hard coding also.
If its cumbersome to change and better modular approach will follow with py_func then its okay.

yanniszark:
@nrchakradhar currently, this is a dirty approach to get it working.
Once I have something working, I can begin to clean it up.

For the separate DAG step, I was actually thinking of including deletion in the kfctl_delete step instead of making a new step.
kfctl would create and delete a cluster in those steps for other configs, so it's not a stretch imo.

yanniszark:
@jlewi this is ready for review.
I used the ksonnet way because it seems that things are still pending on the python front.
Once the dust has settled around py_func, I will port the existing_arrikto test to it.

Initial E2E test for existing_arrikto:
* Initial test runs the same tests as the GCP configs.
* Instead of an on-prem cluster, a GKE cluster is used for starters.

Some changes I had to make:
* `kfctl_go_test`: for GCP, the cluster is created by the kfctl binary. For existing_arrikto, we need to supply a cluster, so a new argument `cluster_creation_script` is introduced.
* `kfctl_is_ready`: the list of deployments to check includes gcp-specific applications and is parameterized based in the python code based on the `basic_auth`, `use_istio` parameters. A new argument `checklist` is introduced, which explicitly states the list of Deployments/StatefulSets to check.
* `kfctl_delete_test`: as with `kfctl_go_test`, we need to delete the cluster we created earlier, as it won't be cleaned up by kfctl. To achieve that, we add a new `cluster_deletion_script` argument.
* Namesuffix is determined implicitly by the basic_auth, istio params. We add a `nameSuffix` param to explicitly set it.

testing/kfctl/kf_is_ready_test.py
```diff
@@ -29,6 +30,20 @@ def test_kf_is_ready(namespace, use_basic_auth, use_istio):
 
   util.load_kube_config()
 
+  # If a checklist is provided, use that instead.
+  if checklist:
+    with open(checklist) as f:
+      items = yaml.safe_load(f)
+    for item in items:
+      if item["kind"] == "Deployment":
+        util.wait_for_deployment(api_client, item["namespace"], item["name"], 10)
```
kkasravi: too bad we can't check status conditions for these

yanniszark: Is this more desirable  over checking that replicas == readyReplicas that the code does right now?
If yes, we could improve the util library in kubeflow/testing.

kkasravi: i've seen crashloopbackoff in a number of scenarios for the pods

no intent to hold up your PR, opening an issue is perhaps a better idea,
where applications and events could be checked

Feel free to open an issue otherwise I can

thanks yannis

testing/kfctl/scripts/create_existing_cluster.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
nrchakradhar: Can the functionality of this script be achieved by 

https://github.com/kubeflow/testing/blob/a841aa0b2b1f6b64e06f5ab2be9f6df2ba3b214e/py/kubeflow/testing/util.py#L166-L170

testing/kfctl/scripts/delete_existing_cluster.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
nrchakradhar: Can this be used here.
https://github.com/kubeflow/testing/blob/a841aa0b2b1f6b64e06f5ab2be9f6df2ba3b214e/py/kubeflow/testing/util.py#L196-L207

testing/kfctl/kfctl_go_test.py
```diff
@@ -211,10 +214,26 @@ def test_build_kfctl_go(app_path, project, use_basic_auth, use_istio, config_pat
...
+  #
   # Do not run with retries since it masks errors
   util.run([kfctl_path, "apply", "-V", "all"], cwd=app_path)
 
-  verify_kubeconfig(app_path)
+  if not cluster_creation_script:
+    verify_kubeconfig(app_path)
+
+
+def filterSpartakus(spec):
```
nrchakradhar: Probably not needed. This is renamed as filter_sparkatus
https://github.com/kubeflow/kubeflow/blob/77e4ee4a050dcd5d60b79bd28bb7923a065b41b4/testing/kfctl/kfctl_go_test.py#L93

Probably  merge issue.

testing/kfctl/scripts/create_existing_cluster.sh
yanniszark: It can, but it basically requires me to translate the gcloud command to a JSON request body, which is not straightforward.
We can leave this to a followup PR is it's more desirable to do it that way.
I also didn't find any other code using that function.

testing/kfctl/scripts/delete_existing_cluster.sh
yanniszark: This is more straightforward to achieve.
I will try to use it here.

testing/kfctl/kfctl_go_test.py
yanniszark: Exactly, this was a merge issue, I will rebase to fix.

yanniszark:
@abhi-g @jlewi @lluunn
Could you please take a look if everything looks ok so we can get this in?
It will be a great help in merging other PRs and ensuring nothing breaks.

testing/kfctl/kf_is_ready_checklists/kfctl_existing_arrikto_checklist.yaml
```diff
@@ -0,0 +1,153 @@
+# Deployments in namespace kubeflow
```
jlewi: Instead of making this Arrikto specific, should we have a file that is components common to all deployments and then a separate file for anything that is config specific?

nrchakradhar: I was thinking of making more changes here, to generate this list from KfDef Spec used in testing. The approach would be: look into the applications used in KfDef spec, search inside of the generated manifests folder for matching deployments/statefulset etc. I can take a look at this once #4148, this PR and #4187 is merged.
Does that sound reasonable.

yanniszark:
> This will not be unique. A given pull request can trigger multiple invocations of the same test. If you have two runs running simultaneously they would

I agree, I should have explained my reasoning here.
I did it this way so the number of clusters wouldn't grow in an unbounded manner, in case there was a resource leak.
It seemed reasonable to only have one cluster per PR.
If it's not, I will change it to use the build id.

> Why not make this a pytest file and then you can just invoke it directly in your workflow by adding it as a step to the graph.

It required less logic in the jsonnet file.
Also, it didn't seem unreasonable for the time being.

> What if we used this PR to define the appropriate pytest modules for the Arrikto test; get it merged and then as soon as #4148 is merged use that as the basis to generate an appropriate workflow which adds the cluster creation and deletion steps as needed?

I agree about using py_func to define tests for existing_arrikto.
However, the v0.7 is very soon and I'd like to have E2E tests to validate nothing breaks.
Do you think that the python-based testing will make it in time?
In the community meeting, @abhi-g proposed merging the jsonnet tests and convert after v0.7 is out.

testing/kfctl/kf_is_ready_checklists/kfctl_existing_arrikto_checklist.yaml
yanniszark: @nrchakradhar  would be great if we could automate this.
@jlewi the current python test is GCP-specific, that's why I had to find a way to specify what I wanted.
Ideally, the E2E test will discover what is needed.
Otherwise, we could define it in checklist-like files, which can be used by any config.

jlewi:
@yanniszark Is there a way to make this test non-blocking until we've observed that the test is running reliably?

yanniszark:
> @yanniszark Is there a way to make this test non-blocking until we've observed that the test is running reliably?

Not that I know of.
I believe the way to achieve this is to declare a new field in the prow_config.yaml workflow definition (eg optional) and then change the code in run_e2e_workflow (in kubeflow/testing) to act accordingly.
We can open an issue to track this feature request.

> PULL_NUMBER also won't be set in tpostsubmits.

Thanks, I used REPO_NAME and BUILD_ID, which are available in both pre and post submits.
This combination should be unique.

> The naming pattern also doesn't follow the regex used by cleanup_ci.py

Changed the cluster name to start with `kfctl`.

Finally, I opened #4189 to track the necessary improvements to this PR.

testing/kfctl/kf_is_ready_checklists/kfctl_existing_arrikto_checklist.yaml
jlewi: @yanniszark see TODO here
https://github.com/kubeflow/kubeflow/blob/e38c6f3234b301f33226706c285aafb9bd32d357/testing/kfctl/kf_is_ready_test.py#L33

There are a set of applications that should be installed across all deployments (e.g. TFJob, PyTorch,, Pipelines etc....).

Think of this as a compliance tests. These are the applications that should be installed on all KF compliant clusters.

So that list whether specified in a YAML file or in python should be the same for all applications.

Then there should be a list of applications that should only be installed in certain settings; e.g. on GCP with IAP.

So the tests that apply to all components should be written once and reused across all components.

So we should create a list of all the K8s resources common to all deployments.
Then based on either the KFDef spec or command line arguments we should add additional expectations.

So to make the test non GCP specific we should conditionally enable the set of K8s resources to look for GCP resources based on whether the GCP plugin spec is provided in the KFDef file.

jlewi: Thanks @yanniszark for updating kfctl_go_test.py.
Can we remove this file now?

testing/kfctl/kf_is_ready_test.py
```diff
@@ -33,36 +35,79 @@ def test_kf_is_ready(namespace, use_basic_auth, use_istio):
...
+    "application-controller-stateful-set",
+    "seldon-operator-controller-manager",
   ]
 
-  ingress_related_deployments = []
+  with open(path.join(app_path, "app.yaml")) as f:
+    kfdef = yaml.safe_load(f)
+  platform = kfdef["spec"]["platform"]
+
+  if platform == "gcp":
```
jlewi: nit: can we combine this with the if/else below.

jlewi:
Thanks @yanniszark can we delete kfctl_existing_arrikto_checklist.yaml now?

jlewi:
FYI: I just lgtm'd #4187 which refactors kfctl_go_test.py. Hopefully this doesn't impact you.

nrchakradhar:
#4232 is also modifying kfctl_go_test.py to include a new flag. Probably the number of flags should be reduced and could be a separate PR though.

yanniszark:
@jlewi tests are failing because we use the extensions v1beta1 client to check deployments.
https://github.com/kubeflow/testing/blob/e1f4fa27a2b6f5be5c4ef27d8a0718a96c04562d/py/kubeflow/testing/util.py#L412

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -142,7 +142,13 @@ func (coord *coordinator) getPackageManagers(kfdef *kfdefsv3.KfDef) *map[string]
 // kftypesv3.LoadKfApp which will try and dynamically load a .so
 //
 func getPackageManager(kfdef *kfdefsv3.KfDef) (kftypesv3.KfApp, error) {
-	switch kfdef.Spec.PackageManager {
+
+	packageManager := kfdef.Spec.PackageManager
```
jlewi: Why are there kfctl changes in this PR?

yanniszark: @jlewi these are minor fixes which became necessary after the new semantics PR was merged.
I have opened a separate PR to add them, but I also include them here to get the tests working.
https://github.com/kubeflow/kubeflow/pull/4244

jlewi: Got it. Thanks

jlewi:
@yanniszark Can you let me know when the coordinator.go changes have been merged and this PR has been rebased to remove those changes. Then I'll approve this one.

yanniszark:
Got this error from the KFServing test:
```
```

Probably a transient network error ?
/retest

jlewi:
@yanniszark Should we enable  this only on post-submit and not presubmit until we have a track record that its running reliably? Reverting PRs that break postsubmits is reasonable.

Should we split the improvements to kfctl_test_ready into a separate PR? Would this help distinguish problems with kfctl_existing_arrikto vs. the tests for the individual applications?

Is the KFServing test just checking that KFServing is deployed correctly? Or is it trying to use KFServing?

jlewi:
We could also guard the additional tests in is_ready to not run them in presubmits so we can verify they are running reliably before invoking them on presubmit.

yanniszark:
@jlewi the last few runs have tests failing for random reasons (see kfserving above) or timing out waiting for a deployment to be ready (GCP test waiting for the metadata service).

> @yanniszark Should we enable this only on post-submit and not presubmit until we have a track record that its running reliably? Reverting PRs that break postsubmits is reasonable.

Do postsubmits run after the merge has been made or after the approve is given?

> Is the KFServing test just checking that KFServing is deployed correctly? Or is it trying to use KFServing?

Not sure, this is not something I have touched.

> Should we split the improvements to kfctl_test_ready into a separate PR? Would this help distinguish problems with kfctl_existing_arrikto vs. the tests for the individual applications?

How about just disabling the check for the metadata deployment?
It seems to be causing most of the failures I have seen.

jlewi:
@yanniszark postsubmits are run after the PR is merged.   My suggestion is to only enable new tests as blocking presubmits after we've baked them using postsubmits to ensure the new tests aren't flaky. This is purely defensive to ensure the new tests don't end up blocking PRs because of flakiness.

If we are monitoring the postsubmits pretty actively and the tests are reliable then we can rollback any breaking PRs.

My suggestion would be to do this as follows

1. modify prow_config.yaml to only trigger existing_arrikto on postsubmit
1. In kfctl_is_ready_test.py only add the new resources to the lists if job_type is not presubmit

jlewi:
even better for kfctl_is_ready_test.py would be to run the test but leverage pytest to mark it as an expected failure.  Off hand I'm not sure exactly how to do this without adding two steps to the graph.

Its possible with pytest its as simple as adding two test functions to the python file one that runs  the new tests and one that runs the old tests and then adding the pytest annotation to indicate the new one is expected to fail.

yanniszark:
@jlewi I generally agree with being conservative.
However, the kfctl_is_ready test is very lacking.
It's basically not checking half the things that have been deployed.
I have tried to include as much as possible there and just made the metadata deployments optional, which seem to cause issues.

I also made the existing_arrikto test a postsubmit.

jlewi:
Thanks.
Looks like you got some conflicts
/lgtm
/approve

yanniszark:
@jlewi I'm getting failed tests from the kfctl-go-build-deploy test.
It's timing out while waiting for components to be ready.
AFAIK, this is a new test, any idea why it's failing when it seems to use more or less the same config as the other GCP tests?

/retest

testing/workflows/components/kfctl_go_test.jsonnet
```diff
@@ -33,7 +33,13 @@ local runPath = srcDir + "/testing/workflows/run.sh";
 local kfCtlPath = srcDir + "/bootstrap/bin/kfctl";
 local kubeConfig = testDir + "/kfctl_test/.kube/kubeconfig";
 
-// Name for the Kubeflow app.
+// cluster_creation_script specifies the script to run in order to create
```
kkasravi: Will we move to the argo_client/tekton_client rather than jsonnet  in a future PR ?

yanniszark: Yes, the issue is https://github.com/kubeflow/kubeflow/issues/4189

yanniszark:
@jlewi could you lgtm again now that I have rebased?

testing/kfctl/kf_is_ready_test.py
```diff
@@ -33,36 +35,88 @@ def test_kf_is_ready(namespace, use_basic_auth, use_istio):
...
       "ml-pipeline-scheduledworkflow",
       "ml-pipeline-ui",
+      "ml-pipeline-viewer-controller-deployment",
+      "mysql",
       "notebook-controller-deployment",
-      "tf-job-operator",
+      "profiles-deployment",
       "pytorch-operator",
-      "katib-controller",
+      "tensorboard",
```
jlewi: Why are we testing tensorboard? That shouldn't even be deployed.

jlewi:
@yanniszark The changes to kfctl_is_ready_test.py seem a bit risky right now. If the test is flaky adding more tests to it is only likely to exacerbate the problem. Can we pull the changes to kfctl_is_ready_test.py into a separate PR?  

I think before we increase coverage of is_read_test we need to split it up; if we test more things in a single test the test is just going to become more unreliable.

jlewi:
@yanniszark I think an easy thing to do would be break it up into multiple tests corresponding to different related groups of servicess

e.g.

def test_kfserving
   _wait_for_deployments(....)
   _wait_for_statefulsets(...)

def test_gcp
    _wait_for_deployments(....)
   _wait_for_statefulsets(...)

That should hopefully give us finer grained signal about what's flaky. We can also easily use annotations to disable one subset or another.

But I'd suggest we do that in a separate PR so we don't keep blocking the arrikto E2E test.

yanniszark:
@jlewi as asked I reverted the extra coverage of the kf_is_ready test.
I filed https://github.com/kubeflow/kubeflow/issues/4294 to track this and I believe it's important to tackle it soon, since we currently don't test a lot of the installed applications.

Can you please comment on the issue with your suggestions on implementation?

Some questions I had on the approach you proposed were:
* How do we skip certain tests if this is split in multiple ones?
* What are test annotations and how are they used? Are they a pytest feature?
Let's continue this on https://github.com/kubeflow/kubeflow/issues/4294.

If the presubmit passes we should be good to go, based on your latest review.

PR #4155, opened 2019-09-19
profile v1beta1

kunmingg:
- Update Profile API to v1beta1
  Two new fields compared with v1alpha1:
  - Plugins: allow platform specific plugins like GCP workload identify setup
  - ResourceQuotaSpec: define resource spec of target namespace owned by the profile
  
  Detailed instructions at README.md 
- Update controller implementation with kube builder v2.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4155)


components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Plugin is for customize actions on different platform.
+type Plugin struct {
```
kkasravi: is this similar to the kfctl plugin?

components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
```diff
@@ -0,0 +1,83 @@
...
+	Owner   rbacv1.Subject `json:"owner,omitempty"`
+	Plugins []Plugin       `json:"plugins,omitempty"`
+	// Resourcequota that will be applied to target namespace
+	ResourceQuotaSpec v1.ResourceQuotaSpec `json:"resourcequotaspec,omitempty"`
+}
+
+type ProfileState string
+
+const (
+	ProfileSucceed ProfileState = "Succeed"
```
kkasravi: should this be Successful? Should it agree with ApplicationStatus or DeploymentStatus

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -249,16 +249,30 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
 			"defaultEdittor", "error", err)
 		return reconcile.Result{}, err
 	}
+	gcpPlugin := &GcpPlugin{}
```
yanniszark: I believe that Plugins should be called in an encapsulated way with a common entrypoint interface.
I wouldn't expect this common code to know about individual plugins.

components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
```diff
@@ -0,0 +1,83 @@
...
+// ProfileSpec defines the desired state of Profile
+type ProfileSpec struct {
+	// The profile owner
+	Owner   rbacv1.Subject `json:"owner,omitempty"`
+	Plugins []Plugin       `json:"plugins,omitempty"`
+	// Resourcequota that will be applied to target namespace
+	ResourceQuotaSpec v1.ResourceQuotaSpec `json:"resourcequotaspec,omitempty"`
+}
+
+type ProfileState string
```
yanniszark: I would suggest we avoid using enum fields in status, as they break if we add new possible values.
Instead, the K8s api convention advises one to use Conditions.
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md

components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
kunmingg: Yes, they'll follow similar pattern.

kunmingg:
@kkasravi 
@yanniszark 
profile plugin will behave similar to kfdef plugin:
each platform can define its own plugin spec and can do customized actions based on each spec.
for example GCP plugin will setup GCP credentials so user can talk to GCP APIs in their namespace.

components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
kunmingg: updated

components/profile-controller/pkg/controller/profile/profile_controller.go
kunmingg: updated

components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
kunmingg: ProfileState is saved as plain string, adding new values should not change existing messages.
We can use conditions in followup PRs if needed.

yanniszark: @kunmingg I'm a bit hesitant to go forward with a Phase enum-type status, because this is the pattern that Kubernetes used and then decided explicitly against it, in favor of conditions.
See [this section](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties) for more information: 

> pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted system-design principles and hampered evolution, since adding new enum values breaks backward compatibility. 


/cc @jlewi @kkasravi

krishnadurai:
This is great! Would be great to have resource quotas for each profile.
/cc @krishnadurai

jlewi:
@kunmingg If this is ready for review can you remove WIP? I generally ignore PRs that have WIP. If you want it to be reviewed but not merged; place a hold

Jeffwan:
This is great. I plan to add aws plugins to add istio policy for every namespace/profile. I think v1beta1 can meet my need.

kunmingg:
Current plan is to enable this new API in v0.7.

components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
jlewi: +1 to using conditions. 

@kunmingg are there problems with using conditions rather than enums?

components/profile-controller/pkg/apis/kubeflow/v1beta1/profile_types.go
```diff
@@ -0,0 +1,83 @@
...
+// Plugin is for customize actions on different platform.
+type Plugin struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Spec	*runtime.RawExtension `json:"spec,omitempty"`
+}
+
+type ProfileCondition struct {
+	Message	string `json:"message,omitempty"`
+	Status	string `json:"status,omitempty"`
```
yanniszark: I believe it's better to follow the standard condition fields here as well.
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

/cc @jlewi

jlewi: +1

kunmingg:
rebased to resolve conflicts.

jlewi:
/lgtm
/approve

/hold

@kunmingg looks like the tests are failing.

@yanniszark Did you want to take another look before we merge this?

components/profile-controller/api/v1beta1/profile_types.go
```diff
@@ -14,41 +13,51 @@ See the License for the specific language governing permissions and
...
+type Plugin struct {
+	metav1.TypeMeta `json:",inline"`
+	Spec            *runtime.RawExtension `json:"spec,omitempty"`
+}
+
+type ProfileCondition struct {
+	Type    string `json:"type,omitempty"`
+	Status  string `json:"status,omitempty" description:"status of the condition, one of True, False, Unknown"`
+	Message string `json:"message,omitempty"`
+}
```
yanniszark: @jlewi @kunmingg should we add all the standard fields here or leave it as is for now?

yanniszark:
Seems great!
I like how the Plugins are abstracted with an interface and treated uniformly.
/lgtm

PR #4166, opened 2019-09-21
Add condition operations to KfDef for reconcile semantics

gabrielwen:
Adds functions to update and read conditions so that reconcile function could read from it and determine whether to terminate or retry.

Relates to #4056 
Blocking #4141

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4166)


gabrielwen:
@jlewi @yanniszark do we need to check duplicated conditions, e.g. plugin condition to a plugin?

bootstrap/pkg/apis/apps/kfdef/v1beta1/application_types.go
```diff
@@ -160,6 +167,8 @@ type KfDefCondition struct {
 	LastUpdateTime metav1.Time `json:"lastUpdateTime,omitempty"`
 	// Last time the condition transitioned from one status to another.
 	LastTransitionTime metav1.Time `json:"lastTransitionTime,omitempty"`
+	// Name of the component this condition points to.
+	Name string `json:"name,omitempty"`
```
yanniszark: Should we use Reason instead of introducing a non-standard field?

jlewi:
@gabrielwen either way. coordinator can get a list of the plugins from the KFDef spec. So it will know the kind for any plugin listed in the KFDef spec. So coordinator.go could programmatically look for appropriate conditions based on the plugin kinds.

So I don't think we need to hard code the conditions.

gabrielwen:
@jlewi @yanniszark could you take another look?  this PR should be ready.

jlewi:
/hold
/lgtm
/approve

@yanniszark any additional comments?

jlewi:
/lgtm
/approve

@gabrielwen you can go ahead and remove the hold unless you want to wait for @yanniszark

Issue #3938, opened 2019-08-18
Define Kubeflow cluster role and combine roles  (was default roles should include virtualservices)


jlewi:
/kind feature

It looks like permission to create virtual services aren't attached by default to the roles created in namespaces. See below

**Why you need this feature:**
[Is your feature request related to a problem? Please describe in details]

editors should probably have the ability to edit virtual services so they can deploy services in their namespaces and configure traffic.

viewers should probably have viewer permissions.

**Describe the solution you'd like:**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Here is the edit role

```
kubectl get clusterrole edit -o yaml
aggregationRule:
  clusterRoleSelectors:
  - matchLabels:
      rbac.authorization.k8s.io/aggregate-to-edit: "true"
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    rbac.authorization.kubernetes.io/autoupdate: "true"
  creationTimestamp: 2019-08-16T04:04:19Z
  labels:
    kubernetes.io/bootstrapping: rbac-defaults
    rbac.authorization.k8s.io/aggregate-to-admin: "true"
  name: edit
  resourceVersion: "319"
  selfLink: /apis/rbac.authorization.k8s.io/v1/clusterroles/edit
  uid: ebc81501-bfda-11e9-970b-42010a8e010d
rules:
- apiGroups:
  - ""
  resources:
  - pods/attach
  - pods/exec
  - pods/portforward
  - pods/proxy
  - secrets
  - services/proxy
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - ""
  resources:
  - serviceaccounts
  verbs:
  - impersonate
- apiGroups:
  - ""
  resources:
  - pods
  - pods/attach
  - pods/exec
  - pods/portforward
  - pods/proxy
  verbs:
  - create
  - delete
  - deletecollection
  - patch
  - update
- apiGroups:
  - ""
  resources:
  - configmaps
  - endpoints
  - persistentvolumeclaims
  - replicationcontrollers
  - replicationcontrollers/scale
  - secrets
  - serviceaccounts
  - services
  - services/proxy
  verbs:
  - create
  - delete
  - deletecollection
  - patch
  - update
- apiGroups:
  - apps
  resources:
  - daemonsets
  - deployments
  - deployments/rollback
  - deployments/scale
  - replicasets
[...]

PR #4167, opened 2019-09-21
Kubeflow K8s roles used instead of K8s default roles

krishnadurai:
Partially Addresses: #3938 
~~This PR adds labels 'rbac.authorization.kubeflow.org/aggregate-to-kubeflow-{role}' to aggregate K8s default ClusterRoles to Kubeflow based ClusterRoles using Aggregated ClusterRoles.~~
This PR enables Kubeflow cluster roles defined in manifests to be associated with users using KFAM and Profiles Controller.

K8s default roles:
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#default-roles-and-role-bindings

Aggregated cluster roles:
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

/cc @johnugeorge 
/assign @krishnadurai

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4167)

### DIFF
--- a/bootstrap/pkg/kfapp/kustomize/kustomize.go
+++ b/bootstrap/pkg/kfapp/kustomize/kustomize.go
@@ -108,6 +108,13 @@ type kustomize struct {
 	configOverwrite bool
 }
 
+// k8sClusterRolePairs are used to define clusterrole names which need to be
+// aggregated to Kubeflow clusterroles.
+type k8sClusterRolePair struct {
+	clusterRoleName         string
+	clusterAggregationLabel string
+}
+
 const (
 	defaultUserId = "anonymous"
 	outputDir     = "kustomize"
@@ -250,6 +257,38 @@ func (kustomize *kustomize) initK8sClients() error {
 	return nil
 }
 
+func applyK8sAggregationLabels(apply *utils.Apply) error {
+	var k8sClusterRolePairs = []k8sClusterRolePair{
+		k8sClusterRolePair{
+			clusterRoleName:         "cluster-admin",
+			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-cluster-admin",
+		},
+		k8sClusterRolePair{
+			clusterRoleName:         "admin",
+			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-admin",
+		},
+		k8sClusterRolePair{
+			clusterRoleName:         "edit",
+			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-edit",
+		},
+		k8sClusterRolePair{
+			clusterRoleName:         "view",
+			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-view",
+		},
+	}
+	for _, clusterRolePair := range k8sClusterRolePairs {
+		labelErr := apply.LabelClusterRole(
+			clusterRolePair.clusterRoleName,
+			clusterRolePair.clusterAggregationLabel,
+			"true",
+		)
+		if labelErr != nil {
+			return labelErr
+		}
+	}
+	return nil
+}
+
 // Apply deploys kustomize generated resources to the kubenetes api server
 func (kustomize *kustomize) Apply(resources kftypesv3.ResourceEnum) error {
 	var restConfig *rest.Config = nil
@@ -261,6 +300,10 @@ func (kustomize *kustomize) Apply(resources kftypesv3.ResourceEnum) error {
 		return err
 	}
 
+	labelErr := applyK8sAggregationLabels(apply)
+	if labelErr != nil {
+		return labelErr
+	}
 	kustomizeDir := path.Join(kustomize.kfDef.Spec.AppDir, outputDir)
 	for _, app := range kustomize.kfDef.Spec.Applications {
 		resMap, err := EvaluateKustomizeManifest(path.Join(kustomizeDir, app.Name))
--- a/bootstrap/pkg/kfapp/kustomize/kustomize.go
+++ b/bootstrap/pkg/kfapp/kustomize/kustomize.go
@@ -108,13 +108,6 @@ type kustomize struct {
 	configOverwrite bool
 }
 
-// k8sClusterRolePairs are used to define clusterrole names which need to be
-// aggregated to Kubeflow clusterroles.
-type k8sClusterRolePair struct {
-	clusterRoleName         string
-	clusterAggregationLabel string
-}
-
 const (
 	defaultUserId = "anonymous"
 	outputDir     = "kustomize"
@@ -257,38 +250,6 @@ func (kustomize *kustomize) initK8sClients() error {
 	return nil
 }
 
-func applyK8sAggregationLabels(apply *utils.Apply) error {
-	var k8sClusterRolePairs = []k8sClusterRolePair{
-		k8sClusterRolePair{
-			clusterRoleName:         "cluster-admin",
-			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-cluster-admin",
-		},
-		k8sClusterRolePair{
-			clusterRoleName:         "admin",
-			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-admin",
-		},
-		k8sClusterRolePair{
-			clusterRoleName:         "edit",
-			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-edit",
-		},
-		k8sClusterRolePair{
-			clusterRoleName:         "view",
-			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-view",
-		},
-	}
-	for _, clusterRolePair := range k8sClusterRolePairs {
-		labelErr := apply.LabelClusterRole(
-			clusterRolePair.clusterRoleName,
-			clusterRolePair.clusterAggregationLabel,
-			"true",
-		)
-		if labelErr != nil {
-			return labelErr
-		}
-	}
-	return nil
-}
-
 // Apply deploys kustomize generated resources to the kubenetes api server
 func (kustomize *kustomize) Apply(resources kftypesv3.ResourceEnum) error {
 	var restConfig *rest.Config = nil
@@ -300,10 +261,6 @@ func (kustomize *kustomize) Apply(resources kftypesv3.ResourceEnum) error {
 		return err
 	}
 
-	labelErr := applyK8sAggregationLabels(apply)
-	if labelErr != nil {
-		return labelErr
-	}
 	kustomizeDir := path.Join(kustomize.kfDef.Spec.AppDir, outputDir)
 	for _, app := range kustomize.kfDef.Spec.Applications {
 		resMap, err := EvaluateKustomizeManifest(path.Join(kustomizeDir, app.Name))
--- a/bootstrap/pkg/utils/k8utils.go
+++ b/bootstrap/pkg/utils/k8utils.go
@@ -15,6 +15,7 @@ package utils
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"github.com/ghodss/yaml"
 	configtypes "github.com/kubeflow/kubeflow/bootstrap/v3/config"
@@ -313,6 +314,54 @@ func (a *Apply) namespace(namespace string) error {
 				Message: fmt.Sprintf("couldn't create %v %v Error: %v",
 					string(kftypes.NAMESPACE), namespace, nsErr),
 			}
+		}
+	}
+	return nil
+}
+
+//LabelClusterRole labels ClusterRoles with labelKey: labelValue pair
+func (a *Apply) LabelClusterRole(clusterRoleName string, labelKey string,
+	labelValue string) error {
+	clusterRole, clusterRoleMissingErr := a.clientset.RbacV1().ClusterRoles().Get(clusterRoleName,
+		metav1.GetOptions{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "rbac.authorization.k8s.io/v1beta1",
+				Kind:       "ClusterRole",
+			},
+		})
+	if clusterRoleMissingErr == nil {
+		var labelPatchMap = map[string]map[string]map[string]string{
+			"metadata": map[string]map[string]string{
+				"labels": map[string]string{
+					labelKey: labelValue,
+				},
+			},
+		}
+		labelPatchJSON, jsonErr := json.Marshal(labelPatchMap)
+		if jsonErr != nil {
+			return &kfapis.KfError{
+				Code: int(kfapis.INVALID_ARGUMENT),
+				Message: fmt.Sprintf("Couldn't marshal labelPatchJSON %v Error: %v",
+					labelPatchMap, jsonErr),
+			}
+		}
+		log.Infof("Labeling ClusterRole: %v", clusterRoleName)
+		_, labelingErr := a.clientset.RbacV1().ClusterRoles().Patch(
+			clusterRoleName,
+			"application/strategic-merge-patch+json",
+			[]byte(labelPatchJSON),
+		)
+		if labelingErr != nil {
+			return &kfapis.KfError{
+				Code: int(kfapis.INTERNAL_ERROR),
+				Message: fmt.Sprintf("Couldn't label ClusterRole %v Error: %v",
+					clusterRole, labelingErr),
+			}
+		}
+	} else {
+		return &kfapis.KfError{
+			Code:    int(kfapis.NOT_FOUND),
+			Message: fmt.Sprintf("Default Kubernetes role %v not found.", clusterRole),
 		}
 	}
 	return nil
--- a/bootstrap/pkg/utils/k8utils.go
+++ b/bootstrap/pkg/utils/k8utils.go
@@ -15,7 +15,6 @@ package utils
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"github.com/ghodss/yaml"
 	configtypes "github.com/kubeflow/kubeflow/bootstrap/v3/config"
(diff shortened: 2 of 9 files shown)


krishnadurai:
Added kubeflow/manifests#374 which shows Kubeflow ClusterRoles and how each application can define their ClusterRoles to aggregate to them.

krishnadurai:
TODOs:
[ ] - Test on a fresh GCP cluster for Profile namespace roles correctness
[ ] - Raise separate manifests PR defining roles for the following Kubeflow Resources:
experiments.kubeflow.org
kfservices.serving.kubeflow.org
notebooks.kubeflow.org
poddefaults.kubeflow.org
profiles.kubeflow.org
pytorchjobs.kubeflow.org
scheduledworkflows.kubeflow.org
tfjobs.kubeflow.org
trials.kubeflow.org
viewers.kubeflow.org

krishnadurai:
/assign @jlewi
/assign @kunmingg 
This is ready for review.

krishnadurai:
/hold

For all reviews to be completed.

jlewi:
Thanks @krishnadurai 

Why are we using go code to add the labels rather than just including the labels in the kustomize manifests?

components/access-management/kfam/bindings.go
```diff
@@ -32,6 +32,15 @@ const ServiceRoleBinding = "servicerolebindings"
 const SERVICEROLEISTIO = "ns-access-istio"
 const USER = "user"
 const ROLE = "role"
+//roleBindingNameMap maps frontend role names to k8s role names and vice-versa
+var roleBindingNameMap = map[string]string {
+	"kubeflow-admin":	"admin",
```
kunmingg: I guess we don't need mapping for both directions?

krishnadurai:
@jlewi @kunmingg PTAL

kunmingg:
For example we have patch like:
https://github.com/kubeflow/manifests/blob/master/gcp/cloud-endpoints/base/gcp-credentials-patch.yaml

I think you can put roles annotation patch at https://github.com/kubeflow/manifests/tree/master/profiles/base

krishnadurai:
@kunmingg In gcp-credentials patch, cloud-enpoints-controller is defined in the same kustomization folder. In the current case, we are not including the default Kubernetes roles within our manifests and looking to patch them - which I'm not sure that it can be done with Kustomize.

kunmingg:
@krishnadurai 
Got it
@jlewi If we want to apply patch with kustomize, we need to call `patch()` inside `kustomize apply()`, since the clusterrole we want to patch is not created by our menifests.

kunmingg:
Alternatively, we can leave k8s roles unchanged, only aggregate CR roles into `kubeflow-edit`.
So we'll sycn three binding:
```
k8s role binding
istio role binding
kubeflow role binding (new, for all CRs)
```

krishnadurai:
> Alternatively, we can leave k8s roles unchanged, only aggregate CR roles into `kubeflow-edit`.
> So we'll sycn three binding:

@kunmingg this seems like a good approach too - just that we would have to maintain a role mapping for each KubeFlow Role and the corresponding isito and k8s roles associated with them at kfam and profile-controller.

kunmingg:
Currently we maintain 2 role mappings, adding one more mapping for robustness should be OK.
My worry is we don't own or create k8s default roles, and our patch could be temporary if user decide to upgrade or replace master.

bootstrap/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -250,6 +257,38 @@ func (kustomize *kustomize) initK8sClients() error {
...
+			clusterRoleName:         "edit",
+			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-edit",
+		},
+		k8sClusterRolePair{
+			clusterRoleName:         "view",
+			clusterAggregationLabel: "rbac.authorization.kubeflow.org/aggregate-to-kubeflow-view",
+		},
+	}
+	for _, clusterRolePair := range k8sClusterRolePairs {
+		labelErr := apply.LabelClusterRole(
```
yanniszark: nit: would it be better to follow the err convention like the rest of the file?

bootstrap/pkg/utils/k8utils.go
```diff
@@ -318,6 +319,52 @@ func (a *Apply) namespace(namespace string) error {
 	return nil
 }
 
+//LabelClusterRole labels ClusterRoles with labelKey: labelValue pair
+func (a *Apply) LabelClusterRole(clusterRoleName string, labelKey string,
```
yanniszark: Since this is a very specific function needed in a certain place only, I believe it would be better to contain it within that file (kustomize.go)?
You can create a k8s client there as well.

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -238,7 +240,7 @@ func (r *ReconcileProfile) Reconcile(request reconcile.Request) (reconcile.Resul
 		RoleRef: rbacv1.RoleRef{
 			APIGroup: "rbac.authorization.k8s.io",
 			Kind:     "ClusterRole",
-			Name:     ADMIN,
+			Name:     KUBEFLOW_PREFIX + ADMIN,
```
yanniszark: I think it's preferable to be explicit here and pass the names we want to see created (eg kubeflow-editor instead of edit):
https://github.com/kubeflow/kubeflow/blob/5ebad34ce3e18727dda8288880cc25b3d230b568/components/profile-controller/pkg/controller/profile/profile_controller.go#L216

krishnadurai:
I see the benefit of this approach you've mentioned here since we do not own the Kubernetes default ClusterRole manifest. I think we can include this change without needing to patch.
In this approach, would it be necessary to have a separate Istio role-binding, since we are defining the roles through our manifest?

Let's wait for @jlewi's input so that I'll be able to implement these changes at one go.

kunmingg:
@krishnadurai 
The challenge around Istio rbac is:
it defines it's own access policy as CRs, not following k8s rbac API, so likely it cannot be aggregated.
Would be great if we can find a way to manage Istio role binding under kubeflow-role?

components/profile-controller/pkg/controller/profile/profile_controller.go
krishnadurai: @yanniszark 
> Regarding the need to patch the default ClusterRoles, isn't it only needed for ClusterRole view?
> We can get the permissions of the other ClusterRoles by adding their aggregationRules labelSelector.

IIUC, does this mean that we aggregate to default ClusterRoles Kubernetes provides and create RoleBindings against it? We were looking to aggregate the default ClusterRoles into our own Kubeflow ones, allowing for custom roles to be composed in the future. 

> 
> A final thought is this:
> Is it a stretch to define our own K8s kubeflow-viewer ClusterRole, with a set of permissions that is fairly stable? (eg Pods, Services, {Stateful, Daemon, Replica}Set, Deployments, Jobs, etc).

It shouldn't be a stretch, though the thinking here was to avoid having to maintain and update our own when Kubernetes provides us with its own default roles.

krishnadurai:
@kunmingg I get what you mean here - the Istio RBAC roles are the ones being defined by the Profile Controller and Manifests. I mistook it for the Istio CR roles for Isito resources (like editing virtualservices) defined in the manifest PRs.

bootstrap/pkg/utils/k8utils.go
krishnadurai: It makes sense - I'll make this change.

components/profile-controller/pkg/controller/profile/profile_controller.go
krishnadurai: I agree explicit naming would be better.

krishnadurai:
> @krishnadurai thanks for the effort on consolidating a strategy for RBAC permissions.
> I have left some comments below.
> Regarding the need to patch the default ClusterRoles, isn't it only needed for ClusterRole view?
> We can get the permissions of the other ClusterRoles by adding their aggregationRules labelSelector.
> 
> A final thought is this:
> Is it a stretch to define our own K8s kubeflow-viewer ClusterRole, with a set of permissions that is fairly stable? (eg Pods, Services, {Stateful, Daemon, Replica}Set, Deployments, Jobs, etc).

components/profile-controller/pkg/controller/profile/profile_controller.go
yanniszark: > > Regarding the need to patch the default ClusterRoles, isn't it only needed for ClusterRole view?
> > We can get the permissions of the other ClusterRoles by adding their aggregationRules labelSelector.

> IIUC, does this mean that we aggregate to default ClusterRoles Kubernetes provides and create RoleBindings against it?

I had a lapse in my thinking there.
I was thinking of something along the lines of:
* We want kubeflow-viewer to get view permissions.
* Add aggregate-to-edit labelSelector to kubeflow-viewer, so that it will get the view permissions, like the edit ClusterRole does.

However, this is not ideal, as other applications in the Cluster may be aggregating to the edit ClusterRole.

To sum up, I see the following solutions:
* Edit the default ClusterRoles and add the necessary labels.
* Define our own base ClusterRoles.

jlewi: @krishnadurai @yanniszark so to summarize the key question is

1) Are Kubeflow roles the result of aggregating default K8s roles
or 
2) Do we define our roles for basic K8s permissions that users should have? i.e. we create kubeflow specific default roles that K8s defines and then we aggregate those roles

Option 1 has the advantages

* Depend on K8s to automatically update the roles with new resources over time

Option 1 has the disadvantage

* Kubeflow needs to modify the default roles to add the aggregation labels
* Kubeflow's default permissions are less explicit over time and potentially different K8s flavors/versions

Option 2 has the advantage 

* Kubeflow doesn't need to modify default cluster roles
* Kubeflow is very explicit about what the default permissions are

Option 2 has the disadvantage

* Kubeflow needs to explicitly modify the default roles over time
* Will we need to parameterize the default role because different K8s versions have different versions of resources? e.g. extension/v1beta1 vs. v1?

I tend to lean towards option 2 because it seems like if we start with 2 we could change our mind and still provide backwards compatiblity easier then with option 1.

If users want need to include the default rules in the aggregation then they could always manually add the label to the default roles. We could use that to validate the utility of doing that and eventually do it automatically.

krishnadurai: @jlewi This sounds good. Let's define our own Kubeflow roles for K8s resources. I'll change this PR to remove the labeling part and retain the Profile Controller and Kfam  parts. Let's have all ClusterRoles aggregate to Kubeflow roles in the manifest PR. I'll raise a separate PR defining a ClusterRole for Kubernetes resources.

krishnadurai: Another question is that should we start explicitly supporting a range of Kubernetes versions for each release if we are already not doing it?

jlewi: @krishnadurai LGTM to the first part.

@krishnadurai Yes I think we will need to start thinking about which versions of K8s we support and how to support multiple versions.

According to
https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles

It looks like role aggregation was introduced in 1.9. I think K8s is currently on 1.16 so that's 7 releases almost two years ago. 

I don't think we ever supported versions that low so I don't think this change is impacting our ability to support K8s versions currently used but let me know if I'm missing something.

krishnadurai: @jlewi I just wanted to ensure that the roles which we are defining would work across these supported K8s versions.

krishnadurai:
Test failed with error:
```
Error: couldn't delete KfApp:  (kubeflow.error): Code 500 with message: coordinator Delete failed for gcp:  (kubeflow.erro
r): Code 400 with message: Error when cleaning IAM policy: googleapi: Error 409: There were concurrent policy changes. Please retry the whole read-modify-write wi
th exponential backoff., aborted
```

jlewi:
@krishnadurai Please let us know when this is ready for another look.

krishnadurai:
@jlewi Sure. Expect this to be ready later today.

krishnadurai:
This PR along with kubeflow/manifests#388 is ready for review.

@kunmingg @yanniszark @jlewi

jlewi:
On my end
/lgtm
/approve

krishnadurai:
What we may show in the frontend can differ from our ClusterRole names in K8s, though we need to keep the RoleBinding names in a namespace consistent with whats shown in the frontend. Frontend shouldn't be aware of the names of these K8s ClusterRoles in the backend.

> Are those KF related?

These are KF related.

> If your using GCP then I'm guessing
> user-xyz-gmail-com-clusterrole-edit
>
> is the granting cluster admin rights to the user running kfctl. So that seems fine.

This RoleBinding grants privileges to a contributor added through Manage Users - here's where the naming can be closer to the frontend names for roles.
 
> Not sure about namespace-admin.

namespaceAdmin is for the owner of the namespace.

> IIRC we were previously using "default-editor" and "default-viewer". Were those created by us or are those default K8s rules.

These are associated with service accounts used for user impersonation through Profiles.

krishnadurai:
@kunmingg @yanniszark can you please take a look so that this can be integrated?

components/access-management/kfam/bindings.go
```diff
@@ -32,6 +32,15 @@ const ServiceRoleBinding = "servicerolebindings"
 const SERVICEROLEISTIO = "ns-access-istio"
 const USER = "user"
 const ROLE = "role"
+//roleBindingNameMap maps frontend role names to k8s role names and vice-versa
```
yanniszark: Why does the KFAM API need to know anything about the frontend?
It seems to me that they should be cleanly separated.
Can you explain why this is needed?

components/profile-controller/pkg/controller/profile/profile_controller.go
```diff
@@ -52,6 +52,22 @@ const USER = "user"
...
+// Kubeflow default role names
+// TODO: Make kubeflow roles configurable (krishnadurai)
+// This will enable customization of roles.
+const(
+	kubeflowAdmin =	"kubeflow-admin"
+	kubeflowEdit =	"kubeflow-edit"
+	kubeflowView =	"kubeflow-view"
+)
+
+//roleBindingNameMap maps frontend role names to k8s role names
```
yanniszark: Again, this seems very weird to me.
The frontend is a client of this API, so it's as if we are letting the client decide what the API is, instead of defining it here.

This brings me to a second observation.
There are 2 places where binding creation takes place.
1. The profile-controller
2. The KFAM API

I believe we should standardize and do it in one place only.
In this case, should the profile-controller also use the KFAM API to create a binding?
I think this will also reduce the need to define the same names in different places.
/cc @kunmingg @jlewi

components/access-management/kfam/bindings.go
krishnadurai: @yanniszark the frontend now sends the values 'admin', 'edit' and 'view' to the back-end to identify roles. This happens to be the same names as the default Kubernetes cluster roles and they were passed down as is to the ClusterRole's creation code.

components/profile-controller/pkg/controller/profile/profile_controller.go
krishnadurai: @yanniszark I agree with your view and I think we should address this. The way I'm thinking this can be done is through making Kfam the sole owner of creating, listing and deleting the RoleBinding resources. Likewise when we should be able to make these roles more composable, I think we could store these role definitions or mappings in a storage instead - this is why I've marked a TODO.
Given the scope of what we are trying to achieve in this PR I think these changes should not halt this PR.

krishnadurai: I'll clarify here that the client is not defining the back-end API parameter values. I think the values used by the back-end API may or should not be the same as what the cluster role names are because the names which we use might clash with other applications in K8s as well. Here I feel the values 'admin', 'edit' and 'view' should be the values accepted by the API whereas the bindings may have a 'kubeflow' prefix.

krishnadurai:
@jlewi I made an error here and as you've suggested - I've corrected it.

What @yanniszark and I were discussing is that it would be better to have all rolesBindings created by KFAM. So here, profile controller would use KFAM to create all roleBindings by an API call.

jlewi:
@krishnadurai thanks 

/lgtm
/approve

/hold

@krishnadurai I put a hold in case you want to wait for @yanniszark to take another look on my end this is LGTM.

@krishnadurai @yanniszark regarding whether or not to use KFAM to create all role bindings; I'd generally like to rely on KFAM less and use the K8s APIServer and roles/rolebindings directly whenever possible rather than creating a layer of indirection around RBAC.

IIRC there were primarily 2 use cases for which we created KFAM

1. To provide a simple API call for the central dashboard to figure out what namespaces to list in the central dashboard
    * The goal here was to abstract from the front end the particulars on how it was determined which
      namespaces a user has access to 

1. A method call to delete profiles

    * We want the owner of a profile to be able to delete that profile but not other profiles
    * RBAC doesn't give us this level of granularity which is why we do it through KFAM.

krishnadurai:
@yanniszark I'll wait for your comments. PTAL.

yanniszark:
@krishnadurai I've got this on my TODOs for tomorrow.
Great work!

yanniszark:
@krishnadurai maybe I missed something.
Didn't we decide on defining explicit kubeflow ClusterRoles?
If yes, is the logic to label existing ClusterRoles not necessary?

krishnadurai:
@yanniszark I think you missed kubeflow/manifests#388 - this defines the Kubernetes default roles. This does not require us to label using go, so I've removed that from this PR.
The parts retained here are the KFAM and Profile controller parts.

yanniszark:
/lgtm

I believe that we need to change the Binding so that it either has the real role name or a magic string.
Right now, it says it has a RoleRef but it's actually a magic string.
Let's open an issue to track that.

krishnadurai:
> I believe that we need to change the Binding so that it either has the real role name or a magic string.
> Right now, it says it has a RoleRef but it's actually a magic string.
> I'll open an issue to track that.

I didn't quite get this part. Can you please explain this in the issue you'll open?

krishnadurai:
@kunmingg @jlewi The build fails with:
```
Unpacking python3-cryptography (2.1.4-1ubuntu1.3) ...
Selecting previously unselected package python3.6-dev.
Preparing to unpack .../126-python3.6-dev_3.6.8-1~18.04.2_amd64.deb ...
Unpacking python3.6-dev (3.6.8-1~18.04.2) ...
ERROR: (gcloud.builds.submit) build 0e95a77e-93ec-40bc-b533-fb8fbc21e0b8 completed with status "TIMEOUT"
```

Is there something I can do about this to make the pre-submits pass?

yanniszark:
https://github.com/kubeflow/kubeflow/issues/4217

components/access-management/kfam/bindings.go
jlewi: @krishnadurai do we need both directions?

components/access-management/kfam/bindings.go
jlewi: @krishnadurai can you add a todo to fix this?

jlewi:
@krishnadurai rebase on master. I think if you do that you won't trigger the jupyter release workflow.

components/access-management/kfam/bindings.go
krishnadurai: @jlewi referring to the comment (here)[https://github.com/kubeflow/kubeflow/pull/4167#pullrequestreview-292398198].

components/access-management/kfam/bindings.go
krishnadurai: I've replied on @yanniszark's issue regarding this: https://github.com/kubeflow/kubeflow/issues/4217#issuecomment-537438075

krishnadurai:
/hold cancel

This requires an LGTM again

PR #4176, opened 2019-09-23
Update C2D UI to improve operation status reporting.

prodonjs:
- Estimate IAP deployment time to 30m
- Check for IAP at 30s intervals and countdown estimated time from 30m.
- Check for IAP status indefinitely (instead of 25m).

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4176)

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -148,7 +148,7 @@ const MYSQL = 'mysql';
 const PASSWORD = 'password';
 const PROFILES = 'profiles';
 const SPARTAKUS = 'spartakus';
-const nameformat = '[a-z]([-a-z0-9]*[a-z0-9])?';
+const NAME_FORMAT = '[a-z]([-a-z0-9]*[a-z0-9])?';
 
 const styles: {[key: string]: React.CSSProperties} = {
     btn: {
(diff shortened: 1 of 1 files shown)


prodonjs:
/assign @kunmingg 
/area front-end

@kunmingg - This addresses the issue by removing the main timeout for IAP-based deployments. I think there still might be room for improvement to detect if the image load returns a 302 asking the user to authorize but I wanted to try to get something in place based on the feedback from Amy and Karl.

components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -765,31 +765,31 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
             // an image served by the target site, the img load is a simple html
             // request and not an AJAX request, thus bypassing the CORS in this
             // case.
-            this._appendLine('Validating if IAP is up and running...');
+            const expectedTimeSecs = 30 * 60; // 30m
             const startTime = new Date().getTime() / 1000;
             const img = document.createElement('img');
-            img.src = dashboardUri + 'assets/favicon-32x32.png' + '?rand=' + Math.random();
+            const iconFilename = this.state.kfversion === Version.V06 ?
+                'favicon-32x32.png' : 'kf-logo_64px.svg';
```
kunmingg: If user change version during wait time, we might be watching the wrong path?
Should we still keep pulling both path for v0.5 and v0.6?

prodonjs: Ahh good catch. I changed the other references to `this.state` to be captured in the closure. I'll fix this.

prodonjs:
@kunmingg - Updated `_redirectToKFDashboard` to use a state captured at the time the Deployment request is issued. Here's a screenshot of what the countdown will look like.

e6iKT2mbpVq

PR #4177, opened 2019-09-23
C2D: Fix bugs in _enableGcpServices.

prodonjs:
- Loosen logic to check for "falsy" values rather than explicit nulls/undefineds.
- Move "sleeps" outside of try block to ensure they are used in each loop iteration.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4177)


prodonjs:
/assign @kunmingg 
/area front-end

@kunmingg - This should address the 401 errors when deploying on a brand new project. The primary issue was the position of the `await wait()` statements, which would not take effect if the catch clauses were reached. That was an omission on my part when we did the big refactoring for v0.6.

Issue #1933, opened 2018-11-09
Fresh install renders 'no matches for kind "Application"'


pdmack:
I've seen this twice but it seems to remedy itself by regenerating and reapplying the ks app.

/cc @kkasravi 

```
+ ks apply default -c application
INFO Applying customresourcedefinitions applications.app.k8s.io
INFO Creating non-existent customresourcedefinitions applications.app.k8s.io
ERROR handle object: patching object from cluster: merging object with existing state: unable to recognize "/tmp/ksonnet-mergepatch884886323": no matches for kind "Application" in version "app.k8s.io/v1beta1"
```

pdmack:
Yes

jlewi:
@kkasravi @pdmack any idea whether this is still happening?

cheyang:
I'm able to reproduce it in brachn `v0.4-branch`. Are there any workarounds?

```
${KUBEFLOW_SRC}/scripts/kfctl.sh apply k8s
+ ks apply default -c application
INFO Updating customresourcedefinitions applications.app.k8s.io
INFO  Creating non-existent customresourcedefinitions applications.app.k8s.io
ERROR handle object: merging object with existing state: unable to recognize "/tmp/ksonnet-mergepatch060322179": no matches for kind "Application" in version "app.k8s.io/v1beta1"
```

xuchenCN:
Hi @cheyang  I encountered the same issue , but its disppear when I re-apply the ksonnet ,  have some idea for that ?

kkasravi:
you should try 
```
      ks show default -c metacontroller -c application > default.yaml
      kubectl apply --validate=false -f default.yaml
```

cheyang:
> Hi @cheyang I encountered the same issue , but its disppear when I re-apply the ksonnet , have some idea for that ?
> BTW , may I ask what do you do to overcome the GFW to pull the gcr.io images ?

You can run this on Alibaba Cloud by following the document below which is in Chinese: https://yq.aliyun.com/articles/686672

13567436138:
when i apply  ks show default -c metacontroller -c application > default.yaml
I got  ks show default -c metacontroller -c application > default.yaml
what‘s going on？

13567436138:
got  ERROR finding app root from starting path: : unable to find ksonnet project

jlewi:
I observed this error. The logs clearly indicate the apps resource CR was created.

```
[...]

PR #4187, opened 2019-09-26
kfctl#36 Split build and deploy to separate tests

nrchakradhar:
For kfctl#36, I have separated build and deploy into separate modules

- kfctl_go_build_test.py
- kfctl_go_deploy_test.py

In kfctl_go_deploy, included an error if kfctl go binary is not found

Once #4148 is merged,
https://github.com/kubeflow/kubeflow/blob/18aa0732a9479c4cce7ad05408288c9a4d5b62cc/py/kubeflow/kubeflow/ci/kfctl_e2e_workflow.py#L514-L520

will be changed to include new step for build which does not need any args at the moment.

Also, after #4154 is merged, some minor changes might be needed.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4187)


jlewi:
Pro tip make sure reviewers are listed as Assignees otherwise it won't show up in their review dashboard.
You can also explicitly /assign reviewers
/assign @jlewi

jlewi:
Thanks @nrchakradhar 

Should we move these files into 
https://github.com/kubeflow/kubeflow/tree/master/py/kubeflow/kubeflow/ci

Then kfctl_go_test.py could import and reuse the functions defined in those files.

This way we don't end up duplicating the code and we can make changes in a single location.

nrchakradhar:
> Then kfctl_go_test.py could import and reuse the functions defined in those files.
> 
> This way we don't end up duplicating the code and we can make changes in a single location.

@jlewi Do we want to have kfctl_go_test.py going forward? I thought it was not required. Once the workflow is moved to python, e2e test can have build, deploy as steps and kfctl_go_test.py can be removed. The test semantics will be a DAG of steps of independent functionalities. This way there will be only ONE way of e2e tests and not combined multiple steps in DAG into a single python test. 

Did not remove the file now in this commit as submit test will fail. When rebasing after #4148 merge, I can delete the file.
Please let me know your thoughts and I can change accordingly.

jlewi:
We do want to get rid of kfctl_go_test.py. But @yanniszark is planning on continuing to use it in the Arrikto E2E test #4154. So it will live on for a while.

In the meantime the semantics of kfctl are changing (#4040) so we need to update the tests to use the new semantics.

So we'd like to be able to have this logic in a single place and not have to update multiple locations.

nrchakradhar:
@jlewi I will make the changes and submit tomorrow. I got busy with other activities.

jlewi:
Yes this is very much what I had in mind. I might suggest moving the new pytest modules into py/kubeflow/kubeflow/ci as well.

The only reason for this is because by convention we always add the py directory to our python path. So this way all the modules are on our python path and we could import them if need be

swiftdiaries:
/cc @swiftdiaries 
+1 to @jlewi 's comments
Can we wait on removing the `kfctl_go_test.py` until #4040 goes in and we can test the new semantics?

nrchakradhar:
@jlewi I have moved the files as per comment. I am not able to resolve the python paths though.
kfctl_go_test.py:5: in <module>
   from kubeflow.kubeflow.ci import kfctl_go_test_utils as kfctl_util
E   ImportError: No module named 'kubeflow.kubeflow'

Can you please help.

jlewi:
It looks like the PYTHONPATH for the step doesn't include kubeflow/kubeflow/py.
If you look at the argo workflow spec
https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubeflow_kubeflow/4187/kubeflow-presubmit/1178293231973044224/kubeflow-presubmit-kfctl-go-iap-istio-4187-b60254a-4224-510f.yaml

the PYTHONPATH is

```
      - name: PYTHONPATH
        value: /mnt/test-data-volume/kubeflow-presubmit-kfctl-go-iap-istio-4187-b60254a-4224-510f/src/kubeflow/kubeflow:/mnt/test-data-volume/kubeflow-presubmit-kfctl-go-iap-istio-4187-b60254a-4224-510f/src/kubeflow/testing/py:/mnt/test-data-volume/kubeflow-presubmit-kfctl-go-iap-istio-4187-b60254a-4224-510f/src/kubeflow/tf-operator/py
```

I think you need to change this line
https://github.com/kubeflow/kubeflow/blob/4a1c06b7f608b434c749c1bca51b7850de61cf27/testing/workflows/components/kfctl_go_test.jsonnet#L100

To

```
 name: "PYTHONPATH",
 value: kubeflowPy + ":" +  kubeflowPy + "/py" + ":" + kubeflowTestingPy, 
```

nrchakradhar:
> I think you need to change this line
> https://github.com/kubeflow/kubeflow/blob/4a1c06b7f608b434c749c1bca51b7850de61cf27/testing/workflows/components/kfctl_go_test.jsonnet#L100

I guess we have switched to using kfctl_e2e_workflow.py. I changed the PYTHONPATH there and pre-submit has passed.

@jlewi Let me know next steps.

jlewi:
@nrchakradhar you are correct; sorry for the misleading information.

jlewi:
@nrchakradhar LGTM but it looks like you might have done a merge and not a rebase because the commit history shows 27 commits. Some of which aren't yours.

Could you please fix this by doing a rebase? There are some instructions in the first comment in this issue.
https://github.com/kubeflow/website/issues/1094

You will want to rebase dropping all commits that aren't yours and part of this PR. You can squash the remaining commits.

nrchakradhar:
@jlewi I have done the changes. PTAL.

nrchakradhar:
@jlewi 
I rebased the changes and after that except for one test all have passed. 
With my limited understanding, my analysis of [logs](http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-kfctl-go-basic-auth-4187-6d88a99-8016-3680?tab=workflow&nodeId=kubeflow-presubmit-kfctl-go-basic-auth-4187-6d88a99-8016-3680-538352680&sidePanel=logs%3Akubeflow-presubmit-kfctl-go-basic-auth-4187-6d88a99-8016-3680-538352680%3Amain) points to "Application" CR not installed. There have been couple of issues related to synchronization like #1933 but are quite old.
The same workflow for other cases with IAP are passing.

Can you please let me know how can I fix this?

Also, I found that its not easy to copy the logs from browser to other textpad/notepad/vim etc. Is there some way to do this?

py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
```diff
@@ -0,0 +1,238 @@
...
+  config_spec["metadata"]["name"] = kfdef_name
+
+  repos = config_spec["spec"]["repos"]
+  if os.getenv("REPO_NAME") == "manifests":
+    # kfctl_go_test.py was triggered on presubmit from the kubeflow/manifests
+    # repository. In this case we want to use the specified PR of the
+    # kubeflow/manifests repository; so we need to change the repo specification
+    # in the KFDef spec.
+    # TODO(jlewi): We should also point to a specific commit when triggering
+    # postsubmits from the kubeflow/manifests repo
```
swiftdiaries: Can we address this comment in this PR?
If not we can come back to it at a later point.

swiftdiaries:
/lgtm
Pending a pass on the presubmit.

py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
nrchakradhar: I can include this, but need information on to change to what as I am still learning all the CI stuff that is happening here.

nrchakradhar:
@jlewi @swiftdiaries Presubmit has passed. PTAL.
I can address any additional comments either in the PR or a follow up as per your suggestions.

py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
swiftdiaries: Cool. We can open an issue and address it soon.

swiftdiaries:
I don't have approve rights, @richardsliu / @kunmingg  could you PTAL?

PR #4191, opened 2019-09-26
Kfctl upgrade py test

jlewi:
Create a skeleton workflow for doing upgrades.

* We need to write an E2E test for kfctl upgrade (kubeflow/kfctl#35).

* This PR defines a new python class to define the upgrade E2E test workflow.

  * It subclasses the existing e2e_workflow.py in order to reuse the existing
    logic for building kfctl, deploying kubeflow, and validating kubeflow.

  * The modified workflow runs kfctl upgrade after first deploying Kubeflow.

    * This is actually just a place holder which prints out the command to
      be executed; this will need to be updated after we actually write a
      pytest to do the upgrade.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4191)

### DIFF
--- a/py/kubeflow/kubeflow/ci/kfctl_e2e_workflow.py
+++ b/py/kubeflow/kubeflow/ci/kfctl_e2e_workflow.py
@@ -577,8 +577,11 @@ class Builder:
               ]
 
       dependencies = [build_kfctl["name"]]
-      endpoint_ready = self._build_step(step_name, self.workflow, E2E_DAG_NAME, task_template,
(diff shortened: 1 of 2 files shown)


richardsliu:
@jlewi Is this ready for review?

jlewi:
@richardsliu I think its outdated. The PR should probably move over to kubeflow/kfctl; do you want to take it over?

jlewi:
Replaced by kubeflow/kfctl#172

Issue #4196, opened 2019-09-27
kubeflow/kfctl breakage


kkasravi:
/kind bug

**What steps did you take and what happened:**
can't build kubeflow/kfctl after syncing with kubeflow/kubeflow/bootstrap

problems:
- bootstrap/pkg/apis/apps/kfdef/v1beta1/*.go imports github/kubeflow/kfctl (circular dependency)
- go.mod imports profile-controller by downloading an older version of kfctl
- pkg/apis/apps/kfdef/testdata is missing a doc.go

kfctl needs to be in lock-step with profile-controller, profile-controller needs to move to v2 so no incidental dependency on kfctl is introduced. 

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`)
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

jlewi:
Why does kfctl need to be in lock step with profile controller?

jlewi:
@kkasravi any update on this? How does this impact our plan to move kfctl to kubeflow/kfctl?

kkasravi:
@jlewi this is fixed.

kkasravi:
no impact - see https://github.com/kubeflow/kfctl/pull/51

PR #4197, opened 2019-09-27
Kfctl breakage

kkasravi:
see #4196

can't build kubeflow/kfctl after syncing with kubeflow/kubeflow/bootstrap

problems:

- bootstrap/pkg/apis/apps/kfdef/v1beta1/*.go imports github/kubeflow/kfctl (circular dependency)
- go.mod imports profile-controller by downloading an older version of kfctl
- pkg/apis/apps/kfdef/testdata is missing a doc.go
- kfctl needs to be in lock-step with profile-controller, profile-controller needs to move to v2 so no incidental dependency on kfctl is introduced.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4197)


kunmingg:
Can we remove v2 from kfctl side?
I guess we don't have dependency version conflict since we already deprecate ksonnet?

richardsliu:
/lgtm
/approve
We should open a follow up issue to remove v2 as @kunmingg mentioned.

PR #4201, opened 2019-09-27
Fix test_endpoint.

jlewi:
* We need to use self.workflow.

* This was causing postsubmits to fail because self.workflow is undefined.

* Add a presubmit test that runs with test_endpoint=true in the event the
  E2E workflow is modified.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4201)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -50,12 +50,26 @@ workflows:
       - presubmit
     include_dirs:
       - bootstrap/*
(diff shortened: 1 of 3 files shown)


jlewi:
Test failures were likely due to hitting SSL certificate quota limits. That should be cleaned up.
/test all

Issue #3703, opened 2019-07-22
KfDef specs should use KfDef v1beta1


jlewi:
The KfDef specs stored here:
https://github.com/kubeflow/kubeflow/tree/master/bootstrap/config

Should be updated to use KfDef.Spec.Applications.
Support for KfDef.Spec.Applications should already be available in v0.6.0

Follow on to #3491

0.7 configs should be updated to use KfDef v1beta1.

jlewi:
GCP configs are updated.

I think Arrikto config will be updated by #3904.

@Jeffwan Do you want to update the AWS configs?

Jeffwan:
Just see this issue. I will make corresponding change.

richardsliu:
Is this done?

jlewi:
@richardsliu no; kfctl doesn't support v1beta1 yet and the KFDef YAML files still need to be updated.
https://github.com/kubeflow/manifests/blob/18a8882d1ae3b4fdae2d1d34f80005948a7566b7/kfdef/kfctl_gcp_iap.yaml#L2

@gabrielwen How far will you get with this this week?

@gabrielwen will be OOO next week so maybe someone else can help pick up remaining work next week?

gabrielwen:
so here we could use v1beta1 in kfctl:
https://github.com/kubeflow/kubeflow/blob/6a0fd60a8bfec01ebff6ff4d1a2760320b399374/bootstrap/pkg/apis/apps/kfdef/kfloader.go#L197

I just mentioned this to @swiftdiaries so he will use this function during `build/apply`.

therefore other remaining work would be migrate v1alpha1 to v1beta1.  I'll write a simple script to do this (basically use the function I mentioned above for transformation).  I might need help for verification after this point.

gabrielwen:
TODOS:
- [ ] Migrate all codes using v1alpha1 to v1beta1.
- [ ] A simple CLI to transform KfDef v1alpha1 to v1beta1.
- [ ] Update all KfDef config files on master to v1beta1.

jlewi:
@lluunn any update on this?

gabrielwen:
@lluunn is updating gcp to use `KfConfig`

@Jeffwan could you also do the same thing for AWS?  starting at this point we should change to KfConfig:
https://github.com/kubeflow/kubeflow/blob/2395c5fc0d384cf6dd4d7167bbe9010cd1315b54/bootstrap/pkg/kfapp/coordinator/coordinator.go#L119

[...]

PR #4210, opened 2019-09-28
KfConfig - Internal data structure for holding config info

gabrielwen:
Adding `KfConfig` as internal data structure for holding configs info.  We will start separating data structures as API (KfDef) and data structure as internal implementations (KfConfig):
- adds a new data structure KfConfig for internal implementations.
- adds converters for KfDef both v1alpha1 and v1beta1.
- removes most functions in v1beta1 as we won’t be operating on it.

Relates to #3703

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4210)


yanniszark:
Is this an implementation similar to the Hub-Spokes model for conversion webhooks? (https://book.kubebuilder.io/multiversion-tutorial/conversion.html).
It's the first time I'm seeing such an approach.
Would it make more sense to operate on v1beta1 and convert v1aplha1 to it via a hub-spokes model?
Will KFConfig become an aggregation of all the different versions?
/cc @jlewi

jlewi:
/lgtm
/approve

@yanniszark When implementing APIs its common practice to use an internal datastructure that is different from your external API for precisely the reasons we are doing in this PR.

The goal of using an internal data structure is not to have to update all the code locations. So if we simply use v1beta1 that doesn't avoid that problem; i.e. we would still have to update all locations e.g. to KFDef.Spec.Platform because that field is not in v1beta1. 

By using an internal datastructure KFConfig and converting v1beta1 to KFConfig; we can set KFConfig.Platform to a suitable value based on the v1beta1 spec (e.g. by looking at the plugin) so that we don't have to update all the code locations at once.

Overtime we can migrate the code to use the new fields in KFConfig (e.g. plugins). Once all the code has been updated we can simply delete field KFConfig.Spec.Platform. Furthermore, if we forgot to fix some references we will get a compile time error.

PR #4139, opened 2019-09-17
[3772][notebook-controller] Reissue pod and sts events as notebook events

fediazgon:
This PR is a partial fix of https://github.com/kubeflow/kubeflow/issues/3772. The notebook-controller creates notebook events by propagating both StatefulSet and Pod events. The UI should fetch these events to determine the status of the notebook server (see complete discussion in the linked issue).

How `kubectl -n kubeflow get events --sort-by=.metadata.creationTimestamp` looks like for different scenarios.

**A. Happy path (pod can be scheduled)**

A.1 Notebook Server creation
```text
LAST SEEN   TYPE     REASON                  OBJECT                                          MESSAGE
7s          Normal   ProvisioningSucceeded   persistentvolumeclaim/workspace-everything-ok   Successfully provisioned volume pvc-7701080a-f089-485b-beef-e62b610bee81
7s          Normal   Provisioning            persistentvolumeclaim/workspace-everything-ok   External provisioner is provisioning volume for claim "kubeflow/workspace-everything-ok"
7s          Normal   SuccessfulCreate        statefulset/everything-ok                       create Pod everything-ok-0 in StatefulSet everything-ok successful
7s          Normal   ExternalProvisioning    persistentvolumeclaim/workspace-everything-ok   waiting for a volume to be created, either by external provisioner "k8s.io/minikube-hostpath" or manually created by system administrator
6s          Normal   Pulled                  pod/everything-ok-0                             Container image "gcr.io/kubeflow-images-public/tensorflow-1.13.1-notebook-cpu:v0.5.0" already present on machine
6s          Normal   Created                 pod/everything-ok-0                             Created container everything-ok
6s          Normal   Scheduled               pod/everything-ok-0                             Successfully assigned kubeflow/everything-ok-0 to minikube
5s          Normal   Created                 notebook/everything-ok                          Reissued from pod/everything-ok-0: Created container everything-ok
5s          Normal   Pulled                  notebook/everything-ok                          Reissued from pod/everything-ok-0: Container image "gcr.io/kubeflow-images-public/tensorflow-1.13.1-notebook-cpu:v0.5.0" already present on machine
5s          Normal   Started                 notebook/everything-ok                          Reissued from pod/everything-ok-0: Started container everything-ok
5s          Normal   SuccessfulCreate        notebook/everything-ok                          Reissued from statefulset/everything-ok: create Pod everything-ok-0 in StatefulSet everything-ok successful
[...]

Issue #3772, opened 2019-07-29
Notebook Controller should surface StatefulSet status in the Notebook CR if Pod is not created


kimwnasptd:
The Notebook Controller sets the `status.containerState` in the Notebook CR only when the Pod is created. So, if there would be a problem with a missing Service Account or an unbound PVC, then the Pod will not be scheduled and thus the status in the Notebook CR will not be updated to provide the relative info.

The Controller should set the status in the Notebook CR even when the Pod is not scheduled. In this case, the Notebook Controller will need to query the [Events](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.12/#event-v1-core) API and select the ones with
```yaml
involvedObject:
    apiVersion: apps/v1
    kind: StatefulSet
    name: <Notebook Name>
    namespace: <Notebook Namespace>
```

From the selected events, it should then only keep the latest one. From this event we need the `.message` field.

Lastly, in order for the jupyter-web-app to show a spinner and the status, the Notebook Controller will need to set the `status.containerState` in the CR to `waiting` and the `waiting.reason` to be the StatefulSet's event message.

/cc @lluunn

fediazgon:
Is this still an issue. Can I work on it?

jlewi:
@fediazgon Yes this is still an issue; it would be great if you could pick it up.

jlewi:
@fediazgon Do you plan on picking this up? If yes do you have an ETA for when you might have an initial PR that people can take a look at?

fediazgon:
@jlewi Yes! First time diving in k8s code and I'm derailing a lot because I'm trying to understand the surrounding code. I'll try to have a PR before Saturday.

fediazgon:
Hey @jlewi @kimwnasptd. Before start coding, I would like some advice to see if the following approach is valid.

Events related to a notebook are named like `notebookName.XXXXXXX`, this means that, if we watch for single Events, we can only capture the event at the beginning of the reconcile call, since later (e.g., after we fetch the notebook instance) we don't know the name of the event. This problem does not exist with pods because we know pods are named like `notebookName-0`. However, for events, the `XXXXXXX` is randomly generated. One possible implementation is:

1) Split the event's name by ".", get the first substring (i.e., "notebookName"). Try to fetch the notebook by using key `types.NamespacedName{Name: substring, Namespace: event.Namespace}`. The previous might fail because the event could be unrelated to a notebook. If a notebook was retrieved, update its state if necessary. 

[...]

PR #4214, opened 2019-09-30
[3772][jupyter-web-app] Surface notebook events as extra information

fediazgon:
Partial fix of https://github.com/kubeflow/kubeflow/issues/3772. Depends on https://github.com/kubeflow/kubeflow/pull/4139 (must be merged first).

https://github.com/kubeflow/kubeflow/pull/4139 reissues pod and sts events as notebook events. This is useful since some information is not present as sts or pod status, for example:

```
2s          Warning   FailedCreate            statefulset/test-nb-server                                       create Pod test-nb-server-0 in StatefulSet test-nb-server failed error: pods "test-nb-server-0" is forbidden: error looking up service account kubeflow/default-editor: serviceaccount "default-editor" not found
```

In the previous scenario, the sts status would only show `createdReplicas = 0`, which may very well be that the sts is waiting for the pod to start.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4214)

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/api.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/api.py
@@ -98,6 +98,19 @@ def list_notebooks(namespace):
     )
 
 
+@auth.needs_authorization("list", "", "v1", "events")
+def list_notebook_events(namespace, nb_name):
+    '''
+    V1EventList with events whose source the Notebook with 'nb_name' from namespace 'namespace'
+    '''
+    return wrap_resp(
+        "notebook-events",
+        v1_core.list_namespaced_event,
+        namespace=namespace,
+        field_selector="involvedObject.kind=Notebook,involvedObject.name=" + nb_name
+    )
+
+
 @auth.needs_authorization("list", "kubeflow.org", "v1alpha1", "poddefaults")
 def list_poddefaults(namespace):
     return wrap_resp(
@@ -110,7 +123,7 @@ def list_poddefaults(namespace):
     )
 
 
-@auth.needs_authorization("get", "core", "v1", "secrets")
+@auth.needs_authorization("get", "", "v1", "secrets")
 def get_secret(name, namespace):
     return wrap_resp(
         "secret",
@@ -120,7 +133,7 @@ def get_secret(name, namespace):
     )
 
 
-@auth.needs_authorization("list", "core", "v1", "namespaces")
+@auth.needs_authorization("list", "", "v1", "namespaces")
 def list_namespaces():
     return wrap_resp(
         "namespaces",
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/base_app.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/base_app.py
@@ -1,3 +1,5 @@
+import datetime as dt
+
 from flask import jsonify, request, Blueprint
 from kubernetes import client
 from . import api
@@ -24,8 +26,21 @@ def get_notebooks(namespace):
     if not data["success"]:
         return jsonify(data)
 
-    data["notebooks"] = [utils.process_resource(nb)
-                         for nb in data["notebooks"]["items"]]
+    items = []
+    for nb in data["notebooks"]["items"]:
+        nb_name = nb["metadata"]["name"]
+        nb_creation_time = dt.datetime.strptime(
+            nb["metadata"]["creationTimestamp"], "%Y-%m-%dT%H:%M:%SZ")
+        nb_events = api.list_notebook_events(namespace, nb_name)
+        if not nb_events["success"]:
+            return jsonify(nb_events)
+        # User can delete and then create a nb server with the same name
+        # Make sure previous events are not taken into account
+        nb_events = filter(lambda e: utils.event_timestamp(e) >= nb_creation_time,
+                           nb_events["notebook-events"].items)
+        items.append(utils.process_resource(nb, nb_events))
+
+    data["notebooks"] = items
     return jsonify(data)
 
 
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -1,8 +1,10 @@
+import datetime
 import logging
 import os
 import sys
 import yaml
 import json
+from collections import defaultdict
 from flask import request
 from kubernetes import client
 import datetime as dt
@@ -17,6 +19,14 @@ CONFIGS = [
 # The values of the headers to look for the User info
 USER_HEADER = os.getenv("USERID_HEADER", "X-Goog-Authenticated-User-Email")
 USER_PREFIX = os.getenv("USERID_PREFIX", "accounts.google.com:")
+
+EVENT_TYPE_NORMAL = "Normal"
+EVENT_TYPE_WARNING = "Warning"
+
+STATUS_ERROR = "error"
+STATUS_WARNING = "warning"
+STATUS_RUNNING = "running"
+STATUS_WAITING = "waiting"
 
 
 # Logging
@@ -249,10 +259,10 @@ def process_pvc(rsrc):
     return res
 
 
-def process_resource(rsrc):
+def process_resource(rsrc, rsrc_events):
     # VAR: change this function according to the main resource
     cntr = rsrc["spec"]["template"]["spec"]["containers"][0]
-    status, reason = process_status(rsrc)
+    status, reason = process_status(rsrc, rsrc_events)
 
     res = {
         "name": rsrc["metadata"]["name"],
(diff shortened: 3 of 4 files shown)


jlewi:
Thanks @fediazgon 

Could you please update the PR description to explain in more detail what this PR is doing? e.g. what status is being added to the notebook? How is that status information is determined?

As mentioned in #3772 I'm still unsure whether we want to go down the road of turning events into status as opposed to just surfacing events. Your answers to the above questions will make it easier to evaluate.

fediazgon:
@kimwnasptd @prodonjs Can someone take a look at it?

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
```diff
@@ -269,37 +279,57 @@ def process_resource(rsrc):
...
     if "deletionTimestamp" in rsrc["metadata"]:
-        return "waiting", "Deleting Notebook Server"
+        return STATUS_WAITING, "Deleting Notebook Server"
 
     # Check the status
     try:
         s = rsrc["status"]["containerState"]
     except KeyError:
-        return "waiting", "No Status available"
+        s = ""
```
kimwnasptd: Could you add a small comment explaining that if no `containerState` is available, then we provide information regarding the Notebook's state from the Events in the if clause bellow?

components/jupyter-web-app/backend/kubeflow_jupyter/common/base_app.py
```diff
@@ -25,8 +27,20 @@ def get_notebooks(namespace):
...
+    for nb in data["notebooks"]["items"]:
+        nb_name = nb["metadata"]["name"]
+        nb_creation_time = dt.datetime.strptime(
+            nb["metadata"]["creationTimestamp"], "%Y-%m-%dT%H:%M:%SZ")
+        nb_events = api.get_notebook_events(namespace, nb_name)
+        nb_events = nb_events["notebook-events"].items if nb_events["success"] else []
+        # User can delete and then create a nb server with the same name
+        # Make sure previous events are not taken into account
+        nb_events = filter(lambda e: utils.event_timestamp(e) >= nb_creation_time,
+                           nb_events)
```
kimwnasptd: Doesn't Kubernetes delete the Notebook's Events once the Notebook instance gets deleted?

fediazgon: No. You can try to create a notebook and then delete it and you will see that the pods and statefulsets events are not deleted either. However, if a new controller starts and tries to watch events from the pod or the ss, it won't see events since they are associated with a resource that no longer exists. But if you ask the master-API they are still there.

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
fediazgon: Done

components/jupyter-web-app/backend/kubeflow_jupyter/common/base_app.py
kimwnasptd: I see, it makes sense then to filter the events and show only the ones that should be for each CR's current life-cycle and not from previous ones (based on the timestamps of the CR and the Events). ACK

components/jupyter-web-app/backend/kubeflow_jupyter/common/base_app.py
```diff
@@ -25,8 +27,20 @@ def get_notebooks(namespace):
...
         return jsonify(data)
 
-    data["notebooks"] = [utils.process_resource(nb)
-                         for nb in data["notebooks"]["items"]]
+    items = []
+    for nb in data["notebooks"]["items"]:
+        nb_name = nb["metadata"]["name"]
+        nb_creation_time = dt.datetime.strptime(
+            nb["metadata"]["creationTimestamp"], "%Y-%m-%dT%H:%M:%SZ")
+        nb_events = api.get_notebook_events(namespace, nb_name)
```
kimwnasptd: Since an API call might fail could you add a simple if statement to catch if something wrong happened and return it to the frontend? This way we won't silently ignore errors that might occur.

[Here](https://github.com/kubeflow/kubeflow/blob/master/components/jupyter-web-app/backend/kubeflow_jupyter/default/app.py#L38) is an example where we handle possible api errors. Ideally I would prefer to work with Python exceptions and (probably) a set of generic decorators to handle these errors in order to not repeatedly write if statements. But until I write down an implementation for that lets just use if statements.

kimwnasptd:
@fediazgon I'm trying to build the notebook-controller but I get an error saying that
```bash
module github.com/kubeflow/kubeflow@latest found (v0.7.0), but does not contain package github.com/kubeflow/kubeflow/components/common/reconcilehelper
```

I also tried to run the code inside my GOPATH but it didn't make a difference. Do you have an idea of what I'm missing?

Also, as a side note, I can [see](https://console.cloud.google.com/gcr/images/kubeflow-images-public/GLOBAL/notebook-controller?gcrImageListsize=30) that the notebook-controller image hasn't been updated/rebuild since Oct 7. cc @lluunn

lluunn:
> @fediazgon I'm trying to build the notebook-controller but I get an error saying that
> 
> ```shell
> module github.com/kubeflow/kubeflow@latest found (v0.7.0), but does not contain package github.com/kubeflow/kubeflow/components/common/reconcilehelper
> ```
> 
> I also tried to run the code inside my GOPATH but it didn't make a difference. Do you have an idea of what I'm missing?
> 
> Also, as a side note, I can [see](https://console.cloud.google.com/gcr/images/kubeflow-images-public/GLOBAL/notebook-controller?gcrImageListsize=30) that the notebook-controller image hasn't been updated/rebuild since Oct 7. cc @lluunn

I just tried and got similar error: `build github.com/kubeflow/kubeflow/components/notebook-controller: cannot load github.com/kubeflow/kubeflow/components/common/reconcilehelper: cannot find module providing package github.com/kubeflow/kubeflow/components/common/reconcilehelper`
Investigating..

For updating the image, the last major change I did was notebook beta, which is in September: https://github.com/kubeflow/manifests/pull/364. I think we do need to update image now.

lluunn:
https://github.com/kubeflow/kubeflow/issues/4582

fediazgon:
> > @fediazgon I'm trying to build the notebook-controller but I get an error saying that
> > ```shell
> > module github.com/kubeflow/kubeflow@latest found (v0.7.0), but does not contain package github.com/kubeflow/kubeflow/components/common/reconcilehelper
> > ```
> > 
> > 
> > I also tried to run the code inside my GOPATH but it didn't make a difference. Do you have an idea of what I'm missing?
> > Also, as a side note, I can [see](https://console.cloud.google.com/gcr/images/kubeflow-images-public/GLOBAL/notebook-controller?gcrImageListsize=30) that the notebook-controller image hasn't been updated/rebuild since Oct 7. cc @lluunn
> 
> I just tried and got similar error: `build github.com/kubeflow/kubeflow/components/notebook-controller: cannot load github.com/kubeflow/kubeflow/components/common/reconcilehelper: cannot find module providing package github.com/kubeflow/kubeflow/components/common/reconcilehelper`
> Investigating..
> 
> For updating the image, the last major change I did was notebook beta, which is in September: [kubeflow/manifests#364](https://github.com/kubeflow/manifests/pull/364). I think we do need to update image now.

I got the same error, but I assumed that it was normal since that change (adding `reconcileHelper`) is not yet in `kubeflow@latest`, which I think is pulled by default if no version is specified in `go.mod`.

You can run this command to pull the master version when building (from _components/notebook-controller_):

```go mod edit -require github.com/kubeflow/kubeflow@master```

**Comment by lluunn** [comment]
@fediazgon reconcileHelper is checked in 2 month ago, do you why it's not in kubeflow@latest?

With `go mod edit -require github.com/kubeflow/kubeflow@master`, I got a different error:
`controllers/notebook_controller.go:476:45: k8s.io/apimachinery/pkg/types.NamespacedName composite literal uses unkeyed fields`

**Comment by fediazgon** [comment]
> @fediazgon reconcileHelper is checked in 2 month ago, do you why it's not in kubeflow@latest?

No idea. I can investigate. I thought changes were cherry-picked from the master branch to the `v0.X branch`.

> With `go mod edit -require github.com/kubeflow/kubeflow@master`, I got a different error:
> `controllers/notebook_controller.go:476:45: k8s.io/apimachinery/pkg/types.NamespacedName composite literal uses unkeyed fields`

That looks like an error from `go vet` (that's on me). I don't know why I cannot reproduce it in my development environment.

**Comment by lluunn** [comment]
> That looks like an error from go vet (that's on me). I don't know why I cannot reproduce it in my development environment.

I think this depends on go vet setting. https://stackoverflow.com/questions/54548441/composite-literal-uses-unkeyed-fields. I will send out a quick fix.

**Comment by lluunn** [comment]
https://github.com/kubeflow/kubeflow/pull/4583

**Comment by jlewi** [comment]
@fediazgon and @kimwnasptd any update on this? Think we could get this merged this week?

**Comment by fediazgon** [comment]
@kimwnasptd I addressed all your comments and changed `list_notebook_events` to use the new auth mechanism.

File: components/jupyter-web-app/backend/kubeflow_jupyter/common/api.py [hunk 4]
```diff
@@ -98,6 +98,19 @@ def list_notebooks(namespace):
     )
 
 
+@auth.needs_authorization("list", "core", "v1", "events")
```
kimwnasptd: @fediazgon it will need to be:
`@auth.needs_authorization("list", "", "v1", "events")` (without the core)
just like with pvcs. When written with `core` it will fail the authorization checks. 

I realized that I've written `"core"` instead of `""` in some other api calls like the `list_namespaces` and `get_secret`. Could you also update these ones as well in a follow up commit?

fediazgon: Done

kimwnasptd:
I tried the changes locally and they work as expected. Thanks for the great work @fediazgon 
/lgtm
/approve

Issue #3945, opened 2019-08-19
Istio-Gateway for created Notebook Servers is not configurable


crauterb:
/kind bug

**What steps did you take and what happened:**

I deployed v0.6.1. of Kubeflow in an existing Kubernetes Cluster with an existing Istio-Setup.
After adapting all _virtual-service_ definitions, I got stuck on deploying new notebooks: I was not able to configure the created virtual service for the created notebook server.

The default version that is deployed is:
```yaml
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: notebook-kubeflow-my-jupyter
spec:
  gateways:
  - kubeflow/kubeflow-gateway
  hosts:
  - '*'
  http:
  - match:
    - uri:
        prefix: /notebook/kubeflow/my-jupyter
    rewrite:
      uri: /notebook/kubeflow/my-jupyter
    route:
    - destination:
        host: my-jupyter.kubeflow.svc.cluster.local
        port:
          number: 80
    timeout: 300s
```
... and if I am not mistaken, this is hard-coded in the [notebook-controller.go](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/pkg/controller/notebook/notebook_controller.go#L431)

**What did you expect to happen:**

I expected to be able to configure the gateway using a `.yaml` file or environment variable.

**Anything else you would like to add:**

I generated the `.yaml` configuration using [this documentation](https://www.kubeflow.org/docs/started/k8s/kfctl-existing-arrikto/). 

**Environment:**

- Kubeflow version: : v.0.6.1
- kfctl version: : v0.6.1-rc.2-1-g3a37cbc6
- Kubernetes platform: on-premise Kubernetes cluster
- Kubernetes client version: (use `kubectl version`): v1.15.2
- Kubernetes server version: (use `kubectl version`): v1.11.3
- OS: Ubuntu

yanniszark:
Seems like a limitation of the notebook controller.
Possible solution:
1. Add option in notebook controller to define the Istio Gateway.
2. Expose that option as a parameter in kubeflow/manifests.
3. Set that parameter to your desired value in the KfDef config file.

```yaml
applications:
- name: notebook-controller
  kustomizeConfig:
    overlays:
    - istio
    - application
    parameters:
    - name: istio-gateway
      value: mygateway
    repoRef:
      name: manifests
      path: jupyter/notebook-controller
```

/cc @kunmingg

[...]

PR #4216, opened 2019-10-01
[3945] Configurable Istio Gateway for Notebook Controller

jeremievallee:
Fix for #3945.

`notebook-controller` checks if `ISTIO_GATEWAY` environment variable is set. If it is, it will use its value as gateway. If empty, it will default to `kubeflow/kubeflow-gateway`, which was the previously hardcoded value.

I will also make a PR on `kubeflow/manifests` in order to add the Kustomize config for this environment variable.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4216)

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -367,7 +367,12 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
 	if err := unstructured.SetNestedStringSlice(vsvc.Object, []string{"*"}, "spec", "hosts"); err != nil {
 		return nil, fmt.Errorf("Set .spec.hosts error: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


jeremievallee:
@krishnadurai the virtual service for a notebook is created by the Notebook Controller and by default it uses the gateway `kubeflow/kubeflow-gateway`. But it should be able to use another gateway if needed.

As an example, in our install of Kubeflow, our gateway is in another namespace, which means when we create a notebook its virtual service is not set up properly, because refering to a gateway that doesn't exist.

I'm not too sure how #4167 solves that problem, as it would be annoying for users to have to manually edit their virtual services. Am I missing something here?

krishnadurai:
@jeremievallee thanks.
I understand now that you would want the kubeflow-gateway in a different namespace and want the virtual services be added to it by default.

krishnadurai:
@jeremievallee could you try rebasing this on master once?

jeremievallee:
@krishnadurai done :)

PR #4220, opened 2019-10-02
Centraldashboard: Updates Stackdriver metrics and links for GKE deployments

### LINKED ISSUE #3535: [CentralDashboard] Correct behavior for Stackdriver links
Stackdriver monitoring for GKE clusters supports two different options (https://cloud.google.com/monitoring/kubernetes-engine/). Our current implementation is based on the legacy option which will continue to allow us to retrieve metrics, but currently will link to a page in the Stackdriver UI that is not relevant for clusters using the new option for monitoring.

We should determine if it's possible to know the type of monitoring in use, and if so, build the correct link to the most relevant Stackdriver page.

/area centraldashboard
/area front-end
/priority p2
/assign @prodonjs

prodonjs:
Fixes #3535

- Uses new kubernetes.io metrics https://cloud.google.com/monitoring/api/metrics_kubernetes
- Metrics charts now link the new Kubernetes metrics dashboard in Stackdriver.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4220)

### DIFF
--- a/components/centraldashboard/app/stackdriver_metrics_service_test.ts
+++ b/components/centraldashboard/app/stackdriver_metrics_service_test.ts
@@ -1,4 +1,5 @@
 import * as monitoring from '@google-cloud/monitoring';
+import * as fetch from 'node-fetch';
 
 import {Interval} from './metrics_service';
 import {StackdriverMetricsService} from './stackdriver_metrics_service';
@@ -17,6 +18,7 @@ describe('StackdriverMetricsService', () => {
   };
   let stackdriverClient: jasmine.SpyObj<monitoring.MetricServiceClient>;
   let service: StackdriverMetricsService;
+  let mockFetch: jasmine.Spy;
 
   beforeEach(() => {
     jasmine.clock().install();
@@ -25,6 +27,12 @@ describe('StackdriverMetricsService', () => {
         'mockStackdriverClient', ['listTimeSeries', 'projectPath']);
     stackdriverClient.projectPath.withArgs(projectId).and.returnValue(
         `projects/${projectId}`);
+    const response = {
+      ok: true,
+      text: () => Promise.resolve('test-cluster'),
+    } as fetch.Response;
+    mockFetch =
+        spyOn(fetch, 'default').and.returnValue(Promise.resolve(response));
     service =
         new StackdriverMetricsService(stackdriverClient, projectId, nodeNames);
   });
@@ -38,7 +46,7 @@ describe('StackdriverMetricsService', () => {
       const response =
           ([
             {
-              metric: {labels: {instance_name: 'node-1-gce-instance-abc'}},
+              resource: {labels: {node_name: 'node-1-gce-instance-abc'}},
               points: [
                 {
                   interval: {
@@ -61,7 +69,7 @@ describe('StackdriverMetricsService', () => {
               ]
             },
             {
-              metric: {labels: {instance_name: 'node-2-gce-instance-xyz'}},
+              resource: {labels: {node_name: 'node-2-gce-instance-xyz'}},
               points: [
                 {
                   interval: {
@@ -89,8 +97,8 @@ describe('StackdriverMetricsService', () => {
           .withArgs({
             name: 'projects/test-project',
             filter:
-                'metric.type="compute.googleapis.com/instance/cpu/utilization" ' +
-                'AND metric.label.instance_name = one_of("node-1", "node-2")',
+                'metric.type="kubernetes.io/node/cpu/allocatable_utilization" ' +
+                'AND resource.label.node_name=one_of("node-1", "node-2")',
             interval: {
               startTime: {seconds: 1557705300},
               endTime: {seconds: 1557705600}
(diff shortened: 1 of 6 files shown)


prodonjs:
/unassign @abhi-g 
/unassign @swiftdiaries 
/assign @avdaredevil 

@avdaredevil - Here is the change I was mentioning. You can see it in https://iap-c2d.endpoints.prodonjs-kubeflow-dev.cloud.goog/ or you can try it in your own cluster using containerImage *gcr.io/prodonjs-kubeflow-dev/centraldashboard:pr4220*.

The only thing this doesn't do is provide the `f.rlabel.resource.labels.cluster_name` query filter to the Stackdriver dashboard, as that would require more work to plumb the cluster-name from the backend to the frontend. One option would be to extract it from the hostname, but I'm not sure it's useful because the user still has to [expand](https://screenshot.googleplex.com/3CvLtUuLDdm.png) the cluster resources anyways.

components/centraldashboard/app/stackdriver_metrics_service_test.ts
```diff
@@ -266,6 +275,26 @@ describe('StackdriverMetricsService', () => {
...
+      // Call again to validate that cluster-name is cached
+      await service.getPodCpuUtilization(Interval.Last5m);
+      expect(mockFetch).toHaveBeenCalledWith(
+          'http://metadata.google.internal/computeMetadata/v1/instance/attributes/cluster-name',
+          {headers: {'Metadata-Flavor': 'Google'}});
+      expect(mockFetch).toHaveBeenCalledTimes(1);
+    });
+
+    it('Filters without cluster name', async () => {
+      const response = {ok: false} as unknown as fetch.Response;
```
avdaredevil: Why do you need to double cast this object?

components/centraldashboard/app/stackdriver_metrics_service_test.ts
```diff
@@ -266,6 +275,26 @@ describe('StackdriverMetricsService', () => {
...
+      mockFetch.and.returnValue(Promise.resolve(response));
+      await service.getPodCpuUtilization(Interval.Last5m);
+      expect(stackdriverClient.listTimeSeries).toHaveBeenCalledWith({
+        name: 'projects/test-project',
+        filter: 'metric.type="kubernetes.io/container/cpu/limit_utilization"',
+        interval:
+            {startTime: {seconds: 1557705300}, endTime: {seconds: 1557705600}},
+        aggregation
```
avdaredevil:
```suggestion
        aggregation,
```

components/centraldashboard/app/stackdriver_metrics_service_test.ts
```diff
@@ -332,9 +361,9 @@ describe('StackdriverMetricsService', () => {
...
             name: 'projects/test-project',
-            filter: 'metric.type="container.googleapis.com/container/' +
-                'memory/bytes_used" AND ' +
-                'metric.label.memory_type = "non-evictable"',
+            filter: 'metric.type="kubernetes.io/container/memory/used_bytes"' +
+                ' AND metric.label.memory_type = "non-evictable"' +
+                ' AND resource.label.cluster_name="test-cluster"',
             interval: {
               startTime: {seconds: 1557705300},
               endTime: {seconds: 1557705600}
```
avdaredevil:
```suggestion
              endTime: {seconds: 1557705600},
```

components/centraldashboard/tsconfig.json
```diff
@@ -1,25 +1,19 @@
...
-        ]
-    },
-    "include": [
-        "app",
-        "types"
-    ],
-    "exclude": [
-        "app/**/*_test.ts",
-        "app/**/test_resources.ts",
+  "compilerOptions": {
```
avdaredevil: Can we please keep the tabbing at 4 spaces?

prodonjs: Sorry, I think my editor is inheriting settings from g3 default which uses 2. I just wanted to fix the trailing comma here which is invalid JSON.

components/centraldashboard/app/stackdriver_metrics_service_test.ts
prodonjs: Normally, TS compiler doesn't like that I am not including the other mandatory properties of the Response interface. I think because we aren't explicitly compiling the test sources we're OK so I have just listed it as fetch.Response to simplify.

prodonjs:
@avdaredevil - I think the image build is working as intended but presubmits are failing across the board. As far as your trailing comments go, I was going through to start adding them but I noticed a number of other places. We should probably add that as a TSLint or ESLint setting and provide instructions on ensuring that they get applied during file saving.

components/centraldashboard/tsconfig.json
avdaredevil: I see

components/centraldashboard/app/stackdriver_metrics_service_test.ts
avdaredevil: I see, thanks!

PR #4235, opened 2019-10-04
Enable finalizer to call RevokePlugin() during profile deletion

kunmingg:
Since profile v1beta1, we support adding custom plugins as:
```
type Plugin interface {
	// Called when profile CR is created / updated
	ApplyPlugin(*ProfileReconciler, *profilev1beta1.Profile) error
	// Called when profile CR is being deleted
	RevokePlugin(*ProfileReconciler, *profilev1beta1.Profile) error
}
```
This PR enables finalizer to execute `RevokePlugin`

- `RevokePlugin` should undo all changes outside k8s cluster made by `ApplyPlugin`.
- For example `RevokePlugin` of `workload_identity` will undo bindings to GCP service account.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4235)

### DIFF
--- a/components/profile-controller/controllers/plugin_workload_identity.go
+++ b/components/profile-controller/controllers/plugin_workload_identity.go
@@ -33,6 +33,7 @@ const WORKLOAD_IDENTITY = "WorkloadIdentity"
 
 const GCP_ANNOTATION_KEY = "iam.gke.io/gcp-service-account"
 const GCP_SA_SUFFIX = ".iam.gserviceaccount.com"
+const WORKLOAD_IDENTITY_ROLE = "roles/iam.workloadIdentityUser"
 
 // GcpWorkloadIdentity: plugin that setup GKE workload identity (credentials for GCP API) for target profile namespace.
 type GcpWorkloadIdentity struct {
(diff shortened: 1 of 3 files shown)


components/profile-controller/controllers/plugin_workload_identity.go
```diff
@@ -123,7 +120,30 @@ func (gcp *GcpWorkloadIdentity) setupWorkloadIdentity(namespace string, ksa stri
 	return err
 }
 
-// RevokePlugin: do nothing today.
+// addBinding add binding for <member, WORKLOAD_IDENTITY_ROLE>
+func addBinding(currentPolicy *iam.Policy, member string) {
```
jlewi: Could/should we add unittests for any of these functions?

kunmingg:
Test error is 
```
couldn't delete KfApp:  (kubeflow.error): Code 500 with message: coordinator Delete failed for gcp:  (kubeflow.error): Code 400 with message: Error when cleaning IAM policy: googleapi: Error 409: There were concu
rrent policy changes. Please retry the whole read-modify-write with exponential backoff., aborted
```
We might want to add retry on this step during delete KfApp.

Issue #3496, opened 2019-06-19
kfctl E2E test deploying on existing kubernetes cluster; simulated on prem cluster


jlewi:
We need an E2E test for deploying on an existing Kubernetes cluster.

There's a couple ways we could do this

1. We could use minikf
1. We could use GKE to simulate an existing kubernetes cluster

jlewi:
Might be good to use a tool like kops or kubeadm to setup a conformant K8s cluster on GCP and use that as our test bed.

jlewi:
Lets use this issue to track an E2E test for 
https://github.com/kubeflow/kubeflow/blob/master/bootstrap/config/kfctl_k8s_istio.yaml

krishnadurai:
@jlewi Sure we'll set up tests with the above config

nrchakradhar:
@krishnadurai 
I remember a PR from @swiftdiaries for using KinD (Kubernetes in Docker) for kubeflow e2e test. It could be a first step towards this testing. Its now possible to have KinD run in Kubernetes also as per [seldon comment](https://github.com/kubernetes-sigs/kind/issues/303#issuecomment-546705931)

Do we consider this or are there better choices?

krishnadurai:
@nrchakradhar the PR which you are mentioning is #4237. 

@swiftdiaries could you please provide the context around KinD and how we can go on to introduce it?

jlewi:
Wouldn't using a tool like

* [kops](https://github.com/kubernetes/kops)
* [kubeadm](https://kubernetes.io/docs/reference/setup-tools/kubeadm/kubeadm/)
* [kubespray](https://github.com/kubernetes-sigs/kubespray)

Be a better tool than KinD for simulating and testing an on prem cluster? IIUC KinD is targeting a local (single machine) environment. 

With tools like kops I would expect we can use Cloud VM's to setup a simulated on prem deployment.

swiftdiaries:
KinD is specifically for CI environments and ephemeral (so it doesn't add to frequent resource quota out  of bound errors that break Prow). The problem with using `gcloud` or `kops` or `kubeadm` is that it creates resources external to the workflow. Istio uses `KinD` a lot and `kubeadm` uses `KinD` to test `kubeadm` 
I am working on a GitHub action to test `kfctl` to remove the dependency with Prow and enable developers to speed up their dev cycles by using their own PR branch for iterating.
https://github.com/swiftdiaries/kind-kfctl-github-actions/
[...]

PR #4237, opened 2019-10-06
[WIP] Adds a presubmit kfctl_k8s_istio E2E test

swiftdiaries:
Addresses: https://github.com/kubeflow/kubeflow/issues/3496

KinD is a lightweight Kubernetes distribution that is meant for
testing environments and local development. We use this primarily for
testing the `kfctl_k8s_istio` config and the related components in Kubeflow CI.
Like KinD, this is NOT meant for production workloads.
Please refer to MiniKF for a production-ready environment :)

This PR adds KinD as a platform and a E2E test for the kfctl_k8s_istio.yaml KfDef.

Needs PR: https://github.com/kubeflow/manifests/pull/440

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4237)


bootstrap/pkg/kfapp/kind/kind.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Create a context for the cluster and validate it
+	ctx := cluster.NewContext(clusterName)
+	KubeconfigPath := ctx.KubeConfigPath()
+	log.Infof("Creating KinD cluster with the kubeconfig placed at: %v", KubeconfigPath)
+
+	kindManifestsDir := kind.KfDef.Status.ReposCache["manifests"].LocalPath + "/kind"
+	kindConfigFilePath := kindManifestsDir + "/kind-config.yaml"
+	// KinD Node Image version
+	imageName := "kindest/node:v1.15.0@sha256:b4d092fd2b507843dd096fe6c85d06a27a0cbd740a0b32a880fe61aba24bb478"
```
nrchakradhar: Question: Is it better to take this also from kind-config.yaml?

bootstrap/pkg/apis/apps/group.go
```diff
@@ -153,6 +151,7 @@ const (
 	GCP              = "gcp"
 	MINIKUBE         = "minikube"
 	EXISTING_ARRIKTO = "existing_arrikto"
+	Kind             = "kind"
```
nrchakradhar: Nit: In LHS, should it be KinD (D capital)?

nrchakradhar:
@swiftdiaries Just for my own understanding, is the pre-submit failure due to missing "plugin" section in [kfctl_k8s_istio_kind.yaml](https://raw.githubusercontent.com/swiftdiaries/manifests/kfctl_k8s_e2e/kfdef/kfctl_k8s_istio_kind.yaml)
  plugins:
    - name: kind
      spec:
...

bootstrap/pkg/kfapp/kind/kind.go
swiftdiaries: Thanks for the review @nrchakradhar, yes it will be. Added.

swiftdiaries:
I forgot to add the plugin in the KfDef, added now.

nrchakradhar:
[KinD issue #864](https://github.com/kubernetes-sigs/kind/issues/864#issuecomment-534885348) may help
how can increase the log level ?

--loglevel string logrus log level [panic, fatal, error, warning, info, debug, trace] (default "warning")

Should we keep log level to debug for some time?

jlewi:
Why are we defining a new Kubernetes in docker plugin in order to test kfctl_k8s_istio E2E test?

The point of kfctl_k8s_istio.yaml is that we can install on an existing cluster; so why can't our E2E test use existing tooling to create the cluster?

So here's my expectation for how we would do E2E testing for kfctl_k8s_istio.

1. We modify [kfctl_e2e_workflow.py](https://github.com/kubeflow/kubeflow/blob/master/py/kubeflow/kubeflow/ci/kfctl_e2e_workflow.py) to create an Argo workflow with suitable steps for creating the cluster
    * We could either subclass that workflow or else just add arguments and if statements

1. We would add a step to create the appropriate K8s cluster using the tooling

    * So we could use kops, Kind, gcloud
    * Whatever makese sense for the test

1. We run kfctl apply `kfctl_k8s_istio.yaml` to install Kubeflow on that existing cluster

BenTheElder:
👋  I would really consider executing the kind CLI, the library is still rough and the interfaces are not final, we're working to improve that but it falls lower priority that improving the behavior of the tool 😅

swiftdiaries:
Thanks for the tip @BenTheElder :) I'm wwitching to using the CLI and I'm in the process of removing the plugin that used the library

swiftdiaries:
Looks like `KinD` isn't able to get the Docker daemon from inside Argo.
Seeing this error:

```
util.py                     45 INFO     Running: /mnt/test-data-volume/kubeflow-presubmit-kfctl-go-k8s-4237-53d884a-7664-92b0/src/kubeflow/kubeflow/testing/kfctl/kubernetes-in-docker crea
te cluster --name=kubeflow_kind --config=https://raw.githubusercontent.com/swiftdiaries/manifests/kfctl_k8s_e2e/kind/kind-config.yaml --image=kindest/node:v1.15.0@sha256:b4d092fd2b507843d
d096fe6c85d06a27a0cbd740a0b32a880fe61aba24bb478 --loglevel=debug
cwd=None
util.py                     60 INFO     Subprocess output:
util.py                     71 INFO     time="21:17:08" level=debug msg="Running: /usr/bin/docker [docker ps -q -a --no-trunc --filter label=io.k8s.sigs.kind.cluster --format {{.Names}}\\
t{{.Label \"io.k8s.sigs.kind.cluster\"}}]"
util.py                     71 INFO     Error: could not list clusters: failed to list nodes: exit status 1
```

swiftdiaries:
Should be able to use [this](https://github.com/kubernetes-sigs/kind/issues/303#issuecomment-518593664) to deploy KinD from a Kubernetes Pod since Argo runs each step as a Pod.

Issue #4020, opened 2019-08-28
CD for jupyter notebook controller


jlewi:
/kind feature

**Why you need this feature:**
We'd like to continually build the jupyter notebook controller from the latest commits and update the  kustomize manifest to use the new image so that master is updated in a timely fashion.

We have some scripts/tooling for jupyter-web app here
https://github.com/kubeflow/kubeflow/tree/master/releasing/auto-update

This probably needs improvement

jlewi:
Current status:

* @kkasravi Defined some initial tekton workflows in #4238
* kubeflow/testing#424 is tracking installing Tekton on our test clusters so we can run these workflows

I think the next critical step is to get Tekton installed on our releasing cluster to begin running these workflow

kkasravi:
@jlewi I updated tektoncd-install to 0.8.0 ([here](https://github.com/kkasravi/kfctl-config-files/tree/master/tektoncd-install)) which allows a Pipeline to be embedded in PipelineRun. This allows all parameters to be defined in PipelineRun however it's dependent on removing builtin.NewLegacyOrderTransformerPlugin from kfctl (so resources are honored in the ordered listed)

kkasravi:
Possible game plan
1. Get the kfctl fix in (noted above)
2. Update tekton-install in manifests to 0.8.0
3. Update the CI for components to use 0.8.0 (embedded Pipeline in PipelineRun)
4. kubeflow/testing#424

lluunn:
cc

jlewi:
@kkasravi  where does kfctl come into play?

* There's a Tekton PipelineRun defined [here](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/ci/pipeline-run.yaml) to do 

  * Instructions for running it are [here](https://github.com/kubeflow/kubeflow/tree/master/components/base)

So I would think we would need to do the following

1. Setup our releasing cluster to run these pipelines

    * Install tekton in our release cluster kubeflow/testing#424
    * Create [required secrets](https://github.com/kubeflow/kubeflow/tree/master/components/base) in the cluster

1. Install the [tekton pipeline templates](https://github.com/kubeflow/kubeflow/tree/master/components/base) in the cluster

1. Do a manual run(s) of the tekton pipeline to verify it works. 

1. Setup a solution to auto-trigger tekton CD workflows on post-submit and periodically

[...]

PR #4238, opened 2019-10-06
Ci for components

kkasravi:
this does ci/cd for the following components
- admission-webhook
- centraldashboard
- jupyter-web-app
- notebook-controller
- profile-controller

using tekton pipelines. The follow use case is enabled:

1. A PR is merged into kubeflow/kubeflow updating the component
1. The merged commit is 1234
1. This tekton pipelinerun is triggered to build the component image from commit @1234.
1. The pipelinerun edits manifests/common/centraldashboard/base/kustomization.yaml (using kubeflow-bot repo) and adds the new image tag
1. The pipelinerun calls `make generate; make test`
1. If successful then
1.   The pipeline checks in the changes
1.   Opens a PR with the updated kubeflow/manifests that uses the newly built image
1.   Approvers LGTM the PR to kubeflow/manifests and it gets merged

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4238)


jlewi:
@kkasravi How are we going to trigger these pipelines?
Do we already have this pipeline setup and working for one application?

jlewi:
Is the base kustomize package really a base package?

IIUC the base package is defining Tekton Tasks and Pipelines. These resources are reused using Tekton not using kustomize.

For each application we are using kustomize to define a PipelineRun that will reference the Tekton resources defined in the base package but not using overlays.

jlewi:
Would it be better to put the kustomize package defining the CI/CD pipeline into the directory releasing?

Longer term should we move the definition of that pipeline into kubeflow/testing?

swiftdiaries:
And optionally, do we have a way to turn on/off the CI? 
For times when we have to update/fix?
Not sure if this is a major concern but when there's a drift, it'd nice to be able to turn it off -> patch in fixes and turn it on again. 
And +1 to the move to `kubeflow/testing`

jlewi:
/hold
  * Were there changes you wanted to get in before commiting it?

/lgtm
/approve

jlewi:
#4020 is the issue tracking this work.

Issue #4240, opened 2019-10-07
kfctl build and apply do not appear to detect the project variable in KFDef


holdenk:
/kind bug

**What steps did you take and what happened:**
Download kfctl_gcp_iap.yaml , replace the email key with my e-mail username, add project: <PROJECT NAME> as the last key as indicated in the "Project should be set to the GCP project you want to use." comment. Run kfctl apply. Receive error project is not configured.

**What did you expect to happen:**

Either applying of kubeflow, or at least an error indicating where to configure the project.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): N/A (did not get that far)
- kfctl version: (use `kfctl version`): 2395c5fc
- Kubernetes platform: (e.g. `minikube`): GCP
- Kubernetes version: (use `kubectl version`): N/A (did not get that far)
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04.3

jlewi:
@holdenk can you provide your KFDef File?

jlewi:
@gabrielwen @lluunn  Are we not attempting to get the default values from gcloud?

lluunn:
dup of this one: https://github.com/kubeflow/kubeflow/issues/3923

jlewi:
Duplicate of #3923

jlewi:
See #4246 
The problem appears to be different from #3923 
In particular, kfctl doesn't appear to respect the email, project, and zone set in the KFDef.

The E2E test is setting the EMAIL and PROJECT environment variable which isn't what it should be doing
https://github.com/kubeflow/kubeflow/blob/9592b48c95537ceec4c9aada83b73b7a738e48a8/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py#L211

This was changed in #4040 which defined the new "-f" flag.

/assign @swiftdiaries

jlewi:
Here's where its getting PROJECT,ZONE, EMAIL from environment variables
https://github.com/kubeflow/kubeflow/blob/9592b48c95537ceec4c9aada83b73b7a738e48a8/bootstrap/pkg/kfapp/gcp/gcp.go#L1964

jlewi:
This was added recently in #4040

PR #4246, opened 2019-10-07
Clarify that we pick up PROJECT, EMAIL, and ZONE from environment variables.

holdenk:
Closes https://github.com/kubeflow/kubeflow/issues/4240

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4246)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -1965,21 +1965,21 @@ func (gcp *Gcp) Generate(resources kftypesv3.ResourceEnum) error {
 	if gcp.kfDef.Spec.Project == "" {
 		return &kfapis.KfError{
 			Code:    int(kfapis.INVALID_ARGUMENT),
-			Message: "Project not specified.",
+			Message: "Project not specified. Set PROJECT environment variable.",
 		}
 	}
 	gcp.kfDef.Spec.Zone = os.Getenv("EMAIL")
 	if gcp.kfDef.Spec.Email == "" {
 		return &kfapis.KfError{
 			Code:    int(kfapis.INVALID_ARGUMENT),
-			Message: "email not specified.",
+			Message: "email not specified. Set EMAIL environment variable.",
 		}
 	}
 	gcp.kfDef.Spec.Zone = os.Getenv("ZONE")
 	if gcp.kfDef.Spec.Zone == "" {
 		return &kfapis.KfError{
 			Code:    int(kfapis.INVALID_ARGUMENT),
-			Message: "zone not specified.",
+			Message: "zone not specified. Set ZONE environment variable.",
 		}
 	}
 	// Set default IPName and Hostname


swiftdiaries:
Thanks for the quick fix @holdenk !
/lgtm

jlewi:
Is this correct? Shouldn't we be using gcloud to get the default values?

jlewi:
This should be where we are getting email
https://github.com/kubeflow/kubeflow/blob/9592b48c95537ceec4c9aada83b73b7a738e48a8/bootstrap/pkg/kfapp/gcp/gcp.go#L230

swiftdiaries:
I went off of the docs here: https://www.kubeflow.org/docs/gke/deploy/deploy-cli/
I'm not sure if that is updated

swiftdiaries:
For some reason, the shelled out commands don't work for me when I run `kfctl`. 
So as a work around, I set the environment variables. 
Would it be better to check here if the Spec has a project/email set and if it's empty then pick it up from the environment variable?

holdenk:
So we ignore the spec currently (I tried setting project and e-mail in my input). I think longer term reading the input spec and figuring out default values from the gcloud command sound lovely, but for right now the error message has no path to resolution for the user.

swiftdiaries:
Would something like this work?
Since the `gcloud config get-value account` is run at the beginning and if the `KfDef` is edited manually, at this point we should have these set.
In the event they aren't, we can set them from the environment variables.
Regarding raising an error, Apply should throw the appropriate error if something's not right. 

```
	if gcp.kfDef.Spec.Project == "" {
	    gcp.kfDef.Spec.Project = os.Getenv("PROJECT")
	}
	
	if gcp.kfDef.Spec.Email == "" {
            gcp.kfDef.Spec.Email = os.Getenv("EMAIL")
	}

	if gcp.kfDef.Spec.Zone == "" {
            gcp.kfDef.Spec.Zone = os.Getenv("ZONE")
	}
```

jlewi:
@holdenk The resolution should be to set project explicitly in your KFDef file. That's what the E2E test is doing so I'd be suprised if setting project, email, and zone explicitly in your KFDef file doesn't work.
https://github.com/kubeflow/kubeflow/blob/9592b48c95537ceec4c9aada83b73b7a738e48a8/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py#L148

Can you provide the steps to reproduce?

swiftdiaries:
I modified the E2E test to pick it up from the environment variable here:
https://github.com/kubeflow/kubeflow/blob/9592b48c95537ceec4c9aada83b73b7a738e48a8/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py#L211
So I guess it ignores the one in the KFDef file.

jlewi:
@swiftdiaries that's a bug. We should not be reading PROJECT and ZONE from environment variables.

Where is the logic to fetch those values from environment variables? 


/assign @gabrielwen

holdenk:
That makes sense then @swiftdiaries 
The repro steps are `mkdir gcp07b && cd gcp07b && kfctl build --file=../gcp07b.yaml` with https://pastebin.com/fbSfaRpz
I can update this to check if the spec values are empty and change the error message to say set env var or in the input spec.

Does that sound good @jlewi & @swiftdiaries ?

jlewi:
@swiftdiaries @holdenk I'm still trying to figure out why using environment variables work. To my knowledge getting PROJECT and ZONE from environment variables was nothing we ever supported and I can't find code explicitly getting those values from environment variables.

My conjecture is that there might be some form of viper magic going on where options can automatically be set via command line or environment variable.

In the past the supported way of setting project and zone were

1. Setting it in KFDef
1. Setting explicitly via command line option

Using environment variables doesn't really make sense because its redundant with gcloud config. So i don't think we want to update the error message to start telling people to use environment variable.

If the code isn't respecting the value in KFDef then that's a huge bug that we need to fix.

swiftdiaries:
Removing the error here should fix it.
While I was testing locally, the shelled out commands from `gcp.go` didn't execute.
I'll file a separate bug for this.

jlewi:
@swiftdiaries the code should not be calling os.Getenv to get PROJECT, ZONE, or email. Those lines are also causing the bug since they are not checking if PROJECT, zone and email are already set.

swiftdiaries:
I'll get a fix out in a PR and link it here.

jlewi:
Thanks @swiftdiaries lets update the test as well please to not set the environment variables.

swiftdiaries:
Got it !

swiftdiaries:
Before we call `gcp.generateDMConfigs()` here:
https://github.com/kubeflow/kubeflow/blob/9592b48c95537ceec4c9aada83b73b7a738e48a8/bootstrap/pkg/kfapp/gcp/gcp.go#L1998

We need to have these set in the Spec. If they are empty, should we surface that out with a `log.Warnf()`?

swiftdiaries:
PR: https://github.com/kubeflow/kubeflow/pull/4252

gabrielwen:
> @swiftdiaries @holdenk I'm still trying to figure out why using environment variables work. To my knowledge getting PROJECT and ZONE from environment variables was nothing we ever supported and I can't find code explicitly getting those values from environment variables.
> 
> My conjecture is that there might be some form of viper magic going on where options can automatically be set via command line or environment variable.
> 
> In the past the supported way of setting project and zone were
> 
> 1. Setting it in KFDef
> 2. Setting explicitly via command line option
> 
> Using environment variables doesn't really make sense because its redundant with gcloud config. So i don't think we want to update the error message to start telling people to use environment variable.
> 
> If the code isn't respecting the value in KFDef then that's a huge bug that we need to fix.

@jlewi 
using env is supported since we started implementing kfctl go binary.  it is derived from the shell script version and back at that time we wanted to keep the same semantics.

holdenk:
Should I close this then since https://github.com/kubeflow/kubeflow/pull/4252 ?

swiftdiaries:
@holdenk sorry about that, could you please close?

PR #4251, opened 2019-10-07
kf-monitoring as monitoring lib

kunmingg:
Add `kf-monitoring` as monitoring metrics library.
The goal is to define a loose structure for all kubeflow metrics so that metrics consuming services can process those metrics based on common structure.

- fixed common metrics names and types.
- fixed common label names: `component, kind, namespace, action, severity`

1. Upstream components can label metrics by labels values.
2. Alert rules are grouped by `metrics name (type) + label action + label severity`
3. Dashboard UI can show metrics filtered by label combinations.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4251)


components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+
+func NewMetricsExporter(component string) *MetricsExporter {
+	return &MetricsExporter{
+		component: component,
+		ErrChan: make(chan error),
+		metricsPort: METRICSPORT,
+	}
+}
+
+func (me *MetricsExporter) SetPort(port int) *MetricsExporter {
```
zhenghuiwang: If port can be changed, why not simply export the field as `Port`?

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+	labels := prometheus.Labels{COMPONENT: me.component, KIND: kind, NAMESPACE: namespace, ACTION: action}
+	requestGauge.With(labels).Add(val)
+}
+
+func (me *MetricsExporter) ObserveRequestExponentialLatency(kind string, namespace string, action string, val float64) {
+	labels := prometheus.Labels{COMPONENT: me.component, KIND: kind, NAMESPACE: namespace, ACTION: action}
+	requestExponentialLatency.With(labels).Observe(val)
+}
+
+func (me *MetricsExporter) serveMetrics() {
```
zhenghuiwang: `serveMetrics()` should follow the same pattern as `server.Serve()` to directly return an error and block the calling thread if no error occurs.

In general, calling a func to do a task in a single thread or multi threads is the caller's decision. The function like `serveMetrics` or `server.Serve()` should not decide it for the caller even if it is blocking. Because in some case, the caller simply wants to be blocked on this call.

For callers which don't, it is straightforward to add an error channel and start a Goroutine. So the `Start()` and `ErrChan` add little value but bloat the API.

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/log"
+	"net"
+	"net/http"
+)
+
+// Common label keys for all metrics signals
+// COMPONENT: name of component outputing the metrics, eg. "tf-operator"
+const COMPONENT = "component"
```
zhenghuiwang: For readability, can you group these constants? Seems there are 3 groups.

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+const SEVERITY_MAJOR = "Major"
+const SEVERITY_CRITICAL = "Critical"
+
+const METRICSPORT = 8079
+const METRICSPATH = "/metrics"
+
+var (
+	// Counter metrics
+	// num of requests counter vec
+	requestCounter = prometheus.NewCounterVec(
```
zhenghuiwang: Nit: It would be more readable, if these vars are only defined in `var (...)`with zero-values, but assigned with values in `init()` right before it is used.

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+func (me *MetricsExporter) SetPort(port int) *MetricsExporter {
+	me.metricsPort = port
+	return me
+}
+
+func (me *MetricsExporter) Start() {
+	go me.serveMetrics()
+}
+
+func (me *MetricsExporter) AddRequestCounter(kind string, namespace string, action string) {
```
zhenghuiwang: Nit: `AddRequestCounter` -> `IncRequestCounter`

kunmingg:
@zhenghuiwang 
Thanks!
Now `ErrChan` is removed and `ServeMetrics()` is blocking.
Port is accessible for callers so can be reset if necessary.

components/kf-monitoring/README.md
```diff
@@ -0,0 +1,22 @@
...
+  - Two major downstream consumers:
+    - [alert rules](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/) that keep scanning metrics
+    - [monitoring dashboard](https://prometheus.io/docs/visualization/grafana/) that display metrics groups
+- structured label group
+  - Each component use labels to identify the metrics they are recording.
+  - Current label group: `component, kind, namespace, action, severity`
+  
+**Integration instruction**
+
+WIP
```
zhenghuiwang: For easy integration, is it possible to integrate this with Istio for services with sensible defaults? (This may overlap with Istio dashboard) e.g.

component -> Istio virtual service name
kind -> "endpointStatus"
action -> Get/Post/... + endpointPath
severity -> 4XX(major) / 5XX(critical)

I think it is hard for some component to change its source code to use this metrics server. For example, Metadata uses MLMD gRPC docker image out of box. Even if some component can, it would be valuable if the external behavior is captured by default and developers only need to capture the internal status.

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/common/log"
+	"net"
+	"net/http"
+)
+
+// Common label keys for all metrics signals
+// >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
+// COMPONENT: name of Component outputing the metrics, eg. "tf-operator"
+const COMPONENT = "Component"
```
zhenghuiwang: Nit: I was thinking group by "()" instead of ">>>"
```
const (
     COMPONENT = ""
     KIND = ""
     ...
)
```

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+	prometheus.MustRegister(requestCounter)
+	prometheus.MustRegister(requestFailureCounter)
+	prometheus.MustRegister(requestGauge)
+	prometheus.MustRegister(requestFailureGauge)
+	prometheus.MustRegister(requestLinearLatency)
+	prometheus.MustRegister(requestExponentialLatency)
+	prometheus.MustRegister(serviceHeartbeat)
+}
+
+type MetricsExporter struct {
```
zhenghuiwang: A public struct (and method) should have a comment like "// MetricsExporter ...".
Can you run `go vet ./...`?

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+	)
+	// Counter metrics for failed requests
+	requestFailureCounter = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name: "request_failure_counter",
+		Help: "Number of request_failure_counter",
+	}, []string{COMPONENT, KIND, NAMESPACE, ACTION, SEVERITY})
+
+	// Gauge metrics
+	requestGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "requests_gauge",
```
yeya24: Requests should not be a gauge value. From my thought, requests are only appropriate for counter.

components/kf-monitoring/MetricsExporter.go
```diff
@@ -0,0 +1,161 @@
...
+		Help: "Number of requests_gauge",
+	}, []string{COMPONENT, KIND, NAMESPACE, ACTION, SEVERITY})
+
+	// Gauge metrics for failed requests
+	requestFailureGauge = prometheus.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "requests_failure_gauge",
```
yeya24: Ditto. A counter is better

components/kf-monitoring/README.md
```diff
@@ -0,0 +1,22 @@
+# kf-monitoring
+
+kf-monitoring is common monitoring util library for Kubeflow components written in go.
```
yeya24: is a common monitoring

components/kf-monitoring/README.md
```diff
@@ -0,0 +1,22 @@
...
+The goal is to provide interface for both metrics generators (kubeflow components) and metrics consumters (alerts / UI).
```
yeya24: s/consumters/consumers/g

jlewi:
@kunmingg what are the next steps here?

Issue #2633, opened 2019-03-06
Build kubeflow docker images on power


dreamryx:
We have test running latest mnist example with kubeflow 4.0 on Power 8 (ppc64le architecture). During the test, we have built a lot of kubeflow docker images on power and want to contribute the dockerfiles. We supposed to have two options:

 - Due to a lot of changes of dockerfiles is just replace the base image, we can follow the way in k8s - create temp docker file and replcace the base image variable to build images for different arch. For example, https://github.ibm.com/platformcomputing/kubernetes/tree/master/cluster/images/hyperkube. To make makefile look nice, this way better have a fixed registry contains images for different platforms with consistent naming style.
 
 - Add new dockerfiles with suffix "-ppc64le". Then each dockerfile can be used to built image for specific platform.

gyliu513:
@jlewi any comments for this? Seems there are also some discussions at https://github.com/kubeflow/kubeflow/issues/925

FYI @hougangliu @jinchihe

jlewi:
IIUC your are proposing modifying all Kubeflow related Docker images in order to build versions for power. 

So this is a pretty big change that touches pretty much all of Kubeflow. If so it would be great to put together a doc documenting this in more detail and discuss it in the community meeting or perhaps the developer summit if your there.

Can/Should we localize all power related changes in a single location

e.g. could we have  a repo kubeflow/power that would contain dockerfiles and other tooling needed to build for power? Can we use automation to keep this in sync with the underlying repos so for example if the TFJob Dockerfile is updated in kubeflow/tf-operator this change is automatically propogated to kubeflow/power

gyliu513:
Thanks @jlewi , your comments really helps!!

Compared with `create a new repo kubeflow/power`, we prefer the big change. As in the future, we may also want to support [s390x](https://en.wikipedia.org/wiki/Linux_on_z_Systems) as well. Creating a repo for each platform maybe difficult to manage.

So we will try to create a design document first and try to review it in community meeting after ready for review.

Besides, we also have another issue https://github.com/kubeflow/kubeflow/issues/2226 which want to unify the kubeflow image build and location. As currently, all of the base images in kubeflow are already multi-arch docker images, so it is pretty simple to enable those images to support POWER, we already have some PRs as follows:
- https://github.com/kubeflow/tf-operator/pull/893/files This is trying to update the hardcode `amd64` to a variable so that it can be configured on different platforms.
[...]

PR #4257, opened 2019-10-09
add the centraldashboard Dockerfile for ppc64le

alongzhi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4257)

### DIFF
--- a/components/centraldashboard/Dockerfile.ppc64le
+++ b/components/centraldashboard/Dockerfile.ppc64le
@@ -1,0 +1,42 @@
+# Step 1: Builds and tests
+FROM node:10-stretch AS build
+
+ARG kubeflowversion
+ARG commit
+ENV BUILD_VERSION=$kubeflowversion
+ENV BUILD_COMMIT=$commit
+ENV CHROME_BIN=/usr/bin/chromium-browser
+ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
+
+# Install the Chromium package and configure environment for testing
+RUN echo deb http://us.ports.ubuntu.com/ubuntu-ports/ bionic universe >> /etc/apt/sources.list && \
+    echo deb http://us.ports.ubuntu.com/ubuntu-ports/ bionic main restricted >> /etc/apt/sources.list && \
+    apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3B4FE6ACC0B21F32 && \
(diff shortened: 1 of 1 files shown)


prodonjs:
@alongzhi - How would you anticipate this being used in Kubeflow deployments? Currently, I don't think any other Docker images are built for ppc64le. I just want to make sure I understand what this will be used for since nothing else appears to be referencing it.

jinchihe:
@prodonjs We are working on Kubeflow supporting POWER platform, and some Dockerfile has been updated. See the ticket for details https://github.com/kubeflow/kubeflow/issues/2633 .We now installed Kubeflow on POWER after rebuild some image for POWER platform, so contribute the Dockerfile. Thanks.

alongzhi:
Hi @prodonjs , Thanks for your reviewing. 

I'm building such an image so that centraldashboard can be run on ppc64le platform. In the past several months, we have updated some components/images to support on ppc64le, centraldashboard will be one of the plan. @jinchihe , could you please share more background about this?

jinchihe:
@alongzhi Seems you created new Dockerfile for POWER platform, is there have big diff with one that for x86_64?

alongzhi:
@jinchihe Yes this Dockerfile for Power has a big difference with x86_64, the major reason is that the chromium-browser is not supported from source alpinelinux.org, so we have to find another one. And some other enhancement to make sure npm build/test pass, thanks.

jinchihe:
@alongzhi OK, great.
/lgtm

prodonjs:
Thanks @jinchihe and @alongzhi for the explanations. I'd like to get one of the project leads like @jlewi to weigh in on this as his comments on #2633 seem to indicate some concern about adding additional Dockerfiles to each component.

/cc @jlewi

jlewi:
@prodonjs @alongzhi I think its fine to include non x86 Dockerfiles to support power PC builds.
My expectation is that it would be the folks supporting the powerpc deployments and not the application owners who would be responsible for maintaining these.

Placing a hold just to make sure we are all on the same page but if we are then I think @alongzhi you can remove this and we are good to go.

/approve
/hold

xauthulei:
retest

prodonjs:
@alongzhi - I would recommend rebasing this change over *master* and then pushing it again. Usually that resolves the failing presubmit tests.

PR #4258, opened 2019-10-09
kfam: add param for enabling patch workload identity plugin to all profiles.

kunmingg:
kfam now has new flag `--workload-identity`
Expected value is a GCP service account that will bind to user pods in current k8s cluster.
If above flag is not "", kfam will patch workloadIdentity plugin to all profiles it create.

Cluster admin can still change GCP identity of certain profile by editing the profile CR directly.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4258)


components/access-management/main.go
```diff
@@ -28,22 +28,26 @@ const USERIDHEADER = "userid-header"
 const USERIDPREFIX = "userid-prefix"
 // set cluster admin user id here.
 const CLUSTERADMIN = "cluster-admin"
+// identity used for workload-identity plugin.
+const WORKLOADIDENTITY = "workload-identity"
```
yanniszark: Default value for WORKLOADIDENTITY should be disabled ("").
Having it enabled by default assumes everyone will use GCP by default.

components/access-management/kfam/api_default.go
```diff
@@ -129,6 +135,22 @@ func (c *KfamV1Alpha1Client) CreateProfile(w http.ResponseWriter, r *http.Reques
 	}
 }
 
+func (c *KfamV1Alpha1Client) PatchProfilePlugins(profile *profilev1beta1.Profile) {
```
yanniszark: Where is this function used?

jlewi:
@kunmingg Why is it a default of kfam and not the profile controller? If we do it in the KFAM API and then if a user creates a profile by the CLI it wouldn't be automatically patched?

Could we instead have the profile controller take an optional default value that would get set on all profiles?

components/access-management/main.go
kunmingg: The current default arg value is "", WORKLOADIDENTITY is off by default.

jlewi:
@kunmingg Why do it in two places? Should we just change this PR to add the flag to the profile controller instead of doing it in KFam?

kunmingg:
Sorry that's what I mean...
I'm going to close this one and move logic to controller side.

jlewi:
@kunmingg got it thanks.

Issue #4231, opened 2019-10-04
Add label to all namespaces to support Katib in 0.7


johnugeorge:
/kind feature

In 0.7 release,  all namespaces where katib experiments have to be deployed should have the following label. 
`     katib-metricscollector-injection: enabled
`
This includes Kubeflow + all user namespaces that are created by profile. 

Related:  https://github.com/kubeflow/katib/issues/841

Sample:  https://github.com/kubeflow/katib/blob/master/manifests/v1alpha3/0-namespace.yaml#L6

jlewi:
Is there a reason katib depends on the label on the namespace and not the label on the individual pods?

johnugeorge:
@jlewi I think, it matters in large clusters where there are non kubeflow namespaces.  Since object selectors are introduced only in k8s 1.15(https://github.com/kubernetes/kubernetes/blob/release-1.15/pkg/apis/admissionregistration/types.go#L264) , there will be  performance impact and pod level web hooks can become a bottleneck for pre 1.15 clusters.  We can remove this label when k8s 1.15 becomes the default. 

/cc @gaocegege

gaocegege:
Namespace selector of webhooks can avoid performance issues. Or the webhook will accept many requests from all namespaces.

Label on the individual pods has the right behavior, too. But the performance issue cannot be avoided. We still get many requests from all namespaces.

PR #4262, opened 2019-10-09
Katib namespace metric collector label added to kfctl and profile-controller

krishnadurai:
Resolves: #4231

/cc @johnugeorge @hougangliu @gaocegege

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4262)

### DIFF
--- a/bootstrap/pkg/utils/k8utils.go
+++ b/bootstrap/pkg/utils/k8utils.go
@@ -16,6 +16,7 @@ package utils
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"github.com/ghodss/yaml"
 	gogetter "github.com/hashicorp/go-getter"
@@ -51,8 +52,9 @@ import (
 )
 
 const (
-	YamlSeparator = "(?m)^---[ \t]*$"
-	CertDir       = "/opt/ca"
+	YamlSeparator              = "(?m)^---[ \t]*$"
+	CertDir                    = "/opt/ca"
+	katibMetricsCollectorLabel = "katib-metricscollector-injection"
 )
 
 func generateRandStr(length int) string {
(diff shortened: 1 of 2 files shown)


bootstrap/pkg/utils/k8utils.go
```diff
@@ -314,6 +341,18 @@ func (a *Apply) namespace(namespace string) error {
 					string(kftypes.NAMESPACE), namespace, nsErr),
 			}
 		}
+	} else {
```
hougangliu: Thanks for your PR
for above branch creating a new namespace, we should also make sure the katibMetricsCollectorLabel added. I think you can add the label in line 335

krishnadurai: @hougangliu Thanks, looks like I missed it in the commit.

krishnadurai:
This is ready for review.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -114,8 +116,11 @@ func (r *ProfileReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error)
 		ObjectMeta: metav1.ObjectMeta{
 			Annotations: map[string]string{"owner": instance.Spec.Owner.Name},
 			// inject istio sidecar to all pods in target namespace by default.
-			Labels: map[string]string{"istio-injection": "enabled"},
-			Name:   instance.Name,
+			Labels: map[string]string{
+				istioInjectionLabel:        "enabled",
+				katibMetricsCollectorLabel: "enabled",
```
kunmingg: Seems the label value is always "enabled".
Then shall we just inject katib metrics collector by default to all katib pods?

bootstrap/pkg/utils/k8utils.go
```diff
@@ -314,6 +348,18 @@ func (a *Apply) namespace(namespace string) error {
...
+		if _, ok := namespaceInstance.ObjectMeta.Labels[katibMetricsCollectorLabel]; !ok {
+			patchErr := a.patchNamespaceWithLabel(
+				namespace, katibMetricsCollectorLabel, "enable",
```
hougangliu: s/enable/enabled

components/profile-controller/controllers/profile_controller.go
hougangliu: metrics collector is injected into katib pods as a sidecar container by mutating webhook, for performance consideration, only the pods whose namespace has "katib-metricscollector-injection" label will be handled.

kunmingg: Can we instead have a 
`kubeflow-managed`: `true`
label on all kubeflow namespaces?

I feel other controllers will need similar label when kubeflow got deployed as part of a giant cluster.

hougangliu: I am not sure
katib can work well all alone, even not as a component of kubeflow

kunmingg: @jlewi what's your preference?

Is `katib namespace` == `kubeflow namespace`?
If they are same set, then we can use 1 label for multiple controllers.

gaocegege: Will istio inject the istio sidecar to every pod in the namespace if we set it to enabled? If that, it may affect the behavior of katib. When users use job to run HPO jobs, we assume that containers[0] in the job is the actual training job.

johnugeorge: @kunmingg This label will be removed after k8s 1.15 becomes the default. See related comment https://github.com/kubeflow/kubeflow/issues/4231#issuecomment-539305378

jlewi: I think we need to establish guidelines around how admission hooks should behave.

I can see reasons for both arguments

* Application specific labels make it easy to selectively enable/disable an application web hook on a namespace level

On the other hand we could potentially get an explosion of labels if every application adds its own.

I'd probably lean towards being able to selectively disable/enable katib on a per namespace level. That feels more defensive. 

Katib is still Alpha so I think we can change the behavior in subsequent releases. So I don't have a strong oppinion either way.

bootstrap/pkg/utils/k8utils.go
krishnadurai: Thanks, done.

components/profile-controller/controllers/profile_controller.go
krishnadurai: > I'd probably lean towards being able to selectively disable/enable katib on a per namespace level. That feels more defensive.

I agree this allows for a cluster-admin to switch preferences for each namespace.

kunmingg: @johnugeorge 
1.15 is required for multi version CR as well, looking forward to it

Issue #3923, opened 2019-08-16
kfctl cannot infer project when not set


lluunn:
/kind bug

**What steps did you take and what happened:**

I am using the latest config from master (removing the dummy value of project/email)
```
kfctl init --config=kfctl_gcp_iap.yaml
kfctl generate all
```

The project field is missing, and hostname treat it as empty string
kf0815-2.endpoints..cloud.goog

It can infer the email though.

gcloud config has both email and project set:

```
$ gcloud config list
[compute]
region = us-central1
zone = us-central1-a
[core]
account = lunkai@google.com
disable_usage_reporting = True
project = kai-test2

Your active configuration is: [default]
```

cc @jlewi

jlewi:
It looks like we have logic to get the Default email
https://github.com/kubeflow/kubeflow/blob/d82aa59103940918a9c8e1ca87ba94f24bec09b2/bootstrap/pkg/kfapp/gcp/gcp.go#L229

We are missing similar logic for project and zone.

Bump to P0 because we should try to fix this in 0.7.X.

This should be a relatively straightforward fix.
Without it users can't deploy in one command e.g.

```
kfctl apply -f https:/.../kfctl_gcp_iap.yaml
```

Because they have to modify the config to set the values.

PR #4263, opened 2019-10-10
[kfctl] set default project and zone for GCP

zhenghuiwang:
fixes https://github.com/kubeflow/kubeflow/issues/3923

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4263)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -106,6 +106,8 @@ type Gcp struct {
 	// Function to get the GcpAccount.
 	// Support injection for testing.
 	gcpAccountGetter func() (string, error)
+	gcpProjectGetter func() (string, error)
+	gcpZoneGetter    func() (string, error)
 
 	runGetCredentials bool
 }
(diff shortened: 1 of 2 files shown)


lluunn:
thanks
/lgtm
/approve

swiftdiaries:
Hey @zhenghuiwang 
Could you please test this with the E2E?
I think [this](https://github.com/kubeflow/kubeflow/blob/master/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py#L149) is the part where we set the project in the config
By removing it, we should see the project automatically inferred.

zhenghuiwang:
> Hey @zhenghuiwang
> Could you please test this with the E2E?
> I think [this](https://github.com/kubeflow/kubeflow/blob/master/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py#L149) is the part where we set the project in the config
> By removing it, we should see the project automatically inferred.

@swiftdiaries thanks for the suggestion, which makes a lot of sense to test it e2e if it is easy to do.

Looking at the pointed code, I find it is a util func not a test. Although the func is named `get_config_spec`, it actually sets the project name explicitly. Deleting that line may surprise the func caller when the default project name is not the same as the name passed in.

PR #4275, opened 2019-10-11
Improve kfctl semantics and ability to rerun app.yaml

jlewi:
* Related to kubeflow/kfctl#49

* Delete NewKfapp its not being called from anywhere.

Refactor NewLoadKfAppFromURI and LoadKfAppCfgFile

* We should only call CreateCfgFile and do the directory check if using a
  remote URI

  * For a local path we already have an appfile so we don't need to recreate
    the CreateCfgFile

  * Calling CreateCfgFile for local paths was a major source of friction
    because it ends up checking if the directory is empty.

    * But when invoking kfctl on a local file we don't need to check if
      the directory is empty because we can assume that's the correct
      ${KFAPP}.

* Once we have a local config file we can treat the two cases the same

* Fix error propogation in apply.go; if BuildKfAppFromURI returns an
error we want to return that error so it ends up being printed out.
  Otherwise we just print out kfApp is nil and swallow the error.

* Improve the error message for project, zone, and config not being set
  by providing the appropriate gcloud command.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4275)

PR #4279, opened 2019-10-11
Add error logging to kfctl apply.go - Remove GetKfAppFromCfgFile

zhenghuiwang:
Fixes https://github.com/kubeflow/kfctl/issues/45

Error logging is missing in this code block in `apply.go`
```
if kind == string(kftypes.KFDEF) {	
	kfApp, err = coordinator.BuildKfAppFromURI(configFilePath)	
}
```

It is hard to see why kfApp is nil if `coordinator.BuildKfAppFromURI`  fails.

Rewriting the whole func to reduce the number of `if-else` statements.

* Delete `GetKfAppFromCfgFile` and call `BuildKfAppFromURI` from URI instead.

  * We want to unify the code paths and avoid unnecessary divergence.

  * The code path should be as follows
  * If the "-f" flag isn't set use cwd to guess the config file path (i.e. app.yaml)
      *At this point the code path should be the same whether -f was supplied or -f is missing
  * If -f is a remote URI copy it to CWD; set configFilePath to the local file
     * At this point the code path is the same for remote and local files and configFilePath should 
       just be pointing at a local file path

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4279)


bootstrap/cmd/kfctl/cmd/apply.go
```diff
@@ -72,15 +52,35 @@ var applyCmd = &cobra.Command{
 			if err != nil || kfApp == nil {
 				return fmt.Errorf("error loading kfapp: %v", err)
 			}
+			return kfApp.Apply(kftypes.ALL)
```
jlewi: I think we can get rid of the call to GetKfApplyFromCFGFile and kfApp.Apply here.
In the if block we just need to populate configFilePath because its not suplied via command line argument.

At that point the code path should be the same as when user supplies -f flag.

jlewi: I think we can also delete GetKfAppFromCfgFile and just use BuildKfAppFromURI call below.

jlewi: We will need to update delete.go to call LoadKfAppFromCfgFile as well rather than GetKfApplyFromCFGFile

bootstrap/cmd/kfctl/cmd/delete.go
```diff
@@ -42,11 +44,15 @@ var deleteCmd = &cobra.Command{
 		if err != nil {
 			return fmt.Errorf("cannot fetch current directory for apply: %v", err)
 		}
-		deleteStorage := deleteCfg.GetBool(string(kftypes.DELETE_STORAGE))
-		kfApp, err = coordinator.GetKfAppFromCfgFile(cwd+"/app.yaml", deleteStorage)
+		kfApp, err = coordinator.BuildKfAppFromURI(filepath.Join(cwd, "app.yaml"))
```
richardsliu: This will attempt to recreate the app.yaml file - why is this being changed?

richardsliu: Discussed with @jlewi , this PR needs to be rebased after https://github.com/kubeflow/kubeflow/pull/4275 is merged.

jlewi:
@zhenghuiwang #4275 is merged can you rebase and update this?

bootstrap/cmd/kfctl/cmd/delete.go
jlewi: With #4275  BuildKFAppFromURI should no longer be attempting to recreate the KFDef file. A new config file should only be created in the event we are using a remote URI. But this isn't a remote URI so we should be good.

zhenghuiwang:
/test kubeflow-presubmit


Error log:
```
util.py                     71 INFO     time="2019-10-14T18:13:46Z" level=error msg="Creating kfctl-2a8a error: &{Code:RESOURCE_ERROR Location:/deployments/kfctl-2a8a/resources/kfctl-2
a8a Message:{\"ResourceType\":\"gcp-types/container-v1beta1:projects.locations.clusters\",\"ResourceErrorCode\":\"FAILED_OPERATION\",\"ResourceErrorMessage\":\"The network \\\"default\
\\" does not have available private IP space in 10.0.0.0/8 to reserve a /14 block for containers for cluster {us-central1-a, 29647740582, kubeflow-ci-deployment, kfctl-2a8a, f465accda3
71eac2931a09c12fa4341efab6b2f6e033c4e45be4e8a1f75433ec}. - https://container.googleapis.com/v1beta1/projects/29647740582/zones/us-central1-a/clusters/kfctl-2a8a\"} ForceSendFields:[] N
ullFields:[]}" filename="gcp/gcp.go:380"
```

PR #4286, opened 2019-10-12
Installing the latest KFP SDK package in the notebook image

Ark-kun:
/assign @jlewi @IronPan

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4286)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -6,7 +6,7 @@ ARG BASE_IMAGE=tensorflow/tensorflow:1.13.2-py3-jupyter
 FROM $BASE_IMAGE
 
 ARG TF_SERVING_VERSION=0.0.0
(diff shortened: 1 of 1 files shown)


Ark-kun:
Otherwise KF 0.7 ships with very old SDK version with many known issues.

jlewi:
@Ark-kun @IronPan I thought we wanted to stop installiing the SDK in the notebook images and instead have samples explicitly install the SDK version they need because the SDK was evolving so rapidly?

jlewi:
See kubeflow/kubeflow#3632

Ark-kun:
> @Ark-kun @IronPan I thought we wanted to stop installiing the SDK in the notebook images and instead have samples explicitly install the SDK version they need because the SDK was evolving so rapidly?

This is also an option. I'm OK with that.
Both options are better than having a version from June that's 11 releases behind.

jlewi:
I think its better to just remove it and have notebooks install the specific version they need.

PR #4297, opened 2019-10-14
remove jsonnet unit test

lluunn:
We don't need them anymore.

/cc @gabrielwen @richardsliu

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4297)


jlewi:
@lluunn I don't think all of our tests have migrated off of ksonnet yet. Should we keep this until they do?

lluunn:
ah good point.

PR #4308, opened 2019-10-15
Fix darwin kfctl

swiftdiaries:
Fixes an issue with the kfctl darwin build and release.
We used to copy the linux build as the MacOS one previously.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4308)


swiftdiaries:
/cc @richardsliu 
since this pertains to the 0.7 release

jlewi:
@swiftdiaries I would suggest rebasing; I just merged the PR to reduce the number of presubmits that are triggered so that might help get a green

swiftdiaries:
Looks like I misconstructed the `go build` with OS and ARCH args

lluunn:
I don't think we need the top level go.mod go.sum?
https://github.com/kubeflow/kubeflow/blob/master/go.mod
https://github.com/kubeflow/kubeflow/blob/master/go.sum

swiftdiaries:
I seem to made a mistake in the `Makefile` for the tarball as well.
I'll open a PR fixing it and remove the top level `go.mod` and `go.sum` files

PR #4311, opened 2019-10-15
Change calls to NewLoadKfAppFromURI

richardsliu:
* Per kubeflow/kfctl#55 NewLoadKfAppFromURI is the preferred method for external callers to create
  KFApps

* Also update the code to use KFConfig.

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4311)


bootstrap/pkg/kfupgrade/kfupgrade.go
```diff
@@ -217,13 +229,23 @@ func MergeKfDef(oldKfDef *kfdefsv3.KfDef, newKfDef *kfdefsv3.KfDef) {
 }
 
 func (upgrader *KfUpgrader) Generate() error {
-	kfApp := kustomize.GetKfApp(upgrader.NewKfDef)
+	kfApp, err := coordinator.BuildKfAppFromURI(upgrader.TargetPath)
```
jlewi: I think you want to use NewLoadKfAppFromURI.

BuildKFAppFromURI is calling init and generate if you need those methods just call them yourself on the kfapp returned by NewLoadKfAppFromUri.

jlewi: You are already calling Generate below.

bootstrap/pkg/kfupgrade/kfupgrade.go
```diff
@@ -75,12 +75,12 @@ func createNewKfApp(baseConfig string, version string, oldKfDef *kfdefsv3.KfDef)
 	newKfDef.Spec.Version = version
 
 	// Make sure the new KfApp is created.
-	_, err = coordinator.CreateKfAppCfgFile(newKfDef)
+	configFile, err := coordinator.CreateKfAppCfgFile(newKfDef)
```
jlewi: I don't think you want to call createKfAppCfgFile. CreateKfAppCfgFile should be an internal method used by NewKfApp which determines what to do when the argument to NewKfApp is a remote URI.

The KfUpgrade should determine the appropriate directory to store KfApp and just write the file to that directory.
i.e. by calling converters.WriteConfigFile
https://github.com/kubeflow/kubeflow/blob/9f5ef66ce3d12ff2b603d07330a16661caa0b286/bootstrap/pkg/apis/apps/configconverters/converters.go#L139

bootstrap/pkg/kfupgrade/kfupgrade.go
```diff
@@ -52,7 +52,7 @@ func createNewKfApp(baseConfig string, version string, oldKfDef *kfdefsv3.KfDef)
 	// Load the new KfDef from the base config
 	newKfDef, err := kfdefsv3.LoadKFDefFromURI(baseConfig)
```
jlewi: Use this function
https://github.com/kubeflow/kubeflow/blob/9f5ef66ce3d12ff2b603d07330a16661caa0b286/bootstrap/pkg/apis/apps/configconverters/converters.go#L37 to load a KFConfig rather than a KFDef

bootstrap/pkg/kfupgrade/kfupgrade.go
```diff
@@ -217,13 +229,23 @@ func MergeKfDef(oldKfDef *kfdefsv3.KfDef, newKfDef *kfdefsv3.KfDef) {
...
+		return err
+	}
+
 	return kfApp.Generate(kftypesv3.K8S)
 }
 
 func (upgrader *KfUpgrader) Apply() error {
-	kfApp := kustomize.GetKfApp(upgrader.NewKfDef)
-	err := kfApp.Generate(kftypesv3.K8S)
+	kfApp, err := coordinator.BuildKfAppFromURI(upgrader.TargetPath)
```
jlewi: use NewKfAppFromURI

richardsliu:
@jlewi PTAL.

jlewi:
/lgtm
/approve

But it looks like you have some conflicts and need to rebase.

PR #4313, opened 2019-10-16
Mark kfctl_delete_test as expected to fail

jlewi:
* See kubeflow/kfctl#56 the test is flaky
* More importantly if the test fails we end up blocking the uploading of
  junit files and we lose signal.
* So to prevent that we mark it as expected to fail. This way the step
  won't exit with non-zero exit code and the cleanup will continue.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4313)


jlewi:
Looks like a test flake running the unittests.

```
go: error loading module requirements
Makefile:73: recipe for target 'generate' failed
make: *** [generate] Error 1
```

/test all

PR #4317, opened 2019-10-16
admission-webhook: support imagepullsecrets in poddefaults

### LINKED ISSUE #4315: admission-controller: allow imagePullSecrets in PodDefaults
/kind feature

We need this feature to satisfy the private registry use-case.
In the general use-case where you want to access a private registry from a K8s cluster, you have the following options:
1. Configure Docker credentials on each node. This isn't ideal as it makes images available to all users in all namespaces and it's also Docker-specific.
2. Use platform-specific registries (eg GKE with GCR). This isn't ideal as it is not portable.
3. Use imagePullSecrets. This is the recommended way (in the k8s docs) for the general case.

In addition, this will enable users to pull images from their own registries, using their own imagePullSecrets which reside in their namespace/Profile.

Extend PodDefaults to include ImagePullSecrets.

/cc @zabbasi  @lluunn @kkasravi @jlewi @lluunn @kunmingg

yanniszark:
Resolves #4315 

This PR adds support for imagePullSecrets in PodDefaults.
This is done in the following steps:
1. Extend API and add new function for merging imagePullSecrets.
2. Improve Makefile to include code generation.
3. Generate new Deepcopy methods for API.

/cc @zabbasi @jlewi @kimwnasptd @elikatsis @kkasravi @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4317)


jlewi:
@yanniszark I thought I responded yesterday but maybe I forgot to submit my comment?

IIUC image pullsecrets can be bound to a service account. So if someone wanted to attach image pull secrets to one or more pods; they could just configure those pods to use the appropriate service account.

Is there an advantage to using an admission hook instead of attaching pull secrets to a service account?

yanniszark:
@jlewi you have actually submitted a comment in the respective issue.
Let's continue the discussion there.

jlewi:
@yanniszark per the discussion in #4315 can we go with using imagepullsecrets in service accounts and close this PR?

yanniszark:
@jlewi let's revisit this after the Kubecon.

mvoitko:
@jlewi how do you propose to add `imagePullSecrets` to service accounts? I hope not manually via kubectl commands)

PR #4318, opened 2019-10-16
Update e2e tests

lluunn:
- always use build_and_apply, since the old semantic is not supported 
- kfctl-go-iap-istio is covered by kfctl-gcp-e2e-changed, and kfctl-gcp-e2e-changed is the same as kfctl-go-iap-endpoint, so combine them.
- combine basic_auth pre and post submit, and remove build_deploy since build_and_apply is true

/cc @kunmingg @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4318)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -28,59 +28,32 @@ workflows:
     params:
       installIstio: true
       workflowName: deployapp-istio
+  ### GCP IAP
   - py_func: kubeflow.kubeflow.ci.kfctl_e2e_workflow.create_workflow
     name: kfctl-go-iap-endpoint
     job_types:
+      - presubmit
       - postsubmit
       - periodic
     include_dirs:
       - bootstrap/*
-      - deployment/*
-      - dependencies/*
       - kubeflow/*
       - testing/*
+      - py/kubeflow/kubeflow/ci/kfctl_e2e_workflow.py
     kwargs:
       use_basic_auth: false
       test_endpoint: true
       config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_gcp_iap.yaml
-  - py_func: kubeflow.kubeflow.ci.kfctl_e2e_workflow.create_workflow
-    component: kfctl_go_test
-    name: kfctl-go-iap-istio
-    job_types:
-      - presubmit
-    include_dirs:
-      - bootstrap/*
-      - kubeflow/*
-      - testing/*
-    kwargs:
-      use_basic_auth: false
-      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_gcp_iap.yaml
-  # If the kfctl_e2e_workflow.py was modified then on presubmit we also want to run with test_endpoint true
-  # to ensure that test still pasess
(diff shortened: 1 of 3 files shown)


lluunn:
/hold

waiting for test result

jlewi:
@lluunn What do you mean by using build_and_apply always?

I think the difference is whether we just do

```
kfctl apply -f spec.yaml
```

Or we do

```
kfctl build -f spec.yaml
kfctl apply -f spec.yaml
```

Per kubeflow/kfctl#57 we shouldn't really need to fully deploy with both semantics in order to know that everything is working.

We should probably have more focused tests to test which ever semantics we don't end up adopting.
 
If we want to pick just one for the E2E test I might suggest

```
kfctl build -f spec.yaml
kfctl apply -f spec.yaml
```

This is primarily because I think just running

```
kfctl apply -f spec.yaml 
```

Is more common; i.e. thats what we do in the demo script. So manual testing is more likely to  reveal problems with that particular syntax.

/cc @swiftdiaries

prow_config.yaml
```diff
@@ -28,59 +28,32 @@ workflows:
     params:
       installIstio: true
       workflowName: deployapp-istio
+  ### GCP IAP
   - py_func: kubeflow.kubeflow.ci.kfctl_e2e_workflow.create_workflow
     name: kfctl-go-iap-endpoint
     job_types:
+      - presubmit
```
jlewi: I don't think we want to run this particular test on presubmit.

The difference between this particular test and kfctl_go_test is that this instance is running test_endpoint which verifies that the IAP endpoint is accessible.

I don't think we want to run that on presubmit. Only postsubmit.

An alternative way to skip the test on presubmits would be using pytest annotations but that's not what we do right now.

prow_config.yaml
```diff
@@ -28,59 +28,32 @@ workflows:
...
-    kwargs:
-      use_basic_auth: false
-      test_endpoint: true
-      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_gcp_iap.yaml
-  # Run basic auth test as part of every periodic and postsubmit run. 
+  ### GCP BasicAuth
   - py_func: kubeflow.kubeflow.ci.kfctl_e2e_workflow.create_workflow
     name: kfctl-go-basic-auth
     job_types:
+      - presubmit
```
jlewi: I would actually suggest we just run basic auth in postsubmit.

Per kubeflow/kfctl#57 all the expensive presubmits are slowing us down. Using an E2E test for Kubeflow as a whole to test that kfctl is something we should fix.

At best we should only be triggering this if the gcp/basic auth code changes.

prow_config.yaml
lluunn: we are already running endpoint test in presubmit: https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml#L60

lluunn:
I think I misunderstood the param.
I thought we are still using old semantics.

swiftdiaries:
We switched to the new semantics for the E2E, but we do need to think about [this](https://github.com/kubeflow/kfctl/issues/57)

PR #4320, opened 2019-10-16
Fix KfConfig conversion bugs

gabrielwen:
forgot to copy kind to KfDef in v1beta1.

Related to #3703 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4320)

### DIFF
--- a/bootstrap/pkg/apis/apps/configconverters/v1beta1.go
+++ b/bootstrap/pkg/apis/apps/configconverters/v1beta1.go
@@ -4,12 +4,37 @@ import (
 	"fmt"
 	"github.com/ghodss/yaml"
 	kfapis "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis"
+	kftypesv3 "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis/apps"
 	kfconfig "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis/apps/kfconfig"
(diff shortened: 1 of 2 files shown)


bootstrap/pkg/apis/apps/configconverters/v1beta1.go
```diff
@@ -12,6 +13,30 @@ import (
...
+		string(kfconfig.AWS_PLUGIN_KIND): true,
+		string(kfconfig.GCP_PLUGIN_KIND): true,
+	}
+
+	_, ok := platforms[pluginKind]
+	return ok
+}
+
+func maybeGetPlatform(pluginKind string) string {
+	platforms := map[string]string{
```
lluunn: (minor) Should we just use switch?

bootstrap/pkg/apis/apps/configconverters/v1beta1.go
```diff
@@ -12,6 +13,30 @@ import (
 type V1beta1 struct {
 }
 
+func isPlatform(pluginKind string) bool {
```
lluunn: where is this used?

gabrielwen: removed.

bootstrap/pkg/apis/apps/configconverters/v1beta1.go
gabrielwen: I don't know, which one is a better solution in golang?

PR #4323, opened 2019-10-17
check ep instead of pod for culling notebooks

### LINKED ISSUE #4322: bug when checking if the notebook needs to be culled
/kind bug

In the current culling logic, we will first check whether the notebook pod exists. However, this pod maybe still in pending state so we cannot access its status. 
So in this case, a better solution is to check whether this Service endpoint > 0, this means the pod is ready and accessible from Service. 

I' d like to open a PR for it.
``` Go
	// Check if the Notebook needs to be stopped
	if podFound && culler.NotebookNeedsCulling(instance.ObjectMeta) {
		log.Info(fmt.Sprintf(
			"Notebook %s/%s needs culling. Setting annotations",
			instance.Namespace, instance.Name))

		// Set annotations to the Notebook
		culler.SetStopAnnotation(&instance.ObjectMeta)
		err = r.Update(ctx, instance)
		if err != nil {
			return ctrl.Result{}, err
		}
```

**Environment:**

- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`)
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

yeya24:
Signed-off-by: yeya24 <yb532204897@gmail.com>

fixes #4322

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4323)

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -168,7 +168,6 @@ func (r *NotebookReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	// Check the pod status
 	pod := &corev1.Pod{}
-	podFound := false
 	err = r.Get(ctx, types.NamespacedName{Name: ss.Name + "-0", Namespace: ss.Namespace}, pod)
 	if err != nil && apierrs.IsNotFound(err) {
 		// This should be reconciled by the StatefulSet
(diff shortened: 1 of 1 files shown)


yeya24:
Sorry for this PR. That' s my mistake because the controller checks the Kubernetes api instead of accessing the service. I will close this now

PR #4337, opened 2019-10-18
Drastically reduce the presubmits triggered for kfctl to reduce pain.

jlewi:
* See kubeflow/kfctl#57 kfctl testing is overly reliant on expensive, flaky
  E2E tests; right now we aren't testing kfctl so much as all of Kubeflow.

* This is cauing a lot of pain and negatively impacting velocity of kfctl.

* Right now we are running multiple variants of the same expensive e2e
  tests on presubmits.

  * These additional tests are providing minimal additional coverage on
    presubmit but the flakiness is a major drag on velocity

* Don't run kfctl-gcp-basic-auth on presubmit; this provides minimal
  additional coverage on presubmit

* Delete kfctl-go-build-deploy - the difference between this test
  and the other e2e tests is that it was invoking kfctl build followed by
  kfctl apply rather than just apply.

  * This provides minimal additional coverage
  * Instead just always run the 2 step process.
  * We should use unittests/mocks or some other cheaper solution to
    capture the differences with just running kfctl apply.

* Delete kfctl-gcp-e2e-changed

  * The only difference between this and kfctl-gcp-iap was that it was
    selectively enabling the endpoint test on some presubmits

  * We should just disable test_endpoint on all presubmits
  * the endpoint_test isn't a kfctl test so much as an all of Kubeflow test

  * Modify test_endpoint.py to use pytest annotations to selectively
    skip the test on presubmit.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4337)

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -29,48 +29,23 @@ workflows:
       installIstio: true
       workflowName: deployapp-istio
   - py_func: kubeflow.kubeflow.ci.kfctl_e2e_workflow.create_workflow
-    name: kfctl-go-iap-endpoint
+    name: kfctl-go-iap
     job_types:
+      - presubmit
       - postsubmit
       - periodic
     include_dirs:
       - bootstrap/*
-      - deployment/*
-      - dependencies/*
-      - kubeflow/*
-      - testing/*
-    kwargs:
-      use_basic_auth: false
-      test_endpoint: true
-      config_path: https://raw.githubusercontent.com/kubeflow/manifests/master/kfdef/kfctl_gcp_iap.yaml
-  - py_func: kubeflow.kubeflow.ci.kfctl_e2e_workflow.create_workflow
-    component: kfctl_go_test
-    name: kfctl-go-iap-istio
-    job_types:
-      - presubmit
-    include_dirs:
-      - bootstrap/*
-      - kubeflow/*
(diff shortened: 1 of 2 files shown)


prow_config.yaml
```diff
@@ -29,48 +29,23 @@ workflows:
...
-      - presubmit
-    include_dirs:
-      - bootstrap/*
-      - kubeflow/*
       - testing/*
       - py/kubeflow/kubeflow/ci/kfctl_e2e_workflow.py 
     kwargs:
       use_basic_auth: false
+      # Run build and then apply rather than just apply
+      build_and_apply: true
```
lluunn: Since kfctl apply only is more common (I suppose?), should we set this to false?

nrchakradhar: This flag itself may not be needed after some more refactoring and re-organization of e2e tests, but its not high on priority I suppose.

jlewi: My thinking is that we get more coverage by doing

```
kfctl build ...
kfctl apply
```

then just doing 

```
kfctl apply
```

My expectation is also that we will have more manual testing of just running

```
kfctl apply
```

So I'd expect if we just run `kfctl apply` has a bug we are more likely to know about it.

PR #4341, opened 2019-10-18
[centraldashboard] Clear iframe state when user navigates away.

### LINKED ISSUE #4334: [CentralDashboard v2] iFrame src updates when location hash changes on sub-apps
## About:
In @prodonjs words:
> There's still another problem that's always existed that I just became aware of. This [screencast](https://screencast.googleplex.com/cast/NjI1MTQ4NzEyMDMyNjY1NnwzYjQ0NDkxZC05MA) taken on the image from 0.6.2 demonstrates the problem. Because Pipelines and Metadata relies on hash-based routing, updating the iframe src doesn't actually force the navigation event. So, if I click on the Exit Handler pipeline, then click back, and then click on the Parallel execution pipeline, the page inside of Pipelines doesn't actually update.

> The solution to this is probably to destroy and recreate the iframe whenever the user navigates away from the top-level route that activates the iframe page. I didn't have the time to implement that today but that's probably a useful thing to open an issue for.

### Meta
/area centraldashboard
/area front-end
/assign @prodonjs 
/priority p1
/cc @kwasi

prodonjs:
Fixes #4334

This ensures that embedded apps like Pipelines and Metadata which
use hash-based routing reload when the user navigates away and comes
back.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4341)

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -239,6 +239,10 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         if (hideSidebar || isIframe !== this.inIframe || isIframe) {
             this.$.MainDrawer.close();
         }
+
+        if (!isIframe) {
+            this.iframeSrc = 'about:blank';
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


prodonjs:
/uncc @swiftdiaries 
/uncc @abhi-g 

/cc @avdaredevil 

@avdaredevil - This should address the issue.

- [Screencast](https://screencast.googleplex.com/cast/NTA0NTIzMzMwNzg3NzM3Nnw1NGVmMTMzZi1iNA)
- [Live](https://kubeflow.endpoints.prodonjs-kubeflow-dev.cloud.goog/)

PR #4347, opened 2019-10-18
Import KindOrderTransformer and KindFilterTransformer into kfctl

jbrette:
Those two transformers (that can eventually be loaded into kustomize) allow kfctl to natively decide the order of resources before running "apply" or "delete".

They could also be used to get kfctl to apply in phases, for instance kfctl apply during phase1 would create the CRDs only, kfctl apply during phase2 would apply everything but the CRDs.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4347)

### DIFF
--- a/bootstrap/pkg/kfapp/kustomize/kindfiltertransformer/KindFilterTransformer.go
+++ b/bootstrap/pkg/kfapp/kustomize/kindfiltertransformer/KindFilterTransformer.go
@@ -1,0 +1,92 @@
+// Copyright 2019 The Kubernetes Authors.
+// SPDX-License-Identifier: Apache-2.0
+package kindfiltertransformer
+
+import (
+	"errors"
+
+	"github.com/ghodss/yaml"
+	// "sigs.k8s.io/yaml"
+
+	"sigs.k8s.io/kustomize/v3/pkg/ifc"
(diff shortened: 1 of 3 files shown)


jlewi:
Thanks @jbrette You might want to add yourself to the github org so you can trigger tests.

Can we create unittests for these files?

yanniszark:
@jbrette @jlewi 
I'm a little hesitant about having a Plugin do magic reordering of resources.
If something doesn't work as we want it, then we have to make changes to the Go code instead of reordering resources in the kustomization file.

I'm thinking that maybe it's better to configure kustomize so that it's keeping the declared order (like kubectl) and be explicit on this.
WDYT?

jlewi:
@jbrette @yanniszark @kkasravi how should we proceed?

Note development of kfctl has now migrated to kubeflow/kfctl so if we decide to move forward with this change we should close this PR and open a new one in kubeflow/kfctl

kkasravi:
@jlewi i think we should leverage these transformers to order CRDs, Cluster Resources first which would allow us to collapse all the *-crds targets. We could possibly also look at building dependencies (eg kfserving depends on knative) with these plugins. We can enable/disable plugins as needed using transformers or generators or a directive in KfDef.

PR #4349, opened 2019-10-18
kfctl keeps local config file name

lluunn:
Related https://github.com/kubeflow/kfctl/issues/54

- If the config file is remote, will save it in current directory with its filename.
- If the config file is local, will keep the name (previously will write out another app.yaml)
- force apply and delete to pass -f

This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4349)


bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -282,14 +281,20 @@ func LoadKfAppCfgFile(cfgfile string) (kftypesv3.KfApp, error) {
...
+
+		_, err = CreateKfAppCfgFile(kfdef)
 		if err != nil {
 			return nil, &kfapis.KfError{
 				Code:    int(kfapis.INVALID_ARGUMENT),
 				Message: fmt.Sprintf("Error creating KfApp from config file: %v", err),
 			}
 		}
 	}
+	// ??
```
gabrielwen: remove?

bootstrap/pkg/kfapp/coordinator/coordinator.go
```diff
@@ -300,6 +305,8 @@ func LoadKfAppCfgFile(cfgfile string) (kftypesv3.KfApp, error) {
 		}
 	}
 
+	// Set the ConfigFileName for kfdef. This is used when we WriteToConfig.
+	kfdef.Spec.ConfigFileName = filepath.Base(appFile)
```
gabrielwen: I'm not sure I understand why we need a new field `ConfigFileName`?  just like what you do here, we could always separate the values we need with `Base` and `Dir`, no?

lluunn: We need to save it so when we write out, we know the filename

bootstrap/pkg/kfapp/coordinator/coordinator.go
lluunn: done, thanks

jlewi:
@lluunn I would suggest not closing the issue until the fix has made it into the branch (if necessary) and RC. Lets continue to use the related issue to track any cherry-picks and RCs.

jlewi:
@lluunn 

> If the config file is remote, will save it in current directory as app.yaml.

Why can't we keep the name of the YAML file rather than using app.yaml? i.e. kfctl_gcp_iap.yaml should always be kfctl_gcp_iap.yaml

bootstrap/cmd/kfdefConverter/cmd/tov1beta1.go
```diff
@@ -40,6 +41,8 @@ var tov1beta1Cmd = &cobra.Command{
 		if outPath == "" {
 			outPath = args[0]
 		}
+		config.Spec.AppDir = filepath.Dir(outPath)
+		config.Spec.ConfigFileName = filepath.Base(outPath)
```
jlewi: Why not store the full  path to config.Spec.ConfigFileName as opposed to just the relative path?

lluunn: Because appDir is the directory.
Or we can change appDir + configfilename to be appFile (full path)?

bootstrap/pkg/apis/apps/configconverters/converters.go
```diff
@@ -49,7 +49,7 @@ func LoadConfigFromURI(configFile string) (*kfconfig.KfConfig, error) {
 	//
 	// TODO(jlewi): Should we use hashicorp go-getter.GetAny here? We use that to download
 	// the tarballs for the repos. Maybe we should use that here as well to be consistent.
-	appFile := path.Join(appDir, KfConfigFile)
+	appFile := path.Join(appDir, "tmp_app.yaml")
```
jlewi: Doesn't this function need to set kfconfig.KfConfig.ConfigFile? So that it can be referenced later.

LoadConfigFromURI is called from coordinator.go 
https://github.com/kubeflow/kubeflow/blob/7727e577413a63cb4134258107be312c3b72fe04/bootstrap/pkg/kfapp/coordinator/coordinator.go#L277

I think LoadConfigFromURI should set the URI from which the kfdefconfig was loaded.

LoadKFAppCfgFile calls CreateKfAppFile to actually persist the file
https://github.com/kubeflow/kubeflow/blob/7727e577413a63cb4134258107be312c3b72fe04/bootstrap/pkg/kfapp/coordinator/coordinator.go#L285

That eventually calls WriteConfigToFile

I think WriteConfigToFile could transform the URI to a local path.

bootstrap/pkg/apis/apps/configconverters/converters.go
```diff
@@ -145,7 +145,20 @@ func isCwdEmpty() string {
 	return cwd
 }
 
-func WriteConfigToFile(config kfconfig.KfConfig, filename string) error {
+func WriteConfigToFile(config kfconfig.KfConfig) error {
```
jlewi: Do you want to take a pointer so that you can potentially change ConfigFileName?
My expectation would be that LoadFromURI sets ConfigFileName originally to the URI e.g.
https://.../kfctl_gcp_iap.yaml

Then somewhere in the code base we would replace that with the local file path.

A natural place to do that would be in WriteConfigToFile.

bootstrap/pkg/apis/apps/configconverters/converters.go
lluunn: I am doing it in LoadConfigFromURI now.

bootstrap/pkg/apis/apps/configconverters/converters.go
lluunn: doing it in LoadConfigFromURI

lluunn:
> @lluunn
> 
> > If the config file is remote, will save it in current directory as app.yaml.
> 
> Why can't we keep the name of the YAML file rather than using app.yaml? i.e. kfctl_gcp_iap.yaml should always be kfctl_gcp_iap.yaml

done

jlewi:
@lluunn Is the PR description still up to date? If not could you please update it?

lluunn:
> @lluunn Is the PR description still up to date? If not could you please update it?

updated

jlewi:
/lgtm
/approve
/hold
for @gabrielwen

PR #4350, opened 2019-10-18
kf_is_ready_test: add test for workload identity and GCP credentials

kunmingg:
And tests to check workload identity policy setup and GCP credentials insertion.
related: https://github.com/kubeflow/testing/pull/497

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4350)

### DIFF
--- a/testing/kfctl/kf_is_ready_test.py
+++ b/testing/kfctl/kf_is_ready_test.py
@@ -7,6 +7,9 @@ import uuid
 import yaml
 from retrying import retry
 
+import googleapiclient.discovery
+from oauth2client.client import GoogleCredentials
+
 import pytest
 
 from kubeflow.testing import util
@@ -19,6 +22,26 @@ def set_logging():
                       datefmt='%Y-%m-%dT%H:%M:%S',
                       )
   logging.getLogger().setLevel(logging.INFO)
+
+def get_platform_app_name(app_path):
+  with open(os.path.join(app_path, "app.yaml")) as f:
+    kfdef = yaml.safe_load(f)
+  app_name = kfdef["metadata"]["name"]
+  platform = ""
+  apiVersion = kfdef["apiVersion"].strip().split("/")
+  if len(apiVersion) != 2:
+    raise RuntimeError("Invalid apiVersion: " + kfdef["apiVersion"].strip())
+  if apiVersion[-1] == "v1alpha1":
+    platform = kfdef["spec"]["platform"]
+  elif apiVersion[-1] == "v1beta1":
+    for plugin in kfdef["spec"].get("plugins", []):
+      if plugin.get("kind", "") == "KfGcpPlugin":
+        platform = "gcp"
+      elif plugin.get("kind", "") == "KfExistingArriktoPlugin":
+        platform = "existing_arrikto"
+  else:
+    raise RuntimeError("Unknown version: " + apiVersion[-1])
+  return platform, app_name
 
 def test_kf_is_ready(record_xml_attribute, namespace, use_basic_auth, use_istio,
                      app_path):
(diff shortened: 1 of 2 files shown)


jlewi:
Looks like test failure was some weird prow issue.

```
https://screenshot.googleplex.com/UvivsyrWiVe
```

testing/kfctl/kf_is_ready_test.py
```diff
@@ -145,6 +145,34 @@ def test_kf_is_ready(namespace, use_basic_auth, use_istio, app_path):
     logging.info("Verifying that deployment %s started...", deployment_name)
     util.wait_for_deployment(api_client, knative_namespace, deployment_name, 10)
 
+  if platform == "gcp":
```
jlewi: This should be a new test function in this module (a pytest module can have multiple test functions)

e.g.

```
test_workload_identity
```

We want separate functions to make it easier to distinguish failures with workload identity versus other tests.

jlewi: Use pytest.Skip to skip the test if we aren't running on GCP
http://doc.pytest.org/en/latest/skipping.html

testing/kfctl/kf_is_ready_test.py
```diff
@@ -145,6 +145,34 @@ def test_kf_is_ready(namespace, use_basic_auth, use_istio, app_path):
...
+    # check secret
+    util.wait_for_secret(api_client, namespace, "user-gcp-sa")
```
jlewi: Why do we need to "wait_for_secret". The secret should already exist right? We don't need to wait for the secret to be ready.

kunmingg: A better name is "check_secret"

kunmingg:
@jlewi presubmit depend on https://github.com/kubeflow/testing/pull/497

testing/kfctl/kf_is_ready_test.py
```diff
@@ -62,34 +84,16 @@ def test_kf_is_ready(namespace, use_basic_auth, use_istio, app_path):
...
-    platform = kfdef["spec"]["platform"]
-  elif apiVersion[-1] == "v1beta1":
-    for plugin in config_spec["spec"].get("plugins", []):
-      if plugin.get("kind", "") == "KfGcpPlugin":
-        platform = "gcp"
-      elif plugin.get("kind", "") == "KfExistingArriktoPlugin":
-        platform = "existing_arrikto"
-  else:
-    raise RuntimeError("Unknown version: " + apiVersion[-1])
+  platform = get_paltform(app_path)
```
lluunn: typo: paltform ->platform

testing/kfctl/kf_is_ready_test.py
```diff
@@ -20,7 +23,26 @@ def set_logging():
...
   logging.getLogger().setLevel(logging.INFO)
 
-def test_kf_is_ready(namespace, use_basic_auth, use_istio, app_path):
+def get_paltform(app_path):
+  with open(os.path.join(app_path, "app.yaml")) as f:
+    kfdef = yaml.safe_load(f)
+  platform = ""
+  apiVersion = kfdef["apiVersion"].strip().split("/")
+  if len(apiVersion) != 2:
+    raise RuntimeError("Invalid apiVersion: " + apiVersion)
```
lluunn: will this work? it's string + list

testing/kfctl/kf_is_ready_test.py
```diff
@@ -145,6 +149,52 @@ def test_kf_is_ready(namespace, use_basic_auth, use_istio, app_path):
     logging.info("Verifying that deployment %s started...", deployment_name)
     util.wait_for_deployment(api_client, knative_namespace, deployment_name, 10)
 
+
+def test_gcp_access(namespace, app_path, app_name, project_id):
```
jlewi: Why do we need app_name and project_id? Can't we get those from app_path?

You should also add a function argument for the record_xml_attribute
Then call 

```
 util.set_pytest_junit(record_xml_attribute, "test_gcp_access")
```

This will set some xml attributes in the junit file that will make it easier to group tests based on workflow.

kunmingg: Updated to get rid of app_name.
Can extract project once kfdef version migration is done.

kunmingg:
rebase to resolve conflict

PR #4359, opened 2019-10-21
Resolves manifest apply retry issues with kfctl

### LINKED ISSUE #4327: kfctl: handle CRD / Webhook establishment
/kind bug

Kfctl will sometimes fail because of the following reasons:
1. CRDs haven't become established yet and CR creation fails.
2. Admission Webhooks haven't become ready yet and CR creation fails.

Kfctl should have some mechanism to counter these kind of scenarios.

Some more details here: https://github.com/kubeflow/kubeflow/issues/3989#issuecomment-532638203
TLDR: the current way of defending against these scenarios is ineffective.

**Possible Solutions:**
1. Retry on failed apply with a backoff.
2. Wait for establishment when encountering a CRD. This requires parsing the yaml contents and isn't straightforward for Webhooks, since the definition is different from the actual deployment.

I recommend going with (1) for now, as the easiest thing to implement immediately.

/cc @kkasravi @jlewi @krishnadurai @swiftdiaries @gabrielwen

krishnadurai:
Resolves #4327 

[cmdutil.CheckErr](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubectl/pkg/cmd/util/helpers.go#L119) exits execution in apply.run(). 

Now CheckError has been moved to kustomization.go instead - so that retry is functional.

P.S. Credits to @johnugeorge for suggesting this.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4359)

### DIFF
--- a/bootstrap/go.sum
+++ b/bootstrap/go.sum
@@ -82,6 +82,7 @@ github.com/davecgh/go-spew v1.1.1 h1:vj9j/u1bqnvCEfJOwUhtlOARqs3+rkHYY13jYWTU97c
 github.com/davecgh/go-spew v1.1.1/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/deckarep/golang-set v1.7.1 h1:SCQV0S6gTtp6itiFrTqI+pfmJ4LN85S1YzhDf9rTHJQ=
 github.com/deckarep/golang-set v1.7.1/go.mod h1:93vsz/8Wt4joVM7c2AVqh+YRMiUSc14yDtF28KmMOgQ=
+github.com/dgrijalva/jwt-go v3.2.0+incompatible h1:7qlOGliEKZXTDg6OTjfoBKDXWrumCAMpl/TFQ4/5kLM=
 github.com/dgrijalva/jwt-go v3.2.0+incompatible/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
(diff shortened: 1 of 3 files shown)


yanniszark:
It looks like we submitted a PR at the same time :stuck_out_tongue: 
https://github.com/kubeflow/kubeflow/pull/4360

I had a bit different approach in my PR, can you take a look and tell me what you think?
For these changes, I have some comments:
1. I don't think we should CheckErr at all. The kustomize.Apply() function returns an error.
2. Why use the more complicated RetryNotify function and not just the backoff.Retry one? It seems like a lot of locs were added with no clear benefit.

krishnadurai:
@yanniszark 
> I had a bit different approach in my PR, can you take a look and tell me what you think?

Sure, I'll review it shortly.

>     1. I don't think we should CheckErr at all. The kustomize.Apply() function returns an error.

The idea here was to exit execution at the point of failure and discontinue further execution.

>     2. Why use the more complicated RetryNotify function and not just the backoff.Retry one? It seems like a lot of locs were added with no clear benefit.

Right now, kustomize.Apply() function takes an entire application definition and its overlays, combines them and writes it into a temporary file. This file is then passed to kubectl's apply function.
In case of failure during kubectl apply, the entire application of the manifest fails. This requires a prompt at the CLI end to notify the user that `kfctl` is going to retry applying the manifest. Otherwise, it may be misleading for the user to see the same output for application reappear several times.
RetryNotify allows us to log output to stdout when a retry is attempted.

jlewi:
@krishnadurai should we close this given #4360?

PR #4360, opened 2019-10-21
kfctl: fix apply retry

yanniszark:
Related https://github.com/kubeflow/kubeflow/issues/4327

Changes the default behavior from exit to panic.
Then recovers the panic inside the Apply function and returns an error.

Copying from the issue:

> It seems that kubectl uses os.Exit() when encountering an error, instead of returning one.
I guess that's to be expected, since it's a CLI tool.

> From comments in the code, I've seen that this behaviour can also be changed to a panic.
This could work for us. For example, the apply.Apply() function could recover the panic and return an error instead.
Will look into it soon.

/cc @krishnadurai @jlewi @kkasravi 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4360)

### DIFF
--- a/bootstrap/pkg/utils/k8utils.go
+++ b/bootstrap/pkg/utils/k8utils.go
@@ -279,6 +279,12 @@ func NewApply(namespace string, restConfig *rest.Config) (*Apply, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Change behaviour of error from exit to panic.
+	cmdutil.BehaviorOnFatal(func(s string, i int) {
+		panic(fmt.Errorf("Encountered error: %s. Exiting with code: %d.", s, i))
+	})
+
 	return apply, nil
 }
 
@@ -305,7 +311,15 @@ func (a *Apply) Apply(data []byte) error {
 			Message: fmt.Sprintf("could not initialize  Error %v", initializeErr),
 		}
 	}
-	err := a.run()
+	var err error
+	func() {
+		defer func() {
+			if temp := recover(); temp != nil {
+				err = temp.(error)
+			}
+		}()
+		err = a.run()
+	}()
 	if err != nil {
 		return err
 	}
@@ -315,7 +329,6 @@ func (a *Apply) Apply(data []byte) error {
 func (a *Apply) run() error {
 	resourcesErr := a.options.Run()
 	if resourcesErr != nil {
-		cmdutil.CheckErr(resourcesErr)
 		return &kfapis.KfError{
 			Code:    int(kfapis.INTERNAL_ERROR),
 			Message: fmt.Sprintf("Apply.Run  Error %v", resourcesErr),
--- a/bootstrap/pkg/kfapp/kustomize/kustomize.go
+++ b/bootstrap/pkg/kfapp/kustomize/kustomize.go
@@ -55,6 +55,8 @@ import (
 	"sigs.k8s.io/kustomize/v3/pkg/validators"
 	"sigs.k8s.io/kustomize/v3/plugin/builtin"
 	"strings"
+	"time"
+
 	// Auth plugins
 	_ "k8s.io/client-go/plugin/pkg/client/auth/azure"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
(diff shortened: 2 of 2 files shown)


bootstrap/pkg/utils/k8utils.go
```diff
@@ -279,6 +279,12 @@ func NewApply(namespace string, restConfig *rest.Config) (*Apply, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// Change behaviour of error from exit to panic.
+	cmdutil.BehaviorOnFatal(func(s string, i int) {
+		panic(fmt.Errorf("Encountered error: %s. Exiting with code: %d.", s, i))
```
jlewi: Why are we doing panic? Doesn't panic attempt to crash the program? Why would we want to crash?

jlewi:
@yanniszark Could you please update the PR description to explain what the issue is and how this is addressing it?

bootstrap/pkg/utils/k8utils.go
yanniszark: @jlewi the kubectl-related libraries use the cmdutil CheckErr function, which uses os.Exit by default.
By changing the exit to a panic, we are given the chance to intercept the panic and return an error instead.

yanniszark:
@krishnadurai I would expect that right now, only the k8sutil Apply-related functions use kubectl libraries which could invoke os.Exit().
As such, all current cases should be covered.

As for the exit vs panic:
I personally vote for panic.
It gives a stack trace which we can use to easily pinpoint the leak and fix it.
In contrast, os.Exit() doesn't give any context about where the issue occured.

krishnadurai:
@yanniszark it seems that apply is the only user of kubectl libraries.

When we propagate the error handled after panic,  I suggest we exit `kfctl` after the configured number of retries.

bootstrap/pkg/utils/k8utils.go
jlewi: @krishnadurai How is that different from the default behavior which is os.Exit? If there is some place in the kfctl code base that is calling a kubectl library (e.g. cmdutil) then if it returns an error by default it would just call os.Exit and terminate execution if I'm understanding @yanniszark correctly.

So if we change the default behavior to panic and don't intercept the panic then the behavior would still be to crash on error.

jlewi:
@yanniszark I would suggest not marking #4327  as fixed by this issue. Assuming this is a fix that needs to be included in 0.7.0 we should leave the bug open until we have an RC with the fix and have verified the fix.

We should use the issue to track any cherry-picks and necessary RCs.

jlewi:
@yanniszark @krishnadurai Do we have an ETA for when we can get a fix?

bootstrap/pkg/utils/k8utils.go
krishnadurai: @jlewi behaviour wise things would be the same for panic and os.Exit. The only difference is what's printed out to stdout. If there's an unhandled panic it would lead to kfctl printing a traceback like the one shown below (simulated this one):

```
panic: Encountered error: error: no objects passed to apply. Exiting with code: 1.

goroutine 1 [running]:
github.com/kubeflow/kubeflow/bootstrap/v3/pkg/utils.NewApply.func1(0xc00067fec0, 0x21, 0x1)
	/Users/duselvar/golang/src/github.com/kubeflow/kubeflow/bootstrap/pkg/utils/k8utils.go:285 +0x12e
k8s.io/kubernetes/pkg/kubectl/cmd/util.checkErr(0x5b65700, 0xc000f18130, 0x58fb7b8)
	/Users/duselvar/golang/pkg/mod/k8s.io/kubernetes@v1.13.4/pkg/kubectl/cmd/util/helpers.go:170 +0x8da
k8s.io/kubernetes/pkg/kubectl/cmd/util.CheckErr(...)
	/Users/duselvar/golang/pkg/mod/k8s.io/kubernetes@v1.13.4/pkg/kubectl/cmd/util/helpers.go:116
github.com/kubeflow/kubeflow/bootstrap/v3/pkg/utils.(*Apply).run(0xc0004b89c0, 0x0, 0x0)
	/Users/duselvar/golang/src/github.com/kubeflow/kubeflow/bootstrap/pkg/utils/k8utils.go:335 +0x62
...
```

krishnadurai:
> @yanniszark @krishnadurai Do we have an ETA for when we can get a fix?

We should look to get this in today.

> And I think it would be better to inform the user of `kfctl` about retry failures as explained in this [comment](https://github.com/kubeflow/kubeflow/pull/4359#issuecomment-544574094).

@yanniszark, in my opinion, this change would be necessary.

yanniszark:
@krishnadurai agreed, I will print messages that inform of the retry.

yanniszark:
@krishnadurai @jlewi updated the PR to include a log message.
Could you take a look?

yanniszark:
@krishnadurai had to rebase on master. Could you lgtm again?

PR #4380, opened 2019-10-23
fix kfctl corner cases

kunmingg:
* SetApplicationParameter will not throw error if application not found.
  * application might be deleted from config file on purpose
* For same reason we print warnings instead of error when default namespace not found.
* Also update cmd help messages to include local config example.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4380)

### DIFF
--- a/bootstrap/pkg/apis/apps/kfconfig/types.go
+++ b/bootstrap/pkg/apis/apps/kfconfig/types.go
@@ -571,11 +571,10 @@ func (c *KfConfig) SetApplicationParameter(appName string, paramName string, val
 
 			return nil
 		}
-		log.Warnf("Application %v not found", appName)
-		return nil
-	}
 
-	return &AppNotFound{Name: appName}
+	}
+	log.Warnf("Application %v not found", appName)
+	return nil
 }
 
 // SetSecret sets the specified secret; if a secret with the given name already exists it is overwritten.
--- a/bootstrap/pkg/apis/apps/kfconfig/types.go
+++ b/bootstrap/pkg/apis/apps/kfconfig/types.go
@@ -571,6 +571,8 @@ func (c *KfConfig) SetApplicationParameter(appName string, paramName string, val
 
 			return nil
 		}
+		log.Warnf("Application %v not found", appName)
+		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


bootstrap/pkg/kfapp/kustomize/kustomize.go
```diff
@@ -204,6 +204,8 @@ func (kustomize *kustomize) Apply(resources kftypesv3.ResourceEnum) error {
 		}
 		return nil
 	}, b)
+	log.Warnf("Default namespace creation skipped")
```
lluunn: should we check err != nil and then print warning?

bootstrap/pkg/apis/apps/kfconfig/types.go
```diff
@@ -571,6 +571,8 @@ func (c *KfConfig) SetApplicationParameter(appName string, paramName string, val
 
 			return nil
 		}
+		log.Warnf("Application %v not found", appName)
```
lluunn: should we do this outside the if at line 577?

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -1809,12 +1809,17 @@ func (gcp *Gcp) ConfigPodDefault() error {
 		return kfapis.NewKfErrorWithMessage(err, "set K8s clientset error")
 	}
 
+	defaultNamespace := kftypesv3.EmailToDefaultName(gcp.kfDef.Spec.Email)
+	_, err = k8sClient.CoreV1().Namespaces().Get(defaultNamespace, metav1.GetOptions{})
+	if err != nil {
+		log.Warnf("Default namespace %v creation skipped", defaultNamespace)
+		return nil
+	}
```
gabrielwen: why this?  the default namespace is created with the default profile in kustomize.go, right?

kunmingg: Currently default profile is one entry application entry from manifest. If user delete this entry default namespace will not be created.

kunmingg:
If default profile is skipped then default namespace + pod-default will be skipped as well.

PR #4384, opened 2019-10-24
[CentralDashboard v2] Polling for profile state (for 20 seconds) to allow a transition to dashboard

avdaredevil:
## About
- Registration polling added to check for a profile before running env-info
- Nits tackled
- Removed an unnecessary wait

### Meta
/area centraldashboard
/area front-end
/priority p1
/assign @avdaredevil
/cc @prodonjs @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4384)

### DIFF
--- a/components/centraldashboard/public/components/registration-page.js
+++ b/components/centraldashboard/public/components/registration-page.js
@@ -87,12 +87,12 @@ export class RegistrationPage extends utilitiesMixin(PolymerElement) {
         );
     }
 
-    async pollProfile(times = 10, delay = 500) {
+    async pollProfile(times, delay) {
         const profileAPI = this.$.GetMyNamespace;
         if (times < 1) throw Error('Cannot poll profile < 1 times!');
         for (let i = 0; i < times; i++) {
             const req = profileAPI.generateRequest();
-            await req.completes;
+            await req.completes.catch(() => 0);
             if (req.response && req.response.hasWorkgroup) return true;
             await this.sleep(delay);
         }
--- a/components/centraldashboard/public/components/registration-page.js
+++ b/components/centraldashboard/public/components/registration-page.js
@@ -87,6 +87,17 @@ export class RegistrationPage extends utilitiesMixin(PolymerElement) {
         );
     }
 
+    async pollProfile(times = 10, delay = 500) {
+        const profileAPI = this.$.getMyNamespace;
+        if (times < 1) throw Error('Cannot poll profile < 1 times!');
+        for (let i = 0; i < times; i++) {
+            const req = profileAPI.generateRequest();
+            await req.completes;
+            if (req.response && req.response.hasWorkgroup) return true;
+            await this.sleep(delay);
+        }
+    }
+
     async finishSetup() {
         const API = this.$.MakeNamespace;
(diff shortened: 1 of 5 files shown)


components/centraldashboard/public/components/registration-page.js
```diff
@@ -87,19 +87,31 @@ export class RegistrationPage extends utilitiesMixin(PolymerElement) {
         );
     }
 
+    async pollProfile(times = 10, delay = 500) {
```
prodonjs: Do you think it makes sense to remove the parameters here and just add these values as module constants? Since this isn't a general purpose function and only gets used once, it seems like that simplifies things.

components/centraldashboard/public/components/registration-page.js
```diff
@@ -87,19 +87,31 @@ export class RegistrationPage extends utilitiesMixin(PolymerElement) {
...
+        await this.sleep(1); // So the errors and callbacks can schedule
         if (this.error && this.error.response) {
             return this.waitForRedirect = false;
         }
-        // If request completed and 6 seconds pass, probably let
-        // the user click next again!
-        await this.sleep(6000);
+        // Poll for profile over a span of 20 seconds (every 300ms)
+        // if still not there, let the user click next again!
+        const success = await this.pollProfile(66, 300);
```
prodonjs: There is no `catch` for the `pollProfile` call. If something were to fail, even due to an intermittent network glitch, it looks like you might get stuck in an inconsistent state.

components/centraldashboard/public/components/registration-page_test.js
```diff
@@ -59,9 +63,41 @@ describe('Registration Page', () => {
...
+
+        await sleep(5);
+        expect(flowcomplete).not.toHaveBeenCalled();
+
+        mockIronAjax(
+            registrationPage.$.GetMyNamespace,
+            {hasWorkgroup: true, hasAuth: true, user: 'test'},
+        );
+
+        await sleep(300);
```
prodonjs: If you want to avoid having real waits in your tests, you can use the Jasmine fake clock to mock out the timer behavior and advance it to run any timers.

https://jasmine.github.io/api/3.5/Clock.html

components/centraldashboard/public/components/registration-page.js
avdaredevil: Yup, will do!

components/centraldashboard/public/components/registration-page.js
avdaredevil: It's supposed to return true or false, but yeah I'll mute network errors.

components/centraldashboard/public/components/registration-page_test.js
avdaredevil: It's not about the timer, as it is about scheduling, does clock control tick scheduling?

avdaredevil: Okay seems like it works, but it is really nuanced (updated tests to use that!)

avdaredevil:
Updated tests with comments, and added the negative case check @prodonjs! Thanks, and PTAL

Issue #4397, opened 2019-10-24
kfctl 2nd apply failed


lluunn:
/kind bug

**What steps did you take and what happened:**
kfctl build -V -f http://...
kfctl apply -V -f kfctl_gcp_iap.yaml
kfctl apply -V -f kfctl_gcp_iap.yaml

The second apply failed with:

```
INFO[0100] Downloading secret user-gcp-sa from namespace kubeflow  filename="gcp/gcp.go:1812"
INFO[0100] Creating secret user-gcp-sa to namespace kubeflow-lunkai  filename="gcp/gcp.go:1818"
INFO[0100] Secret kubeflow-lunkai.user-gcp-sa already exists  filename="gcp/gcp.go:1425"
INFO[0100] Generating PodDefault in namespace kubeflow-lunkai; APIVersion kubeflow.org/v1alpha1  filename="gcp/gcp.go:1753"
Error: poddefaults.kubeflow.org "add-gcp-secret" already exists
Usage:
  kfctl apply -f ${CONFIG} [flags]

Flags:
  -f, --file string   Static config file to use. Can be either a local path or a URL.
                      For example:
                      export CONFIG=https://raw.githubusercontent.com/kubeflow/manifests/v0.7-branch/kfdef/kfctl_existing_arrikto.yaml
                      export CONFIG=https://raw.githubusercontent.com/kubeflow/manifests/v0.7-branch/kfdef/kfctl_aws.yaml
                      export CONFIG=https://raw.githubusercontent.com/kubeflow/manifests/v0.7-branch/kfdef/kfctl_gcp_iap.yaml
                      export CONFIG=https://raw.githubusercontent.com/kubeflow/manifests/v0.7-branch/kfdef/kfctl_k8s_istio.yaml
                      kfctl apply -V --file=${CONFIG}
  -h, --help          help for apply
  -V, --verbose       verbose output default is false

poddefaults.kubeflow.org "add-gcp-secret" already exists
```

**Environment:**
kfctl build from master

cc @jlewi @kunmingg @gabrielwen

jlewi:
Reopening because this hasn't been cherry-picked to the release branch

jlewi:
I checked using basic auth and

kubeflow/kubeflow - v0.7.0-rc.7-3-g8dbde9d8
kubeflow/manifests - v0.7.0-rc.2-19-g317c0d8

rerunning `kfctl apply -V -f ...` succeeded.

PR #4398, opened 2019-10-24
check pod default before post request

lluunn:
Verified that 2nd apply succeeded with the fix.

fix https://github.com/kubeflow/kubeflow/issues/4397

/cc @kunmingg @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4398)

### DIFF
--- a/bootstrap/pkg/kfapp/gcp/gcp.go
+++ b/bootstrap/pkg/kfapp/gcp/gcp.go
@@ -94,6 +94,8 @@ const (
 	GcpPluginName               = "KfGcpPlugin"
 	GcpAccessTokenName          = "accessToken"
 	BasicAuthPasswordSecretName = "password"
+	// The PodDefault in default namespace
+	PodDefaultName = "add-gcp-secret"
 )
 
 // Gcp implements KfApp Interface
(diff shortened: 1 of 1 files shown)


bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -1881,6 +1881,12 @@ func (gcp *Gcp) ConfigPodDefault() error {
 		}
 	}
 
+	getReq := crdClient.Get().Resource(mapping.Resource.Resource).Namespace(defaultNamespace).Name("add-gcp-secret")
```
jlewi: Nit define a constant for the name and use it here and here
https://github.com/kubeflow/kubeflow/blob/0a370635ad1db8139a1cf56bef329f34344fea6d/bootstrap/pkg/kfapp/gcp/gcp.go#L1761

lluunn: done

jlewi:
@abhi-g This is for backwards compatibility but it is in the default namespace created by kfctl. This matches 0.6 behavior so if someone is using PodDefault in that namespace things would still work.

/lgtm
/approve
/hold cancel

Issue #4376, opened 2019-10-23
RC5 Darwin binary name is inconsistent


arun-gupta:
/kind bug

RC5 Darwin binaries are called `kfctl-darwin` and Linux binaries are called `kfctl`. This change happened with RC5. This naming scheme is inconsistent with earlier releases, and is also not inconsistent with `kubectl`.

Can we change back to `kfctl` as the binary name?

@Jeffwan

PR #4406, opened 2019-10-25
rename darwin binary as kfctl

swiftdiaries:
Resolves: https://github.com/kubeflow/kubeflow/issues/4376
/cc @Jeffwan

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4406)

### DIFF
--- a/bootstrap/Makefile
+++ b/bootstrap/Makefile
@@ -102,8 +102,11 @@ build-bootstrap: deepcopy generate fmt vet
 	${GO} build -gcflags '-N -l' -o bin/bootstrapper cmd/bootstrap/main.go
 
 build-kfctl: deepcopy generate fmt vet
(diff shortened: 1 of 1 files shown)


bootstrap/Makefile
```diff
@@ -112,8 +112,8 @@ build-gc: deepcopy generate fmt vet
 build-kfctl-tgz: build-kfctl
 	chmod a+rx ./bin/kfctl
 	rm -f bin/*.tgz
-	cd bin && tar -cvzf kfctl_$(TAG)_linux.tar.gz ./kfctl
-	cd bin && tar -cvzf kfctl_${TAG}_darwin.tar.gz ./kfctl-darwin
+	cd bin && tar -cvzf kfctl_$(TAG)_linux.tar.gz ./kfctl && rm ./kfctl
```
jlewi: Would it be better to use different subdirectories for different architectures?
e.g
bin/${ARCH}/kfctl

Swapping out bin/kfctl for different architectures seems like it could lead to some confusing error

swiftdiaries: Oh yeah much better 😅

PR #4408, opened 2019-10-25
Support set enviroment and command in notebook backend

SaintBacchus:
Add enviroment and command field in create notebook request in kubeflow components of jupter-web-app.
New fileds example:
```json
{
    "env":[
        {
            "name":"Key1",
            "value":"Value1"
        },
        {
            "name":"Key2",
            "value":"Value2"
        }
    ],
    "command":[
        "tini",
        "--"
    ]
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4408)

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -471,3 +471,19 @@ def add_notebook_volume_secret(nb, secret, secret_name, mnt_path, mode):
         "name": secret,
     }
     container["volumeMounts"].append(mnt)
+
+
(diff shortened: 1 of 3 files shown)


nrchakradhar:
@SaintBacchus May be a foolish query: Will 'command' and 'env' settings be available for changing via the Notebook server GUI from where notebooks are created?

For community provided images, my understanding is this is not applicable. However for custom images built by users, this will be useful. Is this understanding correct?

It may be good to document that injecting of command and env gives greater control for users but also can be harmful when used wrongly.

SaintBacchus:
Yes, the command is only used for custom images. It's better to have a GUI for this.

PR #4416, opened 2019-10-25
Fix bug preventing presubmits on kubeflow/manifests branches from using correct commit of kubeflow/manifests

jlewi:
* When triggering tests on kubeflow/manifests we need to modify the
  KFDef spec to change the URI of the kubeflow/manifests repo in the
  KFDef spec.

* There was a bug in the code where we were checking the URI to decide
  whether to override it or not. This code assumed we were on the master
  branch so presubmits on the kubeflow/manifests v0.7-branch weren't using
  the correct commit.

* This change fixes this PR to properly set repos in the KFDef when
  triggering on kubeflow/manifests

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4416)

### DIFF
--- a/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
+++ b/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
@@ -184,19 +184,20 @@ def get_config_spec(config_path, project, email, zone, app_path):
       if repo["name"] !=  manifests_repo_name:
         continue
 
-      version = os.getenv("PULL_NUMBER")
+      version = None
 
-      if not version:
-        # Postsubmit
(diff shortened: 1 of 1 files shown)


jlewi:
Typo

```
kfctl_go_test.py:6: in <module>
   from kubeflow.kubeflow.ci import kfctl_go_test_utils as kfctl_util
E     File "/mnt/test-data-volume/kubeflow-presubmit-kfctl-go-iap-4416-4f0717d-6980-c88f/src/kubeflow/kubeflow/py/kubeflow/kubeflow/ci/kfctl_go_test_util
s.py", line 196
E       elif os.getenv("PULL_PULL_SHA"):
```

PR #4417, opened 2019-10-26
Fix: error isn't logged before retry

krishnadurai:
During apply failure, kustomize.Apply() retries application warning the user about currently encountered error. This error is printed as <nil> as the variable is set is incorrect.

```
WARN[0016] Encountered error during apply: <nil>         filename="kustomize/kustomize.go:183"
WARN[0016] Will retry in 13 seconds. 
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4417)


yanniszark:
Thanks @krishnadurai.
/approve

Issue #2337, opened 2019-01-26
Kubeflow on arm64


vielmetti:
Referencing #819, and similar to #925/#1928 -

Kubernetes currently runs on arm64, and there is interest to get kubeflow running on arm64 as well. I am happy to entertain any suggestions on where to start first in terms of porting, either from the "most difficult projects to tackle first" point of view, or the "easiest possible way to get a proof of concept" as well.

jlewi:
@vielmetti that sounds great. How can we help? 

My suggestion would probably to be take one of the examples e.g. [mnist](https://github.com/kubeflow/examples/tree/master/mnist) try to run it on Arm and just see where things get start.

To run on ARM will all of our docker images need to be rebuilt to support ARM? 

Can you and the other folks working on supporting other platforms like IBM power 8 provide guidance on what's needed from the Kubeflow project in order to support other architectures?

jmreicha:
Still interested in this.

smeyer12:
list of apps we plan to have working on arm64 includes: k8s, tensorflow, caffe, keras, theano, torch/pytorch, kubeflow, jupyter, spark/pyspark, R, julia.

MrXinWang:
Hi @jlewi !

I have setup a kubeflow environment (kubeflow v0.5.1, deployed on microk8s) on our arm64 server with all component images rebuilt for arm64 architecture. Currently all containers work well and I have also run [a demo of object detection](https://github.com/kubeflow/examples/tree/master/object_detection) successfully. 

Now I would like to contribute this to the community, is there any suggestions or comments for me to do this? Should I upload images or Dockerfile of them?

elviraux:
@jlewi were you able to help @MrXinWang ?

jlewi:
@MrXinWang I would suggest talking to @jinchihe and looking at what we are doing for power #4133 we should try to follow the same approach for arm as we are following for power

MrXinWang:
@jlewi @vielmetti Thanks for the kindly suggestion! Working on this.

vielmetti:
It looks like this issue has not had activity for a few months, but also that a patch has been merged to a related project https://github.com/kubeflow/metadata/pull/164 in the intervening time, I'm interested to get a status update.

At the last we looked @MrXinWang was making some progress - hoping to get an update on anything relevant there as time and continued interest allows.

MrXinWang:
Hi @vielmetti , I was on the vacation before and therefore sorry for the late reply.

[...]

PR #4422, opened 2019-10-29
Add arm64 support for centraldashboard and Fix chromium@edge version cannot start bug

MrXinWang:
Referencing [comment ](https://github.com/kubeflow/kubeflow/issues/2337#issuecomment-524690065)and issue [#2337](https://github.com/kubeflow/kubeflow/issues/2337)
 We are working on Kubeflow supporting arm64 platform. We now rebuilt some images and installed Kubeflow on arm64, and hence contribute the Dockerfile.

1. Add arm64 support for centraldashboard, tested on both arm64 and x86_64 machines.
2. Refer to [issue](https://github.com/Zenika/alpine-chrome/issues/39), fix chromium@edge version cannot start bug.

Signed-off-by: Henry Wang <<henry.wang@arm.com>>

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4422)

### DIFF
--- a/components/centraldashboard/Dockerfile
+++ b/components/centraldashboard/Dockerfile
@@ -9,20 +9,31 @@ ENV CHROME_BIN=/usr/bin/chromium-browser
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
 
 # Installs latest Chromium package and configures environment for testing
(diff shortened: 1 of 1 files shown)


MrXinWang:
@prodonjs Thanks for the comment, tests look good now. Are there any extra stuff that I need to do? Thanks

Issue #4421, opened 2019-10-28
kfctl / C2D: throw warning when length(deploy-name.endpoints.project-id.cloud.goog) > 63


kunmingg:
/kind bug
This is a limitation from let's encrypt.
We should throw warning messages and exit when 
length(deploy-name.endpoints.project-id.cloud.goog) > 63

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

amygdala:
maybe we should hash the project+deployment string and use the hash for the cert.

kunmingg:
cert hostname need to match cloud endpoint hostname

PR #4436, opened 2019-10-30
Add a check if hostname is too long for GCP endpoint

zhenghuiwang:
related to https://github.com/kubeflow/kubeflow/issues/4421

needs cherry-pick later.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4436)


kunmingg:
@jlewi 
Now the app name is folder name, seems user have to recreate folder?

zhenghuiwang:
> @jlewi
> Now the app name is folder name, seems user have to recreate folder?

I don't think it is a problem.

Folder name is the default if name in spec is not set. User can still reset a shorter name in spec explicitly without recreating a folder.

Issue #4439, opened 2019-10-31
GCP config yaml spec.secrets.secretSource` is configured empty for basic-auth


sarjeet2013:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

Hitting the following error when running kfctl apply with gcp config 0.7* yaml:
kfctl apply -- -V -f kfctl_gcp_basic_auth.0.7.0.yaml
```
INFO[0036] Creating GCP secrets for basic auth...        filename="gcp/gcp.go:1612"
ERRO[0036] There was a problem getting the password for basic auth; error No secret source provided for secret password  filename="gcp/gcp.go:1555"
INFO[0036] Error is not a KfError; No secret source provided for secret password  filename="apis/kferrors.go:59"
Error:  (kubeflow.error): Code 500 with message: coordinator Apply failed for gcp:  (kubeflow.error): Code 520 with message: gcp apply could not create secrets Error cannot create basic auth login secret; No secret source provided for secret password
```

**What did you expect to happen:**
kfctl apply works fine and able to see the pods in running state.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
I tried to debug the issue but not able to root cause the issue. The following env are set before kfctl is run.

KUBEFLOW_USERNAME=testuser
KUBEFLOW_PASSWORD=testpassword

Here is the excerpt from yaml:
```
secrets:
  - name: password
    secretSource: {}
```

Looking at the code, the above secretSource should have populated with the hashed password. 

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
0.7.0
- kfctl version: (use `kfctl version`):
kfctl v0.7.0-rc.5-20-g618c77be

- Kubernetes platform: (e.g. `minikube`)
GKE

- Kubernetes version: (use `kubectl version`):
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T23:41:55Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #4444, opened 2019-10-31
Store secrets using EnvSource for basic auth

zhenghuiwang:
#4439 

Makes suggested changes in https://github.com/kubeflow/kubeflow/issues/4439#issuecomment-548375713

- Use EnvSource instead of HashSource for secrets
- fix Json name for env secret name 
- remove literal secret from yaml file
- Change `buildBasicAuth` encode the plain text secret to base64

Testing:
- `kfctl apply -f https://raw.githubusercontent.com/kubeflow/manifests/v0.7-branch/kfdef/kfctl_gcp_basic_auth.0.7.0.yaml` works.
- second run `kfctl apply` with generated local yaml works too.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4444)

### DIFF
--- a/bootstrap/pkg/apis/apps/configconverters/v1beta1.go
+++ b/bootstrap/pkg/apis/apps/configconverters/v1beta1.go
@@ -137,12 +137,12 @@ func (v V1beta1) ToKfConfig(kfdefBytes []byte) (*kfconfig.KfConfig, error) {
 		s := kfconfig.Secret{
 			Name: secret.Name,
 		}
-		if secret.SecretSource == nil {
+		// We don't want to store literalSource explictly, becasue we want the config to be checked into source control and don't want secrets in source control.
+		if secret.SecretSource == nil || secret.SecretSource.LiteralSource != nil {
 			config.Spec.Secrets = append(config.Spec.Secrets, s)
 			continue
 		}
 		src := &kfconfig.SecretSource{}
-		// We don't want to store HashSource and LiteralSource explictly.
 		if secret.SecretSource.EnvSource != nil {
 			src.EnvSource = &kfconfig.EnvSource{
 				Name: secret.SecretSource.EnvSource.Name,
--- a/bootstrap/pkg/apis/apps/configconverters/v1beta1.go
+++ b/bootstrap/pkg/apis/apps/configconverters/v1beta1.go
@@ -2,6 +2,7 @@ package configconverters
 
 import (
 	"fmt"
+
 	"github.com/ghodss/yaml"
 	kfapis "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis"
 	kftypesv3 "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis/apps"
@@ -141,11 +142,8 @@ func (v V1beta1) ToKfConfig(kfdefBytes []byte) (*kfconfig.KfConfig, error) {
 			continue
 		}
 		src := &kfconfig.SecretSource{}
-		if secret.SecretSource.LiteralSource != nil {
-			src.LiteralSource = &kfconfig.LiteralSource{
-				Value: secret.SecretSource.LiteralSource.Value,
-			}
-		} else if secret.SecretSource.EnvSource != nil {
+		// We don't want to store HashSource and LiteralSource explictly.
+		if secret.SecretSource.EnvSource != nil {
 			src.EnvSource = &kfconfig.EnvSource{
 				Name: secret.SecretSource.EnvSource.Name,
 			}
--- a/bootstrap/pkg/apis/apps/kfconfig/types.go
+++ b/bootstrap/pkg/apis/apps/kfconfig/types.go
@@ -13,7 +13,7 @@ import (
 	kfapis "github.com/kubeflow/kubeflow/bootstrap/v3/pkg/apis"
 	"github.com/pkg/errors"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 )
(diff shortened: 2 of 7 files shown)


bootstrap/pkg/apis/apps/configconverters/v1beta1.go
```diff
@@ -141,11 +142,8 @@ func (v V1beta1) ToKfConfig(kfdefBytes []byte) (*kfconfig.KfConfig, error) {
 			continue
 		}
 		src := &kfconfig.SecretSource{}
-		if secret.SecretSource.LiteralSource != nil {
-			src.LiteralSource = &kfconfig.LiteralSource{
-				Value: secret.SecretSource.LiteralSource.Value,
-			}
-		} else if secret.SecretSource.EnvSource != nil {
+		// We don't want to store HashSource and LiteralSource explictly.
```
jlewi: nit: Please expand the comment to explain why we don't serialize it (e.g. because we want the config to be checked into source control and don't want secrets in source control).

Can we add a unittest to? e.g. verify that if hashed source or literalsource is set and we call writeconfig the secret gets stripped out.

bootstrap/pkg/apis/apps/configconverters/v1beta1.go
```diff
@@ -136,16 +137,13 @@ func (v V1beta1) ToKfConfig(kfdefBytes []byte) (*kfconfig.KfConfig, error) {
 		s := kfconfig.Secret{
 			Name: secret.Name,
 		}
-		if secret.SecretSource == nil {
+		// We don't want to store literalSource explictly, becasue we want the config to be checked into source control and don't want secrets in source control.
+		if secret.SecretSource == nil || secret.SecretSource.LiteralSource != nil {
```
jlewi: add || secret.SecretSource.hashedSource != nil

as well because we don't want to serialize hashed sources either

gabrielwen: I think hashed source is deprecated?

zhenghuiwang: HashedSource is not a field in SecretSource. We don't need to check that.
https://github.com/kubeflow/kubeflow/blob/f943808a2805d81c90935f77b89b8bb85e43baa0/bootstrap/pkg/apis/apps/kfdef/v1beta1/application_types.go#L89

zhenghuiwang:
Thanks for the comments! @jlewi @gabrielwen 

Please take another look. 

/cc @kunmingg

bootstrap/pkg/apis/apps/configconverters/v1beta1.go
jlewi: Oh great

bootstrap/pkg/apis/apps/kfconfig/types.go
```diff
@@ -514,23 +516,43 @@ func (c *KfConfig) SyncCache() error {
 	return nil
 }
 
-// GetSecret returns the specified secret or an error if the secret isn't specified.
-func (c *KfConfig) GetSecret(name string) (string, error) {
+// Base64Encrypt encrypts a string to based64 encoding.
+func Base64Encrypt(password string) (string, error) {
```
jlewi: Why do this here instead of gcp.go?  Bcrypt encoding secrets is specific to the basic auth secret.
Now that we got rid of HashedLiteralSource in v1beta1  its no longer a part of the generic API for secrets
https://github.com/kubeflow/kubeflow/blob/f943808a2805d81c90935f77b89b8bb85e43baa0/bootstrap/pkg/apis/apps/kfdef/v1beta1/application_types.go#L89

If we make it a public method people will likely use it leading to problems.

bootstrap/pkg/apis/apps/kfconfig/types.go
```diff
@@ -514,23 +516,43 @@ func (c *KfConfig) SyncCache() error {
...
+			Message: fmt.Sprintf("Error when hashing password: %v", err),
+		}
+	}
+	encodedPassword := base64.StdEncoding.EncodeToString(passwordHash)
+
+	return encodedPassword, nil
+}
+
+// GetEncodedSecret returns the encoded secret value or an error if the secret isn't specified.
+func (c *KfConfig) GetEncodedSecret(name string) (string, error) {
```
jlewi: Per above comment I don't think we want to do this. This will cause problems. Code that is calling GetSecret are expecting the raw secret not a bcrypted value.

zhenghuiwang: I will move the logic to build basic auth secret.

bootstrap/pkg/kfapp/gcp/gcp.go
```diff
@@ -213,7 +212,7 @@ func getSA(name string, nameSuffix string, project string) string {
 func (gcp *Gcp) GetK8sConfig() (*rest.Config, *clientcmdapi.Config) {
 	// TODO(jlewi): Should we unify the code by just setting ts and then calling
 	// ts.Tpken to get a token?
-	accessToken, _ := gcp.kfDef.GetSecret(GcpAccessTokenName)
+	accessToken, _ := gcp.kfDef.GetEncodedSecret(GcpAccessTokenName)
```
jlewi: Revert. We need the raw accessToken not the bcrypted value token. If we use the bcrypted value token it won't work.

jlewi:
@zhenghuiwang I think you want to revert a lot of your most recent changes. 

* GetSecret should return raw values.
* The only place we want to deal with bcrypted values is when creating the K8s secret to store the bcrypted password

https://github.com/kubeflow/kubeflow/blob/f943808a2805d81c90935f77b89b8bb85e43baa0/bootstrap/pkg/kfapp/gcp/gcp.go#L1552

So that code should change to something like

```
rawPassword, err := gcp.kfDef.GetSecret(p.Auth.BasicAuth.Password.Name)
encodedPassword, err := bcryptEncodePassword(rawPassword)
```

Since v1beta1 no longer supports hashedLiteralSource we should be able to assume the password is always encoded but if we wanted to provide backwards compatibility just in case we could instead do someihing like

```
rawPassword, err := gcp.kfDef.GetSecret(p.Auth.BasicAuth.Password.Name)

encodePassword:= rawPassowrd

if  secret_type != HashedLiteralSource { # pseudo code
  encodedPassword, err := bcryptEncodePassword(rawPassword)
}

```

zhenghuiwang:
@jlewi thanks for the detailed guide. I made the change.

jlewi:
/lgtm
/approve

Thanks!

PR #4447, opened 2019-11-01
Endpoint basic auth test

### LINKED ISSUE #4442: [GCP] Endpoint test for basic auth
/kind feature

We need an E2E test that the endpoint for the central dashboard is accessible with basic auth.

This would catch bugs like
kubeflow/manifests#593

It should hopefully be relatively straightforward to modify our existing endpoint ready test.
https://github.com/kubeflow/kubeflow/blob/master/testing/kfctl/endpoint_ready_test.py

gabrielwen:
Related to #4442

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4447)

### DIFF
--- a/testing/gcp_util.py
+++ b/testing/gcp_util.py
@@ -135,7 +135,6 @@ def basic_auth_endpoint_ready(url, username, password, wait_min=15):
     sleep(10)
     num_req += 1
     logging.info("Trying url: %s", url)
-    print("Trying url: %s", url)
     try:
       resp = requests.request(
           "GET",
@@ -143,7 +142,7 @@ def basic_auth_endpoint_ready(url, username, password, wait_min=15):
           verify=False)
       logging.info(resp.text)
       if resp.status_code == 200:
-        logging.info("Basic auth is ready for %s!", url)
+        logging.info("Basic auth landing page is ready for %s !", url)
       else:
         logging.info("%s: basic auth is not ready, request number: %s" % (url, num_req))
     except Exception as e:
--- a/testing/gcp_util.py
+++ b/testing/gcp_util.py
@@ -135,8 +135,17 @@ def basic_auth_endpoint_ready(url, username, password, wait_min=15):
     sleep(10)
     num_req += 1
     logging.info("Trying url: %s", url)
+    print("Trying url: %s", url)
     try:
-      pass
+      resp = requests.request(
+          "GET",
+          url,
+          verify=False)
+      logging.info(resp.text)
+      if resp.status_code == 200:
+        logging.info("Basic auth is ready for %s!", url)
+      else:
+        logging.info("%s: basic auth is not ready, request number: %s" % (url, num_req))
     except Exception as e:
       logging.info("%s: basic auth is not ready, exception caught %s, request number: %s" %
                    (url, str(e), num_req))
--- a/testing/gcp_util.py
+++ b/testing/gcp_util.py
@@ -140,11 +140,11 @@ def basic_auth_endpoint_ready(url, username, password, wait_min=15):
           "GET",
           url,
           verify=False)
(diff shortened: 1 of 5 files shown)


jlewi:
kubeflow/kfctl#69 tracking presubmit failure

testing/kfctl/endpoint_ready_test.py
```diff
@@ -28,11 +28,12 @@ def test_endpoint_is_ready(record_xml_attribute, project, app_name):
   """
   util.set_pytest_junit(record_xml_attribute, "test_endpoint_is_ready")
 
-  # Owned by project kubeflow-ci-deployment.
-  os.environ["CLIENT_ID"] = "29647740582-7meo6c7a9a76jvg54j0g2lv8lrsb4l8g.apps.googleusercontent.com"
-  if not gcp_util.endpoint_is_ready(
-      "https://{}.endpoints.{}.cloud.goog".format(app_name, project),
-      wait_min=25):
+  if not use_basic_auth:
```
jlewi: Don't we need to provide username and password in the case of basic auth?

testing/gcp_util.py
```diff
@@ -100,19 +102,22 @@ def endpoint_is_ready(url, wait_min=15):
       resp = requests.request(
           "GET",
           url,
-          headers={
-              "Authorization":
-              "Bearer {}".format(google_open_id_connect_token)
-          },
-          verify=False)
+          verify=False) if use_basic_auth else requests.request(
```
jlewi: Is there a typo here? This doesn't looks like valid syntax

gabrielwen: yes, it is valid - this is like Python version of syntax sugar.

testing/kfctl/endpoint_ready_test.py
gabrielwen: that depends on whether you'd like to test kflogin with POST.

testing/gcp_util.py
jlewi: Google style guide suggests eschewing complex "list comprehensions". If it doesn't fit on one line its a good indicator its too complicated.

Can we use a standard multi-line if/else statement?

testing/kfctl/endpoint_ready_test.py
jlewi: Can we test that if we send a request with basic auth headers containing username/password that we can access the central dashboard? That would probably be my P0.

Testing kflogin would also be good. Is it possible to do both?

I think you will need to modify kfctl_go_build_deploy.py to save the username and password in the test directory (make sure to save it in a location that doesn't get uploaded to GCS for prow).

gabrielwen: I think you are talking about the same thing, tho.  it's actually sending username/pw to kflogin to get the auth token and then we could access the central dashboard.

testing/gcp_util.py
gabrielwen: done.

testing/kfctl/endpoint_ready_test.py
gabrielwen: updated, but you might wanna see if I write to the correct directory - I write to `app_path`

jlewi: Great thanks! Is the test just checking that we get a 200?That's probably good enough for now but is there someway we could verify we actually get the centraldashboard?

gabrielwen: the content shows up as central dashboard.  but I'm not sure if it's a safe way to find a substr of that as it might be changed.

jlewi:
/lgtm
/approve

Thanks  @gabrielwen lets not close the bug until we have a passing test run.

PR #4456, opened 2019-11-04
pass app_path as args

gabrielwen:
Related: kubeflow/kfctl#71

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4456)

### DIFF
--- a/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
+++ b/py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
@@ -91,7 +91,7 @@ def load_config(config_path):
       config_spec = yaml.load(f)
       return config_spec
 
-def set_env_init_args(use_basic_auth, use_istio):
+def set_env_init_args(use_basic_auth, use_istio, app_path):
   # Is it really needed?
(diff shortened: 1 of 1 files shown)


py/kubeflow/kubeflow/ci/kfctl_go_test_utils.py
```diff
@@ -91,7 +91,7 @@ def load_config(config_path):
       config_spec = yaml.load(f)
       return config_spec
 
-def set_env_init_args(use_basic_auth, use_istio):
+def set_env_init_args(use_basic_auth, use_istio, app_path):
```
jlewi: Why not just take config_path

def set_env_init_args(config_path):
  ...

And then determine the settings from the config path?

gabrielwen: I think this is more about the structure of `kfctl_deploy_kubeflow`?  `set_env_init_args` is only called from `kfctl_deploy_kubeflow` and the function already has some values as parameters e.g. `use_basic_auth` and `use_istio`.

PR #4470, opened 2019-11-05
fix json field names

gabrielwen:
Used the wrong names...

Related: #4442

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4470)


gabrielwen:
@richardsliu mentioned we should write to kfctl repo now.
/close

PR #4475, opened 2019-11-06
catch http connection error when endpoint is not ready

gabrielwen:
Related: #4442

Needs to catch exception.  When endpoint is not ready, http raise exception instead of returning non-200 response.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4475)

### DIFF
--- a/testing/gcp_util.py
+++ b/testing/gcp_util.py
@@ -97,7 +97,6 @@ def iap_is_ready(url, wait_min=15):
   end_time = datetime.datetime.now() + datetime.timedelta(
       minutes=wait_min)
   while datetime.datetime.now() < end_time:
-    sleep(10)
     num_req += 1
     logging.info("Trying url: %s", url)
     try:
@@ -120,6 +119,7 @@ def iap_is_ready(url, wait_min=15):
     except Exception as e:
       logging.info("%s: Endpoint not ready, exception caught %s, request number: %s" %
                    (url, str(e), num_req))
+    sleep(10)
   return False
 
 def basic_auth_is_ready(url, username, password, wait_min=15):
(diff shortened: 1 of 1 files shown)


testing/gcp_util.py
```diff
@@ -128,14 +128,20 @@ def basic_auth_is_ready(url, username, password, wait_min=15):
   end_time = datetime.datetime.now() + datetime.timedelta(
       minutes=wait_min)
   while datetime.datetime.now() < end_time:
-    sleep(2)
+    sleep(10)
```
jlewi: Why is the sleep at the start of loop? Doesn't this introduce an unnecessary delate?

testing/gcp_util.py
```diff
@@ -128,14 +128,20 @@ def basic_auth_is_ready(url, username, password, wait_min=15):
...
-        verify=False)
-    if resp.status_code != 200:
+    logging.info("Trying url: %s request number %s" % (get_url, req_num))
+    resp = None
+    try:
+      resp = requests.request(
+          "GET",
+          get_url,
+          verify=False)
+    except Exception as e:
```
jlewi: Is it possible to catch a more narrow Exception e.g. httpException

testing/gcp_util.py
gabrielwen: this was copied from iap_is_ready.  and I think it's fine because the first we try to connect to the endpoint is almost unlikely to be ready, anyway.

testing/gcp_util.py
gabrielwen: updated.

testing/gcp_util.py
jlewi: Please move the sleep to the end of the loop. I think that makes more sense.

gabrielwen: updated.

PR #4476, opened 2019-11-06
update test config to rm kfctl tests

kunmingg:
Remote tests for kfctl code as now kfctl dev happens in it's own repo.
Next step is delete kfctl code from kubeflow repo

Related: https://github.com/kubeflow/kfctl/issues/79

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4476)


jlewi:
@richardsliu @kunmingg should we add these tests in kubeflow/kfctl first?
This way we don't lose signal?

PR #4479, opened 2019-11-07
Fix var name error

gabrielwen:
Related: #4442 

Sorry the var names are wrong...

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4479)


gabrielwen:
Ignoring postsubmit failure.

PR #4484, opened 2019-11-07
Delete the ksonnet registry; we no longer use ksonnet.

jlewi:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4484)


kubeflow/profiles/tests/profiles_test.py
```diff
@@ -1,29 +1,53 @@
-"""Test Profiles custom resource.
-
-todo
-
+"""Test profiles custom resource.
```
Ark-kun: Is this change relevant to the PR?

PR #4489, opened 2019-11-08
Added website versioning steps to the release procedures

sarahmaddox:
In the release procedures, split out the steps for updating version numbers in master website and creating a new website branch. Added a number of steps to the section on updating version numbers in various files

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4489)

### DIFF
--- a/docs_dev/releasing.md
+++ b/docs_dev/releasing.md
@@ -215,7 +215,7 @@ getting the correct installation instructions etc. Here are the steps to follow:
    * Update the version number. For example, if the upcoming release is 
      Kubeflow v0.7.0:
      ```
-     version_menu = "v0.7.0"
+     v0.7.0
      ```
 
 1. Update **all** the shortcodes that hold the config file names and URIs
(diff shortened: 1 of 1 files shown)


docs_dev/releasing.md
```diff
@@ -160,14 +160,92 @@ Alternatively you can use the UI.
...
+
+We usually start updating the website for the upcoming major/minor release when
+we've released a good RC for that release. At this point, it's useful for
+users to start experimenting with the upcoming release.
+
+When we're ready to start updating the docs for the upcoming version, we
+need to update various version numbers on the website, to ensure users are
+getting the correct installation instructions etc. Here are the steps to follow:
+
+1. Make sure the website branch for the current major/minor version has been
```
jlewi: Its somewhat ambiguous which version current is referring to; i.e. is it the version already released or the version we are releasing.

My suggestion would be in the preamble to say something like suppose we are in the process of releasing 0.Y so the docs are currently serving version 0.X. 

Then you can refer to 0.Y and 0.X below and its clear what you are refering to.

jlewi: Or you can just say "if we are releasing 0.7 then make sure the branch for v0.6 exists" that would be consistent with what you have below.

sarahmaddox: Good point. Done.

PR #4493, opened 2019-11-08
Fix KFCTL e2e test build path

gabrielwen:
Needs to point to kfctl repo instead of using kubeflow/kubeflow/bootstrap

/assign @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4493)


kunmingg:
Shall we migrate this file to kfctl repo?

gabrielwen:
> Shall we migrate this file to kfctl repo?

yes, I think so.  but let's fix the test first.

PR #4518, opened 2019-11-22
Show namespace selector when choosing pipelines

Bobgy:
We are working on supporting multi user mode for pipelines in Kubeflow.
Show the selector so it becomes visible.

I would also want to add a "pipelines multi user" feature flag that can turn this on/off. Can someone point me to the right direction?

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4518)

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -203,7 +203,7 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         case IFRAME_LINK_PREFIX:
             this.page = 'iframe';
             isIframe = true;
-            hideNamespaces = this.subRouteData.path.startsWith('/pipeline');
+            hideNamespaces = false;
             this._setActiveMenuLink(this.subRouteData.path);
             this._setIframeSrc();
             break;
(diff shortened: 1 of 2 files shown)


prodonjs:
/uncc @swiftdiaries 
/cc @avdaredevil 

This looks good, but you'll need to remove the test at  https://github.com/kubeflow/kubeflow/blob/183919dccbe1e5c77aece17fd5064992383dddb6/components/centraldashboard/public/components/main-page_test.js#L152.

In terms of setting the flag, we haven't had a mechanism in place for that. Would you expect the flag to be some kind of metadata or tag stored in the K8s cluster that the dashboard would read? Or would it be something provided in a query string or some other user-provided mechanism?

Bobgy:
> In terms of setting the flag, we haven't had a mechanism in place for that. Would you expect the flag to be some kind of metadata or tag stored in the K8s cluster that the dashboard would read? Or would it be something provided in a query string or some other user-provided mechanism?

I expect the flag to live in a configmap. It can be passed to the dashboard pod as an env flag to turn on this behavior.
(Please don't put much efforts on this, whether we need the feature flag is still being discussed on.)

Bobgy:
Hi, I confirmed we don't need the feature flag. I will fix tests

Bobgy:
@prodonjs Can you review again? thanks!

avdaredevil:
/lgtm

Will let @prodonjs do the `/approve`

Bobgy:
@prodonjs friendly reminder

gaoning777:
Thanks, Yuan for quickly solving this.

PR #4525, opened 2019-11-25
configuring scheduler name from configmap

itayvallach:
Set environment variable SCHEDULER_NAME in"notebook-controller-parameters" ConfigMap and it will be used as the notebook scheduler.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4525)

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -328,6 +328,10 @@ func generateStatefulSet(instance *v1beta1.Notebook) *appsv1.StatefulSet {
 			FSGroup: &fsGroup,
 		}
 	}
+	schedulerName := os.Getenv("SCHEDULER_NAME")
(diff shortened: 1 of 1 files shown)


itayvallach:
@googlebot I signed it!

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -328,6 +328,10 @@ func generateStatefulSet(instance *v1beta1.Notebook) *appsv1.StatefulSet {
 			FSGroup: &fsGroup,
 		}
 	}
+	schedulerName := os.Getenv("SCHEDULER_NAME")
+	if len(schedulerName) > 0 {
+		podSpec.SchedulerName = schedulerName
+	}
```
yeya24: Don' t think this is a good way to configure the scheduler in notebook controller. We should control this through the notebook CR itself

PR #4539, opened 2019-11-28
Add metadata field to PodDefaults

### LINKED ISSUE #4502: admission-controller: allow metadata/annotations in PodDefaults
/kind feature

We we're using kube2iam/kiam to grant IAM pemissions, e.g for accessing s3, to pods and want to grant these permissions to jupyter notebooks. For that we'd need to set the `iam.amazonaws.com/role: argus-default` annotation to the jupyter pods by using PodDefaults

Extend PodDefaults to include metadata.

If this is agreed on, I can implement it. Should be pretty straight forward.

Related: #4315

discordianfish:
This copies labels and annotations from the PodDefault's metadata field
to pods.

This closes #4502

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4539)

### DIFF
--- a/components/admission-webhook/pkg/apis/settings/v1alpha1/poddefault_types.go
+++ b/components/admission-webhook/pkg/apis/settings/v1alpha1/poddefault_types.go
@@ -54,8 +54,11 @@ type PodDefaultSpec struct {
 	// +optional
 	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
 
-	// ObjectMeta defines the metadata to inject labels and annotations from into the metadata field of a pod.
-	metav1.ObjectMeta `json:"metadata,omitempty"`
+	// Annotations defines the annotations to inject into the pod.
+	Annotations map[string]string `json:"annotations,omitempty"`
+
+	// Labels defines the labels to inject into the pod.
+	Labels map[string]string `json:"labels,omitempty"`
 }
 
 // PodDefaultStatus defines the observed state of PodDefault
--- a/components/admission-webhook/pkg/apis/settings/v1alpha1/poddefault_types.go
+++ b/components/admission-webhook/pkg/apis/settings/v1alpha1/poddefault_types.go
@@ -53,6 +53,9 @@ type PodDefaultSpec struct {
 	// VolumeMounts defines the collection of VolumeMount to inject into containers.
 	// +optional
 	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
+
+	// ObjectMeta defines the metadata to inject labels and annotations from into the metadata field of a pod.
+	metav1.ObjectMeta `json:"metadata,omitempty"`
 }
 
 // PodDefaultStatus defines the observed state of PodDefault
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -109,7 +109,18 @@ func safeToApplyPodDefaultsOnPod(pod *corev1.Pod, podDefaults []*settingsapi.Pod
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


components/admission-webhook/pkg/apis/settings/v1alpha1/poddefault_types.go
```diff
@@ -53,6 +53,9 @@ type PodDefaultSpec struct {
 	// VolumeMounts defines the collection of VolumeMount to inject into containers.
 	// +optional
 	VolumeMounts []v1.VolumeMount `json:"volumeMounts,omitempty"`
+
+	// ObjectMeta defines the metadata to inject labels and annotations from into the metadata field of a pod.
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
jlewi: Did you consider having separate fields for labels & annotations rather than a metadata field?
metadata has a bunch of fields
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#objectmeta-v1-meta

If we include ObjectMeta does that create confusion about what happens for other fields (e.g. name, namespace, ownerReferences, finalizers etc...)?

yanniszark: How about introducing the whole field but only use those certain parts?
eg StatefulSet uses PersistentVolumeClaim as the type for volumeClaimTemplates, even if it doesn't use all the fields.

We could also maybe use some of them in the future?

jlewi: The downside of using ObjectMeta is that the API is no longer self-specifying and well defined. A user has to read the docs in order to understand the behavior; i.e. which fields are copied and which are ignored. Whereas if we have fields labels and annotations that are copied then it is arguably more self explanatory.

Furthermore we are in effect saying that the behavior for other fields e.g "finalizers, name, namespace, ownerReferences" are undefined. Its not clear whether these fields are intentionally excluded or they are excluded only because we did a partial implementation in the need of expediency.

Do we have a strong reason to expect that we will want to include these fields in the future? finalizers seems like the only other field that we might need to set in the future. Even then I'm not sure if setting finalizers on pods as opposed to the owning object is common.

For all the other fields (name, uid, ownerReferences) etc... setting them in podDefaults seems incorrect.

discordianfish: Yeah I wasn't sure either about using ObjectMeta but with only the annotations/labels fields (what I'm doing here and what @yanniszark describes unless I'm mistaken), or add annotations/labels as separate fields. I've went with using ObjectMeta because having a struct for metadata as oppose to just two map[string]string seemed nicer, but not feeling strongly that way.
I'll update the PR to use Annotations and Labels fields instead.

jlewi: Thanks. I think this is better.

discordianfish:
@jlewi Was about to test this on my cluster. Looks like something isn't working yet. Given this poddefault:
```
kind: PodDefault
metadata:
  name: add-xx-role
spec:
  selector:
   matchLabels:
     add-xx-role: "true"
  desc: "Assume xx-default role"
  annotations:
    iam.amazonaws.com/role: xx-default
```
The pod matches according to the admission controller log:
```
I1203 17:22:02.637042       1 main.go:467] applied poddefaults: add-xx-role successfully on Pod: test-0
```
but the pod doesn't have the annotation. I figured it might be due to the fields not being in the CRD, so I've added them (like this: https://github.com/discordianfish/kubeflow-manifests/commit/89e8cac0a63af034d4bc9499300f4d17ecf543f0) but still same result.

Maybe it's getting overwritten somewhere later in code?

PR #4544, opened 2019-12-04
Fix annotation/label injection in admission hook

discordianfish:
Fix for my broken implementation in #4539 /cc @jlewi 

Before we used the annotations/labels from PodDefaults's embedded
ObjectMeta instead of the correct Annotations/Labels field of the
PodDefaultSpec.

This also adds tests for the (SafeTo)ApplyPodDefaultsOnPod function to
that uncovered the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4544)

### DIFF
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -114,8 +114,8 @@ func safeToApplyPodDefaultsOnPod(pod *corev1.Pod, podDefaults []*settingsapi.Pod
 		defaultLabels      = make([]*map[string]string, len(podDefaults))
 	)
 	for i, pd := range podDefaults {
-		defaultAnnotations[i] = &pd.Annotations
-		defaultLabels[i] = &pd.Labels
+		defaultAnnotations[i] = &pd.Spec.Annotations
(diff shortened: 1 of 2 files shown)


discordianfish:
Btw, tested this in our setup and works as expected.

jlewi:
/lgtm
/approve

@discordianfish the prow dashboard is what I use (and what I recommend others in Kubeflow use) to identify PRs needing attention

https://k8s-gubernator.appspot.com/pr

You can enter a users GitHub id to see what they see; e.g. 
https://k8s-gubernator.appspot.com/pr/jlewi

If you looked at the PR dashboard you would have seen that your PR wasn't showing up for me as needing as attention. Which is why it got overlooked.

The reason is prow uses a simple turn based heuristic. So once I commented on the PR to run the tests prow said it no longer needed my attention. 

But if you looked at your dashboard it was showing up for you; i.e. the ball was in your court.

So if you simply responded on the pr with something like @jlewi hey this is ready for another look. that would have caused prow to move it back up to my needs attention.

PR #4556, opened 2019-12-09
jupyter-web-app: Use SubjectAccessReviews instead of KFAM

### LINKED ISSUE #2981: Notebook Spawner should check for edit permissions before spawning the notebook

kimwnasptd:
closes #2981 

With this PR Jupyter Web App will use k8s `SubjectAccessReviews` to check whether a user is authorized to perform an action on the k8s api server.

This PR should be merged after [manifests/655](https://github.com/kubeflow/manifests/pull/655) is merged, which gives permissions to JWA's ServiceAccount to use SubjectAccessReviews.

/cc @yanniszark 
/cc @kunmingg 
/cc @jlewi 
/assign @kimwnasptd

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4556)

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/api.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/api.py
@@ -77,8 +77,8 @@ def wrap(fn, *args, **kwargs):
 
 # API Functions
 # GETers
-@auth.needs_authorization("get", "core", "v1", "pvcs")
-def get_pvcs(namespace):
+@auth.needs_authorization("list", "core", "v1", "pvcs")
+def list_pvcs(namespace):
     return wrap_resp(
         "pvcs",
         v1_core.list_namespaced_persistent_volume_claim,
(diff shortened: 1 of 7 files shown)


components/jupyter-web-app/backend/kubeflow_jupyter/common/api.py
```diff
@@ -111,149 +77,95 @@ def wrap(fn, *args, **kwargs):
...
-def get_secret(ns, nm, user=None):
-    if not is_authorized(user, ns):
-        return {
-            "success": False,
-            "log": "User '{}' is not authorized for namespace '{}'".format(
-                user, ns
-            )
-        }
-
+# @auth.needs_authorization("list", "storage.k8s.io", "v1", "storageclasses")
```
kunmingg: uncomment?

kimwnasptd: Currently users added with the Profile flow (either owner or edit/view) will get a `RoleBinding` that will give them `kubeflow-admin` permissions on this namespace.

But, in order to list StorageClasses the user will need to have a `ClusterRoleBinding` which is not currently supported.  To keep the current UX I let JWA to use its own credentials for this call. I will open an issue for us to keep track of this detail and discuss whether it makes sense to change the Profiles mechanism to support this case.

Also, as a side note, JWA exposes only one endpoint `/api/storageclasses/default` which will use this function in order to determine if a default storageclass is set. So currently an unauthorized user will not be able to use this function to list storageclasses.

jlewi:
/assign @kunmingg 
/assign @lluunn 

@kunmingg and @lluunn could one of you take the lead on reviewing this PR?

kimwnasptd:
@lluunn @kunmingg kind reminder for this. 

The [PR](https://github.com/kubeflow/manifests/pull/655) for updating the manifests to give permissions to JWA for `SubjectAccessReviews` has also been merged.

PR #4577, opened 2019-12-16
delete kubeflow/kubeflow/py

gabrielwen:
Related: kubeflow/kfctl#77

We've migrated codes to the kfctl repo, deleting the codes here and make kfctl repo source of truth.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4577)


jlewi:
@gabrielwen I think you were a little zealous.

It doesn't look like the following scripts migrated to kubeflow/kfctl so should we keep them around
* rebuild_manifests.sh
* application_util.py
* update_jupyter_web_app.py
* test_data (used by application_util_test.py)

jlewi:
I think only the files named "kfctl_*" need to be deleted?

gabrielwen:
@jlewi updated.  yea on second thought you are right.

Issue #4582, opened 2019-12-16
notebook controller and tensorboard controller doesn't build


lluunn:
/kind bug

**What steps did you take and what happened:**
both cannot find `github.com/kubeflow/kubeflow/components/common/reconcilehelper`
Should be related to https://github.com/kubeflow/kubeflow/pull/4312

https://github.com/kubeflow/kubeflow/pull/4214#issuecomment-566035507

cc @kimwnasptd @abhi-g

lluunn:
cc @quanjielin

jlewi:
It looks like #4631 might not have fixed the build when using Kaniko. When I run in Kaniko in the CD pipelines I get

```
INFO[0027] Using files from context: [/workspace/kubeflow/components/notebook-controller/go.sum] 
INFO[0027] COPY go.sum go.sum                           
INFO[0027] Taking snapshot of files...                  
INFO[0027] RUN go mod download                          
INFO[0027] cmd: /bin/sh                                 
INFO[0027] args: [-c go mod download]                   
go: parsing /common/go.mod: open /common/go.mod: no such file or directory
go: finding github.com/prometheus/client_golang v0.9.0
go: finding github.com/go-logr/logr v0.1.0
go: finding sigs.k8s.io/controller-runtime v0.2.0
go: finding k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
go: finding sigs.k8s.io/controller-tools v0.2.0
go: finding k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
go: finding k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
go: error loading module requirements
error building image: error building stage: waiting for process to exit: exit status 1
```

I suspect the problem now is that we aren't uploading "/common" as part of the kaniko context.

kunmingg:
@jlewi 
can we close?

jlewi:
@kunmingg yes

PR #4312, opened 2019-10-16
initial checkin of tensorboard controller

quanjielin:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4312)

PR #4583, opened 2019-12-16
fix notebook image build

lluunn:
With the change, I can build the notebook image.
The new one is v20191216-v0.7.0-rc.5-61-g2da14890-dirty-d0bbf2, will update the manifest.

/cc @kimwnasptd @fediazgon @jlewi 
related to https://github.com/kubeflow/kubeflow/issues/4582

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4583)

### DIFF
--- a/components/notebook-controller/go.mod
+++ b/components/notebook-controller/go.mod
@@ -4,19 +4,11 @@ go 1.12
 
 require (
 	github.com/go-logr/logr v0.1.0
-	github.com/gogo/protobuf v1.2.1 // indirect
-	github.com/json-iterator/go v1.1.6 // indirect
-	github.com/modern-go/reflect2 v1.0.1 // indirect
-	github.com/onsi/ginkgo v1.8.0 // indirect
-	github.com/onsi/gomega v1.5.0 // indirect
+	github.com/kubeflow/kubeflow v0.7.0-rc.5.0.20191205174032-2da148904248
 	github.com/prometheus/client_golang v0.9.0
-	github.com/spf13/pflag v1.0.3 // indirect
-	golang.org/x/net v0.0.0-20190501004415-9ce7a6920f09 // indirect
-	golang.org/x/sys v0.0.0-20190429190828-d89cdac9e872 // indirect
-	golang.org/x/text v0.3.2 // indirect
-	gopkg.in/yaml.v2 v2.2.2 // indirect
 	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
 	k8s.io/client-go v11.0.1-0.20190409021438-1a26190bd76a+incompatible
 	sigs.k8s.io/controller-runtime v0.2.0
+	sigs.k8s.io/controller-tools v0.2.0 // indirect
 )
--- a/components/notebook-controller/go.mod
+++ b/components/notebook-controller/go.mod
@@ -4,7 +4,6 @@ go 1.12
 
 require (
 	github.com/go-logr/logr v0.1.0
-	github.com/kubeflow/kubeflow v0.7.0-rc.5.0.20191205174032-2da148904248
 	github.com/prometheus/client_golang v0.9.0
 	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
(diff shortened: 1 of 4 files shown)


fediazgon:
Just tried locally and works fine for me.

components/notebook-controller/go.mod
```diff
@@ -4,19 +4,11 @@ go 1.12
 
 require (
 	github.com/go-logr/logr v0.1.0
-	github.com/gogo/protobuf v1.2.1 // indirect
-	github.com/json-iterator/go v1.1.6 // indirect
-	github.com/modern-go/reflect2 v1.0.1 // indirect
-	github.com/onsi/ginkgo v1.8.0 // indirect
-	github.com/onsi/gomega v1.5.0 // indirect
+	github.com/kubeflow/kubeflow v0.7.0-rc.5.0.20191205174032-2da148904248
```
jlewi: Is this pinning to a specific version of kubeflow/kubeflow? Do we want to do that? 
Why is github.com/kubeflow/kubeflow listed go.mod? Isn't github.com/kubeflow/kubeflow the module we are building?

components/notebook-controller/go.sum
```diff
@@ -38,6 +44,13 @@ github.com/kisielk/gotool v1.0.0/go.mod h1:XhKaO+MFFWcvkIS/tQcRk01m1F5IRFswLeQ+o
 github.com/kr/pretty v0.1.0/go.mod h1:dAy3ld7l9f0ibDNOQOHHMYYIIbhfbHSm3C4ZsoJORNo=
 github.com/kr/pty v1.1.1/go.mod h1:pFQYn66WHrOpPYNljwOMqo10TkYh1fy3cYio2l3bCsQ=
 github.com/kr/text v0.1.0/go.mod h1:4Jbv+DJW3UT/LiOwJeYQe1efqtUx/iVham/4vfdArNI=
+github.com/kubeflow/kubeflow v0.7.0-rc.5.0.20191205174032-2da148904248 h1:VrZRKsJuXQ9Wsxmk6Sjcj+ZB74g4FchomlJyHTdCfWk=
```
jlewi: Same question as above

jlewi:
@lluunn Could you explain what the problem is and how this fixes it? Do you know why this wasn't caught by tests? Is our postsubmit and periodic tests red?

jlewi:
@kkasravi could you please take a look at the use of go.mod here and provide some guidance.

It looks like we have multiple modules in kubeflow/kubeflow (e.g. notebook-controller) is one that depends on other modules (e.g. common/reconcile-helper).

When building notebook-controller we'd like it to depend on the other modules checked out in the same repo.

Does that seem right? And if so how do we do that?

lluunn:
The problem is if you try `make build-gcr` from master, it doesn't work. I think presubmit doesn't build the image.
The error was that it cannot load https://github.com/kubeflow/kubeflow/tree/master/components/common/reconcilehelper, which is moved there from notebook controller in https://github.com/kubeflow/kubeflow/pull/4312.

I tried the suggestion in https://github.com/kubeflow/kubeflow/pull/4214#issuecomment-566215307 to do `go mod edit -require github.com/kubeflow/kubeflow@master`. Not sure if it's the best way forward.

jlewi:
@lluunn could you open up a bug about the jupyter E2E test? I would expect that to build the image from head and test it.

It looks to me though like the way you are fixing the bug is pinning to a specific commit of kubeflow/kubeflow to find reconcilehelper

```
require (
  github.com/kubeflow/kubeflow v0.7.0-rc.5.0.20191205174032-2da148904248
)
```

I would expect that means we don't pick up any local changes to kubeflow/kubeflow when building the image. Is that what we want? e.g. if we were to end up making a fix to reconcilehelper would the fix get picked up?

I guess a smoke test would be to add a syntax error in reconcilehelper and see if it still builds. If it does then that would confirm you aren't picking up locally modified code.

swiftdiaries:
One possible approach:
We could move the [reconcilehelper util](https://github.com/kubeflow/kubeflow/blob/master/components/common/reconcilehelper/util.go) into either the notebook-controller or the tensorboard-controller and have the other one share the implementation.

lluunn:
I moved back the reconcile util.

jlewi:
If we think moving the reconcileHelper into the notebook package is the best solution then we can go with that.

However, if IIUC we should be able to make the original package layout work just by setting up go.mod correctly. 

See the second stackoverflow answer here
https://stackoverflow.com/questions/52026284/accessing-local-packages-within-a-go-module-go-1-11

Do we just need to add a replace directive to point the module at a relative path?

See also this doc
https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive

One of the main use cases for the replace directive is a multi-module project and it looks like that might be what we have here.

jlewi:
Closing in favor of kubeflow/kubeflow#4631 which updates go.mod

PR #4614, opened 2020-01-02
Jupyter-Web-App: Support for selecting GPUs

### LINKED ISSUE #4288: Jupyter Web App Better UI for Selecting GPUs
/kind feature

Right now to add GPUs users have to add a free form string like "{ "nvidia.com/gpu": 1}" in the extra resources field.

This isn't very intuitive. It would be better if we could have a form that would just allow a user to enter the number of GPUs.

/cc @kimwnasptd

kimwnasptd:
closes #4288 

Instead of having an `Extra Resources` field we allow the user to select the number and vendor of GPUs for a Notebook.

gpus

/assign @kimwnasptd
/cc @lluunn 
/cc @avdaredevil

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4614)

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -387,6 +387,59 @@ def set_notebook_memory(notebook, body, defaults):
     container["resources"]["requests"]["memory"] = memory
 
 
+def set_notebook_gpus(notebook, body, defaults):
+    gpus = None
+    gpuDefaults = defaults.get("gpus", {})
+    if gpuDefaults.get("readOnly", False):
+        # The server should not allow the user to set the GPUs
+        # if the config's value is readOnly. Use the config's value
+        gpus = gpuDefaults["value"]
+        logger.info(f"Using default GPU config: {gpus}")
+
+    elif "gpus" not in body:
+        # Try to load the default values. If they don't exist, don't use GPUs
+        if "gpus" not in defaults:
+            logger.info("No 'gpus' value in neither the form's body nor in"
+                        " the default config's values. Will not use any GPUs")
+            return
+        else:
+            gpus = gpuDefaults["value"]
+            logger.info(f"Using default GPU config: {gpus}")
+
+    else:
+        # Make sure the GPUs value in the request is properly formatted
+        gpus = body["gpus"]
+        logger.info(f"Using form's GPUs: {gpus}")
+
+        if "num" not in gpus:
+            logger.error("'gpus' must have a 'num' field")
+            return
+
+        if gpus["num"] != "none" and "vendor" not in gpus:
+            logger.error("'gpus' must have a 'vendor' field")
+            return
+
+        if gpus["num"] != "none":
+            try:
+                int(gpus["num"])
+            except ValueError:
+                logger.error(f"gpus.num is not a valid number: {gpus['num']}")
+                return
+
+    # Add the GPU annotation
+    if gpus["num"] == "none":
+        return
+
+    container = notebook["spec"]["template"]["spec"]["containers"][0]
+    vendor = gpus["vendor"]
+    num = int(gpus["num"])
+
+    limits = container["resources"].get("limits", {})
+    limits[f"{vendor}.com/gpu"] = num
+
+    container["resources"]["limits"] = limits
+
+
 def set_notebook_configurations(notebook, body, defaults):
     notebook_labels = notebook["metadata"]["labels"]
 
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -1,19 +1,20 @@
-import datetime
+import datetime as dt
+import json
 import logging
 import os
 import sys
+
 import yaml
-import json
 from collections import defaultdict
 from flask import request
 from kubernetes import client
-import datetime as dt
+
 from . import api
 
 # The backend will send the first config it will successfully load
 CONFIGS = [
     "/etc/config/spawner_ui_config.yaml",
-    "./kubeflow_jupyter/common/yaml/spawner_ui_config.yaml"
+    "./kubeflow_jupyter/common/yaml/spawner_ui_config.yaml",
 ]
 
 # The values of the headers to look for the User info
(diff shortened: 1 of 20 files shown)


kimwnasptd:
This PR is XXL because I formatted the entire frontend code with prettier in the first commit.

discordianfish:
We're using PodDefaults for this. I wonder where to draw the line between things that should be configured by PodDefaults vs what should be exposed directly in the UI.

I feel like GPUs are important/generic enough to warrant their own UI options and the PodDefaults afaik don't support customization (e.g request n GPUs), but maybe that's an issue on it's own.

discordianfish:
Oh and this here wouldn't be sufficient for us in the future, since we'd need to add tolerations for the taints we'll have on our GPU nodes

jlewi:
@kimwnasptd I'd highly recommend doing the PR formatting in its own PR so that its easier to review the changes specific to this PR and if necessary revert them.

jlewi:
@kimwnasptd Is a drop down for number of GPUs better then just letting the user enter an integer?

How is the GPU vendor populated? I suspect this could be different on various platforms? Does the vendor correspond to the key in the limits field of the pod spec?

@discordianfish yes it might be worth extending PodDefaults to allow injecting resources. One reason I could forsee doing this would be in the case clusters have multiple types of GPUs. On GKE if you have multiple types of GPUs we would also need to add a node selector (https://github.com/kubeflow/manifests/issues/375). So in this case, I could imagine defining different PodDefaults corresponding to different resource classes. In the UI the user could then select which class of resources to use

Also we might want to look into replacing our PodDefaults with [PodPreset](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.17/#podpreset-v1alpha1-settings-k8s-io)

discordianfish:
@jlewi Oh I didn't realize that resources aren't supported in PodDefaults.. And they aren't supported in PodPresets either.

I also just came across https://github.com/kubernetes/kubernetes/issues/69721#issuecomment-455677705:
> BTW, PodPreset is not actively being worked on AFAIK, and may eventually be converted to an extension or deprecated entirely.

So maybe this warrants a new strategy how to handle site specific notebook configuration.

kimwnasptd:
@jlewi ACK for the formatting.

Regarding the drop-down I saw it on GCP when creating a VM with GPUs. I don't have any hard preference on this, I'll make it a number input instead.

The vendor field had two hardcoded values AMD and INTEL. I just took a look on k8s [Device Plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/) and realized that this field will need to at least be more configurable wrt the values it contains. I'll add code to allow the admin to set in JWA's config what values they want the drop-down to show and on what vendor value each one will correspond to.

And yes, the vendor key corresponds to the key in the limits field of the pod spec.

jlewi:
Thanks @kimwnasptd That sounds good to me.

I don't have a strong oppinion about whether GPUs should be an integer or a drop down.

> Regarding the drop-down I saw it on GCP when creating a VM with GPUs. I don't have any hard preference on this, I'll make it a number input instead.

I think you are right and it probably makes sense to follow what they did.

My guess is the reason its a drop down is to constrain the number of GPUs and prevent invalid input.

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
```diff
@@ -387,6 +387,61 @@ def set_notebook_memory(notebook, body, defaults):
     container["resources"]["requests"]["memory"] = memory
 
 
+def set_notebook_gpus(notebook, body, defaults):
+    gpus = None
+    gpuDefaults = defaults.get("gpus", {})
+    if gpuDefaults.get("readOnly", False):
```
jlewi: Does Kubernetes not support any form of introspection that would allow us to avoid explicitly configuring this?

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
```diff
@@ -387,6 +387,61 @@ def set_notebook_memory(notebook, body, defaults):
...
+        # The server should not allow the user to set the GPUs
+        # if the config's value is readOnly. Use the config's value
+        gpus = gpuDefaults["value"]
+        logger.info(f"Using default GPU config: {gpus}")
+
+    elif "gpus" not in body:
+        # Try to load the default values. If they don't exist, don't use GPUs
+        if "gpus" not in defaults:
+            logger.info(
+                "No 'gpus' value in neither the form's body nor in"
```
jlewi: nit: "in neither" -> "in either the form's body or in the default config's values"

jlewi:
@kimwnasptd one nitpicky comment otherwise looks good.

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
kimwnasptd: The `readOnly` field in JWA's config is for the admin to specify that the users should not be able to edit the values of that field in the UI's form. At this point we make sure to use the default value in the config if the admin has set it to be `readOnly`.

It's not clear to me how we could leverage k8s for this one. Was it a miss interpretation on what this small section of code was doing or am I missing something?

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
kimwnasptd: Thanks, done

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
jlewi: @kimwnasptd apologies I think I commented on the wrong line. I meant could we not use introspection to some how get the names for the limit values e.g. "nvidia.com"?

jlewi:
/lgtm
/approve

/hold

@kimwnasptd you can go ahead and remove the hold to merge this. I just put a hold so you could see my response to your comment.

components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
kimwnasptd: Oh I see what you mean. We could maybe check the `node.capacity` or `node.allocatable` to determine the extra available resources on the nodes. But we will need to define how to filter what devices we want to show in JWA.

I also think that this will be the next improvement of this feature but I'd like to dig a little bit more into it first to assess our options.

kimwnasptd:
@jlewi thanks, lets open another issue to track what alternatives we have for auto-detecting the available devices and how we could further improve the GPU selection in JWA.

/hold cancel

Issue #4619, opened 2020-01-06
Add licenses for dependencies in Profile controller Images


jlewi:
/kind feature

Related to https://github.com/kubeflow/testing/issues/539

Need to add all license content for thrid-party dependencies in profile controller image.

We might want to switch to a distroless base image as well.
https://github.com/GoogleContainerTools/distroless/blob/master/examples/go/Dockerfile

jlewi:
@kunmingg Could you take care of this please?

PR #4622, opened 2020-01-06
generate license for profile controller image

kunmingg:
add third party dependency license for profile controller image following:
https://github.com/kubeflow/metadata/tree/master/third_party_licenses

fix https://github.com/kubeflow/kubeflow/issues/4619

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4622)


kunmingg:
@zhenghuiwang 
Any requirement for license file format or location in our image?

abhi-g:
Generally License file goes in the top directory, for eg. where README.md
might be. Also, common practice is to name is LICENSE.TXT

zhenghuiwang:
/lgtm

I don't think there are requirements on where to put all dependency licenses, but I see others usually put them into `third_party/licenses.txt` as you do.

PR #4625, opened 2020-01-07
generate license for kfam image; switch to a distroless base image

kunmingg:
kfam image:
generate dependency license 
switch to a distroless base image

related: https://github.com/kubeflow/kubeflow/issues/4619

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4625)


components/access-management/Dockerfile
```diff
@@ -1,19 +1,20 @@
...
 COPY . .
 RUN go mod download
-RUN go build -gcflags 'all=-N -l' -o /opt/kubeflow/access-management main.go
+RUN go build -gcflags 'all=-N -l' -o access-management main.go
+RUN chmod a+rx access-management
 
-RUN chmod a+rx /opt/kubeflow/access-management
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:latest as serve
```
jlewi: Why static? It looks like the goland example is using base
https://github.com/GoogleContainerTools/distroless/blob/master/examples/go/Dockerfile

kunmingg: updated

Issue #4551, opened 2019-12-06
Notebook images referencing out of date Fairing


joeliedtke:
/kind bug

**What steps did you take and what happened:**

1. Create a Kubeflow Notebooks instance from the TF 1.14 CPU image or TF 2.0 CPU image (have not tested with GPU)
1. Create a Python3 notebook
1. Run the following commands in the notebook:

   * from kubeflow import fairing: Result, ModuleNotFoundError
   * import fairing: Result, loaded
   * !pip show fairing: Result, fairing 0.5 is installed in the default notebook image

**What did you expect to happen:**

Current version of Fairing should be installed in the notebook image.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: 0.7.0
- kfctl version: (use `kfctl version`):
- Kubernetes platform: GKE
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

joeliedtke:
It looks like the following line needs to be either be updated with each release, or be updated to install Fairing from PyPi.

https://github.com/kubeflow/kubeflow/blob/b813c9afaf2b053928935ba8ab8665a261591517/components/tensorflow-notebook-image/Dockerfile#L107

Issue #1428, opened 2018-08-25
Any reason to use ksonnet over skaffold in Quick Start


JCzz:
Hi

Any reason for using ksonnet over skaffold in you "Quick Start" https://www.kubeflow.org/docs/started/getting-started/

Thanks

ai4life:
Why use ksonnet vs skaffold, as well as draft, gitkube, helm, metaparticle, forge.sh, psykube?

JCzz:
Sorry for not making it clearer; I am learning the tools around Kubernetes, so I'm curious to see if there is any difference in using the Ksonnet or Skaffold, as Ksonnet is described as a requirement? 

```
Requirements:
ksonnet version 0.11.0.
```

PR #381, opened 2018-03-07
Fix tfjob dashboard  ambassador route

wbuchwalter:
Needed for: https://github.com/kubeflow/tf-operator/pull/441

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/381)

PR #4645, opened 2020-01-12
Updates to install Fairing from PyPI

joeliedtke:
Updates the Dockerfile for the TF notebook images to install Fairing from PyPI, instead of installing Fairing v0.5 from a GitHub commit.

This change fixes [Kubeflow/Kubeflow #4551](https://github.com/kubeflow/kubeflow/issues/4551).

This change should help fix:
-  [Kubeflow/Website #1428](https://github.com/kubeflow/website/issues/1428)
-  [Kubeflow/Fairing #381](https://github.com/kubeflow/fairing/issues/381): This issue describes that they couldn't run the Kubeflow Notebook example because the notebook was running an old release of Fairing.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4645)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -104,7 +104,7 @@ RUN pip3 --no-cache-dir install \
     jupyter-console==6.0.0 \
     jupyterlab \
     xgboost \
-    git+https://github.com/kubeflow/fairing@a9bb9d5cc1c9f1d75efa31198ddbdccfe176b7bf
+    kubeflow-fairing
 
 COPY --chown=jovyan:users requirements.txt /tmp
 
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -104,7 +104,7 @@ RUN pip3 --no-cache-dir install \
     jupyter-console==6.0.0 \
     jupyterlab \
     xgboost \
-    kubeflow-fairing
+    kubeflow-fairing==0.7.1
 
 COPY --chown=jovyan:users requirements.txt /tmp
 


components/tensorflow-notebook-image/Dockerfile
```diff
@@ -104,7 +104,7 @@ RUN pip3 --no-cache-dir install \
     jupyter-console==6.0.0 \
     jupyterlab \
     xgboost \
-    git+https://github.com/kubeflow/fairing@a9bb9d5cc1c9f1d75efa31198ddbdccfe176b7bf
+    kubeflow-fairing
```
abhi-g: should we pin this to the latest version i.e. 0.7.1?

joeliedtke: That's a good question.

@abhi-g and @jinchihe, What are your preferences? 

1. In each Kubeflow release a Fairing maintainer could pin the version that they prefer to the release. Each Kubeflow release will require an update to this Dockerfile.
2. The notebook images could pick up the most recent version of Fairing at the time of release.  Depending on the order of operations, the most recent version of Fairing may not be included in the notebook image. However, no changes are required to this Dockerfile for each release.

abhi-g: I prefer to pin to 0.7.1 because we know for now its tested with the rest of the KF components for training and serving. It does require some work to remember to update the version when releasing Fairing but thats trivial. 

If in the next few releases we get the indication that there is continuous activity and Fairing is keeping pace with rest of KF, we can choose to switch to install the latest release.

joeliedtke: Ok, I've added a commit to pin kubeflow-fairing to 0.7.1. I've built the image and tested it in KF 0.7.

PTAL

jinchihe:
/lgtm

Thanks @joeliedtke

edi-bice-by:
Installed KF 1.1 and launched a TF 2.1 CPU notebook server.

From what I see from the terminal things have not changed (probably a regression)

pip show fairing
Name: fairing
Version: 0.5
Summary: Python SDK for building, training, and deploying ML models
Home-page: https://github.com/kubeflow/fairing
Author: Kubeflow Authors
Author-email: None
License: UNKNOWN
Location: /usr/local/lib/python3.6/dist-packages
Requires: oauth2client, requests, future, cloudpickle, notebook, google-cloud-storage, google-api-python-client, httplib2,google-auth, setuptools, docker, kubernetes, six, numpy, tornado
Required-by:

$ pip show kubeflow-fairing
WARNING: Package(s) not found: kubeflow-fairing

PR #4659, opened 2020-01-17
[CentralDashboard] Namespace defaulting and validation

### LINKED ISSUE #3863: [CentralDashboard v2] Default namespace to OwnedNamespace or Kubeflow
## About:
`Select namespace` as a state makes little sense as it:
- Causes unnecessary User Interactions on the dashboard
- Leads to an uneven UX for sub-apps that don't handle a non-selected namespace

### Meta:
/kind feature
/priority p1
/area centraldashboard
/area front-end
/assign @avdaredevil 
/cc @kimwnasptd @prodonjs

avdaredevil:
## About
This PR introduces a validation layer into the namespace selector (anytime namespaces or selected is changed):
- Defaulting logic added for Namespace selector:
  - If namespace selected is valid, keep
  - If there is an owned namespace, use that
  - If not, go for `kubeflow` if available
  - If not, leave unselected
- Invalidate a namespace that's not part of `namespaces` and use the logic defined above

### Meta
/area centraldashboard
/area front-end
/priority p1
/assign @avdaredevil
/cc @kwasi @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4659)

### DIFF
--- a/components/centraldashboard/public/components/namespace-selector.js
+++ b/components/centraldashboard/public/components/namespace-selector.js
@@ -161,12 +161,9 @@ export class NamespaceSelector extends PolymerElement {
         const {namespaces} = this;
         if (!namespaces) return;
         const nsSet = new Set(namespaces.map((i) => i.namespace));
-        const owned = namespaces.find((n) => n.role == 'owner');
-
         if (nsSet.has(this.selected)) return;
 
-        // eslint-disable-next-line
-        console.log({own: (owned||[]).namespace, namespaces, ns: this.selected, yeah: namespaces.includes('kubeflow')});
+        const owned = namespaces.find((n) => n.role == 'owner');
         this.selected = (owned && owned.namespace)
             || (nsSet.has('kubeflow')
                 ? 'kubeflow'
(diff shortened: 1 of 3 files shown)


components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -153,6 +154,24 @@ export class NamespaceSelector extends PolymerElement {
...
+     */
+    validate() {
+        const {namespaces} = this;
+        if (!namespaces) return;
+        const nsSet = new Set(namespaces.map((i) => i.namespace));
+        if (nsSet.has(this.selected)) return;
+
+        const owned = namespaces.find((n) => n.role == 'owner');
+        // eslint-disable-next-line
+        console.log({own: (owned||[]).namespace, namespaces, ns: this.selected, yeah: namespaces.includes('kubeflow')});
```
prodonjs: I don't think there are any other `console.log` statements for non-error conditions so I'd suggest removing ithis.

avdaredevil: Good catch, I forgot to remove this line. Thanks!

avdaredevil: Good catch, I forgot to remove this before submitting the PR. Thanks!

PR #4662, opened 2020-01-17
[CentralDashboard] Registration flow onError delay removed

avdaredevil:
## About
There should be no delay in registration flow retry if an error comes back early. This PR fixes that

### Meta
/area centraldashboard
/area front-end
/priority p1
/assign @avdaredevil
/cc @kwasi @prodonjs

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4662)


avdaredevil:
@jlewi Could we get this merged?

Issue #4071, opened 2019-09-04
Profile & Kfam: Version stability


kunmingg:
To guarantee backward compatibility:
- Profile CRD should be locked in v0.7.0
- Kfam API should be locked in v0.7.0

----

Will create v1 API which will be same as v1beta1 and achieve compatibility for free.

jlewi:
@kunmingg Is there more work to be done here? 

It looks like we have a v1beta1.
https://github.com/kubeflow/kubeflow/tree/master/components/profile-controller/api

I'm guessing we need a v1 for 1.0. Do we have an issue for that?

kunmingg:
Issue description updated.
Will create v1 API

jlewi:
@kunmingg Looks like #4663 was merged. What is the remaining work to close this out?

kunmingg:
This one can be closed

Issue #4070, opened 2019-09-04
Profile & Kfam: Upgradeability


kunmingg:
Profile & Kfam: Upgradeability

jlewi:
Bump to P0 because this is required for 1.0.

@kunmingg is there any work to be done? Should this just work out of the box?

kunmingg:
Will test out once https://github.com/kubeflow/kubeflow/issues/4071 is closed

kunmingg:
Verified

PR #4663, opened 2020-01-17
update profile api version to v1

kunmingg:
* create profile API version v1 (which has same spec as v1beta1)
* profile controller will use v1 as storage version but serve both profile v1 and v1beta1

profile upgrade from v1beta1 to v1 on existing cluster:
* apply v1 CRD and controller manifests
* existing v1beta1 profiles will continue to work with new controller.
  

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4663)


kunmingg:
related:
https://github.com/kubeflow/kubeflow/issues/4071

kunmingg:
related:
https://github.com/kubeflow/kubeflow/issues/4070

PR #4664, opened 2020-01-19
Fix docker builds of notebook and tensorboard controller

jlewi:
* The notebook-controllers and tensorboard-controllers now depend on
  the go package components/common

* We need to rewrite the Dockerfiles so that the context is now

  ${KUBEfLOW_REPO}/common

  * so that components/common can be included in the context and copied
    to the Dockerfile

* Create skaffold configs to make it easier to do remote builds with Kaniko

  * The skaffold configs are currently written assuming the kubeflow-ci cluster
    is used to build the images. This could be generalized in the future.

* Remove the code to build the notebook-controller with GCB; we can just
  use skaffold and kaniko to do efficient remote builds.

* Related to #4582 - Jupyter image doesn't build.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4664)


kimwnasptd:
@jlewi when running the `make docker-build` to create a Docker image for the notebooks controller locally I bumped into the following two issues:
1. I get an error when the `check-license` rule runs:
```
< sigs.k8s.io/controller-tools
+ echo 'Please update the license file for changed dependencies.'
Please update the license file for changed dependencies.
+ exit 1
```
2. (By removing the `test` rule dependency for `docker-build`) I get an error from docker 
```
docker build .. -t gcr.io/kubeflow-images-public/notebook-controller
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /path/to/kubeflow/components/Dockerfile: no such file or directory
```

Is the second issue resolved when using kaniko? If yes will that mean that we'll have to use kaniko also for local testing/dev (building docker image locally)?

jlewi:
@kimwnasptd looks like the `docker build` just needed to be executed fromt he parent directory. PTAL.

Thanks for catching this.

Issue #4076, opened 2019-09-04
Profile & Kfam: Prometheus metrics


kunmingg:
See parent issue: https://github.com/kubeflow/kubeflow/issues/4058
And common formatting: https://github.com/kubeflow/kubeflow/issues/4099

yeya24:
I' d like to work on this feature. But what metrics do we need?

kunmingg:
@yeya24 
Thanks!
How about start with metrics in https://github.com/kubeflow/kubeflow/issues/4099?
Any suggestion on how to show error messages through prometheus metrics?

yeya24:
> @yeya24
> Thanks!
> How about start with metrics in #4099?
> Any suggestion on how to show error messages through prometheus metrics?

Actually, error messages is not very suitable for prometheus metrics. They should be deal with some logging systems. 

At most times, we can just use a simple counter type to represent the total times of error.  If the type of error is limited, we can add the error type as a label of the metrics.

kunmingg:
Sounds good, how about categorize errors by calling path and / or error severity?

yeya24:
> Sounds good, how about categorize errors by calling path and / or error severity?

Actually, I don't think this make any sense. This should be done in logging.

kunmingg:
Can we make `reuqest path` and `error severity` as 2 tags in errors?
So we alert is triggered, we can include them in alert messages, and oncall people will know which call path has errors, and roughy what kind of error it is?
For example for [kfam](https://github.com/kubeflow/kubeflow/blob/master/components/access-management/api/swagger.yaml) we can include the paths as error metrics tags

jlewi:
@yeya24 Any update on this?

Bump to P0 because this is required for 1.0.

yeya24:
@jlewi  Actually currently I am not working on this because I saw this PR https://github.com/kubeflow/kubeflow/pull/4251.
I am wondering do we need to wait until that PR merges?

PR #4674, opened 2020-01-21
kfam: prometheus monitoring integration

kunmingg:
requestCounter: 
normal request count with tags: `COMPONENT, KIND, REQUSER, ACTION, PATH`

requestErrorCounter:
failed request count with tags: `COMPONENT, KIND, REQUSER, ACTION, PATH, SEVERITY`

serviceHeartbeat:
counter that increase value by 1 every 10 seconds, indicating the kfam backend is alive.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4674)

### DIFF
--- a/components/access-management/kfam/monitoring.go
+++ b/components/access-management/kfam/monitoring.go
@@ -1,0 +1,67 @@
+package kfam
+
+import (
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const KFAM = "kfam"
+const COMPONENT = "component"
+const KIND = "kind"
+// User that make the request
+const REQUSER = "user"
+const ACTION = "action"
+const PATH = "path"
+const SEVERITY  = "severity"
+const SEVERITY_MINOR = "minor"
+const SEVERITY_MAJOR = "major"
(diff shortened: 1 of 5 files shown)


components/access-management/kfam/monitoring.go
```diff
@@ -0,0 +1,67 @@
+package kfam
```
zhenghuiwang: Do we expect other KF components to follow this pattern(same set of counter names, types etc)? If so, can it be put as a library outside `kfam`?

kunmingg: We can put that effort into v1.0.1 and avoid blocking release of v1.0.0?
@jlewi

kunmingg:
related: https://github.com/kubeflow/kubeflow/issues/4076

components/access-management/kfam/monitoring.go
jlewi: +1 I suspect we will likely need to do some more refactoring post 1.0 as we will identify missing metrics etc... so I would be in favor of getting this merged now and then taking a more holistic look at our monitoring.

PR #4687, opened 2020-01-23
Delete dead code for releasing and updating images.

jlewi:
* These scripts were a much earlier attempt to try to manage image tags
  as part of releases.

* These scripts were abandoned several releases ago.

  * We have new infrastructure
    see: https://github.com/kubeflow/testing/tree/master/apps-cd
    to make it easier to consistently apply semantic tags e.g "vx.y.z"
    to images and update the kustomize manifests

  * We should no longer need to worry about dealing with tags being moved
    because the plan is to use tags + image shas see: kubeflow/testing#558

* update_components.sh is obsolete since we no longer use ksonnet.

Related to kubeflow/testing#573

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4687)


jlewi:
Looks like there was a problem with the test infra and the tests never ran.

/test all

jlewi:
Job is crashing

```
Traceback (most recent call last):  File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main    "__main__", fname, loader, pkg_name)  File "/usr/lib/python2.7/runpy.py", line 72, in _run_code    exec code in run_globals  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 532, in <module>    final_result = main()  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 522, in main    return run(args, file_handler)  File "/src/kubeflow/testing/py/kubeflow/testing/run_e2e_workflow.py", line 293, in run    ks_cmd = ks_util.get_ksonnet_cmd(w.app_dir)  File "/src/kubeflow/testing/py/kubeflow/testing/ks_util.py", line 46, in get_ksonnet_cmd    with open(app_yaml_file) as app_yaml:IOError: [Errno 2] No such file or directory: '/src/kubeflow/kubeflow/components/jupyter-web-app/releaser/app.yaml'
```

jlewi:
Looks like the same error.

Issue #4617, opened 2020-01-03
Allow for service account created by profiles controller to have extra SCC granted


crobby:
/kind feature

**Why you need this feature:**
For openshift to be able to spawn a Jupyter notebook, the default-editor service account currently needs to have an SCC that will allow the fsGroup: 100, which is what's being used by the notebook-controller.  Right now, we have to go back after the profile is created and manually add an scc to allow that permission.

**Describe the solution you'd like:**
I suppose there's more than one way to handle this.  
1)  Allow some sort of config to the profiles service that can attach a list of SCCs to a newly created SA in a profile.
2)  Have the notebook-controller allow us to set an fsGroup.  I'm not sure if this is as desirable of a solution since I'm not entirely sure of the reasoning behind using fsGroup: 100, but it seems like it has a shot at working.

**Anything else you would like to add:**

vpavlin:
I see the default here https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L51

This comment explains that it is needed for GKE (https://github.com/kubeflow/kubeflow/pull/2463#issuecomment-463161704)

For the OpenShift, it would be best if we could skip the `securityContext` creation completely an leave it up to the Pod Security Policies controlle (https://kubernetes.io/docs/concepts/policy/pod-security-policy/) to generate it.

What do you think about adding an option which would turn off the `securityContext` creation in the code?

crobby:
So, I've taken a closer look at this and I'm a little unsure of which approach to take for the fix.  As it is right now, it is possible for me to create a notebook that runs as I intend on Openshift.  If I create a Notebook yaml that contains a podSpec with securityContext: {}, the notebook controller respects that and I get a Notebook that spins-up as I intended (no fsGroup: 100).  That makes me think that possibly a UI tweak [to be able to specify whether or not a special securityContext should be included in the Notebook that is being generated] might be appropriate to allow the flexibility regardless of platform.

Alternatively, maybe some sort of flag to the NotebookController that would allow the code that adds "fsGroup: 100" to run or not as desired would also work.

Any thoughts?

jlewi:
The fixed image is being cherry-picked onto master in 
https://github.com/kubeflow/manifests#935

If we want to cherry-pick onto v1 lets verify on master first.

PR #4713, opened 2020-01-31
Adding env var to suppress automatic additon of fsGroup in notebook pod

crobby:
Allowing for an env var ADD_FSGROUP to be set to false to suppress the automatic addition of fsGroup: 100 in the pod's security context.
If the env var is not specified or set to true, the current behaviour will be unchanged.
This addresses issue #4617.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4713)

### DIFF
--- a/components/notebook-controller/README.md
+++ b/components/notebook-controller/README.md
@@ -33,6 +33,12 @@ That is, the user should specify what and how to run.
 
 All other fields will be filled in with default value if not specified.
 
+## Environment parameters
+
+ADD_FSGROUP:  If the value is true or unset, fsGroup: 100 will be included
+in the pod's security context. If this value is present and set to false, it will suppress the
+automatic addition of fsGroup: 100 to the security context of the pod.  
+
 ## Implementation detail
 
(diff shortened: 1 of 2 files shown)


crobby:
@googlebot I fixed it

vpavlin:
@pdmack Any thoughts? My only concern with the PR is loading the environment variable "deep" in code which makes the configuration less transparent.

@crobby It would be probably also good to add a not to the controller README about the env var

crobby:
> @pdmack Any thoughts? My only concern with the PR is loading the environment variable "deep" in code which makes the configuration less transparent.
> 
> @crobby It would be probably also good to add a not to the controller README about the env var

I went with this approach since it mirrors how ISTIO_GATEWAY/USE_ISTIO are already used in the controller.  If another method is preferable, I'm open to that.

Yes, a note in the README is a good idea.  I will add that.

Thanks

pdmack:
@crobby doesn't this also need to be brought forward to master?
/cc @vpavlin

Issue #4705, opened 2020-01-30
Jupyter UI shows error while waiting for pod to bind PVC


jtfogarty:
/kind bug

**What steps did you take and what happened:**
Created a Notebook Server using Rook/Ceph as the storage layer.

Pod created but the UI shows a red X in the status.  

![Screen Shot 2020-01-30 at 2 20 05 PM](https://user-images.githubusercontent.com/6708730/73487270-43213480-436c-11ea-8bf9-a91892cfa9d6.png)

**What did you expect to happen:**
Pod to start without issue

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
- kfctl version:  v1.0-rc.1-0-g963c787
- Kubernetes platform: v1.15.7
- Kubernetes version: v1.15.7
- OS (e.g. from `/etc/os-release`): NAME="CentOS Linux"
VERSION="7 (Core)"

jtfogarty:
The pod eventually starts and the status turns green

yanniszark:
@jtfogarty is this still an issue?
I suspect this is because the Rook (Ceph) provisioner is taking some time to provision the Persistent Volume, so in the meantime you get this message.

jlewi:
Downgrading to P1 because this doesn't seem like its release blocking.

jtfogarty:
@yanniszark Yes, this is still an issue

jlewi:
I observed the same behavior on GCP using the default storage class (not ROK).

Bump back to P0; I think this is going to confuse a lot of people.

@fediazgon @kimwnasptd Is this related to the changes in the UI to surface events? Is there an easy way to show something less confusing to users?

fediazgon:
Yes, it is related to [this change](https://github.com/kubeflow/kubeflow/pull/4214/files#diff-d575d8cda5f10ed63f52ac8e7a65b733R328). Right now the UI shows that status when the event matches the following predicate `eventType == 'Warning' and eventReason in ["FailedCreate", "FailedScheduling"]`.

One solution would be to limit the events that are shown by looking at the event's description. @kimwnasptd WDYT?

jlewi:
@fediazgon thanks. What if we just returned STATUS_WAITING rather than STATUS_ERROR
https://github.com/kubeflow/kubeflow/pull/4214/files#diff-d575d8cda5f10ed63f52ac8e7a65b733R329

[...]

PR #4729, opened 2020-02-04
Jupyter displays Warn event message when hover over the spinner

fediazgon:
Jupyter UI should show the event's message for any `WARNING` event. The UI should not display an error ❌ next to the affected notebook since it is difficult to distinguish which of these errors are temporary.

Related to: #4705

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4729)

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -320,13 +320,13 @@ def find_error_event(rsrc_events):
     of why the resource could not be created. For a Notebook, it can be due to:
 
           EVENT_TYPE      EVENT_REASON      DESCRIPTION   
-          Warning         FailedCreate      serviceaccount not found (originated in statefulset)
-          Warning         FailedScheduling  Insufficient CPU (originated in pod)
+          Warning         FailedCreate      pods "x" is forbidden: error looking up service account ... (originated in statefulset)
+          Warning         FailedScheduling  0/1 nodes are available: 1 Insufficient cpu (originated in pod)
 
     '''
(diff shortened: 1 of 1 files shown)


components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
```diff
@@ -24,6 +25,8 @@
 EVENT_TYPE_NORMAL = "Normal"
 EVENT_TYPE_WARNING = "Warning"
 
+EVENT_ERROR_MSG_REGEX = re.compile(
```
kimwnasptd: Should we change the `Insufficient cpu` to just `Insufficient` in order to also include other cases such as not enough memory or gpus?

jlewi: See comment 
https://github.com/kubeflow/kubeflow/issues/4705#issuecomment-582396315

I think we should focus on surfacing events rather than trying to correctly classify them. 

Insufficient resources could be a permanent error e.g. user requests 32 CPUs and only has 16 CPU machines.

Or it could be temporary, because of autoscaling or just because some other workloads will finish allowing it to be scheduled.

jlewi:
/lgtm
/approve
/hold
@kimwnasptd do you want to take another look?

@fediazgon You can send me a PR adding yourself to the org
https://github.com/kubeflow/internal-acls/blob/master/github-orgs/kubeflow/org.yaml

If you want to be able to auto-trigger the tests

jlewi:
I think @kimwnasptd is done for the day; so I'm going to remove the hold as I'd like to get a fix cherry-picked onto the branches today.

/hold cancel

Issue #4746, opened 2020-02-09
Expose more attributes options in profile controller


Jeffwan:
/kind feature

**Why you need this feature:**
Currently, profile controller create istioRBAC with attributed `request.headers`

```
Properties: map[string]string{fmt.Sprintf("request.headers[%v]", r.UserIdHeader): r.UserIdPrefix + profileIns.Spec.Owner.Name},
```
https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/controllers/profile_controller.go#L389

In AWS, there's no header can be used directly, we consider to use `claim.*` after istio decode our jwt token. In this case, we need to make some changes in profile controller in order to create right istio rbac rules. 

https://istio.io/docs/reference/config/policy-and-telemetry/attribute-vocabulary/

One example will be:
```
request.auth.claims[username]: kubeflow-user
request.auth.claims[email]: kubeflow-user@example.com
```

**Describe the solution you'd like:**
In order to make it generic, I will introduce two more arguments to describe the attribute name and type. This will give enough flexibility to handle most of the cases.

Jeffwan:
I will close this issue temporarily since we use different way to resolve the problem. I will reopen once we do use claim in this way

PR #4747, opened 2020-02-09
Expose more istio attribute options in profile controller

Jeffwan:
Address Issue https://github.com/kubeflow/kubeflow/issues/4746

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4747)

### DIFF
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -85,6 +85,8 @@ type ProfileReconciler struct {
 	Log              logr.Logger
 	UserIdHeader     string
 	UserIdPrefix     string
+	AttributeName    string
+	AttributeType    string
 	WorkloadIdentity string
 }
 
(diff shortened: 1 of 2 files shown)


components/profile-controller/controllers/profile_controller.go
```diff
@@ -377,6 +379,14 @@ func (r *ProfileReconciler) updateIstioRbac(profileIns *profilev1.Profile) error
 			}
 		}
 	}
+
+	// Check all options here https://istio.io/docs/reference/config/policy-and-telemetry/attribute-vocabulary/
+	var propertyKey string
+	if r.AttributeType == "map" {
```
jlewi: Would it possible to write a unittest to ensure we generate the expected ServiceRoleBinding?

jlewi:
@Jeffwan so just to double check; it looks like you added some command line arguments. If those flags aren't set and the default values are used the behavior matches the current behavior?

components/profile-controller/controllers/profile_controller.go
Jeffwan: Definitely. I should. I will update the PR later.

kunmingg:
Thanks for the update!

Jeffwan:
I will close this PR at this moment and I move to a different solution without profile controller change. https://github.com/kubeflow/manifests/pull/908  I will reopen whenever there's other similar requirement.

Issue #4403, opened 2019-10-24
Fine Grained IAM roles for AWS


Jeffwan:
/kind feature

**Why you need this feature:**
[Is your feature request related to a problem? Please describe in details]

Right now, we attach IAM policy to node groups to make sure AWS specific components have permission to talk to AWS services.  The problem is this way only provides fine-grained roles at the node level rather than pod level. 

In the context of access control in Amazon EKS, the community came up with a number of open source solutions, such as kube2iam, kiam, and Zalando’s IAM controller. As EKS already open source a solution, we want to use it in Kubeflow as well. 

**Describe the solution you'd like:**
[A clear and concise description of what you want to happen.]
Use IAM for pods to manage permission

https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/specifying-conditions.html
https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

The solution is open source, we'd like to make this work with any AWS setups (like kops) as well.

PR #4751, opened 2020-02-10
Add AWS IAM Role for ServiceAccount support in profile controller

Jeffwan:
kubeflow/kubeflow#4403

Profile controller side changes

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4751)


jlewi:
/assign @kunmingg 

@kunmingg How risky do you think this change is? If we cherry-pick it onto 1.0 are you concerned anything other than AWS would be impacted? The code looks pretty isolated so it looks low risk to me.

kunmingg:
@Jeffwan 
Thanks for the new plugin! Really cool.
Can we 
* add new plugin at https://github.com/kubeflow/kubeflow/tree/master/components/profile-controller#plugins?
* add one profile sample with new plugin at: https://github.com/kubeflow/kubeflow/tree/master/components/profile-controller/config/samples?
* add 1 or 2 tests to verify plugin works as expected?

@jlewi 
Should be fine to cherry-pick, each plugin's logic should only apply when set.

Jeffwan:
@kunmingg Thanks. I will update the PR.

Jeffwan:
@kunmingg  Please have another look. doc and examples have been updated. UT and manual test pass as well.

kunmingg:
/lgtm
Thanks!

How does user bind multiple aws role per profile?

Jeffwan:
@kunmingg 
Thanks for helping review. Currently, only one role can be attached. To support multiple roles, we can either accept string list for `awsIamRole`, or propose group concept in profile controller and grant user the right group of permissions

Issue #4749, opened 2020-02-10
Profile controller uses project id from user sa for identity namespace


daikeshi:
/kind bug

**What steps did you take and what happened:**
We have teams in different gcp projects are using a shared Kubeflow cluster.  When I try to use a gcp service account from a different project for `gcpServiceAccount` field in a user's profile, profile-controller throws an error: 
```Identity namespace does not exist (${PROJECT_ID}.svc.id.goog)```

It seems the project id used in identity namespace is inferred from user's gcp service account. https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/controllers/plugin_workload_identity.go#L54

**What did you expect to happen:**
I think it might make more sense to use the id of the project where Kubeflow cluster was created in.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
- kfctl version: (use `kfctl version`): 0.7.1
- Kubernetes platform: (e.g. `minikube`): GKE
- Kubernetes version: (use `kubectl version`): v1.14.8-gke.33
- OS (e.g. from `/etc/os-release`): mac OS Mojave v10.14.6

jlewi:
I don't know that we will fix this issue.

Per kubeflow/manifests#1136 I think we are moving away from profiles and bootstrapping namespaces using GitOps and declarative packages.

We are also moving to using Cloud Config Connector.
https://github.com/kubeflow/gcp-blueprints

So in this new world if you wanted to configure which GCP service account is bound to a KSA you would just modify the K8s resource and if necessary create an IAMPolicyMember to grant the appropriate workload identity on the GCP SA.

daikeshi:
Tha's OK. We currently use a fork of profile controller (with the fix) as the workaround. 
We will need some migration after Kubeflow deprecates profile in the future, but it should be fine.

jlewi:
Removing from 1.1 because I don't think anyone is working on it.

daikeshi:
Hey @jlewi, the fix I made has already been merged in master with this PR https://github.com/kubeflow/kubeflow/pull/4754. I think we just need to publish a new image and update the deployment manifest. Will that be automatically taken care of in Kubeflow 1.1 release so we can remove our fork? Let me know if there's anything I can help with.

jlewi:
kubeflow/kubeflow#5067 is tracking the release of the profile controller for KF 1.1

PR #4754, opened 2020-02-11
Use project id from the default credentials for identity namesapce

daikeshi:
This PR is trying to address https://github.com/kubeflow/kubeflow/issues/4749. @kunmingg 

We can use the project id from the default credentials for the identity namespace when we config the workload identity. This is more reliable than using the project id from the user sa, which can be in a different gcp project.  

Please let me know if this makes sense. I'm happy to help fix the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4754)

### DIFF
--- a/components/profile-controller/controllers/plugin_workload_identity.go
+++ b/components/profile-controller/controllers/plugin_workload_identity.go
@@ -89,8 +89,8 @@ func (gcp *GcpWorkloadIdentity) updateWorkloadIdentity(namespace string, ksa str
 	}
 	ctx := context.Background()
 	gcpSa := gcp.GcpServiceAccount
-	// Get credentials.
-	client, err := google.DefaultClient(context.Background(), iam.CloudPlatformScope)
+	// Get client.
+	client, err := google.DefaultClient(ctx, iam.CloudPlatformScope)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


components/profile-controller/controllers/plugin_workload_identity.go
```diff
@@ -102,14 +102,21 @@ func (gcp *GcpWorkloadIdentity) updateWorkloadIdentity(namespace string, ksa str
...
 	currentPolicy, err := iamService.Projects.ServiceAccounts.GetIamPolicy(saResource).Context(ctx).Do()
 	if err != nil {
 		return err
 	}
 
-	// update policy
-	bindingMember := fmt.Sprintf("serviceAccount:%v.svc.id.goog[%v/%v]", projectID, namespace, ksa)
+	// Update policy
+	// Use ProjectID from the default credentials for identity namespace.
+	bindingMember := fmt.Sprintf("serviceAccount:%v.svc.id.goog[%v/%v]", credentials.ProjectID, namespace, ksa)
```
kunmingg: Can we check if `credentials.ProjectID` is empty here?

daikeshi: Sure, I added a check and it will fallback to `projectID` if `credentials.ProjectID` is empty.

components/profile-controller/controllers/profile_controller_test.go
```diff
@@ -52,7 +52,7 @@ func TestUpdateNamespaceLabels(t *testing.T) {
 	for _, test := range tests {
 		updateNamespaceLabels(test["current"])
 		if !reflect.DeepEqual(test["expected"], test["current"]) {
-			t.Errorf("Expect:\n%v; Output:\n%v", test[current], test["expected"])
+			t.Errorf("Expect:\n%v; Output:\n%v", test["current"], test["expected"])
```
daikeshi: it seems quotes are missing here. it failed locally,

PR #4775, opened 2020-02-18
Distroless latest for admission-webhook Dockerfile

jtfogarty:
This is part of an effort to reduce critical vulnerabilities.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4775)

### DIFF
--- a/components/admission-webhook/Dockerfile
+++ b/components/admission-webhook/Dockerfile
@@ -13,7 +13,7 @@ ENV GO111MODULE=on
 RUN go build  -o webhook .
 
 # Copy the controller-manager into a thin image
-FROM ubuntu:latest
+FROM gcr.io/distroless/base:latest
 WORKDIR /
 COPY --from=builder /go/src/github.com/kubeflow/kubeflow/components/admission-webhook/webhook .
 ENTRYPOINT ["/webhook"]
--- a/components/admission-webhook/Dockerfile
+++ b/components/admission-webhook/Dockerfile
@@ -10,10 +10,10 @@ COPY . .
 ENV GO111MODULE=on
 
 # Build
-RUN go build  -o webhook .
+RUN CGO_ENABLED=0 GOOS=linux go build -o webhook -ldflags "-w" -a .
 
-# Copy the controller-manager into a thin image
-FROM gcr.io/distroless/base-debian10:latest
+# Copy the controller-manager into a distroless image
+FROM gcr.io/distroless/static:latest
 WORKDIR /
 COPY --from=builder /go/src/github.com/kubeflow/kubeflow/components/admission-webhook/webhook .
 ENTRYPOINT ["/webhook"]
--- a/components/admission-webhook/Dockerfile
+++ b/components/admission-webhook/Dockerfile
@@ -13,7 +13,7 @@ ENV GO111MODULE=on
 RUN CGO_ENABLED=0 GOOS=linux go build -o webhook -ldflags "-w" -a .
 
 # Copy the controller-manager into a distroless image
-FROM gcr.io/distroless/static:latest
+FROM gcr.io/distroless/static-debian10:fd0d99e8c54d7d7b2f3dd29f5093d030d192cbbc
 WORKDIR /
 COPY --from=builder /go/src/github.com/kubeflow/kubeflow/components/admission-webhook/webhook .
 ENTRYPOINT ["/webhook"]
--- a/components/admission-webhook/Dockerfile
+++ b/components/admission-webhook/Dockerfile
@@ -13,7 +13,7 @@ ENV GO111MODULE=on
 RUN go build  -o webhook .
 
 # Copy the controller-manager into a thin image
-FROM gcr.io/distroless/base:latest
+FROM gcr.io/distroless/base-debian10:latest
 WORKDIR /
 COPY --from=builder /go/src/github.com/kubeflow/kubeflow/components/admission-webhook/webhook .
 ENTRYPOINT ["/webhook"]


components/admission-webhook/Dockerfile
```diff
@@ -13,7 +13,7 @@ ENV GO111MODULE=on
 RUN go build  -o webhook .
 
 # Copy the controller-manager into a thin image
-FROM ubuntu:latest
+FROM gcr.io/distroless/base:latest
```
jlewi: Do you want to use base-debian10? 
see
https://github.com/GoogleContainerTools/distroless#base-operating-system

Do you need base or could you get by with just static?
https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md

jtfogarty:
I tested this and gcr.io/distroless/base:latest works, gcr.io/distroless/static:latest does not

components/admission-webhook/Dockerfile
```diff
@@ -13,7 +13,7 @@ ENV GO111MODULE=on
 RUN go build  -o webhook .
```
swiftdiaries: I'm assuming that this part is what prevents us from using the static image.

`CGO_ENABLED=0 GOOS=linux go build -o webook -ldflags "-w" -a .`
Could you please try this instead?
Without `CGO` we should be able to build with static

components/admission-webhook/Dockerfile
swiftdiaries: https://github.com/kubeflow/kubeflow/pull/4775/files#r382505839

components/admission-webhook/Dockerfile
jtfogarty: Using base-debian10

jtfogarty:
@jlewi using base-debian10 for distroless.

tested fine

components/admission-webhook/Dockerfile
jtfogarty: @swiftdiaries where is that line `CGO_ENABLED=0 GOOS=linux go build -o webook -ldflags "-w" -a .` 
it's not in the Makefile

jlewi: I think @swiftdiaries is suggesting changing the existing go-build line to what he provided. I think setting CGO_ENABLED=0 removes a dependency on some c stuff in the compiled binary which might allow the use of static instead of base.

jtfogarty: @swiftdiaries 
I have the below in the Dockerfile;
```
# Build
#RUN go build  -o webhook .
RUN CGO_ENABLED=0 GOOS=linux go build -o webook -ldflags "-w" -a .
# Copy the controller-manager into a distroless image
FROM gcr.io/distroless/base-debian10:latest
WORKDIR /
COPY --from=builder /go/src/github.com/kubeflow/kubeflow/components/admission-webhook/webhook .
```

But I'm getting the below errors;
```
COPY failed: stat /var/lib/docker/overlay2/a11ca9ddcef163c8b24f630971afd113cf3f45adae11a9fd0cb81ef5a862e6e5/merged/go/src/github.com/kubeflow/kubeflow/components/admission-webhook/webhook: no such file or directory
```

krishnadurai: @jeff in the line:
```
RUN CGO_ENABLED=0 GOOS=linux go build -o webook -ldflags "-w" -a .
```
Webhook is mis-spelt.

jlewi:
/lgtm
/approve
/hold

@jtfogarty Using the distroless image base is a huge improvement over using Debian. It looks like we might be able to get some additional improvements by following @krishnadurai 's suggestion to see if we can make it work with static-debian10. I'll leave that up to you. If you want to submit as is you can remove the hold.

jtfogarty:
@jlewi I tested with static, now I'm testing with static-debian10.  I think this is the desired state.  It should be finished shortly

jtfogarty:
@jlewi There is no static-debian10:latest
https://console.cloud.google.com/gcr/images/distroless/GLOBAL/static-debian10?gcrImageListsize=30

Currently, the docker file uses static:latest.

Please advise

krishnadurai:
@jtfogarty you could possibly test it against this tag:
```
gcr.io/distroless/static-debian10@sha256:4433370ec2b3b97b338674b4de5ffaef8ce5a38d1c9c0cb82403304b8718cde9
```

Which is the latest published image.

jtfogarty:
@krishnadurai ok, this is running

@jlewi do you want me to hardcode the image sha256:

krishnadurai:
In hindsight, I believe it would be better to use the tag for the above image:

`fd0d99e8c54d7d7b2f3dd29f5093d030d192cbbc`

jlewi:
* I think we want to use the debian10 images not the debian9 images on the assumption debian10 includes patches/fixes that we want
* It looks like static corresponds to debian9 so I think we should use static-debian10
* It looks like the lack of "latest" is a known issue https://github.com/GoogleContainerTools/distroless/issues/436
  * We should probably add that as a TODO to switch to the latest tag
  * I think the recommended practice is to use the latest tag so that everytime we rebuild we get updates/fixes as opposed to have to manually update the base image
* I have no strong opinion on using tags vs. shas since in this case we'd actually like to use latest 

@jtfogarty Thanks for your forebearance. It might be worth capturing all of this info somewhere; maybe 
https://github.com/kubeflow/community/tree/master/proposals
or
https://github.com/kubeflow/community/tree/master/guidelines

krishnadurai:
@jlewi thanks for summarising this. I'll take a stab at making a guideline for distroless images.

swiftdiaries:
@jtfogarty Thank you for the patience

PR #4777, opened 2020-02-18
Fix notebook controller event filtering

### LINKED ISSUE #4776: notebook-controller: Pod without "-" in name breaks notebook-controller
/kind bug

1. Create a Pod without any dashes ("-") in its name.
2. Notebook-controller goes into CrashLoopBackOff.

* Notebook-controller not to crash.

The offending code seems to be the event predicate filtering.
It filters all Pods instead of just the ones belonging to notebook-controller.
The code should be changed to:
* Only get the notebook-related Pod events (check ownerReference).
* Handle the case where "-" doesn't exist in name (maybe log an error).

Stack trace:
```
E0218 15:17:23.036298       1 runtime.go:69] Observed a panic: "slice bounds out of range" (runtime error: slice bounds out of range)
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/runtime/runtime.go:76
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/runtime/runtime.go:65
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/runtime/runtime.go:51
/usr/local/go/src/runtime/panic.go:522
/usr/local/go/src/runtime/panic.go:54
/workspace/notebook-controller/controllers/notebook_controller.go:470
/workspace/notebook-controller/controllers/notebook_controller.go:549
/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0/pkg/predicate/predicate.go:63
/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0/pkg/source/internal/eventsource.go:68
/go/pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/shared_informer.go:554
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:265
/go/pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/shared_informer.go:548
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:152
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:153
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:88
/go/pkg/mod/k8s.io/client-go@v11.0.1-0.20190409021438-1a26190bd76a+incompatible/tools/cache/shared_informer.go:546
/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:71
/usr/local/go/src/runtime/asm_amd64.s:1337
panic: runtime error: slice bounds out of range [recovered]
        panic: runtime error: slice bounds out of range

```

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): v1.0-rc.0-44-g1913078

/cc @lluunn @jlewi 
/priority p0

yanniszark:
Related to #4776
@jlewi should we close #4776 when this is merged, or wait for the manifests as well?

This PR fixes event filtering/reconciling for the notebook controller.
- Add check to avoid crashing if the notebook name doesn't contain a dash ("-").
- Fix reconciliation based on events.

/cc @jlewi @lluunn

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4777)

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -83,12 +83,17 @@ func (r *NotebookReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
 	log := r.Log.WithValues("notebook", req.NamespacedName)
 
+	// TODO(yanniszark): Can we avoid reconciling Events and Notebook in the same queue?
 	event := &v1.Event{}
 	var getEventErr error
 	getEventErr = r.Get(ctx, req.NamespacedName, event)
 	if getEventErr == nil {
 		involvedNotebook := &v1beta1.Notebook{}
-		involvedNotebookKey := types.NamespacedName{Name: nbNameFromInvolvedObject(&event.InvolvedObject), Namespace: req.Namespace}
+		nbName, err := nbNameFromInvolvedObject(r.Client, &event.InvolvedObject)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		involvedNotebookKey := types.NamespacedName{Name: nbName, Namespace: req.Namespace}
 		if err := r.Get(ctx, involvedNotebookKey, involvedNotebook); err != nil {
 			log.Error(err, "unable to fetch Notebook by looking at event")
 			return ctrl.Result{}, ignoreNotFound(err)
@@ -464,12 +469,30 @@ func isStsOrPodEvent(event *v1.Event) bool {
 	return event.InvolvedObject.Kind == "Pod" || event.InvolvedObject.Kind == "StatefulSet"
 }
 
-func nbNameFromInvolvedObject(object *v1.ObjectReference) string {
-	nbName := object.Name
-	if object.Kind == "Pod" {
-		nbName = nbName[:strings.LastIndex(nbName, "-")]
+func nbNameFromInvolvedObject(c client.Client, object *v1.ObjectReference) (string, error) {
+	name, namespace := object.Name, object.Namespace
+
+	if object.Kind == "StatefulSet" {
+		return name, nil
 	}
-	return nbName
+	if object.Kind == "Pod" {
+		pod := &corev1.Pod{}
+		err := c.Get(
+			context.TODO(),
+			types.NamespacedName {
+				Namespace: namespace,
+				Name:      name,
+			},
+			pod,
+		)
+		if err != nil {
+			return "", err
+		}
+		if nbName, ok := pod.Labels["notebook-name"]; ok {
+			return nbName, nil
+		}
+	}
+	return "", fmt.Errorf("object isn't related to a Notebook")
 }
 
 func nbNameExists(client client.Client, nbName string, namespace string) bool {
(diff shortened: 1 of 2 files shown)


components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -528,9 +529,10 @@ func (r *NotebookReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	eventToRequest := handler.ToRequestsFunc(
 		func(a handler.MapObject) []ctrl.Request {
+			event := a.Object.(*v1.Event)
 			return []reconcile.Request{
 				{NamespacedName: types.NamespacedName{
-					Name:      a.Meta.GetName(),
+					Name:      nbNameFromInvolvedObject(&event.InvolvedObject),
```
fediazgon: I am not sure about this change. Later, in the reconcile call, we need to fetch the actual event:

```
getEventErr = r.Get(ctx, req.NamespacedName, event)
```

fediazgon: I am not sure about this change. Later, in the reconcile call, we need to fetch the actual event:

```
getEventErr = r.Get(ctx, req.NamespacedName, event)
```

yanniszark: Oh wow, I hadn't seen that because I never expected to see two kinds inside the same reconcile queue.
This seems like a very bad practice.

Thanks for pointing it out, with the new handling we never enqueue events.

yanniszark:
@fediazgon thanks for the review, I removed the now obsolete code from Reconcile.
What do you think?

components/notebook-controller/controllers/notebook_controller.go
fediazgon: I don't think that code is obsolete. The goal of enqueueing events that were originated in a pod or sts related to a notebook is:
 
* Fetch the event.
* Fetch the notebook related to the previous event (using `nbNameFromInvolvedObject`).
* Reissue the event as if it was the notebook the source of the event.

That code was introduced to solve https://github.com/kubeflow/kubeflow/issues/3772 and it is used by the jupyte web-app to enrich the information shown to the user (see [this](https://github.com/kubeflow/kubeflow/pull/4214/files#diff-e737059d17e1c37ba095426315b34847R110)).

So I guess the only change needed in this PR is to fix `nbNameFromInvolvedObject`, right?

yanniszark: You're right, the reissuing is the only thing that's not handled.
I'm not very fond of this practice because it can lead to name collisions (event name == sts name) and skipped reconciliations.
They should be very rare in practice though.
I'm ok with leaving it like this for now.

fediazgon:
LGTM

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -466,7 +467,7 @@ func isStsOrPodEvent(event *v1.Event) bool {
 
 func nbNameFromInvolvedObject(object *v1.ObjectReference) string {
 	nbName := object.Name
-	if object.Kind == "Pod" {
+	if object.Kind == "Pod" && strings.Contains(nbName, "-") {
```
jlewi: I'm confused about what the function nbNameFromInvolvedObject is doing

It looks to me like the code is looking for pods that are associated with the notebook. So it expects pods of the form 
${NBNAME}-0

Should we be more stringent in matching names? e.g. should we specifically look for pods named

nbname-0

So we don't match nbname-11234

Or whatever?

Can we add a unittest to catch this and other cases?

jlewi:
@yanniszark lets not close the issue until its been cherry-picked into 1.0 and the fix has been verified.

components/notebook-controller/controllers/notebook_controller.go
yanniszark: @jlewi I think it's fine, because after that the predicate will try to find the notebook with that name and skip it if it doesn't exist.
So it's ok to be a little lax in this check.
Otherwise, I'll get to this tomorrow.

jlewi: Can we add the unittest? Since this caused a bug it would be good to have a test to prevent a regression.

jlewi:
@yanniszark Could you add the unittest and get this merged?

yanniszark:
@jlewi pushed updated code that:
1. Finds notebook from notebook-name label on Pod.
2. Adds unittest for updated nbNameFromInvolvedObject function.

jlewi:
This is great. Thanks @yanniszark 

/lgtm
/approve

Issue #4015, opened 2019-08-27
[CentralDashboard v2] Create end-to-end testing for Releases


avdaredevil:
## About
We want to have end-to-end verification of health / state status of central-dashboard. This would allow us to make sure the state we expect in CentralDash is exactly what the backends expect

### Context:
- The 2 routes most helpful for this are:
  - `//<host>:<port>/debug` - Lists important state info, and environment data that can be used for quick verifications
  - `//<host>:<port>/api/workgroup/env-info` - Lists metrics, platform data, user info, namespaces (_you'll only get scoped data if your headers are set correctly for MUI_)

### Meta
/area centraldashboard
/area front-end
/priority p1
/kind feature
/assign @avdaredevil @jlewi 
/cc @prodonjs

jlewi:
I think it would be valuable to write a short doc/tech note about how we could do testing for the central dashboard.

When doing the 0.7 release we hit kubeflow/kubeflow#4415. This is an issue where the self registration flow is not working in the case where Kubeflow clusters don't have identity.

prodonjs:
@avdaredevil - I don't think I have the bandwidth to carry it forward, but I put together a rough pass at an E2E test in https://github.com/prodonjs/kubeflow/commit/bca6def173959c2721d1effdbe2e4e91cf41c94f. This was intended to test the IAP-scenario, but it could be easily adapted to handle basic-auth as well and let us have coverage for both registration scenarios.

Maybe we can schedule some time to discuss next week.

jlewi:
@prodonjs @avdaredevil is there anyone that could pick up https://github.com/prodonjs/kubeflow/commit/bca6def173959c2721d1effdbe2e4e91cf41c94f and push it over?

prodonjs:
What's the deadline? I'm on leave for another couple of weeks so I won't be available to carry it through myself.

@avdaredevil - if you have the cycles and need a consult or have questions, I may be able to respond.

jlewi:
Thanks @prodonjs

jlewi:
I looked at https://github.com/prodonjs/kubeflow/commit/bca6def173959c2721d1effdbe2e4e91cf41c94f

I see some javascript to drive the Kubeflow central dashboard.

I think some of the additional work needed is

* Figuring out how to drive this in our E2E tests?
* Figuring out how to get credentials and talk to the KF cluster?
  * I think we will want to access central dashboard over the external accesspoint so that redirects would work
[...]

PR #4773, opened 2020-02-17
[Jupyter-Web-App]:Add basic E2E Tests

kimwnasptd:
This test has two basic test cases for now.

* Navigate to index page and ensure that no error appeared
* Navigate to form page and ensure that no error appeared

cc @jlewi @fediazgon 
/assign @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4773)

PR #4780, opened 2020-02-19
[CentralDashboard][Testing] End-to-end testing added

avdaredevil:
## About
End-to-end testing added! Features include:
- Error surfacing
- HTML mismatch verification
- User dirty state verification
- Registration flow logic (thanks to @prodonjs)
- Comments explaining all ENV values, and where to get them from, including docs
- Small fix in API_Workgroup (type related - non-functional).
- API_Workgroup now is better at server-side error logging
- Tabbed output
- New API added in Central-Dash (`DELETE /api/workgroup/nuke-self`) which allows a user to delete their own workgroup programmatically (_note the use of `DELETE` method_)
- The test can now self-clean after a run, otherwise has clean instructions in case there's a mismatch

### Meta
/area centraldashboard
/area front-end
/priority p0
/assign @avdaredevil
/cc @kwasi @prodonjs @jlewi @truptinayak

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4780)


kunmingg:
lgtm for me, thanks!

jlewi:
/assign @kimwnasptd 

@kimwnasptd Do you also want to take a look at this since you are also working on E2E testing for our UIs?

jlewi:
@avdaredevil lets not close #4015 until we have results showing up in our test and our prow job. I believe there will be follow on work in order to report results to prow.

Could you explain and maybe capture in a README how the test works? A couple key questions,

* How are you driving the UI? Are you using a framework like selenium and a browser?
* How does this compare to what @kimwnasptd is doing in kubeflow/kubeflow#4773 to test the Jupyter Web App?

* It looks like this is using a framework called jasmine? What does that do?
* It looks like you are using [puppeteer](https://github.com/puppeteer/puppeteer) to create a browser?
  * How is this different from selenium?
  * What does this mean in terms of requirements for a container that is running puppeter? Do we have to have chrome installed?

kimwnasptd:
@jlewi thanks for adding me in the loop, start looking on it now

components/centraldashboard/test/e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+async function interceptHeadersForDebugging(page: Page) {
+    await page.setRequestInterception(true);
+    page.on("request", request => {
+        console.log("Request:", request.url(), request.headers());
+        request.continue();
+    });
+}
+
+/**
+ * This function should help any dev down the line quickly inspect all
```
kimwnasptd: nit: Is this comment the same with the one above?

components/centraldashboard/test/e2e.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    KF_HOST,   // The url to your cluster, usually: https://<cluster>.endpoints.<project>.cloud.goog/
+    CLIENT_ID, // This is the CLIENT_ID token used to create the cluster (used in http://deploy.kubeflow.cloud/)
+    SERVICE_ACCOUNT_EMAIL, // The email address for the service account you created <name>@<project>.iam.gserviceaccount.com
+    SERVICE_ACCOUNT_KEY,   // Should be a JSON file downloaded from cloud console service accounts
+} = process.env;
+
+['KF_HOST', 'CLIENT_ID', 'SERVICE_ACCOUNT_EMAIL', 'SERVICE_ACCOUNT_KEY'].forEach(envVar => {
+    if (!process.env[envVar]) {console.log(`${envVar} environment variable must be set`);process.exit(1);}
+});
+if (!/^https?:\/\/\S+/.test(KF_HOST)) {console.log('Invalid HOST url provided, must be like http*://*');process.exit(1);}
```
kimwnasptd: nit: `Invalid HOST url provided, must be like https*://*` (the protocol should be https in the end)

kimwnasptd:
Regarding how it compares with JWA's E2E tests, they are very similarly structured.
First both get the JWT for accessing the IAP endpoint and then configure the browser/driver to use this token on each request to have access to the cluster.

In these tests @avdaredevil  ensures that the Central Dashboard's login workflows and user handling are working as intended, as well that the dashboard has properly rendered (Apoorv correct me if something I say is off). 

JWA's test take for granted that we have a Central Dashboard that will have a namespace selector and will interact both with the Dashboard (to change the namespace) as well as with the iframed page (Jupyter Web App).

components/centraldashboard/test/e2e.test.ts
avdaredevil: Fixed, thanks for catching that!

components/centraldashboard/test/e2e.test.ts
avdaredevil: Right, usually clusters redirect, so I allowed users to pass in `http` or `https`, but I've changed the log statement AND regex to simply only allow `https`

avdaredevil:
In addendum to @kimwnasptd's comments and @jlewi's open questions. I've added the docs to elaborate on the methodology, stack, running instructions, open questions, etc.

jlewi:
Thanks @kimwnasptd my question about comparison to JWA was more about the test frameworks and test infra being used as opposed to test coverage.

Ideally I'd like to see as much sharing of test frameworks between the various UIs (e.g. Jupyter, Central Dashboard, Metadata, Katib, etc...).

Ultimately its up to the application developers but I think reusing frameworks and sharing knowledge will be beneficial.

My ask to both of you would be to consider presenting at the community meetings to describe how you are doing UI testing.

jlewi:
Thanks @avdaredevil for driving this; I'll leave some comments in #4015 about next steps.

/lgtm
/approve

/hold
To give @kimwnasptd a chance to review.

kimwnasptd:
/lgtm
Nice work Apoorv!

I agree with @jlewi's point regarding the reuse of frameworks. Since all of the web apps deployed with kubeflow will be interacting with the dashboard, it would be more maintainable to have reusable code in testing. This way, we could have the code for interacting with the dashboard in one place for all the different tests to use.

So I would suggest to encourage the use of one single testing framework for our different Kubeflow UIs. Regarding which framework that should be, I think a factor we should consider is, as Jeremy [described](https://github.com/kubeflow/kfctl/pull/249#discussion_r382366651), how to integrate it with our testing infrastructure.

I haven't worked with Jasmine, so I can't really evaluate how easy it will be to integrate it with prow logging, although I'm open to learning/using it for JWA's tests in order to have a common testing framework between apps.

avdaredevil:
Thanks, Kimonas!

Right makes perfect sense! What framework are you guys currently using?

We mirrored the testing development practices similar to the frameworks we use internally. As a result, we're currently using Jasmine for Unit / End-to-End testing. It's very similar to `mocha` but I believe more feature-complete due to it being an older codebase

As far as the integration layer is concerned, how does the framework actually report data to prow? Is there a well-known API, or does it use exit codes?

jlewi:
@avdaredevil @kimwnasptd I'm removing the hold. Lets continue the discussion and frameworks else where (mailing list). As I mentioned before I would encourage both of you to present at upcoming community meeting to share knowledge about how you are doing E2E testing. Please consider adding yourselves to the community meeting agendas.

/hold cancel
/lgtm
/approve

Issue #4789, opened 2020-02-21
Notebook image continuous builds need to be updated to use Tekton 


jlewi:
/kind feature

On postsubmits we were running ksonnet workflows to auto build the notebook images and 
push them to gcr.io/kubeflow-images-public

We need to make two updates

1. We want to get rid of ksonnet and Argo and switch to using Tekton for CI/CD pipelines
1. We want to run these Tekton pipelines in our release infrastructure
    * see https://github.com/kubeflow/testing/tree/master/apps-cd

Today we have release infrastructure for continuously building Kubeflow Docker Images and updating our Kustomize manifests to reference the new images. This is described at
https://github.com/kubeflow/testing/tree/master/apps-cd

We'd like to do the same for the notebook images. However, what exists needs to be modified because for the notebook images we need to update the configmap listing the notebook images and not the kustomization.yaml file.

/cc @Tabrizian

Tabrizian:
@jlewi I can work on this if this is not already done...

marcindulak:
Will this setup create a chance so more notebook image types will be maintained?

The tensorflow notebook images are well maitained

![Screenshot from 2020-02-23 11-23-33](https://user-images.githubusercontent.com/3178318/75110504-49cf5e80-562f-11ea-9d6a-4d51082cd8b3.png)

but pytorch and other frameworks seem not

![Screenshot from 2020-02-23 11-23-08](https://user-images.githubusercontent.com/3178318/75110509-5358c680-562f-11ea-93cf-ac4636605a09.png)

Tabrizian:
@marcindulak, I am not sure whether it is in our plan but it is definitely a requirement. Meanwhile, you can easily create your own custom images and use them for the notebooks.

jlewi:
@Tabrizian if you could take a look at this that would be fantastic.

Tabrizian:
@jlewi Sorry for the delay, I was looking into this and I am not sure about exactly what I am supposed to do. As far as I understood, I should add an entry to the file below for building the tensorflow images and remove the ks and jsonnet files, right?

https://github.com/kubeflow/testing/blob/master/apps-cd/applications.yaml

jlewi:
@Tabrizian sorry for the delay. I think its a bit more involved then that. 

Here's how our CD currently works.

* We have a [Tekton pipeline](https://github.com/kubeflow/testing/blob/master/apps-cd/pipelines/base/pipeline.yaml) that consists of two tasks

[...]

PR #4791, opened 2020-02-21
Remove dead tests and notebook image release on postsubmit.

jlewi:
* We no longer want to run building and publishing of tensorflow notebook
  images on postsubmit because we won't have permission in the CI infrastructure
  to push to gcr.io/kubeflow-images-public.

  * See https://github.com/kubeflow/kubeflow/issues/4789 for fixing
    Notebook image CD

* The minikube tests aren't being maintained so just get rid of them
  * We have migrated off of ksonnet to using python workflows
    which are triggered off the kfctl repo.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4791)


gabrielwen:
@jlewi let me know when it's ready.

Issue #4809, opened 2020-03-01
Admission Controller failed to build 


Jeffwan:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

```
$ make build
go build -gcflags 'all=-N -l' -o bin/webhook main.go
go: downloading k8s.io/api v0.0.0-20190222213804-5cb15d344471
go: downloading k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
go: downloading sigs.k8s.io/controller-runtime v0.1.10
go: extracting k8s.io/api v0.0.0-20190222213804-5cb15d344471
go: extracting sigs.k8s.io/controller-runtime v0.1.10
go: downloading k8s.io/client-go v10.0.0+incompatible
go: extracting k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
go: downloading golang.org/x/net v0.0.0-20190415214537-1da14a5a36f2
go: extracting k8s.io/client-go v10.0.0+incompatible
go: downloading golang.org/x/oauth2 v0.0.0-20190402181905-9f3314589c9a
go: downloading golang.org/x/crypto v0.0.0-20190418165655-df01cb2cc480
go: extracting golang.org/x/net v0.0.0-20190415214537-1da14a5a36f2
go: extracting golang.org/x/crypto v0.0.0-20190418165655-df01cb2cc480
go: downloading golang.org/x/sys v0.0.0-20190418153312-f0ce4c0180be
go: extracting golang.org/x/sys v0.0.0-20190418153312-f0ce4c0180be
go: extracting golang.org/x/oauth2 v0.0.0-20190402181905-9f3314589c9a
# command-line-arguments
./main.go:399:18: undefined: codecs
./main.go:420:15: undefined: getCrdClient
./main.go:509:18: undefined: codecs
./main.go:550:14: undefined: configTLS
make: *** [build] Error 2
```

We actually can find this file under package `main` here, 
https://github.com/kubeflow/kubeflow/blob/6409fdec4b181592a138f4e6fb8a55e89be771ea/components/admission-webhook/scheme.go#L28

https://github.com/kubeflow/kubeflow/blob/6409fdec4b181592a138f4e6fb8a55e89be771ea/components/admission-webhook/Makefile#L7

The problem is if other files use `main` package, go run need to supply all files as argument like 

```shell
go build -gcflags 'all=-N -l' -o bin/webhook main.go config.go scheme.go
or 
[...]

PR #4810, opened 2020-03-01
Fix admission webhook build issue

Jeffwan:
Address issue #4809

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4810)

### DIFF
--- a/components/admission-webhook/Makefile
+++ b/components/admission-webhook/Makefile
@@ -4,9 +4,8 @@ TAG ?= $(eval TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always
 GOLANG_VERSION ?= 1.12
 
 build:
-	go build -gcflags 'all=-N -l' -o bin/webhook main.go
+	go build -gcflags 'all=-N -l' -o bin/webhook .
 
 build-gcr:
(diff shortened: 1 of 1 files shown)


jlewi:
/lgtm
/approve

Do we want to run builds on presubmit e.g. by using Kaniko to ensure everything is working?
/hold

Jeffwan:
@jlewi That will be better. There're a few components in kubeflow/kubeflow. It might be hard to build all of them. Otherwise, hook probably need to detect the project changes belong to and build that specific one.

jlewi:
prow_config.yaml supports triggering certain workflows only when certain paths are modified. So we could have different workflows for the different applications and only trigger them when the corresponding source code is modified.

@gabrielwen is working on supporting Tekton in our E2E workflows. My hope is that will allow us to create reusable Tekton tasks for things like build a container, run unittests etc...

If you wanted to contribute to that I think you could try to define a Tekton Pipeline that we could run to build and test the admission hook. 

If you want to submit this as is just remove the hold.

Jeffwan:
/hold cancel

Cool. Let me check Tekton tasks and see if I can build some CI for projects under kubeflow components.

Issue #4811, opened 2020-03-01
Profile Controller: Run unit tests when building docker image


yanniszark:
/kind feature

**Why you need this feature:**
To validate a profile-controller image doesn't contain code that breaks unit tests.

**Describe the solution you'd like:**
The profile controller Dockerfile should also run unit tests.

cc @jlewi

PR #4812, opened 2020-03-01
Profile Controller: Containerized unit tests

yanniszark:
**Resolves:** https://github.com/kubeflow/kubeflow/issues/4811

**Description:**

Ensure unit tests are run when building a profile-controller docker image.
/cc @jlewi @kunmingg

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4812)


jlewi:
Don't we want to run unittests on presubmits so that we don't check in bad code?

yanniszark:
@jlewi that's true as well.
I think it's still a best practice to run unit tests on image builds.
We can open an issue for adding them to our presubmits as well.

What do you think?

jlewi:
@yanniszark Isn't doing them in presubmits more important? If you do them in image builds isn't that sort of too late? At that point we've already checked in broken code.

If our builds are reproducible (and they should be) then with high fidelity if we know at check in time that tests are passing then building from the same commit should also pass. If our builds/tests are not reproducible then we should aim to fix that.

So in this PR can we aim to run the unittests in pre (& postsubmits) and surface the results in prow?

We do this for kfctl. If you look at the prow link for a kfctl PR you can see all the unitest results
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_kfctl/251/kubeflow-kfctl-presubmit/1232071461976412160/


Could we just add a step to one of the test workflows (or create a new one) to run the go tests and copy the junit files to the artifacts dir so we surface the results? Here's what kfctl does:
https://github.com/kubeflow/kfctl/blob/master/testing/workflows/components/unit_tests.jsonnet#L111

Lets not use ksonnet; lets use python to define any workflows.

components/profile-controller/README.md
```diff
@@ -94,6 +94,22 @@ Plugin owners have full control over plugin spec struct and implementation.
...
+For example:
+
+```bash
+wget -O kubebuilder.tag.gz 'https://github.com/kubernetes-sigs/kubebuilder/releases/download/v2.3.0/kubebuilder_2.3.0_linux_amd64.tar.gz'
+mkdir -p bin/.kubebuilder
+tar xvfz kubebuilder.tag.gz -C bin/.kubebuilder --strip-components=2
+rm kubebuilder.tag.gz
+
+# Point to kubebuilder assets directory and run tests
+KUBEBUILDER_ASSETS=$(pwd)/bin/.kubebuilder make test
```
Jeffwan: em. Do we need kubebuilder to run tests? generate will need `controller-gen`. 

```
test: generate fmt vet
	go test ./... -coverprofile cover.out
	$(GO) test ./... -v -coverprofile cover.out
``

PR #4814, opened 2020-03-02
Fix the kubeflow git link for clone.

Tatsinnit:
New to kubeflow but I hope I will help in improving and contributing whenever I can.

The fix in this PR is to rectify the `https` git clone link, with `:` in url it will fail the clone with message like this: `'https://github.com:kubeflow/kubeflow/': URL using bad/illegal format or missing URL` 

* Fix is to rectify the https url to this; `https://github.com/kubeflow/kubeflow`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4814)

### DIFF
--- a/bootstrap/developer_guide.md
+++ b/bootstrap/developer_guide.md
@@ -75,7 +75,7 @@ git clone git@github.com:kubeflow/kubeflow.git
 or
 
 ```sh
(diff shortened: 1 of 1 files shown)


Tatsinnit:
`@googlebot I signed it!`

PR #4853, opened 2020-03-11
static-config-server

kunmingg:
Tiny server that serve static config contents within k8s cluster.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4853)


yanniszark:
@kunmingg can you share details on why this is needed?
Is there an open issue?

kunmingg:
@yanniszark 
Our current use case is hosting jwt public key in private cluster.
Related: 
https://github.com/kubeflow/manifests/pull/999

yanniszark:
@kunmingg thanks for the explanation :)

Issue #4863, opened 2020-03-16
Notebook Images : Change owner to jovyan of /home/jovyan 


jlewi:
/kind bug

We create the home directory /home/jovyan here
https://github.com/kubeflow/kubeflow/blob/296bf145c33afd65b60f2e0e3e62b781c1330386/components/tensorflow-notebook-image/Dockerfile#L90 

but we don't change the owner and group to be jovyan and users respectively.

When using the docker image with Kubeflow notebooks that doesn't matter because we end up overwriting /home/jovyan with a volume that will have user and group set correctly.

However it can lead to unexpected behavior when people use the image directly; e.g. as the base image with func_to_container_op.

/cc @Tabrizian

imskr:
Can I fix this?

jlewi:
@imskr I believe this is already fixed by #4867 

I believe we still need to build updated docker images and then update the configmap on master to use them.

jlewi:
#4789 is tracking continuous builds.

jlewi:
This might not make it into 1.1 as the continuous builds #4789 is still in the works.

PR #4867, opened 2020-03-17
Fix Jupyter Notebook $HOME Permission

Tabrizian:
About #4863

But the permission of $HOME was not being affected in the script not sure whether this PR solves the problem.

/cc @jlewi

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4867)


components/tensorflow-notebook-image/Dockerfile
```diff
@@ -86,7 +86,6 @@ ENV LANGUAGE en_US.UTF-8
 # but allow for non-initial launches of the notebook to have
 # $HOME provided by the contents of a PV
 RUN useradd -M -s /bin/bash -N -u $NB_UID $NB_USER && \
-    chown -R ${NB_USER}:users /usr/local/bin && \
```
jlewi: Why did you remove that line?
Don't we just want to add

```
chown -R ${NB_USER}:users ${HOME}
```

And then you could verify by  building the docker image and doing ls -la /home
hopefully /home/jovyan should be owned by user jovyan and group users.

Tabrizian: Sorry, I was really stupid. Fixed!

jlewi:
Thanks!
/lgtm
/approve

Not sure about the tests though.

jlewi:
Its the TF notebook release that's failing.
http://testing-argo.kubeflow.org/workflows/kubeflow-test-infra/kubeflow-presubmit-tf-notebook-release-4867-2a0406e-6592-abaf?tab=workflow

It looks like building the jupyter images is failing. So could be an issue in the Dockerfile

Tabrizian:
@jlewi This needs a LGTM only

Issue #4878, opened 2020-03-20
Add `~/.local/bin` to Jupyter Notebook PATH


Jeffwan:
/kind feature

**Why you need this feature:**

pip install command add binary to ~/.local./bin which is not included in the PATH.  It would be better for us to use it directly without `export PATH`. 

![image](https://user-images.githubusercontent.com/4739316/77190216-13ea9200-6a96-11ea-9221-3ce49eb1e037.png)

```
echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

This is a problem of Ubuntu actually, our notebook images is built on top of tensorflow official image which uses ubuntu:1804 as base image. 

https://unix.stackexchange.com/a/392710

**Describe the solution you'd like:**
Add `~/.local/bin` into PATH. 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

Jeffwan:
PR merged. I can close the issue

PR #4879, opened 2020-03-20
Add $HOME/.local/bin into PATH

PatrickXYS:
## Issue to fix
Issue #4878 mentions that we want to add `$HOME/.local/bin` into `PATH`. Thus, we don't have to explicitly use `$HOME/.local/bin/cmd` to invoke the cmd located in the `$HOME/.local/bin`.

## Change to make
I add a `ENV` layer in the `DockerFile` of `tensorflow-notebook-image` for this functionality.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/kubeflow/4879)

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -19,6 +19,7 @@ ENV NB_USER $NB_USER
 ENV NB_UID 1000
 ENV HOME /home/$NB_USER
 ENV NB_PREFIX /
+ENV PATH $HOME/.local/bin:$PATH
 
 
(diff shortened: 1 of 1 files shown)


Jeffwan:
@PatrickXYS 


It should be 

```
ENV PATH="$HOME/.local/bin:${PATH}"

```

PatrickXYS:
@Jeffwan Seems like test works fine, are we good to go?

Jeffwan:
> @Jeffwan Seems like test works fine, are we good to go?

I can not approve change in this folder. Wait for someone to review and approve.

Issue #4921, opened 2020-04-06
Kubeflow deploy UI: print out error details from upstream.


kunmingg:
/kind feature

Currently when calling API enabling failed we'll print out general message suggesting user to enable APIs manually.
We should print the upstream error details as well. 

**Why you need this feature:**
[Is your feature request related to a problem? Please describe in details]

**Describe the solution you'd like:**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #4933, opened 2020-04-08
print orignial error when failed enabling GCP APIs

kunmingg:
fix https://github.com/kubeflow/kubeflow/issues/4921

Print out original error message from GCP API as error title.


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -689,7 +689,7 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
                 dialogAsCode: false,
                 dialogBody: 'Please enable cloud resource manager API: https://console.developers.google.com/apis/api/cloudresourcemanager.googleapis.com/' +
                 ' and iam API: https://console.developers.google.com/apis/api/iam.googleapis.com/',
-                dialogTitle: 'Please enable APIs for your project and try again',
+                dialogTitle: "Error calling GCP API:" + err.toString(),
```
jlewi, line 692: nit: Leave a space after the ":"

PR #4940, opened 2020-04-10
Cleanup APT in container images after all installs

saschagrunert:
We now run a dedicated step cleaning up the apt cache, which should
reduce the image size a bit.

/kind cleanup


components/tensorflow-notebook-image/Dockerfile
```diff
@@ -26,36 +26,31 @@ ENV PATH $HOME/.local/bin:$PATH
...
+    lsb-release \
+    openssh-client \
+    python3-dev \
+    python3-pip \
+    python3-setuptools \
+    sudo \
+    unzip \
+    vim \
+    wget \
+    zip
```
saschagrunert, line 49: Just fixed the intends and sorted them alphabetically

PR #4948, opened 2020-04-13
refactoring: deduplication

marload:
I removed the duplicate code.

Thx!


richardsliu:
@marload I think the original version is clearer for readability reasons.

PR #4967, opened 2020-04-17
add deprecation warning on kubeflow c2d

kunmingg:
add deprecation warning on kubeflow c2d

### DIFF
--- a/components/gcp-click-to-deploy/src/DeployForm.tsx
+++ b/components/gcp-click-to-deploy/src/DeployForm.tsx
@@ -235,6 +235,11 @@ const styles: {[key: string]: React.CSSProperties} = {
         fontSize: 20,
         margin: '30px 60px',
     },
+    warning: {
+        color: '#ff0000',
+        fontSize: 18,
+        margin: '30px 60px',
+    },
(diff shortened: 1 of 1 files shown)


components/gcp-click-to-deploy/src/DeployForm.tsx
```diff
@@ -310,6 +315,8 @@ export default class DeployForm extends React.Component<any, DeployFormState> {
             <div>
                 <div style={styles.text}>Create a Kubeflow deployment</div>
 
+                <div style={styles.warning}>Warning: Kubeflow Deploy Site will be deprecated soon. </div>
```
amygdala, line 318: tiny nit: maybe 'This Kubeflow Deploy...' ? Just in case they're wondering if they're on the deploy site.

jlewi: @kunmingg Could you also add instructions point folks at the instructions for using the CLI? 
Thanks

jlewi:
Thanks @kunmingg for turning this around so quickly. One minor comment.

kunmingg:
updated

Issue #4970, opened 2020-04-20
Manage contributor add failure - rolebinding unique name issue


lalithvaka:
/kind bug

**What steps did you take and what happened:**
Manage contributor screen. Adding a contributor creates a role binding / service role binding with name including userID's first character by ignoring the digits and not able to add other users with similar numbers.
our Org uses UserIDs as a1234@xx.org, a2345@xx.org. 

![image](https://user-images.githubusercontent.com/8292387/79807750-b2317800-8320-11ea-8939-a32d2bf33618.png)
![image](https://user-images.githubusercontent.com/8292387/79807788-c2495780-8320-11ea-9654-67fdbca24cca.png)

**What did you expect to happen:**
rolebinding and service role bindings names to use digits too

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):1.0.2
- kfctl version: (use `kfctl version`): 
- Kubernetes platform: (e.g. `minikube`)
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

yanniszark:
This is the code that decides the name of RoleBindings:
https://github.com/kubeflow/kubeflow/blob/8f8e348bc8df0b65f09c3082b1f559fff20d62e6/components/access-management/kfam/bindings.go#L56-L72
We should probably amend that logic.
I'm not sure what priority to put on this.
We should decide our plans for the contributor UI first, because if we are moving towards a GitOps approach, the contributor UI becomes less valuable.
On the other hand, this should be an easy fix if we can decide on what naming rule to use.

/area multiuser
cc @jlewi @kunmingg

lalithvaka:
Thank you @yanniszark , I was able to fix this code(with my limited go knowledge) to support numbers in the email address. This changes  email address to i397401@zq.msds.kp.org to i397401-zq-msds-kp-org which is being used in the role binding names.
```
//getBindingName returns bindingName, which is combination of user kind, username, RoleRef kind, RoleRef name.
func getBindingName(binding *Binding) (string, error) {
	// Only keep lower case letters, replace other with -
	reg, err := regexp.Compile("[^a-z0-9]+")
	if err != nil {
		return "", err
	}
	nameRaw := strings.ToLower(
		strings.Join([]string{
			binding.User.Kind,
[...]

PR #4979, opened 2020-04-27
Pr/email escape (issue 4970)

lalithvaka:
Updating the getBindingName to not ignore the digits in the email while creating a binding name. This is my first ever pull request. Apologies for lot of spaces with no real changes in the file.

```go
//getBindingName returns bindingName, which is combination of user kind, username, RoleRef kind, RoleRef name.
func getBindingName(binding *Binding) (string, error) {
	// Only keep lower case letters, replace other with -
	reg, err := regexp.Compile("[^a-z0-9]+")
	if err != nil {
		return "", err
	}
	nameRaw := strings.ToLower(
		strings.Join([]string{
			binding.User.Kind,
			url.QueryEscape(reg.ReplaceAllString(binding.User.Name, "-")),
			binding.RoleRef.Kind,
			binding.RoleRef.Name,
		}, "-"),
	)

	return reg.ReplaceAllString(nameRaw, "-"), nil
}
```

### DIFF
--- a/components/access-management/kfam/bindings.go
+++ b/components/access-management/kfam/bindings.go
@@ -21,7 +21,6 @@ import (
 	"strings"
 
 	istiorbac "github.com/kubeflow/kubeflow/components/profile-controller/api/istiorbac/v1alpha1"
-
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
@@ -57,7 +56,7 @@ type BindingClient struct {
 
 //getBindingName returns bindingName, which is combination of user kind, username, RoleRef kind, RoleRef name.
 func getBindingName(binding *Binding) (string, error) {
-	// Only keep lower case letters, replace other with -
+	// Only keep lower case letters and numbers, replace other with -
 	reg, err := regexp.Compile("[^a-z0-9]+")
 	if err != nil {
 		return "", err
@@ -167,7 +166,6 @@ func (c *BindingClient) Delete(binding *Binding) error {
 }
 
 func (c *BindingClient) List(user string, namespaces []string, role string) (*BindingEntries, error) {
-
 	bindings := []Binding{}
 	for _, ns := range namespaces {
 		roleBindingList, err := c.kubeClient.RbacV1().RoleBindings(ns).List(metav1.ListOptions{})
--- a/components/access-management/kfam/bindings.go
+++ b/components/access-management/kfam/bindings.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	istiorbac "github.com/kubeflow/kubeflow/components/profile-controller/api/istiorbac/v1alpha1"
+
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
@@ -32,14 +33,15 @@ const ServiceRoleBinding = "servicerolebindings"
 const SERVICEROLEISTIO = "ns-access-istio"
 const USER = "user"
 const ROLE = "role"
+
 //roleBindingNameMap maps frontend role names to k8s role names and vice-versa
-var roleBindingNameMap = map[string]string {
-	"kubeflow-admin":	"admin",
-	"kubeflow-edit":	"edit",
-	"kubeflow-view":	"view",
-	"admin":					"kubeflow-admin",
-	"edit":						"kubeflow-edit",
-	"view":						"kubeflow-view",
+var roleBindingNameMap = map[string]string{
+	"kubeflow-admin": "admin",
+	"kubeflow-edit":  "edit",
+	"kubeflow-view":  "view",
+	"admin":          "kubeflow-admin",
+	"edit":           "kubeflow-edit",
+	"view":           "kubeflow-view",
 }
 
 type BindingInterface interface {
@@ -55,19 +57,21 @@ type BindingClient struct {
 
 //getBindingName returns bindingName, which is combination of user kind, username, RoleRef kind, RoleRef name.
 func getBindingName(binding *Binding) (string, error) {
-	nameRaw := strings.ToLower(
-		strings.Join([]string{
-			binding.User.Kind,
-			url.QueryEscape(binding.User.Name),
-			binding.RoleRef.Kind,
-			binding.RoleRef.Name,
-		},"-"),
-	)
+	//Lalith Change
 	// Only keep lower case letters, replace other with -
-	reg, err := regexp.Compile("[^a-z]+")
+	reg, err := regexp.Compile("[^a-z0-9]+")
 	if err != nil {
 		return "", err
 	}
+	nameRaw := strings.ToLower(
+		strings.Join([]string{
+			binding.User.Kind,
+			url.QueryEscape(reg.ReplaceAllString(binding.User.Name, "-")),
+			binding.RoleRef.Kind,
+			binding.RoleRef.Name,
+		}, "-"),
+	)
+
 	return reg.ReplaceAllString(nameRaw, "-"), nil
 }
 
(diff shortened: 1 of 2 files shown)


lalithvaka:
To provide additional details:  This is in reference to the issue I have submitted earlier - https://github.com/kubeflow/kubeflow/issues/4970

Under Manage Contributors screen, when an email with digits (a247267@zq.msds.kp.org) gets added, the role binding gets created with name as user-a-zq-msds-kp-org-clusterrrole-edit and this will block adding addition emails as contributors with emails starting with a247267@zq.msds.kp.org etc.. Same would happen with lalith1.vaka@google.com lalith.vaka123@google.com. 

This fix would honor the digits in addition to characters in the user email while creating Role Binding names. And in the above example the role binding name would be created as user-a247267-zq-msds-kp-org-clusterrrole-edit.

yanniszark:
@lalithvaka thanks a lot for the PR!
This seems like a great opportunity to add a couple of unit tests to that function, to ensure it yields a correct result.
I would recommend the following cases:
* only letters
* only numbers
* letters and numbers (starting with letter)
* letters and numbers (starting with numbers)

Here are some details on table-driven tests in Go: https://github.com/golang/go/wiki/TableDrivenTests
Here is an example from our code base: https://github.com/kubeflow/kubeflow/blob/c7150a18b3a567a7f1d59893a0ef2adc1245ae14/components/profile-controller/controllers/profile_controller_test.go
Would that be possible?

jlewi:
A unittest would be great.
/ok-to-test

lalithvaka:
> @lalithvaka thanks a lot for the PR!
> This seems like a great opportunity to add a couple of unit tests to that function, to ensure it yields a correct result.
> I would recommend the following cases:
> * only letters
> * only numbers
> * letters and numbers (starting with letter)
> * letters and numbers (starting with numbers)
> 
> Here are some details on table-driven tests in Go: https://github.com/golang/go/wiki/TableDrivenTests
> Here is an example from our code base: https://github.com/kubeflow/kubeflow/blob/c7150a18b3a567a7f1d59893a0ef2adc1245ae14/components/profile-controller/controllers/profile_controller_test.go
> Would that be possible?

Thank you for your help @yanniszark . Submitted the unit tests as well.

Cc
@jlewi

components/access-management/kfam/bindings.go
```diff
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	istiorbac "github.com/kubeflow/kubeflow/components/profile-controller/api/istiorbac/v1alpha1"
+
```
yanniszark, line 24: Can you remove this newline?

components/access-management/kfam/bindings.go
```diff
@@ -55,19 +57,20 @@ type BindingClient struct {
 
 //getBindingName returns bindingName, which is combination of user kind, username, RoleRef kind, RoleRef name.
 func getBindingName(binding *Binding) (string, error) {
+	// Only keep lower case letters, replace other with -
```
yanniszark, line 60: I believe we keep lower case letters and numbers with this regex.
Can you update the comment to reflect that?

components/access-management/kfam/bindings.go
```diff
@@ -164,6 +167,7 @@ func (c *BindingClient) Delete(binding *Binding) error {
 }
 
 func (c *BindingClient) List(user string, namespaces []string, role string) (*BindingEntries, error) {
+
```
yanniszark, line 170: Can you delete this newline?

components/access-management/kfam/bindings_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+// Building Binding Object from k8s.io/api/rbac/v1
+func getBindingObject(binding string) *Binding {
+
+	return &Binding{
+		User: &rbacv1.Subject{
+			Kind: "user",
```
yanniszark, line 14: nit: Can you use the `rbacv1.UserKind` constant instead?
It's a best practice to use the constants provided by the package, as any changes to the constant will be immediately reflected in our code.

components/access-management/kfam/bindings_test.go
```diff
@@ -0,0 +1,72 @@
...
+	in       *Binding
+	out      string
+	hasError bool
+}{
+	{getBindingObject("lalith.vaka@zq.msds.kp.org"), "user-lalith-vaka-zq-msds-kp-org-clusterrole-edit", false},
+	{getBindingObject("397401@zq.msds.kp.org"), "user-397401-zq-msds-kp-org-clusterrole-edit", false},
+	{getBindingObject("lalith.397401@zq.msds.kp.org"), "user-lalith-397401-zq-msds-kp-org-clusterrole-edit", false},
+	{getBindingObject("397401.vaka@zq.msds.kp.org"), "user-397401-vaka-zq-msds-kp-org-clusterrole-edit", false},
+	{getBindingObject("i397401@zq.msds.kp.org"), "user-i397401-zq-msds-kp-org-clusterrole-edit", false},
+}
```
yanniszark, line 36: Since these tests belong to the `TestGetBindingName` function, should we move this variable declaration inside the function instead of being a global one?

components/access-management/kfam/bindings_test.go
```diff
@@ -0,0 +1,72 @@
...
+	file, err := os.Open("./testdata/file.go")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	*/
+	//format := "--- %s: %s (%s)\n"
+	for _, tt := range tests {
+
+		s, error := getBindingName(tt.in)
```
yanniszark, line 52: Since `error` is an already declared interface type, let's use a different name for this variable.
A common name for error variables in Go is `err`.

components/access-management/kfam/bindings_test.go
```diff
@@ -0,0 +1,72 @@
...
+	/* Read the test data from a file if needed
+
+	// create a testdata folder under this package and add files as needed
+	file, err := os.Open("./testdata/file.go")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	*/
+	//format := "--- %s: %s (%s)\n"
```
yanniszark, line 49: Should this be removed?

components/access-management/kfam/bindings_test.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, tt := range tests {
+
+		s, error := getBindingName(tt.in)
+
+		if tt.hasError {
+			// expected an error
+			if error == nil {
+				t.Errorf("got %q, want %q", tt.in, s)
+			} else {
+				//t.Logf("getBindingName() with args %v PASSED and got value '%v'", tt.in, error)
```
yanniszark, line 59: Should this else clause be removed?

components/access-management/kfam/bindings_test.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+
+		} else {
+			//expected a value
+			if s != tt.out {
+				t.Errorf("got %q, want %q", s, tt.out)
+			} else {
+				//t.Logf("getBindingName() with args %v PASSED and got value '%v'", tt.in, s)
```
yanniszark, line 67: Should this else clause be removed?

components/access-management/kfam/bindings_test.go
lalithvaka: I left this behind for future enhancements if someone is interested in using a file based table test data and do not know how.

components/access-management/kfam/bindings_test.go
lalithvaka: agree

components/access-management/kfam/bindings_test.go
lalithvaka: Changed the error name

components/access-management/kfam/bindings_test.go
lalithvaka: Changed

components/access-management/kfam/bindings.go
lalithvaka: sure

components/access-management/kfam/bindings.go
lalithvaka: changed

components/access-management/kfam/bindings.go
lalithvaka: yes

components/access-management/kfam/bindings_test.go
```diff
@@ -44,27 +31,33 @@ func TestGetBindingName(t *testing.T) {
...
+	}
 
+	//format := "--- %s: %s (%s)\n"
+	for _, tt := range tests {
+		s, errorReturned := getBindingName(tt.in)
 		if tt.hasError {
 			// expected an error
-			if error == nil {
+			if errorReturned == nil {
 				t.Errorf("got %q, want %q", tt.in, s)
```
yanniszark, line 55: Last nit:
In this case an unexpected error happened, so the return values shouldn't matter much, but you probably want to see the error.
How about logging: `t.Errorf("Unexpected error occured: %v", errorReturned)`

components/access-management/kfam/bindings_test.go
yanniszark: Do you think a file-based method will be needed for this test?
In any case, I am ok either way, I will leave it up to you.

yanniszark:
@lalithvaka awesome, one small final nit and we should be ready to go!

yanniszark:
@jlewi @kunmingg since I'm not an approver for this code, can you help as well?

PR #4990, opened 2020-04-29
Don't show green check erroneously in JWA

### LINKED ISSUE #4712: Notebook Server UI Status shows Green even if Pod has failed
/kind bug

Created a Notebook server as in this issue #4706 

Status shows green but pod has failed

Status to show an error state

**Environment:**

- kfctl version:  v1.0-rc.1-0-g963c787
- Kubernetes platform: v1.15.7
- Kubernetes version: v1.15.7
- OS (e.g. from `/etc/os-release`): NAME="CentOS Linux"
VERSION="7 (Core)"

kimwnasptd:
closes #4712 

The most important change is in the Notebooks Controller. I've made the controller to use the `containerState` of the container that has the same name with the Notebook. This is currently the case if someone creates a Notebook via the Notebook Manager UI, but the users should be aware of this in case they create Notebooks by hand.

I would also like to make a small PR in the docs afterwards to make sure this information is more available to end users.

/cc @jlewi 
/assign @kimwnasptd

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -3,11 +3,12 @@ import json
 import logging
 import os
 import sys
-
-import yaml
 from collections import defaultdict
+
 from flask import request
 from kubernetes import client
+
+import yaml
 
 from . import api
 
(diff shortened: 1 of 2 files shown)


jlewi:
Is it possible to write unittests?

Issue #4808, opened 2020-02-28
Admission Webhook prevents Kubeflow pod deployment.


Tomcli:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
When `admission-webhook-mutating-webhook-configuration` deployed before the Admission Webhook pod is ready. The webhook is preventing pod deployment under the kubeflow namespace including the Admission Webhook pod. Thus, it permanently stops all the pod deployment unless we delete the `admission-webhook-mutating-webhook-configuration`.

**What did you expect to happen:**
admission-webhook at least should not prevent the Admission Webhook pod deployment. 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 1.0
- kfctl version: (use `kfctl version`): 1.0-rc4
- Kubernetes platform: (e.g. `minikube`)
- Kubernetes version: (use `kubectl version`): 1.16
- OS (e.g. from `/etc/os-release`):

jlewi:
@Tomcli is this an issue withe admission webhook not being properly scoped to only apply to profile created namespaces?

Tomcli:
Yes, there's no namespace selector for the admission webhook. So probably it's trying to watch pods from all namespaces.

animeshsingh:
think this is critical - we deployed Tekton with alongside Kubeflow on the same cluster, and it started watching that as well, which is precisely the kind of behavior we need to limit Kubeflow for.

Tomcli:
If we are going to add a namespace selector for the admission webhook, what label name would you prefer for the admission webhook.

yanniszark:
@Tomcli that's a good question!
Currently, here is the code making the namespaces: https://github.com/kubeflow/kubeflow/blob/d37bf05548bcf33ccd20c5c24e2b60068f90552b/components/profile-controller/controllers/profile_controller.go#L120-L130
It doesn't apply any kubeflow labels.
I think it's a good idea to label the profile-created namespaces with a specific label.
How about `app.kubernetes.io/part-of: kubeflow-profile`, inspired from the [kubernetes recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels).

jlewi:
@Tomcli Would you be willing to write a unittest?

Tomcli:
[...]

PR #4991, opened 2020-04-29
Add new profile namespace label for user namespaces

Tomcli:
Related https://github.com/kubeflow/kubeflow/issues/4808

This is one of the PR for adding a new label to indicate the Kubeflow-owned namespaces. We will be using this new `app.kubernetes.io/part-of: kubeflow-profile` as the namespace selector for the Admission Webhook. Once this PR is patched into the latest profile controller release, then we can update the Admission Webhook base manifest to the following so that it only watches namespaces with the kubeflow-profile label.
 
```yaml
apiVersion: admissionregistration.k8s.io/v1beta1
kind: MutatingWebhookConfiguration
metadata:
  name: mutating-webhook-configuration
webhooks:
- clientConfig:
    caBundle: ""
    service:
      name: $(podDefaultsServiceName)
      namespace: $(podDefaultsNamespace)
      path: /apply-poddefault
  name: $(podDefaultsDeploymentName).kubeflow.org
  namespaceSelector:
    matchLabels:
      app.kubernetes.io/part-of: kubeflow-profile
  rules:
  - apiGroups:
    - ""
    apiVersions:
    - v1
    operations:
    - CREATE
    resources:
    - pods
```

cc @animeshsingh

### DIFF
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -65,6 +65,7 @@ const (
 var kubeflowNamespaceLabels = map[string]string{
 	"katib-metricscollector-injection":      "enabled",
 	"serving.kubeflow.org/inferenceservice": "enabled",
+	"app.kubernetes.io/part-of":             "kubeflow-profile",
 }
 
 const DEFAULT_EDITOR = "default-editor"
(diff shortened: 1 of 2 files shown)


yanniszark:
@Tomcli thanks, this is very useful!
/lgtm

@jlewi I don't have approve powers in this context, so I will need you to also confirm.

Tomcli:
@jlewi do you think we can add this new label to all the namespaces created by profile?

PR #5015, opened 2020-05-14
Enable kfam, admission_webhook on arm64

MrXinWang:
This commit adds container building support for ingress_setup, kfam, and admission_webhook on arm64.

Signed-off-by: Henry Wang <Henry.Wang@arm.com>

### DIFF
--- a/components/access-management/Dockerfile
+++ b/components/access-management/Dockerfile
@@ -5,7 +5,12 @@ WORKDIR /workspace
 
 COPY . .
 RUN go mod download
-RUN go build -gcflags 'all=-N -l' -o access-management main.go
+RUN if [ "$(uname -m)" = "aarch64" ]; then \
+        CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -gcflags 'all=-N -l' -o access-management main.go; \
+    else \
+        CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -gcflags 'all=-N -l' -o access-management main.go; \
+    fi
+
 RUN chmod a+rx access-management
 
 # Use distroless as minimal base image to package the manager binary
--- a/components/admission-webhook/Dockerfile
+++ b/components/admission-webhook/Dockerfile
@@ -4,13 +4,17 @@ FROM golang:${GOLANG_VERSION} as builder
 
 # Copy in the go src
 WORKDIR /go/src/github.com/kubeflow/kubeflow/components/admission-webhook
-COPY pkg/  pkg/ 
+COPY pkg/  pkg/
 COPY . .
 
 ENV GO111MODULE=on
 
 # Build
-RUN CGO_ENABLED=0 GOOS=linux go build -o webhook -a .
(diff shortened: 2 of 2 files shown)


jlewi:
Is there an issue tracking this? In particular is there an issue spelling out the motivation for making things run on Arm?

MrXinWang:
Hi @jlewi ! Thanks for the reply :))

Yes of course, please see [kubeflow#2337](https://github.com/kubeflow/kubeflow/issues/2337) and these two comments made by myself ([#1](https://github.com/kubeflow/kubeflow/issues/2337#issuecomment-615005699) [#2](https://github.com/kubeflow/kubeflow/issues/2337#issuecomment-628400573)) about current status on arm64. Thanks!

components/ingress-setup-image/Dockerfile.base.arm64
```diff
@@ -0,0 +1,39 @@
+FROM docker:17.12.0-ce as static-docker-source
```
jlewi, line 1: I believe this docker image is specific to setting up IAP on GCP.
So I'm not really sure it makes sense to provide an arm64 variant.

MrXinWang: Hi @jlewi , thanks for reviewing! The reason why I added this base image is that `google/cloud-sdk:alpine` is an x86_64 only image. In order to build this `ingress-setup-image` on arm64, I have to add this google cloud base image and use `docker --buildarg` to avoid breaking existing x86_64 building strategy.

My previous idea before our discussion in [kubeflow#2337](https://github.com/kubeflow/kubeflow/issues/2337) was to support as much component as I could. So if you think it is unnecessary to add arm64 support for this `ingress-setup-image`, I would also happy to remove this.

jlewi: Yes. Please remove. This image is only used for GCP. If you are on GCP you aren't running on Arm64 so there's no reason to maintain an Arm64 variant of this image.

components/ingress-setup-image/README.md
```diff
@@ -2,9 +2,18 @@
...
 ```
 docker build . -t gcr.io/kubeflow-images-public/ingress-setup:latest
 docker push gcr.io/kubeflow-images-public/ingress-setup:latest
 ```
+
+on AArch64 machines, run
+
+```
+docker build . -t google-cloud-sdk:debian10 -f Dockerfile.base.arm64
+docker build . -t gcr.io/kubeflow-images-public/ingress-setup:latest --build-arg BASE_IMAGE=google-cloud-sdk:debian10
```
jlewi, line 16: Please revert for reasons mentioned in other comment.

components/ingress-setup-image/Dockerfile.base.arm64
MrXinWang: Thanks for the explanation @jlewi. The `ingress-setup-image` code has been reverted.

components/ingress-setup-image/README.md
MrXinWang: Done.

PR #5016, opened 2020-05-14
Bumping central dashboard node version to v12.18.3

MrXinWang:
When building current master branch central dashboard on arm64, error message `Segmentation fault (core dumped)` will occur. By bumping base image node version to 10.13, this error can be fixed.

I also tested this commit on x86_64 machine to make sure this will not break current x86_64 build.

Signed-off-by: Henry Wang <Henry.Wang@arm.com>

### DIFF
--- a/components/centraldashboard/Dockerfile
+++ b/components/centraldashboard/Dockerfile
@@ -1,5 +1,5 @@
 # Step 1: Builds and tests
-FROM node:10-alpine AS build
+FROM node:12.18.3-alpine AS build
 
 ARG kubeflowversion
 ARG commit
@@ -39,7 +39,7 @@ RUN npm rebuild && \
     npm prune --production
 
 # Step 2: Packages assets for serving
-FROM node:10-alpine AS serve
+FROM node:12.18.3-alpine AS serve
 
 ENV NODE_ENV=production
 WORKDIR /app


MrXinWang:
Hi @prodonjs @swiftdiaries ! Any comments?

axelgobletbdr:
Without this fix, centraldashboard needs to run as privileged pod on Kubernetes. This is a severe security issue. Therefore, we would be very happy if this PR was merged.

axelgobletbdr:
@MrXinWang I see that there are also even newer node versions available. Why did you choose this exact version? Maybe it would make the PR even stronger if an even higher node version was used.

MrXinWang:
Thanks @axelgobletbdr , do you know any reviewer(s) would be interested in this? Probably we can let them take a look.

axelgobletbdr:
@avdaredevil @swiftdiaries can you look into this PR? It solves a big security risk in central dashboard.

components/centraldashboard/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Step 1: Builds and tests
-FROM node:10-alpine AS build
+FROM node:10.13-alpine AS build
```
Bobgy, line 2: https://hub.docker.com/_/node
node:10-alpine is node 10.22

Why do we need to revert to node:10.13?

Bobgy:
/assign @Bobgy 
I can help with the review

StevenReitsma:
Some context of people with similar issues with the `node-alpine` image: 
- https://github.com/nodejs/docker-node/issues/1158
- https://github.com/google/gvisor/issues/267#issuecomment-562204277

The issue seems to be that Node tries to use the `membarrier` syscall in Alpine-based images, which is not allowed on secure clusters.

I don't believe downgrading to 10.13 would be the solution but using `node-slim` instead of `node-alpine` should work. We can test whether that solves the issue on our end @axelgobletbdr?

MrXinWang:
> I don't believe downgrading to 10.13 would be the solution but using node-slim instead of node-alpine should work.

Thanks @StevenReitsma ! I remember the original goal of this PR is to solve the docker image building on arm64, but please do test it :) If convert the base image to `node-slim` would help, I can update this PR.

> Why do we need to revert to node:10.13?
Hi @Bobgy ! If I remember the story clearly, when this PR was proposed (several months ago), the version of node was not this high, therefore at that time the version proposed by this PR was an upgrade....I will try to test if this PR is still valid with the lastest version ASAP. Thanks!

axelgobletbdr:
I agree that downgrading is not the right decision here @StevenReitsma . I also understand that things have changed in these 3 months @MrXinWang . @MrXinWang can you test whether @StevenReitsma's proposed changes would solve the issue?

swiftdiaries:
Can we upgrade to Node's LTS version `v12.18.3`? @MrXinWang 

@axelgobletbdr the original intent for this PR was to support ARM builds. Could you open a separate PR with the intent of addressing the security issue? I believe it'll be better for lineage tracking of changes 

WDYT? @Bobgy

MrXinWang:
@swiftdiaries As long as the change won't break anything, I am happy with any version :) v12.8.3 seems perfect to me.

@axelgobletbdr Yes of course! I will do it ASAP tomorrow when I am back to the office :)

swiftdiaries:
Awesome ! Thank you so much for the quick reply :) 
Sorry for the super late turn around in merging this PR

axelgobletbdr:
@swiftdiaries I created a separate issue for the security problem: https://github.com/kubeflow/kubeflow/issues/5188. I tested bumping the base image to `node:12.18.3-alpine` and can confirm that the centraldashboard can now run with restricted permissions. A separate PR should not be necessary here as this PR will change the same lines of code.

@MrXinWang can you test whether  `node:12.18.3-alpine`  solves your problems as well? If so, please adapt the PR such that it can be merged.

swiftdiaries:
@axelgobletbdr 🙌🏾 thanks for the quick work testing that out

MrXinWang:
@axelgobletbdr @swiftdiaries @Bobgy Rebased and updated. PTAL, thanks!

swiftdiaries:
/approve 
Thank you for the quick turn around @MrXinWang 🙌🏾

StevenReitsma:
Thanks all!

PR #5019, opened 2020-05-18
Add yanniszark to access-management, profile-controller

yanniszark:
Adding `yanniszark` to access-management, profile-controller components.
I am familiar with the codebase and they are also directly related with the multi-user cuj, which I'm very familiar with.
This way, I can be of help to people contributing fixes to these components.

/cc @jlewi 

### DIFF
--- a/components/access-management/OWNERS
+++ b/components/access-management/OWNERS
@@ -1,4 +1,3 @@
 approvers:
-  - kunmingg
   - yanniszark
 reviewers:


components/access-management/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+  - kunmingg
```
jlewi, line 2: Remove @kunmingg; @kunmingg is focused on other things so this is all you for the time being.

Also don't duplicate approvers under reviewers; an approver is automatically a reviewer.

yanniszark:
@jlewi thanks for the review!
I amended the PR and only added myself to approvers.

PR #5069, opened 2020-06-15
tensorboard-controller: Mount GCP secret only when accessing Google storage

### LINKED ISSUE #5065: Tensorboard controller creates servers that always mount user-gcp-sa secret
/kind bug

Currently, `user-gcp-sa` secret [is always mounted](https://github.com/kubeflow/kubeflow/blob/master/components/tensorboard-controller/controllers/tensorboard_controller.go#L131-L147) by the pod running the Tensorboard server, regardless of the logs storage being a Google cloud bucket or not.

This secret is required only when accessing a Google storage bucket, so we should limit the secret mount to that use case.

I've bumped into this bug while working on the controller as part of my [Google Summer of Code project](https://summerofcode.withgoogle.com/projects/#4660815581413376) for building a Tensorboard UI that also supports PVCs. So, I will prepare a PR to fix this behavior.

cc @jlewi @kimwnasptd @elikatsis

kandrio98:
Closes #5065

The PR also contains a minor fix related to double import of a package.

### DIFF
--- a/components/tensorboard-controller/controllers/tensorboard_controller.go
+++ b/components/tensorboard-controller/controllers/tensorboard_controller.go
@@ -127,40 +127,39 @@ func (r *TensorboardReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func generateDeployment(tb *tensorboardv1alpha1.Tensorboard, log logr.Logger) *appsv1.Deployment {
-	volumeMounts := []corev1.VolumeMount{
-		{
+	var volumeMounts []corev1.VolumeMount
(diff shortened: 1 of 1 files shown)


kimwnasptd:
The code looks good so its LGTM from me.
/lgtm

@jlewi I can see that we don't have an OWNERS file for the [tensorboard-controller](https://github.com/kubeflow/kubeflow/tree/master/components/tensorboard-controller), so I'll leave the approve part to you. Should we create one for this component?

jlewi:
@kimwnasptd yes please create an OWNERs file

/approve

PR #5083, opened 2020-06-19
Fix notebook controller rbac gen

HumairAK:
[make manifest](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/Makefile#L48) for `notebook-controller` doesn't seem to generate the rbac files as described [here](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L75). Adding a new line after the gen markers seems to solve this problem!


HumairAK:
@googlebot I signed it!

Issue #4883, opened 2020-03-23
Do not allow a Notebook Server to be created if Existing Workspace or Data Volumes is selected and the Volume does not exist.


jtfogarty:
/kind feature

**Why you need this feature:**
This will prevent the user from attempting to create a Notebook server that will not start

**Describe the solution you'd like:**
When the user select Existing for Workspace or Data Volumes, do not allow the user to type the volume name, Only allow them to select a name

**Anything else you would like to add:**
none

lalithvaka:
I totally agree with this issue. I see our teams doing this mistake and expecting the notebook servers to come up. Existing option should be followed by a drop down only. Or we should. think about redesigning the volume selection process.
cc @avdaredevil

PR #5086, opened 2020-06-21
show all existing PVCs as dropdown under volumes for a new notebook server

lalithvaka:
Resolves issue #4883 
This PR enhances two things
1. Shows all Existing PVCs in the profile as a dropdown while creating a new notebook server under both workspace / data volumes when type drop down selected as "Existing"
2. Hide Mode and Size form fields under volumes when selected existing PVCs so that its not confusing

Follow are the screenshots of before and after
**Before the fix**
image

**After the fix**
image

### DIFF
--- a/components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.html
+++ b/components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.html
@@ -1,42 +1,49 @@
 <div *ngIf="volume" [formGroup]="volume" class="volume-wrapper">
-  <mat-form-field appearance="outline" id="type">
-    <mat-label>Type</mat-label>
-    <mat-select formControlName="type">
-      <mat-option
+ 
+  <!-- Volume Type Input -->
+    <mat-form-field appearance="outline" id="type">
+      <mat-label>Type</mat-label>
+      <mat-select  (selectionChange)="selectType($event)" formControlName="type">
+        <mat-option *ngFor="let i of types"
         [disabled]="newTypeIsDisabled()"
         [matTooltip]="newTypeTooltip()"
-        value="New"
-        >New</mat-option
-      >
-      <mat-option value="Existing">Existing</mat-option>
-    </mat-select>
-  </mat-form-field>
+        value={{i}}>{{i}}</mat-option>
+      </mat-select>
+    </mat-form-field>
+  
+    <!-- Volume Name Input -->
+    <mat-form-field *ngIf="typeSelected === 'New'" class="wide" appearance="outline" id="name">
+      <mat-label>Name</mat-label>
+      <input matInput formControlName="name"/>
+    </mat-form-field>
+    
+    <mat-form-field *ngIf="typeSelected === 'Existing'" class="wide" appearance="outline" id="name">
+      <mat-label>Name</mat-label>
+      <mat-select formControlName="name">
+        <mat-option *ngFor="let i of existingPVCs"
+          value={{i}}>{{i}}</mat-option>
+      </mat-select>
+    </mat-form-field>
 
-  <!-- Volume Name Input -->
-  <mat-form-field appearance="outline" id="name">
-    <mat-label>Name</mat-label>
-    <input matInput formControlName="name" />
-  </mat-form-field>
-
-  <!-- Size Input -->
-  <mat-form-field appearance="outline" id="size">
-    <mat-label>Size</mat-label>
-    <input matInput formControlName="size" />
-  </mat-form-field>
-
-  <!-- Mode Input -->
-  <mat-form-field appearance="outline" id="mode">
-    <mat-label>Mode</mat-label>
-    <mat-select formControlName="mode">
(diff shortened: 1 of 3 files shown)


lalithvaka:
@avdaredevil , please let me know if you can review this PR as well?

components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.html
```diff
@@ -1,42 +1,49 @@
...
+        <mat-option *ngFor="let i of types"
         [disabled]="newTypeIsDisabled()"
         [matTooltip]="newTypeTooltip()"
-        value="New"
-        >New</mat-option
-      >
-      <mat-option value="Existing">Existing</mat-option>
-    </mat-select>
-  </mat-form-field>
+        value={{i}}>{{i}}</mat-option>
```
avdaredevil, line 10:
```suggestion
        value="{{i}}">{{i}}</mat-option>
```

components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.ts
```diff
@@ -58,6 +63,14 @@ export class VolumeComponent implements OnInit, OnDestroy {
     }
   }
 
+// ----- onChange of the New / Existing Volume -----
+  selectType(event): void {
+    this.typeSelected = event.value;
+    if (this.typeSelected === 'New') {
+      this.volume.controls.name.setValue(this.currentVolName);
+    }
```
avdaredevil, line 71: *nit*: Prevent unnecessary code depth
```suggestion
    if (this.typeSelected !== 'New') return;
    this.volume.controls.name.setValue(this.currentVolName);
```

components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.html
lalithvaka: Will test this and commit again - Definitely not an expert on right way to code this.

components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.ts
lalithvaka: Will test this and commit again

components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.html
lalithvaka: Submitted the commit again after testing.

components/jupyter-web-app/frontend/src/app/resource-form/volume/volume.component.ts
lalithvaka: Updated and submitted the commit

kimwnasptd:
@avdaredevil thanks for taking a look!

@lalithvaka I'll take a more thorough look at this asap.
With a first glance could you make the `Mount Point` input wide enough to fill the space left from the two previous inputs, when selecting an existing volume?

lalithvaka:
@kimwnasptd Can you please review this? If you have any additional changes, I will submit the pull request. Am wondering if this can be pushed to 1.1 release as well!

lalithvaka:
@kimwnasptd , any progress on this? Thank you.

kimwnasptd:
@lalithvaka my only objection is about removing the inputs since the delete button looks too spaced out in the screenshots.
I'd suggest we keep the inputs as is, they are disabled either way so users wouldn't be able to modify them, and lets have a separate PR for hiding them.

the code is LGTM also from me

lalithvaka:
@kimwnasptd , I have tested this feature with our end user community. They all liked the dropdown but they got confused on the "Size" and "Mode" fields. They were all assuming that any existing PVC they pick, they are expecting the fields disabled are reflection of what the actual size and mode of the selected PVC which is not the case. Hence to remove the confusion, I got rid of the fields on the selection of Existing PVCs. I totally agree with you that the spacing of the delete button needs some adjustment. 

Please let me know if you can approve or merge this?

avdaredevil:
Moving ahead on this in favor of getting it in 1.1
/lgtm
/approve

Thanks for making this change @lalithvaka!

/cc @jlewi

jlewi:
@lalithvaka Looks like there are test failures

lalithvaka:
@jlewi , @kimwnasptd , @avdaredevil , pre submit test failures are fixed. Can you please lgtm / approve this PR? I would like to submit a cherry pick for this into 1.1. Thank you.

thesuperzapper:
@lalithvaka there is an issue with this if you type the name of a notebook which has existed in past, the UI gets a bit confused, and changes to "existing" but leaves the "name" field as text field, rather than a dropdown.

lalithvaka:
@thesuperzapper That was the behavior before I submitted the PR to show the drop down list for the existing PVCs. I thought about how to address that but couldn't come up with a better option with out totally revamping the existing design.  So I left it as it just shows the user that name is an existing PVC. I am not a UX/UI expert, if you think of anything better, please go ahead and submit an additional PR.

PR #5088, opened 2020-06-23
Add OWNERs file to tensorboard controller

kimwnasptd:
We need to have an OWNERs file for the Tensorboard Controller

Since me, @elikatsis and @kandrio98 have taken an extensive look on the inner workings of the tensorboard controller, as part of the [GSoC Project](https://summerofcode.withgoogle.com/projects/#4660815581413376) , I'll add us to this file. 

@jlewi @quanjielin please take a look and tell me if the proposed roles are OK with you.

/cc @jlewi @quanjielin 
/assign @kimwnasptd

### DIFF
--- a/components/tensorboard-controller/OWNERS
+++ b/components/tensorboard-controller/OWNERS
@@ -1,0 +1,5 @@
+approvers:
+  - elikatsis
+  - kandrio98
+  - kimwnasptd
+reviewers:


components/tensorboard-controller/OWNERS
```diff
@@ -0,0 +1,10 @@
+approvers:
+  - elikatsis
+  - kandrio98
+  - kimwnasptd
+  - quanjielin
```
jlewi, line 5: I don't believe @quanjielin is working on this anymore.

components/tensorboard-controller/OWNERS
```diff
@@ -0,0 +1,10 @@
...
+reviewers:
```
jlewi, line 6: approvers shouldn't be listed as approvers

components/tensorboard-controller/OWNERS
kimwnasptd: Do you propose to not include him in approvers, reviewers or both?

components/tensorboard-controller/OWNERS
kimwnasptd: I think I'm missing something here, where should approvers be listed?

jlewi: My bad typo; meant to say that approvers should be listed as reviewers.

components/tensorboard-controller/OWNERS
jlewi: Both; I don't think she is working on KF anymore; 

@quanjielin are you still working on TB controller or can we remove you from the OWNERs file?

quanjielin: feel free to remove me from the owner files.

kimwnasptd: ACK, I removed her from the OWNERs file

components/tensorboard-controller/OWNERS
kimwnasptd: I see. Which people should I assign as reviewers?

jlewi: @kimwnasptd You can just leave it blank if there is no one that makes sense to add as a reviewer.

kimwnasptd:
@jlewi the PR has passed the tests, could you give this again a LGTM?

PR #5090, opened 2020-06-24
Update kubeflow-fairing version in notebook image

jinchihe:
Update kubeflow-fairing version to 1.0.0 in notebook image


jlewi:
@jinchihe If you want these changes in the 1.1 images please see if there is an open PR for building the docker images for the notebooks and use that to track ensuring this PR makes it in.

PR #5108, opened 2020-07-06
Remove sync_images.py

marload:
The existing late return has been changed to an early return.

Thank You!


jlewi:
This script is no longer used. Do you want to just delete it?

kfctl now has a subcommand "mirror" which outputs either Cloud Build or Tekton workflows to mirror your images to a different registry.

marload:
@jlewi 
Are you just saying to delete file `sync_image.py`?

jlewi:
Yes

PR #5109, opened 2020-07-06
Remove test_deploy.py

marload:
I removed duplication. Maintenance will be easier.

Thanks You!


jlewi:
Can we just delete this script? I don't believe we are using it anymore since it is using ksonnet to deploy and we don't use ksonnet to deploy Kubeflow anymore.

marload:
@jlewi 
I applied!

PR #5110, opened 2020-07-06
Refactoring: Deduplication

marload:
Unnecessary Raise has been removed.

Thanks You!


jlewi:
See comment in #5109 can we just delete this script?

marload:
@jlewi 
Are you just saying to delete file `test_deploy.py`?

PR #5118, opened 2020-07-08
Fix #4842 Pod resources limit

### LINKED ISSUE #4842: Pod resources limit
/kind bug

When I create jupyter notebook container. I found in pod yaml that the cpu num setting in jupyter create page is equivalent to resource request items. So this means no cpu limit is set in pod. For a multi-user server, a single user may use 100% cpu. Other users' task and whole kubeflow platform will hang up. No enough resources will also cause k8s kill some system pods or pod health check failed and these pods will repeat to restart automatically again and again.   
`
"resources": { 
     "limits": { 
        "nvidia.com/gpu": "4"
    },
    "requests": {
        "cpu": "16",
        "memory": "92Gi"
    }
`
The cpu and memory resource setting in jupyter create page should be resources limit items in pod yaml as GPU settings have done.

In addition, when I use kubeflow V0.4 last year, this limit feature is correct.

**Environment:**

- Kubeflow version: 1.0.0
- kfctl version:1.0.0
- Kubernetes platform:  Microk8s 1.15
- Kubernetes version: 1.15.10
- OS : Ubuntu 18.04

axelgobletbdr:
Fix #4842 - issue was closed earlier, but not solved within Kubeflow repo by author

Sets CPU and memory limits of Notebooks equal to requests. This prevents
- out of memory problems due to over-provisioning
- users using more resources than they should

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -369,6 +369,7 @@ def set_notebook_cpu(notebook, body, defaults):
         logger.info("Using default CPU: " + cpu)
 
     container["resources"]["requests"]["cpu"] = cpu
+    container["resources"]["limits"]["cpu"] = cpu
 
 
 def set_notebook_memory(notebook, body, defaults):
@@ -385,6 +386,7 @@ def set_notebook_memory(notebook, body, defaults):
         logger.info("Using default Memory: " + memory)
 
     container["resources"]["requests"]["memory"] = memory
+    container["resources"]["limits"]["memory"] = memory
 
 
 def set_notebook_gpus(notebook, body, defaults):
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/yaml/notebook.yaml
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/yaml/notebook.yaml
@@ -18,5 +18,8 @@ spec:
             requests:
               cpu: "0.1"
               memory: "0.1Gi"
+            limits:
+              cpu: "0.1"
+              memory: "0.1Gi"
       ttlSecondsAfterFinished: 300
       volumes: []


axelgobletbdr:
@vkoukis @prodonjs could you look into this PR?

axelgobletbdr:
@StefanoFioravanzo @SachinVarghese @lalithvaka this PR has been open for several weeks now. Could you or some other maintainers look into this?

apryiomka:
what about the resource limits on the Notebook CR? It currently does not specify one?

axelgobletbdr:
Hi @apryiomka 

The notebook CRD already has the option to add limits: https://github.com/kubeflow/kubeflow/blob/f100b901199113e669974a889c69ad297cae7b9d/components/notebook-controller/config/crd/bases/kubeflow.org_notebooks.yaml#L1370

In this file, I also added limit defaults to the notebook CR (which will most of the time get overriden by the jupyter web app Python service: https://github.com/kubeflow/kubeflow/blob/730fd014fa43162943c117d3404556d03b048ac2/components/jupyter-web-app/backend/kubeflow_jupyter/common/yaml/notebook.yaml#L21

Can you elaborate more on your exact concerns?

apryiomka:
@axelgobletbdr, i would argue most of the notebooks will not be submitted using CR, but using the Notebook Web App. Meaning the web app needs to impose the limits when the CR is created and submitted to k8s.

axelgobletbdr:
@apryiomka no problem! It is good to be thorough and ask questions

axelgobletbdr:
@kimwnasptd can you look into this PR?

kimwnasptd:
@axelgobletbdr the code looks good. Ideally the UI should fully expose to the users both `requests` and `limits` and allow them to fine tune both.

Lets merge this as is for now and later on, depending on user's feedback, we could:
* Extend the UI to expose both limits and resources
* Extends the web app's config to allow admins to configure these values

axelgobletbdr:
@kimwnasptd I agree that setting the requests and limits to the same value is not always optimal. For now, this is a safe option that will give predictable performance of your cluster. 

Option 1 is only a good idea if the UI user knows the impact of setting limits > requests. I doubt that this will be the case for many users. Advanced users can always decide to create a Notebook resource. 

Option 2 might be a more suitable option, if users need a different request-limit ratio in the future.

Thanks again for merging the request!

apryiomka:
i agree on the UI part, but also keep in mind that most of the Notebook users are Data Scientists who care not about the difference between the two. Most they care about is _i need this number of CPUs, GPUs and GB of memory_. They don't necessarily anticipate the notebook container going over those resource. If that is the case, they can aways create a larger notebook.

I do think that requests / limit is more suitable for none-UI / user base workflows, for instance, i have a service that i don't want to go down in case my requested resources exceeded. In this case it makes sense to have limit larger than the request.

thesuperzapper:
I think setting a limit is an **extremely bad idea**, as it can result in an unexpected shutdown of notebook instance (and losing anything currently in memory on the pod).

There is also an important distinction to make about what a "request" actually means. Lets not forget that in the end we are just spawning Pods on Docker, **which has access to all of the host's resources**, meaning the code running in the pod has no idea that it will be suddenly killed if it uses more than some arbitrary number.

@jlewi This PR needs to be reverted.

thesuperzapper:
I have made #5226 to revert this change.

axelgobletbdr:
Hi @thesuperzapper, setting requests == limits will:
- cause your pod to shutdown if you use more memory than your limit. This way, you will lose unsaved changes. Jupyter occasionally autosaves, so the damage should be minimal. Furthermore, you know how much memory you are allowed to use (you set the limit yourself), so you should be aware which operation might cause an OOM. 
- never cause the node run out of memory, since scheduling occurs based on resource requests. Therefore, your pod will only crash if you do not adhere to your own resource limits, rather than by the fault of some other user on the same node.

setting limits > requests (or omitting limits completely) will:
- allow you to use more resources than you scheduled, if that is necessary occasionally. This might make resource utilization better, but makes your physical limit dependent on the memory usage of other pods. 
- can crash your pod even though you did not hit your memory limits (because node is OOM), causing very unexpected behavior. 

All in all, I think setting limits close or equal to requests makes the whole system much more predictable.

What exact concerns do you have? I think a pod shutdown is very expected behavior if I use more resources than I requested.

thesuperzapper:
@alexlatchford 
Firstly, I think this limits feature should be put behind a config in `spawner_ui_config.yaml` (defaulting to off), as it significantly changes how notebooks are provisioned. (Which is why I have made PR: #5226)

**In response to your points:**
1. Just because a notebook is saved, dose not mean it is safe to suddenly kill the Notebook pod.
1. Data Scientists run LONG jobs, and when the pod is suddenly killed, they have to run all code again.
1. Jupyter is not the only notebook type people run in Kubeflow. (We also run RStudio and Theia) 

**Solution:**
* From a resource perspective, notebooks really need to be scheduled on their own node. They are **development environments** which allow arbitrary code execution (not well behaved web microservices).
* In my deployments, users pick if they need 'shared resources' (where they will contend with other notebooks), or if they need a 'solo resources' (if they need an entire node to themselves).
* I am currently working on a PR to introduce `NodeSelectors`/`Tolerations` to the notebook spawner UI, which will make this possible for everyone else to do. NOTE: in this setup you don't even let the user set a resource request, as they request a node type using the `NodeSelectors`

axelgobletbdr:
HI @thesuperzapper, I see that your use case is different from the use cases of the other people in this thread, and I do respect that. I understand that a toggle for this feature helps your use case. I think it would me more constructive to make an issue for that with good arguments, rather than reverting this commit, which many others value. Could me make a separate issue? Then we can discuss this matter further in that issue and create a solution that works for anyone.

thesuperzapper:
@axelgobletbdr I fail to see how you have users which will accept their pods being randomly killed due to spiking over a resource limit?

axelgobletbdr:
@thesuperzapper, a node has finite memory. There will always be a point where a pod will crash due to memory issues. This can be when the node is full (like in your case, when you have 1 notebook per node) or when you hit your pod resource limit. The behavior is the same for both cases: you have a memory budget and you have to stay below it or else you will suffer the consequences. 

When only crashing when the node hits an OOM (the case you would like to see), you will have more headroom for memory spikes, that is correct. On the other hand, if you are working with multiple users on the same node, other users can 'steal' your memory, making the amount of memory you can use less obvious. If two users decide to load a large dataset on the same time, they will suffer the consequences. When only scheduling 1 user per node, you do not have this issue, but this is not a feature of kubeflow at this point. Also, depending on the nodes you are using, it is not feasible to run 1 user per node. We for example run a rather small number of pretty large nodes. These nodes are way too big for the average workload of a user. We would need a lot of these machines and they would have a lot of idle resources when we schedule one user per node. 

You should only use what you request, such that other users can have a pleasant experience within the cluster as well. Therefore we would like to keep these limits (potentially with a toggle that we will discuss in the other story) within kubeflow.

You can always set the requests to the size of your nodes if you want to schedule only one user per node. Then your limits are equal to the node limits as well, and you have only one user per node. Then the issues you are having are resolved as well, and you are following the standard way of how kubernetes virtualizes resources.

thesuperzapper:
@axelgobletbdr, the PR which adds tolerations/affinity is: #5237

thesuperzapper:
For those watching, PR #5280 has put this feature behind config flags which are disabled by default: 
* `spawnerFormDefaults.cpu.setLimit`
* `spawnerFormDefaults.memory.setLimit`

Additionally, warnings are displayed to users in the Notebook Spawner UI if those flags are enabled:

picture_6

PR #5139, opened 2020-07-16
Update kubeflow-fairing to avoid mismatching with kfserving

jinchihe:
Due to the issue https://github.com/kubeflow/examples/issues/806
we need to upgrade kubeflow-fairing to v1.0.1, thanks.


jinchihe:
/assign @jlewi
/assign @pdmack 

Thanks.

PR #5142, opened 2020-07-16
Deduplicate contributing docs

alexlatchford:
Similar vain as: https://github.com/kubeflow/community/pull/185

Following the link from the `k8s-ci-robot` suggesting to join the community links (see https://github.com/kubeflow/pipelines/pull/4228#issuecomment-659685533) to this page but this page doesn't actually have any instructions on how to join the community!


alexlatchford:
Ah looks like because I used the Github editor to produce the change it didn't set my email in the `git commit` and thus didn't sign the CLA, is this a blocker? See here for verification: https://github.com/kubeflow/pipelines/pull/4228#event-3555251381

Happy to reopen with a proper commit if so.

PR #5074, opened 2020-06-16
Added gpu details to the notebook server table

lalithvaka:
Resolves issue #5073 
Current Notebook server dashboard displays CPU, Memory and Image name but not the GPUs used. Displaying GPUs used by notebook servers help avoid confusion with in a project team on whether the resources are still available to use. This pull request adds both GPUs as well as GPU tooltip.
Screenshot
image

PR #5162, opened 2020-07-24
Automated cherry pick of #5074: added gpu details to the notebook server table #5086: show all existing PVCs as dropdown under volumes while Cherry pick of #5074 #5086 on v1.1-branch. #5074: added gpu details to the notebook server table #5086: show all existing PVCs as dropdown under volumes while

lalithvaka:


jlewi:
@lalithvaka Should this be the last cherry-picking of anything which isn't a critical bug fix? Perhaps we should start thinking about 1.2?

As is this cherry-pick will likely not be included in 1.1 because it arrived after 1.1-RC.2 which is the one we are pushing.

lalithvaka:
@jlewi this is the last one am tracking besides the documentation [PR]() https://github.com/kubeflow/website/pull/2032 for some of these changes.

Issue #4758, opened 2020-02-12
Persistent Volume Management UI


jl-massey:
/kind feature

**Why you need this feature:**
We have a large number of Kubeflow users for our on-premises installation.  We find that creating notebooks, deleting, then recreating leaves orphaned PV's.   If a notebook is deleted, then creating a new notebook requires great memory or docs for the previously used volume name.

**Describe the solution you'd like:**
From the kubeflow central dashboard, either add a resource management page (accessed from tab on the left) OR a panel to the initial page that lists all persistent volumes in the selected namespace.   Next to each volume, the ability to copy the name, see statistics (size, type, etc.) and delete (with confirmation) unused volumes.

**Anything else you would like to add:**
This is something that is needed to ensure our focus remains on USING the platform, rather than playing cleanup behind dozens of users, and answer requests to find 'lost' volumes.

jbottum:
@jl-massey Hi, we (Arrikto) has been scoping this requirement.    We would like to set-up some time to discuss.   Please let me know if you would like to set-up at time next week.

kimwnasptd:
@jl-massey we (Arrikto) also believe that currently the handling of PVCs/PVs is a pain point.

We have developed a Managment UI, similar to the one for managing Notebooks, that will handle PVCs. This will allow users to keep track of the PVCs in the different Namespaces and perform CRUD actions from a UI.

I'm planning to demo this in the following community meeting on Tuesday, June 9th. Feel free to join to see it in action.

Anujay-Saraf:
**@kimwnasptd** This is really helpful in enterprise level as well, as PVC's management time is a pain and can you please share the process/manifest to add this volume manager UI, with existing or new kf installation process along with GKE based cluster, similar to your **MiniKF release**.

https://journal.arrikto.com/democratizing-the-use-of-pvcs-with-the-introduction-of-a-volume-manager-ui-a5bf616570d3

DavidSpek:
To add on to this. I have been working on adding support to Kale for taking snapshots for storage backends that support this using the generic kubernetes implementation. Once that is further along it would also greatly benefit from a UI. The logic to restore a notebook from (one of) the PVCs that was attached to it is already there and working, it's mainly the implementation of what metadata to use and where to place it that needs to be looked at. 
https://github.com/kubeflow-kale/kale/pull/217

bravo325806:
This Volume Manager UI can be deployed on on-presmises kubernetes?
[...]

Issue #3578, opened 2019-06-30
Web-app to manage tensorboard instances 


jlewi:
We should consider creating a standalone web app similar to our jupyter web app that makes it easy for folks to create/delete tensorboard instances.

I suspect copying and modifying the jupyter web app to create a web app for managing tensorboard would be pretty straightforward.

Some pointers to get started

* Here is a link to the current [jupyter web app code](https://github.com/kubeflow/kubeflow/tree/master/components/jupyter-web-app)
* The notebook in kubeflow/examples#/723 has an example of deploying TensorBoard by creating
   1. An ISTIO virtual service
   1. A K8s service
   1. A Deployment
* The code for the tensorboard controller is [here](https://github.com/kubeflow/kubeflow/tree/master/components/tensorboard-controller)

This project could be broken down into several pieces

1. Create a Web App for spinning up TensorBoard instances
1. Create a Docker image for building it
1. Create a kustomize manifest for deploying it
1. Update Kubeflow's [CD Pipelines](https://github.com/kubeflow/testing/tree/master/apps-cd)  to continually update the kustomize manifest with the latest image

Pipelines (@neuromage ) create a Viewer CRD which can be used for tensboard.
https://github.com/kubeflow/pipelines/tree/master/backend/src/crd/controller/viewer

Currently I believe that is integrated into pipelines and used to auto-visualize tensorboard data reported by pipelines.

@kimwnasptd Thoughts? Any interest in possibly tackling this as part of 0.7?

/cc @karthikv2k 
/cc @neuromage 
/cc @vkoukis

kimwnasptd:
@jlewi yes I would like to contribute to this for 0.7. We could have a first implementation that would provide a very similar UX with the jupyter webapp and then iterate on that.

jlewi:
@kimwnasptd any update on this?

I think there was a mention of the fact that you had a prototype for a UI? Do you have an ETA for when you will have something ready to demo and then a potential PR?

A TensorBoard controller was recently added
https://github.com/kubeflow/kubeflow/tree/master/components/tensorboard-controller

gaocegege:
tensorboard CRD seems to be very simple now. I am wondering if we should use deployment directly. Is there any benefit of introducing a new CRD here?

jlewi:
@kimwnasptd are you interested in potentially being a mentor for any GSOC students interested in working on this?

KaviyaPeriyasamy:
[...]

PR #5164, opened 2020-07-27
Common code between the different python backends

kimwnasptd:
Right now JWA's backend is written in Python Flask. We also have the [Tensorboards web app](https://github.com/kubeflow/kubeflow/issues/3578) whose backend will also be written in Python and Flask. Lastly, the [Volumes web app](https://github.com/kubeflow/kubeflow/issues/4758), that we'd like to contribute, will be also written with the same technologies.

I think it would be a good next step to create a common python codebase/library that all these apps could use, at some extend. This common code will include:
* How we handle exceptions in the backend
* Common routes for serving static files and their cache control policy
* Authorization checks with SubjectAccessReview
* Authentication checks on the Kubeflow headers
* Common helper functions for dates, yaml parsing etc
* health/liveness probes

I have prepared this code while working on the Volumes web app and it's the first step towards pushing the backend code upstream. It would also help our running [GSoC project with Tensorboard](https://summerofcode.withgoogle.com/projects/#4660815581413376) as the backend could use a lot of this common code. The Jupyter web app should also gradually use this code.

I'm also adding @kandrio98 to the loop since this affects his GSoC project as well as to get him more exposed to our review process.

@jlewi if you have a better candidate in mind for reviewing this PR let me know

/cc @jlewi 
/assign @kimwnasptd

### DIFF
--- a/components/crud-web-apps/common/README.md
+++ b/components/crud-web-apps/common/README.md
@@ -1,0 +1,48 @@
+# Common code for CRUD web apps
+
+Since our CRUD web apps like the Jupyter, Tensorboards and Volumes UIs are similarly build with Angular and Python/Flask we should factor the common code in to modules and libraries.
+
+This directory will contain:
+1. A Python package with a base backend. Each one of the mentioned apps are supposed to extend this backend.
+2. An Angular library that will contain the common frontend code that these apps will be sharing
+
+## Backend
+
+The backend will be exposing a base backend which will be taking care of:
+* Serving the Single Page Application
+* Adding liveness/readiness probes
+* Authentication based on the `kubeflow-userid` header
+* Authorization using SubjectAccessReviews
+* Uniform logging
+* Exceptions handling
+* Common helper functions for dates, yaml file parsing etc.
+
+### How to use
+
+In order to use this code during the development process one could use the `-e` [flag](https://pip.pypa.io/en/stable/reference/pip_install/#install-editable) with `pip install`. For example:
+
+```bash
+# I'm currently in /components/crud-web-apps/volumes/backend
+cd ../../common/backend && pip install -e .
+```
+
+This will install all the dependencies of the package and you will now be able to include code from `kubeflow.kubeflow.crud_backend` in you current Python environment.
+
+In order to build a Docker image and use this code you coud build a wheel and then install it:
+```dockerfile
+FROM python:3.7 AS backend-kubeflow-wheel
+
+WORKDIR /src
+COPY ./components/crud-web-apps/common/backend .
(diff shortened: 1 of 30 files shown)


jlewi:
@kimwnasptd Who else will be owning and maintaining this code? They should be the ones reviewing it.

What about @StefanoFioravanzo and what about @lalithvaka

StefanoFioravanzo:
@jlewi I think you are suggesting creating an OWNERS file for the `py/` directory? It would definitely make sense since we have are going to rely heavily on this code for various applications. It's ok for me to be part of that file, @elikatsis as well is involved in the development of this common code. We could start with us + @kimwnasptd being part of the OWNERS

jlewi:
What project or application is this code for? Please open up a separate PR creating an appropriate directory with an OWNERs file. The OWNERs should indicate who will be responsible for reviewing and maintaining this code. 

If this code is primarily for the jupyter web app, why not put it in
https://github.com/kubeflow/kubeflow/blob/master/components/jupyter-web-app/OWNERS

kimwnasptd:
@jlewi this will be common backend code between the jupyter web app, tensorboards web app and the volumes web app.

The jupyter web app will start using the common code more gradually over time, but the tensorboards and volumes web apps will be using this common code out of the box.

jlewi:
Can we structure the code so that all code lives under an appropriate project directory with a suitable OWNERs file that indicates who is responsible for this project.

All projects will need to find WG sponsorship or they will be archived. Ownership of the projects and boundaries should be clearly delineated by directory structure and OWNERs files.

kimwnasptd:
@jlewi since this will be common code for different projects, what structure do you think would fit best here?

For the approvers/reviewers we could add @StefanoFioravanzo @elikatsis and I. Once we decide on where this code should be placed I'll create a new small PR just for adding the OWNERS file in the agreed place[s].

jlewi:
@kimwnasptd I think my comment above applies.

> Can we structure the code so that all code lives under an appropriate project directory with a suitable OWNERs file that indicates who is responsible for this project.

So pick an appropriate directory to serve as the parent directory for all the web apps that you will be developing. Put an OWNERs file there to indicate clear ownership and then organize the code under that.

kimwnasptd:
I updated the code to live under the `crud-web-apps` directory.

@elikatsis @StefanoFioravanzo take a look when when you have the time. I'll be on PTO for the next couple weeks so I won't be able to respond frequently during that time, but I'll try to answer any questions that might come up.

/cc @elikatsis 
/cc @StefanoFioravanzo 
/uncc @jlewi

PR #5180, opened 2020-08-01
tensorboard web-app: Create Tensorboard web-app backend

kandrio98:
This PR contains the code for the Tensorboard web-app backend, which includes routes for `GET`, `POST` and `DELETE` requests.
This is actually part of my [GSoC-2020 project](https://summerofcode.withgoogle.com/projects/#4660815581413376).

The backend is created with Python/Flask, so it also uses the common code from `kubeflow.kubeflow.crud_backend`.
As a result, this PR should be merged after [#5164](https://github.com/kubeflow/kubeflow/pull/5164).

/cc @jlewi @kimwnasptd @elikatsis 
/assign @kandrio98

### DIFF
--- a/components/crud-web-apps/tensorboards/backend/app/utils.py
+++ b/components/crud-web-apps/tensorboards/backend/app/utils.py
@@ -10,11 +10,7 @@ def parse_tensorboard(tensorboard):
         "name": tensorboard["metadata"]["name"],
         "namespace": tensorboard["metadata"]["namespace"],
         "logspath": tensorboard["spec"]["logspath"],
-        "age": {
-            "uptime": helpers.get_uptime(
-                tensorboard["metadata"]["creationTimestamp"]),
-            "timestamp": tensorboard["metadata"]["creationTimestamp"],
-        },
+        "age": helpers.get_age(tensorboard),
     }
 
     return parsed_tensorboard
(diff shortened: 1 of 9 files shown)


components/crud-web-apps/tensorboards/backend/app/utils.py
```diff
@@ -0,0 +1,39 @@
...
+def parse_tensorboard(tensorboard):
+    """
+    Process the Tensorboard object and format it as the UI expects it.
+    """
+
+    parsed_tensorboard = {
+        "name": tensorboard['metadata']['name'],
+        "namespace": tensorboard['metadata']['namespace'],
+        "logspath": tensorboard['spec']['logspath'],
+        "age": {
```
kimwnasptd, line 13: Could you add a `get_age(k8s_object)` in our common [helpers.py](https://github.com/kubeflow/kubeflow/blob/e99b5e18697a15088abea12543bd4e3f180ff984/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/helpers.py) file that would return a dictionary with the `uptime` and `timestamp` values? Since other web apps will want to use this convention, for returning age information to their frontends, it would be a good idea to add this logic into a common place.

For now it could only work with dict objects, like CRs, and later one we could extend it to work with class objects as well, such as PVCs for example.

kandrio98: Thanks for the comment. I totally agree, so I just added a new commit that adds `get_age(k8s_object)` to the common code.

PR #5184, opened 2020-08-04
Create a new directory in components for web apps

### LINKED ISSUE #5183: Create a parent directory for future web apps and their common code
/kind feature

Since we will be having two new web apps, one for [Tensorboards](https://github.com/kubeflow/kubeflow/issues/3578) and one for [Volumes](https://github.com/kubeflow/kubeflow/issues/4758), simply putting their code and the common code they will be sharing under the `/components` directory will make it difficult to navigate and organize.

We should have a parent directory for these. In this directory we will be also placing common code that all of these web apps, and the Jupyter one, will be using

kimwnasptd:
closes #5183 

Since we want to also have some common code between our web apps we should create a parent directory for any future web app we want to develop.

The code for the web apps, common or not, should be organized under this directory.

This PR will also be the first step for merging the code from #5164 

cc @elikatsis @StefanoFioravanzo @kandrio98 

/cc @jlewi 
/assign @kimwnasptd


jlewi:
web apps is overly general. Not all web apps will use this code. Can you scope it down to the web applications that you will own? Which apps are those? I believe they are 3; jupyter UI, tensorboard, and volume controller. Furthermore, I believe based on your previous PR your writing python code for these web apps? 

Per: kubeflow/kubeflow#3578 I believe what you are building is a web app that will manage web apps? So maybe call this something like "web-app-manager"?

kimwnasptd:
> web apps is overly general. Not all web apps will use this code. Can you scope it down to the web applications that you will own?

What about `crud-web-apps`? I can't think of a better name for generalizing these three web apps and their code together. I think `web-app-manager` would not be a good fit, since the folder is not an app per se but a directory that contains the code for other apps.

> Which apps are those? I believe they are 3; jupyter UI, tensorboard, and volume controller.

Yes, it will be for the {Jupyter, Tensorboard, Volumes} UIs. 

> I believe based on your previous PR your writing python code for these web apps?

#5164 has the common backend/Python code, which essentially exposes a base backend that adds authnz/logging/probes/serving to any app that will be extending it. So the three mentioned apps will be extending this base app and will be adding their own routes on top. Next, I'll be creating a PR that will contain common Angular code that the three apps' frontends will be sharing.

So the structure will be looking like this [ we could also discuss the naming for the common code in their respective PRs ]:
* `crud-web-apps`
  * `jupyter`
    * frontend
    * backend
  * `tensorboards`
    * ...
  * `volumes`
    * ...
  * common
    * frontend
      * kubeflow-common-lib
        * package.json
        * ...
    * backend
      * setup.py
      * kubeflow
        * kubeflow
           * crud_backend
   
> Per: #3578 I believe what you are building is a web app that will manage web apps? So maybe call this something like "web-app-manager"?

This folder will not be a web app manager, but rather a reordering in order to include the web apps and their common code together. Per #3578, the web app that will be handling a more generalized Kubeflow CR should also live inside this folder, since it will be utilizing the common code. 

I hope the above clarifies a little bit what I have in mind.

components/web-apps/OWNERS
```diff
@@ -0,0 +1,11 @@
+approvers:
+  - elikatsis
+  - kimwnasptd
+  - StefanoFioravanzo
+reviewers:
+  - avdaredevil
```
jlewi, line 6: Have @avdaredevil and @prodonjs agreed to participate in this? 

Do not duplicate reviewers and approvers. Please refer to
https://www.kubeflow.org/docs/about/contributing/#maintaining-owners-files

kimwnasptd: My bad, I just copied the reviewers from the jupyter web app.
Lets start with an empty list of reviewers for now

Issue #5190, opened 2020-08-05
API calls on centraldashboard become slow with a large number of profiles


zachomedia:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

We started seeing some slow startup time when a user loads the Kubeflow dashboard, and after some tracing tracked down a slow call to /api/workgroup/env-info.

After some investigation, it led us to the kfam service where obtaining role bindings takes 3-6 seconds:

```
GET /kfam/v1/bindings?user=$USER ReadBinding 6.064392368s
```

**What did you expect to happen:**

The call succeed quickly and not slow down with increased profile counts

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

We have approximately 123 profiles in our environment for our users. The speed of the call seems to have slowed down as the profile count increased.

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 1.0.1
- kfctl version: (use `kfctl version`): 1.0.1
- Kubernetes platform: (e.g. `minikube`) AKS
- Kubernetes version: (use `kubectl version`): 1.15.10
- OS (e.g. from `/etc/os-release`): Ubuntu 16.04

sylus:
This is a pretty big deal for us and would expect this would be reproducible by numerous other orgs.

Jeffwan:
@zachomedia @sylus Could you share how many profiles do you have in that cluster? Since it makes call to Kube-apiserver. Any performance issue on master instance may increase latency. Do you have way to know the `6s` trace? I am curious which part contribute more.  active request like `List` to APIServer won't be efficient, I agree we probably need to add some cache and user won't see user immediately but I think this should be acceptable. An option to control cache policy will give some flexibility.

zachomedia:
@Jeffwan We currently have 122 profiles in the cluster. I will see if we can get a way to break down the trace more and get a breakdown on the 6s response time.

Jeffwan:
@zachomedia  I will mark this as part of 1.2 improvement. I think we did tuning on jupternotebook on profile controller but not on UI yet. Might be some tuning to do on some components to improve performance

sylus:
Would this tuning done on the profile controller be applicable to jupyter notebook? If so we could take a look and try to help also. Keen to move this forward.

derekwyatt:
[...]

PR #5202, opened 2020-08-11
Use informer / indexer to cache rolebindings

thetko85:
To solve https://github.com/kubeflow/kubeflow/issues/5190

Found most of the issue to occur when listing the rolebindings when there are many namespaces as it does a loop. Query the shared index to list and get the rolebindings.

* Note: first commit and if there are procedures I am missing, please let me know.

### DIFF
--- a/components/access-management/kfam/api_default.go
+++ b/components/access-management/kfam/api_default.go
@@ -12,13 +12,14 @@ package kfam
 
 import (
 	"encoding/json"
-	log "github.com/sirupsen/logrus"
 	istioRegister "github.com/kubeflow/kubeflow/components/access-management/pkg/apis/istiorbac/v1alpha1"
 	profileRegister "github.com/kubeflow/kubeflow/components/access-management/pkg/apis/kubeflow/v1beta1"
 	profilev1beta1 "github.com/kubeflow/kubeflow/components/profile-controller/api/v1beta1"
+	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/informers"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -28,6 +29,7 @@ import (
 	"path"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 	"strconv"
+	"time"
 )
 
 type KfamV1Alpha1Interface interface {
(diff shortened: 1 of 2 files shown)


thetko85:
@googlebot I fixed it.

yanniszark:
@thetko85 thanks a lot for the pr.
Because the client-go and controller-runtime packages already solve this problem and are widely used, I would strongly suggest reusing them.
More specifically, client-go informers or controller-runtimer manager can be used for getting a caching client without needing to reimplement the caching logic.
Can you take a look at those? I'll try to come back with some examples but feel free to explore.

thetko85:
@yanniszark thanks for pointing me in the right direction. makes a lot more sense now that you mentioned it. I have refactored the code to utilise the informer and the indexer to get the rolebindings.

let me know if there is something that I am missing.

Cheers,

yanniszark:
@thetko85 thanks for the prompt response and nice work on using an Informer!
Now that you use an informer, the `RoleBindingRepo` abstraction is not really necessary.
You can use `factory.Rbac().V1().RoleBindings().Lister()` which has Get/List methods that use the informer's cache.

thetko85:
@yanniszark Thanks for the pointers. I have refactored the code to utilise the lister api. One question I do have is the stopper channel for the informers. Are there any implications for not explicitly stopping the channel in cases of panics or errors to the http server?

yanniszark:
@thetko85 sorry for the late response, this is a PTO period.

> Are there any implications for not explicitly stopping the channel in cases of panics or errors to the http server?

That's a good question. I expect that the informer uses the stop channel so that it is informed of when the program needs to shut down, so that it can stop and cleanup the watches on the API Server. In a previous controller I wrote, I see that I created a channel that captures certain shutdown signals:
https://github.com/rook/rook/blob/0bcf040bd7653d8ce6b13a276ee24125f30ceb4c/cmd/rook/cassandra/operator.go#L80-L85

I would suggest doing the same here.

yanniszark:
On another sidenote, in case you want to improve the code even more, we can use the controller-runtime package's caching client. This client encapsulates caching for all types, so you don't need to make a lister/informer dance for every type that's needed:
https://godoc.org/sigs.k8s.io/controller-runtime/pkg/manager#DefaultNewClient

yanniszark:
ping @thetko85, any updates on this?
The only changes needed is to add a stop channel to the informer that catches signals, as detailed here:
https://github.com/kubeflow/kubeflow/pull/5202#issuecomment-676575622

DavidSpek:
@yanniszark Is this something we should try and have included in release 1.3?

yanniszark:
@DavidSpek it seems the original author won't continue this effort. Do you want to open a new PR for this? I don't think it's a blocker, but it's good to continue the effort :)

DavidSpek:
@yanniszark I will open a PR for this then. I'm not sure how best to test image though (notebook controller is pretty easy to see if it is broken).

DavidSpek:
@yanniszark I'm closing this PR as I have opened https://github.com/kubeflow/kubeflow/pull/5673 instead.
/close

PR #5218, opened 2020-08-15
Tensorboard controller: Add scheduling functionality for Tensorboard servers that use RWO PVCs as log storages

### LINKED ISSUE #5039: Extend Tensorboard Controller to Support PVCs
/kind feature

Currently, the Tensorboard controller does not fully support arbitrary PVCs being mounted to pods running the Tensorboard server. In fact, it can only support PVCs named [tb-volume](https://github.com/kubeflow/kubeflow/blob/4bf2d4131a5aebe0f1b53d36dd94031f88426c0b/components/tensorboard-controller/controllers/tensorboard_controller.go#L156-L164).

I would like to extend the Tensorboard controller so that it fully supports any PVC created by the user in the namespace of the Tensorboard object applied. This is actually a part of my [Google Summer of Code project](https://summerofcode.withgoogle.com/projects/#4660815581413376) for building a Tensorboard UI that also supports PVCs.

My proposition would be to extend the Tensorboard controller so that it obtains the name of the PVC by the `spec.logspath` field of the Tensorboard CR applied by the user. In detail, if a user wants to mount a PVC to a pod running the Tensorboard server, then the value of `spec.logspath` should follow the format: `pvc://{pvc-name}/{local-path}` (similar to what `kfserving` uses in its `InferenceService` CRDs). This way, I will be able to extend the code of the controller so that it creates a deployment that mounts the desired PVC to the pod running the Tensorboard server. 
The extended controller will also be backwards compatible.

cc @jlewi @kimwnasptd @elikatsis

kandrio98:
closes #5039

The added scheduling feature is disabled by default. 

It is enabled only if the `RWO_PVC_SCHEDULING` env var is set to `true` when running the Tensorboard controller.

/cc @jlewi @kimwnasptd @elikatsis
/assign @kandrio98

### DIFF
--- a/components/tensorboard-controller/controllers/tensorboard_controller.go
+++ b/components/tensorboard-controller/controllers/tensorboard_controller.go
@@ -68,7 +68,10 @@ func (r *TensorboardReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
 	}
 
 	// Reconcile k8s deployment.
-	deployment := generateDeployment(instance, logger)
+	deployment, err := generateDeployment(instance, logger, r)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
 	if err := ctrl.SetControllerReference(instance, deployment, r.Scheme); err != nil {
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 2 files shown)


components/tensorboard-controller/controllers/tensorboard_controller.go
```diff
@@ -126,27 +130,69 @@ func (r *TensorboardReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
-					ClaimName: "tb-volume",
+					ClaimName: pvcname,
 				},
 			},
 		})
+
+		if err, sch := rwoPVCScheduling(); err != nil {
+			return nil, err
+		} else if err == nil && sch {
```
kimwnasptd, line 171: nit: I think you don't need the `err == nil` check here since this holds from the previous if check

components/tensorboard-controller/controllers/tensorboard_controller.go
```diff
@@ -281,3 +328,102 @@ func isCloudPath(path string) bool {
...
+	pod = findRunningPod(pods)
+
+	//If there is no running pod that uses the PVC, then: nodename == "".
+	//Else, nodename contains the name of the node that the pod is running on.
+	nodename = extractNodeName(pod)
+
+	//In this case, there is a running pod that uses the PVC, therefore we create
+	//a nodeAffinity field so that the Tensorboard server will be scheduled (if possible)
+	//on the same node as the running pod.
+	if nodename != "" {
```
kimwnasptd, line 393: nit: You could have an `if nodename == "" {` here and return nil if that holds.

Then you could move the code below one tab to the left, outside of an if clause, to make it a little bit more simple to read

kimwnasptd (commented):
Just some nit comments and lgtm overall.

PR #5226, opened 2020-08-19
Revert "Fix #4842 Pod resources limit (#5118)"

### LINKED ISSUE #4842: Pod resources limit

thesuperzapper:
This reverts commit 4b06248b

Setting a limit is an extremely bad idea, as it can result in an unexpected shutdown of notebook instance (and losing anything currently in memory on the pod).

There is also an important distinction to make about what a "request" actually means. Let's not forget that in the end we are just spawning Pods on Docker, which has access to all of the host's resources, meaning the code running in the pod has no idea that it will be suddenly killed if it uses more than some arbitrary number.

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/common/utils.py
@@ -390,7 +390,6 @@ def set_notebook_cpu(notebook, body, defaults):
         logger.info("Using default CPU: " + cpu)
 
     container["resources"]["requests"]["cpu"] = cpu
-    container["resources"]["limits"]["cpu"] = cpu
 
 
 def set_notebook_memory(notebook, body, defaults):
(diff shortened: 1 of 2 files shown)


thesuperzapper:
I have a pending PR which will put this feature behind a config flag, so there would be no need to revert it.

PR #5259, opened 2020-08-28
Tensorboard web-app: Add functionality to inform TWA frontend about the status of Tensorboard servers

### LINKED ISSUE #5166: Tensorboard CR doesn't contain information about the Tensorboard Server being ready or not
/kind feature

Currently, the `status` field of the `Tensorboard CR` does not contain any information about the Tensorboard Server being ready or not ([see CRD](https://github.com/kubeflow/kubeflow/blob/0360d956a02970c0aae27bce591d34d903ce2d88/components/tensorboard-controller/config/crd/bases/tensorboard.kubeflow.org_tensorboards.yaml#L42-L63) for Tensorboard).

I believe that being able to know if a `Tensorboard Server` is ready is a necessary UX functionality for the `Tensorboard web-app` (which I'm currently building for my [GSoC project](https://summerofcode.withgoogle.com/projects/#4660815581413376)). The frontend of the web-app should inform users about the Tensorboard objects that are ready or, more specifically, about the **Tensorboard Servers** that are ready to connect. This can be done by listing the Tensorboard CRs in a specific namespace, and getting the necessary information from their `status` fields. But, in the present, no such information is provided by the Tensorboard CRs.

**The solution I propose:** The Tensorboard Controller should check the `readyReplicas` field of the Tensorboard `deployment.status` and keep it in sync with the `readyReplicas` field of the `tensorboard.status`.

For the above solution to be implemented, we will need to:

- Extend the `Tensorboard CRD` so that Tensorboard CRs will contain a `readyReplicas` field in theιr status.
- Extend the Tensorboard Controller so that it checks the `readyReplicas` field of the Tensorboard `deployment.status` and copies it's value to the `readyReplicas` field of the Tensorboard CR.

This way, the frontend of the web-app will be able to check the `tensorboard.status` and it will inform the user that the corresponding Tensorboard Server is ready to connect.

It is a really basic fuctionality, but since the current Tensorboard CR doesn't provide any information about the status of the Tensorboard Server, it would be a necessary addition for a descent, more intuitive UX.

cc @jlewi @elikatsis @kimwnasptd

kandrio98:
closes #5166 

This is part of my [GSoC-2020 project](https://summerofcode.withgoogle.com/projects/#4660815581413376). 

It also contains a small code that adds a PVC GET route of the Tensorboard web-app backend. 

/cc @kimwnasptd @elikatsis
/assign @kandrio98

### DIFF
--- a/components/crud-web-apps/tensorboards/backend/app/utils.py
+++ b/components/crud-web-apps/tensorboards/backend/app/utils.py
@@ -7,9 +7,11 @@ def parse_tensorboard(tensorboard):
     """
     
     if tensorboard.get("status", {}).get("readyReplicas", 0) == 1: 
-        status = "ready"
+        phase = "ready"
+        message = "The Tensorboard server is ready to connect"
     else:
-        status = "unavailable"
+        phase = "unavailable"
+        message = "The Tensorboard server is currently unavailble"
 
     parsed_tensorboard = {
         "name": tensorboard["metadata"]["name"],
@@ -17,7 +19,8 @@ def parse_tensorboard(tensorboard):
         "logspath": tensorboard["spec"]["logspath"],
         "age": helpers.get_age(tensorboard),
         "status": {
-            "phase": status, 
+            "phase": phase,
+            "message": message, 
         }
     }
 
--- a/components/crud-web-apps/tensorboards/backend/app/utils.py
+++ b/components/crud-web-apps/tensorboards/backend/app/utils.py
@@ -5,15 +5,30 @@ def parse_tensorboard(tensorboard):
     """
     Process the Tensorboard object and format it as the UI expects it.
     """
+    
+    if tensorboard.get("status", {}).get("readyReplicas", 0) == 1: 
+        status = "ready"
+    else:
+        status = "unavailable"
 
     parsed_tensorboard = {
         "name": tensorboard["metadata"]["name"],
         "namespace": tensorboard["metadata"]["namespace"],
         "logspath": tensorboard["spec"]["logspath"],
         "age": helpers.get_age(tensorboard),
+        "status": {
+            "phase": status, 
+        }
     }
 
     return parsed_tensorboard
+
+def getPVCName(pvc):
+    """
+    Return name of given PVC.
+    """
+
+    return pvc.metadata.name
 
 
 def get_tensorboard_dict(namespace, body):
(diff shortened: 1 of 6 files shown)


components/crud-web-apps/tensorboards/backend/app/utils.py
```diff
@@ -5,16 +5,34 @@ def parse_tensorboard(tensorboard):
     """
     Process the Tensorboard object and format it as the UI expects it.
     """
+    
+    if tensorboard.get("status", {}).get("readyReplicas", 0) == 1: 
+        phase = "ready"
```
kimwnasptd, line 10: Could you use the [class attributes](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/status.py#L1) instead of string values here?
It should be `STATUS_PHASE.READY`

components/crud-web-apps/tensorboards/backend/app/utils.py
```diff
@@ -5,16 +5,34 @@ def parse_tensorboard(tensorboard):
...
+    else:
+        phase = "unavailable"
+        message = "The Tensorboard server is currently unavailble"
 
     parsed_tensorboard = {
         "name": tensorboard["metadata"]["name"],
         "namespace": tensorboard["metadata"]["namespace"],
         "logspath": tensorboard["spec"]["logspath"],
         "age": helpers.get_age(tensorboard),
+        "status": {
```
kimwnasptd, line 21: Could you use the helper function from the [status.py](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/status.py#L16) instead?

components/crud-web-apps/tensorboards/backend/app/utils.py
```diff
@@ -5,16 +5,34 @@ def parse_tensorboard(tensorboard):
     """
     Process the Tensorboard object and format it as the UI expects it.
     """
+    
+    if tensorboard.get("status", {}).get("readyReplicas", 0) == 1: 
+        phase = "ready"
+        message = "The Tensorboard server is ready to connect"
+    else:
+        phase = "unavailable"
```
kimwnasptd, line 13: same here as above

kimwnasptd (commented):
some small comments, lgtm otherwise

Issue #5287, opened 2020-09-08
Use nonroot user as default user in notebook controller container


MartinForReal:
/kind feature

**Why you need this feature:**
Need to follow container security best practise: use nonroot user as default user

**Describe the solution you'd like:**
change the default user account to nonroot user

**Anything else you would like to add:**
none

Issue #5250, opened 2020-08-28
Failed to create statefulset on kubernetes 1.18+ when new notebook resource is defined because finalizer access is not defined in rbac configuration


MartinForReal:
/kind bug

**What steps did you take and what happened:**
Apply a notebook resource in kubernetes 1.18 cluster
and controller reported that 

```
 cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: 
```

**What did you expect to happen:**
a statefulset is created without error 

**Anything else you would like to add:**

**Environment:**

- Kubeflow version: 1.0.1
- kfctl version: (use `kfctl version`): none
- Kubernetes platform: (e.g. `minikube`): 1.18.6
- Kubernetes version: (use `kubectl version`):1.18.6
- OS (e.g. from `/etc/os-release`):centos 7.8

MartinForReal:
close this issue in favor of #5288 #5286 #5285

PR #5288, opened 2020-09-08
Use nonroot user in notebook controller

MartinForReal:
Signed-off-by: MartinForReal <fanshangxiang@gmail.com>
Fixed: #5287

### DIFF
--- a/components/notebook-controller/Dockerfile
+++ b/components/notebook-controller/Dockerfile
@@ -6,11 +6,10 @@
 #
 # This is necessary because the Jupyter controller now depends on
 # components/common
-ARG GOLANG_VERSION=1.12.5
+ARG GOLANG_VERSION=1.13
 FROM golang:${GOLANG_VERSION} as builder
 
 WORKDIR /workspace
-
 # Copy the Go Modules manifests
 COPY notebook-controller /workspace/notebook-controller
 COPY common /workspace/common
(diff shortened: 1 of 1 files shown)


MartinForReal:
related #5250

MartinForReal:
ping ? @StefanoFioravanzo

MartinForReal:
Could you please help review this pr? Thanks! @kimwnasptd

components/notebook-controller/Dockerfile
```diff
@@ -6,11 +6,10 @@
 #
 # This is necessary because the Jupyter controller now depends on
 # components/common
-ARG GOLANG_VERSION=1.12.5
+ARG GOLANG_VERSION=1.13
```
naveensrinivasan, line 9: Why change the `go` version as part of this PR. IMO that should be a separate PR

DavidSpek:
The version update for GO in this PR is also being done by https://github.com/kubeflow/kubeflow/pull/5394

DavidSpek:
I just built the image without but without the update to the Go version (as this could cause conflicts with the PR mentioned above), and deployed it on my Kubeflow 1.2 installation and there don't seem to be any problems. Creating and deleting notebooks works as expected and no errors in the controller pod logs.

yanniszark:
The Go version was updated as part of https://github.com/kubeflow/kubeflow/pull/5394.
@MartinForReal could you update your PR so that it only changes the image?

DavidSpek:
For those interested, the difference between `base` and `static` is that `static` does not contain glibc, libssl, openssl. 
The `static` image does not have a `debug` flag, so no shell is present. If a shell would be needed the alternative is `base:debug-nonroot`. 
To check what PID the process runs as I built the image using `base:debug-nonroot` and this is the output of `ps aux`:
```
PID   USER     TIME  COMMAND
    1 nonroot   0:00 /manager
   48 nonroot   0:00 sh
   53 nonroot   0:00 ps aux
```

Other info from the BUILD file for the image that might be interesting to indicate what is different from the root image:
```
passwd_entry(       
name = "nonroot_user",       
gid = 65532,       
home = "/home/nonroot",       
info = "nonroot",       
shell = "/sbin/nologin",       
uid = 65532,       
username = "nonroot",     )
```

DavidSpek:
@MartinForReal could you please remove the Go version update from this PR so that it can be merged?

DavidSpek:
Closing in favor of https://github.com/kubeflow/kubeflow/pull/5428

/close

Issue #5282, opened 2020-09-07
Notebook controller is only checking the status of one (the first) container


gilbeckers:
/kind bug

**What steps did you take and what happened:**
When you add a new notebook server, the status of the newly added notebook becomes ready to soon. Namely, if you click on connect, a "no healthy upstream" message shows the first seconds. I tried to fix this by adding a readiness probe to the notebook container (via a custom admission controller). However, the problem stayed the same: the green check mark appears under status too soon.

**What did you expect to happen:**
Correct behavior of the status icon of the notebook server in the Kubeflow dashboard.

**Anything else you would like to add:**
I think I know what the problem is and why the status is wrongly reported in the Kubeflow dashboard: The notebook controller is only checking the status of the first container https://github.com/kubeflow/kubeflow/blob/c4717d84b6d927b7cf2305511002094e9d19f5a4/components/notebook-controller/controllers/notebook_controller.go#L210 

As a result, it's only taking the status of the istio-proxy container in consideration, and the status of the notebook container is ignored. I think this is a little bug, as the controller should definitely check the status of all containers, and not only the first one.

I'm new to this community, but if it's ok I would like to fix this and add a new pull request.

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 1.0.2
- kfctl version: (use `kfctl version`): 1.0.2
- Kubernetes platform: (e.g. `minikube`): minikube
- Kubernetes version: (use `kubectl version`): client v1.19.0 , server v1.18.3
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

gilbeckers:
**Update:**

I managed to fix the wrongly reported status of a new created notebook server in the kubeflow dashboard. 
Now, the status of nb server becomes ready only after the two containers (isto-proxy and notebook) are both Running and Ready. 

In order to fix this, two changes were needed: 
- The netbook controller was only checking the Status of the first container, resulting in only the istio container being taken into account (as stated in the first comment above). I changed this so that now all containers in the pod are considered.
https://github.com/gilbeckers/kubeflow/blob/db06cc850e9ef8fa602ba825444099cca2ee0f95/components/notebook-controller/controllers/notebook_controller.go#L217

[...]

PR #5292, opened 2020-09-09
Add liveness and readness probe for jupyter notebook

MartinForReal:
Add liveness and readness probe for jupyter notebook
Fixed: #5282
Signed-off-by: MartinForReal <fanshangxiang@gmail.com>


MartinForReal:
@googlebot I fixed it.

thesuperzapper:
@MartinForReal Is this disable-able/configurable for non-Jupyter docker images?

Or even just JupyterHub ones?

MartinForReal:
@thesuperzapper  "/notebook/" + instance.Namespace + "/" + instance.Name + "/api/status" is the default URI generated by notebook controller.  No matter what the underlying image is, the container should launch a Jupyter notebook instance.  /api/status is a standard API provided by Jupiter notebook. So I think it should work well with all kinds of Jupyter instance.  The notebook culling feature in the notebook controller is a good example. 

I can add one extra env flag for this, just like what  ENABLE_CULLING did.

How do you think?

thesuperzapper:
@MartinForReal, I am talking about when the image launched by the notebook controller is NOT a Jupyter image, for example: RStudio or Theia. 

It would be good if we can at least prepare for that, as this change would cause those images to perpetually restart.

MartinForReal:
Why don't we create another controller for RStudio or Theia? @thesuperzapper

thesuperzapper:
@MartinForReal, I am fairly sure we can just generalise the existing notebook-controller to support any docker image which exposes a web socket.

Currently you have to hack it a bit for non-jupyter, but it does work as long as you:
1. use port `8888`
1. embed some HTTP path rewriting for `./notebook/{namespace}/{notebook-name}/`
1. ensure the HTTP header has: `Access-Control-Allow-Origin: *`

We can make this way easier if we:
1. make the target port configurable
1. manage the HTTP path rewriting OUTSIDE the docker image:
   * like with the `istio gateway` OR a basic sidecar which just forwards the traffic to the docker image
1. inject the `Access-Control-Allow-Origin: *` header
1. remove the hardcoding to use `/home/jovyan` if no mount point is specified (probably just fail instead)
1. don't implement things like this PR, without making them general enough to handle all types of docker image:
    * not all docker images expose something at `/api/status`

MartinForReal:
Well, it is interesting but I'm afraid I have different ideas. 

I think there are a lot of solutions available which help us create three objects ( one statefulset, one service, and one virtualservice) such as helm, oam and etc. And they are generic enough so that we can apply them without any modification to our apps. To be more specific: what are the pros and cons when this solution is compared with helm operator + customized helm charts?

By implementing the controller, we can customize the resource lifecycle to support Jupiter operations,  especially day2 operations.

thesuperzapper:
@MartinForReal when I say "support any docker image which exposes a web socket" I am actually meaning any "web-based IDE"

I think it would be crazy for us not to consolidate this into a single controller, as the code duplication would be crazy.

For example, I believe these are the IDE's we need to support:
* [Jupyter](https://jupyter.readthedocs.io/en/latest/)
* [JupyterLab](https://jupyterlab.readthedocs.io/en/stable/)
* [RStudio](https://hub.docker.com/r/rocker/rstudio)
* [Theia](https://theia-ide.org/)

MartinForReal:
But I don't think we can find a general solution that provides monitoring, error handling, authentication without specific customization. And we can extract common libraries from these implementations.

I don't think code duplication issue does exist.

thesuperzapper:
@MartinForReal why not put a `notebookType` field in the Notebook CRD, with values like:
-  `jupyter`
- `rstudio`
- `theia`
- `generic`

The setting of `notebookType` would control all the IDE specific features, like idle detection and liveness checks.

Why do you need a new controller, and therefore CRD for each notebook type?

gilbeckers:
I'm also trying to fix this problem, see #5282

Instead of adding the readiness probe to the notebook container via the notebook controller, we were thinking to do this using a new custom admission controller. The idea is that at the creation of a notebook pod, the admission control would jump in and mutate the pod as desired.  

The admission controller is configured via a MutatingWebhookConfiguration to act on the creation of Pods. Then, at create, the admission controller filters the right pod and only allows notebook pods to pass. The next step is to find the right container, as there are two (isto-proxy and the container running the nb image): this is done by looking at the environment variables. The nb container has a var named "NB_PREFIX", The final step is to add the readiness probe to the container.

MartinForReal:
@thesuperzapper I don't see any improvement when it is compared with the new controller. because it makes validation logic and control logic much more complicated and we also need to add extra logic to query different kinds of instance. Cluster-API is a good example of this kind of design. But you can find that It is difficult to validate cr and implement providers. And there is no point in making it general because we can not share any code except object creation. 

> The setting of notebookType would control all the IDE specific features, like idle detection and liveness checks.

thesuperzapper:
@MartinForReal but if you think about it from a user perspective, having a new resource type for each new type of IDE is cumbersome, as
all the resource configs are the same, like volumes, name, etc. (Only the UI itself is different)

 Having multiple Resource types also causes more pain in the frontend UI design, and just pushes all the complexity into the JavaScript/Python app.

Under your design (multiple controllers) how would you propopose supporting a "generic" IDE type which could be used by Kubeflow admins for whatever internal tool they might have?

Finally, I believe there is some code which is shared, especially complicated logic like:
* "is there currently someone using a pod"
* "when was the last time someone accessed the UI of the pod"

MartinForReal:
@thesuperzapper well, I can hardly agree with you because daemonset, statefulset, and deployment all have a field which is the type of corev1.podSpec But it doesn't mean they should be managed by the same controller. and I think it is a common practise to apply Backend for FrontEnd pattern here. And we have graphql and other tools and technology which can simplify works in front end. 

And I highly doubled that requirement is realistic because the user does care what kind of ide they are using.  And kubeflow admin always manage objects using labeling strategies.

thesuperzapper:
@MartinForReal regardless of how we manage the Notebook, please make this PR not break non-jupyter docker images.

thesuperzapper:
@MartinForReal, after thinking about it more, you are partially right but I think I have an idea which will work for both of us.

We can create an annotation for the Notebook resource called `kubeflow-resource-type`:
* initially `kubeflow-resource-type` should support two values: `jupyter` and `generic`
* when it is `jupyter` 
   * we should enable features like injecting liveness probes (for example, like this PR)
* when it is `generic`:
   * we disable injection of jupyter specific features 
* NOTE: this field has an implied default value of `jupyter` when missing for backwards compatibility

Note there is already precedent for annotations like this, see [kubeflow-resource-disabled](https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/pkg/culler/culler.go#L37).

---

If you do this, I can implement a user friendly way for Kubeflow administrators to provide non-jupyter images in the `jupyter-web-app` spawner UI.

thesuperzapper:
@kimwnasptd you might want to chip in

kimwnasptd:
First off, supporting more types of Notebooks is something we want to put effort in for Kubeflow 1.2 and forward. 


FYI we are [moving towards](https://github.com/kubeflow/kubeflow/issues/3578#issuecomment-655724933) abstracting the way we deploy our apps in Kubeflow. In the near future we'll be discussing on what technical details to abstract in order to deploy a stateful application on Kubeflow, such as a Notebook, an IDE, Tensorboard etc, via a more generic CutomResource.

But until then, for this PR I also aggre with @thesuperzapper to not introduce extra functionality that only works for Jupyter. What I propose is:
* lets use a `notebook-type: {jupyter,theia,rstudio}` annotation in a Notebook CR to express the type of the Notebook. In case it is missing it will default to `jupyter` to maintain backward compatibility
* if it's `jupyter` then we'll use the path @MartinForReal has already implemented in this PR
* if it's not `jupyter` then the controller will log that it doesn't support setting readiness/liveness probes for the provided type of Notebook.

This way we move forward with @MartinForReal's implementation, but we also allow for other people to add support for other types of Notebooks in the future.

@MartinForReal could you update the code to align with the proposed plan?

yanniszark:
Hi all, this is a very interesting issue!
I have some questions regarding the liveness probe. Namely:
* Will the liveness probe interfere with culling? AFAIK, probing is performed by the kubelet and traffic passes through the Istio sidecar container, so it probably counts as traffic. Maybe an exec liveness probe would be better here, as it would go through localhost?

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -323,6 +322,36 @@ func generateStatefulSet(instance *v1beta1.Notebook) *appsv1.StatefulSet {
 			},
 		}
 	}
+	if len(container.Ports) == 1 {
```
gilbeckers, line 325: Here you assume there is only one container running in the pod, while in fact there are two; istio-proxy and notebook container. As result, the liveness- and readiness probe will only be added to the istio-proxy container and the notebook container will stay untouched.

MartinForReal: istio-proxy is added by istio sidecar injector.  And only when pod is created, the istio sidecar injector can add new containers. 
The controller watches the change of notebook resource and creates the correlated statefulset. The controller can only find one container when it generate statefulset definitions. FYI

thesuperzapper: We could also target the port based off the name `notebook-port`

gilbeckers: Oops, sorry I did't know that :blush:. Thanks for the information

vinayan3:
@yanniszark Is there still an appetite for landing this? I just encoutered a case where the notebook's server crashed on startup and never came alive. It'd be great to have a liveness probe.

PR #5303, opened 2020-09-12
Deprecate HTTP header X-Auth-Token

### LINKED ISSUE #5302: Deprecate header "X-Auth-Token" by using "Authorization"
/kind feature

The problem was revealed from this comment https://github.com/kubeflow/kubeflow/issues/4912#issuecomment-665387957

The header `X-Auth-Token` is designed to authenticate request that doesn't contain secure cookie. e.g., API requests from notebook. However, based on https://github.com/arrikto/oidc-authservice/pull/25 `X-Auth-Token` has been replaced by `Authorization`. which means, if it upgrades oidc-authservice for multi-tenant kubeflow, some kubeflow UX will be broken. 

in existing code where it sends the header `X-Auth-Token`, send its value along with the header `Authorization`. 
Then trying merge https://github.com/kubeflow/manifests/pull/1431

This will also encourage the use of standard HTTP headers to be proxy/firewall friendly.

shawnzhu:
Fixes #5302

### DIFF
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/rok/rok.service.ts
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/rok/rok.service.ts
@@ -67,7 +67,9 @@ export class RokService extends BackendService {
     return this.http
       .head<any>(url, {
         headers: new HttpHeaders({
+          // X-Auth-Token header is deprecated by Authorization header
           'X-Auth-Token': this.csrfToken,
+          'Authorization': this.csrfToken,
         }),
         observe: 'response',
       })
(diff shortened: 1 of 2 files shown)


yanniszark:
/cc @kimwnasptd 
I think the `X-Auth-Token` header here is used for a different purpose (CSRF), so we may want to keep it.

kimwnasptd:
@yanniszark correct, this is only used in the Rok flavor of the UI and it is used for setting up communication with Rok.

This does not affect the default Jupyter web app that gets deployed on Kubeflow

shawnzhu:
I've summarized what I learned here: https://github.com/kubeflow/kubeflow/issues/5302#issuecomment-692124126

> I think the `X-Auth-Token` header here is used for a different purpose (CSRF), so we may want to keep it.

> correct, this is only used in the Rok flavor of the UI and it is used for setting up communication with Rok.

like to learn more about this since I could not find anything else about the usage of `X-Auth-Token` where the only place I saw is it can be accepted by OIDC-authservice for authentication purpose.

Issue #5308, opened 2020-09-15
Can't import notebook controller API


jrykr:
/kind bug

**What steps did you take and what happened:**

1. Create a go.mod that requires kubeflow.
2. Import a version from "github.com/kubeflow/kubeflow/components/notebook-controller/api/..."
3. Build fails trying to resolve notebook-controller's dependencies as follows:

```bash
$ go build
go: finding github.com/kubeflow/kubeflow/components/notebook-controller latest
go: github.com/kubeflow/kubeflow/components/notebook-controller@v0.0.0-20200908101143-7f5e242f4671 requires
        github.com/kubeflow/kubeflow/components/common@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000
```

```
// kubeflow/components/notebook-controller/go.mod
// ...
require (
  github.com/kubeflow/kubeflow/components/common v0.0.0-00010101000000-000000000000
  // ...
)

replace github.com/kubeflow/kubeflow/components/common => ../common
```

**What did you expect to happen:**

I would like to be able to use the notebook controller API.

**Anything else you would like to add:**

- It looks like the reason for the relative require was because of failed builds caused by factoring out `common` #4582.
- I'll put in a PR for what worked for me. I just added a valid pseudo-version, which made the relative replace unnecessary. 

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): v1beta1
- kfctl version: (use `kfctl version`): n/a
- Kubernetes platform: (e.g. `minikube`): n/a
- Kubernetes version: (use `kubectl version`): n/a
- OS (e.g. from `/etc/os-release`): Mac OS X 10.15.6

jrykr:
Made PR #5309

yanniszark:
PR is merged so this issue should be resolved.

PR #5309, opened 2020-09-15
Use valid commit for kubeflow/components/common module.

jrykr:
Makes the notebook controller API importable for 3rd party applications. 

See #5308

### DIFF
--- a/components/notebook-controller/go.mod
+++ b/components/notebook-controller/go.mod
@@ -4,7 +4,7 @@ go 1.12
 
 require (
 	github.com/go-logr/logr v0.1.0
-	github.com/kubeflow/kubeflow/components/common v0.0.0-00010101000000-000000000000
+	github.com/kubeflow/kubeflow/components/common v0.0.0-20200908101143-7f5e242f4671
 	github.com/prometheus/client_golang v0.9.0
 	k8s.io/api v0.0.0-20190409021203-6e4e0e4f393b
 	k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d
@@ -13,4 +13,7 @@ require (
 	sigs.k8s.io/controller-tools v0.2.0 // indirect
 )
 
+// Ensure we build the notebook-controller with the latest `common`
+// module. However, because this module's `replace` will be ignored by
+// other modules, we still specify a commit in the `require` directive.
 replace github.com/kubeflow/kubeflow/components/common => ../common


swiftdiaries:
Could you please open a PR to add yourself to the Kubeflow org?

jrykr:
PR opened for Kubeflow org addition.

MartinForReal:
From my point of view, the notebook controller should depend on the code in the same commit. If we remove the replace statement here, we should maintain a release tag for the common component which is something we won't do in the near future.

jrykr:
@MartinForReal  The problem with the current situation is that there are lots of reasons for a team to want to manage notebook servers outside of kubectl or jupyter-web-app, but builds that depend on the notebook controller API will fail. To me, this is worth solving because we do want to encourage a simple workflow for development on top of Kubeflow. 

What about leaving the replace statement unchanged while adding a valid pseudo-version? I think this would:

1. Enable third-party builds, i.e., builds where notebook-controller is not the main module.
2. Avoid creating the obligation to maintain a release tag in components/common.
3. Continue using the relative path when notebook-controller is the main module.

Thoughts?

yanniszark:
> What about leaving the replace statement unchanged while adding a valid pseudo-version?
If this works as described, it would be ideal.
Have you tested it in a sample repo?

jrykr:
> have you tested it in a sample repo?

@yanniszark  Yes, I have confirmed this. 

Amended accordingly.

yanniszark:
@jrykr thanks! Is it possible to also add a comment with the explanation of what we did as a comment in `go.mod` (and in the commit)?
Something over the replace statement, like:
```
We use the latest `kubeflow/common` to build the notebook-controller, but also import a
certain commit to support importing the Notebook API. According to the go.mod docs [1]:

> exclude and replace directives only operate on the current (“main”) module. exclude and
> replace directives in modules other than the main module are ignored when building the
> main module. The replace and exclude statements, therefore, allow the main module
> complete control over its own build, without also being subject to complete control by
> dependencies.

[1]: https://github.com/golang/go/wiki/Modules#gomod
```

jrykr:
@yanniszark Makes sense. Amended with a shorter comment in `go.mod`, expanded in the commit message.

jrykr:
@yanniszark Bumping this.

yanniszark:
For some reason, it didn't pick up the approve...

/approve

PR #5314, opened 2020-09-18
Correct ContainerStatus of Notebook CR

### LINKED ISSUE #5282: Notebook controller is only checking the status of one (the first) container

gilbeckers:
The Notebook Controller doesn't set the `ContainerState` of the CR correctly. In some cases the first container is the istio-sidecar which results in an incorrect state being shown to the Notebook CR. This is fixed now by showing the `ContainerState` of Notebook container to the Notebook CR `ContainerState`. 

The Notebook container is selected by matching with [the name of the notebook](https://github.com/kubeflow/kubeflow/blob/7f5e242f4671f453eaee72e4319a5f7f190f7725/components/jupyter-web-app/backend/kubeflow_jupyter/common/yaml/notebook.yaml#L14). A warning is logged in case no container is found.

This PR fixes #5282.

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -213,30 +213,35 @@ func (r *NotebookReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		if len(pod.Status.ContainerStatuses) > 0 {
 			notebookContainerFound := false
 			for i := range pod.Status.ContainerStatuses {
-				if pod.Status.ContainerStatuses[i].Name == instance.Name &&
-					pod.Status.ContainerStatuses[i].State != instance.Status.ContainerState {
-					log.Info("Updating Notebook CR state: ", "namespace", instance.Namespace, "name", instance.Name)
-					cs := pod.Status.ContainerStatuses[i].State
-					instance.Status.ContainerState = cs
-					oldConditions := instance.Status.Conditions
-					newCondition := getNextCondition(cs)
-					// Append new condition
-					if len(oldConditions) == 0 || oldConditions[0].Type != newCondition.Type ||
-						oldConditions[0].Reason != newCondition.Reason ||
-						oldConditions[0].Message != newCondition.Message {
-						log.Info("Appending to conditions: ", "namespace", instance.Namespace, "name", instance.Name, "type", newCondition.Type, "reason", newCondition.Reason, "message", newCondition.Message)
-						instance.Status.Conditions = append([]v1beta1.NotebookCondition{newCondition}, oldConditions...)
-					}
-					err = r.Status().Update(ctx, instance)
-					if err != nil {
-						return ctrl.Result{}, err
-					}
-					notebookContainerFound = true
-					break
+				if pod.Status.ContainerStatuses[i].Name != instance.Name {
+					continue
 				}
+				if pod.Status.ContainerStatuses[i].State == instance.Status.ContainerState {
+					continue
+				}
+
+				log.Info("Updating Notebook CR state: ", "namespace", instance.Namespace, "name", instance.Name)
+				cs := pod.Status.ContainerStatuses[i].State
+				instance.Status.ContainerState = cs
+				oldConditions := instance.Status.Conditions
+				newCondition := getNextCondition(cs)
+				// Append new condition
+				if len(oldConditions) == 0 || oldConditions[0].Type != newCondition.Type ||
+					oldConditions[0].Reason != newCondition.Reason ||
+					oldConditions[0].Message != newCondition.Message {
+					log.Info("Appending to conditions: ", "namespace", instance.Namespace, "name", instance.Name, "type", newCondition.Type, "reason", newCondition.Reason, "message", newCondition.Message)
+					instance.Status.Conditions = append([]v1beta1.NotebookCondition{newCondition}, oldConditions...)
+				}
+				err = r.Status().Update(ctx, instance)
+				if err != nil {
+					return ctrl.Result{}, err
+				}
+				notebookContainerFound = true
+				break
+
 			}
 			if !notebookContainerFound {
-				log.Error(nil, "Could not found the Notebook container, will not update the status of the CR. No container has the same name as the CR.", "CR name:", instance.Name)
+				log.Error(nil, "Could not find the Notebook container, will not update the status of the CR. No container has the same name as the CR.", "CR name:", instance.Name)
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -206,23 +206,34 @@ func (r *NotebookReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+					}
+					notebookContainerFound = true
+					break
+				}
 			}
-			err = r.Status().Update(ctx, instance)
-			if err != nil {
-				return ctrl.Result{}, err
+			if !notebookContainerFound {
+				log.Info("WARNING: Notebook container is not found, so could not update State of Notebook CR")
```
yanniszark, line 236: I don't think that a log.Info should emit a `WARNING` message.
If the message is indeed a warning, use `log.Warn`.
This sounds like an error though.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -206,23 +206,34 @@ func (r *NotebookReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-			// Append new condition
-			if len(oldConditions) == 0 || oldConditions[0].Type != newCondition.Type ||
-				oldConditions[0].Reason != newCondition.Reason ||
-				oldConditions[0].Message != newCondition.Message {
-				log.Info("Appending to conditions: ", "namespace", instance.Namespace, "name", instance.Name, "type", newCondition.Type, "reason", newCondition.Reason, "message", newCondition.Message)
-				instance.Status.Conditions = append([]v1beta1.NotebookCondition{newCondition}, oldConditions...)
+
+		if len(pod.Status.ContainerStatuses) > 0 {
+			notebookContainerFound := false
+			for i := range pod.Status.ContainerStatuses {
```
kimwnasptd, line 212: Could you add a comment here to document that the status of the CR will be updated based on the `ContainerState` of the container that has the same name with the CR?

components/notebook-controller/controllers/notebook_controller.go
kimwnasptd: I agree with @yanniszark.
Although AFAIK `logr` does not have a [Warning level](https://github.com/go-logr/logr#why-not-more-named-levels-like-warning). So we should do this an Error instead.

@gilbeckers could you also make the error message more explicit as to why it couldn't find the Notebook Container? Something like: `Could not find the Notebook container. No container has the same name with the CR '{container-name}'. Will not update the status of the CR.`

kimwnasptd (commented):
@gilbeckers 
> This PR partially fixes #5282 (the fact that the status icon of newly created notebook server in the Kubeflow dashboard is not correct)

I think this PR fixes #5282 entirely since the issue was only about the notebooks controller. Could you open a new issue regarding the status icon of the newly created notebook server in the dashboard not being correct?

gilbeckers:
> I think this PR fixes #5282 entirely since the issue was only about the notebooks controller. Could you open a new issue regarding the status icon of the newly created notebook server in the dashboard not being correct?

I updated my original comment

kimwnasptd:
@gilbeckers apologies for the long delay. I would like to push this effort forward as it's something we want to fix for the next release.

Looking again at the code I have some more comments before merging this, mostly to simplify a little bit the code. @gilbeckers would you be interested in putting some small extra effort so we can merge this?

gilbeckers:
@kimwnasptd I'm ok with putting some extra effort in this one. But can you please give some more info on what you have in mind, so that I can estimate what you mean with small effort.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -206,23 +206,37 @@ func (r *NotebookReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-				log.Info("Appending to conditions: ", "namespace", instance.Namespace, "name", instance.Name, "type", newCondition.Type, "reason", newCondition.Reason, "message", newCondition.Message)
-				instance.Status.Conditions = append([]v1beta1.NotebookCondition{newCondition}, oldConditions...)
+
+		// Update status of the CR using the ContainerState of
+		// the container that has the same name as the CR.
+		// If no container of same name is found, the state of the CR is not updated.
+		if len(pod.Status.ContainerStatuses) > 0 {
+			notebookContainerFound := false
+			for i := range pod.Status.ContainerStatuses {
+				if pod.Status.ContainerStatuses[i].Name == instance.Name &&
```
kimwnasptd, line 216: Could you break this if statement to the following ones:

```golang
if pod.Status.ContainerStatuses[i].Name != instance.Name {
    continue
}

if pod.Status.ContainerStatuses[i].State == instance.Status.ContainerState {
    continue
}

log.Info("Updating Notebook CR state: ", "namespace", instance.Namespace, "name", instance.Name)
cs := pod.Status.ContainerStatuses[i].State
...
```

kimwnasptd (commented):
It's just a small reorganization on some if statements to make the code a little bit more clear and not too much nested.

Ideally I think we should work on moving a lot of the logic of the `Reconcile` function into smaller functions. But this should be an independent effort on its own.

gilbeckers:
@kimwnasptd I added your requested changes

kimwnasptd:
The code looks good. Thanks for the contribution @gilbeckers!
/lgtm
/approve

Issue #5310, opened 2020-09-16
Refactor Jupyter web app to utilize the common code


kimwnasptd:
/kind feature

**Why you need this feature:**
We've already created a common Python package for the backend #5164 and an Angular library for the frontend #5252 . We also decided to move our CRUD web apps under the `/components/crud-web-apps` directory #5183.

Following up, we should refactor both the frontend and the backend of JWA to utilize this code. This will help us reduce code duplication significantly and it will also reduce friction on adding new features going forward, since the common logic will be moved elsewhere.

This will be used as a tracking issue for this effort.

**Describe the solution you'd like:**
The changes that need to be done are the following, the order matters:
- [ ] Stop adding more features to `/components/jupyter-web-app` to ensure that the two apps will be synced in terms of features
- [x] Create a new `notebooks` folder under the `/components/crud-web-apps` that will contain the code for the refactored Jupyter web app.
- [x] PR the refactored backend code under `/components/crud-web-apps/jupyter/backend`
- [x] PR the refactored frontend code under `/components/crud-web-apps/jupyter/frontend`
- [x] PR the required {Docker,Make}files as well as READMEs
- [x] Update the CI to build the new web app when the source code is modified from a new PR
- [x] Update our manifests to use the refactored web app

After that we should continue adding new features to the newest web app.

**Anything else you would like to add:**
* The refactored web app will be identical to the current one and it should provide the same functionality the current one does.
* We should be creating issues for each one of the above steps and link them to this one so we can keep track of the work getting done on this
* I'm aiming to have this work merged for Kubeflow 1.2, as well as many of the features from #5280
* We (Arrikto) have already been using this new web app for MiniKF and have tested it extensively

cc @jlewi @jbottum @thesuperzapper

DavidSpek:
@kimwnasptd I believe this issue can be (almost) closed.

kimwnasptd:
Yes it can most certainly be closed. I think all of the above bullets are satisfied so we can move along and close this

/close

PR #5280, opened 2020-09-07
jupyter-web-app rewrite

thesuperzapper:
Hi all, this is a pretty big PR which rewrites large portions of the `jupyter-web-app` UI.

This PR is a first step towards a better UI, I expect significant iteration before Kubeflow 1.2.

## Backend Changes:
- upgrade `Flask_Cors`, `Flask`, and `kubernetes` python packages
  - WARNING: don't forget to run `pip install -r requirements.txt` before testing
- implement PATCH API for Notebooks (used by start/stop feature)

## UI changes (General):
  - specify typography/font settings in `index.html` `<body>` with angular class, rather than star selector in `styles.scss`
  - move from NPM `material-icons` to `material-design-icons-iconfont` (actually has all of the material icons)

## UI changes (Notebook Table):
- overview:
  - see: **((PICTURE 1))**
- reduced the number of columns in the table, and made rows "expandable" to display further info: 
  - see: **((PICTURE 2))**
- implemented "start/stop" feature for notebook servers (as discussed in #4857)
  - see: **((PICTURE 3))**
  - this feature is created by PATCHing a `kubeflow-resource-stopped=true` annotation onto the Notebook Resource
- improved clarity of warning popup for delete/pause of notebooks 
  - see: **((PICTURE 4))**
- ensure notebook age is correctly reported if web server is not in UTC timezone

## UI changes (Notebook Spawner):
- **General:**
  - see: **((PICTURE 5))**
  - improved config descriptions
  - disabled form fields values are now submitted in the backend REST API POST (this is safe because our backend ignores when the user manipulates fields which are `readonly`)
- **Image:**
  - if images are `readonly` users can now select from the administrator provided list, rather than only the default image
- **CPU/RAM:**
  - administrators can set CPU/RAM to `null` (and no cpu/ram request will be made)
    - this allows spawning with only affinity configs (where requests may get in the way)
  - limits introduced by PR #5118 are now disabled by default
    - can be re-enabled with: `.cpu.setLimit` and `.memory.setLimit`
  - warnings will display in the UI if limits are enabled: 
    - see: **((PICTURE 6))**
- **GPU:**
  - administrators can set a list of valid "numbers of GPUs" for the dropdown: `.gpus.value.nums`
- **Volumes (General):**
  - administrators can set default values for volumes added with "new volume": `.dataVolumes.default.*`
  - UI now displays existing volume size
  - UI now disables 'new' type if no DefaultStorageClass exists in the cluster (and we are using `class={none}`)
[...]

PR #5316, opened 2020-09-21
Refactor the JWA backend to utilize common code

kimwnasptd:
first step for #5310 

Will move the backend under `/components/crud-web-apps/jupyter/backend`.

All the backend requests will have the following structure:
```json
{
  "some-field": "some-value",

  "status": 200,
  "success": true,
  "user": "kimwnasptd",
  "log": "this field will be present only when the success is false",
}
```

Need to append a commit for tolerations and affinity.

We (Arrikto) are deploying this jupyter web app on MiniKF and have extensively tested it. So @elikatsis @yanniszark and @StefanoFioravanzo will most probably not have something big to propose in the review.

cc @thesuperzapper

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/routes/patch.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/routes/patch.py
@@ -1,0 +1,80 @@
+import datetime as dt
+
+from flask import request
+from werkzeug import exceptions
+
+from kubeflow.kubeflow.flask_rest_backend import api, decorators, logging
+
+from .. import status
+from . import bp
+
+log = logging.getLogger(__name__)
+
+STOP_ATTR = "stopped"
+ATTRIBUTES = set([STOP_ATTR])
+
+
+# Routes
+@bp.route(
+    "/api/namespaces/<namespace>/notebooks/<notebook>", methods=["PATCH"]
+)
+@decorators.request_is_json_type
+def patch_notebook(namespace, notebook):
+    request_body = request.get_json()
+    log.info("Got body: %s", request_body)
+
+    if request_body is None:
+        raise exceptions.BadRequest("Request doesn't have a body.")
+
+    # Ensure request has at least one valid command
+    if not any(attr in ATTRIBUTES for attr in request_body.keys()):
+        raise exceptions.BadRequest(
+            "Request body must include at least one supported key: %s"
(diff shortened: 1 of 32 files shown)


kimwnasptd:
The last commit for supporting Affinity and Tolerations is there. This is ready for review

components/crud-web-apps/jupyter/backend/apps/common/routes/patch.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # start/stop a notebook
+    if STOP_ATTR in request_body:
+        start_stop_notebook(namespace, notebook, request_body)
+
+    return api.success_response()
+
+
+# helper functions
+def start_stop_notebook(namespace, notebook, request_body):
```
thesuperzapper, line 44: Why does this need to be a separate function?

The patch function is sufficient?

thesuperzapper:
Should we rename the `/jupyter` prefix on the URL to something more generic?

thesuperzapper:
This looks pretty good as a first pass (although I am still sceptical we need a fully separate code path for ROK).

I assume we will implement the changes from https://github.com/kubeflow/kubeflow/pull/5280 when we make the new `notebook-front-end`, as there are lots of backend changes required for those frontend changes.

components/crud-web-apps/jupyter/backend/apps/common/routes/patch.py
kimwnasptd: I was thinking that If in the future we want to handle other cases in the PATCH request, then each one of them should be its own function and not have all the logic unfolded in the PATCH handler

kimwnasptd: I was thinking that If in the future we want to handle other cases in the PATCH request, then each one of them should be its own function and not have all the logic unfolded in the PATCH handler

kimwnasptd (commented):
@thesuperzapper 
> Should we rename the /jupyter prefix on the URL to something more generic?

Was thinking of keeping it as is for now until we replace the the code entirely. Then once we have completely replaced the app we can change the naming both in the routes as well in the manifests naming.

kimwnasptd:
I think we can move on with merging this.
Next step will be to submit the PR for the frontend

PR #5323, opened 2020-09-26
use buster-slim to build the frontend

yhwang:
buster supports more architectures then stretch.
Since it's only used to build the juypter-web-app frontend,
also switch to slim version. The intermidate image size is
reduced from 1.2 GB to 550 MB.

### DIFF
--- a/components/jupyter-web-app/Dockerfile
+++ b/components/jupyter-web-app/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 0: UI Build Stage
-FROM node:10 as build-stage
+FROM node:10-buster-slim as build-stage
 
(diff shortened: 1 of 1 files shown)


Jeffwan:
Thanks for the improvement. @yhwang 

/lgtm
/approve

PatrickXYS:
Feel free to bring it up in WG-Notebook weekly meeting

DavidSpek:
@StefanoFioravanzo @kimwnasptd  Given that we are moving to using the JWA that uses the common code (and that has been moved to buster-slim in https://github.com/kubeflow/kubeflow/pull/5555), can this PR be closed?

PR #5353, opened 2020-10-22
Allow intra namespace traffic by default

robbertvdg:
We have strict RBAC authorization enabled by default for all user traffic. This gives problems when a user deploys an app or database in their namespace. Connections will be terminated due to missing authorization (in our case a valid JWT).

We would like to make an exception for traffic in the user's own namespace, which the following property takes care of.


robbertvdg:
@yanniszark can you review this?

yanniszark:
@robbertvdg thanks for the contribution!
/ok-to-test

PR #5374, opened 2020-10-31
Fixes the default leader election ID

naveensrinivasan:
The default leader election  ID is `controller-leader-election-helper` which could conflict when multiple controllers run within the same namespace. This is a required field in later versions of controller-runtime.

More details on this https://github.com/kubernetes-sigs/controller-runtime/pull/446

### DIFF
--- a/components/notebook-controller/main.go
+++ b/components/notebook-controller/main.go
@@ -47,19 +47,22 @@ func init() {
 }
 
 func main() {
-	var metricsAddr string
+	var metricsAddr, leaderElectionNamespace string
 	var enableLeaderElection bool
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "",
+		"Determines the namespace in which the leader election configmap will be created.")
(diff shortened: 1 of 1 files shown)


naveensrinivasan:
@MartinForReal Why are the tests failing? Can you point me in the direction to resolve it?

naveensrinivasan:
image

The `argo` UI fails to load and I don't see any documentation on how to debug the issue.

MartinForReal:
@naveensrinivasan I think it is a good idea to provide a default value for leader-election-namespace. How about using the namespace in which the controller resides as the default namespace?

naveensrinivasan:
> @naveensrinivasan I think it is a good idea to provide a default value for leader-election-namespace. How about using the namespace in which the controller resides as the default namespace?

@MartinForReal  If the `LeaderElectionNamespace` is empty it would use the `namespace` one https://github.com/kubernetes-sigs/controller-runtime/blob/9975e296899a4843fe69e9d1675fee10957ab694/pkg/leaderelection/leader_election.go#L72

The reason I provided an option was to test the controller from the terminal without deploying it as a container.  This is standard practice for `LeaderElectionNamespace` for dev-test cycle.

MartinForReal:
@StefanoFioravanzo Could you please review this pr? Thanks!

PR #5378, opened 2020-11-04
Implemented functional tests using ginkgo for notebook controller

naveensrinivasan:
The notebook controller can be tested using [sigs.k8s.io/controller-runtime/pkg/envtest ](https://github.com/kubernetes-sigs/controller-runtime/tree/master/pkg/envtest)which comes as part of kubebuilder. With this, we should be able to measurable test coverage.

This will also make the `dev-test` cycle easy for any new developer to be onboarded, to migrate to a higher version of `kubebuilder`, and for refactoring the existing code.

This test is only a start and does not cover comprehensively all the areas, I am planning to add furthermore to it.  

I didn't want a large PR.


naveensrinivasan:
@MartinForReal  ping

yanniszark:
@naveensrinivasan this is a great effort! Thanks a lot for taking the time to write those tests and make a PR.
I can take this one myself.

/unassign @MartinForReal 
/assign @yanniszark 

Keep in mind we have some small issues with CI tests (https://github.com/kubeflow/testing/issues/774), but I let's see if they are resolved by the time I finish this review. I will add it to my backlog for tomorrow.

PatrickXYS:
You can test until https://github.com/kubeflow/testing/issues/774 issue resolved.

components/notebook-controller/controllers/functional_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	beta1 "github.com/kubeflow/kubeflow/components/notebook-controller/api/v1beta1"
```
yanniszark, line 29: nit: Can you please use `nbv1beta1`, for uniformity across the `notebook-controller` codebase?

components/notebook-controller/controllers/functional_test.go
```diff
@@ -0,0 +1,94 @@
...
+				}
+				return true, nil
+			}, timeout, interval).Should(BeTrue())
+			/*
+				Checking for replica count
+			*/
+			By("By checking that the Notebook is running with at least 1 replica")
+			Eventually(func() bool {
+				return notebook.Status.ReadyReplicas > 0
+			}, timeout, interval).ShouldNot(Equal(0))
```
yanniszark, line 91: Is this correct? You return a bool but then check if it equal to zero (False).
So in this case, you actually check if `notebook.Status.ReadyReplicas <= 0`.
I attempted to change this to:

```go
			By("By checking that the Notebook is running with at least 1 replica")
			Eventually(func() int32 {
				// Shouldn't we get the statefulset again?
				return notebook.Status.ReadyReplicas
			}, timeout, interval).ShouldNot(BeZero())
```
but the test failed.

Which leads me to ask, in controller-runtime's test environment, is the statefulset controller running?
If it is running, what happens to the Pods it creates? I assume that they can't be scheduled, as there is no Node for them to be scheduled to. Can you dig a bit deeper on what happens with Pods in controller-runtime's test env and if they have any suggestions?

Finally, you wrapped the replica check in an `Eventually` block, but since the block doesn't get the statefulset again, the retries are meaningless.

yanniszark (changes requested):
@naveensrinivasan I made the first review pass on this. It looks great overall, but it seems you have a bug in the final part of the test. Please take a look at the comments for more information.

Finally, I would like to point out that this would also make a good E2E test and we could benefit from code re-use. Not necessary for now, just something to think about. Cheers!

components/notebook-controller/controllers/functional_test.go
naveensrinivasan: I agree on Pod it creates can't be scheduled without a Node. I will check on that.

Also 🤦  on the replica check in the `Eventually` block without getting the state refreshed.
It should be more like this.

``` go
By("By checking that the Notebook is running with at least 1 replica")
			Eventually(func() int32 {
				err := k8sClient.Get(ctx, notebookLookupKey, createdNotebook)
				if err != nil {
					return 0
				}
				return createdNotebook.Status.ReadyReplicas
			}, timeout, interval).ShouldNot(BeZero())
```

If the controller runtime does not have an option to create `pod` then probably this test would have to be changed. 
But I am not confident about the controller runtime not having that. Let me check and keep you posted.

components/notebook-controller/controllers/functional_test.go
naveensrinivasan: Agreed.

components/notebook-controller/controllers/functional_test.go
naveensrinivasan: @yanniszark  I checked on the `envtest` and confirmed that the `statefulset` controller isn't running.  Only the `api server` itself is run by default. So we cannot check the `pod`. So I am refactoring my tests to check for `statefulset`

naveensrinivasan:
@yanniszark  The ideal thing for the tests should be able to either be able to run against a live cluster or local envtest. That would IMO the right approach the solve for e2e tests. Here is an example of using `kind` https://github.com/microsoft/k8s-cronjob-prescaler/blob/master/Makefile

components/notebook-controller/Makefile
```diff
@@ -25,6 +28,9 @@ check-license:
 
 # Run tests
 test: generate fmt vet manifests check-license
+	mkdir -p ${ENVTEST_ASSETS_DIR}
+	test -f ${ENVTEST_ASSETS_DIR}/setup-envtest.sh || curl -sSLo ${ENVTEST_ASSETS_DIR}/setup-envtest.sh https://raw.githubusercontent.com/kubernetes-sigs/controller-runtime/master/hack/setup-envtest.sh
```
yanniszark, line 32: Can you pin the script at a particular commit?
This way, we can be sure an update in this external repo doesn't break us.

components/notebook-controller/Makefile
```diff
@@ -25,6 +28,9 @@ check-license:
...
+	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR);
```
yanniszark, line 33: This won't work as it is, because make's default shell is `/bin/sh`.
See: https://stackoverflow.com/a/43566158

components/notebook-controller/controllers/suite_test.go
```diff
@@ -0,0 +1,105 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	beta1 "github.com/kubeflow/kubeflow/components/notebook-controller/api/v1beta1"
```
yanniszark, line 35: You missed the ones in this file :)

yanniszark (changes requested):
Thanks for the fixes @naveensrinivasan!

All in all, a couple of nits here and there. In addition, I see that you have added a script to download the binary dependencies needed to run the tests. Personally, I'm not a big fan of shell scripting, it is just too error-prone. In a recent project, I used a Python script to download deps which I feel is much more readable and easy to reason about. See: https://github.com/arrikto/oidc-authservice/blob/master/hack/binary_deps.py

What do you think? We can also address this in a follow-up PR, which will tackle downloading deps. Whatever you feel is best.

naveensrinivasan:
@yanniszark  I am not a fan of shell script either!  Why not use https://github.com/magefile/mage to download the dependencies and scripting? This makes it easier for someone to come on board to use one language which is `go`. 

Let me know your thoughts.

components/notebook-controller/Makefile
naveensrinivasan: Thanks, I will do that. 👍

naveensrinivasan:
@yanniszark  I would like a follow-up PR for downloading the binary dependencies.

yanniszark:
@naveensrinivasan thanks, this is almost ready.
The only think remaining is that `make test` doesn't work.
It doesn't work because of the `check-license` target. I don't believe it makes sense for `check-license` to be a prerequisite for test right now, let's fix one thing at a time. Can you remove the `check-license` dep from the `test` target of the Makefile?
Also, when running `make test`, it modifies some files (YAML configuration and code formatting). Can you also submit those in a commit? In the future, we should automate these checks. After these two things, we should be ready to merge.

naveensrinivasan:
@yanniszark  Removed the `check-license` and included the `YAML` that was updated.

yanniszark:
/lgtm
/approve

Looks great!

PR #5385, opened 2020-11-05
Fixes the leader election ID

naveensrinivasan:
The default leader election ID is `controller-leader-election-helper` which could conflict when multiple controllers run within the same namespace. This is a required field in later versions of controller-runtime.

More details on this kubernetes-sigs/controller-runtime#446

Similar to https://github.com/kubeflow/kubeflow/pull/5374

### DIFF
--- a/components/profile-controller/main.go
+++ b/components/profile-controller/main.go
@@ -48,7 +48,7 @@ func init() {
 }
 
 func main() {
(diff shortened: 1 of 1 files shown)


yanniszark:
@naveensrinivasan thanks! I can take this one as well!
/assign @yanniszark

PR #5387, opened 2020-11-08
Making e2e tests better.

naveensrinivasan:
The goal is to make the `e2e` testing better and be able to run them either on a `production` like cluster or use something like with an in-memory cluster  [kind](https://github.com/kubernetes-sigs/kind) with using [sigs.k8s.io/controller-runtime/pkg/envtest](sigs.k8s.io/controller-runtime/pkg/envtest). 

 This also provides a measurable `coverage` on tests.

The changes in this PR

* Use `ginkgo` and `gomega` as the testing framework
* Use [sigs.k8s.io/controller-runtime/pkg/envtest](sigs.k8s.io/controller-runtime/pkg/envtest)  for testing the API
* Separate out `e2e` tests with notebook_controller_integration_test.go and notebook_controller_envtest_test.go
* notebook_controller_integration_test.go - Expects the notebook controller to run to satisfy the desired state. The ginkgo runner can identify these tests using `--focus="E2E TEST:"`. This can be configured to use an actual `k8s` cluster like `kind` or a larger cluster.
* notebook_controller_envtest_test.go - Expects the API Server to accept the requests for the notebook and does not expect the tests to run the notebook controller to satisfy the desired state. This can be used as a quick sanity test.
* Makefile - The changes to the Makefile are
   +  Downloading and setting up `kind` cluster.
   + Downloading and setting up `ENV` tests.
   + Downloading `ginkgo`
   + Running `e2e` tests

This is based on my other PR https://github.com/kubeflow/kubeflow/pull/5378 which is still in review.


naveensrinivasan:
@yanniszark Took your idea and opened this PR to have the `e2e` tests better. I would appreciate your feedback.

yanniszark:
@naveensrinivasan thanks a lot for the PR, a couple of general comments:
- I believe the standard practice is for E2E tests to live in a separate top-level folder, as they test the whole software, not just one package of it. So I don't think we should put an E2E test under `controllers`. How about adding an `e2e` top-level folder?
- I see you used kind, I've heard of it before. Personally, I have used `k3d`, because it is super fast (11,6s cluster creation time on my PC). How does kind fare? To give you an example, I made a recent blog post about how we used k3d for e2e testing in a recent project:
https://www.arrikto.com/devops/k3d-github-actions-kubernetes-e2e-testing-made-easy/
https://github.com/arrikto/oidc-authservice/blob/master/e2e/e2e_test.go

In addition, I would propose having the cluster-creation logic in Go (not Makefile), so we can have more control (e.g., check an envvar and skip creation if it is set, to support existing clusters). Example code:
https://github.com/arrikto/oidc-authservice/blob/2015dee56f3ee7e7b24a7d90cdb7c932d044b8fb/e2e/e2e_test.go#L416 (you can separate the set image part).

So, the next steps I would recommend are:
1. Create a top-level folder named `e2e` and add a new test (e.g., `e2e/notebook_test.go`).
1. Write the E2E test. Mostly the same as what you have, but also code to:
    - Create the cluster or using existing one, controlled by an envvar.
    - Apply the manifests to deploy the notebook-controller (here we can also explore starting the controller with `go run`, for a quicker test).

What do you think?

naveensrinivasan:
Appreciate the feedback.

-  I agree with the E2E being a top-level folder.
- Yes, `k3d` is way faster than `kind` and it makes sense to use that over `kind`.
- I can move the code to `go` for cluster creation. Thanks to your existing code, I can reuse some of them.

>Apply the manifests to deploy the notebook-controller (here we can also explore starting the controller with go run, for a quicker test).

I understand the manifest deployment with `go`, but I don't understand using `go run` for it.  Could you probably expand on it?

yanniszark:
> I understand the manifest deployment with go, but I don't understand using go run for it. Could you probably expand on it?

What I meant is that once you have a cluster, you can deploy the controller either with a docker image and YAML (closer to the real environment but slower, better for CI) or by using `go run <>` (faster but less realistic, better for development).

naveensrinivasan:
closing this for https://github.com/kubeflow/kubeflow/pull/5400

PR #5394, opened 2020-11-10
Upgrading the `go` compiler version.

naveensrinivasan:
The `go`  compiler `1.12` and `1.11` are no more supported.  

>Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including critical security problems, in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

https://golang.org/doc/devel/release.html

Go also guarantees backward compatibility https://golang.org/doc/go1compat

### DIFF
--- a/components/notebook-controller/Dockerfile
+++ b/components/notebook-controller/Dockerfile
@@ -6,7 +6,7 @@
 #
 # This is necessary because the Jupyter controller now depends on
 # components/common
-ARG GOLANG_VERSION=1.12.5
+ARG GOLANG_VERSION=1.15
 FROM golang:${GOLANG_VERSION} as builder
 
 WORKDIR /workspace
--- a/components/notebook-controller/Makefile
+++ b/components/notebook-controller/Makefile
@@ -2,8 +2,8 @@
 # Image URL to use all building/pushing image targets
 IMG ?= gcr.io/kubeflow-images-public/notebook-controller
 TAG ?= $(eval TAG := $(shell date +v%Y%m%d)-$(shell git describe --tags --always --dirty)-$(shell git diff | shasum -a256 | cut -c -6))$(TAG)
-GOLANG_VERSION ?= 1.11.2
 SHELL := /bin/bash
+GOLANG_VERSION ?= 1.15
 
 # Whether to use cached images with GCB
 USE_IMAGE_CACHE ?= true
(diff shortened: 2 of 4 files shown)


yanniszark:
/assign @yanniszark 
Thanks @naveensrinivasan! Can you rebase your changes ontop of the latest master?
Give me a ping when this PR is ready for review.

naveensrinivasan:
@yanniszark It is rebased against master. Thanks

naveensrinivasan:
@yanniszark Ping

components/notebook-controller/Dockerfile
```diff
@@ -6,7 +6,7 @@
 #
 # This is necessary because the Jupyter controller now depends on
 # components/common
-ARG GOLANG_VERSION=1.12.5
+ARG GOLANG_VERSION=1.15.4
```
Jeffwan, line 9: minor: it would be great to keep golang version same as golang used in k8s dependency.

naveensrinivasan: @Jeffwan  Thanks, I agree with the recommendation and have updated it.

yanniszark: @Jeffwan thanks for the recommendation!
Can you expand a bit on it? I have never encountered it before.
What is the reason for matching the versions of Go?

yanniszark:
@naveensrinivasan sorry for the delay!
/hold
/lgtm
/approve

Everything is good, but I think commit messages could be a bit better.
I was thinking a single commit (instead of 3), similar to this:
```
notebook-controller: Upgrade go version

Upgrade go version of the notebook-controller to 1.15, across the Dockerfile,
Makefile and README. We used the same Golang version as our Kubernetes
dependency, after @Jeffwan's suggestion.
```

What do you think? If you believe it's too much, feel free to remove the hold.

yanniszark:
Also:
/ok-to-test

naveensrinivasan:
> @naveensrinivasan sorry for the delay!
> /hold
> /lgtm
> /approve
> 
> Everything is good, but I think commit messages could be a bit better.
> I was thinking a single commit (instead of 3), similar to this:
> 
> ```
> notebook-controller: Upgrade go version
> 
> Upgrade go version of the notebook-controller to 1.15, across the Dockerfile,
> Makefile and README. We used the same Golang version as our Kubernetes
> dependency, after @Jeffwan's suggestion.
> ```
> 
> What do you think? If you believe it's too much, feel free to remove the hold.

@yanniszark  I can do that.

DavidSpek:
@yanniszark Since the commits were squashed the /lgtm label needs to be applied again.

Issue #5389, opened 2020-11-09
Script to install all the dependencies 


naveensrinivasan:
/kind feature

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
To set up development ENV for contributing it would be great to have a script to download all the dependencies.

**What did you expect to happen:**
The bootstrap script that download's all the dependencies.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`)
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

Issue #3657, opened 2019-07-15
E2E test for jupyter web app


jlewi:
We should have E2E tests for the Jupyter web app using a framework like selenium.

See kubeflow/testing#288

kimwnasptd:
I can try and take that one.

I'll start with some basic end-to-end cases such as creating and deleting Notebooks and will gradually try to add more advanced ones.

/assign kimwnasptd

jlewi:
@kimwnasptd any update on this?

jlewi:
Bump to P0 because this is a requirement for 1.0.

kimwnasptd:
@jlewi I'm reading the [e2e instructions](https://github.com/kubeflow/testing#writing-an-argo-workflow-for-an-e2e-test) in the testing repo and I have some questions.

1. Does each test (.py file) need to be encapsulated in an argo workflow in order to be triggered from prow?
    * If yes, where should this bundle (.py + argo app) live in the code base?
    * Will the argo workflow also need to be added to the repo's `prow_config.yaml` to be run as a postsubmit job?
2. The test code will need to have access to a kubeflow cluster with the latest jupyter-web-app image deployed. How will it know this endpoint? I see the [testing#273](https://github.com/kubeflow/testing/issues/273) but I'm not sure of its current state. 
3. Will the test code be responsible for setting up the kubeflow cluster or is this automated and a kubeflow endpoint could be provided as an ENV Var (for example) in the container that will run the test?

jlewi:
>Does each test (.py file) need to be encapsulated in an argo workflow in order to be triggered from prow?
If yes, where should this bundle (.py + argo app) live in the code base?

An E2E test will be an Argo workflow that contains steps for building and deploying the application and then testing it. The invocation of the python code (e.g. pytests) to test your application will be steps in this workflow.

* The code for the primary E2E test for kubeflow is defined in [kfctl_e2e_workflow.py](https://github.com/kubeflow/kfctl/blob/master/py/kubeflow/kfctl/testing/ci/kfctl_e2e_workflow.py)

* So you can use that as the basis for defining an E2E workflow for the jupyter web app

* If you want to create a presubmit test that you would write an E2E workflow that would have Argo steps for the following

   * Build the jupyter web app
   * Modify the manifests to use your new image
   * Deploy Kubeflow using the updated image
   * Run one or my pytests to verify the jupyter web app is working

[...]

PR #5400, opened 2020-11-12
Implemented E2E testing

naveensrinivasan:
* Implemented E2E testing with `gomega` and `ginkgo`
  * The tests could use a pre-existing cluster or create a new `k3d`
  * Ran `go mod tidy`


naveensrinivasan:
@yanniszark ping

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,240 @@
...
+		return defaultVal
+	}
+	return envVar
+}
+
+// Cleaning the ENV
+var _ = SynchronizedAfterSuite(func() {
+
+}, func() {
+	if !useExistingCluster {
```
yanniszark, line 102: Maybe check clusterName directly?

yanniszark (changes requested):
@naveensrinivasan thanks a lot for the PR! I have left some small comments on the code.
I also wanted to make 2 more general notes:

1. In the code you import the image of the notebook-controller, but you never actually use it, because the test doesn't apply the notebook-controller to the cluster. Instead, it creates a manager and runs the controller locally. What I expected the test to do was to apply manifests  to the cluster (e.g., with `kustomize build` and `kubectl apply`) and then test the behavior of the notebook-controller. What do you think?

1. We are starting to gather binary dependencies. I think so far we have k3d, kubebuilder and its binaries, ginkgo. We currently have Makefile targets to install them. I think it would be much better if we consolidated the handling of these binary dependencies in a separate script. See example: https://github.com/arrikto/oidc-authservice/blob/master/hack/binary_deps.py
We can address this in another PR if you'd like.

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,240 @@
...
+		u := &unstructured.Unstructured{}
+		if err := crdscheme.Convert(obj, u, nil); err != nil {
+			fmt.Println(err, "error converting to unstructured object", "object-kind", obj.GetObjectKind())
+			continue
+		}
+		res = append(res, u)
+	}
+	fmt.Println(len(res))
+	return res
+}
```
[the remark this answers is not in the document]
naveensrinivasan: The code doesn't use the `kustomize` or `kubectl` to deploy the CRD's. I tried deleting the CRD's and it failed. Here is the source of the code which was used as reference https://github.com/kubernetes-sigs/controller-runtime/blob/f52b6180d515bca5f8faa551c3784fee5ce89a83/pkg/envtest/envtest_test.go#L61

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,240 @@
...
+var (
+	crdscheme          = scheme.Scheme
+	letterRunes        = []rune("abcdefghijklmnopqrstuvwxyz") // for generating random names for tests.
+	useExistingCluster bool                                   // flag for using an existing cluster for testing.
+	cfg                *rest.Config
+	k8sClient          client.Client // You'll be using this client in your tests.
+	testEnv            *envtest.Environment
+	clusterName        string                                           // cluster that was dynamically created.
+	crds               []*apiextensionsv1beta1.CustomResourceDefinition // crds installed into the cluster.
+	k8sDockerImage     string
```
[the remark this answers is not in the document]
naveensrinivasan: I can change that.

naveensrinivasan:
>In the code you import the image of the notebook-controller, but you never actually use it, because the test doesn't apply the notebook-controller to the cluster. Instead, it creates a manager and runs the controller locally. What I expected the test to do was to apply manifests to the cluster (e.g., with kustomize build and kubectl apply, see https://github.com/arrikto/oidc-authservice/blob/2015dee56f3ee7e7b24a7d90cdb7c932d044b8fb/e2e/e2e_test.go#L448-L473) and then test the behavior of the notebook-controller. What do you think?


The code applies the CRD using `envtest`  instead of `kubectl apply`
```
// Installing CRD's
	crds, err = envtest.InstallCRDs(cfg,
		envtest.CRDInstallOptions{ErrorIfPathMissing: false,
			Paths: []string{filepath.Join("..", "config", "crd", "bases")}})
```

The test doesn't run the controller locally. By Applying the CRD and deploying docker image
 ```
cmd = exec.Command("k3d", "image", "import", "-c", clusterName, imageName)
	cmd.Stderr, cmd.Stdout = os.Stderr, os.Stdout

yanniszark:
> We have a tracking issue for this #5389 and IMO we should have a discussion over there and then make a decision.

ACK on the binary deps issue. Let's continue the discussion there.

> The test doesn't run the controller locally. By Applying the CRD and deploying the docker image the controller is utilized to test it within the cluster. The controller isn't run locally.

I believe this is not true.
1. If the controller is running inside the docker image, then why are you creating a manager and starting the controller in the go code?
```go
	err = (&controllers.NotebookReconciler{
		Client:        k8sManager.GetClient(),
		Log:           ctrl.Log.WithName("controllers").WithName("notebook-controller"),
		Scheme:        k8sManager.GetScheme(),
		Metrics:       controllermetrics.NewMetrics(k8sManager.GetClient()),
		EventRecorder: k8sManager.GetEventRecorderFor("notebook-controller"),
	}).SetupWithManager(k8sManager)
	Expect(err).ToNot(HaveOccurred())
```

1. The `k3d image import` command just imports the image in the cluster's docker registry. It doesn't create a Pod running the image.

I would expect the test to use the kustomize manifests under `config` in order to deploy the notebook-controller. For example, using `kustomize build config/default`.

naveensrinivasan:
> > We have a tracking issue for this #5389 and IMO we should have a discussion over there and then make a decision.
> 
> ACK on the binary deps issue. Let's continue the discussion there.
> 
> > The test doesn't run the controller locally. By Applying the CRD and deploying the docker image the controller is utilized to test it within the cluster. The controller isn't run locally.
> 
> I believe this is not true.
> 
> 1. If the controller is running inside the docker image, then why are you creating a manager and starting the controller in the go code?
> 
> ```go
> 	err = (&controllers.NotebookReconciler{
> 		Client:        k8sManager.GetClient(),
> 		Log:           ctrl.Log.WithName("controllers").WithName("notebook-controller"),
> 		Scheme:        k8sManager.GetScheme(),
> 		Metrics:       controllermetrics.NewMetrics(k8sManager.GetClient()),
> 		EventRecorder: k8sManager.GetEventRecorderFor("notebook-controller"),
> 	}).SetupWithManager(k8sManager)
> 	Expect(err).ToNot(HaveOccurred())
> ```
> 
> 1. The `k3d image import` command just imports the image in the cluster's docker registry. It doesn't create a Pod running the image.
> 
> I would expect the test to use the kustomize manifests under `config` in order to deploy the notebook-controller. For example, using `kustomize build config/default`.

Yes that is correct the tests are running within the `envtest` and not inside the docker image. I will change the code to the do the kustomize build config/default

naveensrinivasan:
The changes to run notebook controller has been though because the manifests that are required are totally out of sync from the source for example
- RBAC for notebook controller does not have permission defined for `events` https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/config/rbac/role.yaml 

whereas the one that is deployed does have it

https://github.com/kubeflow/manifests/blob/a95c7d0a0eda0066c987df19b1b0ad00d4413101/tests/stacks/generic/test_data/expected/rbac.authorization.k8s.io_v1_clusterrole_notebook-controller-role.yaml

- Another example of the manifests that have been out of sync is when deploying using 
```

**Comment by yanniszark** [comment]
@naveensrinivasan thanks for raising your concerns. Let me try to address them:

> The changes to run notebook controller for e2e within k3d has been though because the manifests that are required are totally out of sync from the source for example

The source for notebook-controller manifests is the config folder. Please don't mind the manifests repo right now.
It will be refactored to copy and reuse the manifests developed in this repo (the upstream app repo).

> RBAC for notebook controller does not have permission defined for events https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/config/rbac/role.yaml

I think it should have! Can you add the relevant comment here:
https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L75-L80
so that kubebuilder picks it up and generates the correct RBAC?

> Another example of the manifests that have been out of sync is when deploying manifests it fails because of this bug kubernetes-sigs/controller-tools#444

Jeez, that seems like a handful. I suspect we're going to have to deal with this for all of Kubeflow.
Instead of trying to solve it here, I would propose to open an issue and use K8s < 1.18 for the E2E tests.

Do the above help you continue? I feel E2E testing is very important and I think you do as well, so I will do my best to unblock your efforts 😄

**Comment by naveensrinivasan** [comment]
>I think it should have! Can you add the relevant comment here:
https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L75-L80
so that kubebuilder picks it up and generates the correct RBAC?

I am aware of that but it looks like there is a bug in regeneration https://github.com/kubernetes-sigs/kubebuilder/issues/1496 🤦

**Comment by naveensrinivasan** [comment]
This PR needs to be merged in https://github.com/kubeflow/kubeflow/pull/5426

File: components/notebook-controller/e2e/suite_test.go [hunk 4]
```diff
@@ -0,0 +1,240 @@
...
+func getEnvDefault(variable string, defaultVal string) string {
+	envVar := os.Getenv(variable)
+	if len(envVar) == 0 {
+		return defaultVal
+	}
+	return envVar
+}
+
+// Cleaning the ENV
+var _ = SynchronizedAfterSuite(func() {
```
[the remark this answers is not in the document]
naveensrinivasan: I agree. Thanks

components/notebook-controller/e2e/suite_test.go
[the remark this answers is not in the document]
naveensrinivasan: I have removed this by just deleting the CRD using `kubectl delete `

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// getEnvDefault returns the ENV variable or the default value.
+func getEnvDefault(variable string, defaultVal string) string {
+	envVar := os.Getenv(variable)
+	if len(envVar) == 0 {
+		return defaultVal
+	}
+	return envVar
+}
```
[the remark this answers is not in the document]
naveensrinivasan: Agreed. Thanks

components/notebook-controller/Makefile
```diff
@@ -95,3 +94,16 @@ build-gcr: test
...
+	curl -sS https://raw.githubusercontent.com/rancher/k3d/1030f0fa47b65d78eb9ffbdf94f0623a08d19b97/install.sh  | TAG=v3.1.0 bash
+endif
+e2e-tests: k3d
+		# Docker image that is used for spinning up the k8s cluster.
+		# K8S_DOCKER_IMAGE=docker.io/rancher/k3s:v1.16.14-k3s1
+		# if TEST_USE_EXISTING_CLUSTER is false then a new k3d cluster would be created to test controller.
+		# TEST_USE_EXISTING_CLUSTER=false
+		# Image to load into the test cluster.
+		# TEST_IMAGE=$(IMG):latest
+		TEST_USE_EXISTING_CLUSTER=false TEST_IMAGE=$(IMG):latest ginkgo --focus="E2E TEST:" -v  --cover  --coverprofile cover.out  ./e2e/...
```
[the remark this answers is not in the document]
naveensrinivasan: Yes, Agreed. Thanks

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,240 @@
...
+	err = cmd.Run()
+	if err != nil {
+		return err
+	}
+
+	kubeconfigPath := path.Join(os.Getenv("HOME"), fmt.Sprintf(".k3d/kubeconfig-%s.yaml", clusterName))
+	os.Setenv("KUBECONFIG", kubeconfigPath)
+	imageName := os.Getenv(TEST_IMAGE)
+	fmt.Println(imageName)
+	cmd = exec.Command("k3d", "image", "import", "-c", clusterName, imageName)
```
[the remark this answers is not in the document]
naveensrinivasan: I will do that. Thanks

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,240 @@
...
+	// +kubebuilder:scaffold:imports
+)
+
+// These tests use Ginkgo (BDD-style Go testing framework). Refer to
+// http://onsi.github.io/ginkgo/ to learn more about Ginkgo.
+
+const (
+	K8S_DOCKER_IMAGE          = "K8S_DOCKER_IMAGE"
+	TEST_USE_EXISTING_CLUSTER = "TEST_USE_EXISTING_CLUSTER"
+	DOCKER_IMAGE              = "docker.io/rancher/k3s:v1.16.14-k3s1" // image used for spinning k8s cluster.
```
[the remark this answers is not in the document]
naveensrinivasan: Agreed. Thanks

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,240 @@
...
+	TEST_IMAGE                = "TEST_IMAGE"
+)
+
+var (
+	crdscheme          = scheme.Scheme
+	letterRunes        = []rune("abcdefghijklmnopqrstuvwxyz") // for generating random names for tests.
```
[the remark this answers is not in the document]
naveensrinivasan: Yes, I have moved them to the function.

naveensrinivasan:
@yanniszark  Updated the code.

naveensrinivasan:
@yanniszark Ping

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -74,10 +74,15 @@ type NotebookReconciler struct {
 
 // +kubebuilder:rbac:groups=apps,resources=statefulsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=apps,resources=statefulsets/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
```
yanniszark, line 77: Let's remove this, the controller never touches deployments.
I saw the manifests, they must be wrong.

components/notebook-controller/e2e/e2e_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	nbv1beta1 "github.com/kubeflow/kubeflow/components/notebook-controller/api/v1beta1"
+)
```
yanniszark, line 31: nit: Shouldn't this be:
```
import (
	"context"
	"fmt"
	"time"

	nbv1beta1 "github.com/kubeflow/kubeflow/components/notebook-controller/api/v1beta1"
	. "github.com/onsi/ginkgo"
	. "github.com/onsi/gomega"
	appsv1 "k8s.io/api/apps/v1"
	v1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/types"
)
```

2 blocks, not 3

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,231 @@
...
+var _ = SynchronizedBeforeSuite(func() []byte {
+	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+	By("Creating a k8s cluster.")
+	if !useExistingCluster {
+		err := bootstrapk8sCluster()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(err).NotTo(HaveOccurred())
+	}
+	existingCluster := true
+	testEnv = &envtest.Environment{UseExistingCluster: &existingCluster}
```
yanniszark, line 101: I think `testEnv` doesn't do anything at all. Should we remove it?

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+// Setting up the ENV. This will run once before all the tests.
+var _ = SynchronizedBeforeSuite(func() []byte {
+	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))
+	By("Creating a k8s cluster.")
+	if !useExistingCluster {
+		err := bootstrapk8sCluster()
+		Expect(err).NotTo(HaveOccurred())
+		Expect(err).NotTo(HaveOccurred())
```
yanniszark, line 98: double line

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	existingCluster := true
+	testEnv = &envtest.Environment{UseExistingCluster: &existingCluster}
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).ToNot(HaveOccurred())
+	Expect(cfg).ToNot(BeNil())
```
yanniszark, line 105: I think `testEnv` doesn't do anything at all. Should we remove it?

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,231 @@
...
+		By("Deleting the deployed CRD's in the existing cluster.")
+		// Deleting CRD's that were deployed as part of the test.
+		err := deleteKustomizations()
+		Expect(err).Should(BeNil())
+	}
+})
+
+// bootstrapk8sCluster creates a k3d cluster with a random name. Loads the docker image into the cluster.
+// The docker image name retrieved from the ENV variable TestImage.
+func bootstrapk8sCluster() error {
```
yanniszark, line 141: nit: `bootstrapK3DCluster` for uniformity?

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,231 @@
...
+func deleteK3DCluster(clusterName string) error {
+	return exec.Command("k3d", "cluster", "delete", clusterName).Run()
+}
+
+// applyKustomizations applies kustomizations into the k8s clusters.
+func applyKustomizations() error {
+	kubectlApply := exec.Command("kubectl", "apply", "-f",
+		filepath.Join("..", "config", "crd", "bases"))
+	kubectlApply.Stdout, kubectlApply.Stderr = os.Stdout, os.Stderr
+	err := kubectlApply.Run()
```
yanniszark, line 180: Is this needed? The `default` kustomization you install below also includes the crd :)

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,231 @@
...
+	return cmd.Run()
+}
+
+// deleteK3DCluster deletes the dynamically created k3d cluster.
+func deleteK3DCluster(clusterName string) error {
+	return exec.Command("k3d", "cluster", "delete", clusterName).Run()
+}
+
+// applyKustomizations applies kustomizations into the k8s clusters.
+func applyKustomizations() error {
```
yanniszark, line 176: Should we make `applyKustomizations` take a list of paths as input, so it's re-usable?

components/notebook-controller/e2e/suite_test.go
```diff
@@ -0,0 +1,231 @@
...
+
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// deleteKustomizations deletes the CRD,RBAC,Deployment and services of the notebook controller from the cluster.
+func deleteKustomizations() error {
```
yanniszark, line 206: Should we make deleteKustomizations take a list of paths as input, so it's re-usable?

components/notebook-controller/config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: IMAGE_URL
+      - image: gcr.io/kubeflow-images-public/notebook-controller:v20201122-v1.2-rc.0-5-g87c6d9c0-dirty-1a1de6
```
yanniszark, line 11: It's not necessary to commit this change. Unfortunately, making the branch dirty after a test is a side-effect of kustomize not allowing configuration with envvars, because everything should be recorded in git. So you can leave this one, maybe we can make this experience better in the future 🤔

components/notebook-controller/e2e/e2e_test.go
yanniszark, line 31: nit: 3 blocks vs 2.

yanniszark (changes requested):
@naveensrinivasan this is awesome, I'm really looking forward to including this in our CI! :D

A general comment I have is that it would help me (the reviewer) immensely if the commits were a bit better organized so that each commit focuses on one thing and explains what it does. For example, a good series of commits would be:

- notebook-controller: Update outdated kubebuilder RBAC annotations
- notebook-controller: Generate manifests
- notebook-controller: Implement E2E test
- notebook-controller: Update Makefile with targets for e2e testing
- notebook-controller Misc Makefile changes


Some other things I noticed are that:
- Because of the changed `sed` command in the Makefile, the `docker-build` target failed. Can you explain what you tried to do?
- The `manifests` target changes the CRD definition. I assume you made manual changes to the definition in order to work with k8s 1.18? Can we keep the controller-gen goodness for now and tackle the k8s 1.18 issue another time? We have to tackle it for all of KF and I don't think this is the right place. My expectation is that nothing should change if I run `make manifests` or `make generate`.

The good news is that I managed to run the e2e test (for k8s 1.17) on my pc and it worked! We are VERY close :D

naveensrinivasan:
@yanniszark Ping. The longer it takes to merge the PR, the higher likely would it becoming stale.

yanniszark:
@naveensrinivasan I did a checkout of your code, but it seems that your commits are interleaved with commits from master. I'm guessing this happened because you merged master on your branch. For pull requests, we always use rebase. So that new code commits are always on top of existing code commits. Can you please rebase ontop of latest master and remove all merge commits?

naveensrinivasan:
I am on vacation and will comeback and do it. Thanks @yanniszark

DavidSpek:
Does this PR only implement E2E testing for the notebook controller? If so, it might be a good idea to reflect that in the title as I thought it might be related to https://github.com/kubeflow/kubeflow/issues/3657 as well.

yanniszark:
@DavidSpek good suggestion! @naveensrinivasan any updates on this?

yanniszark:
@naveensrinivasan any updates on the PR? The main thing needed is for you to rebase your code on top of the latest master and remove the merge commits.

DavidSpek:
bump

Issue #5419, opened 2020-11-20
Clean up dead code in kubeflow/kubeflow?


jlewi:
Follow on to kubeflow/community#454?

The question posed in that issue is:

> What's the kubeflow/kubeflow positioning? Should we deprecate it once notebook and control plane components move out of it?

With the exception of notebooks and one or two other packages (e.g. PodDefaults) that should probably move into WG owned  repositories (#5418) it looks like most of the code in kubeflow/kubeflow is dead code. We should probably delete it and aim to whiddle down the repo to just a readme.

Looking at the top level directories

* ~~bootstrap - this can be deleted; all relevant code should now be in kubeflow/kfctl~~
* ~~build - This looks like some utilities for managing license files in code~~
  * We should probably keep it but might want to move it
  * kubeflow/testing is generally where we've been gathering shared engprod resources 
* components - This directory should be whiddled down to nothing
  * Sub directories should either be moved into WG owned repositories or be deleted if they are no longer being developed/maintained
* docs/gke - This should probably move to a GCP directory (gcp-blueprints?) or be deleted if no longer used
* hack - should probably move to kubeflow/testing for shared engprod
* ~~metric-collector - I don't believe this is being maintained so we can probably just delete it~~
   * If a WG wants to claim it they should move it to one of their repos
* ~~py/kubeflow - I don't think this is being used anymore so it can probably be deleted~~
  * It looks like its just engprod scripts so if its being used it should move to kubeflow/testing
* ~~releasing - dead code; just delete~~
* ~~scripts~~
  * This looks like outdated scripts that have been superseded by other solutions
  * WGs should probably claim anything if they still need it.
  * I would suggest just deleting the directory to shakeout who might still be depending on it
* testing
  * Similar to scripts
  * Any generic testing code still in use should live in kubeflow/testing
* ~~tf-controller-examples~~
  * Delete; looks very stale
  * Any operator specific examples should move into training operator repos
* ~tools/gcb - delete~

Jeffwan:
`docs_dev` seems need some updates, it should be renamed to `docs`?

code snippet in `docs/gke` seems still helpful. I can leave to @Bobgy to remove to right place.

Jeffwan:
I will move some useful scripts in `hack` and `scripts` to kubeflow/testing

Bobgy:
I'll take a look about gcp stuff, thanks a lot for taking care of others @Jeffwan !

DavidSpek:
[...]

PR #5424, opened 2020-11-21
Clean up dead codes

Jeffwan:
Resolve part of https://github.com/kubeflow/kubeflow/issues/5419

I will leave testing related codes in a separate PR. 

Remove dead codes in 

- tf-controller-examples
- tools
- bootstrap
- metrics-collector
- releasing
- scripts
- hack 
- build


Bobgy:
/lgtm
/approve
Thanks @Jeffwan !

Since tests are passing, I think we can just continue. If anyone find out anything they want to claim, they can always find them in git history.

PR #5425, opened 2020-11-21
Remove components no longer being maintained

Jeffwan:
Resolve part of #5419 

- Reorganize tensorflow image folder. 
  - Move contrib images to this folder
  - Move README.md and deprecate outdated build_image.py
- Remove components specific to GCP. I am not sure if they are being used. If so, maybe it's better to move to gcp-blueprints? 
  - private-utils
  - static-config-server
  - echo-server
  - ingress-setup-image. 
- Remove openmpi-controller. User should use kubeflow/mpi-operator instead

GCP part needs review from @Bobgy  /cc @kubeflow/google


alsrgv:
cc @tgaddair

Bobgy:
@jlewi can you help me confirm is any of the gcp stuff is still being used? I took a glance and it seems they aren't.

jlewi:
* echo-server is a useful debug utility for IAP; its not specific to IAP but I think that's the only thing its used for
* ingress-setup-image I thought we might be using that docker image to run some of the scripts to setup IAP but maybe not.

Either way I think its fine to delete these as if we are still using them we should probably move them to a GCP specific repo.

/lgtm
/approve

Thanks @Jeffwan for the cleanup.

PR #5426, opened 2020-11-22
Rollback the CRD to controller-gen 2.0 instead of 2.5. The controller-gen 2.5 has some breaking changes.

naveensrinivasan:
In my previous commit  https://github.com/kubeflow/kubeflow/commit/b78618d3da77535154a5452a056a2ba56db2eab7, the CRD was generated with controller-gen `2.5` instead of `2.0` 

This will make the CRD similar to one in the manifests 
 https://github.com/kubeflow/manifests/blob/59aeca3ae76957926b8ba7f443d229dda51dc6b6/jupyter/notebook-controller/base/crd.yaml

### DIFF
--- a/components/notebook-controller/config/crd/bases/kubeflow.org_notebooks.yaml
+++ b/components/notebook-controller/config/crd/bases/kubeflow.org_notebooks.yaml
@@ -3,18 +3,14 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  annotations:
-    controller-gen.kubebuilder.io/version: v0.2.5
   creationTimestamp: null
   name: notebooks.kubeflow.org
 spec:
   group: kubeflow.org
   names:
     kind: Notebook
-    listKind: NotebookList
     plural: notebooks
-    singular: notebook
-  scope: Namespaced
+  scope: ""
   subresources:
     status: {}
   validation:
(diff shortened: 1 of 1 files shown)


Jeffwan:
@naveensrinivasan Any reason to use elder controller-tools? We should always use newer version. I think it doesn't matter there's difference between manifest in `kubeflow/manifest`.  If `kubelfow/manifest` is outdated, we can have the update

naveensrinivasan:
> @naveensrinivasan Any reason to use elder controller-tools? We should always use newer version. I think it doesn't matter there's difference between manifest in `kubeflow/manifest`. If `kubelfow/manifest` is outdated, we can have the update

@Jeffwan I agree we need to use the latest version of controller-gen, but that has an impact on changing other things like this https://github.com/kubernetes-sigs/controller-tools/issues/444. The previous commit was unintentional. I have created an issue for this https://github.com/kubeflow/kubeflow/issues/5433 and IMO that should be a separate PR.

A

naveensrinivasan:
The ideal way to upgrade this is when [kubebuilder 3.0](https://github.com/kubernetes-sigs/kubebuilder/milestone/21) is released. 

This is to address things like
```
* spec.validation.openAPIV3Schema.properties[spec].properties[template].properties[spec].properties[initContainers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
* spec.validation.openAPIV3Schema.properties[spec].properties[template].properties[spec].properties[containers].items.properties[ports].items.properties[protocol].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```

This is because we have `corev1 objects` embedded within the CRD. This issue was addressed specifically by upstream https://github.com/kubernetes/kubernetes/commit/d3e641e84ec9e5a9e551ef394509f65c6e6782b6 .

There are hacky ways of fixing the generated CRD with `sed` to add additional tags. But those aren't clean IMO.

DavidSpek:
Since the last comment Kubebuilder v3.0.0-alpha.0 was released (December 12th), and v3.0.0-alpha.1 4 days ago on Jan 8th. As it seems like the release of Kubebuilder v3.0 will be released relatively soon (previous time frame from alpha to release being similar to the time that has passed for this PR already), is it worth waiting for the release?

DavidSpek:
Kubebuilder V3.1.0 has been released 6 days ago, so we should probably aim to upgrade rather than downgrade.

PR #5427, opened 2020-11-22
Fix notebook controller RBAC permission

naveensrinivasan:
The notebook controller doesn't have the necessary markers for generating correct RBAC permissions. 
- Updated the code based on the manifests for generating the correct `RBAC` permissions https://github.com/kubeflow/manifests/blob/a95c7d0a0eda0066c987df19b1b0ad00d4413101/tests/stacks/generic/test_data/expected/rbac.authorization.k8s.io_v1_clusterrole_notebook-controller-role.yaml
- Updated the controller-gen to `2.0` instead of `2.5` which caused the CRD to be regenerated. 

These changes are required to address the issues with https://github.com/kubeflow/kubeflow/pull/5400

### DIFF
--- a/components/notebook-controller/config/crd/bases/kubeflow.org_notebooks.yaml
+++ b/components/notebook-controller/config/crd/bases/kubeflow.org_notebooks.yaml
@@ -3,18 +3,14 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  annotations:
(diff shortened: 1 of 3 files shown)


naveensrinivasan:
FYI @yanniszark

yanniszark:
/assign @yanniszark
Thanks!

DavidSpek:
@yanniszark If the RBAC permissions for the notebook controller are indeed incorrect, this should probably be fixed for the release.

PR #5469, opened 2020-12-18
centraldashboard: redirect non-iframed apps

StefanoFioravanzo:
**Problem statement**:

Centraldashboard provides a shared JS library that Kubeflow apps can use
to get the selected namespace. This PR extends this shared lib by allowing
apps to redirect themselves under a Centraldashboard iFrame, if they are
not iFramed. This is particularly useful when one open a KF app link in a
new tab, thus losing the `/_/` iFrame prefix and consequently not seeing
anymore the Centraldashboard sidebar.

**Solution**:

* Implement a new `dashboard-settings` API which returns the `settings`
  field of the Centraldashboard ConfigMap.
* Extend the shared lib to call the settings endpoint if the app
  is not iFramed. If `DASHBOARD_FORCE_IFRAME` in the settings is set to
  `true`, then redirect the app under the `/_/` prefix.
* An app can opt out from the global config by setting the 
  `disableForceIframe` flag in the `init` function.

/cc @kimwnasptd
/cc @elikatsis

### DIFF
--- a/components/centraldashboard/config/centraldashboard-config.yaml
+++ b/components/centraldashboard/config/centraldashboard-config.yaml
@@ -79,5 +79,5 @@ data:
     }
 kind: ConfigMap
 metadata:
-  name: centraldashboard-links-config
+  name: centraldashboard-config
   namespace: kubeflow
--- a/components/centraldashboard/config/centraldashboard-config.yaml
+++ b/components/centraldashboard/config/centraldashboard-config.yaml
@@ -1,5 +1,9 @@
 apiVersion: v1
 data:
+  settings: |-
+    {
+      DASHBOARD_FORCE_IFRAME: true
+    }
   links: |-
     {
       "menuLinks": [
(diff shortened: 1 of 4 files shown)


components/centraldashboard/config/centraldashboard-config.yaml
```diff
@@ -1,5 +1,9 @@
 apiVersion: v1
 data:
+  settings: |-
```
kimwnasptd, line 3: Does this file live alongside the source code in order to allow the developers to easily apply it and play around with it?

I'm expecting that the manifests will also define such a ConfigMap object. We will also need to update the manifests with this change right?

kimwnasptd (commented):
Thanks for the PR @StefanoFioravanzo! The redirect functionality makes sense.

Left some questions around the ConfigMap. The code LGTM

components/centraldashboard/config/centraldashboard-config.yaml
StefanoFioravanzo: Yeap this is just an example so that we can give people a quick reference on how to configure centraldashboard

And yes, we will need to update the manifests with this change as well. I will make sure to include this when we refactor the manifests in Jan

kimwnasptd: ACK, thanks for confirming!

PR #5471, opened 2020-12-21
Migrate to controller-runtime 0.7.0

jbrette:
- Consolidate Makefile and Dockerfile
- Update all the go.mod to use controller-runtime 0.7.0
- Update all the go.mod to use Kubernetes 1.19.2
- Update go code to account for new context parameters
- Update code-generator to latest version
- Run go mod tidy
- Run "make fmt".
- Update golang to 1.15


naveensrinivasan (commented):
This is a large PR. Without tests (pre-existing and new), it would be risky. I would recommend having tests that would help with confidence in these changes.

PR #5463, opened 2020-12-15
web-apps(front): Update the common frontend library

kimwnasptd:
This is a prerequisite for #4758 

This PR updates the common code and introduces the following features:
* Introduce new reusable components 
* Improve the way we use the common table component in all the crud-web-apps
* Adds frontend CSRF protection in the common frontend library #5461
* Adds unit tests to the common frontend library #5462

Also cc @DavidSpek, after we merge this I'll be able to push the frontend PR for the Volumes UI

We (Arrikto) are already using this code from the web apps in MiniKF and we've tested it extensively. 

/cc @StefanoFioravanzo 
/cc @elikatsis 
/cc @thesuperzapper 
/assign @kimwnasptd

PR #5472, opened 2020-12-21
web-apps(back): Add CSRF protection mechanisms to the backends

### LINKED ISSUE #5461: Ensure the crud-web-apps are CSRF protected
/kind feature

We need to ensure that our [crud-web-apps](https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps) are CSRF protected for insecure methods [POST, PUT, PATCH etc]

We can follow the [double-submit](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie) solution from OWASP. Angular has support for sending this cookie in a [header](https://angular.io/guide/http#security-xsrf-protection). 

This solution should be implemented in the common frontend and backend code to ensure that all of our web apps [jupyter, tensorboards etc] are protected.

cc @kubeflow/wg-notebook-leads

kimwnasptd:
closes #5461
This is the backend PR. In order to fully close the issue we also need to merge #5463 

I also added some small changes wrt:
* parsing strings. We should gradually move away from `f"strings"` 
* allow the backend to skip the Subject Access Review check in some cases

/cc @elikatsis 
/assign @kimwnasptd

### DIFF
--- a/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/csrf.py
+++ b/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/csrf.py
@@ -1,0 +1,112 @@
+"""
+Cross Site Request Forgery Blueprint.
+
+This module provides a Flask blueprint that implements protection against
+request forgeries from other sites. Currently, it is only meant to be used with
+an AJAX frontend, not with server-side rendered forms.
+
+The module implements the following protecting measures against CSRF:
+- Double Submit Cookie.
+- Custom HTTP Headers.
+- SameSite cookie attribute.
+
+To elaborate, the `Double Submit Cookie` procedure looks like the following:
+1. Browser requests `index.html`, which contains the compiled Javascript.
+2. Backend sets the `CSRF_COOKIE` by calling `set_cookie`. If the cookie
+   already exists, `set_cookie` overrides it with a new one. The cookie
+   contains a random value.
+3. Frontend (`index.html`) is loaded and starts making requests to the backend.
+   For every request, the frontend reads the `CSRF_COOKIE` value and adds a
+   `CSRF_HEADER` with the same value.
+4. Backend checks that the value of `CSRF_COOKIE` matches the value of
+   `CSRF_HEADER`. All endpoints are checked, except the index endpoint and
+   endpoints with safe methods (GET, HEAD, OPTIONS, TRACE).
+
+Custom Headers (`CSRF_HEADER`) provide an extra layer of protection, as
+cross-origin requests cannot include custom headers (assuming CORS is not
+misconfigured) because of the Same-Origin policy.
+
+The SameSite cookie attribute provides another layer of protection, but may
+impede usability so it is configurable. This attribute controls whether a
+cookie is sent by the browser when a cross-site request is made. It defaults to
+"Strict".
+
+References:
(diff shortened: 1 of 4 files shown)


kimwnasptd:
Also 
/cc @thesuperzapper

kimwnasptd:
Since @elikatsis is on PTO, @yanniszark could you take a look at these backend changes?

yanniszark:
@kimwnasptd sure! I see that you have included 4 commits, but 1 is for CSRF and the other are code refactorings. Did you mean to include them in this PR or did they slip?

components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/csrf.py
```diff
@@ -0,0 +1,108 @@
...
+4. Backend checks that the value of `CSRF_COOKIE` matches the value of
+   `CSRF_HEADER`. All endpoints are checked, except the index endpoint and
+   endpoints with safe methods (GET, HEAD, OPTIONS, TRACE).
+
+Custom Headers (`CSRF_HEADER`) provide an extra layer of protection, as
+cross-origin requests cannot include custom headers (assuming CORS is not
+misconfigured) because of the Same-Origin policy.
+
+The SameSite cookie attribute provides another layer of protection, but may
+impede usability so it is configurable. This attribute controls whether a
```
yanniszark, line 30: The SameSite attribute is currently not configurable. Should we change the docstring to reflect that? I see that you've added a note that we may need to make it configurable in the future.

kimwnasptd:
Good catch @yanniszark! I must have slipped them while I was rebasing my PR branches.
I'll create a new separate PR with the small fixes and I'll only keep the CSRF changes here

yanniszark (changes requested):
Thanks @kimwnasptd. On the CSRF commit, I only have a small nit in the docstring.

components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/csrf.py
kimwnasptd: Lets make it configurable then, it should be a small change.

I'll make it look for a `CSRF_SAMESITE` env var with a default value to `Strict`. If the value provided by the user is not `Strict`, `Lax` or `None` then again it will default to `Strict`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#SameSite

kimwnasptd: force pushed. The PR should only contain one commit responsible for the CSRF functionality.
I also made the `SameSite` attribute of the cookie configurable via the `CSRF_SAMESITE` variable.

I also think a good next step would be to document the ENV Vars that each web app uses in the respective READMEs. This will make clear to the users how they can configure each app.

yanniszark: That's great! One question: maybe it's better to start the env var section in the README now and document this specific settings with this PR, which introduces it. You can add a note that the section is a work in-progress so that readers don't get confused. What do you think?

kimwnasptd: ACK! I modified the README of the common code.
I also created #5483 as an umbrella issue for the web apps

yanniszark:
/lgtm
Thanks!

PR #5474, opened 2020-12-21
centraldashboard: Improve sidebar UX

StefanoFioravanzo:
This PR improves the Centraldashboard sidebar in the following ways:

- Add icons to sidebar items. To apply this, change the menu links in the configmap as follows

 ```json
"menuLinks": [
{
    "link": "/jupyter/",
    "text": "Notebooks",
    "icon": "book"
},
]
```

- Make the sidebar persistent when in widescreen and collapsible with the viewport is shrinked.

- Make it possible to declare collapsible sections. E.g. the following config will create a collapsible section titled "Experiments", with two items inside ( A single item can be of type `type: item`, which is the default if omitted):

```json
{
  "type": "section",
  "text": "Experiments",
  "icon": "done-all",
  "items": [
    {
      "link": "/pipeline/#/experiments",
      "text": "Pipelines"
    },
    {
      "link": "/katib/",
      "text": "HP Tuning"
    }
  ]
}
```

/cc @kimwnasptd 
/cc @elikatsis


components/centraldashboard/public/components/main-page.js
```diff
@@ -220,34 +218,34 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
...
             break;
         case 'manage-users':
-            this.sidebarItemIndex = this.menuLinks.length
-                                    + this.externalLinks.length + 1;
             this.page = 'manage-users';
             hideTabs = true;
             allNamespaces = true;
-            hideSidebar = true;
+            hideSidebar = false;
+            // need to ues the shadowRoot selector instead of this.$ because
```
kimwnasptd, line 239: small typo, ues -> uses

components/centraldashboard/public/components/main-page.js
```diff
@@ -303,36 +308,58 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
...
-                .findIndex((m) => path.startsWith(m.link));
-            if (menuLinkIndex >= 0) {
-                this.sidebarItemIndex = menuLinkIndex + this.menuLinks.length+1;
-            } else {
-                this.sidebarItemIndex = -1;
-            }
+    _setActiveMenuLink(path, hashPath) {
+        const htmlElements = this._clearActiveLink();
+        let matchPath = path;
+        if (hashPath && hashPath !== '') {
```
kimwnasptd, line 320: the ```&& hasPath !== ''``` is redundant. It will not go into the if body when ```hashPath``` is `''`

components/centraldashboard/public/components/resources/kubeflow-icons.js
```diff
@@ -4,6 +4,11 @@
 */
 import '@polymer/iron-icon/iron-icon.js';
 import '@polymer/iron-iconset-svg/iron-iconset-svg.js';
+import '@polymer/iron-icons/communication-icons.js';
```
kimwnasptd, line 7: Do we need all of these new imports or only a subset of them?

kimwnasptd (commented):
Left some small comments, the code LGTM overall. Nice work @StefanoFioravanzo!

components/centraldashboard/public/components/main-page.js
StefanoFioravanzo: Good catch, thanks!

components/centraldashboard/public/components/main-page.js
StefanoFioravanzo: Oh you are right, thanks!

components/centraldashboard/public/components/resources/kubeflow-icons.js
StefanoFioravanzo: This is a good point but:

1. I wouldn't know how to import just specific icons from these files, I had tried but didn't find a way
2. Even if it was possible to import just specific icons, I would prefer to import these bundles so that people can then configure the sidebar with whatever icons they like (via ConfigMap) without having to change the source code

kimwnasptd: Makes sense, lets keep them then

kimwnasptd:
@StefanoFioravanzo the changes LGTM. 

I see there are conflicts with the latest master, maybe because of the previous PR on the Central Dashboard? Could you rebase the changes on top of the latest master?

StefanoFioravanzo:
@kimwnasptd Everything should be okay now!

kimwnasptd:
Nice! 
/lgtm
/approve

Issue #5004, opened 2020-05-07
Email validation regex is too restrictive


reem:
/kind bug

**What steps did you take and what happened:**

Attempted to add a contributor with an email address on a `.network` domain. Received "Contributor doesn't look like a valid email address"

**What did you expect to happen:**

User is added as normal.

**Anything else you would like to add:**

This regex: https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/app/api_workgroup.ts#L11-L12

Is used to validate emails, and is too simple. It demands that your TLD contains only 2 or 3 characters, which rejects many valid domains.

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`)
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

jlewi:
Did you try adding the user directly by creating appropriate RBAC and ISTIORBAC resources?

reem:
This is still an issue in the latest kubeflow release, the regex is still rejecting completely valid emails, issue is not stale. Adding the users directly is a very involved workaround for end users to have to go through.

reem:
Now that pipelines are scoped by user/namespace, this has become a pretty strong blocker for granting access to important namespaces for these users.

yanniszark:
/reopen
/lifecycle frozen

Hi @reem! The contributor UI is not a very well-maintained component.
Its backend (KFAM) is following a road to deprecation.

So I have 2 suggestions:
First, did this approach work for you?
> Did you try adding the user directly by creating appropriate RBAC and ISTIORBAC resources?

It's not UI-driven, but enables you to understand the necessary permissions and possibly build tooling on top of that.

Another way would be to make a contribution to relax this validation check, if the UI is useful for your case.

kim-sardine:
There's a conflict between this regex pattern and [manual-profile-creation](https://www.kubeflow.org/docs/components/multi-tenancy/getting-started/#manual-profile-creation). "manual-profile-creation" have no validation on email. It's just a string.

on centraldashboard's "Manage Contributors" page, I can't remove manually-created contributor(static user) from namespace  because of this regex.

DavidSpek:
@kim-sardine Could you maybe raise a new issue about this, as this issue is regarding the regex being too restrictive. The problem of no email validation happening in the profile CR is quite different.

[...]

PR #5506, opened 2021-01-08
Expand the email regex to allow more TLDs

reem:
Fixes https://github.com/kubeflow/kubeflow/issues/5004 by allowing top level domains up to 7 characters in length

### DIFF
--- a/components/centraldashboard/app/api_workgroup.ts
+++ b/components/centraldashboard/app/api_workgroup.ts
@@ -9,7 +9,7 @@ import {
 } from './api';
 
 // From: https://www.w3resource.com/javascript/form/email-validation.php
-const EMAIL_RGX = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/;
+const EMAIL_RGX = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,7})+$/;
 
 // Valid actions for handling a contributor
 type ContributorActions = 'create' | 'remove';


reem:
@googlebot I signed it!

DavidSpek:
@yanniszark Is there a reason to use a self made regex instead of a [RFC 5322](https://www.ietf.org/rfc/rfc5322.txt) compliant regex for email addresses? A quick search gives the following regex, which is a corrected version of one that seems very wide spread:
```
(?:[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*|"(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21\x23-\x5b\x5d-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])*")@(?:(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?|\[(?:(?:(2(5[0-5]|[0-4][0-9])|1[0-9][0-9]|[1-9]?[0-9]))\.){3}(?:(2(5[0-5]|[0-4][0-9])|1[0-9][0-9]|[1-9]?[0-9])|[a-z0-9-]*[a-z0-9]:(?:[\x01-\x08\x0b\x0c\x0e-\x1f\x21-\x5a\x53-\x7f]|\\[\x01-\x09\x0b\x0c\x0e-\x7f])+)\])
```
Source: https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression

yanniszark:
@DavidSpek I agree that we should not reinvent the wheel.
The code actually mentions an external source: https://www.w3resource.com/javascript/form/email-validation.php
Which prescribes: `/^[a-zA-Z0-9.!#$%&'*+/=?^_`{|}~-]+@[a-zA-Z0-9-]+(?:\.[a-zA-Z0-9-]+)*$/`
But the code doesn't actually use that!

@reem can you try the regex prescribed by
https://stackoverflow.com/questions/201323/how-to-validate-an-email-address-using-a-regular-expression
and mentioned by @DavidSpek and report if either one fixes your issue?

DavidSpek:
@yanniszark I looked over the comment with the source so excuse my claim about reinventing the wheel. I also check the source and the regex that is now in the code is not used on that page except for a comment. 
An uncorrected version of the regex I mentioned above can be tested here:
https://www.regextester.com/115911
Filling in the corrected version yields the same results.

reem:
Honestly validating emails using a regex in general is incorrect and likely to reject many valid emails especially as TLDs continue to evolve. This issue itself is proof of that. If anything I would suggest replacing the regex with a much much simpler check or simply validating it on the server using an actual email parser. The number of different and disagreeing regex's you can find on the internet for this task is ample evidence and I'm sure we can all relate to having our valid emails rejected by overly zealous validation.

That said I could still update this PR to use the linked even more complex regex here, though my time to thoroughly test the change is limited.

DavidSpek:
@reem I saw that as well. For now, I believe the regex from my comment above should solve issues for most people. After that it might be worth looking at properly validating emails.

DavidSpek:
@reem Would you be able to commit the suggested change so this PR can be merged?

DavidSpek:
@reem Are you wanting to continue with this PR?

DavidSpek:
@yanniszark Given that there hasn't been a response on the PR for a while, should I recreate it so it can be merged and included in the 1.3 release?

DavidSpek:
@yanniszark @kimwnasptd I'm closing this PR as the author has not responded and I have made the discussed changes in https://github.com/kubeflow/kubeflow/pull/5679. 
/close

PR #5513, opened 2021-01-13
Centraldashbaord: Fix namespace navigation

### LINKED ISSUE #5484: Cards Recent Pipelines/Runs in Central Dashboard should be populated with items in the selected namespace
/kind feature

Central Dashboard tries to populate *Recent Pipelines* and *Recent Pipeline Runs* cards with the 5 most recently created Pipelines and Runs respectively with no regards to selected namespace.
However, it is not compatible with the changes done to resolve issue [#4197](https://github.com/kubeflow/pipelines/issues/4197). That is, we should populate the card with the last 5 elements that correspond to the selected namespace. 
On the other hand, in multiuser mode and with the new Pipeline API, it is mandatory to pass namespace to both, the pipelines and runs endpoints, otherwise we get an error.

Make Central Dashboard to pass selected namespace to the corresponding Pipelines API endpoints.
If no namespace is selected, then do not populate the card but show a message like 'Choose a namespace to see Pipelines/Runs'. In other words, the behavior should be similar to the *Recent Notebooks* cards

This issue is related to [#4197](https://github.com/kubeflow/pipelines/issues/4197) resolved by PR [4835](https://github.com/kubeflow/pipelines/pull/4835), so Central Dashboard can be compatible with that.

tasos-ale:
This PR improves and fixes how Centraldashabord handles namespace selection:
- Correctly track the selected namespace when it is changed in the namespace select box
- Use the tracked namespace in the homepage cards
- Call Kubeflow Pipelines API using the namespace filter

Closes #5484

/cc @kimwnasptd
/cc @StefanoFioravanzo

### DIFF
--- a/components/centraldashboard/public/components/dashboard-view.js
+++ b/components/centraldashboard/public/components/dashboard-view.js
@@ -15,9 +15,10 @@ import './notebooks-card.js';
 import './pipelines-card.js';
 import './resource-chart.js';
 import {getGCPData} from './resources/cloud-platform-data.js';
+import utilitiesMixin from './utilities-mixin.js';
 
 
-export class DashboardView extends PolymerElement {
+export class DashboardView extends utilitiesMixin(PolymerElement) {
     static get template() {
         return html([`
             <style include="card-styles">
@@ -34,7 +35,10 @@ export class DashboardView extends PolymerElement {
         return {
             documentationItems: Array,
             quickLinks: Array,
-            namespace: String,
+            namespace: {
+                type: Object,
+                observer: '_namespaceChanged',
+            },
             platformDetails: Object,
             platformInfo: {
                 type: Object,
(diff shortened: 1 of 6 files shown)


elikatsis:
@googlebot I consent

DavidSpek:
I believe https://github.com/kubeflow/pipelines/pull/4933 is slightly relevant for this PR.

elikatsis:
@DavidSpek 
>I believe kubeflow/pipelines#4933 is slightly relevant for this PR.

Could you elaborate?
I think they are irrelevant. That KFP PR touches the KFP python client making it return a different URL prefix so that the links the client outputs are inside the centraldashboard.
This is, IMHO, incorrect. As @StefanoFioravanzo has commented (https://github.com/kubeflow/pipelines/pull/4933#discussion_r556480791), now the centraldashboard library takes care of this (https://github.com/kubeflow/kubeflow/pull/5469).

DavidSpek:
@elikatsis With slightly relevant I meant that this PR (but as it turns out more https://github.com/kubeflow/kubeflow/pull/5469) is relevant to https://github.com/kubeflow/pipelines/pull/4933 and if might need to be reworked.

StefanoFioravanzo:
Thanks @tasos-ale these changes look great!

/lgtm
/approve

Issue #4838, opened 2020-03-09
Enable JupyterLab view by default


Jeffwan:
/kind feature

**Why you need this feature:**
https://github.com/kubeflow/kubeflow/issues/4818

Currently, user has to change url to `/lab` to switch to lab view. I am thinking if we want to enable lab view as default option and user can still switch to default view. 

Any concerns or insights? If not, I will update container images.

hemantha-kumara:
@Jeffwan Have you tried setting in `--NotebookApp.default_url='/lab'` option in Dockerfile CMD?
`jupyter notebook --notebook-dir=/home/cjup --ip=0.0.0.0 --no-browser --allow-root --port=8888 --NotebookApp.token='' --NotebookApp.password='' --NotebookApp.allow_origin='*' --NotebookApp.base_url=${NB_PREFIX} --NotebookApp.default_url='/lab'`

We have created custom image with above mentioned command and by default /lab is opening whenever we connect notebook.

hemantha-kumara:
@Jeffwan 
As discussed in #4818 it would be good if we could handle settings from configmap instead of changing in all Dockerfile.

May be we can put /lab by default in configmap.

Jeffwan:
@hemantha-kumara Make sense. Trying to see any concerns to make that. If not, I will go change configmap

jrykr:
@Jeffwan is the plan still to implement this via configmap? I just came across this issue after I too changed to `/lab` manually. Looks like the issue has gone stale, and I couldn't locate a solution elsewhere.

Jeffwan:
@jrykr I feel like it's much easier to change image. This should be default view. I think I will add it in next release.

DavidSpek:
I agree that Jupyter Lab should be the default view and that it should be set in the images. This should be discussed with the notebook WG.

PR #5515, opened 2021-01-13
Change default view to lab

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/4838.

### DIFF
--- a/components/tensorflow-notebook-image/Dockerfile
+++ b/components/tensorflow-notebook-image/Dockerfile
@@ -123,5 +123,5 @@ RUN docker-credential-gcr configure-docker && chown ${NB_USER}:users $HOME/.dock
 EXPOSE 8888
 USER jovyan
 ENTRYPOINT ["tini", "--"]
-CMD ["sh","-c", "jupyter notebook --notebook-dir=/home/${NB_USER} --ip=0.0.0.0 --no-browser --allow-root --port=8888 --NotebookApp.token='' --NotebookApp.password='' --NotebookApp.allow_origin='*' --NotebookApp.base_url=${NB_PREFIX}"]
+CMD ["sh","-c", "jupyter lab --notebook-dir=/home/${NB_USER} --ip=0.0.0.0 --no-browser --allow-root --port=8888 --NotebookApp.token='' --NotebookApp.password='' --NotebookApp.allow_origin='*' --NotebookApp.base_url=${NB_PREFIX}"]
 


kimwnasptd:
@DavidSpek  from what I remember this shouldn't work since JupyterLab was had different cmd arguments. The `--NotebookApp.ip` argument for example should be `--LabApp.ip`. At least that was the case a [couple of years back](https://github.com/kubeflow/kubeflow/pull/2349/files#diff-f3f0008fed2b80d1da069d653213382a430869c171abf1447c1529af87aa50a8R19), but I'm not sure if they've changed it.

Have you tried building a docker image of a Notebook and test it?

DavidSpek:
@kimwnasptd I've been using this CMD string for months with my personal images that are all using the jupyter-stack as a base. Also, `--NotebookApp.ip` doesn't currently exist `--NotebookApp.token=''` `--NotebookApp.password=''` `--NotebookApp.allow_origin='*'` `--NotebookApp.base_url` are all fine from my experience. I can double check the documentation if you'd like though.

DavidSpek:
@kimwnasptd Found it quickly, there are no `--LabApp.xx` arguments. https://jupyter-notebook.readthedocs.io/en/stable/config.html

kimwnasptd:
Aren't the docs for Notebooks different for labs?
https://jupyter-notebook.readthedocs.io/en/stable/config.html
https://jupyterlab.readthedocs.io/en/stable/

The only references to `LabApp` I can find are in the [Advanced Usage](https://jupyterlab.readthedocs.io/en/stable/user/directories.html?highlight=labapp#disabling-rebuild-checks) section where they mention the `LabConfig`. I can't seem to find any more references to `--LabApp` so maybe indeed these are remnants from the past.

I'll try out the command with a "latest" Notebook docker image and will resume the review of this.

DavidSpek:
@kimwnasptd Given https://github.com/kubeflow/kubeflow/issues/5516 I'd say this PR has been tested at this point.

DavidSpek:
@kimwnasptd I think this can be merged.

kimwnasptd:
I tested the above change be editing the `args` field of the pod template spec inside a notebook and the change works as expected. So it seems that the `--LabApp.*` args are no longer used.

LGTM from my end, but I'll wait for @StefanoFioravanzo to take a look and merge this, since he has more experience wrt the actual Notebook environments.

/lgtm

DavidSpek:
@kimwnasptd The Kale Dockerfiles seem to use the `--LabApp` annotation. So just to be sure I hadn't missed something looked it up again and I see lots of references to `--LabApp` usage on Google, but I can't find it in the docs myself. Either way, it seems like both `--LabApp` and `--NotebookApp` work regardless. Maybe @StefanoFioravanzo knows what the difference is.

StefanoFioravanzo:
@kimwnasptd @DavidSpek These `--LabApp` and `--NotebookApp` arguments correspond to the keys used in the Jupyter config file that one can generate with `jupyter {notebook,lab} --generate-config`.

By looking at the two configs, the `lab` version still uses `NotebookApp` as prefix, apart from a few of them that use the `LabApp` one.

All the CLI overrides we provide in the Dockerfile entrypoint are listed as `NotebookApp` fields even in the `lab` config. So I am good with it.

(Since we can't find any Lab-specific documentation on this, I believe they indeed support both version, but let's just stick with what the default configuration file generates.)

PR #5522, opened 2021-01-17
Use node:12 as specified in the docs

DavidSpek:
The docs specify that the crud-web-apps use node 12 in multiple places. However, the Dockerfiles use node 10 for the build. This PR updates the Dockerfiles to use node 12


DavidSpek:
@kimwnasptd I can confirm that node:12-buster-slim works for building the TWA when https://github.com/kubeflow/kubeflow/pull/5528 and https://github.com/kubeflow/kubeflow/pull/5531 are applied. So this should be ready for merging.

kimwnasptd:
Thanks for the contribution David!

/lgtm
/approve

PR #5528, opened 2021-01-18
twa(front): Add missing dependencies for tensorboard web app

DavidSpek:
This PR adds the missing dependencies needed to build the tensorboards web app. 
closes https://github.com/kubeflow/kubeflow/issues/5529
/cc @kimwnasptd


kimwnasptd:
thank you for your PR @DavidSpek! 
As I also mentioned in the issue there are two more [peerDependencies](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/package.json#L14-L15) from the library that we need to add
```
date-fns: ^1.29.0,
lodash-es: 4.17.11
```

Also, please modify the PR's description so that it will also close the issue once the PR is merged.

DavidSpek:
@kimwnasptd I added date-fns and lodash-es. The reason I did not link the issue is because there still is a bug when trying to build the TWA frontend even after adding these dependencies.

kimwnasptd:
> The reason I did not link the issue is because there still is a bug when trying to build the TWA frontend even after adding these dependencies.

ACK! Lets just mention something like "partially closes \<issue\>" in the future, so that we still link to the parent issue, in case we have a chain of PRs.

The changes look good. I also tested the web app and the frontend builds/starts as expected. Thanks for the effort @DavidSpek!

/lgtm
/approve

PR #5530, opened 2021-01-18
add instructions for building crud docker images directly

DavidSpek:
This PR adds a small README to the crud-web-apps directory explaining how to build the docker images directly. 

/cc @kimwnasptd

### DIFF
--- a/components/crud-web-apps/README.md
+++ b/components/crud-web-apps/README.md
@@ -1,0 +1,8 @@
+# Build instructions for docker images
+To build the Jupyter and Tensorboards docker images directly (without the Makefile) for development purposes, the following commands can be used from this directory.
+
+## Building the Jupyter Images
+`cp jupyter/Dockerfile.dockerignore .dockerignore && docker build --pull -f "jupyter/Dockerfile" -t <image-tag> . && rm .dockerignore`
+
+## Building the Tensorboard Images
+`cp tensorboards/Dockerfile.dockerignore .dockerignore && docker build --pull -f "tensorboards/Dockerfile" -t <image-tag> . && rm .dockerignore`


components/crud-web-apps/README.md
```diff
@@ -0,0 +1,8 @@
+# Build instructions for docker images
+To build the Jupyter and Tensorboards docker images directly (without the Makefile) for development purposes, the following commands can be used from this directory.
```
kimwnasptd, line 2: I don't like the phrasing of this sentence, as it implies that for development purposes we need to run the command directly and can't use the Makefifle.

We can use the Makefile to build Jupyter app's image by running
```bash
REGISTRY_PROJECT=my-repo make docker-build
```

[ Tensorboard's makefile uses slightly different vars but we should iron them out and have the same template when we setup the CI/CD ]

components/crud-web-apps/README.md
```diff
@@ -0,0 +1,8 @@
...
+
+## Building the Jupyter Images
+`cp jupyter/Dockerfile.dockerignore .dockerignore && docker build --pull -f "jupyter/Dockerfile" -t <image-tag> . && rm .dockerignore`
```
kimwnasptd, line 5: I think it's difficult to read one big command so could you instead use the triple \` markdown and break the commands with `\`? It would look like this
```bash
cp jupyter/Dockerfile.dockerignore \
&& docker build --pull -f "jupyter/Dockerfile" -t ${IMG} \
&& rm .dockerignore
```

kimwnasptd (commented):
Thanks for the PR @DavidSpek!

Let's not include these commands in a global README, but rather add them to each app's README. This way the developers won't have to jump between directories to work with the components. Everything relative to an app will be in its own README.

Let's create a distinct `Build image` section for each app where we:
1. Mention how to build the images using the Makefiles
2. Also explain how to run the Makefile's command by hand

Ideally we shouldn't need to manually copy paste the dockerignore file but AFAIK there is no command line option to configure the location of the dockerignore.

DavidSpek:
Arguable this PR is not very good and unnecessary so I am closing it so that it doesn't stay in the backlog forever.

PR #5538, opened 2021-01-19
cluster admin support multiple user split by comma

qqsun8819:
now kfam can only accept a single cluster admin in command line
we should support pass multiple cluster admins  into kfam ,and then we can update profiles-profiles-config after kubeflow is deployed


DavidSpek:
@qqsun8819 Would you be able to rebase this PR on the master branch. This way the changes to our CI/CD will be picked up and we can continue with it more quickly.

Issue #5457, opened 2020-12-13
Enable Dependabot for security updates


naveensrinivasan:
/kind feature

Enable the Dependabot for security updates within the Github project.
![image](https://user-images.githubusercontent.com/172697/101999789-82b7ed00-3cae-11eb-9daf-dfe6017a1804.png)

**Why you need this feature:**
[Is your feature request related to a problem? Please describe in details]
There are critical security vulnerabilities that aren't patched. This will enable addressing those issues proactively. 

Here is an example of security issues reported

![image](https://user-images.githubusercontent.com/172697/101999863-11c50500-3caf-11eb-90d0-b7c413d76874.png)

**Describe the solution you'd like:**
[A clear and concise description of what you want to happen.]

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

DavidSpek:
@naveensrinivasan Given the open PRs from the dependabot, isn't this already enabled?

naveensrinivasan:
Yes!

DavidSpek:
/reopen 

While dependabot is turned on, it does not seem like it is properly configured. This should be solved with https://github.com/kubeflow/kubeflow/pull/5542

DavidSpek:
Closing this as Renovate will be used (hopefully) in the near future. 
/close

PR #5542, opened 2021-01-20
tests: add dependabot config script

DavidSpek:
closes: https://github.com/kubeflow/kubeflow/issues/5457

Inspired by https://github.com/kubeflow/pipelines/issues/4682 I created a script that will create a config file for depandabot so that it knows what directories to scan. It will scan the repository for files named `*ockerfile*`, `package*.json`, `*requirements.txt` and `go.*`. It is setup for dockerfiles, npm, gomod and python at the moment, and I believe this should cover almost all the code in the repo. It is trivial to further customize what folders are selected if further customization is needed. It also parses the closest OWNERS file for a given directory and assigns the approvers and adds the reviewers to the PRs it creates. 

As it stands now, there are about 95 PRs that will be created with this configuration, so it might be advisable to have some form of plan to implement it in stages or be ready to quickly go through lots of the PRs. Another option is to create a target branch for all these PRs so they can be merged into that first rather than master.

For reference, the PRs that will be created can be found here: https://github.com/DavidSpek/kubeflow/pulls


yanniszark:
@DavidSpek is it possible to turn the dependabot on for only some components?
For example, I'd like to have some testing CI before going forward with this in notebook-controller:
https://github.com/kubeflow/kubeflow/pull/5400

CI validation would enable us to merge with confidence. Merging 95 PRs without some CI in place doesn't seem viable.

Issue #5552, opened 2021-01-25
Reduce the size of crud-web-apps images


kimwnasptd:
/kind feature

**Why you need this feature:**
Right now the crud-web-apps are using the `python:3.7` base docker image, as the last step of their multi build process. This can lead into images of almost 1GB for just some python code + html,css,js files.

**Describe the solution you'd like:**
We could use the `-slim-buster` variants of the Python docker images to significantly reduce the size of the produced images.

DavidSpek:
I'll look into this a bit. I'm assuming you are referring to the CPU image, and note the GPU image which will be larger due to the NVIDIA drivers and such.

kimwnasptd:
Woo hoo!

kimwnasptd:
I'm reopening the issue since our current images use `make` to run as an entrypoint, which is not installed in the buster-slim versions of the docker images.

I'll fix the entrypoint and test it e2e after I fix #5568

DavidSpek:
@kimwnasptd can you take a look at the linked PR that should close this issue and fix the web app? https://github.com/kubeflow/kubeflow/pull/5606

PR #5555, opened 2021-01-26
Move JWA python images to slim-buster

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5552

Reduced the image size of the JWA from 928 MB to 163 MB. Similarly, the TWA image is now 162 MB.


kimwnasptd:
Thanks for picking it up @DavidSpek 

/lgtm
/approve

Issue #5482, opened 2020-12-22
Run pre/post-submit tests with the Optional-test infra


kimwnasptd:
/kind feature

**Why you need this feature:**
We will need to be able to run pre/post-submit jobs as a part of the [Notebooks WG](https://github.com/kubeflow/community/blob/master/wgs.yaml#L211) and a testing infrastructure. This way we can:
* Run unit and e2e tests to ensure future PRs don't break the existing functionality
* Automatically build new images for our artifacts, when their code is modified from a PR
* Create PR to update the manifests with the newly built images, from the previous step

**Describe the solution you'd like:**
We [ the Notebooks WG ] would like to integrate with the Optional-test infra, similar to how the [kfctl](https://github.com/kubeflow/kfctl/pull/427) repo does

A first good step would be to create a presubmit job that will run the unit tests for the [crud-web-app's common code](https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps/common/frontend/kubeflow-common-lib).

kimwnasptd:
cc @kubeflow/wg-notebook-leads 

/assign @kimwnasptd 
/assign @PatrickXYS

PatrickXYS:
There's a new doc I wrote recently for onboarding optional-test-infra https://github.com/kubeflow/testing/blob/master/aws/Doc/onboarding.md

It may be helpful for WG-notebook, will come back in Jan for further discussion

Aut0R3V:
Hey, anything I can do here?

kimwnasptd:
Hey there @Aut0R3V, the work for this issue will be broken down into two main tasks:
1. Configure the `kubeflow/kubeflow` repo and the Optional-test infra prow cluster to be able to run pre/post submit jobs
2. Write the python files that prow jobs will execute to generate an Argo workflow that will be run

We are working with @PatrickXYS on the first step. Once the Optional-test infra prow cluster is ready to run jobs from the `kubeflow/kubeflow` repo then we can coordinate on writing the python files for the Argo workflows.

I'll be posting updates of the entire process here so you, and everyone interested in this, can be in the loop

kimwnasptd:
@PatrickXYS for the first step of the [on-boarding process](https://github.com/kubeflow/testing/blob/master/aws/Doc/onboarding.md), could you create a PR similar to kubeflow/internal-acls#373 to give write permissions for the `kubeflow/kubeflow` to the third-party-bots?

[...]

PR #5564, opened 2021-01-27
web-apps(front): Make the library tests run in CI

kimwnasptd:
A first stop for adding a presubmit #5482 job for running unit tests in the common front end code.

The presubmit tests will run inside a container, as an Argo task/step. In order to run the tests inside a container, without any interactions, we will need to make the following changes:
1. Run a headless version of Chrome, to run the tests
2. Don't run the testing process, `ng test`, in watch mode. By default watch mode is on, in order for the process to track changes and run the tests while code is modified.

With the above changes we can run the tests of the common frontend code by navigating to the frontent code, where the `package.json` lives, and running:
```bash
docker run --rm -t \
  -v $(pwd):/mnt/project \
  --user 1000:1000 \
  --workdir /mnt/project \
  -e NG_CLI_ANALYTICS='false' \
  node:12.20.1-stretch-slim \
  npm ci && npm run test-ci
```

/cc @StefanoFioravanzo 
/cc @elikatsis 
/assign @kimwnasptd

### DIFF
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/package.json
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/package.json
@@ -6,6 +6,7 @@
     "start": "ng serve",
     "build": "ng build && npm run copyCSS && npm run copyAssets",
     "test": "ng test",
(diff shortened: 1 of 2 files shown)


kimwnasptd:
@StefanoFioravanzo could you also add an lgtm?

PR #5567, opened 2021-01-28
web-apps: Fixes to the common code

kimwnasptd:
This PR introduces various fixes into both the common frontend and backend code. 
Since there are a lot of small ones I didn't create separate issues for each one of them, but here's a list of what's included:

Backend:
* Allow backend to skip authz checks when listing Pods
* Add ability to fetch pod logs
* Correctly calculate the default storage class

Frontend:
* Global SCSS classes for common types
* Export common modules to be used from the apps without requiring explicit imports
* Make the backgrounds of the apps white, and not grey due to CentralDashboard
* Move the toolbar buttons to the right, like the Pipelines UI
* Create a component for showing explanatory text. This will be useful for forms
* Create a component for a bottom bar for submitting CRs in forms

The bulk of the code is the two new components, but they are only view components [without any specialized logic].
I also included some basic tests for the new components and `ng test` passes as expected.

/cc @StefanoFioravanzo 
/cc @elikatsis 
/assigh kimwnasptd

### DIFF
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/form/submit-bar/submit-bar.component.scss
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/form/submit-bar/submit-bar.component.scss
@@ -1,0 +1,45 @@
+:host {
+  display: block;
+  width: 100%;
+}
+
+.link {
+  color: blue;
+  text-decoration: underline;
+}
+
+.link {
+  color: blue;
+  text-decoration: underline;
+  cursor: pointer;
+}
+
+.bar {
+  padding: 0.5rem 0;
+}
+
+.bar > * {
+  margin-top: auto;
+  margin-bottom: auto;
+}
+
+.bar > button:first-child {
+  margin-left: 35%;
+}
+
+.bar > button:nth-child(2) {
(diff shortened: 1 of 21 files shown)


components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/form/submit-bar/submit-bar.component.scss
```diff
@@ -0,0 +1,45 @@
+:host {
+  display: block;
+  width: 100%;
+}
+
+.link {
+  color: blue;
+  text-decoration: underline;
+}
```
tasos-ale, line 9:
```suggestion

```
You probably don't need that, you have `.link` class twice.

tasos-ale (commented):
Hey, @kimwnasptd I took the liberty of taking a quick look at the PR and left some comments, but of course, I'll also let the other reviewers take a stab at this.

kimwnasptd:
@tasos-ale thank you very much for the review! All of your comments make sense, I'll integrate them in this PR.

tasos-ale:
@googlebot I fixed it

kimwnasptd:
@tasos-ale thanks again for the feedback. I've accepted your suggestions and pushed another commit for the button change.

PR #5569, opened 2021-01-29
web-apps(back): Export package code for wheel

### LINKED ISSUE #5568: Common crud-apps backend code wheel is missing the code
/kind bug

1. Build a docker image for on of the crud web apps
2. Run the image and start a python process
3. import `kubeflow` module
4. Rases a `ModuleNotFound` exception, despite the fact that it should be [installed](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/Dockerfile#L38)

The `kubeflow` module should be imported as expected.

The build process uses the [setuptools.find_packages()](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/backend/setup.py#L18) to figure out which packages to include. But the directory does not define any Python packages with `__init__.py` files. To fix this we should define such files.

/assign @kimwnasptd

kimwnasptd:
closes #5568 

We should define `__init__.py` files to denote our packages.

/cc @elikatsis 
/assign @kimwnasptd

### DIFF
--- a/components/crud-web-apps/common/backend/.gitignore
+++ b/components/crud-web-apps/common/backend/.gitignore
@@ -1,1 +1,3 @@
 kubeflow.egg-info
+dist
+build


elikatsis (changes requested):
I believe that we don't need `components/crud-web-apps/common/backend/__init__.py`.
This is because it is the root of the python package, and `setup.py` denotes it. We need to add `__init__.py` files in the directories that are _below_ the root of the package.

For context, here's some good documentation on python packages structure: https://docs.python-guide.org/writing/structure/

kimwnasptd:
@elikatsis you are right! We shouldn't have __init__.py files in the root directory.
I pushed a new commit

Issue #3362, opened 2019-05-27
Support theia or microsoft visual studio code


jlewi:
It is often useful to have a proper IDE when developing models. For example, even when using Jupyter, users may choose to organize python code into modules. They might also want to create additional files like YAML files.

Jupyter's editor capabilities are limited.

Microsoft visual studio code is really nice. In addition, there are an increasing number of plugins adding Kubernetes support to visual studio code.

It would be nice if users could easily run Microsoft visual studio code in a side car of their jupyter pods so they could combine VS code's editor capabilities with jupyter.

coder.com provides a remote server for running visual studio code.

I did some prototyping and it seems pretty straightforward to run vscode in a sidecar of jupyter
https://github.com/jlewi/kubeflow-dev/tree/master/vscode

One issue I've run into is that cdr/code-server#670 prevents it from working behind a reverse proxy.

/cc @kimwnasptd 
/cc @aronchick

LeonardAukea:
@jlewi Nice idea. Would it not make more sense to use vs-code IDE locally and easily communicate with kubeflow apis? In the case of connecting to a server -> https://devblogs.microsoft.com/python/remote-python-development-in-visual-studio-code/
https://github.com/Microsoft/vscode-dev-containers . 
So what I'm getting at is the ability to attach vscode to a running server.

jlewi:
I don't think the too are mutually exclusive.
There are two use cases for running vscode inside the cluster

1. For demos/tutorials it would be nice to have an IDE installed in the cluster and accessible via web browser so we don't have to deal with getting an editor installed locally
1 Being able to edit code stored inside a pod or PVC in the cluster without out having to copy the files back and forth to your local machine to edit.

jlewi:
There's also Theia
https://theia-ide.org/

jlewi:
Thei might be a better option for licensing and Kubernetes reasons. Thei is fully open source.

LeonardAukea:
This is maybe a little much -> https://github.com/ml-tooling/ml-workspace but it has some good things.  I will probably spend a little time next week, to see if I can get something working. Our data sience teams and me included are not solely using notebooks. Also, now that VS Code has nice .ipynb support+ipykernel, jupyterlab is not getting used as much as before.

fojtami1:
[...]

PR #5570, opened 2021-01-30
Add experimental R-Studio images

DavidSpek:
Related to the discussion https://github.com/kubeflow/kubeflow/issues/3362 and https://github.com/kubeflow/kubeflow/issues/2208.

This PR adds experimental R-Studio images based of off the popular images from the [rocker project](https://www.rocker-project.org/). The way this image works is by having an embedded instance of Apache2 rewrite the url base from the `NB_PREFIX` environment variable. Due to RStudio having hardcoded redirects to `/auth-sign-in` and `/auth-sign-out` for loading the authentication cookie, even with authentication disabled, it is necessary to also inject the `NB_PREFIX` to the `www-root-path` variable in the `rserver.conf`. Without this, the R-Studio interface will not load and redirect to `<kubeflow_url>/auth-sign-in?appUri=` and it does not seem like this can be fixed with Apache or NGINX by the rewrtire rules ([related merge](https://github.com/rstudio/rstudio/pull/7628), [related code](https://github.com/rstudio/rstudio/blob/ddcd7191ec89c4da00e77afae7e9f27e61e87c36/src/gwt/www/templates/encrypted-sign-in.htm#L312). This image does not use a Jupyter image as base with R-Studio accessible through the `jupyterlab/server-proxy` extension, and thus is more lightweight and smaller (1.96GB regular, 7.1GB gpu). The GPU image uses `rocker/ml:4.0.3-cuda11.1` as its base.

There are some things that are non-optimal, and thus why I think the image should be seen as experimental. 
1. The JWA only allows mounting the workspace volume to `/home/jupyter`, but the default user in this image is `rstudio`. While the image can probably be changed, the hardcoded workspace volume probably needs a better solution.
2. While `kubectl` does work when access from inside the pod (with `kubectl exec -it`), it does not work in the terminal of R-Studio and results in `The connection to the server localhost:8080 was refused - did you specify the right host or port?`. I have not yet been able to figure out how to debug this. 
3. I have not yet tested installing or using the KFP SDK. 
4. I have not yet tested Shiny, this will possibly require some additions to the Apache proxy configuration. 

Any feedback to improve the image or overcome the above mentioned points (mainly 1 and 2) are much appreciated. 

For people that would like to test these images, they are available as `davidspek/rstudio-kubeflow:0.28` and `davidspek/rstudio-kubeflow:0.27-cuda`

### DIFF
--- a/components/notebook-controller/config/base/cluster-role-binding.yaml
+++ b/components/notebook-controller/config/base/cluster-role-binding.yaml
@@ -1,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: role-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: role
+subjects:
+- kind: ServiceAccount
+  name: service-account
--- a/components/notebook-controller/config/base/cluster-role.yaml
+++ b/components/notebook-controller/config/base/cluster-role.yaml
@@ -1,0 +1,107 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: role
+rules:
+- apiGroups:
+  - apps
+  resources:
+  - statefulsets
+  - deployments
+  verbs:
+  - '*'
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - services
+  verbs:
+  - '*'
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+- apiGroups:
+  - kubeflow.org
+  resources:
+  - notebooks
+  - notebooks/status
+  - notebooks/finalizers
+  verbs:
+  - '*'
+- apiGroups:
+  - networking.istio.io
+  resources:
+  - virtualservices
+  verbs:
+  - '*'
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: kubeflow-notebooks-admin
+  labels:
+    rbac.authorization.kubeflow.org/aggregate-to-kubeflow-admin: "true"
+aggregationRule:
+  clusterRoleSelectors:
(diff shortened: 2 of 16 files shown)


DavidSpek:
Regarding issue 2, it seems to be a user permission error or something similar. While `exed` into the pod directly, the root user is able to use kubectl as one would expect. However, when executing `sudo -u rstudio kubectl get pods -n admin` to run kubectl as the `rstudio` user, the error `The connection to the server localhost:8080 was refused - did you specify the right host or port?` also comes up. 
This seems to indicate that the issue is not related to R-Studio itself or the web terminal.

StefanoFioravanzo:
Thanks @DavidSpek this is awesome! I will definitely need to take some time to take a deep look at this. Since you mentioned the `group 1337` thing, note that it comes from Istio. The Istio sidecar runs with user UID `1337` and uses that UID to mount a secret into the Pod. I don't remember the details but you can look at the `securityContext` of the Pod, specifically the part changed by the Istio mutating webhook

DavidSpek:
@StefanoFioravanzo Do you have any idea how Kubectl chooses what config to load and why it isn't finding the service account other than as the root user?

StefanoFioravanzo:
@DavidSpek let's take it from the beginning. You said:

> While kubectl does work when access from inside the pod (with kubectl exec -it), it does not work in the terminal of R-Studio and results in The connection to the server localhost:8080 was refused - did you specify the right host or port?. I have not yet been able to figure out how to debug this.

So, what user are you running with when logged into the container? What user are you running with when using the R-Studio terminal? Make sure you are using the same user in both places first and let's see if you see a conflicting behavior

DavidSpek:
@StefanoFioravanzo When logged into the container directly it uses the Root user. In R-Studio the user `rsudio` is used. Kubectl works for the root user, but not when combined with `sudo`. For the user `rstudio` kubectl doesn't work, also not when using `sudo`. 

Also, for point 1 regarding the workspace volume, would a drop list with values `/home/jovyan` and `/home/rstudio` be an option?

DavidSpek:
It does seem like the R-Studio user can read everything in ` /var/run/secrets/kubernetes.io/serviceaccount/` without an issue.

DavidSpek:
/hold
A newer version is available in https://github.com/kubeflow/kubeflow/pull/5582 that also contains a workaround for `kubectl`

PR #5571, opened 2021-01-31
Add experimental VS-Code image

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/3362

This PR adds an experimental VS Code image which makes use of the [code server](https://github.com/cdr/code-server) package. The way this image works is by having an embedded instance of Apache2 rewrite the url base from the NB_PREFIX environment variable, similar to the experimental R-Studio images. To make the image more easy to customize, a custom image was created which makes use of the [s6 overlay](https://github.com/just-containers/s6-overlay), again similar to the R-Studio images. With this overlay, scripts placed in `/etc/cont-init.d` or `/etc/services.d/` are executed at container start and support the use of environment variables. The image in its current state is only 753 MB and has the KFP SDK installed. 

There are some things that are non-optimal, and thus why I think the image should be seen as experimental.
1. I have not yet found a way to have the s6 overlay start as user `jovyan`. While passing the environment variable `USER` during `docker run` did seem to work, having it in the Dockerfile did not. However, code-server itself is run as user `jovyan`. This is accomplished by the line `exec s6-setuidgid jovyan /usr/local/bin/code-server ...`. 
2. <s>Sadly `kubectl` does not seem to work, neither in the browser based terminal nor from inside the pod. </s> This seems to have been due to the method of installation, when installed from apt it works in the web UI, as well as from the pod directly.
3. While the KFP SDK is installed, I have not yet tested it. 
4. Further customization of the image will be necessary, such as deciding on a default set of extensions to install (adding the `--install-extension` flag to code-server) and other tools such as tensorflow or pytorch
5. Consider the creation of a GPU image that includes the necessary drivers

As always, any feedback to improve the image or overcome the above mentioned points (mainly point 1) are much appreciated.

For those that would like to test this image, it is available as `davidspek/vs-code-kubeflow-custom:0.31`
Kubectl was added in 0.31

### DIFF
--- a/components/vs-code-experimental/Dockerfile
+++ b/components/vs-code-experimental/Dockerfile
@@ -1,0 +1,105 @@
+FROM ubuntu:20.04
+
+# ARG OVERLAY_VERSION=2.1.0.2
+# ARG OVERLAY_ARCH="amd64"
+# ARG S6_OVERLAY_MD5HASH=7e81e28fcb4d882d2fbc6c7f671758e2
+
+ENV NB_PREFIX /
+
+RUN apt-get update && \
+    DEBIAN_FRONTEND=noninteractive apt-get install -y wget tzdata locales curl sudo && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+    # cd /tmp && \
+    # wget https://github.com/just-containers/s6-overlay/releases/download/v$S6_OVERLAY_VERSION/s6-overlay-amd64.tar.gz && \
+    # echo "$S6_OVERLAY_MD5HASH *s6-overlay-amd64.tar.gz" | md5sum -c - && \
+    # tar xzf s6-overlay-amd64.tar.gz -C / --exclude='./bin' && \
+    # tar xzf s6-overlay-amd64.tar.gz -C /usr ./bin && \
+    # rm s6-overlay-amd64.tar.gz
+
+# set version for s6 overlay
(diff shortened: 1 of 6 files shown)


StefanoFioravanzo:
@DavidSpek This looks awesome! Just an initial comment for now regarding where we should place this in the repository. Since you opened both this PR and #5570 , we should first discuss if it makes sense to create a new folder under `components/` that would be maintained by the Notebooks WG, or if we should place these somewhere else.

I don't think we should have multiple images, which are supposed to be consumed by JWA, all under `components/`. @kimwnasptd WDYT?

DavidSpek:
@StefanoFioravanzo Along with that, I'd like to bring up the tensorflow images. Do we really want/need all those versions, or can it just be the latest 1.x and 2.x? Seeing as the non latest versions have serious security vulnerabilities. Having only 2 versions of tensorflow, and then RStudio and VS-Code would mean that a simpler setup of the Dockerfiles and method of building the images can be used.

StefanoFioravanzo:
@DavidSpek I completely agree

DavidSpek:
@StefanoFioravanzo I created https://github.com/kubeflow/kubeflow/issues/5575 to track reworking the tensorflow notebooks.

DavidSpek:
/hold

Some new changes have been implemented in https://github.com/kubeflow/kubeflow/pull/5582/files

PR #5574, opened 2021-02-01
Add initial owners for testing dir

kimwnasptd:
Create an initial OWNERS file for the `testing` dir. Adding myself and @PatrickXYS for now since we are actively working on it for now. 

/cc @Bobgy 
/assign @kimwnasptd

### DIFF
--- a/py/OWNERS
+++ b/py/OWNERS
@@ -1,0 +1,3 @@
+approvers:
+- kimwnasptd
+- PatrickXYS


kimwnasptd:
I think I've made a mistake in this PR. My end goal is to add python code which would generate Argo Workflows for the new Optional-test infra CI. Initially I thought this code should go in the `/testing` dir of kubeflow/kubeflow. But after looking at it more I think my code should live under the `/py/kubeflow/kubeflow/ci` folder. 

@Bobgy I changed the file of the PR, could you take a look now? Sorry for the back and forth.
cc @PatrickXYS

kimwnasptd:
@Bobgy could you take a look at this PR? 

We will need the permissions to move forward with setting up the CI in #5577

Bobgy:
Any strong reason you'd like to mix it with existing code? If you are building sth for yourself, did you consider your own repo or starting a new folder here?

kimwnasptd:
@Bobgy let me provide some context, because we are not trying to add anything new here. We (Notebooks WG leads) are in the process of setting up our CI for running tests on our web apps, controllers, notebooks, etc. Traditionally, this repo was using the `/py/kubeflow/kubeflow/ci` folder to include the code for the CI, so we thought it would make sense to follow exactly what Google was doing, and add any future code there as well. This is why we need permissions, so we can merge these PRs.

This is one PR for example, which adds python files for the CI in the `/py/kubeflow/kubeflow/ci` folder:
https://github.com/kubeflow/kubeflow/pull/5577

Bobgy:
@kimwnasptd Yes, I understand the tradition, but since now we encourage more separation of the test-infras.
GCP test infra and AWS test infra will likely to be operated on their own, so I think it's better we just use separate folders to avoid potentially breaking each other's test infra.

Bobgy:
Hold on, it's my mistake, this is in kubeflow/kubeflow repo.

/lgtm
/approve

Yes, you can be an owner

PR #5576, opened 2021-02-01
Add PyTorch image

DavidSpek:
In spirit of adding experimental R-Studio and VS-Code images, Kubeflow really needs a PyTorch image. So here it is. It uses the latest pytorch 1.7.1 with cuda 11.0 and cudnn8 as a base. Image size is 8.83GB, could probably be optimized further but the pytorch base image is already 5.55GB. The `requirements.txt` doesn't pin any versions other tha kfp at the moment, as I am a bit lazy and Renovate will pin these for us once setup (we should strive to use the latest versions of software anyway). 
The default shell is set to `/bin/sh` for some reason, might be good to change that to `/bin/bash`. I believe the terminal functionality has expanded a bit in JupyterLab 3. 


DavidSpek:
Conda permission are messed up atm. Looking into the best way to fix this. One option is to add permission fix script, another option is to change the base image to the jupyter stacks and install cuda and pytorch on that.

blairdrummond:
FYI You might want to add this flag to the jupyter command

https://github.com/StatCan/kubeflow-containers/pull/203

It was recently introduced, but it's now needed in order to use the `/metrics` endpoint (unless jupyter changes something upstream)

DavidSpek:
@blairdrummond Thanks for catching that. For others, the change is to add `--NotebookApp.authenticate_prometheus=False` to the `CMD` command.

DavidSpek:
/hold

newer version of pytorch images are in https://github.com/kubeflow/kubeflow/pull/5582

PR #5577, opened 2021-02-01
Add initial python files for CI

kimwnasptd:
A first Argo Worflow generator script which can be used as a Prow presubmit Job. This is a first step for #5482.
This workflow runs two basic tests on the common frontend code:
1. it runs it's unittests #5564 
2. checks if the npm module, from the common code, can be build

@PatrickXYS since the [run_e2e_workflow.py](https://github.com/kubeflow/testing/blob/master/py/kubeflow/testing/run_e2e_workflow.py#L232) imports the pythonpaths from the workflow jobs, defined in the `prow_config.yaml`, I made all the imports in my code to use packages like `kubeflow.testing` and `kubeflow.kubeflow.ci` [and no relative imports].

I also defined a file with common code, since some of this logic will be needed for future workflows. Again I think we shouldn't have a problem with imports but please take a look to be sure.

I've created two helper files, a Makefile and a `_runner.py` for checking the generated Argo Workflow. The code is also conditional and can completely omit any secrets in the Argo Workflow. This is useful for testing the produced workflow in a local cluster with argo.

Lastly, please also take a look at the credentials-secrets mounts/volumes as well as to the name of the NFS volume. I'm expecting this NFS volume will also be in the Optional-test infra. If not let's discuss what changes we'll need to make

/cc @PatrickXYS 
/assign @kimwnasptd 

note that we will need to wait for #5574 to be merged first, but we can run the review process in parallel

### DIFF
--- a/py/kubeflow/kubeflow/ci/workflow_utils.py
+++ b/py/kubeflow/kubeflow/ci/workflow_utils.py
@@ -1,0 +1,249 @@
+import logging
+import os
+
+from kubeflow.testing import argo_build_util
+
+# The name of the NFS volume claim to use for test files.
+NFS_VOLUME_CLAIM = "nfs-external"
+# The name to use for the volume to use to contain test data
+DATA_VOLUME = "kubeflow-test-volume"
+
+E2E_DAG_NAME = "e2e"
+EXIT_DAG_NAME = "exit-handler"
+
+
+LOCAL_TESTING = os.getenv("LOCAL_TESTING", "False")
+CREDENTIALS_VOLUME = {"name": "aws-secret",
+                      "secret": {"secretName": "aws-secret"}}
+CREDENTIALS_MOUNT = {"mountPath": "/root/.aws/",
+                     "name": "aws-secret"}
+
+AWS_WORKER_IMAGE = ("527798164940.dkr.ecr.us-west-2.amazonaws.com/"
+                    "aws-kubeflow-ci/test-worker:latest")
+PUBLIC_WORKER_IMAGE = "gcr.io/kubeflow-ci/test-worker:latest"
+
+
+class ArgoTestBuilder:
+    def __init__(self, name=None, namespace=None, bucket=None,
+                 test_target_name=None,
+                 **kwargs):
+        self.name = name
+        self.namespace = namespace
+        self.bucket = bucket
+        self.template_label = "argo_test"
+        self.test_target_name = test_target_name
+        self.mkdir_task_name = "make-artifacts-dir"
+
+        # *********************************************************************
+        #
+        # Define directory locations
+        #
+        # *********************************************************************
+
+        # mount_path is the directory where the volume to store the test data
+        # should be mounted.
+        self.mount_path = "/mnt/" + "test-data-volume"
+        # test_dir is the root directory for all data for a particular test
+        # run.
+        self.test_dir = self.mount_path + "/" + self.name
+        # output_dir is the directory to sync to GCS to contain the output for
+        # this job.
+        self.output_dir = self.test_dir + "/output"
+
+        self.artifacts_dir = "%s/artifacts/junit_%s" % (self.output_dir, name)
+
+        # source directory where all repos should be checked out
+        self.src_root_dir = "%s/src" % self.test_dir
+        # The directory containing the kubeflow/kubeflow repo
+        self.src_dir = "%s/kubeflow/kubeflow" % self.src_root_dir
+
+        # Root of testing repo.
+        self.testing_src_dir = os.path.join(self.src_root_dir,
+                                            "kubeflow/testing")
+
+        # Top level directories for python code
+        self.kubeflow_py = self.src_dir
+
+        # The directory within the kubeflow_testing submodule containing
+        # py scripts to use.
+        self.kubeflow_testing_py = "%s/kubeflow/testing/py" % self.src_root_dir
+
+        self.go_path = self.test_dir
+
+    def _build_workflow(self):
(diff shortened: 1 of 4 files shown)


PatrickXYS:
/cc @kimwnasptd 

To make CI work on the optional-test-infra, you also need to 

Configure Presubmit on Optional-Test-Infra Prow Cluster, check [config.yaml](https://github.com/kubeflow/testing/blob/master/aws/User/clusters/kubeflow-shared-test-infra-poc/namespaces/prow/configmap/config.yaml#L119-L131)

Ref: https://github.com/kubeflow/testing/blob/master/aws/Doc/onboarding.md

kimwnasptd:
@PatrickXYS in which order do you think we should the PRs?

I was thinking the following one:
1. Create the PR for the `py_func` that will generate the Argo Workflow
2. Extend the `prow_config.yaml` in `kubeflow/kubeflow` with a presubmit job which will use the `py_func` from the previous step
3. Extend the `config.yaml` as you mentioned.

Do you think we should do them in a different order?

PatrickXYS:
> @PatrickXYS in which order do you think we should the PRs?
> 
> I was thinking the following one:
> 
> 1. Create the PR for the `py_func` that will generate the Argo Workflow
> 2. Extend the `prow_config.yaml` in `kubeflow/kubeflow` with a presubmit job which will use the `py_func` from the previous step
> 3. Extend the `config.yaml` as you mentioned.
> 
> Do you think we should do them in a different order?

It's good to go with the order you proposed

py/kubeflow/kubeflow/ci/workflow_utils.py
```diff
@@ -0,0 +1,249 @@
...
+                        "workflow_template": self.template_label,
+                    },
+                    argo_build_util.get_prow_labels()
+                ]),
+            },
+            "spec": {
+                "entrypoint": E2E_DAG_NAME,
+                # Have argo garbage collect old workflows otherwise we overload
+                # the API server.
+                "ttlSecondsAfterFinished": 7 * 24 * 60 * 60,
```
PatrickXYS, line 102: In new argo cluster, we installed v2.12.3 argo workflow, this parameter has been deprecated. 

https://github.com/argoproj/argo/blob/master/examples/gc-ttl.yaml

You can instead use ttlStrategy

py/kubeflow/kubeflow/ci/workflow_utils.py
```diff
@@ -0,0 +1,249 @@
...
+EXIT_DAG_NAME = "exit-handler"
+
+
+LOCAL_TESTING = os.getenv("LOCAL_TESTING", "False")
+CREDENTIALS_VOLUME = {"name": "aws-secret",
+                      "secret": {"secretName": "aws-secret"}}
+CREDENTIALS_MOUNT = {"mountPath": "/root/.aws/",
+                     "name": "aws-secret"}
+
+AWS_WORKER_IMAGE = ("527798164940.dkr.ecr.us-west-2.amazonaws.com/"
```
PatrickXYS, line 21: We now migrated to new test-infra, so the worker image should be `public.ecr.aws/j1r0q0g6/kubeflow-testing:latest`

py/kubeflow/kubeflow/ci/workflow_utils.py
```diff
@@ -0,0 +1,249 @@
...
+                    "aws-kubeflow-ci/test-worker:latest")
+PUBLIC_WORKER_IMAGE = "gcr.io/kubeflow-ci/test-worker:latest"
```
PatrickXYS, line 23: You don't need this anymore I think, you can simply use our public ECR test worker image

py/kubeflow/kubeflow/ci/workflow_utils.py
PatrickXYS: Or make it simpler, you don't need to set it up, we have already configured default workflow spec 

```
configmap -o yaml
apiVersion: v1
data:
  workflowDefaults: |
    metadata:
      annotations:
        argo: workflows
    spec:
      ttlStrategy:
        secondsAfterSuccess: 604800
        secondsAfterCompletion: 604800
        secondsAfterFailure: 604800
      activeDeadlineSeconds: 86400
...
```

PatrickXYS (commented):
Left some comments, just some changes related to new test-infra migration. That's because there are something different between old and new test-infra.

E.g, argo workflow upgrade

py/kubeflow/kubeflow/ci/workflow_utils.py
kimwnasptd: ACK, will use the public version only then

py/kubeflow/kubeflow/ci/workflow_utils.py
kimwnasptd: ACK, I'll use the public one

py/kubeflow/kubeflow/ci/workflow_utils.py
kimwnasptd: OK, I'll just omit this field entirely. There no issue by completely omitting the field and the configmap's value will be used right?

kimwnasptd:
Thanks for your comments @PatrickXYS! I've added a commit to address your points

py/kubeflow/kubeflow/ci/workflow_utils.py
PatrickXYS: Yeah default workflow should handle it.

PatrickXYS (commented):
/lgtm
/approve

kimwnasptd:
@PatrickXYS please write the approve command one more time now that you have the necessary permissions

PR #5585, opened 2021-02-03
Add redirection links to kubeflow app repo

PatrickXYS:
**What issue it solves**
Help https://github.com/kubeflow/community/issues/476

**What change it makes**
Add redirection links to kubeflow application repos which are sponsored by WG. 

When users create an issue in kubeflow/kubeflow, they will be able to be redirected to disired repo and create issue there.

/cc @Bobgy @kubeflow/wg-notebooks-leads

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,30 +1,30 @@
 blank_issues_enabled: true
 contact_links:
-  - name: Issue related to pipelines
+  - name: Issues related to Pipelines
     url: https://github.com/kubeflow/pipelines/issues/new/choose
     about: Create issue in kubeflow/pipelines repo
-  - name: Issue related to notebooks
+  - name: Issues related to Notebooks
     url: https://github.com/kubeflow/kubeflow/issues/new
     about: Creat issue in kubeflow/kubeflow repo
-  - name: Issue related to katib
+  - name: Issues related to Katib
     url: https://github.com/kubeflow/katib/issues/new/choose
     about: Create issue in kubeflow/katib repo
-  - name: Issue related to kfserving
+  - name: Issues related to Kfserving
     url: https://github.com/kubeflow/kfserving/issues/new/choose
     about: Create issue in kubeflow/kfserving repo
-  - name: Issue related to tf-operator
+  - name: Issues related to Tf-operator
     url: https://github.com/kubeflow/tf-operator/issues/new/choose
     about: Create issue in kubeflow/tf-operator repo
-  - name: Issue related to pytorch-operator
+  - name: Issues related to Pytorch-operator
     url: https://github.com/kubeflow/pytorch-operator/issues/new/choose
     about: Create issue in kubeflow/pytorch-operator repo
-  - name: Issue related to mpi-operator
+  - name: Issues related to Mpi-operator
     url: https://github.com/kubeflow/mpi-operator/issues/new/choose
     about: Create issue in kubeflow/mpi-operator repo
-  - name: Issue related to xgboost-operator
+  - name: Issues related to Xgboost-operator
     url: https://github.com/kubeflow/xgboost-operator/issues/new/choose
     about: Create issue in kubeflow/xgboost-operator repo
-  - name: Issue related to mxnet-operator
+  - name: Issues related to Mxnet-operator
     url: https://github.com/kubeflow/mxnet-operator/issues/new/choose
     about: Create issue in kubeflow/mxnet-operator repo
 
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,36 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Issue related to manifests
+    url: https://github.com/kubeflow/manifests/issues/new
+    about: Create issue in kubeflow/manfiests repo
+  - name: Issue related to deployment
+    url: https://github.com/kubeflow/kfctl/issues/new
+    about: Create issue in kubeflow/kfctl repo
+  - name: Issue related to pipelines
+    url: https://github.com/kubeflow/pipelines/issues/new/choose
+    about: Create issue in kubeflow/pipelines repo
+  - name: Issue related to notebooks
+    url: https://github.com/kubeflow/kubeflow/issues/new/choose
+    about: Creat issue in kubeflow/kubeflow repo
+  - name: Issue related to katib
+    url: https://github.com/kubeflow/katib/issues/new/choose
+    about: Create issue in kubeflow/katib repo
+  - name: Issue related to kfserving
+    url: https://github.com/kubeflow/kfserving/issues/new/choose
+    about: Create issue in kubeflow/kfserving repo
+  - name: Issue related to tf-operator
+    url: https://github.com/kubeflow/tf-operator/issues/new
+    about: Create issue in kubeflow/tf-operator repo
+  - name: Issue related to pytorch-operator
+    url: https://github.com/kubeflow/pytorch-operator/issues/new
+    about: Create issue in kubeflow/pytorch-operator repo
+  - name: Issue related to mpi-operator
+    url: https://github.com/kubeflow/mpi-operator/issues/new
+    about: Create issue in kubeflow/mpi-operator repo
+  - name: Issue related to xgboost-operator
+    url: https://github.com/kubeflow/xgboost-operator/issues/new
+    about: Create issue in kubeflow/xgboost-operator repo
+  - name: Issue related to mxnet-operator
+    url: https://github.com/kubeflow/mxnet-operator/issues/new
+    about: Create issue in kubeflow/mxnet-operator repo
+
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,16 +1,10 @@
 blank_issues_enabled: true
 contact_links:
-  - name: Issue related to manifests
-    url: https://github.com/kubeflow/manifests/issues/new
-    about: Create issue in kubeflow/manfiests repo
-  - name: Issue related to deployment
-    url: https://github.com/kubeflow/kfctl/issues/new
-    about: Create issue in kubeflow/kfctl repo
   - name: Issue related to pipelines
     url: https://github.com/kubeflow/pipelines/issues/new/choose
     about: Create issue in kubeflow/pipelines repo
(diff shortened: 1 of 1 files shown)


Bobgy:
Thank you! This looks awesome!
/cc @kubeflow/wg-automl-leads @kubeflow/wg-deployment-leads @kubeflow/wg-pipeline-leads @kubeflow/wg-manifests-leads @kubeflow/wg-training-leads @kubeflow/wg-serving-leads

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,36 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Issue related to manifests
+    url: https://github.com/kubeflow/manifests/issues/new
```
Bobgy, line 4: I think we can use `issues/new/choose` for all repos, because it auto redirects to `issues/new` if no templates are setup. When repos decide to set up some templates, they'll get what they want immediately.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,36 @@
...
+    about: Create issue in kubeflow/manfiests repo
+  - name: Issue related to deployment
```
Bobgy, line 6: nit: `Issues related to kfctl`?

I think that makes users easier to identify the place to go.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,36 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Issue related to manifests
```
Bobgy, line 3: Discussion: do we really need this?

I think when people come and want to create an issue, the corresponding group that should handle it should be:
* application owners if it's application specific (including manifest issues)
* distribution owners if it's distribution specific

I don't think there's a case they need to go to manifests WG.

Bobgy:
Another place that would benefit from this a lot is kubeflow/website.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,36 @@
...
+    url: https://github.com/kubeflow/manifests/issues/new
+    about: Create issue in kubeflow/manfiests repo
+  - name: Issue related to deployment
+    url: https://github.com/kubeflow/kfctl/issues/new
+    about: Create issue in kubeflow/kfctl repo
+  - name: Issue related to pipelines
+    url: https://github.com/kubeflow/pipelines/issues/new/choose
+    about: Create issue in kubeflow/pipelines repo
+  - name: Issue related to notebooks
+    url: https://github.com/kubeflow/kubeflow/issues/new/choose
```
Bobgy, line 13: Will this be a self-link? I guess you need to remove the `/choose`.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,36 @@
...
+    about: Creat issue in kubeflow/kubeflow repo
+  - name: Issue related to katib
+    url: https://github.com/kubeflow/katib/issues/new/choose
+    about: Create issue in kubeflow/katib repo
+  - name: Issue related to kfserving
+    url: https://github.com/kubeflow/kfserving/issues/new/choose
+    about: Create issue in kubeflow/kfserving repo
+  - name: Issue related to tf-operator
```
terrytangyuan, line 21: I feel like there are too many repos and users who want to file an issue would get lost. Some alternatives (just my $0.02):
1. Categorize the redirect links by working groups.
2. Encourage users to still file issues here but ask them to add labels that are relevant to each working group.

PatrickXYS: > Categorize the redirect links by working groups.

This might be a good idea, what if one WG maintain multiple repos, e.g, training WG, it might doesn't make sense if user has an issue related to pytorch-operator, but sent in tf-operator / common repo.

.github/ISSUE_TEMPLATE/config.yml
PatrickXYS: good idea

.github/ISSUE_TEMPLATE/config.yml
terrytangyuan: Yeah that would be an issue. Perhaps we can use kubeflow/common to track all training operator related issues. Or follow the second suggestion I had above. BTW, is it possible to do multi-stage selections in the issue template?

Bobgy: Yes, multi-stage sounds like a good idea.
Maybe, each WG can decide on a main repo, and add additional redirects using the same approach in the main repo?

I don't like the label approach, because it's hard to configure notification in a proper way.

PatrickXYS: I digged a bit and don't think there's multi-stage in the issue template.  

Not sure what's the best way for a WG who maintains multiple repo, but as of now, currently we only have training-WG maintainning multiple repos, so I think this should be fine without multistage.

.github/ISSUE_TEMPLATE/config.yml
PatrickXYS: Sounds good

.github/ISSUE_TEMPLATE/config.yml
PatrickXYS: Agree

.github/ISSUE_TEMPLATE/config.yml
Bobgy: In fact, I think training WG can maintain a readme.md page somewhere, maybe in GitHub, maybe in kubeflow.org.

The page should describe where issues should be reported. Then here in this repo, we'll just need to provide a link to that page.


For discussion, I think we can go with current PR first if it takes some time to set up that page.

terrytangyuan (approved):
/lgtm

Bobgy:
/lgtm
@PatrickXYS Do you want to remove the WIP in the title?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,30 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Issue related to pipelines
```
Bobgy, line 3:
```suggestion
  - name: Issues related to Kubeflow Pipelines
```

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,30 @@
...
+    url: https://github.com/kubeflow/pipelines/issues/new/choose
+    about: Create issue in kubeflow/pipelines repo
+  - name: Issue related to notebooks
```
Bobgy, line 6:
```suggestion
  - name: Issues related to Notebooks
```

Bobgy (commented):
Just to be exact, I think we need to match project names exactly as their brandings in https://www.kubeflow.org/docs/components/.

Bobgy:
/cc @joeliedtke 
for review

.github/ISSUE_TEMPLATE/config.yml
PatrickXYS: Should we add `Kubeflow` as prefix of components? I think this might be a little bit redundant. 

How about 

```
name: Issues related to Pipelines
```

Bobgy:
Thank you!

Regarding "Kubeflow Pipelines" versus "Pipelines", I'd keep the full name, because it's now an exact phrase referring to the project.

For the same reason, documentation page shows the full name now.

I think other WGs can decide their canonical brand names too.

PatrickXYS:
@Bobgy  That makes sense, updated the PR with official branding names

thesuperzapper:
Perhaps we should make a seperate website (under kubefow.org) to point users to the right place for their issue/question/request.

We can then put the link to this website in all the issue templates for all the repos.

Bobgy:
Shall we move forward?
I think @thesuperzapper 's idea is a great next step.

PatrickXYS:
I think another website would be a supplyment to current solution, because as what I saw today, when user want to file an issue, normally they won't go to another website and find entry-point from there.

User "might" (I could be wrong since I'm sharing my personal guess) go ahead to github repo like kubeflow/kubeflow, and then send issue here.

thesuperzapper:
Thinking more about the idea for a https://issues.kubeflow.org web-app:
- I think the right way to go is labels, and a single main repo (`kubeflow/kubeflow`) which stores issues
- because lots of issues are not specific to any one part of kubeflow, and need visibility across the Working Groups

The question is, how does the idea for the web app enable users to find/apply the right labels to issues they raise on `kubeflow/kubeflow`?
- perhaps a bot could auto close issues if they are un-labeled for over 30min
- we could have labels automatically applied from the web-app, based on some flow chart (or text analysis if we get them to write the issue directly on the web-app)
- the user might OAUTH with GitHub on the web-app so we can raise issues on their behalf

Obviously users could still manually add labels in the case they know what they need, or the web-app is down.

Thoughts?

PatrickXYS:
> Thinking more about the idea for a https://issues.kubeflow.org web-app:
> 
> * I think the right way to go is labels, and a single main repo (`kubeflow/kubeflow`) which stores issues
> * because lots of issues are not specific to any one part of kubeflow, and need visibility across the Working Groups
> 
> The question is, how does the idea for the web app enable users to find/apply the right labels to issues they raise on `kubeflow/kubeflow`?
> 
> * perhaps a bot could auto close issues if they are un-labeled for over 30min
> * we could have labels automatically applied from the web-app, based on some flow chart (or text analysis if we get them to write the issue directly on the web-app)
> * the user might OAUTH with GitHub on the web-app so we can raise issues on their behalf
> 
> Obviously users could still manually add labels in the case they know what they need, or the web-app is down.
> 
> Thoughts?

Well said, how about putting those words to be considered as one of long-term solution?  Maybe put in the issue instead https://github.com/kubeflow/community/issues/476

In short term, I think this PR will handle most (>=50%) of the issues, we can monitor how it goes in the near future, if we still find we need more automated way to handle massive number of issues, let's move forward with long-term solution.

thesuperzapper:
I made the comment: https://github.com/kubeflow/community/issues/476#issuecomment-777860362

I'm not against doing this PR, but I do think we should consider going the "label approach", rather than fragmenting the issues across many different repos, (and this PR will further fragment the issue landscape).

There is also a good chance that there are too many new issue templates in this PR, and it will overwhelm users.

Bobgy:
I worry the label approach needs too much tooling to make it efficient, e.g. notification.

Agree with Patrick, we can get a lot cheaply with this PR. Let's see how it works and improve

PR #5606, opened 2021-02-13
Fix entrypoint of crud-jwa

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5552

More specifically, this addresses the absence of `make` in the buster-slim version of the images. As described [here](https://github.com/kubeflow/kubeflow/issues/5552#issuecomment-769792232). 

/cc @kimwnasptd

### DIFF
--- a/components/crud-web-apps/jupyter/Dockerfile
+++ b/components/crud-web-apps/jupyter/Dockerfile
@@ -47,4 +47,4 @@ COPY ./jupyter/backend/entrypoint.py .
 COPY --from=frontend /src/dist/default/ /src/apps/default/static/
 COPY --from=frontend /src/dist/rok/ /src/apps/rok/static/
 
-ENTRYPOINT ["/bin/bash","-c","APP_PREFIX=/jupyter gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app"]
+ENTRYPOINT ["/bin/bash","-c","gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app"]
--- a/components/crud-web-apps/jupyter/Dockerfile
+++ b/components/crud-web-apps/jupyter/Dockerfile
@@ -43,9 +43,8 @@ RUN pip3 install -r requirements.txt
 
 COPY ./jupyter/backend/apps/ ./apps
 COPY ./jupyter/backend/entrypoint.py .
-COPY ./jupyter/backend/Makefile .
 
 COPY --from=frontend /src/dist/default/ /src/apps/default/static/
 COPY --from=frontend /src/dist/rok/ /src/apps/rok/static/
 
-ENTRYPOINT make run
+ENTRYPOINT ["/bin/bash","-c","APP_PREFIX=/jupyter gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app"]


thesuperzapper:
@kimwnasptd should we just merge this so that the docker image is valid (we can revisit what do do with the mostly unneeded Make file once we automate image building)

components/crud-web-apps/jupyter/Dockerfile
```diff
@@ -43,9 +43,8 @@ RUN pip3 install -r requirements.txt
 
 COPY ./jupyter/backend/apps/ ./apps
 COPY ./jupyter/backend/entrypoint.py .
-COPY ./jupyter/backend/Makefile .
 
 COPY --from=frontend /src/dist/default/ /src/apps/default/static/
 COPY --from=frontend /src/dist/rok/ /src/apps/rok/static/
 
-ENTRYPOINT make run
+ENTRYPOINT ["/bin/bash","-c","APP_PREFIX=/jupyter gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app"]
```
kimwnasptd, line 50: Please remove the `/bin/bash -c` prefix and don't set the `APP_PREFIX` ENV var

kimwnasptd (commented):
Left some small comments

kimwnasptd:
@thesuperzapper 
> (we can revisit what do do with the mostly unneeded Make file once we automate image building)

I don't think this sentence holds. The Makefile is not only intended for the CD part, it's also to show how one can run the app and create a container image for it

components/crud-web-apps/jupyter/Dockerfile
DavidSpek: @kimwnasptd Do I need to test if this works, or have you already used it with this change?

DavidSpek:
@kimwnasptd I have made the suggested changes.

components/crud-web-apps/jupyter/Dockerfile
DavidSpek: @kimwnasptd removing the "/bin/bash" seems to break things as well as it doesn't know what to execute.

DavidSpek:
@kimwnasptd I have reverted the change you suggested. Removing the bash statement breaks the container, and setting the APP_PREFIX is also done in the Makefile so I'm not sure why that should suddenly be removed.

components/crud-web-apps/jupyter/Dockerfile
kimwnasptd: I'm not sure I understand what you mean.
What is the entrypoint you tried?

DavidSpek: I tried what you suggested (on a different code base, but derived from this app) and got command not found in container startup. 

What I tried:
ENTRYPOINT ["gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app"]

kimwnasptd: Ah I see what's going on. I had in mind the _shell_ form of the entrypoint
```
ENTRYPOINT gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app
```
which is the exact same thing as using the _exec_ form starting with `["/bin/bash", "-c", ...]`

It's OK to leave the list as you have it to make it explicit we use the shell form. Could you just remove the ENV var? I prefer if we set it in the manifests instead of providing a value for it inside the image

DavidSpek: Makes sense. I pushed the change with the APP_PREFIX removed.

kimwnasptd:
The changes look good, thanks David!

/lgtm
/approve

DavidSpek:
@kimwnasptd we need to not forget to add `APP_PREFIX` to the new manifests once the PR has been merged.

DavidSpek:
@kimwnasptd I still see some GCP stuff in the presubmit test, is that what is causing it to fail or is it because the Dockerfile as it is in this PR will not work with the current manifests?

kimwnasptd:
> @kimwnasptd we need to not forget to add APP_PREFIX to the new manifests once the PR has been merged.

Very good point! Could you add a comment to the open PR that introduces the manifests of JWA and link to the comment in this issue as well?

> @kimwnasptd I still see some GCP stuff in the presubmit test, is that what is causing it to fail or is it because the Dockerfile as it is in this PR will not work with the current manifests?

@DavidSpek could you rebase on top of the latest master and force push? This should include the latest changes to the `prow_config.yaml` we did, that removed this GCP test

DavidSpek:
@kimwnasptd can you `/lgtm` this so it can be merged?

kimwnasptd:
/lgtm
we should be good to go now

PR #5610, opened 2021-02-15
(JWA): Allow setting container port

DavidSpek:
This PR adds the ability to set the container port for the "Notebook" server, which is an important step towards providing support for more generic web-apps. This PR relies on https://github.com/kubeflow/kubeflow/pull/5606. 
Together with https://github.com/kubeflow/kubeflow/pull/5601, it should be possible to launch any web-app, provided that app supports being hosted at a subfolder URI. For things like jupyter or R-Studio, the `NB_PREFIX` and `www-root-path` must still be set. For code-server, the official image at `codercom/code-server:3.8.1` can now be used. However, the user in that image is not `jovyan` so the workspace volume will not be correct and authentication is enabled, so it might still be better to use a custom built image. 

image

`davidspek/jupyter-web-app:0.12`
/cc @kimwnasptd @thesuperzapper @StefanoFioravanzo

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -151,6 +151,14 @@ def set_notebook_memory(notebook, body, defaults):
     container["resources"]["requests"]["memory"] = memory
 
 
+def set_notebook_port(notebook, body, defaults):
+    container = notebook["spec"]["template"]["spec"]["containers"][0]
+
+    containerPort = get_form_value(body, defaults, "containerPort")
+
+    container["ports"][0]["containerPort"] = containerPort
+
+
 def set_notebook_tolerations(notebook, body, defaults):
     tolerations_group_key = get_form_value(body, defaults, "tolerationGroup")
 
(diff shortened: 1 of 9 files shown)


blairdrummond:
Maybe instead of *directly* configuring port, homedir, and pullpolicy, what if this was saved into a config with a name like "RStudio" or "VS-Code" (I think you had this idea @DavidSpek ) and then then can create custom configs using a similar form.

I don't assume that our users will know what a pullpolicy is or even what a port is, by default, using a more semantic dropdown like "jupyter, rstudio, etc", would be good, and then allow users to create custom configurations if they need

DavidSpek:
@blairdrummond The problem I had with calling something "RStudio" or "Code-Sever" (VS-Code), is that the configuration can vary greatly depending on the image that is being used. 
I am now seeing if I can change https://github.com/kubeflow/kubeflow/pull/5601 to be a checkbox under advanced as well. 

What I want to do eventually is have the port, `use-root-url` (for the Istio rewrite), and PodDefaults configurable for each image in the `spawner_ui_config.yaml`. This way administrators can customize the settings for each image in the drop down list. 
The next step, which will need some careful consideration, is to have the image list from `spawner_ui_config.yaml` transferred to each user's namespace (if it is not ReadOnly), and allow users to edit what images they have in their drop down list (or even add most recently used), along with the above mentioned config values. This way, a user can an image they use, set the port, uri-rewrite and poddefault for that image once and never need to touch it again. The problem I see with this, is that the JWA will need to be able to read and edit the image list configmap in each user's namespace securely.

DavidSpek:
bump

PR #5393, opened 2020-11-10
Make Profile Controller compatible with Istio v1.4+ (v1.7.4)

ericandrewmeadows:
With Istio v1.4 and above, `AuthorizationPolicy` was implemented to consolidate resources from `ClusterRbacConfig`, `ServiceRole`, and `ServiceRoleBinding`.  During this, there was also a discovery of a bug that shows up because the Istio CRDs are not being validated by the `istiod-istio-system` `ValidatingWebhook` - when the `UserIdHeader` is empty, the header restriction on the conditional is not allowed because empty headers are not valid.

Related issue - https://github.com/kubeflow/kubeflow/issues/5313

Istio version: v1.7.4
Kubeflow version: v1.1.0
Kubernetes version: 1.16.13-gke.403
Golang version: 1.14.3

PR #5600, opened 2021-02-12
Include workflow for ui lib unittests

kimwnasptd:
As part of setting up our CI/CD #5482 we want to extend the prow config to also run the unittests of the shared ui as a presubmit job.

/cc @PatrickXYS

PR #5617, opened 2021-02-20
profile-controller: Migrate to Istio AuthorizationPolicy

### LINKED ISSUE #5397: [istio-authz-profile-controller] Support Istio Authorization Policy in profile-controller
/kind feature

Umbrella issue: https://github.com/kubeflow/kubeflow/issues/5313

In order to use the `profile-controller` with Istio >= 1.6
- We use Istio authorization to limit access to network endpoints, like Jupyter Notebooks.
- Early Istio used Istio RBAC.
- Newer Istio deprecated Istio RBAC and moved to Istio AuthorizationPolicy.

1. Add an envvar option `ISTIO_AUTHZ_TYPE`, with possible values of `rbac`(default) and `authorizationpolicy`.
2. Implement the `authorizationpolicy` option.

This way we also ensure we don't break people that can't move to the new CRD yet.

yanniszark:
**Issues resolved by this PR:**

Closes #5397

Code based on initial PR by @ericandrewmeadows (#5393)
I'm making this PR because the owner of PR #5393 has not shown any activity for 2 weeks now.

**Description:**

The Profile Controller creates Istio RBAC objects to give access to
network resources. However, the Istio RBAC API has been deprecated since
Istio 1.6.

Migrate to the AuthorizationPolicy v1beta1 API to allow for the use of a
modern Istio version.

For reference, these are the objects created before and after this PR:
Before (Istio RBAC):

<details>

```yaml
apiVersion: rbac.istio.io/v1alpha1
kind: ServiceRole
metadata:
  annotations:
    role: admin
    user: user
  creationTimestamp: "2021-01-28T12:56:00Z"
  generation: 1
  name: ns-access-istio
  namespace: kubeflow-user
  ownerReferences:
  - apiVersion: kubeflow.org/v1
    blockOwnerDeletion: true
    controller: true
    kind: Profile
    name: kubeflow-user
    uid: 70bc30c1-8e08-4684-92aa-4e0c372e76f2
  resourceVersion: "6434"
  selfLink: /apis/rbac.istio.io/v1alpha1/namespaces/kubeflow-user/serviceroles/ns-access-istio
  uid: c376e58d-feea-4ee1-955a-0f0814db66d2
spec:
  rules:
  - services:
    - '*'

---

apiVersion: rbac.istio.io/v1alpha1
kind: ServiceRoleBinding
metadata:
  annotations:
    role: admin
    user: user
  creationTimestamp: "2021-01-28T12:56:00Z"
  generation: 1
  name: owner-binding-istio
  namespace: kubeflow-user
  ownerReferences:
  - apiVersion: kubeflow.org/v1
    blockOwnerDeletion: true
    controller: true
    kind: Profile
    name: kubeflow-user
    uid: 70bc30c1-8e08-4684-92aa-4e0c372e76f2
  resourceVersion: "6436"
  selfLink: /apis/rbac.istio.io/v1alpha1/namespaces/kubeflow-user/servicerolebindings/owner-binding-istio
  uid: c450184a-95d5-4250-902e-0198a2b8adb5
spec:
  roleRef:
    kind: ServiceRole
    name: ns-access-istio
  subjects:
  - properties:
      request.headers[kubeflow-userid]: user
```

</details>

After(Istio AuthorizationPolicy):

<details>

```yaml
kind: AuthorizationPolicy
metadata:
  annotations:
    role: admin
    user: user
  creationTimestamp: "2021-02-06T18:49:45Z"
  generation: 4
  name: ns-owner-access-istio
  namespace: kubeflow-user
  ownerReferences:
  - apiVersion: kubeflow.org/v1
    blockOwnerDeletion: true
    controller: true
    kind: Profile
    name: kubeflow-user
    uid: 70bc30c1-8e08-4684-92aa-4e0c372e76f2
  resourceVersion: "3491700"
  selfLink: /apis/security.istio.io/v1beta1/namespaces/kubeflow-user/authorizationpolicies/ns-owner-access-istio
  uid: 7fdc5818-c483-4b25-98f4-21279f1644db
spec:
  rules:
  - when:
    - key: request.headers[kubeflow-userid]
      values:
      - user
  - when:
    - key: source.namespace
      values:
      - kubeflow-user
```


PatrickXYS:
```
MountVolume.SetUp failed for volume "gcp-credentials" : object "kubeflow-test-infra"/"kubeflow-testing-credentials" not registered
```

Should be why it failed to start pod / container

DavidSpek:
@PatrickXYS Does this have to do with https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml still pointing to the ksonnet code?

kimwnasptd:
The problem is that currently the only test defined in `kubeflow/kubeflow`'s [prow_config.yaml](https://github.com/kubeflow/kubeflow/blob/0c4a738625f52092f66ec462e800ddd2837e0f6d/prow_config.yaml) is meant to run in GCP, and thus the error @PatrickXYS mentioned.

That test must be removed. I'll create a new issue for removing this workflow and I'll push a commit to our existing PR that adds unittest workflows https://github.com/kubeflow/kubeflow/pull/5600

yanniszark:
@StefanoFioravanzo @elikatsis could you take a look at this when possible?

LeonardAukea:
@yanniszark We had to set `ARG GOLANG_VERSION=1.15` in the Dockerfile du to a module dependency. Other than that, it seems to work as intended.

yanniszark:
@leonardaukea thanks for the comments! I'm not clear on some things:

> We had to set ARG GOLANG_VERSION=1.15 in the Dockerfile du to a module dependency

I agree with upgrading Go, but why did you have to do it in this case? Running `make build`, which builds using Docker, works. Does it not work for you?

> Would it not be possible to have multiple owners given the AuthorizationPolicy:

In this PR, we only aim to port the existing functionality to AuthorizationPolicy. This sounds like a new feature, so I would recommend opening an issue and explaining it in detail.

LeonardAukea:
> @LeonardAukea thanks for the comments! I'm not clear on some things:
> 
> > We had to set ARG GOLANG_VERSION=1.15 in the Dockerfile du to a module dependency
> 
> I agree with upgrading Go, but why did you have to do it in this case? Running `make build`, which builds using Docker, works. Does it not work for you?
> 
> > Would it not be possible to have multiple owners given the AuthorizationPolicy:
> 
> In this PR, we only aim to port the existing functionality to AuthorizationPolicy. This sounds like a new feature, so I would recommend opening an issue and explaining it in detail.

So the issue was that we did not use the makefile, where this PR includes an updated version. we updated the default in the Dockerfile itself to `v1.15` Since that is what the department was asking for.:  https://github.com/kubeflow/kubeflow/blob/1ed877cb73ec72413b6b273872bf5f6926c4e6dd/components/profile-controller/Dockerfile#L2

image

So even in this PR. the version of go in the `Makefile` was `1.14.3` when we pulled it, but it seems to have been updated.

yanniszark:
Thanks for clarifying @LeonardAukea! I updated the Go version since it's was easy to include, as some extra small yak-shaving. I re-tested the image so all is good and we also have a later Go version :)
Finally, feel free to create an issue with your proposal whenever you like!

zijianjoy:
Thank you so much for the change!

Question: Do we plan to upload a new version of profile-controller to gcr.io/kubeflow-images-public/profile-controller ? Currently the old profile controller with kubeflow deployment on GKE was broken, we are working on a fix, but cannot proceed without a working profile-controller image. And the latest image of profile-controller is 9 months old.

kimwnasptd:
> Do we plan to upload a new version of profile-controller to gcr.io/kubeflow-images-public/profile-controller ? Currently the old profile controller with kubeflow deployment on GKE was broken, we are working on a fix, but cannot proceed without a working profile-controller image. And the latest image of profile-controller is 9 months old.

@zijianjoy yes, we are planning to release new images for all of the components tracked from the `kubeflow/kubeflow` repo for the 1.3 release. Although the images will be pushed to AWS and not GCP since we've transitioned to the Optional-test infra #5482

PR #5582, opened 2021-02-03
Proposed notebooks refactor

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5575

In this PR I've applied the changes I proposed in the above issue. I have removed the old releaser, a startup script that was not used, outdated README info, the image building script as a new solution will need to be made, and, most importantly, the version folders which haven't been updated in a year and contain only versions with many security vulnerabilities. 

Remarks:

- While I am aware that duplication of code is not best practice, and having many essentially duplicate Dockerfiles seems like something to avoid, the reasoning behind this decision is that this allows Renovate (or other dependency management bots) to automatically create PRs for updating the images. It might be possible to avoid the need for this duplication with an advanced Renovate configuration, but before that is looked into this is the easiest solution. 
- All pip dependencies are listed in the `requirements.txt` file, again this allows Renovate (or other dependency management bots) to automatically create PRs for updating the dependencies. The versions are not pinned at the moment, but Renovate should create a PR to pin them. 
- Due to the expansion of the image selection with PyTorch, R-Studio and Code Server (VS-Code), it seemed like a good idea to have a more general `notebook-images` folder (while maybe not the best name anymore). 
- In the new `notebook-images` folder, a folder for `third-party` images is created, where I placed the `contrib` folder that contains some community examples I believe. Also, it contains a folder for AWS images which is something that was previously discussed. If this third-party folder moves forward, it might be worth considering removing or reducing the GCP, AWS and Azure specific packges from the default images. 

Update:
New images have been created according to what was discussed in the Notebooks Working Group meeting of 04-02-2021. A base image was created starting from`buster-slim`, on top of which PyTorch and TensorFlow images have been made. 
A couple notes:
- Python is installed through conda, but other packages are installed with pip. Thus users can use both pip and conda for packages. This choice was made as pip resulted in a smaller image. 
- PyTorch is installed with pip rather than conda, as the pip image was 3 GB smaller than the conda image, and that was even before I added the `--no-cache-dir` to the pip command. 

| Image      | Size | File | Packages | Dockerhub |
| ----------- | ----------- | ----------- | ----------- | ----------- |
| webide-base | 381 MB | [link](https://github.com/kubeflow/kubeflow/pull/5582/files#diff-8f96b5653ae574488a22135cc94a0a8d803bc07ec66a30f15185a042ee22a6f9) | <details><summary>APT</summary><br>nodejs, yarn</details> | `davidspek/kubeflow-ubuntu-webide-base:0.2` |
[...]

PR #5622, opened 2021-02-23
Add notebook-server-base dockerfile and CI/CD

DavidSpek:
This PR adds the Web-IDE base image from https://github.com/kubeflow/kubeflow/pull/5582 that will serve as the base image for all Jupyter, R-Studio and Code-Server (VS-Code) images downstream. More details about the downstream use of this image can be seen in the PR linked above. 

Initial image was built and pushed to test CI/CD and facilitate downstream testing:
public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/base:master-9d688f73

### DIFF
--- a/components/example-notebook-servers/base/Dockerfile
+++ b/components/example-notebook-servers/base/Dockerfile
@@ -1,18 +1,21 @@
 FROM ubuntu:20.04
 
+# common environemnt variables
 ENV NB_USER jovyan
 ENV NB_UID 1000
 ENV NB_PREFIX /
 ENV HOME /home/$NB_USER
+ENV SHELL /bin/bash
 
-# set version for s6 overlay
-ARG OVERLAY_VERSION="v2.2.0.3"
-ARG OVERLAY_ARCH="amd64"
+# args - software versions
+ARG ARCH="amd64"
+ARG S6_VERSION="v2.2.0.3"
+ARG KUBECTL_VERSION="v1.20.5"
 
-# Install -- Linux Dependencies
-ENV DEBIAN_FRONTEND noninteractive
-RUN apt-get -yq update \
-    && apt-get -yq install --no-install-recommends \
+# install - usefull linux packages
+RUN export DEBIAN_FRONTEND=noninteractive \
+ && apt-get -yq update \
+ && apt-get -yq install --no-install-recommends \
     apt-transport-https \
     bash \
     bzip2 \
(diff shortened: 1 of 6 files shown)


components/notebook-images/OWNERS
```diff
@@ -0,0 +1,9 @@
+approvers:
+  - elikatsis
+  - kimwnasptd
+  - StefanoFioravanzo
+  - thesuperzapper
+  - DavidSpek
+reviewers:
+labels:
+  - notebook-servers
```
thesuperzapper, line 9: Lets leave this out, as that label does not exist.

components/example-notebook-servers/base/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+# Set locale
+RUN echo "en_US.UTF-8 UTF-8" > /etc/locale.gen && \
+    locale-gen
+ENV LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8 \
+    LANGUAGE=en_US.UTF-8 \
+    SHELL=/bin/bash
+
+# add s6 overlay
+ADD https://github.com/just-containers/s6-overlay/releases/download/${OVERLAY_VERSION}/s6-overlay-${OVERLAY_ARCH}-installer /tmp/
```
kimwnasptd, line 43: @DavidSpek just to have it documented, why do you install s6 in the base image?

DavidSpek: S6 is the init system that starts the processes in the container. It is an alternative to `Tini`, and is used in all the upstream containers. 
Very short description of why S6 over Tini. S6 can properly handle multiple processes in a container (Tini is only really built to handle 1 process), and it allows you to easily run scripts on container start and such.

DavidSpek: Specifically this I am planning to write detailed documentation for as S6 is very powerful.

kimwnasptd: Thanks for the info!

> Specifically this I am planning to write detailed documentation for as S6 is very powerful.

That's awesome! This will also be needed in order to explain users our reasons and pain points we want to avoid by using it.

So the plan is for every following Notebook server right?

And lastly
> It is an alternative to Tini, and is used in all the upstream containers.

Do you have a short list of projects using it readily available? I believe this will also help users understand why it's useful

DavidSpek: Yes, RStudio and VS-Code (Code-Server) were already using S6, so I moved Jupyter to use that as well (first I was using tini there). So now every notebook downstream uses S6 for the init system. 

The Rocker project uses S6 for their init system. https://github.com/rocker-org/rocker-versioned2
Another popular container image for Code-Server with 10M+ pull on dockerhub is using S6 for the init system. https://hub.docker.com/r/linuxserver/code-server

I can take a bit of time later on to come up with a small list that we can actually use in the docs if you like.

kimwnasptd: It's OK, just these are more than fine for now. 
I'm good, again thanks for the info!

[ feel free to resolve this ]

DavidSpek:
@kimwnasptd @StefanoFioravanzo Could one of you maybe merge this PR so I can get started rebasing/fixing the next downsteam ones (Jupyter, RStudio, VSCode base images). The Dockerfile isn't too exiting, it install some APT packages, creates the user account, sets some variables, install `kubectl` and adds the S6 overlay.

thesuperzapper:
@googlebot I consent.

thesuperzapper:
Looks good!

/approve
/lgtm

Issue #5575, opened 2021-02-01
Tracking: Rework Notebook images


DavidSpek:
/kind feature

## Goals

- Provide sensible Jupyter and RStudio dockerfiles
- Host the container images produced by said dockerfiles on ECR

## Background

The current tensorflow-notebook-image has not been updated in over a year, this is predominantly due to there not being a clear owner until the notebook working group was formed. Furthermore, the current build process is extremely complicated, it uses over [340,000 lines of code](https://raw.githubusercontent.com/kubeflow/kubeflow/master/components/tensorflow-notebook-image/releaser/lib/v1.9.7/swagger.json) to build a single container image. Clearly this is not best practice, and sets a bad example for users to build their own images.

A long standing request from the community is to support notebooks other than Jupyter. To date, supporting RStudio has required an [embedded Apache server](https://github.com/thesuperzapper/kubeflow-images/blob/master/rstudio/Dockerfile) to rewrite URL paths as R-Studio always hosts links at the HTTP root.

## Proposal

To enable native RStudio support, we propose adding the following annotations to the Notebook CRD:
- `notebooks.kubeflow.org/http-rewrite-uri`: sets `http[0].rewrite.uri` in the Notebook’s Istio VirtualService
- `notebooks.kubeflow.org/http-headers-request-set`: sets `http[0].headers.request.set` in the Notebook’s Istio VirtualService

To help help users understand how to make a working RStudio/VSCode container image, we will provide the following new Dockerfiles (and host build versions of them them):

\* denotes default images for the spawner_ui_config.yaml

Server Type | Name | Base Image | Extends Base With | PR
-- | -- | -- | -- | --
N/A | notebook-server-base | ubuntu:20.04 | common linux utilities | https://github.com/kubeflow/kubeflow/pull/5622
RStudio | notebook-server-rstudio | notebook-server-base |   | https://github.com/kubeflow/kubeflow/pull/5624
RStudio | \*notebook-server-rstudio-tidyverse | notebook-server-rstudio | Tidyverse (from conda) | https://github.com/kubeflow/kubeflow/pull/5711
VSCode | notebook-server-codeserver | notebook-server-base |   | https://github.com/kubeflow/kubeflow/pull/5625
VSCode | \*notebook-server-codeserver-python | notebook-server-codeserver | Conda installed Python | https://github.com/kubeflow/kubeflow/pull/5700
[...]

PR #5625, opened 2021-02-23
Add notebook-server-codeserver dockerfile and CI/CD

DavidSpek:
This PR is part of the effort tracked in issue https://github.com/kubeflow/kubeflow/issues/5575 and proposed in https://github.com/kubeflow/kubeflow/pull/5582. 
Its functionality depends on https://github.com/kubeflow/kubeflow/pull/5601.
The image uses the common base image from https://github.com/kubeflow/kubeflow/pull/5622 and installs [Code-Server](https://github.com/cdr/code-server) (VS-Code) in a way that is compatible with Kubeflow. 

Initial images were built and pushed to test CI/CD and facilitate downstream testing:
public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/codeserver:master-28af7716

### DIFF
--- a/components/example-notebook-servers/codeserver/Dockerfile
+++ b/components/example-notebook-servers/codeserver/Dockerfile
@@ -1,62 +1,25 @@
-FROM public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/base:master-9d688f73
+FROM public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/base:master-4892c583
 
 USER root
 
-ENV CODE_VERSION 3.9.1
+# args - software versions
+ARG CODESERVER_VERSION=3.9.1
 
-WORKDIR /tmp
+# install - code-server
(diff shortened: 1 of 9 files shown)


haijohn:
Hi, @DavidSpek, Thanks for the great work. I test this image `davidspek/kubeflow-ubuntu-code-server-experimental:0.4`  on kubeflow 1.12 notebook server , and mounted a pvc on the pod, but the pod logs the following error:
```
[2021-02-26T02:08:58.054Z] error EACCES: permission denied, mkdir '/home/jovyan/.config'
```
I found that the original notebook image use root user and this new image use jovyan. jovyan do not has permission on the mounted pvc.

DavidSpek:
The original image shouldn't be using root, that is a bad idea for many reasons. Also, please note you will not be able to use this image without the update notebook controller (and jupyter web app).

thesuperzapper:
@googlebot I consent.

thesuperzapper:
Looks good!
/approve
/lgtm

PR #5627, opened 2021-02-23
Add notebook-server-tensorflow dockerfiles and CI/CD

DavidSpek:
This PR adds the TensorFlow CPU and CUDA Jupyter base images from #5582 that will serve as a base for users to customize by adding their own choice of packages, as well as the base for the opinionated images that contain a set of commonly used tools to get users started. More details about the downstream use of this image can be seen in the PR linked above.

Initial images were built and pushed to test CI/CD and facilitate downstream testing:
public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow:master-28af7716
public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda:master-28af7716

The CUDA image is getting OOMKilled. Will need to look for a solution.

After fixing the OOMKilled error, the Pod got killed due to the `activeDeadlineSeconds` being reached. I have now made this value configurable again and doubled it for this task. The 50 minute deadline was not reached due to the image taking this long to build, but because it takes a very long time to push the image to ECR. 
On further inspection it looks like it isn't actually the pushing of the image that is taking a long time. It might be due to the CUDA drivers crating a large amount of small files in the container. The step that is taking a long time is `Taking snapshot of full filesystem...`, and as explained [here](https://github.com/GoogleContainerTools/kaniko/issues/875#issuecomment-641486550) this might cause long build times with Kaniko as "it has to stop and stat everything in the filesystem in order tot ake a snapshot".

After increasing the `activeDeadlineSeconds` the build and push is now successful. For future reference, this took [1h5m57s](https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-nb-j-tf-5627-85f8319-3104-fa28?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-nb-j-tf-5627-85f8319-3104-fa28-2958196520) for TensorFlow with CUDA.

Strangely enough without pushing the image the build time for the CUDA image is 12m28s, so it might be pushing to ECR that is taking a long time after all. 


DavidSpek:
@PatrickXYS While trying to build the TensorFlow CUDA image I am hitting OOMKilled errors. I'm guessing this is because that image is quite large. To try and mediate this I am now testing with the pod memory limit set to `8Gi`. However, I was wondering if there is some maximum memory limit I need to keep in mind for pods in the testing infra.

PatrickXYS:
Official doc for setting requests/limits: https://argoproj.github.io/argo-workflows/cost-optimisation/#set-resources-requests-and-limits

The hard limit for memory is 16G per worker node.

DavidSpek:
Thanks for the quick response. The image isn't pushed to ECR yet, but it has made it past the point where it was previously failing and to the end of the build with 8GB for the Kaniko pod. To make it easier to override the default 4G per pod, I made a memory override parameter in the `build_task_template` in `workflow_utils.py`.

thesuperzapper:
@googlebot I consent.

thesuperzapper:
Looks good!
/approve
/lgtm

PR #5640, opened 2021-02-26
(CI/CD): Init python script for building component images with Argo and Kaniko

DavidSpek:
This is a python script I created for building the component images using Argo an Kaniko. I have not yet tested it, as I am unsure how to do this locally. It still requires some work with regards to the destinations and naming/tagging scheme, as well as how to handle the `Dockerfile.dockerignore` files needed for the Jupyter and Tensorboards Web Apps. Hopefully this can serve as a good starting point.

/cc @kimwnasptd  @PatrickXYS @kubeflow/wg-notebooks-leads


kimwnasptd:
Thanks for helping there @DavidSpek! 

Here are some first comments and questions that I believe we need to address:
* We don't want to have a single test that will be building all the images at once, but many smaller independent ones. We don't want to build the Tensorboards web app when a PR only modified the code for the Jupyter web app
* As you mentioned we need to document how will the maintainers be able to test the code. How can a reviewer review the code if they don't know how to run it?
* What names will we use for our images

For the last two I don't have yet a definitive answer, but I'm also looking into it so I'll provide more feedback in the next days.

Could you also raise these questions in the issue #5482  so we can have a first documentation of them there?

DavidSpek:
@kimwnasptd 
Regarding your first point, this makes complete sense. When I started on this I had in mind building all the images for a release, but it is better to split it up and have the image building run with each new commit to a folder. I'll split up the unit script once there is some more clarity.

DavidSpek:
Also, seeing as we will be running a Kaniko job for each image, I will likely change the Kaniko arguments to use git directly.

DavidSpek:
@kimwnasptd I have added presubmit testing files for all the components as well that do not push to the registry, but do check if the image can actually be built. As I found out this is an issue for the admission-webhook, which also has a PR to fix it https://github.com/kubeflow/kubeflow/pull/5661. 

@PatrickXYS Another thing I would like to bring up, is that the `prow_config.yaml` included in a PR **MUST NOT** be used, rather the `prow_config.yaml` from the master should be used. This needs to change **ASAP** as now anybody can create a PR with arbitrary code and have it pushed to the ECR registry as one of the components.

DavidSpek:
@kimwnasptd @PatrickXYS building the image for access-management is getting OOM killed, which is why the presubmit test is failing. This registry also still needs to be created: https://github.com/kubeflow/kubeflow/issues/5482#issuecomment-791906596.

PatrickXYS:
> This needs to change ASAP as now anybody can create a PR with arbitrary code and have it pushed to the ECR registry as one of the components.

This is not true, only kubeflow member can do it. Otherwise, new users need a `/need-to-test` tag to trigger presubmit

DavidSpek:
@PatrickXYS That's true, but it still might be better to pull the config from master to avoid such a situation.

PatrickXYS:
Given this scenario, WG folks need to be responsible to prevent it happen, because it's your responsibility to own and manage artifacts. 

One solution I can come up with is: for presubmit only push image to private ECR, for postsubmit push image to public ECR

py/kubeflow/kubeflow/cd/access_management.py
```diff
@@ -0,0 +1,44 @@
+""""Argo Workflow for building Access Management's OCI image using Kaniko"""
```
PatrickXYS, line 1: I feel like you can abstract this Builder python file given you have multiple almost same file here, add it in e.g, `py/kubeflow/kubeflow/cd/base_builder.py`

And for each Argo Workflow, just reference it should be fine

py/kubeflow/kubeflow/cd/access_management_runner.py
```diff
@@ -0,0 +1,12 @@
+# This file is only intended for development purposes
```
PatrickXYS, line 1: Same here:

I feel like you can abstract this Builder python file given you have multiple almost same file here, add it in e.g, py/kubeflow/kubeflow/cd/base_runner.py

And for each Argo Workflow, just reference it should be fine

py/kubeflow/kubeflow/ci/access_management_tests.py
```diff
@@ -0,0 +1,43 @@
+""""Argo Workflow for testing Access Management's OCI image using Kaniko"""
```
PatrickXYS, line 1: Same herer: 

I feel like you can abstract this Builder python file given you have multiple almost same file here, add it in e.g, py/kubeflow/kubeflow/ci/base_builder.py

And for each Argo Workflow, just reference it should be fine

py/kubeflow/kubeflow/ci/access_management_tests_runner.py
```diff
@@ -0,0 +1,12 @@
+# This file is only intended for development purposes
```
PatrickXYS, line 1: Same here:

I feel like you can abstract this Builder python file given you have multiple almost same file here, add it in e.g, py/kubeflow/kubeflow/ci/base_runner.py

And for each Argo Workflow, just reference it should be fine

DavidSpek:
Thanks for the review. Indeed many files that are almost the same so I will try to optimize this tomorrow. The `*_tests` files I am expecting will have more tests added to them in the future so I'm not sure what the best way would be to abstract the common sections will be. 
Regarding the presubmit and postsubmit tests, I have the presubmit test just build the image as a check but it is never pushed. This is a very simple way to test PRs for some very basic requirements.

PatrickXYS:
> Regarding the presubmit and postsubmit tests, I have the presubmit test just build the image as a check but it is never pushed. This is a very simple way to test PRs for some very basic requirements.

Yes simple in development, difficult in security. How would you prevent malicious/careless push in presubmit?

DavidSpek:
The presubmit is setup to only build the image, the destination is a string that isn't associated with a container registry and the kaniko `--no-push` flag is used. So as long as the Argo workflow isn't mistakenly changed, I think this should be fine.

PatrickXYS:
> The presubmit is setup to only build the image, the destination is a string that isn't associated with a container registry and the kaniko `--no-push` flag is used. So as long as the Argo workflow isn't mistakenly changed, I think this should be fine.

I agree, I'm just saying the worst case, someone removed `--no-push` in their PR

PatrickXYS:
But anyway that's just artifacts management, be prepared for the worst case is what I want to bring up.

DavidSpek:
For that reason, is it possible to use the prow config from the master branch instead of the one included in the PR?

PatrickXYS:
> For that reason, is it possible to use the prow config from the master branch instead of the one included in the PR?

Prow job is PR-based, you can use periodic prowjobs for desired functionality

py/kubeflow/kubeflow/ci/access_management_tests.py
DavidSpek: I'm not sure it makes sense for these CI files, as I am assuming they will be extended with more tests similar to https://github.com/kubeflow/kubeflow/blob/master/py/kubeflow/kubeflow/ci/common_ui_tests.py

PatrickXYS: For extensibility, I would suggest using parent class as base, and child class for patch.

prow_config.yaml
```diff
@@ -23,3 +92,62 @@ workflows:
...
+    include_dirs:
+      - components/centraldashboard/*
+    kwargs: {}
+  # Create and push Tensorboard Web App OCI image to ECR
+  - py_func: kubeflow.kubeflow.cd.twa.create_workflow
+    name: twa-build
+    job_types:
+      - postsubmit
+    include_dirs:
+      - components/crud-web-apps/common/frontend/kubeflow-common-lib/*
```
kimwnasptd, line 125: lets also include the directory of the common code here as well
`components/crud-web-apps/common/backend`

prow_config.yaml
```diff
@@ -23,3 +92,62 @@ workflows:
       - components/crud-web-apps/common/frontend/kubeflow-common-lib/*
```
kimwnasptd, line 92: lets also include the directory of the common code here as well
components/crud-web-apps/common/backend

prow_config.yaml
```diff
@@ -23,3 +92,62 @@ workflows:
...
       - components/crud-web-apps/jupyter/*
     kwargs: {}
+  # Create and push Access Management OCI image to ECR
+  - py_func: kubeflow.kubeflow.cd.access_management.create_workflow
+    name: a-man-build
```
kimwnasptd, line 97: lets rename this to `ac-mgr-build`

py/kubeflow/kubeflow/cd/access_management_runner.py
```diff
@@ -0,0 +1,5 @@
...
+from kubeflow.kubeflow.cd import base_runner
+
+base_runner.main(component_name="access_management",
+                 workflow_name="a-man-build")
```
kimwnasptd, line 5: lets rename this to `ac-mgr-build`

kimwnasptd (commented):
Thanks for your work @DavidSpek! I've left some small comments.

The changes LGTM overall, so once these are resolved we should be good to merge from my side.

py/kubeflow/kubeflow/ci/admission_webhook_tests_runner.py
```diff
@@ -0,0 +1,5 @@
+# This file is only intended for development purposes
+from kubeflow.kubeflow.ci import base_runner
+
+base_runner.main(component_name="admission_webhook_tests",
+                 workflow_name="ad-wh-tests")
```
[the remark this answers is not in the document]
DavidSpek: I'll need to check the length, but it might be that the 1 extra character will cause the length to be over them max 63 characters.

prow_config.yaml
DavidSpek: Same here.
I'll need to check the length, but it might be that the 1 extra character will cause the length to be over them max 63 characters.

py/kubeflow/kubeflow/ci/access_management_tests.py
DavidSpek: @kimwnasptd What is your thought on this?

kimwnasptd: @PatrickXYS to make sure I'm synced with the proposed changes; the argument is to create a common base class for building a workflow that always has a step for kaniko-build, with `--no-push`, since we essentially use it in all of the component CI files?

py/kubeflow/kubeflow/ci/admission_webhook_tests_runner.py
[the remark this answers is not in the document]
kimwnasptd: Here's the error message I had bumped into while also testing this https://github.com/kubeflow/kubeflow/pull/5653#issuecomment-788179056
```
Invalid value: \"kubeflow-kubeflow-presubmit-shared-ui-tests-5653-bd89676-2624-af49\": must be no more than 63 characters"
```

The workflow name above without the `shared-ui-tests`, which is the value we defined in that PR, has a length of 51 chars. So as long as we use at most 12 chars in our names we should be OK

prow_config.yaml
kimwnasptd: As above, IIUC we should be OK https://github.com/kubeflow/kubeflow/pull/5640#discussion_r589421496

py/kubeflow/kubeflow/ci/admission_webhook_tests_runner.py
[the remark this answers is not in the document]
DavidSpek: The situation for the runner is different than what runs on AWS, but I kept the naming consistent for the sake of sanity.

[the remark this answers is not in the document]
kimwnasptd: > I kept the naming consistent with the prow_config.yaml for the sake of sanity.

That's good, I also proposed this name to the workflow to keep the names the same everywhere so we will still be consistent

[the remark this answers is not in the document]
DavidSpek: @kimwnasptd I just pushed your suggested changes and added some empty files to test if there are any new issues regarding the name length.

py/kubeflow/kubeflow/ci/access_management_tests.py
DavidSpek: @kimwnasptd I'm also not completely clear what the proposed implementation would look like. Either way, I'm not sure if this is a blocker as it testing if images can be built seems like something that should be added sooner rather than later. This way we would have discovered that the admission-webhook was broken before https://github.com/kubeflow/kubeflow/pull/5661.
Also, https://github.com/kubeflow/kubeflow/issues/5482#issuecomment-791906596 still needs to be resolved. Should the registry be called `access-management` or should it be `kfam`?

DavidSpek:
@kimwnasptd @PatrickXYS It looks like the build for `access-management` (a.k.a. `kfam`) failed again with message `OOMKilled`. I think it has done this 3 or 4 times now and only built successfully once, so this is something that will need addressing.

py/kubeflow/kubeflow/ci/admission_webhook_tests_runner.py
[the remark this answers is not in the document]
DavidSpek: @kimwnasptd It seems like there were no problems with the test builds after changing the names, so I am resolving this. If this issue does pop up it can be quickly fixed in the future.

py/kubeflow/kubeflow/ci/access_management_tests.py
PatrickXYS: This is only NIT change, if we feel it's urgent, feel free to move forward

PatrickXYS:
> It looks like the build for access-management (a.k.a. kfam) failed again with message OOMKilled

Mostly like 1 out of 10 nodes running out of memory, next time when OOMKilled happened, please have resources quota to higher number to avoid the issue 

https://kubernetes.io/docs/concepts/policy/resource-quotas/

DavidSpek:
@PatrickXYS Will need to look into increasing the resource quota for that workflow then. Is there an easy way to set this with the current setup?

py/kubeflow/kubeflow/ci/access_management_tests.py
DavidSpek: I think this can be dealt with in the future when the tests for each component get expanded. For now, I think it is most important images start getting built and the PRs that are being merged for the release are at least tested if they break the image building.

DavidSpek:
@PatrickXYS @kimwnasptd I think this is good for being merged. The few small things that might remain can be dealt with as we encounter them. I think it is more important to have this merged so that the incoming PRs for the upcoming release are tested if they produce an image that can be built and that the images start being built.

DavidSpek:
Png @Bobgy I'm guessing you will need to approve this as there are lots of changes to the `prow_config.yaml`

Bobgy:
/approve
You can decide when to merge by LGTM

py/kubeflow/kubeflow/ci/access_management_tests.py
kimwnasptd: Regarding the changes for using a different class I agree with @DavidSpek that this isn't a blocker, so let's go with this implementation for now and if in the future we find ourselves repeating too much code then lets go over this argument.
Does this sound OK to you @PatrickXYS ?

kimwnasptd: > Should the registry be called access-management or should it be kfam?

@yanniszark what are your thoughts on this? I guess we should stick to `access-management` for now?

I'll also give a heads up to our tracking issue to add this registry as well

DavidSpek: @kimwnasptd I actually just ran into an issue where using the common `kaniko_builder.py` actually complicates things slightly. For this situation specifically it is easily fixed, however; for the CI testing I think it could just complicate things with little added value.

kimwnasptd: > @kimwnasptd I actually just ran into an issue where using the common kaniko_builder.py actually complicates things slightly. For this situation specifically it is easily fixed

Could you provide some more details for this? Just so that we can identify the pattern/pain-point once we see it reoccurring

yanniszark: >  I guess we should stick to access-management for now?
@kimwnasptd I would vote for `kfam`, as it is what we usually call the component.

thesuperzapper: Also kfam actually stands for Kubeflow Access Management (just like we don't call it kubeflow-pipelines, but kubeflow/pipelines)

DavidSpek: @yanniszark @kimwnasptd I think this discussion should move to https://github.com/kubeflow/kubeflow/issues/5675. Then this PR can be merged and I will update all the registry naming depending on the outcome of https://github.com/kubeflow/kubeflow/issues/5675.

thesuperzapper:
I think this is in a state where we can merge it (given how time pressured we are, and the general agreement),=.

Any further changes/nit-picks can be fixed in future PRs.

/lgtm

PR #5644, opened 2021-02-27
Tesorboard-Controller: use updateStatus instead of update

Garrybest:
The status of `Tensorboard` should be a subresource of the CRD, which is like [`Notebook`](https://github.com/kubeflow/kubeflow/blob/743310d977f6dbac55a0676c94ce645b0276d05f/components/notebook-controller/api/v1/notebook_types.go#L62). Therefore, when updating the instance, we use `r.Status().Update(ctx, instance)` instead of `r.Update(ctx, instance)`.

### DIFF
--- a/components/tensorboard-controller/api/v1alpha1/tensorboard_types.go
+++ b/components/tensorboard-controller/api/v1alpha1/tensorboard_types.go
@@ -50,6 +50,7 @@ type TensorboardStatus struct {
 }
 
 // +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
 
 // Tensorboard is the Schema for the tensorboards API
 type Tensorboard struct {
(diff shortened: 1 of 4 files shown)


Garrybest:
PTAL😊 @elikatsis @kimwnasptd

kimwnasptd:
Thank you for your PR @Garrybest! 

/lgtm
/approve
/hold

I'll merge this PR once we first merge our postsubmit job that will be building the controller's image #5640

DavidSpek:
@kimwnasptd I'm not sure, but I think it might be the case that this PR would require a rebase after https://github.com/kubeflow/kubeflow/pull/5640 is merged. This is the case for presubmit tests in any case.

DavidSpek:
@Garrybest Can you rebase this PR on master?

Garrybest:
> @Garrybest Thanks for the quick rebase. If the image builds successfully I will merge this PR.

@DavidSpek OK, my pleasure. :-)

DavidSpek:
@Garrybest If you are interested, you can see the workflow running here: https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-tb-c-test-5644-7674ecc-8752-bff0

Garrybest:
@DavidSpek Well, it seems something went wrong. Is the presubmit job failed because of my PR?

Issue #5641, opened 2021-02-26
(JWA): CPU in request body wrong type


DavidSpek:
/kind bug

While working on the JWA that uses the common code, I am experiencing a type error regarding CPU needing to be of type `string` rather than `number`. 

Error output:
```
[422] Notebook.kubeflow.org "new-test" is invalid: spec.template.spec.containers.resources.requests.cpu: Invalid value: "number": spec.template.spec.containers.resources.requests.cpu in body must be of type string: "number" http://127.0.0.1:5000/api/namespaces/admin/notebooks
```

This error can be solved by adding the following code to the `getSubmitNotebook()` function in `form-defaul.components.ts`. However, I'm not sure if this is an actual fix for the problem, or just a workaround. 

```
  // Ensure CPU input is a string
  if (typeof notebook.cpu === 'number') {
    notebook.cpu.toString();
  }
```
@kimwnasptd Maybe you can add some input on this.

DavidSpek:
Has been fixed

PR #5646, opened 2021-02-27
(JWA): Add server type selector for jupterlab, vs-code and r-studio

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/3362
This PR adds a selector to the notebook creation page allowing the user to select what type of server they want to start. The options are: Jupyter, VS-Code and R-Studio. When VS-Code or R-Studio are selected, the settings required to run the selected server type are added to the Notebook using the updates to the notebook-controller from https://github.com/kubeflow/kubeflow/pull/5660. These settings are also applied to custom images (if this is unwanted, leave jupyterlab selected). To go along with this, the `spawner_ui_config.yaml` has been modified to have 3 different types of images (`jupyterImage`, `vscodeImage` and `rstudioImage`). 

Server type selector:
image

When VS-Code is selected, the drop down list shows entries from the `vsCodeImage` section of the `spawner_ui_config.yaml`:
image

The same goes for R-Studio:
image

The index page has a new column that show what type of server a give notebook is:
image

/cc @kimwnasptd @StefanoFioravanzo @thesuperzapper @elikatsis @yanniszark


castrojo:
> We need to ask permission for using the logos from the relevant parties.

Adding these links for reference: 

Rstudio: https://rstudio.com/about/trademark/
Jupyter: https://jupyter.org/governance/trademarks.html

I'll chase down VSCode's usage.

DavidSpek:
Thanks for the links @castrojo. Regarding Jupyter, I think it is pretty clear from the text already:
```
All trademarks are subject to “nominative use rules” that allow use of the trademark to name the trademarked entity in a way that is minimal and does not imply a sponsorship relationship with the trademark holder.
As such, stating accurately that software or a service integrates with Jupyter software, that it uses Jupyter software, that it is compatible with Jupyter software, or that it contains Jupyter software, is always allowed. In those cases, you may use the word “Jupyter”, unaltered Jupyter logos, or images or videos of Jupyter applications to indicate this, without our prior approval. This is true both for non-commercial and commercial uses.
```

DavidSpek:
@castrojo I've just sent an email to R-Studio regarding the use of the logo.

kimwnasptd:
I'll also start reviewing this PR so that we can have the necessary infra ready for the new R-Studio and VSCode images. This can be done in parallel with our discussion around the image names #5575. I'll also try to update our [design doc](https://docs.google.com/document/d/14ve8inClzkhIILAIp2FodemRtoS2hqanX8jll2lmRUE/edit#heading=h.47k606j3mc2v) with the designs we will make here.

@DavidSpek is the PR ready for review?

DavidSpek:
@kimwnasptd Yes it's ready for review.

kimwnasptd:
Starting looking into this. I'm good with how the form looks and how the config map will be updated. Although we should redesign its fields at some point in the future but for now it's more than fine.

The first thing we should change though is how the SVGs are sent to the frontend. Let's have the frontend directly fetch them from the backend and not inline them in the config. I'm looking into how we can do this and what parts need to be changed, so I'll provide some more feedback and snippets/examples in a little bit.

Thanks for the effort @DavidSpek!

DavidSpek:
@kimwnasptd Two quick things.
1. The request sent with the notebook CRD needs a small fix to send a string with JSON for the RStudio headers. I will do this in a moment. 
2. Regarding the SVG, the reason for having them in the yaml is so that people can very easily swap them out without needing to rebuild the JWA image (this might be needed due to licensing). I wasn't able to get the frontend to grab the SVG from the yaml directly. 

Regarding the spawner_ui.yaml in general, me and Mathew were already thinking about what would be a better modular structure and started on a proposal for this. An important idea here is that an arbitrary number of image groups can be in the yaml, and the frontend picks up on this.

kimwnasptd:
> Regarding the SVG, the reason for having them in the yaml is so that people can very easily swap them out without needing to rebuild the JWA image (this might be needed due to licensing). I wasn't able to get the frontend to grab the SVG from the yaml directly.

I know, but we shouldn't include static files into the configuration yaml, as it seems a little bit dirty. The browser will also not be able to cache these svgs in case they are embedded in the config. 

Providing a URL for them instead, in the configuration yaml, will make it easy for people to select any SVG they want and the browser will also be able to cache them.

kimwnasptd:
> The request sent with the notebook CRD needs a small fix to send a string with JSON for the RStudio headers. I will do this in a moment.

Makes sense, give me a heads up when you're ready

DavidSpek:
@kimwnasptd I just fixed the JSON part so that is ready. Regarding the SVG in the yaml, it isn't very pretty, but the way it works is that the backend dumps the SVG's to a file in the `assets` where other SVGs are located if they don't exist on container startup. So I believe browser caching will work as expected. Using a URL could break existing deployments if the SVG isn't hosted at that location anymore. Grabbing the SVG from the yaml directly would break browser caching I think so that should probably be avoided then.

kimwnasptd:
Some first comments after diving a little bit deeper into the code and the commits.

Before jumping into different parts of the review I'd like to invest some time and discuss a little bit more about the structure of the commits. These are some tips to make a PR easier for the reviewer to review.

The first thing to remember is that a lot of reviewers will be reviewing the code by looking at the commits one by one and not directly from the GitHub UI, where it shows the entire list of changes. With this in mind some tips are:
1. Try to organize and group your commits based on the functionality they provide.
    * This will make it easier for the reviewer to review chunks of new functionality and not have to keep a state in mind while looking at other commits to have the bigger picture
    * There's no definitive formula for that tip and I believe every developer has room of improvement in this area. Here's an example of this improvement that @elikatsis mentioned for my code https://github.com/kubeflow/kubeflow/pull/5693#discussion_r595051280
2. Don't include commits that change the code that another commit introduced in the same PR.
    * This goes hand to hand with the previous tip. Having all the the relevant changes of a functionality grouped into a single commit drastically improves the review process. The reviewer can completely isolate specific commits of implementation.
    * Fixup or squash commits that should be squashed. Don't have commits titles that start with `Fixup ...`
4. Once the review process starts, and the reviewer submits the first review changes, then from there on submit specific commits that address the reviewer's comments.
    * This will make it easy for the reviewer to be able to quickly check your latest commits and understand if the changes you did align with the proposed changes they had in mind
    * If you rebase the commits then try to avoid squashing the review commits. Here's an example PR in which I marked the review commits with a `review: ` prefix to make it easier for you and @elikatsis to be able to glance at the commits and see their changes https://github.com/kubeflow/kubeflow/pull/5693
5. Don't name commits with titles like `small fixes` as the person looking at it cannot deduce what the changes are about. Ideally these fixes should be integrated into other commits, or be separate commits for each smaller fix

These are just some friendly tips from a quick look on the PR that I believe could drastically improve the PRs you submit and make them really easy to review. And again let me stress out that I'm not trying to neither undermine your work nor try and talk from a high horse. Every developer always has room to improve on how they use git and collaborate with each other, myself included!

These are only meant to be tips from on fellow developer to another :)

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.html
```diff
@@ -4,8 +4,23 @@
   ML packages"
   icon="fa:fab:docker"
 >
+  <section>
```
kimwnasptd, line 7: You don't need this wrapper element. The `<mat-button-toggle-group>` will suffice

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.html
```diff
@@ -4,8 +4,23 @@
...
+    <div class="server-type">Choose server type:</div>
```
kimwnasptd, line 8: I propose to remove this text and only have the buttons. It will make the form more clean

DavidSpek:
@kimwnasptd Thank you for the comments, I really appreciate them. I was actually assuming you (and others) would look at the overall code changes and not each commit separately. I was actually meaning to rebase the PR to a single commit, but hadn't gotten around to it yet. Would a single commit work for you, or should I split up the various sections that are touched in this PR into a few commits?

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.html
```diff
@@ -4,8 +4,23 @@
...
+    <mat-button-toggle-group [formControl]="parentForm.get('serverType')" aria-label="Server Type">
+      <mat-button-toggle value="jupyter" aria-label="Use JupyterLab based server">
+        <mat-icon style="height:72px !important; width:200px !important;" svgIcon="jupyterlab"></mat-icon>
```
kimwnasptd, line 11: Don't modify html elements directly with styles, always use a class instead that wraps what the style changes are. Also try to avoid as much as possible `!important` in CSS. Instead try to use more specific css selectors.

Here you could introduce and use the following CSS class for each `<mav-icon>`
```css
.server-type {
  height: 32px;
  width: 150px;
}
```

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.html
```diff
@@ -4,8 +4,23 @@
   ML packages"
   icon="fa:fab:docker"
 >
+  <section>
+    <div class="server-type">Choose server type:</div>
+    <mat-button-toggle-group [formControl]="parentForm.get('serverType')" aria-label="Server Type">
```
kimwnasptd, line 9: Add a wrapper class here to give this group a bottom margin

```css
.server-type-wrapper {
  margin-bottom: 1rem;
}

**Review by kimwnasptd — COMMENTED** [review]
Here are some first review comments around how the buttons look like. Reducing the height of the buttons will make them look more inline with the other inputs in the form

Screenshot from 2021-03-16 21-11-41

**Comment by DavidSpek** [comment]
I was also not happy with the button design as it is now. Your suggested changes look a lot better and more like what I had in mind. I'll get to adding them first thing tomorrow morning.

**Comment by kimwnasptd** [comment]
> Would a single commit work for you, or should I split up the various sections that are touched in this PR into a few commits?

I would say always go for smaller commits and not a single one. Here's also an article that has helped me and I always try to follow for my PRs https://mtlynch.io/code-review-love/

File: components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.html (hunk 1 above)
Comment (kimwnasptd) [reply components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.html:7]: Looked at it again and in conjunction with how we should place the `Custom image` checkbox.

I propose to 
1. use a wrapper `div` around the `<mat-button-toggle-group>` and the checkbox
2. that `div` will have the class `flex-column` which is defined in the common ui lib
3. move the checkbox on top of the buttons group

The end result will look like this 
Screenshot from 2021-03-16 22-01-49

**Comment by DavidSpek** [comment]
@kimwnasptd I've just pushed a commit with your suggested changes to the button toggle styling. If you could also help me with having the index show small logos rather than the server-type text that would be amazing. 
With small logo I mean:
https://raw.githubusercontent.com/jupyter/jupyter.github.io/master/assets/main-logo.svg
and similar for VSCode and RStudio

**Comment by kimwnasptd** [comment]
@DavidSpek first of all thanks for converting the commits and reorganizing them. It's a very good improvement and helps the review a lot!

> If you could also help me with having the index show small logos rather than the server-type text that would be amazing.

This is definitely doable, I'll add some review comments about it and provide some guidance and next steps

File: components/crud-web-apps/jupyter/frontend/src/app/pages/index/index-default/config.ts [hunk 5]
```diff
@@ -59,6 +59,11 @@ export const defaultConfig = {
         tooltipField: 'name',
       }),
     },
+    {
+      matHeaderCellDef: 'Type',
+      matColumnDef: 'type',
+      value: new PropertyValue({ field: 'serverType'}),
```
kimwnasptd, line 65: Instead of a `PropertyValue` object you can use a [ComponentValue](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/resource-table/types/component-value.ts) instead. In this case you can provide your own Angular component for the specific column.

That component will need to implement the [TableColumn](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/resource-table/component-value/component-value.component.ts#L5-L7) class, which only means that it should have a property `element` in which the table will be injecting the element value of the row. Inside the component you can then do anything you want and show any html [ the svg in this case ] fits your needs.

Here's an example of this 
https://github.com/kubeflow/katib/blob/master/pkg/new-ui/v1beta1/frontend/src/app/pages/experiments/config.ts#L76
https://github.com/kubeflow/katib/blob/master/pkg/new-ui/v1beta1/frontend/src/app/pages/experiments/optimal-trial/experiment-optimal-trial.component.ts

kimwnasptd (commented):
To show an SVG for the new column you will need to
1. Create a new component that will be taking as input [more on this in the comment below] a Notebook that corresponds to a row and then render the SVG
2. Use the `ComponentValue` instead in the table's config

components/crud-web-apps/jupyter/frontend/src/app/pages/index/index-default/config.ts
DavidSpek: Thanks for the instructions. At first I was looking for an example to follow in the `kubeflow/kubeflow` repo but couldn't find any. I had already thought `ComponentValue` would need to be used, but couldn't figure out how to implement this. I'll look into adding this tomorrow. I'll use the same method to dump the SVGs for VSCode and RStudio from the yaml as I am already using so administrators can customize them easily if needed (the VSCode option will likely work for most web-applications).

kimwnasptd (commented):
We briefly talked about it abouve but I'll also add it in the review. Let's not inject SVG code into the ConfigMap's yaml.

The biggest majority of the users will not need to edit these SVG files for the longest time they use the app. But if we inject these into the yaml then we will be introducing and showing a lot of boileplate code to many users that will not need to see/edit this info.

Adding a couple svgs in the image's source code and recompiling isn't a difficult task which I think is acceptable from users to do if they would like to swap things around.

components/crud-web-apps/jupyter/frontend/src/app/pages/index/index-default/config.ts
kimwnasptd: Lets not import svgs into the YAML, I explained the reasoning in the comment below https://github.com/kubeflow/kubeflow/pull/5646#pullrequestreview-614626085.

But lets include the extra links into the `environment.*.ts` files https://github.com/kubeflow/kubeflow/pull/5646#discussion_r596260211

DavidSpek:
@kimwnasptd Could we do some sort of combination of the things you describe and what is being done now? More specifically, allow the user to set the URLs for the icons in the config map. Due to the use of trademarks and the applicable guidelines I think it is import to make it very easy for people to remove the logos (preferably without needing to build their own image).

components/crud-web-apps/jupyter/backend/apps/common/form.py
```diff
@@ -135,6 +135,25 @@ def set_notebook_image_pull_policy(notebook, body, defaults):
     )
 
 
+def set_notebook_http_rewrite_uri(notebook, body, defaults):
```
kimwnasptd, line 138: We can remove this function, the backend can set the necessary annotations by looking at the server type

DavidSpek (commented):
t

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
```diff
@@ -18,29 +18,106 @@ spawnerFormDefaults:
...
+    # Uncomment the following line to only enable standard container Images
+    readOnly: false
+  imageRStudio:
+    # The container Image for the user's RStudio Server
+    # If readonly, this value must be a member of the list below
+    value: public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/rstudio:master-28af7716
+    # The list of available standard container Images
+    options:
+      - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/rstudio:master-28af7716
+    logo: |-
```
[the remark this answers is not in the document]
kimwnasptd: @DavidSpek taking your comment https://github.com/kubeflow/kubeflow/pull/5646#issuecomment-801308346 here to have everything coordinated in this place

> @kimwnasptd Could we do some sort of combination of the things you describe and what is being done now? More specifically, allow the user to set the URLs for the icons in the config map. 

This is the initial idea I had https://github.com/kubeflow/kubeflow/pull/5646#issuecomment-797521345, although thinking about it more I believe it will be more involved to get it working. In this case we will need to first fetch the yaml [ and the request could fail ] and then inject the urls and use them.

> Due to the use of trademarks and the applicable guidelines I think it is import to make it very easy for people to remove the logos (preferably without needing to build their own image).

Could you elaborate a little bit more on this? How does the trademarks and applicable guidelines fit into the picture?

I agree that we should make it easy for people to swap the logos they prefer, but these users are the minority and most probably won't have a hard time to rebuild the image.

components/crud-web-apps/jupyter/backend/apps/common/form.py
kimwnasptd, line 138: Lets remove this helper function since, for now, the backend should just add any necessary annotation based on the server type found in the request

components/crud-web-apps/jupyter/backend/apps/common/form.py
```diff
@@ -135,6 +135,25 @@ def set_notebook_image_pull_policy(notebook, body, defaults):
...
+    notebook_annotations = notebook["metadata"]["annotations"]
+    http_rewrite_uri = get_form_value(body, defaults, "httpRewriteURI")
+    if http_rewrite_uri:
+        notebook_annotations["notebooks.kubeflow.org/http-rewrite-uri"] = http_rewrite_uri
+
+
+def set_http_headers_request_set(notebook, body, defaults):
```
kimwnasptd, line 145: Same as above, lets remove this one as well

components/crud-web-apps/jupyter/backend/apps/common/form.py
```diff
@@ -135,6 +135,25 @@ def set_notebook_image_pull_policy(notebook, body, defaults):
...
+    notebook_annotations = notebook["metadata"]["annotations"]
+    http_headers_request_set = get_form_value(body, defaults, "httpHeadersRequestSet")
+    if http_headers_request_set:
+        notebook_annotations["notebooks.kubeflow.org/http-headers-request-set"] = http_headers_request_set
+
+
+def set_server_type(notebook, body, defaults):
```
kimwnasptd, line 152: Lets extend this function to add the necessary headers and Istio URI rewrite in case of rstudio and vscode

components/crud-web-apps/jupyter/backend/apps/common/form.py
```diff
@@ -135,6 +135,25 @@ def set_notebook_image_pull_policy(notebook, body, defaults):
...
+    notebook_annotations = notebook["metadata"]["annotations"]
+    notebook_annotations["notebooks.kubeflow.org/server-type"] = get_form_value(body, defaults, "serverType")
```
kimwnasptd, line 154: Lets also add some logic here that would handle incorrect values.

The expected values here, from the request, would be:
* `None`, which means `"jupyter"`
* `"jupyter"`
* `"rstudio"`
* `"vscode"`

If this field is present in the request but has a different value from the expected ones then the backend should raise a `400` error

components/crud-web-apps/jupyter/backend/apps/default/routes/post.py
```diff
@@ -27,6 +27,9 @@ def post_pvc(namespace):
 
     form.set_notebook_image(notebook, body, defaults)
     form.set_notebook_image_pull_policy(notebook, body, defaults)
+    form.set_notebook_http_rewrite_uri(notebook, body, defaults)
+    form.set_http_headers_request_set(notebook, body, defaults)
```
kimwnasptd, line 31: As mentioned above lets remove these two functions

components/crud-web-apps/jupyter/backend/apps/default/routes/post.py
```diff
@@ -27,6 +27,9 @@ def post_pvc(namespace):
...
+    form.set_server_type(notebook, body, defaults)
```
kimwnasptd, line 32: Could you also add this function to the Rok flavor of the app as well? In the `backend/apps/rok/routes/post.py` file

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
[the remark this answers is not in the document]
DavidSpek: This was very briefly discussed on above. However, the RStudio guidelines for using their trademark are very strict (Jupyter is the complete opposite and definitely not a problem, VSCode is still unknown). For RStudio, the only permitted use that could apply is: "Your project deploys RStudio products in unmodified form supplied by RStudio, PBC **for internal use only**;". In other words, EKF can probably not use the RStudio logo (you build your own images anyways, but still), and any vendor that offers Kubeflow-as-a-Service (StatCan, though it might pass as internal for now) or provides a distribution. These uses would all need written approval from RStudio to use their logo in the UI. And VSCode is still unknown at this point. 
This is the reason why I would like to have a solution that doesn't require building a new image, but rather just changing a configmap or something similar. This logo issue is something that will also be faced if the Spawner UI becomes more configurable in the future (the admin can create whatever groups of images he wants, and the logos will need to come from somewhere).

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
```diff
@@ -106,7 +106,49 @@ export class FormDefaultComponent implements OnInit, OnDestroy {
...
+        delete notebook.imageVSCode;
+        delete notebook.imageRStudio;
+        // Set base URI to / for Istio rewrite
+        notebook.httpRewriteURI = '/';
+      } else if (notebook.serverType == 'rstudio') {
+        // Remove unnecessary images from payloaf
+        delete notebook.imageVSCode;
+        delete notebook.imageRStudio;
+        // Set base URI to / for Istio rewrite
+        notebook.httpRewriteURI = '/';
```
kimwnasptd, line 124: Lets also remove the `notebook.httpHeadersRequestSet` and `notebook.httpRewriteURI` from this part of the code as well

kimwnasptd (commented):
The next review batch is around how we handle the http-headers-set and istio-uri-rewrite between the fronend and the backend.

Since right now the frontend does not allow the user to modify the headers to set and the Istio rewrite we should not expect the frontend to send these fields to the backend. For the first iteration the backend can expect only the server-type and set the required annotations for the server to work.

In the future we can tackle how the frontend should expose these fields, how to include them in the config yaml, extend the backend to expect these values etc as a feature of its own.

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
```diff
@@ -106,7 +106,49 @@ export class FormDefaultComponent implements OnInit, OnDestroy {
...
+        // Add X-RStudio-Root-Path header to requests for RStudio
+        const headerValue = '/notebook/' + notebook.namespace.toString() + '/' + notebook.name.toString() + '/';
+        notebook.httpHeadersRequestSet = JSON.stringify({ 'X-RStudio-Root-Path': headerValue });
+      }
+
+    // Ensure CPU input is a string
+    if (typeof notebook.cpu === 'number') {
```
DavidSpek, line 150: @kimwnasptd I just remembered I needed to add this here to fix https://github.com/kubeflow/kubeflow/issues/5641. Without this, the JWA would not work.

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.html
```diff
@@ -9,7 +9,11 @@
         <app-form-image
           [parentForm]="formCtrl"
           [images]="config?.image?.options"
-          [readonly]="config?.image?.readOnly"
+          [jupyterReadonly]="config?.image?.readOnly"
+          [imagesVSCode]="config?.imageVSCode?.options"
```
kimwnasptd, line 13: These will need some slight modification to look at, like `config?.image?.vscode` instead

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
```diff
@@ -106,7 +106,49 @@ export class FormDefaultComponent implements OnInit, OnDestroy {
 
     // Use the custom image instead
     if (notebook.customImageCheck) {
+      // Remove unnecessary images from payloaf
+      delete notebook.imageVSCode;
```
kimwnasptd, line 110: IIUC we always want to remove these extra fields right? In this case we can do this outside of all the if statements once.

Also since we will expect the frontend to only send the `serverType`, and not the headers/uri-rewrite, this section will be significantly simplified.

kimwnasptd (commented):
And finally for the last batch of review, lets reorganize the newly introduced sections in the yaml like this:
```yaml
image:
  value: some-default-image
  readOnly: false
  jupyter:
    - image1
    - image2
  rstudio:
    - image1
    - image2
    - image3
  vscode:
    - image1
```

My reasoning is mostly affected from how we want to handle the `readOnly` field for the image. The concept of `readOnly` is that the admin wants to block the user from specifying a field and have the backend use the value from the config instead.

Following this principle the user should either be able to edit the entire image section, independently of what server type is used, or not. So if the `image` option in the yaml is readonly then the user could not select any image to use, and the input should always be disabled.

WDYT?

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.html
DavidSpek: Not just these, but also how the types.ts and config.ts work if I'm correct (not looking at the code at the moment).

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
DavidSpek: There isn't really a reason to remove them, but I thought it would be better not to send unused stuff in the request. Indeed this can most likely be simplified.

DavidSpek:
@kimwnasptd I think it makes sense to only have one readOnly flag. Then again, I don't know all the intricacies of how some admins would like to lock things down for their specific environment. 
Regarding the way readOnly works. At the moment it seems to disable the drop down list completely. However, I would expect this to limit people to only use only images listed in the yaml. I think this is a separate issue that needs fixing (I'm not against adding the fix in here though, but I haven't looked at it yet).

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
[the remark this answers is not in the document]
DavidSpek: @kimwnasptd I think I just thought of the solution for this. If we use links for the logos that direct to the SVGs hosted in this repository, we can swap the SVGs for something else if using the logos turns out to be a problem. I believe if this would the propagate to all running installations as well (if the cache of the original SVG is cleared). Does this make sense?

components/crud-web-apps/jupyter/backend/apps/common/form.py
DavidSpek: Does adding error logic make sense here? The frontend doesn't rely on user input so unless there is a MITM change to the request a wrong input cannot happen. Also `none` cannot happen, as one of the toggles is always selected. It's not as if the JWA backend is decoupled from the frontend (they exist in the same container and are built together). 
Related to this, how would I go about causing a 400 error specifically in the backend code?

kimwnasptd: Indeed, the frontend sends only specific values but it's a good practice for the backend to make it explicit to the user/frontend that the data it received was not formatted as expected.

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
kimwnasptd: Oh maybe I didn't express it clearly, I agree that the frontend should not be sending the `imageVSCode` and `imageRStudio` fields. 

My comment is to write the `delete notebook.imageVSCode` [ and for rstudio ] commands in one place at the top and not do it in multiple places inside the different ifs

DavidSpek: Sorry, I think my comment wasn't clear. We are completely on the same page ;).

components/crud-web-apps/jupyter/backend/apps/common/form.py
DavidSpek: Should I handle the error with `log.error`?

kimwnasptd: And you could look at this snippet for how to raise this error

https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/backend/apps/common/form.py#L31-L34

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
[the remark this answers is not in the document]
kimwnasptd: From what I understand the issue is not how the icon will appear to the UI [ via the backend directly, fetched from somewhere public etc ] but the fact that we use this logo in the software. With that I mean that if someone would use the app he would end up seeing the logo, which could imply that this is endorsed from RStudio etc.

Also an issue with fetching resources in the UI directly from the public internet is that the app unable will not work in air-gaped environments, where the users would only have access to specific IPs.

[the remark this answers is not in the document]
kimwnasptd: So judging from our current situation, where we are tied by time constraints and the trademark situation would need some communication and tip toeing into what it's acceptable I would propose the following solution to unblock ourselves:

1. Don't use the logos in this PR. We could use the [group options](https://v8.material.angular.io/components/select/overview#creating-groups-of-options) from Angular material, or just use buttons with text [ would prefer the first approach tbh ] to show the images for the different servers
2. Open an issue about it and involve @castrojo in this. And once we figure out what we are allowed to do with the rstudio trademark we can open a new PR and add this functionality. We've done it once here so this won't be difficult to do in the future

WDYT?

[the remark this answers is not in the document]
DavidSpek: Pulling it from the repo wouldn't solve your first point, but it does blur the distribution and way the logo is implemented in the software a bit. It could even directly link to the original source of the RStudio SVG. Indeed, for airgaped environment this might pose a slight problem. However, I think that people in that situation could build their own image (and might already be doing so as they would have a private registry as well). Also, I don't think the functionality of the JWA would actually break, the button would just be empty. 

I think the other problem with this is decision is who is responsible for the decision to include the logos or not and who would be responsible if RStudio isn't happy. If I'm the person responsible since it is my PR that adds their logo, I'm fine with that (and would prefer this actually). If I can be the person responsible for this, I'd go for adding the logos and remove them if there is a complaint. I think an argument can be made that the logo is being used according to the guidelines, if internal use is interpreted as "the kubeflow/wg-notebooks developers deploying unmodified RStudio binaries supplied by RStudio to test their functionality in Kubeflow during development".

components/crud-web-apps/jupyter/backend/apps/common/form.py
DavidSpek: Thanks for the link. Sorry for wasting your time on this, I should have looked at the code a bit better and I would have found it.

kimwnasptd: You are not wasting anyone's time. Your efforts are as important as mine and no releases were delayed because of a question :)

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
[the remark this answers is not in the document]
DavidSpek: I only just saw your second comment. Personally, I would then be more in favor of using the button toggle that is there now, but using an SVG with text in it for VSCode and RStudio. This still makes it rather easy for users to customize if they want the actual logos (which they might be allowed to use in their situation). It also allows us to keep using the Jupyter logo (as we can definitely use that one), and works together with https://github.com/kubeflow/kubeflow/pull/5646#pullrequestreview-614539090 as well (assuming I get this implemented quickly).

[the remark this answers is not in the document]
kimwnasptd: Let's go with using the svgs. We'll use links to fetch them from where they publicly host them. Distributions could make a small change and add their svgs to the app and simply change these links, which is something that we were OK with from the beginning. Lets just add a small section for this in the app's README.

Regarding the responsibility part, I'm not a layer on this so not an expert but we are a team and all of us move forward together. We are not going to point fingers to people if things go south. And also since I'm a reviewer and agreed to move forward with this so I'm also responsible.

But if indeed people complain about the usage then we will sincerely apologize and just change this to something like we discussed above.

DavidSpek:
@kimwnasptd Adding to my comment above, the way yous example yaml only allows for 1 default image, however, there needs to be a default for each of the server types. While busy with the `readOnly` logic anyway we should fix it if it is indeed broken (only default value can be selected rather than any of the options in the drop down list). Actually, now that I think about it a readOnly section for the drop down lists that disables it doesn't make much sense (if the admin only wants 1 image, they would only have 1 image in the config). What I think is the intended usage is enabling/disabling the custom image check box, which should just get it's own value in the config.

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.ts
```diff
@@ -10,11 +12,19 @@ import { Subscription } from 'rxjs';
...
+  @Input() imagesVSCode: string[];
+  @Input() vscodeReadonly: boolean;
+  @Input() imagesRStudio: string[];
+  @Input() rstudioReadonly: boolean;
 
   subs = new Subscription();
 
-  constructor() {}
+  constructor(iconRegistry: MatIconRegistry, sanitizer: DomSanitizer) {
+    iconRegistry.addSvgIcon('jupyterlab', sanitizer.bypassSecurityTrustResourceUrl('static/assets/jupyterlab-wordmark.svg'));
```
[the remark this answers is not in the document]
DavidSpek: Just to be clear from what was discussed above, should the URL in the environment be were the SVG is hosted publicly, or should that first be downloaded by the backend and placed in the `static/assets/` directory? Using the public URL directly means I'll need to deal with CORS.

kimwnasptd:
Your comments make sense. Regarding the readonly handling, lets rename the `readOnly` field to `allowCustomImage` [ or something similar ].

About the default values for the different server types one option would be to make the image section be:
```yaml
image:
  allowCustomImage: true
  jupyter:
    options:
      - image1
    default: image1
  vscode:
...
```

components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-image/form-image.component.ts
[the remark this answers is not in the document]
DavidSpek: @kimwnasptd Friendly ping on this in case you missed it, so I know what steps I need to take to get this resolved.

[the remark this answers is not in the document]
kimwnasptd: Thanks for the ping! 

Good catch on the CORS issue. Lets dump the SVGs in the `static/assets` folder. And also define the links that the frontend will use in the `environment` files as mentioned above.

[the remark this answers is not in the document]
DavidSpek: Cool! This shouldn't take long to implement then.

DavidSpek:
The failed test was the access-management build: https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-ac-mgr-tests-5646-2605340-5088-94f9?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-ac-mgr-tests-5646-2605340-5088-94f9-681790487. 
See my comment here: https://github.com/kubeflow/kubeflow/pull/5693#issuecomment-802229766

DavidSpek:
@kimwnasptd How important is it to have the image groups (jupyter, vscode, rstudio) togetehr under `image`? I got started on trying to implement this and I am noticing that this change would require a change in most of the logic in the frontend (maybe also backend), which I think will start getting close to what Mathew and I are creating a proposal for. If the following structure can be used, I can get this part of the review sorted very quickly:

```
image: #for backwards compatibility
  value:
  options:
    -
imageVSCode:
  value:
  options:
    -
imageRStudio:
  value:
  options:
    -
allowCustomImage: true
```

kimwnasptd:
> @kimwnasptd How important is it to have the image groups (jupyter, vscode, rstudio) togetehr under image?

I'm OK with not putting them under `image` and using your mentioned format. The move to replace `readOnly` with `allowCustomImage` is a good next step forward that unties our hands from needing to group them together. 

My first thought would be to rename `image` to `imageJupyter`. But considering that the UI doesn't handle gracefully the case where the config is not as expected I would also vote for keeping the `image` name for backwards compatibility. Let's just add a comment below it that these images are for the jupyter servers.

DavidSpek:
@kimwnasptd Great timing! I just pushed the commit that implements this. There is already a comment that specifies the `image:` is for Jupyter images.

components/crud-web-apps/jupyter/frontend/src/app/pages/index/index-default/config.ts
DavidSpek: I think this is the only thing left to try and implement (the I've already added the SVGs for this as well).

kimwnasptd:
@DavidSpek looking at the commits I see you've addressed all of my comments, thanks!

I'll test the PR tomorrow in my cluster and if everything works as expected I'll merge the PR

DavidSpek:
@kimwnasptd Awesome! For what it's worth I've just deployed this latest iteration along with the notebook-controller that's on ECR and it is working as expected (got my cluster somewhat fixed again).

kimwnasptd:
I tested the changes locally and in my cluster and the app works as we described.

Thank you for your efforts on this @DavidSpek, it is a very nice addition! 

I'll merge this PR and move forward with updating the manifests. I was preparing a PR that adds some missing ENV Vars plus the use of the AWS image, so I'll include these changes there as well. I'll also modify the spawner.config in the manifests to match the one from this PR.

I'll just use the Notebook images in the configmap from this PR, but we can update to use the latest ones in a subsequent PR

/lgtm
/approve

PR #5642, opened 2021-02-26
[WIP] Test optional-test-infra

PatrickXYS:
**DO NOT MERGE!**

This PR is used to test against kubeflow/kubeflow configurations on optional-test-infra

/hold

PR #5653, opened 2021-03-01
CI: Correct the name of the workflow

kimwnasptd:
right now the PR's tests fail, for the common code, because the name of the workflow is not DNS1123 compatible 
https://github.com/kubeflow/kubeflow/pull/5642#issuecomment-786877663

This is needed to fix our CI tests #5482 

I'll use this PR to fix the test-case. I also introduced a test file just to trigger the tests. Once we ensure the CI works as expected then I'll push a commit to delete it

/cc @PatrickXYS

### DIFF
--- a/py/kubeflow/kubeflow/ci/shared_ui_tests.py
+++ b/py/kubeflow/kubeflow/ci/shared_ui_tests.py
@@ -51,6 +51,20 @@ class Builder(workflow_utils.ArgoTestBuilder):
 
         return ui_build
 
+    def _create_exit_handler(self, task_template):
+        ui_build = argo_build_util.deep_copy(task_template)
+
+        ui_build["name"] = "rm-node-modules"
+        ui_build["container"]["image"] = "node:12.20.1-stretch-slim"
+        ui_build["container"]["command"] = ["rm"]
+        ui_build["container"]["args"] = ["-r", "node_modules"]
+
+        ui_dir = ("%s/components/crud-web-apps/common/"
+                  "frontend/kubeflow-common-lib/") % self.src_dir
+        ui_build["container"]["workingDir"] = ui_dir
+
+        return ui_build
+
     def build(self):
         """Build the Argo workflow graph"""
         workflow = self.build_init_workflow()
@@ -68,11 +82,18 @@ class Builder(workflow_utils.ArgoTestBuilder):
                                         ui_tests_task,
                                         [modules_install_task["name"]])
 
-        # run common ui frontend tests
+        # build the node module from the lib source code
         build_step = self._create_ui_build_task(task_template)
         argo_build_util.add_task_to_dag(workflow, workflow_utils.E2E_DAG_NAME,
                                         build_step,
                                         [modules_install_task["name"]])
+
+        # remove node_modules folder as exit handler
+        rm_node_modules = self._create_exit_handler(task_template)
+        argo_build_util.add_task_to_dag(workflow, workflow_utils.EXIT_DAG_NAME,
+                                        rm_node_modules,
+                                        [build_step["name"],
+                                         ui_tests_task["name"]])
 
         # Set the labels on all templates
         workflow = argo_build_util.set_task_template_labels(workflow)
--- a/py/kubeflow/kubeflow/ci/shared_ui_tests.py
+++ b/py/kubeflow/kubeflow/ci/shared_ui_tests.py
@@ -88,12 +88,10 @@ class Builder(workflow_utils.ArgoTestBuilder):
                                         build_step,
                                         [modules_install_task["name"]])
 
-        # remove node_modules folder as exit handler
+        # EXIT-HANDLER: remove node_modules folder as exit handler
         rm_node_modules = self._create_exit_handler(task_template)
         argo_build_util.add_task_to_dag(workflow, workflow_utils.EXIT_DAG_NAME,
-                                        rm_node_modules,
-                                        [build_step["name"],
-                                         ui_tests_task["name"]])
+                                        rm_node_modules, [])
 
         # Set the labels on all templates
         workflow = argo_build_util.set_task_template_labels(workflow)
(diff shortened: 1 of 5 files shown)


kimwnasptd:
Renaming the workflow's name to replace `-` with `_` resolved our last problem with DNS1123 names but we now bump into a new one, which is the 63 char limit.
http://prow.kubeflow-testing.com/view/s3/aws-kubeflow-jenkins/pr-logs/pull/kubeflow_kubeflow/5653/kubeflow-kubeflow-presubmit/1366457561569562624
```
Invalid value: \"kubeflow-kubeflow-presubmit-shared-ui-tests-5653-bd89676-2624-af49\": must be no more than 63 characters"
```

I'll rename the test to something else, like simply `common-ui` to further test that it works as expected. If we find ourselves bumping into it too often then lets discuss if we could slightly change the naming scheme of the workflows.

kimwnasptd:
Getting closer. Looking at the argo ui I can see the following 
https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-common-ui-5653-61f29a5-0128-c964?tab=workflow
```
Failed: invalid spec: templates.exit-handler must have at least one task
```

Although this is weird since I didn't see this error in my local cluster. In any case I'll add a task for the `template.exit-handler` to remove the installed `node_modules` folder and see how it goes.

kimwnasptd:
This time the workflow gets submitted successfully, but fails because the image that runs the tests is missing a dependency [ ChromeHeadless ]. I'll see if I can use another image.

Also another thing I noticed is that it takes ~5 minutes to install the node modules, so maybe it makes sense to not remove them after each workflow, via the `exit-handler`. 

@PatrickXYS if I don't delete the `node_modules` after the worfklow runs then they will still be available for the next run, since they will be installed in the shared NFS volume, right? Also there won't be an issue if the workflow doesn't have an `exit_handler` dag right?

PatrickXYS:
> if I don't delete the node_modules after the worfklow runs then they will still be available for the next run, since they will be installed in the shared NFS volume, right? Also there won't be an issue if the workflow doesn't have an exit_handler dag right?

Though it's installed in the shared NFS

kimwnasptd:
>Though it's installed in the shared NFS volume, but the thing will not work as you expected.
> 
> Because for every single test, the mount path is different such as /mnt/test-data-volume/kubeflow-kubeflow-presubmit-common-ui-5653-f381623-4832-76b4. For different test, it's super hacky to find previous mount path

Makes sense. Let's not go down this road then.

Now, for the testing step, I need to have an oci image with Chrome and node installed. I've used the [browserless/chrome](https://hub.docker.com/r/browserless/chrome) image and run the tests successfully locally inside the container.

@PatrickXYS do you think there's going to be an issue with this image in the AWS cluster?

PatrickXYS:
> 01 03 2021 19:44:58.513:ERROR [launcher]: No binary for ChromeHeadless browser on your platform.
  Please, set "CHROME_BIN" env variable.

From the log, would setting up the env var be a feasible solution to fix the issue? 

> Now, for the testing step, I need to have an oci image with Chrome and node installed. I've used the browserless/chrome image and run the tests successfully locally inside the container.

I don't think there's a reason that a container image not working properly in the EKS cluster, or more specifically in the EC2 worker node.

kimwnasptd:
> From the log, would setting up the env var be a feasible solution to fix the issue?

Τhe issue is that Chrome is not installed at all in the official `node` image.

> I don't think there's a reason that a container image not working properly in the EKS cluster, or more specifically in the EC2 worker node.

ACK, I was thinking if there was a security concern wrt what images should be run in the cluster. I guess no so we should be good to go. The test also passed https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-common-ui-5653-e340f60-8064-71ed?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-common-ui-5653-e340f60-8064-71ed-1717785880

I'll test a commit that breaks the tests to see if this also works as expected and lastly I'll push another commit to rename the `shared_ui_tests` to `common_ui_tests` everywhere, like the workflow name.

py/kubeflow/kubeflow/ci/shared_ui_tests.py
```diff
@@ -27,7 +27,7 @@ def _create_ui_tests_task(self, task_template):
         ui_tests = argo_build_util.deep_copy(task_template)
 
         ui_tests["name"] = "shared-ui-tests"
-        ui_tests["container"]["image"] = "node:12.20.1-stretch-slim"
+        ui_tests["container"]["image"] = "browserless/chrome"
```
PatrickXYS, line 30: Should we tag to a fixed version?

PatrickXYS (commented):
overall lgtm, only nit change

py/kubeflow/kubeflow/ci/shared_ui_tests.py
kimwnasptd: That's a very good point! Doing it now

kimwnasptd: I pinned it to the [1.44-chrome-stable](https://hub.docker.com/layers/browserless/chrome/1.44-chrome-stable/images/sha256-06c88d84d658321f1db07ca7f3500774494adbfb9295c2131ecbccad7ba826f7?context=explore) tag. It worries me that in the base image they just install the latest nodejs. But since we'll be using a specific commit I think we'll be OK, but lets just have it in mind.

I also tried to run the tests with this tag and they run as expected

kimwnasptd:
OK, the tests do fail as expected with the new commit. Task failed successfully :stuck_out_tongue: 

I'll move forward with reverting the test-breaking commit and rename the files.

kimwnasptd:
And we should be good to roll! The tests pass and fail as expected.

We should keep in mind the following:
1. The 63 chars limitation, which means that the workflow's name can not be more than ~12 chars
2. The name of the workflow needs to be DNS1123 competible, since it will be used in the Argo Worfklow CR's name
3. We use the [browserless/chrome:1.44-chrome-stable](https://hub.docker.com/layers/browserless/chrome/1.44-chrome-stable/images/sha256-06c88d84d658321f1db07ca7f3500774494adbfb9295c2131ecbccad7ba826f7?context=explore) image for running the tests, which includes both `node` and `chrome`

@PatrickXYS feel free to add any more comments you might have. If you are OK then @Bobgy I'll need to request some of your cycles to help merge this, since we also changed the `prow_config.yaml` file.

Bobgy:
/lgtm
/approve
Sure

Issue #5588, opened 2021-02-08
As a user of kubeflow, I am able to access the jupyter servers (notebooks) inside other namespaces although i am not assigned as a contributor


Hmr-ramzi:
/kind bug

**What steps did you take and what happened:**
Created 2 users (X and Y). Logged in with user X, Created a jupyter server (notebook server) A. Logged in with user Y, edited the url to access the jupyter server A. I was able to access it using url (/notebook/X/A/tree?) 

**What did you expect to happen:**
To get a 403 error instead

**Anything else you would like to add:**
None

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 1.2
- kfctl version: (use `kfctl version`): 1.2
- Kubernetes platform: (e.g. `minikube`) Vanilla
- Kubernetes version: (use `kubectl version`): 1.17
- OS (e.g. from `/etc/os-release`): N/A

jbottum:
@DavidSpek have you validated this behavior ?

DavidSpek:
@jbottum I have not, as I am currently busy with developing on the new jupyter web app that uses the common code. However, this is very likely related to a known issue that occurs in multiple web apps. See https://github.com/kubeflow/katib/issues/1322, specifically https://github.com/kubeflow/katib/issues/1322#issuecomment-687640209.

connorlwilkes:
I have also encountered this issue. 

It appears that given a URL any user can access anyone else's notebook instances. This is problematic for security particularly in a multi tenant environment.

You could resolve this in the GCP world by utilising the IAP ingress header (https://cloud.google.com/iap/docs/identity-howto) and adding a match on the header in the 'match' block of the VS. This would mean only users with that header would then be able to access this notebook instance. Unsure whether such headers are injected with any other auth method?

yanniszark:
@Hmr-ramzi @connorlwilkes could you please try installing latest Kubeflow 1.3 Release candidate from these instructions:
https://github.com/kubeflow/manifests/tree/v1.3-branch#readme

and let us know what happened? Please keep in mind that some distributions in 1.2 didn't have support for multi-user isolation, so you may have bumped into that.

connorlwilkes:
> @Hmr-ramzi @connorlwilkes could you please try installing latest Kubeflow 1.3 Release candidate from these instructions:
> https://github.com/kubeflow/manifests/tree/v1.3-branch#readme
> 
[...]

PR #5660, opened 2021-03-03
(Notebook-controller): Add `http-rewrite-uri` and `http-headers-request-set` annotations

DavidSpek:
This PR contains the notebook controller code originally found in https://github.com/kubeflow/kubeflow/pull/5601 and https://github.com/kubeflow/kubeflow/pull/5618. It allows users to configure the Istio VirtualService spec that the notebook controller creates by adding annotations to the notebook resource metadata. Specifically, two option for the VirtualService spec can be customized: the HTTP rewrite rule and request headers the Istio proxy adds to requests.

Currently, the notebook-controller configures the Istio rewrite rule according to the following example snippet:
```
http:
  - match:
      - uri:
          prefix: /notebook/<namespace>/<notebook-name>/
    rewrite:
      uri: /notebook/<namespace>/<notebook-name>/
```
This is fine for applications such as Jupyter that allow you to specify (with an environment variable) the base-uri they are hosted at. Thus, JupyterLab is hosting it's service on the pod at `localhost:8888/notebook/<namespace>/<notebook-name>/`. However, many applications do not allow you to specify the base-uri or this cannot easily be done with an environment variable. Examples of such applications are R-Studio and VS-Code. To support these applications, this PR allows users to set the rewrite rule by adding the annotation `notebooks.kubeflow.org/http-rewrite-uri` with as value the path the application is hosted at. For most applications, including R-Studio and VS-Code, this will simply be `/`. Thus, by adding `notebooks.kubeflow.org/http-rewrite-uri: /` as an annotation to the notebook resource, the Istio rewrite snippet will look like this:
```
http:
  - match:
      - uri:
          prefix: /notebook/<namespace>/<notebook-name>/
    rewrite:
      uri: /
```

For some applications, it might be necessary to change or add request headers to the application. In particular, R-Studio requires the `X-RStudio-Root-Path` header to be added with the value `/notebook/<namespace>/<notebook-name>/` (for those familiar, this is supposed to do the same as setting `www-root-path` in the `reserver.conf` file, but this broken in same situations and also requires a script that runs on container startup to configure). Thus, to be able to support R-Studio, the functionality to change an existing or add a new request headers via an annotations on the notebook resource was added in this PR. For example, for R-Studio the annotation `notebooks.kubeflow.org/http-headers-request-set: "X-RStudio-Root-Path: /notebook/<namespace>/<notebook-name>/"` is added to the notebook resource. The snippet of the resulting Istio VirtualService will look like:
```
http:
  - headers:
      request:
        set:
          X-RStudio-Root-Path: /notebook/<namespace>/<notebook-name>/
```

The notebook-controller supports having multiple headers set in the annotations as a literal string. For example, to add a second header and value to the Istio VirtualService the annotation can be set as follows:
`notebooks.kubeflow.org/http-headers-request-set: "X-RStudio-Root-Path: /notebook/admin/new-r-studio-no-root/\nkey1: value1"`
The snippet of the resulting Istio VirtualService will look like:
```
http:
  - headers:
      request:
        set:
          X-RStudio-Root-Path: /notebook/<namespace>/<notebook-name>/
          key1: value1
```
Note that `\n` is used to separate the first and second header key:value pairs and that double quotes are used for the value of the annotation. This is because the multiple values must be entered into the annotations as a literal string. 

Together, these two additional configuration parameters greatly expands the usability of the notebook controller for different types of services, with R-Studio and VS-Code being two of them. 

To test this PR, the image of the notebook-controller deployment can be changed to `davidspek/notebook-controller-update:0.3`. 

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -17,6 +17,7 @@ package controllers
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -46,6 +47,8 @@ import (
 
 const DefaultContainerPort = 8888
 const DefaultServingPort = 80
+const AnnotationRewriteURI = "notebooks.kubeflow.org/http-rewrite-uri"
+const AnnotationHeadersRequestSet = "notebooks.kubeflow.org/http-headers-request-set"
 
 // The default fsGroup of PodSecurityContext.
 // https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#podsecuritycontext-v1-core
@@ -400,7 +403,19 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
 	namespace := instance.Namespace
 	clusterDomain := "cluster.local"
 	prefix := fmt.Sprintf("/notebook/%s/%s/", namespace, name)
+
+	// unpack annotations from Notebook resource
+	annotations := make(map[string]string)
+	for k, v := range instance.ObjectMeta.Annotations {
+		annotations[k] = v
+	}
+
 	rewrite := fmt.Sprintf("/notebook/%s/%s/", namespace, name)
+	// If AnnotationRewriteURI is present, use this value for "rewrite"
+	if _, ok := annotations[AnnotationRewriteURI]; ok && len(annotations[AnnotationRewriteURI]) > 0 {
+		rewrite = annotations[AnnotationRewriteURI]
+	}
+
 	if clusterDomainFromEnv, ok := os.LookupEnv("CLUSTER_DOMAIN"); ok {
 		clusterDomain = clusterDomainFromEnv
 	}
@@ -424,8 +439,29 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
 		return nil, fmt.Errorf("Set .spec.gateways error: %v", err)
 	}
 
+	headersRequestSet := make(map[string]string)
+	// If AnnotationHeadersRequestSet is present, use its values in "headers.request.set"
+	if _, ok := annotations[AnnotationHeadersRequestSet]; ok && len(annotations[AnnotationHeadersRequestSet]) > 0 {
+		requestHeadersBytes := []byte(annotations[AnnotationHeadersRequestSet])
+		if err := json.Unmarshal(requestHeadersBytes, &headersRequestSet); err != nil {
+			// if JSON decoding fails, set an empty map
+			headersRequestSet = make(map[string]string)
+		}
+	}
+	// cast from map[string]string, as SetNestedSlice needs map[string]interface{}
+	headersRequestSetInterface := make(map[string]interface{})
+	for key, element := range headersRequestSet {
+		headersRequestSetInterface[key] = element
+	}
+
+	// the http section of the istio VirtualService spec
 	http := []interface{}{
 		map[string]interface{}{
+			"headers": map[string]interface{}{
+				"request": map[string]interface{}{
+					"set": headersRequestSetInterface,
+				},
+			},
 			"match": []interface{}{
 				map[string]interface{}{
 					"uri": map[string]interface{}{
@@ -449,6 +485,8 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
 			"timeout": "300s",
 		},
 	}
+
+	// add http section to istio VirtualService spec
 	if err := unstructured.SetNestedSlice(vsvc.Object, http, "spec", "http"); err != nil {
 		return nil, fmt.Errorf("Set .spec.http error: %v", err)
 	}


kimwnasptd:
I'll also start reviewing this PR so that we can have the necessary infra ready for the new R-Studio and VSCode images. This can be done in parallel with our discussion around the image names #5575. I'll also try to update our [design doc](https://docs.google.com/document/d/14ve8inClzkhIILAIp2FodemRtoS2hqanX8jll2lmRUE/edit#heading=h.47k606j3mc2v) with the designs we will make here.

@DavidSpek is the PR ready for review?

DavidSpek:
@kimwnasptd In my view it is ready for merging even.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -400,7 +400,21 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
 	namespace := instance.Namespace
 	clusterDomain := "cluster.local"
 	prefix := fmt.Sprintf("/notebook/%s/%s/", namespace, name)
-	rewrite := fmt.Sprintf("/notebook/%s/%s/", namespace, name)
+	var rewrite string
```
kimwnasptd, line 403: I think we should use the _short deceleration operator_ here `:=` instead of `var` since AFAIK `var` also makes the variable have package level scope.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -449,6 +464,32 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
...
+	// split the literal string into keys and values and add them
+	// to the headers variable. Finally, add the headers section to
+	// the variable with the http spec.
+	if _, ok := annotations["notebooks.kubeflow.org/http-headers-request-set"]; ok && len(annotations["notebooks.kubeflow.org/http-headers-request-set"]) > 0 {
+		requestHeaders := strings.Split(annotations["notebooks.kubeflow.org/http-headers-request-set"], "\n")
+		for _, kv := range requestHeaders {
+			if len(strings.Split(kv, ": ")) == 2 {
+				k := strings.Split(kv, ": ")[0]
+				v := strings.Split(kv, ": ")[1]
+				headers["request"].(map[string]interface{})["set"].(map[string]interface{})[k] = v
```
kimwnasptd, line 486: I think we should not let users arbitrarily add headers to the requests that pass from the ingress gateway. For example they could even override the ISTIO `userid-header`, the `X-Forwarded-Proto` so that the server assumes http instead of https etc.

I think we should reconsider the design we take here
also cc @yanniszark who has a lot of experience on security

kimwnasptd:
@DavidSpek IIUC in order to run R-Studio we need the following settings:
1. Istio to do the path/prefix rewriting to root
2. Configure the `X-RStudio-Root-Path` header to use the value of the prefix

Instead of adding annotations that allow users to configure each one of the above I would propose that we introduce a single `notebooks.kubeflow.org/type: {jupyter,rstudio,vscode}` [ if omitted then `jupyter` is assumed ] annotation. So if the Notebook CR has the annotation `notebooks.kubeflow.org/type: rstudio` then the controller will apply these specific two settings.

Ideally this should be in the CR's spec, but since we can't change the `spec` right now we kind of have our hands tied there.

Also cc @elikatsis and @yanniszark who have more experience on k8s controllers

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: Without this, rewrite cannot be referenced in the if statement starting on line 412.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: I believe if we change the method to `add` instead of `set` the header only gets added if it is not already present.

kimwnasptd:
meant to write
/cc @elikatsis 
/cc @yanniszark

DavidSpek:
`notebooks.kubeflow.org/server-type` is already being used for keeping track what type on notebook a user has launched. It would be possible to do what you describe, but especially regarding the path/prefix rewriting the ability to configure this makes it possible to launch any web-app in a container no matter where it is hosted at. By making the annotation configurable, this opens up the possibility to further customize the image settings in the `spawner_ui_config.yaml` in the future. Regarding the header, the security concern might be valid and so it might be better to use the `notebooks.kubeflow.org/server-type` annotation for this.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: If this is a security issue, I think https://github.com/kubeflow/kubeflow/issues/5588 is a much larger problem.

DavidSpek:
@kimwnasptd That is indeed the aim. I think I know a method to not need to remove the header annotation, but make it so that the annotation is managed by the controller (so a user cannot change it). Having the `server-type` annotation also setting other things might not be the best solution either. You cannot change what an annotation does easily as will expect it to do a certain thing, and having dedicated annotations means less assumptions are made about the image that is being run. This still allows for the functionality I know some users and administrators require.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: @kimwnasptd I don't think this is a big problem, as the rest of the virtual service is not touched. So all the user can do is mess up the requests going to the service he already has access to, I don't think this would actually enable somebody to be able to access a notebook in another namespace (unlike the issue linked above). But I am all about learning so I'd like to know the actual dangers/problems this could have.

DavidSpek:
@kimwnasptd I just mentioned this above, but I will mention it again here. I don't think having users be able to change the request headers set by Istio is a (new) security issue, as users can already create and edit virtual services in their namespace as they please.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: @elikatsis Makes sense. Just validated that it still works and pushed the change.

kimwnasptd:
> I don't think having users be able to change the request headers set by Istio is a (new) security issue, as users can already create and edit virtual services in their namespace as they please.

Even if this is an existing security issue then this does not mean should be introducing new code that preserves this issue. 

The controller should only expose the bare minimum and relevant configuration necessary for the user. So unless there are specific use cases where the user need to set other specific headers we should not give them the freedom to arbitrarily configure what headers the ingress gateway sets.

> This still allows for the functionality I know some users and administrators require.

Can you provide some more details about this? What are the use cases and what extra headers did the users need to configure?

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: @kimwnasptd To continue on your last comment below, there are a large amount of header configurations that can be used with RStudio. `X-RStudio-Root-Path` is one of course, `X-RStudio-Request` is another. However, much more can be set with headers: https://docs.rstudio.com/connect/admin/appendix/configuration/#ProxyAuth.Settings. I could be mistaken, but I thought you could also configure or use custom headers with JupyterLab as well. I will need to look at it further tomorrow morning, however, I'm sure there are other applications for which setting headers to configure them is supported and would be useful. I think with the URI rewriting, the configurable request headers, and allowing to set the application port (not included here but I do have a PR open for this), the notebook controller will be compatible with most web-applications a user could want to deploy. 

However, seeing as the issue described here is an issue that is already present, I do not think this is a blocker. Fixing the underlying issue will be a larger task and not something that will be done before the 1.3 release.

thesuperzapper: @kimwnasptd There is a very simple solution here, just reject the annotation if it sets a restricted header name.

This should be easy to implement quickly, and I think we can start with a blacklist (which can be extended in future if issues are found).

So @kimwnasptd what headers should be blacklisted?

DavidSpek: I added a blacklist, so I believe this issue is now resolved.

kimwnasptd: I took a more extensive look at the docs and I propose the following solution:

The headers that are used to configure parts of R-Studio are all non-standard headers, starting with `X-*`. Such example headers are
* `X-RStudio-Request`
* `X-RStudio-Root-Path`
* `X-Forwarded-Host`
* `X-Forwarded-Proto`
* `X-Auth-Token`

So I propose to go the other way around from blacklisting specific headers. Let's only allow the users to edit/set non-standard headers that start with `X-*`. This way we provide all the necessary configuration for R-Studio and at the same time don't allow the users to edit any header arbitrarily.

And if in the future we find users that need to specifically set the well defined headers for a Notebook we can discuss based on the specific use case on how to proceed and allow users to add more headers.

This is the case for the docs I looked up to now
https://docs.rstudio.com/ide/server-pro/access-and-security.htmlhttps://docs.rstudio.com/connect/admin/authentication/proxied/
https://docs.rstudio.com/connect/admin/appendix/configuration/#ProxyAuth.Settings

kimwnasptd:
@DavidSpek was it my idea or did a workflow run for `common-ui` tests?

kimwnasptd:
Indeed it does! This is the relevant log from prow
http://prow.kubeflow-testing.com/view/s3/aws-kubeflow-jenkins/pr-logs/pull/kubeflow_kubeflow/5660/kubeflow-kubeflow-presubmit/1369922260923584512

```
Triggering workflow common-ui because components/crud-web-apps/common/frontend/kubeflow-common-lib/package-lock.json in dir components/crud-web-apps/common/frontend/kubeflow-common-lib/* is modified.
```

I'll investigate this further
cc @PatrickXYS

kimwnasptd:
There are also tests run for Tensorboard controller. @DavidSpek please don't run a retest command for a little bit until we gather the necessary details.

DavidSpek:
@kimwnasptd There are 2 issues here, the wrong tests are being run and the tests are failing. Regardless of what tests are run, they shouldn't be failing.

components/notebook-controller/controllers/notebook_controller.go
kimwnasptd: And lastly, the controller will set accordingly the `X-RStudio-Root-Path` header if the `server-type` is `rstudio`. But the user could still be able to override this by manually specifying a different value.

This way in order for a CR to support RStuido it will only need the `notebooks.kubeflow.org/server-type` annotation, which provides a good default mechanism and extensibility to the users.

DavidSpek:
This is the relevant log for the test actually failing. Guess 3:30 am is not the best time to try and quickly fix code. 
https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-nb-c-tests-5660-cffe277-4512-10a2?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-nb-c-tests-5660-cffe277-4512-10a2-3885565091

DavidSpek:
@kimwnasptd I rebased on master and fixed the issue. Now it is only executing the notebook controller presubmit test. 
https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-nb-c-tests-5660-a642f7a-2752-bcd8

DavidSpek:
@kimwnasptd test has passed, so this should be good to go.

kimwnasptd:
> Guess 3:30 am is not the best time to try and quickly fix code.

Ahh, fan times.

> @kimwnasptd I rebased on master and fixed the issue. Now it is only executing the notebook controller presubmit test.

Awesome! So the takeaway is that if the PR is not on top of the latest master then some extra tests might be triggered, since some extra file modifications are detected. This also makes sense. I'll have it in the back of my mind for future PRs.

> @kimwnasptd test has passed, so this should be good to go.

Nice! I've left some review comments this morning https://github.com/kubeflow/kubeflow/pull/5660#discussion_r592145006. The rest LGTM so once that's done we should be set.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: @kimwnasptd A white-list rather than a blacklist would indeed be a possibility. However, by backing this into the controller administrators cannot easily customize these values. Also, using `X-*` for headers seems to have been deprecated int 2012: see https://tools.ietf.org/html/rfc6648. There might be a way to make this configurable, but I'm not sure this can be done on time. 

Regarding your last comment, are you suggesting that the current mechanisms added to the controller stay as they are, but the controller itself adds the annotations to the CR when `notebooks.kubeflow.org/server-type: rstudio` is present? Because I don't think this would work, as the controller will forcefully change the annotations if they are edited. I think what you are proposing, is essentially already what is being done except for that it is done in the frontend/backend rather than the controller. Users even looking into this feature are probably already aware, or capable of, figuring out how to change the virtual service spec.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -395,12 +395,35 @@ func virtualServiceName(kfName string, namespace string) string {
...
+	// Create a map of all the annotations and assign to a variable
+	annotations := make(map[string]string)
+	for k, v := range instance.ObjectMeta.Annotations {
+		annotations[k] = v
+	}
+
+	rewrite = fmt.Sprintf("/notebook/%s/%s/", namespace, name)
+	// If key "notebooks.kubeflow.org/http-rewrite-uri" is present in the annotations
+	// set the rewrite variable to the value of base-url
+	if _, ok := annotations["notebooks.kubeflow.org/http-rewrite-uri"]; ok && len(annotations["notebooks.kubeflow.org/http-rewrite-uri"]) > 0 {
```
yanniszark, line 423: @DavidSpek these values are used a few times:
"notebooks.kubeflow.org/http-rewrite-uri"
"notebooks.kubeflow.org/http-headers-request-set"

Can you make them into constants, so that they're declared in one place?

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -395,12 +395,35 @@ func virtualServiceName(kfName string, namespace string) string {
 	return fmt.Sprintf("notebook-%s-%s", namespace, kfName)
 }
 
+// Function for the VirtualService header blacklist
+func Find(slice []string, val string) (int, bool) {
```
yanniszark, line 399: Two comments here:
1. nit: Let's use lowercase function names for functions that we do not export (uppercase functions are exported and become part of the package api in Go)
2. The function is called `Find`, but you only use its second argument. I think what you want is a function that checks for existence, cause that's what you check. So how about renaming this to `exists` and returning only a bool?

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -449,6 +473,39 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
...
+	// the variable with the http spec.
+	if _, ok := annotations["notebooks.kubeflow.org/http-headers-request-set"]; ok && len(annotations["notebooks.kubeflow.org/http-headers-request-set"]) > 0 {
+		requestHeaders := strings.Split(annotations["notebooks.kubeflow.org/http-headers-request-set"], "\n")
+		for _, kv := range requestHeaders {
+			if len(strings.Split(kv, ": ")) == 2 {
+				k := strings.Split(kv, ": ")[0]
+				v := strings.Split(kv, ": ")[1]
+				_, key := Find(blacklist, k)
+				_, value := Find(blacklist, v)
+				if !key && !value {
```
yanniszark, line 500: I'm not familiar with this syntax in Go and haven't seen in elsewhere. Can you rewrite this statement to make it more clear what it is you're checking?

components/notebook-controller/controllers/notebook_controller.go
kimwnasptd: > Also, using X-* for headers seems to have been deprecated int 2012: see https://tools.ietf.org/html/rfc6648

This is a very good argument to make to the folks maintaining the R-Studio code :) 

> However, by backing this into the controller administrators cannot easily customize these values.

What extra headers do you think they would need to set aside, from the `X-*`  headers R-Studio knows/expects?

> Regarding your last comment, are you suggesting that the current mechanisms added to the controller stay as they are, but the controller itself adds the annotations to the CR when notebooks.kubeflow.org/server-type: rstudio is present?

No, the controller will not add any extra annotations if the `server-type` is present. It will only configure the vsvc accordingly.
Let me try to clarify with two examples:

CR 1:
```yaml
metadata:
  annotations:
    notebooks.kubeflow.org/server-type: rstudio
```

In this case the controller will see that the Notebook is for R-Studio so it will set the `/` url rewrite in the vsvc and also set the `X-RStudio-Root-Path` header in the vsvc for the prefix

CR 2:
```yaml
metadata:
  annotations:
    notebooks.kubeflow.org/server-type: rstudio
    notebooks.kubeflow.org/http-rewrite-uri: /some/path
```

In this case the controller will first do exactly what it would for CR 1. But then it would also detect the `notebooks.kubeflow.org/http-rewrite-uri` and use this value for the Istio rewrite, instead of the default it previously put. 

Also the same approach will be used with the `notebooks.kubeflow.org/http-headers-request-set`. If this annotation is present then the controller will use the `X-*` headers mentioned in this annotations instead

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -449,6 +473,39 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
...
+	}
+
+	// Blacklist for request headers that are not allowed to be set
+	blacklist := []string{"userid-header", "X-Forwarded-Proto"}
+
+	// If key "notebooks.kubeflow.org/http-headers-request-set" is present on the notebook resource,
+	// split the literal string into keys and values and add them
+	// to the headers variable. Finally, add the headers section to
+	// the variable with the http spec.
+	if _, ok := annotations["notebooks.kubeflow.org/http-headers-request-set"]; ok && len(annotations["notebooks.kubeflow.org/http-headers-request-set"]) > 0 {
```
kimwnasptd, line 492: Also NIT, lets use `http-request-headers-set` [ switch the headers/request words ] instead

DavidSpek: The current naming for these annotations was chosen as this is follows the structure of the resulting VirtualService yaml. 
`notebooks.kubeflow.org/http-rewrite-uri`: sets `http[0].rewrite.uri`
`notebooks.kubeflow.org/http-headers-request-set`: sets `http[0].headers.request.set`

kimwnasptd: ACK. I think it comes more natural to pronounce `http request headers set` rather than `http headers request set`. Could you make this small change?

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -449,6 +473,39 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
...
+		requestHeaders := strings.Split(annotations["notebooks.kubeflow.org/http-headers-request-set"], "\n")
+		for _, kv := range requestHeaders {
+			if len(strings.Split(kv, ": ")) == 2 {
+				k := strings.Split(kv, ": ")[0]
+				v := strings.Split(kv, ": ")[1]
```
yanniszark, line 497: I believe we shouldn't use our own serialization format, but instead take a page from Kubernetes' book.
Kubernetes uses annotations for alpha features. How are those annotations written? With JSON. For example, the `scheduler.alpha.kubernetes.io/affinity` annotation.

Using JSON allows us to leverage existing battle-tested parsers and not implement our own. What do you think?

yanniszark (changes requested):
@DavidSpek thanks for your effort on this! I have left a few comments on the code. The main change I believe we should make is to use JSON for these complex fields in annotations.

Finally, do you have an example Notebook CR you have tested this with?

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: Not sure how to rewrite it. It is using the `Find` function and only continues if the value `k` and `v` are not in the blacklist variable. I'm going to rewrite this to be a whitelist instead, but more that that I am not sure how I would do it.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: I have no idea how to start implementing this, and I don't believe there is enough time to do so anymore either. I'm mainly following what is already being done in the code elsewhere. If you want to improve the code, I'm happy to include it in the PR.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: Makes sense.

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: How do you want it to return only the bool?

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: @kimwnasptd This is now implemented, as well as a whitelist instead of a blacklist

components/notebook-controller/controllers/notebook_controller.go
yanniszark: This is what I mean:
```go
func exists(slice []string, val string) bool {
	for i, item := range slice {
		if item == val {
			return true
		}
	}
	return false
}
```

components/notebook-controller/controllers/notebook_controller.go
DavidSpek: Sure, I could honestly not care less. Pushed the commit.

components/notebook-controller/controllers/notebook_controller.go
yanniszark: @DavidSpek it's a small change. Here is an example you can follow to unmarshal a json map (the use-case we have here):
https://play.golang.org/p/TR-6kdk6s91

components/notebook-controller/controllers/notebook_controller.go
yanniszark: I got confused because of the variable names. I think you should rewrite it like this:
```go
if exists(blacklist, k) {
  continue
}
```
to essentially rule out the bad case and then continue.

BTW, I saw you check both the key and value. Is this right? Shouldn't we check on the header name?

yanniszark (commented):
@DavidSpek I've left another round of comments.

DavidSpek:
@yanniszark I've been testing this with the changes in https://github.com/kubeflow/kubeflow/pull/5646. I think the notebook CR to test it is rather self explanatory (add the annotations).

components/notebook-controller/controllers/notebook_controller.go
thesuperzapper: I think with the blacklist there is very little additional risk from this PR, and I think it's good to merge.

Regarding the server-type thing, we can do that in the future but its not necessary right now. 

Adding a whitelist would only act to reduce the usefulness of this feature for users why might have their own container images.

components/notebook-controller/controllers/notebook_controller.go
kimwnasptd: Thank you!

thesuperzapper: The reason for the strange ordering of the words is because it maps to the Istio VirtualService `http[0].headers.request.set`, therefore, for consistency, let's leave it as is.

components/notebook-controller/controllers/notebook_controller.go
kimwnasptd: > Regarding the server-type thing, we can do that in the future but its not necessary right now.

@thesuperzapper I can understand your urge to have this effort merged as soon as possible, but this is not a big feature so lets get it right from the beginning and not worry about it in the future. The functionality I described is almost there.

> Adding a whitelist would only act to reduce the usefulness of this feature for users why might have their own container images.

Could you elaborate on this? What use cases do you think we limit? With the mention mechanism, of allowing the user to set any `X-*` header they want, we still support all of the configuration options from R-Studio.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -449,6 +478,42 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
...
+	// Create a template map for the headers section
+	// of the Istio VirtualService
+	headers := map[string]interface{}{
+		"request": map[string]interface{}{
+			"set": map[string]interface{}{},
+		},
+	}
+
+	// whitelist for request headers that are not allowed to be set
+	whitelist := []string{"X-Forwarded-Proto", "X-RStudio-Request"}
```
kimwnasptd, line 491: @DavidSpek as mentioned in https://github.com/kubeflow/kubeflow/pull/5660#discussion_r592294749 you don't need to worry about specifying specific headers.

When parsing the full list of provided headers from `notebooks.kubeflow.org/http-request-headers-set` just keep and set those that start with `X-`

DavidSpek:
@yanniszark As you can see the code has changed a lot and is now using `json.Unmarshal`. However, the problem that is now being faced, is when the annotation is not valid (it is a string, or it contains only a  `key` and no `value`) the notebook controller goes into crashbackloop. Here is an output of the log:

```
2021-03-11T16:51:34.006Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8080"}
2021-03-11T16:51:34.007Z	INFO	setup	starting manager
2021-03-11T16:51:34.007Z	INFO	controller-runtime.manager	starting metrics server	{"path": "/metrics"}
2021-03-11T16:51:34.007Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "notebook", "source": "kind source: /, Kind="}
2021-03-11T16:51:34.108Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "notebook", "source": "kind source: /, Kind="}
2021-03-11T16:51:34.209Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "notebook", "source": "kind source: /, Kind="}
2021-03-11T16:51:34.310Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "notebook", "source": "kind source: networking.istio.io/v1alpha3, Kind=VirtualService"}
2021-03-11T16:51:34.411Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "notebook", "source": "kind source: /, Kind="}
2021-03-11T16:51:34.612Z	INFO	controller-runtime.controller	Starting EventSource	{"controller": "notebook", "source": "kind source: /, Kind="}
2021-03-11T16:51:34.714Z	INFO	controller-runtime.controller	Starting Controller	{"controller": "notebook"}
2021-03-11T16:51:34.714Z	INFO	controller-runtime.controller	Starting workers	{"controller": "notebook", "worker count": 1}
2021-03-11T16:51:34.714Z	INFO	controllers.Notebook	Updating StatefulSet	{"notebook": "admin/test", "namespace": "admin", "name": "test"}
2021-03-11T16:51:34.750Z	INFO	controllers.Notebook	Updating Notebook CR state: 	{"notebook": "admin/test", "namespace": "admin", "name": "test"}
2021-03-11T16:51:34.779Z	INFO	culler	Culling of idle Pods is Disabled. To enable it set the ENV Var 'ENABLE_CULLING=true'
2021-03-11T16:51:34.779Z	INFO	controllers.Notebook	Updating StatefulSet	{"notebook": "admin/test-vs-code", "namespace": "admin", "name": "test-vs-code"}
E0311 16:51:34.791493       1 runtime.go:78] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 565 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x14ab0e0, 0x20ceae0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/runtime/runtime.go:74 +0xa6
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/runtime/runtime.go:48 +0x89
panic(0x14ab0e0, 0x20ceae0)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.(*Unstructured).GetNamespace(0x0, 0xc00049d7c9, 0x5)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/apis/meta/v1/unstructured/unstructured.go:234 +0x5d
sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.validateOwner(0x1824880, 0xc00041ce00, 0x1824d00, 0x0, 0xc00041ce00, 0x2097501)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/controller/controllerutil/controllerutil.go:147 +0x97
sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.SetControllerReference(0x1824880, 0xc00041ce00, 0x1824d00, 0x0, 0xc0003ff260, 0xc001743d80, 0x0)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/controller/controllerutil/controllerutil.go:64 +0x14f
github.com/kubeflow/kubeflow/components/notebook-controller/controllers.(*NotebookReconciler).reconcileVirtualService(0xc00078d100, 0xc00041ce00, 0xc00004400a, 0x4)
	/workspace/notebook-controller/controllers/notebook_controller.go:498 +0x15a
github.com/kubeflow/kubeflow/components/notebook-controller/controllers.(*NotebookReconciler).Reconcile(0xc00078d100, 0xc00049d7c9, 0x5, 0xc00049d754, 0xc, 0x109c6b0c6, 0xc000708fc0, 0xc001706378, 0xc001706360)
	/workspace/notebook-controller/controllers/notebook_controller.go:184 +0x1a36
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0002f20b0, 0x14f7d60, 0xc00025e740, 0xc00040f100)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:255 +0x166
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0002f20b0, 0xc000645700)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:231 +0xb0
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker(0xc0002f20b0)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:210 +0x2b
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1(0xc000629bd0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/wait/wait.go:152 +0x5f
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc000629bd0, 0x3b9aca00, 0x0, 0x16e5101, 0xc0000461e0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/wait/wait.go:153 +0x105
k8s.io/apimachinery/pkg/util/wait.Until(0xc000629bd0, 0x3b9aca00, 0xc0000461e0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/wait/wait.go:88 +0x4d
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:192 +0x435
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xbb93bd]
goroutine 565 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/runtime/runtime.go:55 +0x10c
panic(0x14ab0e0, 0x20ceae0)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
k8s.io/apimachinery/pkg/apis/meta/v1/unstructured.(*Unstructured).GetNamespace(0x0, 0xc00049d7c9, 0x5)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/apis/meta/v1/unstructured/unstructured.go:234 +0x5d
sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.validateOwner(0x1824880, 0xc00041ce00, 0x1824d00, 0x0, 0xc00041ce00, 0x2097501)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/controller/controllerutil/controllerutil.go:147 +0x97
sigs.k8s.io/controller-runtime/pkg/controller/controllerutil.SetControllerReference(0x1824880, 0xc00041ce00, 0x1824d00, 0x0, 0xc0003ff260, 0xc001743d80, 0x0)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/controller/controllerutil/controllerutil.go:64 +0x14f
github.com/kubeflow/kubeflow/components/notebook-controller/controllers.(*NotebookReconciler).reconcileVirtualService(0xc00078d100, 0xc00041ce00, 0xc00004400a, 0x4)
	/workspace/notebook-controller/controllers/notebook_controller.go:498 +0x15a
github.com/kubeflow/kubeflow/components/notebook-controller/controllers.(*NotebookReconciler).Reconcile(0xc00078d100, 0xc00049d7c9, 0x5, 0xc00049d754, 0xc, 0x109c6b0c6, 0xc000708fc0, 0xc001706378, 0xc001706360)
	/workspace/notebook-controller/controllers/notebook_controller.go:184 +0x1a36
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0002f20b0, 0x14f7d60, 0xc00025e740, 0xc00040f100)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:255 +0x166
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0002f20b0, 0xc000645700)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:231 +0xb0
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker(0xc0002f20b0)
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:210 +0x2b
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1(0xc000629bd0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/wait/wait.go:152 +0x5f
k8s.io/apimachinery/pkg/util/wait.JitterUntil(0xc000629bd0, 0x3b9aca00, 0x0, 0x16e5101, 0xc0000461e0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/wait/wait.go:153 +0x105
k8s.io/apimachinery/pkg/util/wait.Until(0xc000629bd0, 0x3b9aca00, 0xc0000461e0)
	/go/pkg/mod/k8s.io/apimachinery@v0.17.9/pkg/util/wait/wait.go:88 +0x4d
created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func1
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.11/pkg/internal/controller/controller.go:192 +0x435
```

DavidSpek:
@googlebot I consent

components/notebook-controller/controllers/notebook_controller.go
kimwnasptd: @DavidSpek show your comment below about the error with Unmarshal https://github.com/kubeflow/kubeflow/pull/5660#issuecomment-797378906. Writing it here to keep it in one place.

I'm not sure how you used the Unmarshal function but I believe your could also try and catch the error and not let it panic. Here's a very similar code I've seen https://github.com/kubeflow/katib/blob/master/pkg/new-ui/v1beta1/backend.go#L73

kimwnasptd:
For the google bot I think @thesuperzapper will also need to use the `I consent` command

components/notebook-controller/controllers/notebook_controller.go
yanniszark: @DavidSpek in the example I gave you, I omitted the error handling. Let me give you an example with error handling. This way you can catch the bad case in the controller, log it and emit an event:
https://play.golang.org/p/nxmAh3VaNZr

DavidSpek: @yanniszark Thanks for the example. I just noticed the actual latest changes were not pushed yesterday. I just pushed them and I believe the error handling from you example is included in there. However, the controller crashing is still a problem.

yanniszark: Then we need to find the cause. Can you point to the exact line in your code that crashes?
What have you tried / what do you think is wrong?

thesuperzapper:
@googlebot I consent.

thesuperzapper:
We have pushed changes, and tested it locally, it works as far as we can tell.

I have not included the header blacklist, as I don't believe there is any significant attack vector opened by allowing users to change headers on HTTP connections to Notebook pods they already have access to. (But if you disagree, please rase another PR)

/approve
/lgtm

kimwnasptd:
@thesuperzapper in the future please don't merge PRs without the final ACK of the other involved reviewers.

Regarding the argument of allowing users to set any header, and not only ones that start with `X-` which is what R-Studio expects; I've specifically asked more than once what are the use cases that it limits and why we shouldn't do it https://github.com/kubeflow/kubeflow/pull/5660#discussion_r592245141 https://github.com/kubeflow/kubeflow/pull/5660#discussion_r592294749 but never got an answer.

We should not be giving users more options than what is needed. If in the future we find ourselves limited from this constraint then we can update our code.

Please don't merge code again without the final ACK of the involved reviewers

PR #5663, opened 2021-03-05
First iteration and common utils for Jupyter web app CD

kimwnasptd:
This is part of our effort to use the Optional-test infra for 1.3 going forward #5482 

This PR introduces the following assumptions:
1. The produced Argo Workflow will **not** try to push the produced image to a registry [ `--no-push` Kaniko arg] if the code is not run in the upstream AWS cluster [ `LOCAL_TESTING=True` env var ]. This will make it easier for more people to run the code in their Kubernetes cluster with Argo installed
2. We expect some workflows to not have an `exit dag`. This is the case for now at least, where the CD only builds our public images and pushes them to the registry
3. We will be using `gcr.io/kaniko-project/executor:v1.5.0`. This is a newer version from what's used from [KFServing](https://github.com/kubeflow/kfserving/blob/ec9e8ca569ea173de3717a41130e37cfb756ac8d/test/workflows/components/workflows.libsonnet#L47)
4. The produced images will use the SHA of the PR as the tag, which is what [KFServing does](https://github.com/kubeflow/kfserving/blob/ec9e8ca569ea173de3717a41130e37cfb756ac8d/test/workflows/components/workflows.libsonnet#L348) 

@DavidSpek in the end while working on trying out Kaniko I ended up writing code for it. So let's use this PR to test the first iteration only with JWA and then once we ensure it works lets use your PR #5640 to add workflows for the other components as well.

@PatrickXYS I have two questions:
1. Can we use the `public.ecr.aws/j1r0q0g6/jupyter-web-app` yet, or do you still need to do some setup?
2. What kind of credentials do I need to inject to the workflow in order for the Kaniko container to have permissions to push to the public AWS ECR? Is it just the [docker-config](https://github.com/kubeflow/kfserving/blob/ec9e8ca569ea173de3717a41130e37cfb756ac8d/test/workflows/components/workflows.libsonnet#L197-L202) [ and if yes, where should I mount it ] ?

/cc @PatrickXYS

### DIFF
--- a/py/kubeflow/kubeflow/cd/jwa.py
+++ b/py/kubeflow/kubeflow/cd/jwa.py
@@ -1,0 +1,48 @@
+""""Argo Workflow for building Jupyter web app's OCI image using Kaniko"""
+import os
+
+from kubeflow.kubeflow import ci
+from kubeflow.testing import argo_build_util
+
+TAG = os.getenv("PULL_BASE_SHA", "kaniko-local-test")
+AWS_REGISTRY = "public.ecr.aws/j1r0q0g6/jupyter-web-app"
+
+
+class Builder(ci.workflow_utils.ArgoTestBuilder):
+    def __init__(self, name=None, namespace=None, bucket=None,
+                 test_target_name=None, **kwargs):
+        super().__init__(name=name, namespace=namespace, bucket=bucket,
+                         test_target_name=test_target_name, **kwargs)
+
+    def build(self):
+        """Build the Argo workflow graph"""
+        workflow = self.build_init_workflow(exit_dag=False)
+        task_template = self.build_task_template()
+
+        # Build JWA using Kaniko
+        dockerfile = ("%s/components/crud-web-apps"
+                      "/jupyter/Dockerfile") % self.src_dir
+        context = "dir://%s/components/crud-web-apps" % self.src_dir
+        destination = "%s:%s" % (AWS_REGISTRY, TAG)
+
+        kaniko_task = self.create_kaniko_task(task_template, dockerfile,
+                                              context, destination)
+        argo_build_util.add_task_to_dag(workflow,
+                                        ci.workflow_utils.E2E_DAG_NAME,
+                                        kaniko_task, [self.mkdir_task_name])
+
+        # Set the labels on all templates
+        workflow = argo_build_util.set_task_template_labels(workflow)
+
+        return workflow
+
+
+def create_workflow(name=None, namespace=None, bucket=None, **kwargs):
+    """
+    Args:
+        name: Name to give to the workflow. This can also be used to name
+              things associated with the workflow.
+    """
+    builder = Builder(name=name, namespace=namespace, bucket=bucket, **kwargs)
+
+    return builder.build()
--- a/py/kubeflow/kubeflow/cd/jwa.py
+++ b/py/kubeflow/kubeflow/cd/jwa.py
@@ -1,10 +1,7 @@
 """"Argo Workflow for building Jupyter web app's OCI image using Kaniko"""
-import os
-
 from kubeflow.kubeflow import ci
 from kubeflow.testing import argo_build_util
 
-TAG = os.getenv("PULL_BASE_SHA", "kaniko-local-test")
 AWS_REGISTRY = "public.ecr.aws/j1r0q0g6/jupyter-web-app"
 
 
@@ -23,7 +20,7 @@ class Builder(ci.workflow_utils.ArgoTestBuilder):
         dockerfile = ("%s/components/crud-web-apps"
                       "/jupyter/Dockerfile") % self.src_dir
         context = "dir://%s/components/crud-web-apps" % self.src_dir
-        destination = "%s:%s" % (AWS_REGISTRY, TAG)
+        destination = AWS_REGISTRY
 
         kaniko_task = self.create_kaniko_task(task_template, dockerfile,
                                               context, destination)
--- a/py/kubeflow/kubeflow/cd/jwa.py
+++ b/py/kubeflow/kubeflow/cd/jwa.py
@@ -1,8 +1,7 @@
 """"Argo Workflow for building Jupyter web app's OCI image using Kaniko"""
 from kubeflow.kubeflow import ci
+from kubeflow.kubeflow.cd import config
 from kubeflow.testing import argo_build_util
-
-AWS_REGISTRY = "public.ecr.aws/j1r0q0g6/jupyter-web-app"
 
 
 class Builder(ci.workflow_utils.ArgoTestBuilder):
@@ -20,7 +19,7 @@ class Builder(ci.workflow_utils.ArgoTestBuilder):
         dockerfile = ("%s/components/crud-web-apps"
                       "/jupyter/Dockerfile") % self.src_dir
         context = "dir://%s/components/crud-web-apps" % self.src_dir
-        destination = AWS_REGISTRY
+        destination = config.JUPYTER_WEB_APP_IMAGE
 
         kaniko_task = self.create_kaniko_task(task_template, dockerfile,
                                               context, destination)
--- a/py/kubeflow/kubeflow/ci/workflow_utils.py
+++ b/py/kubeflow/kubeflow/ci/workflow_utils.py
@@ -183,6 +183,30 @@ class ArgoTestBuilder:
 
         return task_template
 
+    def create_kaniko_task(self, task_template, dockerfile, context,
+                           destination, no_push=False):
+        """
+        A task for building images inside a cluster container using Kaniko.
+        If we are testing the workflow locally then we won't be pushing images
+        to any registries. This will make it easier for people to try out and
+        extend the code.
+        """
+        kaniko = argo_build_util.deep_copy(task_template)
+
+        kaniko["name"] = "kaniko-build-push"
+        kaniko["container"]["image"] = "gcr.io/kaniko-project/executor:v1.5.0"
+        kaniko["container"]["command"] = ["/kaniko/executor"]
+        kaniko["container"]["args"] = ["--dockerfile=%s" % dockerfile,
+                                       "--context=%s" % context,
+                                       "--destination=%s" % destination]
+
+        # don't push the image to a registry if trying out the produced
+        # Argo Workflow yaml locally
+        if LOCAL_TESTING == "True" or no_push:
+            kaniko["container"]["args"].append("--no-push")
+
+        return kaniko
+
     def _create_checkout_task(self, task_template):
         """Checkout the kubeflow/testing and kubeflow/kubeflow code"""
         main_repo = argo_build_util.get_repo_from_prow_env()
--- a/py/kubeflow/kubeflow/ci/workflow_utils.py
+++ b/py/kubeflow/kubeflow/ci/workflow_utils.py
@@ -201,6 +201,12 @@ class ArgoTestBuilder:
         """
         kaniko = argo_build_util.deep_copy(task_template)
 
+        # append the tag base-commit[0:7]
+        if ":" not in destination:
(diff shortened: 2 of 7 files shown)


kimwnasptd:
Also while working on the PR I made the workflow to be a presubmit job so that I can easily check the produced workflows. After we've confirmed the PR works as expected I'll move the workflow to be a postsubmit job, before we merge.

Looking at the produced Argo Workflow everything goes as expected:
1. The image's tag is the SHA of the PR
2. The Kaniko container fails because it doesn't has enough permissions

https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-jwa-build-5663-64a400b-3968-932b?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-jwa-build-5663-64a400b-3968-932b-4173484142&sidePanel=logs%3Akubeflow-kubeflow-presubmit-jwa-build-5663-64a400b-3968-932b-4173484142%3Amain

py/kubeflow/kubeflow/cd/jwa.py
```diff
@@ -0,0 +1,48 @@
+""""Argo Workflow for building Jupyter web app's OCI image using Kaniko"""
+import os
+
+from kubeflow.kubeflow import ci
+from kubeflow.testing import argo_build_util
+
+TAG = os.getenv("PULL_BASE_SHA", "kaniko-local-test")
```
DavidSpek, line 7: I think a better naming scheme would be: <branch-name>-<commit-id>. The commit ID is the first 7 characters of the SHA1 hash.

DavidSpek: Also, by using the branch-name, fixes cherry picked fixes for specific Kubeflow release version can easily be distinguished from the containers built from the master branch.

PatrickXYS:
> Can we use the public.ecr.aws/j1r0q0g6/jupyter-web-app yet, or do you still need to do some setup?

This is done 

> What kind of credentials do I need to inject to the workflow in order for the Kaniko container to have permissions to push to the public AWS ECR? Is it just the docker-config [ and if yes, where should I mount it ] ?

You can follow kaniko official doc https://github.com/GoogleContainerTools/kaniko#pushing-to-amazon-ecr, we've already set up credentials and configs as configmap and secret stored in the cluster. So all you have to do is to specify those configurations in your argo workflow.

DavidSpek:
It looks like the volume mount:
```
- name: aws-secret
  mountPath: /root/.aws/
```
Is already in the pod spec, so you'd think that it would be able to push to the registry. I've only looked quickly, but I don't see the accompanying `volumes:` mapping.

py/kubeflow/kubeflow/cd/jwa.py
kimwnasptd: > The commit ID is the first 7 characters of the SHA1 hash.

Could you elaborate on this one? AFAIK the commit ID is the SHA1 of the important parts of the commit (date, log message etc). Could you point me to some documentation about the 7 chars?

> I think a better naming scheme would be:
> Also, by using the branch-name, fixes cherry picked fixes for specific Kubeflow release version can easily be distinguished from the containers built from the master branch.

Yeap, let's go with `${PULL_BASE_REF}-${PULL_BASE_SHA}[0:7]`. This was also the case with the previous images https://console.cloud.google.com/gcr/images/kubeflow-images-public/GLOBAL/jupyter-web-app?gcrImageListsize=30

kimwnasptd:
@PatrickXYS 
> This is done

That's awesome, thanks!

> You can follow kaniko official doc https://github.com/GoogleContainerTools/kaniko#pushing-to-amazon-ecr, we've already set up credentials and configs as configmap and secret stored in the cluster. So all you have to do is to specify those configurations in your argo workflow.

Thanks for the link! Once you posted it I remembered I had glanced over that section in the README but had completely forgotten about it. I'll add the `docker-config` and the `aws-secret`.

@DavidSpek 
> It looks like the volume mount:
> 
>   name: aws-secret
>   mountPath: /root/.aws/
> 
> Is already in the pod spec, so you'd think that it would be able to push to the registry. I've only looked quickly, but I don't see the accompanying volumes: mapping.

@DavidSpek I can see it in prow's logs http://prow.kubeflow-testing.com/view/s3/aws-kubeflow-jenkins/pr-logs/pull/kubeflow_kubeflow/5663/kubeflow-kubeflow-presubmit/1367813324770643968#1:build-log.txt%3A318. Did you had something else on your mind?

DavidSpek:
@kimwnasptd Yeah I figured that wouldn't be the case, I was looking in the wrong place. Isn't the aws secret already mounted then?

kimwnasptd:
Yeap, so we only need to add the `docker-config` volume/volumeMount  so that the docker process will have permissions to push to public ECR [ @PatrickXYS correct me if anything I say is off ]

DavidSpek:
@kimwnasptd Yes I think that is all.

kimwnasptd:
The workflow successfully build the image and pushed it. We have the first public image with the new Optional-test infra folks! :partying_face: 

This is supper exciting!

https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-jwa-build-5663-02f77a1-9840-696f?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-jwa-build-5663-02f77a1-9840-696f-2146802049

kimwnasptd:
I can also see it in the AWS ECR Gallery https://gallery.ecr.aws/j1r0q0g6/jupyter-web-app.

Changing the tag per @DavidSpek's comments, moving the workflow to postsubmits and we should be ready to go!

DavidSpek:
@kimwnasptd Once this PR has been merged I will rework my PR for all the (other) images using the code you created here.

DavidSpek:
@PatrickXYS Related to this PR, is there a way to setup image scanning with some form of GitHub integration (a badge, an issue, anything that can display the results really)? https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-scanning.html

py/kubeflow/kubeflow/cd/jwa.py
```diff
@@ -0,0 +1,48 @@
...
+AWS_REGISTRY = "public.ecr.aws/j1r0q0g6/jupyter-web-app"
```
PatrickXYS, line 5: NIT: since WG-notebooks have multiple ECR registries (7 now, maybe more in the future), I would recommend you maintain a static configmap or mapping in other python files.

E.g, py/kubeflow/kubeflow/cd/config.py

PatrickXYS:
> Related to this PR, is there a way to setup image scanning with some form of GitHub integration (a badge, an issue, anything that can display the results really)

This is customer-faced bonus feature, currently ECR don't support that nor our test-infra.

The badge / issue stuff, normally they're supported by official product, since test-infra is self-hosted, we do not plan to add it on server side.

Feel free to do so in argo workflow, it won't be hard as long as you get familiar with those Github APIs it should be fine.

Note: ECR image scanning is designed for security check

py/kubeflow/kubeflow/ci/workflow_utils.py
```diff
@@ -201,6 +201,12 @@ def create_kaniko_task(self, task_template, dockerfile, context,
         """
         kaniko = argo_build_util.deep_copy(task_template)
 
+        # append the tag base-commit[0:7]
+        if ":" not in destination:
+            sha = os.getenv("PULL_BASE_SHA", "12341234kanikotest")
+            base = os.getenv("PULL_BASE_REF", "master")
+            destination += ":%s-%s" % (base, sha[0:8])
```
DavidSpek, line 208: I think you made a typo here. I believe it should be sha[0:7] and not sha[0:8]. The comment does state sha[0:7].

py/kubeflow/kubeflow/cd/jwa.py
kimwnasptd: That's a good idea, it will help us have all the names organized in one place.
Adding it

py/kubeflow/kubeflow/ci/workflow_utils.py
kimwnasptd: I didn't calculate correctly in the message above, I'll comment there as well.

I had in mind to have the same length it previously had https://console.cloud.google.com/gcr/images/kubeflow-images-public/GLOBAL/jupyter-web-app?gcrImageListsize=30

DavidSpek: I was going with the first 7 characters as this is the Commit ID you see in GitHub. This way, it is easy for users to track down what commit and image originates from (and from which branch). 
image

kimwnasptd: Resolving this one and continuing the discussion on the above conversation

py/kubeflow/kubeflow/cd/jwa.py
kimwnasptd: In the end I had to do `base-commit[0:8]` in order for the parsed length to be the same with what it was for the previous images.

@DavidSpek the images now have described format. Should I go on and resolve this conversation?

py/kubeflow/kubeflow/cd/jwa.py
kimwnasptd: @PatrickXYS I added another commit that introduces this `config.py` file.
PTAL and if you are OK I can resolve this conversation

PatrickXYS: LGTM feel free to resolve

PatrickXYS (approved):
/lgtm
/approve
/hold

Feel free to unhold if you feel it's good to go

py/kubeflow/kubeflow/cd/jwa.py
DavidSpek: @kimwnasptd Does the parsed length matter for something? 
Documentation on Git Commit: https://stackoverflow.com/questions/43763896/whats-the-difference-between-commit-id-and-sha1-hash-in-git

DavidSpek:
@kimwnasptd Could you rename commit b53cc02. It should be postsubmit job :). For later reference it might otherwise be confusing.

py/kubeflow/kubeflow/cd/jwa.py
kimwnasptd: Can't think of something, we should be good to go

kimwnasptd:
@DavidSpek I pushed two commits that moved the workflow to be a postsubmit job. I had explicitly made it a presubmit in the beginning so we could check/run the produced Argo Workflow while I was adding extra commits.

But now that we can confirm it works I made it to be a postsubmit job, so we should be all set

py/kubeflow/kubeflow/cd/jwa.py
DavidSpek: Alright, this can be resolved. Please see my comment below though.

DavidSpek:
@kimwnasptd Cool!

kimwnasptd:
@Bobgy, at your convenience, could you merge this PR? We've also added a postsubmit workflow to `prow_config.yaml`, for building the Jupyter web app image after a PR is merged.

We are super excited for this

DavidSpek:
@kimwnasptd First thing tomorrow morning I'll rework my PR following this one for all the other components (it's getting a bit late now). This way, only 1 more PR will need to be merged by @Bobgy to get all the image building up and running.

DavidSpek:
@kimwnasptd Did the force push go through? I still see the mistake in the commit hash. If it is being difficult it isn't a blocker of course. 

/lgtm

DavidSpek:
@kimwnasptd I just noticed 2 things. First thing, the Dockerfile location is the location within the build context according to the Kaniko docs (though it doesn't seem to matter). But more importantly, when building the JWA with the make file, it copies the `Dockerfile.dockerignore` to the context directory (this is also done for the TWA). However, this isn't done with the current build process so I'm not sure if that will cause a problem or larger images.

PR #5679, opened 2021-03-10
Expand the email regex to allow more TLDs

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5004

Adds the changes proposed in https://github.com/kubeflow/kubeflow/pull/5506. 

/cc @yanniszark

### DIFF
--- a/components/centraldashboard/app/api_workgroup.ts
+++ b/components/centraldashboard/app/api_workgroup.ts
@@ -8,8 +8,8 @@ import {
     ERRORS,
 } from './api';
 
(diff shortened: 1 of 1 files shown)


DavidSpek:
Ping @kimwnasptd @StefanoFioravanzo Could one of you quickly take a look at this?

DavidSpek:
Just tested the changes in this PR manually in an actual deployment (previously only in a regex tester) and it works as expected. Before this change TLDs longer than 3 characters were not valid. After changing the regex longer TLDs can be used.

image

PR #5681, opened 2021-03-10
feat(jupyter): Support hiding image registry/tag

saffaalvi:
Closes https://github.com/StatCan/kubeflow/issues/51

In the notebook server creation view, the fully qualified name of available images is currently displayed as:
- gcr.io/kubeflow-images-public/tensorflow-1.5.1-notebook-cpu:v0.5.0

We implemented an optional feature to hide the registry and tag in the image dropdown list as the names can get unwieldly when the registry domain is much longer than, for example, `gcr.io`. Tags can also be the same (e.g. for organizations that use full commit SHAs for tags). 

This is an optional feature for a more approachable and understandable list, with the new images displayed in the form:
- kubeflow-images-public/tensorflow-1.5.1-notebook-cpu

This PR adds two new configuration options to the `crud-web-apps/jupyter-web-app` supported `spawnerFormDefaults`:
- hideRegistry
- hideTag

If these values are set as `true`, the notebook spawner UI will hide these portions of the image tags in the list of available images. Options were set to "hide" rather than "show" for backwards compatibility: the absence of the values will default to false, which results in the original behaviour of showing the entire image tag.

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
+++ b/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
@@ -49,6 +49,8 @@ spawnerFormDefaults:
   allowCustomImage: true
   # If true, users can input custom images
   # If false, users can only select from the images in this config
+  hideRegistry: false
+  hideTag: false
   imagePullPolicy:
     # Supported values: Always, IfNotPresent, Never
     value: IfNotPresent
--- a/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
+++ b/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
@@ -25,8 +25,6 @@ spawnerFormDefaults:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda-full:master-1831e436
-    hideRegistry: true
-    hideTag: false
   imageGroupOne:
     # The container Image for the user's Group One Server
     # The annotation `notebooks.kubeflow.org/http-rewrite-uri: /`
@@ -48,6 +46,9 @@ spawnerFormDefaults:
     # The list of available standard container Images
     options:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/rstudio-tidyverse:master-164fa2ea
+  # If true, hide registry and/or tag name in the image selection dropdown
+  hideRegistry: true
+  hideTag: false
   allowCustomImage: true
   # If true, users can input custom images
   # If false, users can only select from the images in this config
--- a/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
+++ b/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
@@ -25,7 +25,7 @@ spawnerFormDefaults:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda-full:master-1831e436
-    hideRegistry: false
+    hideRegistry: true
     hideTag: false
   imageGroupOne:
     # The container Image for the user's Group One Server
--- a/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
+++ b/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
@@ -25,6 +25,8 @@ spawnerFormDefaults:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda-full:master-1831e436
+    hideRegistry: false
+    hideTag: false
   imageGroupOne:
     # The container Image for the user's Group One Server
     # The annotation `notebooks.kubeflow.org/http-rewrite-uri: /`
@@ -49,8 +51,6 @@ spawnerFormDefaults:
   allowCustomImage: true
   # If true, users can input custom images
   # If false, users can only select from the images in this config
-  hideRegistry: false
-  hideTag: false
   imagePullPolicy:
     # Supported values: Always, IfNotPresent, Never
     value: IfNotPresent
--- a/components/crud-web-apps/jupyter/frontend/package-lock.json
+++ b/components/crud-web-apps/jupyter/frontend/package-lock.json
@@ -2535,9 +2535,9 @@
       "integrity": "sha512-I99sngh224D0M7XgW1s120zxCt3VYQ3IQsuw3P3jbq5GG4yc79+ZjyKznyOGIQrflfylLgcfekeZW/vk0yng6A=="
     },
     "@types/underscore": {
-      "version": "1.10.24",
-      "resolved": "https://registry.npmjs.org/@types/underscore/-/underscore-1.10.24.tgz",
-      "integrity": "sha512-T3NQD8hXNW2sRsSbLNjF/aBo18MyJlbw0lSpQHB/eZZtScPdexN4HSa8cByYwTw9Wy7KuOFr81mlDQcQQaZ79w=="
+      "version": "1.11.0",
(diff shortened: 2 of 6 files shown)


DavidSpek:
@saffaalvi Awesome feature and thanks for the contribution. Is StatCan already using these changes in production? Also, can you sign the CLA so this can move forward (hopefully) quickly.

blairdrummond:
We sure are @DavidSpek !

Screenshot from 2021-03-11 12-38-48

https://github.com/StatCan/kubeflow/commit/17d85bd

DavidSpek:
@blairdrummond Wesome, thanks for confirming! Looking through the PR, do I assume correctly that the drop down list in the notebook spawner UI also shows these shortened image names?

@kimwnasptd I think this is a PR we should try and have available in 1.3 as well (if at all possible, depending how the rest of the PRs go of course). Given that this already deployed in production by StatCan makes me think it might be possible. Me and Mathew were already discussing such a feature as the image registry names will be quite long (for example: `public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-full:<some_tag>`). We were initially thinking of shortening it to `jupyter-pytorch-full:<some_tag>`, but every bit helps. Potentially, this PR could be slightly extended to make the prefix it can remove a bit more configurable.

saffaalvi:
@DavidSpek Thank you for your interest in this feature. I am currently in the process of getting the CLA signed so hopefully that will be done as soon as possible. And yes, your assumption is correct that the image names drop down list in the notebook spawner UI shows these shortened image names. 

Screen Shot 2021-03-17 at 3 19 54 PM

saffaalvi:
> @kimwnasptd I think this is a PR we should try and have available in 1.3 as well (if at all possible, depending how the rest of the PRs go of course). Given that this already deployed in production by StatCan makes me think it might be possible. Me and Mathew were already discussing such a feature as the image registry names will be quite long (for example: `public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-full:<some_tag>`). We were initially thinking of shortening it to `jupyter-pytorch-full:<some_tag>`, but every bit helps. Potentially, this PR could be slightly extended to make the prefix it can remove a bit more configurable.

@DavidSpek  It’s definitely possible to extend this PR to shorten the image name even further. In terms of making the prefix it can remove a bit more configurable, in this PR, we use the 2 attributes `hideRegistry` and `hideTag` in the `spawner-config.yaml` file to remove those parts from the image name. Would it be better to add or substitute another attribute to remove everything before the `jupyter-pytorch-full:<some_tag>` to shorten the name further? Or would you prefer another approach to configure the names (instead of attributes in the config file)?

kimwnasptd:
@saffaalvi @blairdrummond thank you for your effort!

Also apologies for the very late reply on this, we are trying to wrap up the efforts for 1.3. I'll  try to start allocating some cycles on this from tomorrow. At a first glance I really like this fix as it really simplifies what the user sees. I would even suggest to have the `hideRegistry` option set to `true` by default.

I'll also leave more concrete review comments once I take a more deep dive into the code and run it locally.

/ok-to-test

saffaalvi:
@googlebot I signed it

saffaalvi:
@DavidSpek @kimwnasptd I signed the CLA, is there anything else needed from my end to move along with testing this PR?

DavidSpek:
@saffaalvi Sorry for the late response. It's been very busy with the 1.3 release the past while. As soon as that has settled I'll take a closer looks at your PRs.

thesuperzapper:
@saffaalvi can you rebase to fix conflicts? (and please do testing after the rebase)

DavidSpek:
@wg102 I thought you might be the one that is now taking care of this, so I thought I'd ping you.

JessicaBarh:
@googlebot I fixed it.

saffaalvi:
@googlebot I consent.

JessicaBarh:
@googlebot I consent.

DavidSpek:
@wg102 Here is the log output why the CI failed:

```
W
> frontend@0.0.0 format:check /mnt/test-data-volume/kubeflow-kubeflow-presubmit-jwa-tests-5681-02ccc72-5408-c652/src/kubeflow/kubeflow/compone
nts/crud-web-apps/jupyter/frontend
> prettier --check 'src/**/*.{js,ts,html,scss,css}' || node scripts/check-format-error.js
Checking formatting...
[warn] src/app/pages/form/form-default/form-image/form-image.component.ts
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
-----------------------------------------------
Please run `npm run format:write` to format your code.
-----------------------------------------------
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! frontend@0.0.0 format:check: `prettier --check 'src/**/*.{js,ts,html,scss,css}' || node scripts/check-format-error.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the frontend@0.0.0 format:check script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2021-06-03T22_26_41_572Z-debug.log
```

DavidSpek:
@JessicaBarh Thanks for pushing that fix, the CI tests now all passed.

DavidSpek:
Just tested and confirmed everything works as expected.
/cc @thesuperzapper
/lgtm

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
```diff
@@ -25,6 +25,8 @@ spawnerFormDefaults:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda-full:master-1831e436
+    hideRegistry: false
```
kimwnasptd: We will also need to apply these changes to the manifests files, but lets do this after the refactored config file

kimwnasptd (commented):
@saffaalvi @wg102 @JessicaBarh adding parameters to the ConfigMap for handling the behavior is the way to go, but these fields are added in a semantically wrong place. These settings should not belong to the `spawnerFormDefaults` section of the ConfigMap, since these fields affect also the main page of the app and not just the form.

Although until this point we didn't need global settings and thus the backend only exposed the `spawnerFormDefaults` section of the form. So it's totally understandable that you added these fields there, but lets try and put them in the proper place and have a semantically correct design.

I'll propose some concrete steps to follow so that we can place these fields outside the `spawnerFormDefaults` section. It should not be too involved.

DavidSpek:
@kimwnasptd This PR only changes the spawner UI and not the Notebook index page. The index page already shortens the image URI to the last `/`. Also, only files in the forms section of the frontend are changed in this PR.

kimwnasptd:
Ah, you are correct. But my requested changes remain, we should not place these fields under the `spawnerFormDefaults` section of the ConfigMap since:
1. The `spawnerFormDefaults` section is only used for handling the values of the form fields and what parts of the form the user should be able to interact with
2. It's not semantically correct to place these changes into the `spawnerFormDefaults.image` part of the ConfigMap since that part should only affect the form default for that Notebook type

DavidSpek:
I agree, it should be placed in the same place where we moved the `allowCustomImage` to when we added the multiple image groups. Which would be [here](https://github.com/kubeflow/kubeflow/pull/5681/files#diff-fb292765cd2e7795a7fd8793c0d1f7b465c38e9ec6757c340be0d6732c0bd5c2L49)

JessicaBarh:
@DavidSpek @kimwnasptd thanks, I will make the changes!

DavidSpek:
@JessicaBarh Thanks for fixing this so quickly!

/lgtm

kimwnasptd:
Looking at the PR again and re-evaluating my initial expectations that this feature was changing the main page. @JessicaBarh  I would expect the app to show the image names uniformly accross different pages. 

I'm fine with having a mechanism to define what to prune from the image (registry, tag) but we should be doing the same thing for both the main page and the form page. 

To summarize what I have in mind, the app should:
1. Use the same mechanism for showing image names accross all of its pages
2. Show a tooltip wherever the user hovers on top of a pruned image, so that they can always see the full value

Due to the following reasoning we will need to add the parameters to the root of the ConfigMap [not inside the `spawnerFormDefaults`] since these will need to be global settings. I'll provide more specific details on how to achieve this in a follow up message

kimwnasptd:
I evaluated the changes I had in mind. @JessicaBarh in the end it will be a little bit involved since we will also need to make some changes to the backend. 

I'm OK with adding the params under the `spawnerFormDefaults` for now and once we have e2e testing we can safely make a PR with my, more involved, proposed changes. 

My last review comment is to add a material tooltip to entry of the mat-select so that when the user hovers on top they can see the full image name, even if it is pruned.

JessicaBarh:
> I evaluated the changes I had in mind. @JessicaBarh in the end it will be a little bit involved since we will also need to make some changes to the backend.
> 
> I'm OK with adding the params under the `spawnerFormDefaults` for now and once we have e2e testing we can safely make a PR with my, more involved, proposed changes.
> 
> My last review comment is to add a material tooltip to entry of the mat-select so that when the user hovers on top they can see the full image name, even if it is pruned.

Hi @kimwnasptd you're right I think a tooltip is a great idea so the full name can still be visible. Just added it to my last commit, is this what you had in mind? 

image

kimwnasptd:
@JessicaBarh yeap, this is what I meant. Thanks for adding it!

I think this PR is good to go. Thank you for your contribution and time on this @JessicaBarh!
/lgtm
/approve

PR #5685, opened 2021-03-11
Centraldashboard: Fix active menu item for path-based URLs

StefanoFioravanzo:
This PR fixes two bugs resulting in Centraldashboard not showing an active menu item:

## 1. No active menu item upon refresh

When refreshing the page, the main page fetches menu links by calling the `/api/dashboard-links` endpoint. The callback of this endpoint fires *after* the callback for URL changes, which in turn triggers the activation of the selected menu item. This results in no active menu link. The first commit fixes this by adding an observer to the list of menu items and forcing the activation of one of them, based on the URL.

## 2. Missing support for path-based URLs

We were taking into consideration Apps that do hash-based routing, so URLs with path-based routing would not be matched against the correct menu item. The second commit handles this.

---

This PR also introduces some unit tests to make sure that the selection of the active menu item happens as expected.

/assign @StefanoFioravanzo 
/cc @kimwnasptd 
/cc @elikatsis 
/cc @tasos-ale

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -73,6 +73,7 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
             menuLinks: {
                 type: Array,
                 value: [],
+                observer: '_menuLinksChanged',
             },
             externalLinks: {
                 type: Array,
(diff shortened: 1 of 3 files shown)


kimwnasptd:
Tested this and the correct item is selected once I refresh the central dashboard. The tests are also passing.

Thanks for the PR @StefanoFioravanzo!

/lgtm
/approve

Issue #5553, opened 2021-01-25
Deprecate the old Jupyter web app


kimwnasptd:
/kind feature

**Why you need this feature:**
Right now we only support and develop the jupyter web app that lives under `components/crud-web-apps`. The `components/jupyter-web-app` is not being developed anymore and it will be completely phased out after the 1.3 release, in which we will be using the crud-web-apps one.

The situation right now can be confusing since we maintain two distinct versions of the app, while commit to only one.

**Describe the solution you'd like:**
We should gradually deprecate this `components/jupyter-web-app`. All of it's features, and even more, are already included in the jupyter-web-app. The deprecation should happen in two phases:
1. we edit the README of the `components/jupyter-web-app` to clearly state that the app is no longer maintained and will be phased out/removed in the future
2. purge it's files after the 1.3 release, in which we will have completely switched to the crud-web-apps one

cc @kubeflow/wg-notebooks-leads

DavidSpek:
@kimwnasptd Maybe it is an idea to put the code in a `deprecated` folder during (and maybe for a while after) the transition. Pipelines as a similar folder for their components: https://github.com/kubeflow/pipelines/tree/master/components/deprecated.

DavidSpek:
We should probably come up with a plan when would be the correct time to remove the JWA from the deprecated folder. Not that I think the time is now, but it might be something that stays there indefinitely and gets forgotten if there is no plan.

kimwnasptd:
The plan was to remove this after 1.3 so I believe it's a good thing to do this now, at the beginning of the new cycle.

I've pushed a PR for this #5959

PR #5686, opened 2021-03-11
Parameterize Jupyter route handling

knkski:
Allows setting the prefix under which Jupyter mounts its routes. As part of this, also explicitly responds with index.html to / and /new, instead of any 404's.

### DIFF
--- a/components/jupyter-web-app/backend/kubeflow_jupyter/default/app.py
+++ b/components/jupyter-web-app/backend/kubeflow_jupyter/default/app.py
@@ -1,9 +1,8 @@
-from flask import Flask, request, jsonify, send_from_directory
-from ..common.base_app import app as base
-from ..common import utils, api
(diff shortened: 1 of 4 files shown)


kimwnasptd:
@knkski this web app is deprecated and we've moved to the one under `/components/crud-web-apps/jupyter` which is utilizing a lot of common code between our web apps. That app has an [APP_PREFIX](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/backend/entrypoint.py#L28) env var that can control the prefix of the app.

We have an issue about deprecating the app, but I never moved forward with removing the old code https://github.com/kubeflow/kubeflow/issues/5553. I'll push this effort to remove the old web app, or at least in a different folder, for 1.3 in order to avoid further confusions.

Apologies for the effort and time you put on this.

PR #5689, opened 2021-03-12
Delete old TensorFlow notebook image

DavidSpek:
In accordance to what was discussed and decided upon in the Notebooks Working Group meeting, this PR removes the current TensorFlow notebook image folder. 

https://github.com/kubeflow/kubeflow/issues/5575#issuecomment-797405383

I'm interested to see what happens with the `Languages` percentage of the repo after this change. This is the current state:
image


kimwnasptd:
> I'm interested to see what happens with the Languages percentage of the repo after this change 

Indeed!

It's been a long time, they served their purpose well. On to new images!

/lgtm
/approve

Issue #5615, opened 2021-02-17
Kubflow's Jupyter web app gets binded on 0.0.0.0:500 instead of [::]:5000 in IPv6 kubernetes cluster.


thirdeye-jarvis:
Jupyter Wsgi web app is binded on 0.0.0.0:500 in IPv6 kubernetes cluster due to which I am not able to edit and launch notebooks from Kubeflow dashboard. It should be running on [::]:5000 to establish proper connectivity, 

what's the way of switching this jupyter deployment to Ipv6 ?

Does Kubeflow support IPv6??

![viber_image_2021-02-17_12-13-33](https://user-images.githubusercontent.com/79189189/108167871-61470780-711c-11eb-87a1-94906ca229b1.jpg)

DavidSpek:
@kimwnasptd Does the crud-web-app version of JWA properly support IPv6? If not, is this something that should be included in 1.3?

DavidSpek:
bump

Issue #5605, opened 2021-02-12
Issues with Metadata writer, Central Dashboard, Argo UI  & Ml pipeline pods while deploying Kubeflow in an IPv6 environment.


ayush7007:
/kind bug

I am trying to setup Kubeflow on an IPv6 environment and facing issues with Metadata writer, Central Dashboard, Argo UI & Ml pipeline pods. Not sure whether Kubeflow works fine with Ipv6 as I see no params to switch between Ipv4 and Ipv6. Please suggest if Kubeflow supports IPv6 ? 
Below are the issues I am facing- 

**What steps did you take and what happened:**

1. Metadata writer not able to connect to Metadata store - 
![image](https://user-images.githubusercontent.com/25534418/107885062-08118500-6f1e-11eb-9792-ee28bf67da2a.png)
![image](https://user-images.githubusercontent.com/25534418/107826914-c3fe7300-6dab-11eb-9c72-5566a91ea2be.png)

Metadata writer is trying to connect to GRPC service which is running on  0.0.0.0:8000  in IPv6 environment inside grpc pod( refer screenshot in this post below) and throwing “could not connect to Metadata store”
![image](https://user-images.githubusercontent.com/25534418/108199364-e9d89e80-7142-11eb-9134-e672fd490b38.png)

Grpc deployment logs -
![image](https://user-images.githubusercontent.com/25534418/108199438-0a085d80-7143-11eb-8abf-ed2d2a2ee3f9.png)

2. Kubeflow Central Dashboard is throwing below error where it is splitting the ipv6 link based on colon and assigning incorrect host and port values - 
![image](https://user-images.githubusercontent.com/25534418/108199328-da595580-7142-11eb-9501-238ed684850e.png)

3. Argo UI is running on 0.0.0.0:8001 in IPv6 cluster and is in Not ready state( throwing connection refused) 
![image](https://user-images.githubusercontent.com/25534418/107826644-39b60f00-6dab-11eb-9cc0-4de9103212d7.png)

**What did you expect to happen:**

1. Metadata GRPC is running on 0.0.0.0:8000  whereas it should bind on [::]:8000 in an IPv6 environment, What's the correct way to bind here in case of IPv6?
![image](https://user-images.githubusercontent.com/25534418/107885082-31caac00-6f1e-11eb-8573-b307cca9c5c2.png)

This is hardcoded in Metadata server code -  
[...]

PR #5693, opened 2021-03-12
Fixes in Tensorboard web app

kimwnasptd:
These are some small fixes required for the Tensorboards web app to be able to work under a prefix and inside a container, like jupyter-web-app.

/cc @elikatsis 
/cc @StefanoFioravanzo 

The last remaining part then will be to create the manifests for this app cc @yanniszark

### DIFF
--- a/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/helpers.py
+++ b/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/helpers.py
@@ -24,7 +24,7 @@ def get_prefixed_index_html():
     with open(os.path.join(static_dir, "index.html"), "r") as f:
         index_html = f.read()
         index_prefixed = re.sub(
-            r"\<base href=\".*\"\>", '<base href="%s">' % prefix, index_html,
+            r"\<base href=\".*\".*\>", '<base href="%s">' % prefix, index_html,
         )
 
         return index_prefixed
--- a/components/crud-web-apps/tensorboards/Dockerfile
+++ b/components/crud-web-apps/tensorboards/Dockerfile
@@ -48,4 +48,4 @@ COPY ./tensorboards/backend/Makefile .
 
 COPY --from=frontend /src/dist/ /src/app/static/
 
-ENTRYPOINT gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app
+ENTRYPOINT ["/bin/bash","-c","gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app"]
--- a/components/crud-web-apps/tensorboards/Dockerfile
+++ b/components/crud-web-apps/tensorboards/Dockerfile
@@ -48,4 +48,4 @@ COPY ./tensorboards/backend/Makefile .
 
 COPY --from=frontend /src/dist/ /src/app/static/
 
-ENTRYPOINT make run
+ENTRYPOINT gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app
--- a/components/crud-web-apps/tensorboards/backend/app/utils.py
+++ b/components/crud-web-apps/tensorboards/backend/app/utils.py
@@ -23,14 +23,6 @@ def parse_tensorboard(tensorboard):
     return parsed_tensorboard
 
 
-def getPVCName(pvc):
-    """
-    Return name of given PVC.
-    """
-
-    return pvc.metadata.name
-
-
 def get_tensorboard_dict(namespace, body):
     """
     Create Tensorboard object from request body and format it as a Python dict.
--- a/components/crud-web-apps/tensorboards/backend/app/utils.py
+++ b/components/crud-web-apps/tensorboards/backend/app/utils.py
@@ -1,12 +1,11 @@
-from kubeflow.kubeflow.crud_backend import helpers, status
+from kubeflow.kubeflow.crud_backend import status
 
 
 def parse_tensorboard(tensorboard):
     """
     Process the Tensorboard object and format it as the UI expects it.
     """
-    
-    if tensorboard.get("status", {}).get("readyReplicas", 0) == 1: 
+    if tensorboard.get("status", {}).get("readyReplicas", 0) == 1:
         phase = status.STATUS_PHASE.READY
         message = "The Tensorboard server is ready to connect"
     else:
@@ -17,11 +16,12 @@ def parse_tensorboard(tensorboard):
         "name": tensorboard["metadata"]["name"],
         "namespace": tensorboard["metadata"]["namespace"],
         "logspath": tensorboard["spec"]["logspath"],
-        "age": helpers.get_age(tensorboard),
+        "age": tensorboard["metadata"]["creationTimestamp"],
         "status": status.create_status(phase, message, "")
     }
 
     return parsed_tensorboard
+
 
 def getPVCName(pvc):
     """
--- a/components/crud-web-apps/tensorboards/frontend/package.json
+++ b/components/crud-web-apps/tensorboards/frontend/package.json
@@ -4,7 +4,7 @@
   "scripts": {
     "ng": "ng",
     "start": "npm run copyLibAssets && ng serve --base-href /tensorboard/ --deploy-url /tensorboard/",
-    "build": "npm run copyLibAssets && ng build --prod --base-href /tensorboard/ --deploy-url /tensorboard/static/",
+    "build": "npm run copyLibAssets && ng build --prod --base-href /tensorboards/ --deploy-url static/",
     "build:watch": "npm run copyLibAssets && ng build --watch --deploy-url static/ --outputPath ../backend/app/static/ --outputHashing all",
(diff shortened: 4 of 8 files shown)


components/crud-web-apps/tensorboards/frontend/package.json
```diff
@@ -4,7 +4,7 @@
   "scripts": {
     "ng": "ng",
     "start": "npm run copyLibAssets && ng serve --base-href /tensorboard/ --deploy-url /tensorboard/",
-    "build": "npm run copyLibAssets && ng build --prod --base-href /tensorboard/ --deploy-url /tensorboard/static/",
+    "build": "npm run copyLibAssets && ng build --prod --base-href /tensorboards/ --deploy-url static/",
```
DavidSpek, line 7: In the line above the base-href is `/tensorboard/` instead of `/tensorboards/` which it is here.

components/crud-web-apps/tensorboards/Dockerfile
```diff
@@ -48,4 +48,4 @@ COPY ./tensorboards/backend/Makefile .
 
 COPY --from=frontend /src/dist/ /src/app/static/
 
-ENTRYPOINT make run
+ENTRYPOINT gunicorn -w 3 --bind 0.0.0.0:5000 --access-logfile - entrypoint:app
```
nrchakradhar, line 51: Will this have any incompatibilities for IPv6?

DavidSpek: @nrchakradhar Good point. These same entrypoint is being used by the Jupyter-Web-App. Do you have suggestions how to make this IPv6 compatible?

nrchakradhar: I do not know anything about gunicorn, but from its documentation, the following may work:
https://docs.gunicorn.org/en/stable/settings.html
Multiple addresses can be bound. ex.:

$ gunicorn -b 127.0.0.1:8000 -b [::1]:8000 test:app
will bind the test:app application on localhost both on ipv6 and ipv4 interfaces.

I found an interesting observation here(its quite old though)
https://github.com/benoitc/gunicorn/issues/1138#issuecomment-158358666

Can we make this configurable from manifests so that IPv6 support is provided. There are few issues created already for IPv6 support: #5615 #5605

components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/helpers.py
```diff
@@ -24,7 +24,7 @@ def get_prefixed_index_html():
     with open(os.path.join(static_dir, "index.html"), "r") as f:
         index_html = f.read()
         index_prefixed = re.sub(
-            r"\<base href=\".*\"\>", '<base href="%s">' % prefix, index_html,
+            r"\<base href=\".*\".*\>", '<base href="%s">' % prefix, index_html,
```
elikatsis, line 27: What's the issue we are trying to fix here? Do we need to match **anything** that follows the `href` in the `<base>` tag?

components/crud-web-apps/tensorboards/Dockerfile
elikatsis, line 51: Can we use the `/bin/bash -c` notation here so that
1. we explicitly use `bash` instead of `sh` (https://docs.docker.com/engine/reference/builder/#shell-form-entrypoint-example), and
2. it is uniform with what we do for other web apps (JWA https://github.com/kubeflow/kubeflow/blob/28af7716bc22f25158ddde05af58d53f4fafc939/components/crud-web-apps/jupyter/Dockerfile#L50)

components/crud-web-apps/tensorboards/backend/app/utils.py
```diff
@@ -17,19 +16,12 @@ def parse_tensorboard(tensorboard):
...
         "namespace": tensorboard["metadata"]["namespace"],
         "logspath": tensorboard["spec"]["logspath"],
-        "age": helpers.get_age(tensorboard),
+        "age": tensorboard["metadata"]["creationTimestamp"],
         "status": status.create_status(phase, message, "")
     }
 
     return parsed_tensorboard
 
-def getPVCName(pvc):
```
elikatsis, line 26: Could you split the changes of this commit in 2?
1. date component related changes
2. PVC name related changes

When this PR is squashed & merged we will still maintain a description of this change in the final commit message

elikatsis (commented):
Thanks @kimwnasptd !

I left some comments here and there

components/crud-web-apps/tensorboards/frontend/package.json
kimwnasptd: I'm not sure I understand what the question or suggested change is here. Could you rephrase this?

DavidSpek: There is an inconsistency in the usage of `--base-href /tensorboard/ --deploy-url /tensorboard/"` on Line 6 and `--base-href /tensorboards/` on Line 7.

components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/helpers.py
kimwnasptd: @elikatsis I wanted to catch the use case where the `base` tag could look like this, which includes a space as well
```html
<base href="/" >
```

components/crud-web-apps/tensorboards/Dockerfile
kimwnasptd: Makes sense, will add a commit for this

components/crud-web-apps/tensorboards/backend/app/utils.py
kimwnasptd: makes sense

components/crud-web-apps/tensorboards/frontend/package.json
kimwnasptd: I see what you mean. The `start` script should be completely removed, since the users should run the UI locally with `build:watch` for now.

I'll push a commit to remove this script

components/crud-web-apps/tensorboards/Dockerfile
kimwnasptd: @nrchakradhar thank you for your comments. I'm planning on tackling the IPv6 support for all the web apps after 1.3.

The first step is to bind the IPv6 equivalent addresses as well, as you've pointed out in your links. But I want to do a good deep dive on this and test some edge cases for which I don't have the cycles right now.

components/crud-web-apps/tensorboards/frontend/package.json
kimwnasptd: Added a commit that removes the unused npm script

components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/helpers.py
elikatsis: Then I'd probably suggest using a regex like `r"\<base href=\".*\"\s?\>"` just to make things a bit more strict. Would this work?

kimwnasptd: Yeap, I'll use that instead

kimwnasptd: @elikatsis in the end it's not just the white spaces. We have two cases:
1. `<base href="/tensorboards/">`
2. `<base href="/" />`

There's also an extra `/` character in the end that I missed. So if you are OK I'll stick to the current catch-all rule I had which captures this as well, since we will be complicating the regex more to handle the `/`.

components/crud-web-apps/tensorboards/backend/app/utils.py
kimwnasptd: force pushed the changed commits. Resolving this one

components/crud-web-apps/tensorboards/Dockerfile
kimwnasptd: pushed a commit, will resolve this

components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/helpers.py
elikatsis: Sure let's move on with this as is and come back to it in the future if there's any reason to do so.

kimwnasptd:
@elikatsis I've also pushed my commits that address the review comments. The tests also pass, so feel free to merge if you don't have any further suggestions for the time

elikatsis:
Thanks @kimwnasptd !

/lgtm
/approve

DavidSpek:
@kimwnasptd  @PatrickXYS Just want to highlight here that merging this PR has also triggered a postsubmit for the Volumes and Jupyter Web Apps. It seems like prow is not taking the diff from the PR into account, but rather the diff compared to Master. This way, if a PR is not up to date with the latest commit in master, it will also trigger pre- and postsubmit jobs for any other components that have since changed in the master branch. 
So in this case, I think the images created with the postsubmit build might actually be a regression of the latest changes in the Volumes and Jupyter Web Apps.

PR #5673, opened 2021-03-09
Use informer / indexer to cache rolebindings

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5190

Recreated from https://github.com/kubeflow/kubeflow/pull/5202 due to inactivity of the contributor. 

/cc @yanniszark

PR #5697, opened 2021-03-15
Fix access-management OOMKilled during container build

DavidSpek:
The Kaniko build process for access-management has a tendency to get OOMKilled (see https://github.com/kubeflow/kubeflow/pull/5673). This PR makes the `memory` and `activeDeadlineSeconds` configurable for workflows and increases the memory limit from 4Gi to 6Gi for building the Kaniko image. The `activeDeadlineSeconds` is not used in this PR specifically, but is used in other PRs I have open so adding it in here will make the large amount of rebasing I am going to need to do a little easier.

### DIFF
--- a/py/kubeflow/kubeflow/cd/kaniko_builder.py
+++ b/py/kubeflow/kubeflow/cd/kaniko_builder.py
@@ -9,10 +9,11 @@ class Builder(ci.workflow_utils.ArgoTestBuilder):
         super().__init__(name=name, namespace=namespace, bucket=bucket,
                          test_target_name=test_target_name, **kwargs)
 
-    def build(self, dockerfile, context, destination):
+    def build(self, dockerfile, context, destination,
+              mem_override=None, deadline_override=None):
         """Build the Argo workflow graph"""
         workflow = self.build_init_workflow(exit_dag=False)
-        task_template = self.build_task_template()
+        task_template = self.build_task_template(mem_override, deadline_override)
 
         # Build component OCI image using Kaniko
         dockerfile = ("%s/%s") % (self.src_dir, dockerfile)
--- a/py/kubeflow/kubeflow/cd/access_management.py
+++ b/py/kubeflow/kubeflow/cd/access_management.py
@@ -12,4 +12,5 @@ def create_workflow(name=None, namespace=None, bucket=None, **kwargs):
 
     return builder.build(dockerfile="components/access-management/Dockerfile",
                          context="components/access-management/",
-                         destination=config.ACCESS_MANAGEMENT_IMAGE)
+                         destination=config.ACCESS_MANAGEMENT_IMAGE,
+                         mem_override="6Gi")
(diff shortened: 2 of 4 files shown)


PatrickXYS:
> activeDeadlineSeconds configurable for workflows

I don't understand, OOM is not related to activeDeadlineSeconds.

PatrickXYS (commented):
What's `components/access-management/empty` for?

PatrickXYS:
> The activeDeadlineSeconds is used in another PR, thought I may as well add it here to make all the rebasing I am going to need to do a little easier. 

Make sense, can you make it clear in the PR description?

py/kubeflow/kubeflow/cd/kaniko_builder.py
```diff
@@ -9,10 +9,12 @@ def __init__(self, name=None, namespace=None, bucket=None,
         super().__init__(name=name, namespace=namespace, bucket=bucket,
                          test_target_name=test_target_name, **kwargs)
 
-    def build(self, dockerfile, context, destination):
+    def build(self, dockerfile, context, destination,
+              second_dockerfile=None, second_destination=None,
```
kimwnasptd, line 13: I think these args aren't used anywhere

DavidSpek: They are used [here[(https://github.com/kubeflow/kubeflow/pull/5627/files#diff-d863e305a849abf34adc003f1657663d9261aa4c3fab0849dc71dfb761b3475d). Like I mentioned above I've added these changes to this PR to make the huge amount of rebasing I will need to do later on a little easier.

kimwnasptd: As @PatrickXYS mentioned above, could you edit the description to reflect what changes were made to assist the rebases of future PRs? 

This will help us keep track of the progress and chain between these PRs

DavidSpek: Sure thing, I had actually not intended the `second_*` being in here (which is why I didn't add the lines below), classic copy/paste error haha. I've removed the `second_dockerfile=None, second_destination=None` from this PR.

DavidSpek:
@kimwnasptd Can this be merged?

kimwnasptd:
Yeap, we should be good to go. Thanks for the PR

/lgtm
/approve

DavidSpek:
There was another build that failed with OOMKilled. Might need to up it to 8Gi of ram for this component if this continues to be a problem.

PatrickXYS:
Instead of keep increasing the requests, please investigate why it takes more than 8G memory.

DavidSpek:
I completely agree, but I'm not sure there is enough time for this in the immediate future before the release. After the release I want to look into optimizing the Kaniko builds by using more of the available arguments to speed things up and make the builds more efficient.

PR #5684, opened 2021-03-11
Volumes Management UI

kimwnasptd:
This PR is for #4758 
I don't want to close this issue as we can link other issues to it and have it as an umbrella one.

This web app is a slightly modified version from the one MiniKF has for the last year now. It should allow users to manage their lifecycle of their PVCs in the cluster via a UI. This is a basic UI that can act as a first iteration on how we manage our volumes/PVCs on Kubeflow.

It's structure is almost identical to our Jupyter web app and Tensorboards web app. It also utilizes the common code we have for both the frontend and the backend.

@elikatsis @StefanoFioravanzo feel free to take another look at it
/cc @elikatsis 
/cc @StefanoFioravanzo 
/assign @kimwnasptd 

The next steps will be:
* setting up the manifests for this app, cc @yanniszark 
* integrating it in the CentralDashboard side panel, cc @StefanoFioravanzo 
* adding it to our CI/CD

Issue #5566, opened 2021-01-28
Differences between standalone and kubeflow modes of our web apps


kimwnasptd:
There are many cases where users would like to install specific components of Kubeflow and not the entire solution. To accommodate this need we will have to engineer our web apps to be deploy-able as standalone components as well.

This issue aims to provide a cohesive story around how we want our web apps to behave when they are installed as part of Kubeflow or as standalone components. It's an effort to document the technical details as well as the design decisions to achieve this.

Once we have a solid understanding and ironed out most of the important details then we can convert the knowledge and discussions from this issue into a design doc.

## Kubeflow mode
When we say that an app is deployed alongside Kubeflow this means the following things:
1. There is **always** a logged in user. 
    * This is enforced in a central place which can be IAP, Dex etc. The user first logins to this central entity and then navigates to Kubeflow. This entity places the username in a configurable header in each request that passes from it
    * The backend of each app will be checking that header to deduce the identity of the user and will creating [SubjectAccessReviews](https://kubernetes.io/docs/reference/kubernetes-api/authorization-resources/subject-access-review-v1/) to ensure the user is authorized to perform an action https://github.com/kubeflow/community/pull/334
2. There is a CentralDashboard which
    * decides what Namespaces to show to the user and feeds to the app the selected Namespace
    * has a left hand sidebar, which can have subsections https://github.com/kubeflow/kubeflow/pull/5474. This side bar can be used for navigating the user between different pages of the underlying deployed app

## Standalone mode
This means that only this app is exposed and no other Kubeflow component. This introduces the following consequences that the web apps need to handle:
1. There is no CentralDashboard
    * the app must handle it's navigation to different pages on each own
    * it must handle Namespace selection on its own
2. There is no guarantee that there will be an authentication system, a logged in user. This leads into two scenarios:
    * if there is such a system then it must be setting the user identity in a specific header. This header is configurable for our web apps. If a custom one is used the admin can let the app know which header to look for. In this case the app relies on `SubjectAccessReviews` to ensure the user is authorized.
    * if there is no such a system then we expect that every user is admin and can do anything they want in any namespace. In this case the backend must be configurable to not perform any authnz checks

kimwnasptd:
cc @kubeflow/wg-notebooks-leads 

And also since this affects the web app for katib https://github.com/kubeflow/katib/issues/1421 https://github.com/kubeflow/katib/issues/1090 I'll also
cc @kubeflow/wg-automl-leads

[...]

PR #5706, opened 2021-03-17
Centraldashboard: update menu items in manifests

StefanoFioravanzo:
This PR updates the Centraldashboard ConfigMap to populate the sidebar menu. As part of Kubeflow 1.3 we want to make Centraldashboard the unified place for navigating between Kubeflow components. We have been discussing this in the context of the Web Apps here https://github.com/kubeflow/kubeflow/issues/5566#issue-796194300.

Quoting the relevant passage:

> Kubeflow mode
> When we say that an app is deployed alongside Kubeflow this means the following things:
>
> 1. [...]
> 2. There is a CentralDashboard which:
    - decides what Namespaces to show to the user and feeds to the app the selected Namespace
    - **has a left hand sidebar, which can have subsections https://github.com/kubeflow/kubeflow/pull/5474. This side bar can be used for navigating the user between different pages of the underlying deployed app**

Pipelines already supports hiding its sidebar in Kubeflow mode: https://github.com/kubeflow/pipelines/pull/5200

Katib comes with a new beta UI, which will replace the existing one in the near future. Still, for Kubeflow 1.3, Centraldashboard will expose only the existing one and users will be able to provision the new UI by thenmselves and add a sidebar item by editing the ConfigMap, if they wish. See https://github.com/kubeflow/katib/issues/1421#issuecomment-800147949. The existing UI stil does not support hiding the sidebar, so we can include a single menu item to direct to the Katib main page.

Proposed menu structure:

- `Home`
- `Notebooks`      -> Jupyter Web App
- `Pipelines`      -> KFP Pipelines Page
- `Experiments` (Collapsible Section)
	- `Pipelines`       -> KFP Experiments Page
	- `AutoML`       -> Katib main page
- `Runs`           -> KFP Recurring Runs Page
- `Recurring Runs` -> KFP Recurring Runs Page
- `Artifacts`      -> KFP Artifacts Page
- `Executions`     -> KFP Executions Page

Two top-level menu items that we hope we can include are:

- `Volumes`: Blocking issue https://github.com/kubeflow/kubeflow/pull/5684 and missing manifests
- `Tensorboard`: Blocking issue https://github.com/kubeflow/kubeflow/pull/5693 and missing manifests 

Screenshot 2021-03-17 at 10 37 35

@kubeflow/wg-automl-leads
@kubeflow/wg-notebooks-leads
@kubeflow/wg-pipeline-leads
@kubeflow/wg-serving-leads
@kubeflow/wg-training-leads

### DIFF
--- a/components/centraldashboard/manifests/base/configmap.yaml
+++ b/components/centraldashboard/manifests/base/configmap.yaml
@@ -4,79 +4,118 @@ data:
     {
       "menuLinks": [
         {
-          "link": "/pipeline/",
-          "text": "Pipelines"
-        },
-        {
+          "type": "item",
           "link": "/jupyter/",
-          "text": "Notebook Servers"
+          "text": "Notebooks",
+          "icon": "book"
         },
         {
-          "link": "/katib/",
-          "text": "Katib"
-        }
+          "type": "item",
+          "link": "/pipeline/#/pipelines",
+          "text": "Pipelines",
+          "icon": "kubeflow:pipeline-centered"
+        },
+        {
+          "type": "section",
+          "text": "Experiments",
+          "icon": "done-all",
+          "items": [
+            {
+              "link": "/pipeline/#/experiments",
+              "text": "Pipelines"
+            },
+            {
+              "link": "/katib/",
+              "text": "AutoML"
+            }
+          ]
+        },
+        {
+          "type": "item",
+          "link": "/pipeline/#/runs",
+          "text": "Runs",
+          "icon": "maps:directions-run"
+        },
+        {
+          "type": "item",
+          "link": "/pipeline/#/recurringruns",
+          "text": "Recurring Runs",
+          "icon": "device:access-alarm"
+        },
+        {
+          "type": "item",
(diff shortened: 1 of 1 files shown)


StefanoFioravanzo:
Thanks @Bobgy . Ouch, I should have added the `/hold` label since the PR was `approved` by default. If anyone has comments and/or wants to revisit this, feel free to comment here. I will also open a general tracking issue for the sidebar menu so we can document choices and keep track of all PRs.

Bobgy:
@StefanoFioravanzo it's configure prow to require explicit approval mode. We plan to do it for KFP repo. I wonder if you like that for other repos

StefanoFioravanzo:
@Bobgy That would be very helpful indeed. I was under the assumption that Prow didn't have such an option. Could you link to some issue or PR in KFP where you are discussing this? cc @kimwnasptd as well on this

Bobgy:
@StefanoFioravanzo https://github.com/GoogleCloudPlatform/oss-test-infra/pull/746

It's also possible to configure at org level, depends on people interest

PR #5707, opened 2021-03-17
Add prow config for new notebook servers

DavidSpek:
After realizing that @Bobgy would need to approve each of the PRs for the new notebook images due to the `prow_config.yaml` being edited, I thought it would be best to move those changes to a dedicated PR so @thesuperzapper and I can work more quickly when merging the notebook image PRs. 

@kimwnasptd can you also read over this quickly, as it is essential for being able to move fast with the other PRs and getting the images built.

### DIFF
--- a/prow_config.yaml
+++ b/prow_config.yaml
@@ -94,6 +94,94 @@ workflows:
       - components/crud-web-apps/common/backend/*
       - components/crud-web-apps/volumes/*
     kwargs: {}
+  # Run build test for notebook-server-base OCI image
+  - py_func: kubeflow.kubeflow.ci.notebook_servers.notebook_server_base_tests.create_workflow
+    name: nb-base
+    job_types:
+      - presubmit
+    include_dirs:
+      - components/example-notebook-servers/base/*
+    kwargs: {}
+  # Run build test for notebook-server-jupyter OCI image
+  - py_func: kubeflow.kubeflow.ci.notebook_servers.notebook_server_jupyter_tests.create_workflow
+    name: nb-jupyter
+    job_types:
+      - presubmit
+    include_dirs:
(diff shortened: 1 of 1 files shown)


kimwnasptd:
@DavidSpek I was just thinking about the same think and was about to create an issue! I completely agree on this approach, as it will both help decouple the effort of merging the Dockerfiles and working on the tests.

I'll also be keeping an eye in the PRs and try to also bring up any points wrt the CI/CD here as well.

Thank you for opening the PR!

Bobgy:
/approve
/hold
You can decide when to merge by yourselves.

Thank you for saving me efforts!

DavidSpek:
@kimwnasptd Can this be merged? Then I can get started with rebasing and fixing Notebook PRs for this.

kimwnasptd:
@DavidSpek just to make sure, will merging this PR now give us any trouble for merging the following notebook PRs?

My concern is about triggering tests for the Notebooks if the base would be modified. In that case, since the files for the other Notebooks would be missing we wouldn't be able to merge the first PR [ which is for the base Notebook image I assume? ]. But looking at the modified yaml the jobs for the other Notebooks don't run when the base is changes so we should be good.

DavidSpek:
@kimwnasptd Indeed that should not be a problem. All the image builds are setup to only run when that image itself is changed (this method would tie in nicely with Renovate because of this). Merging this should not cause any problems. However, I am continuously monitoring the prow dashboard so if something unexpected happens I'll quickly create a PR to fix it.

kimwnasptd:
Thanks for confirming. Lets move on with this PR and if something is different from our current best understanding we can come back and reiterate.

/lgtm
/approve

PR #5710, opened 2021-03-17
web-apps: Extend common code for the apps to work in standalone mode

### LINKED ISSUE #5566: Differences between standalone and kubeflow modes of our web apps

kimwnasptd:
Partially fixes #5566 

This PR makes the following changes:
* Introduce a new `APP_NO_AUTHNZ` env var that configures the app to completely skip authnz checks
* Extend the Namespace Service to expose if the Central Dashboard is found

With these changes the admin could configure the apps to skip any authnz checks, which is the case for standalone mode as we've described in #5566.

The next step will be to make our different web apps use the updated `NamespaceService` to understand if the Central Dashboard is present. In that case they should show the namespace selector, which was previously found in the Dashboard.

### DIFF
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/namespace.service.ts
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/namespace.service.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { ReplaySubject, Subscription, fromEvent } from 'rxjs';
+import { ReplaySubject, Subscription, fromEvent, BehaviorSubject } from 'rxjs';
 
 declare global {
   interface Window {
@@ -15,9 +15,11 @@ export class NamespaceService {
 
   // Observable string sources
   private selectedNamespaceSource = new ReplaySubject<string>(1);
+  private dashboardConnectedSource = new BehaviorSubject<boolean>(true);
 
   // Observable string streams
   selectedNamespace$ = this.selectedNamespaceSource.asObservable();
+  dashboardConnected$ = this.dashboardConnectedSource.asObservable();
 
   constructor() {
     fromEvent(window, 'load').subscribe(_ => {
@@ -34,7 +36,13 @@ export class NamespaceService {
             cdeh.onNamespaceSelected = this.updateSelectedNamespace.bind(this);
           },
         );
-      } else if (this.currNamespace === undefined) {
+
+        return;
+      }
+
+      this.dashboardConnectedSource.next(false);
+
+      if (this.currNamespace === undefined) {
         this.updateSelectedNamespace('kubeflow-user');
       }
     });
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/namespace.service.ts
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/namespace.service.ts
@@ -1,5 +1,6 @@
 import { Injectable } from '@angular/core';
 import { ReplaySubject, Subscription, fromEvent, BehaviorSubject } from 'rxjs';
+import { DashboardState } from '../enums/dashboard';
 
 declare global {
   interface Window {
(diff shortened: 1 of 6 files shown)


elikatsis (commented):
After discussing live for a bit, I think we've concluded to:
1. use a `DISABLE_AUTH` Python variable,
2. make the environment variable `APP_DISABLE_AUTH`,
3. make it as follows
   ```python
   DISABLE_AUTH = os.getenv("APP_DISABLE_AUTH", "false").lower == "true"
   ```
4. and move the logic to a different module imported by both `authn.py` and `authz.py`

components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/services/namespace.service.ts
```diff
@@ -15,9 +15,11 @@ export class NamespaceService {
 
   // Observable string sources
   private selectedNamespaceSource = new ReplaySubject<string>(1);
+  private dashboardConnectedSource = new BehaviorSubject<boolean>(true);
```
tasos-ale, line 18: Hello @kimwnasptd. I think this Observable should not start with `true` as a default value. If someone uses the `dashboardConnected$` property and the `centraldashboard` is not ready, it will wrongly inform the user that it is ready. Maybe you can use `null` as the unknown state or even better use an enum. Something like this:
```
enum DashboardState {
  Unknown = 0, // or Connecting
  Connected,
  Disconnected,
}
```

kimwnasptd: Thank you for the feedback @tasos-ale! 

The reason I used `true` as a default value here is because if it starts as `null` then the app will think that the dashboard is not present and try to make a request to fetch the namespaces. 

So in this case I take for granted the opposite, which is expect the dashboard to be present and if after the check it's missing then the apps will make the request to fetch the namespaces

kimwnasptd: but indeed the enumeration is a better way to express this. I'll use this enum instead of a boolean value and the apps will fetch the namespaces only if the state is `Disconnected`

kimwnasptd: pushed a commit for the described changes

kimwnasptd:
@elikatsis @tasos-ale I pushed some review commits with the proposed changes. The tests also pass so if you don't have any further comments we can go on and merge this PR

PR #5718, opened 2021-03-18
profiles: Consolidate manifests

yanniszark:
Umbrella Issue: kubeflow/manifests#1769

```
config
├── base
├── base_v3
├── certmanager
├── crd
├── default
├── manager
├── overlays
│   ├── application
│   ├── debug
│   ├── devices
│   ├── istio
│   └── test
├── rbac
├── samples
└── webhook
```

#### Explanation

The profile-controller has multiple versions of manifests:
- `base` and `overlays`, originally made for `kfctl`.
    - Of the various overlays, we only really use `istio`.
    - `application` installs the Application CR which we want to move away from so this is a good chance to remove it.
- `base_v3`, made to work better with `kustomize`.
- Rest of the folders, generated by kubebuilder.

The purpose of this PR is to simplify and consolidate the structure.

#### Recommended end state

The final proposed structure is:
```
config
├── base
├── crd
├── default
├── manager
├── overlays
│   ├── kubeflow
│   └── standalone
├── rbac
└── samples
```

The breakdown is the following:
- `crd`, `default`, `manager`, `rbac`, `samples`: Kubebuilder-generated structure. We keep this in order to be compatible with kubebuilder workflows. This is not meant for the consumer of the manifests.
- `base`, `overlays`: Kustomizations meant for consumption by the user:
    - `overlays/kubeflow`: Installs `profile-controller` as part of Kubeflow. The resulting manifests should be the same as the result of the [deprecated `base_v3` from kubeflow/manifests](https://github.com/kubeflow/manifests/tree/30bac3c09b11012312d6ce8d2712902648c54a09/apps/profiles/upstream/base_v3). At a glance, it makes the following changes:
        - Use namespace `kubeflow`.
        - Remove namespace resource.
        - Add KFAM container.
        - Add KFAM Service and VirtualService.
    - `overlays/standalone`: Install `profile-controller` in its own namespace. Useful for testing or for users that prefer to install just the controller.


yanniszark:
To review:
- Build `base_v3` manifests, which we presented to users in 1.2: https://github.com/kubeflow/manifests/tree/30bac3c09b11012312d6ce8d2712902648c54a09/apps/profiles/upstream/base_v3
- Build `config/overlays/kubeflow`.
- Compare.

DavidSpek:
@yanniszark Sorry for the unrelated comment on this PR, but the idea just popped into my head. Would it be useful to have a presubmit test that tries to deploy the manifests of each component? Now it just tried to build the image (which is quite useless here, this folder should maybe be ignored for that). However, I think it might be an idea to have the presubmit test build the image, push it to a private registry, edit the manifests to use this private image, then try and deploy the manifests (then maybe run more tests on that deployment). Or even just as a separate thing test deploying manifests as they are when they have been changed.

yanniszark:
@DavidSpek I think that's a great idea for enhancing our CI. It's what some components already do (e.g., pytorch-operator).
Do you want to open an issue in order to sketch what you have in mind?

DavidSpek:
@yanniszark I was just about to give my brain a break for the day, it's not being very useful anymore at the moment. I'll create an issue for this tomorrow. If I have the time and energy I'll also try to create a PoC PR for one of the components tomorrow or in the weekend.

kimwnasptd:
I'll move on with merging this PR. @DavidSpek feel free to create the issue for the e2e testing at your convenience.

/lgtm
/approve

PR #5720, opened 2021-03-19
centraldashboard: Consolidate manifests

yanniszark:
Umbrella-Issue: kubeflow/manifests#1769

Then, let's take a look at `centraldashboard`. This is the current structure:

```.
├── base
├── base_v3
└── overlays
    ├── application
    ├── istio
    └── stacks
```

#### Explanation

- `base_v3` combines `base`, `overlays/application` and `overlays/istio` by importing specific files directly from these folders. Since we now use plain kustomize, we remove this and standardize on `base` and `overlays`.
- `overlays/application`: Adds an Application CR resource for the poddefaults webhook. We want to remove Application CRs, so this is a good opportunity to do so.
- `overlays/stacks`: Patches the CentralDashboard with values from a central ConfigMap. If a distro wants this, they can introduce their own patch.
- `overlays/istio`: The only useful overlay. Adds an Istio VirtualService for CentralDashboard. We keep this.

#### Recommended end state

```
.
├── base
└── overlays
    └── istio
```


yanniszark:

yanniszark:
To review:
- Build `base_v3` manifests, which we presented to users in 1.2: https://github.com/kubeflow/manifests/tree/30bac3c09b11012312d6ce8d2712902648c54a09/apps/centraldashboard/upstream/overlays/stacks
- Build `config/overlays/istio`.
- Compare.

StefanoFioravanzo:
Thanks @yanniszark !
/lgtm
/approve

PR #5721, opened 2021-03-19
jupyter-web-app: Consolidate manifests

yanniszark:
Umbrella-Issue: kubeflow/manifests#1769

```.
├── base
├── base_v3
└── overlays
    ├── application
    ├── aws
    └── istio
```

#### Explanation

- `base_v3` combines `base`, `overlays/application` and `overlays/istio` by importing specific files directly from these folders. Since we now use plain kustomize, we remove this and standardize on `base` and `overlays`.
- `overlays/application`: Adds an Application CR resource. We want to remove Application CRs, so this is a good opportunity to do so.
- `overlays/aws`: Adds a jwa config for AWS. This belongs to an AWS distro. We should remove this.
- `overlays/istio`: The only useful overlay. Adds an Istio VirtualService for CentralDashboard. We keep this.

#### Recommended end state

```.
├── base
└── overlays
    └── istio
```


yanniszark:

To review:
- Build `base_v3` manifests, which we presented to users in 1.2: https://github.com/kubeflow/manifests/tree/30bac3c09b11012312d6ce8d2712902648c54a09/apps/jupyter/jupyter-web-app/upstream/base_v3
- Build `config/overlays/istio`.
- Compare.

kimwnasptd:
Thank you @yanniszark. 

I'll probably need to push a small one to update the config-yaml and use the latest image for #5646 , but this can be done in a follow up PR. cc @DavidSpek 

/lgtm 
/approve

DavidSpek:
Have the role bindings for being able to find the available GPU vendors been added in this PR?

PR #5727, opened 2021-03-19
Manifests for Tensorboards web app

kimwnasptd:
Manifests for the Tensorboards web app.

The app will be exposed under the `/tensorboards` prefix. The next step will now be to include it in the dashboard.

/cc @yanniszark


yanniszark:
Thanks @kimwnasptd 
/lgtm
/approve

PR #5728, opened 2021-03-19
Manifests for the Volumes web app

kimwnasptd:
Manifests for the Volumes web app #4758.

The app will be exposed under the `/volumes` prefix. The next step will now be to include it in the dashboard.

/cc @yanniszark


yanniszark:
Thanks @kimwnasptd 
/lgtm
/approve

PR #5730, opened 2021-03-19
Manifests for Tensorboard controller

kimwnasptd:
Manifests for the Tensorboard Controller

/cc @yanniszark

### DIFF
--- a/components/tensorboard-controller/.gitignore
+++ b/components/tensorboard-controller/.gitignore
@@ -22,5 +22,3 @@ bin
 *.swp
 *.swo
 *~
-
-manager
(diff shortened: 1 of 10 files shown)


DavidSpek:
@kimwnasptd I think you mean `/tensorboards` and not `/volumes` :). 

Related to this, I'm wondering the the `Jupyter Web App` hosted at `/jupyter` shouldn't be `Notebooks Web App` hosted at `/notebooks` to be more cohesive with the other apps. Also, since more "notebooks" in the form of VS-Code and RStudio will be supported it definitely isn't restricted to just Jupyter anymore. Not suggesting that this change needs to be rushed in before the release now, but it's something that should be discussed in the future.

kimwnasptd:
> @kimwnasptd I think you mean /tensorboards and not /volumes :).

Too much copy paste late at night never leads to good things. Thanks for the note!

And yes, we should definitely use a different naming scheme for our web app since it won't only now focus on Jupyter. We'll have a lot of exciting goals to decide for 1.4.

PR #5732, opened 2021-03-20
Add listing nodes to JWA ClusterRole for GPU vendor detection

DavidSpek:
This fix is needed so that the JWA is able to detect which GPU vendors are available in the cluster. Without it an error will be displayed to the users. 

/cc @kimwnasptd @yanniszark


kimwnasptd:
Thanks for the PR @DavidSpek!

/lgtm
/approve

PR #5733, opened 2021-03-20
Prepare CI/CD configs for Notebook Server images

DavidSpek:
Similarly to https://github.com/kubeflow/kubeflow/pull/5707, I've added all the changes to the CI/CD makefiles and the `config.py` that contains the registry destination for the notebook images into this PR. This way, the Notebook PRs won't need as much rebasing after each merge and can be merged a bit more easily. 

@kimwnasptd @thesuperzapper If one of you could quickly browse over this PR and merge it that would be great. I'm now going to remove these sections from the Notebook PRs and run through them one more time so that during the merging all that will need updating is the image tag of the base images.


thesuperzapper:
Looks good!

/approve
/lgtm

PR #5739, opened 2021-03-20
centraldashboard: Add missing links ConfigMap

yanniszark:
/cc @StefanoFioravanzo @kimwnasptd 

For some reason, I missed this in the first PR.

### DIFF
--- a/components/centraldashboard/manifests/base/configmap.yaml
+++ b/components/centraldashboard/manifests/base/configmap.yaml
@@ -53,7 +53,7 @@ data:
           "link": "/pipeline/#/executions",
           "text": "Executions",
           "icon": "av:play-arrow"
-        },
+        }
       ],
       "externalLinks": [ ],
         "quickLinks": [
@@ -115,7 +115,7 @@ data:
             "link": "https://www.kubeflow.org/docs/started/requirements/"
           }
         ]
-}
+    }
 kind: ConfigMap
 metadata:
   name: centraldashboard-config
--- a/components/centraldashboard/manifests/base/deployment.yaml
+++ b/components/centraldashboard/manifests/base/deployment.yaml
@@ -39,5 +39,5 @@ spec:
         - name: REGISTRATION_FLOW
           value: $(CD_REGISTRATION_FLOW)
         - name: DASHBOARD_LINKS_CONFIGMAP
-          value: centraldashboard-links-config
+          value: $(CD_CONFIGMAP_NAME)
       serviceAccountName: centraldashboard
--- a/components/centraldashboard/manifests/base/kustomization.yaml
+++ b/components/centraldashboard/manifests/base/kustomization.yaml
@@ -9,7 +9,7 @@ resources:
 - role.yaml
 - service-account.yaml
 - service.yaml
-- links-configmap.yaml
+- configmap.yaml
 namespace: kubeflow
 commonLabels:
   kustomize.component: centraldashboard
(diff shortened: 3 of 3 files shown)


DavidSpek:
@yanniszark This file was deleted/renamed and updated in https://github.com/kubeflow/kubeflow/pull/5706, so I don't think this PR is correct.

yanniszark:
@DavidSpek thanks for the heads-up. Here is what went wrong:
- PR https://github.com/kubeflow/kubeflow/pull/5706 introduced the new ConfigMap file, but didn't update kustomization.yaml. Thus, the kustomization was broken.

The extra things I fixed are:
- Name the ConfigMap `centraldashboard-links-config`, same as before.
- Fix a small JSON error inside the ConfigMap.

cc @StefanoFioravanzo plz take a look

andreyvelich:
@yanniszark @StefanoFioravanzo As I can see, for Kubeflow 1.3 we moved Katib UI to the `Experiments/AutoML` sub-section.
Will it confuse the user ?

Maybe we should keep Katib in the Root dashboard access, like Pipelines and Notebooks ?
We haven't changed Katib UI yet to hide the menu part and we haven't discussed how we should manage the pages in the Katib UI.

As well, in the docs we introduce the old menu design: https://www.kubeflow.org/docs/components/central-dash/overview/.

Will it be easier for the user to see the Katib UI in the root menu section?

/cc @kimwnasptd @gaocegege @johnugeorge

StefanoFioravanzo:
@yanniszark Note this https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/app/k8s_service.ts#L5 
The default name is `centraldashboard-config`. Let's stick with it since the configmap is not just about links.

StefanoFioravanzo:
@andreyvelich Not problem with me moving the Katib menu item one level up in the sidebar. If you confirm this and the others agree, then let's do it

DavidSpek:
I don't really have an opinion on this, but removing Katib from that section would make it contain only one item. Kind of defeats it's purpose, but assuming other links will be added to it relatively soon it should be fine.

yanniszark:
> @yanniszark Note this https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/app/k8s_service.ts#L5
The default name is centraldashboard-config. Let's stick with it since the configmap is not just about links.

Got it, thanks! I kept the name as `centraldashboard-config` and edited the manifests accordingly.

andreyvelich:
> @andreyvelich Not problem with me moving the Katib menu item one level up in the sidebar. If you confirm this and the others agree, then let's do it

Sure. What do you think @gaocegege @johnugeorge @kimwnasptd ?

> I don't really have an opinion on this, but removing Katib from that section would make it contain only one item. Kind of defeats it's purpose, but assuming other links will be added to it relatively soon it should be fine.

Yes, we have plans with @kimwnasptd to integrate various pages for the new Katib UI and directly add them to the Central Dashboard.
From my understanding, it makes more sense if Katib have independent section in the Central Dashboard with the different sub-sections in the future.

StefanoFioravanzo:
Thanks @yanniszark ! 
/lgtm
/approve

StefanoFioravanzo:
@andreyvelich let's discuss and address this tomorrow in another PR @kimwnasptd will open to add some other menu items to Centraldashboard

yanniszark:
Thanks @StefanoFioravanzo!
@andreyvelich let's address the Katib links issue in a follow-up PR.

andreyvelich:
Sure thank you @yanniszark @StefanoFioravanzo !

yanniszark:
I'm taking the thumbs-up in Stefano's post to mean ok then :) 
/remove-hold

PR #5746, opened 2021-03-22
Establish release process

yanniszark:
Add the necessary tooling to release components in this repo. The release process must generate:
- Images.
- Manifests.

The envisioned release process is similar to what Katib, KFServing and others follow:

1. Create a new release branch:
```sh
VERSION="1.3"
git checkout -b release-$VERSION origin/master
```
2. Edit files under manifests (e.g., Deployments) to the release image tag. This can be done manually or with a `sed` command:
```sh
TAG=$VERSION
components/releasing/update-manifests-images/update-manifests-images
```
3. Bump version in `VERSION` file:
```sh
echo "$VERSION" > VERSION
```
4. Commit changes to the release branch:
```sh
git commit -s -a -m "Release $VERSION"
```
5. PR merged (after CI validates and builds all images for that commit).
6. CI job (postsubmit) kicks in and builds the images for `$VERSION`.

/hold


DavidSpek:
RStudio and RStudio-Tidyverse have been merged, I think you will need to rebase on master to have them included. The test seems to have failed due to the pod building the pytorch container being deleted. 
https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-nb-j-pt-5746-6604369-9008-65bb?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-nb-j-pt-5746-6604369-9008-65bb-4233577146

DavidSpek:
@yanniszark I'm not sure the `VERSION="1.3"` tag for the example notebook images makes a lot of sense due to how they inherit from each other. They are also not really part of Kubeflow in the same way as an image for the notebook-controller or the central-dashboard.

yanniszark:
@DavidSpek thanks for the suggestions. Here are my thoughts:

>  I'm not sure the VERSION="1.3" tag for the example notebook images makes a lot of sense due to how they inherit from each other. They are also not really part of Kubeflow in the same way as an image for the notebook-controller or the central-dashboard.

They are code living in the repo and currently we version the whole code together. The `VERSION` tag of the image will correspond to a tag in git, aka a commit.

> Also, the way the current tagging is done in the CD is <branch>-<commit-id>, to for the 1.3 branch it would be release-1.3-1a2b3c4d. I think this would make more sense that hard pinning the tag to 1.3 in the case that a fix is cherry picked into the release branch.

I'm not sure I follow. The VERSION number will correspond to a single commit, same as the tag you are talking about (`release-1.3-1a2b3c4d`). We are using a tag in order to be able to store the reference in our manifests in the same commit. Did you mean something else?

DavidSpek:
@yanniszark What I meant was that the `VERSION` tag would essentially be the same as `<branch-name>`. However, if you use `VERSION` as the tag, what happens if a fix needs to get cherry picked into the release branch? Then there would be new images with the same `VERSION` tag as the old ones (thus replacing them), correct? If that is the case, you can essentially never provide updated images for a release.

yanniszark:
> However, if you use VERSION as the tag, what happens if a fix needs to get cherry picked into the release branch? Then there would be new images with the same VERSION tag as the old ones (thus replacing them), correct?

Let me go through the steps of cherry-picking to show how it wouldn't happen like this:
1. Cherry-pick a fix ontop of release branch and make PR.
2. PR does not trigger releasing jobs because it doesn't change the VERSION file.
3. After all cherry-pick PRs are in, we create a final release PR that updates the VERSION file and the manifests references.
4. Once the release PR is merged, the postsubmit release job pushes the release images.

Does that make sense? Did you mean something else?

yanniszark:
All blocking PRs are ready from the notebooks side.
We are ready to merge this PR!
cc @kimwnasptd @DavidSpek @StefanoFioravanzo @elikatsis

prow_config.yaml
```diff
@@ -180,6 +200,7 @@ workflows:
     job_types:
       - presubmit
     include_dirs:
+      - components/releasing/version/*
```
DavidSpek, line 203: Why is this line `components/releasing/version/*` while all the others are `components/releasing/version/*`?

prow_config.yaml
```diff
@@ -116,6 +128,7 @@ workflows:
...
```
DavidSpek, line 131: Same here, shouldn't this be `releasing/version/*` like the others?

prow_config.yaml
```diff
@@ -353,3 +374,189 @@ workflows:
...
+    kwargs:
+      release: true
+  # Create and push notebook-server-rstudio OCI image to ECR
+  - py_func: kubeflow.kubeflow.cd.notebook_servers.notebook_server_rstudio.create_workflow
+    name: nb-rstudio
+    job_types:
+      - postsubmit
+    include_dirs:
+      - components/releasing/version/*
+      - components/example-notebook-servers/rstudio/*
```
DavidSpek, line 487: Shouldn't this be `releasing/version/*` like the others?

prow_config.yaml
yanniszark: > Why is this line components/releasing/version/* while all the others are components/releasing/version/*?

I must be missing something, because both things look the same to me :P

prow_config.yaml
```diff
@@ -353,3 +374,189 @@ workflows:
...
+    kwargs:
+      release: true
+  # Create and push notebook-server-rstudio-tidyverse OCI image to ECR
+  - py_func: kubeflow.kubeflow.cd.notebook_servers.notebook_server_rstudio_tidyverse.create_workflow
+    name: nb-rs-tidy
+    job_types:
+      - postsubmit
+    include_dirs:
+      - components/releasing/version/*
+      - components/example-notebook-servers/rstudio-tidyverse/*
```
DavidSpek, line 560: Shouldn't this be `releasing/version/*` like the others?

prow_config.yaml
yanniszark: Ok I got what you mean!

releasing/update-manifests-images
```diff
@@ -0,0 +1,200 @@
...
+        "kustomization": "components/profile-controller/config/base/kustomization.yaml",
+        "images": [
+            {
+                "name": "public.ecr.aws/j1r0q0g6/notebooks/profile-controller",
+                "newName": "public.ecr.aws/j1r0q0g6/notebooks/profile-controller",
+            },
+        ],
+    },
+    {
+        "name": "KFAM",
```
DavidSpek, line 132: Small nit, this name stands out compared to the rest. People not familiar with the code might now know what `KFAM` means. Not a blocker for the PR but `Access Management` might be better to use.

yanniszark: Sounds reasonable

yanniszark:
@DavidSpek great catch on the `include_dirs`. The TLDR is that a rebase went wrong.
I fixed these and also the KFAM naming. Please take a look!

prow_config.yaml
DavidSpek: Sorry, copy/paste error. I thought the PR wasn't held so I jumped out of bed and looked through it very quickly.

DavidSpek:
Looks good, awesome work! Excited to see what happens once merged. 
/lgtm

yanniszark:
/remove-hold

@Bobgy we need your final blessing here in order to merge, since this PR touches the `prow_config.yaml` file. :D

PR #5761, opened 2021-03-24
Allow user to add/delete labels to user namespace using ConfigMap. Fix #5712

### LINKED ISSUE #5712: [Kubeflow 1.3] Allows dynamic label configuration on profile-controller for Istio upgrade
/kind bug

To prepare for Kubeflow 1.3 release, we need to upgrade Istio from 1.4 to 1.9: https://github.com/kubeflow/gcp-blueprints/issues/144. However, in order to inject sidecar for new Istio versions, we need to remove the label: `istio-injection:enabled` for any namespace created by users, and enable automatic sidecar injection in new version. See [reference](https://cloud.google.com/service-mesh/docs/proxy-injection#enabling_automatic_sidecar_injection). With the `istio-injection` existing, automatic side injection is disabled in new version. Therefore, we need to allow users of profile-controller to configure labels in the following two places:

https://github.com/kubeflow/kubeflow/blob/28af7716bc22f25158ddde05af58d53f4fafc939/components/profile-controller/controllers/profile_controller.go#L131
https://github.com/kubeflow/kubeflow/blob/28af7716bc22f25158ddde05af58d53f4fafc939/components/profile-controller/controllers/profile_controller.go#L68-L72

Since we are upgrading istio versions across Kubeflow (https://github.com/kubeflow/manifests/issues/1735) in v1.3 release, we need to make the corresponding change before feature cut.

cc @yanniszark @Bobgy @kimwnasptd

zijianjoy:
Address requirement https://github.com/kubeflow/kubeflow/issues/5712.

The profile-controller applies a hardcoded list of labels to Profile
namespaces. Make this list of labels configurable with a ConfigMap. The
new list of steps for the profile-controller is:

1. **(new)** Profile-controller starts and reads the file given by the
   `--namespace-labels-path` CLI flag (defaults to
   `/etc/profile-controller/namespace-labels.yaml`). This file contains
   a YAML dict of label key/values to add to each Profile namespace.
2. Profile-controller creates namespace.
3. Profile-controller merges existing namespace labels with given
   labels. A given label is applied ONLY if it doesn't exist already.
4. **(new)** If a given label value is empty (``), remove the label from
   the namespace if it already exists.
5. Watch the ConfigMap-backed file and trigger reconciliations.
6. Exit if there is no namespace labels file.

We are using a ConfigMap instead of a plain CLI flag, so that users can
change the list of labels without restarting the profile-controller.


Bobgy:
> If we don't provide namespace-labels.properties, will it still work? ConfigMap creation might fail. How do we avoid breaking default deployment (non-GCP)?

I think we should keep current manifests intact and build an overlay with the configmap, so only people who want to use the configmap can use it.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -616,15 +624,59 @@ func removeString(slice []string, s string) (result []string) {
...
+			} else if len(arr) == 2 {
+				// Add or overwrite label map if value exists.
+				kubeflowNamespaceLabels[arr[0]] = arr[1]
+			} else {
+				// Set value to be empty if nothing exists after "=".
+				// It means this label needs to be removed.
+				kubeflowNamespaceLabels[arr[0]] = ""
+			}
+		}
+	}
```
Bobgy, line 652: It's a nice implementation of the syntax you are proposing.

However, it seems a little unnecessary if we can just use YAML format as the config format, so people do not need to reinvent the wheels and consumers won't guess about how the format is parsed. There are standard libraries for parsing YAML in go: https://github.com/go-yaml/yaml/tree/v2.

Using the standards mean that e.g. it's very easy to insert comments in the configfile, without you needing to implement logic to ignore comments.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -616,15 +624,59 @@ func removeString(slice []string, s string) (result []string) {
 	return
 }
 
+func mergeLabelPropertiesToMap(logger logr.Logger) {
+	file, err := os.Open(NAMESPACE_LABELS_PROPERTIES)
```
Bobgy, line 628: A controller runs in reconcile loops, this method is called very very often, so it'd be better to load the config file at controller startup and fail the controller if any errors show up.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -55,6 +58,9 @@ const USER = "user"
 const ROLE = "role"
 const ADMIN = "admin"
 
+// External input for labels to be added to namespace.
+const NAMESPACE_LABELS_PROPERTIES = "/etc/config/labels/namespace-labels.properties"
```
Bobgy, line 62: it's not a good practice to hard-code config file location, it's reasonable to specify a default location, but the default location should typically be overridable from a command line argument.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -616,15 +624,59 @@ func removeString(slice []string, s string) (result []string) {
...
+}
+
 func updateNamespaceLabels(ns *corev1.Namespace) bool {
 	updated := false
+
 	if ns.Labels == nil {
 		ns.Labels = make(map[string]string)
 	}
+
 	for k, v := range kubeflowNamespaceLabels {
```
Bobgy, line 666: Note: `istioInjectionLabel` is not part of `kubeflowNamespaceLabels` map. How do you plan to let us update it?

components/profile-controller/controllers/profile_controller.go
```diff
@@ -616,15 +624,59 @@ func removeString(slice []string, s string) (result []string) {
...
-			updated = true
+		_, ok := ns.Labels[k]
+		if len(v) == 0 {
+			// When there is an empty value, k should be removed.
+			if ok {
+				delete(ns.Labels, k)
+				updated = true
+			}
+		} else {
+			if !ok {
```
Bobgy, line 675: For discussion,

Not sure what would be our intended behavior, if we start profile controller with default labels, and later decided to configure custom labels. Existing namespaces won't be updated to the new labels, because the labels already exist, so we skip the update process.

Bobgy (commented):
thank you @zijianjoy for the quick PR!
I'm doing an initial pass of review

Bobgy:
Note, if you feel current design is too complex considering my comments, you can evaluate whether https://github.com/kubeflow/kubeflow/issues/5712#issuecomment-803987717 can be a good candidate design with a simpler implementation.

zijianjoy:
> > If we don't provide namespace-labels.properties, will it still work? ConfigMap creation might fail. How do we avoid breaking default deployment (non-GCP)?
> 
> I think we should keep current manifests intact and build an overlay with the configmap, so only people who want to use the configmap can use it.

That makes sense, to confirm: we will need to create an empty file for `configMapGenerator` in `overlay/kubeflow/kustomization`, and create non-empty file `configMapGenerator` in gcp distribution, and overwrite the empty `configMapGenerator`. Correct?

components/profile-controller/controllers/profile_controller.go
zijianjoy: Thank you for the suggestion! I have switched to yaml format in new commit.

components/profile-controller/controllers/profile_controller.go
zijianjoy: If the configmap is updated, we want to read from the latest data instead of cache for label configuration. So we will need to read from VolumeMount. Is it very resource intense to read a file? If yes, what is your suggestion of reading new ConfigMap data?

components/profile-controller/controllers/profile_controller.go
zijianjoy: Make sense, I changed it to a CLI argument with default value. Thank you!

components/profile-controller/controllers/profile_controller.go
zijianjoy: I have thought about it: So the current `kubeflowNamespaceLabels` will not overwrite any existing labels, but our new feature is going to enforce addition/removal/setting of a label. So I created another layer to `enforce` labels received from VolumeMount, and keep the `kubeflowNamespaceLabels` intact.

components/profile-controller/controllers/profile_controller.go
zijianjoy: See the same comment from above, which is to **enforce** labels from VolumeMount, and only **add** label from `kubeflowNamespaceLabels` if label doesn't exist. 


> I have thought about it: So the current `kubeflowNamespaceLabels` will not overwrite any existing labels, but our new feature is going to enforce addition/removal/setting of a label. So I created another layer to `enforce` labels received from VolumeMount, and keep the `kubeflowNamespaceLabels` intact.

components/profile-controller/controllers/profile_controller.go
Bobgy: Note this is a controller, not a service. It's not a big deal to restart an instance to update configs.

Regarding live configmap refresh, I believe there is a config library -- viper which KFP is using too, it can be set-up to watch file/configmap changes, however I think that's not worth the complexity.

zijianjoy: Thank you for confirming! I think it makes sense and updated the code to reflect one-time file read. I would love to learn about viper later on.

zijianjoy:
@Bobgy I created and pushed image to `gcr.io/jamxl-kfp-dev/profile-controller:kf-v1.3-rc0` using the current change. We can test GCP deployment based on this image. Would you like to help me confirm that you are able to pull this image? Thank you!

components/profile-controller/config/overlays/kubeflow/namespace-labels.yaml
```diff
@@ -0,0 +1,3 @@
+# Below is a list of labels to be enforced.
+istio-injection: ''
```
Bobgy, line 2: Shall we comment them out? This is not default value that work for others.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -617,15 +630,48 @@ func removeString(slice []string, s string) (result []string) {
...
+
+	for k, v := range enforcedKubeflowNamespaceLabels {
+		existingValue, ok := ns.Labels[k]
+		if len(v) == 0 {
+			// When there is an empty value, k should be removed.
+			if ok {
+				delete(ns.Labels, k)
+			}
+		} else {
+			if !ok || existingValue != v {
```
Bobgy, line 646: The if seems unnecessary, no matter what, you will set this label

zijianjoy: You are right! Removed this if condition.

components/profile-controller/config/overlays/kubeflow/namespace-labels.yaml
zijianjoy: Yes I will comment the enforcement here. I will add those labels from gcp-blueprint side.

yanniszark:
@Bobgy @zijianjoy  some quick questions from going through the changed files:
- Should we have labels hardcoded in our source code AND labels in a configmap? Would it not be better to just use the ConfigMap and have the default values be the ones we have now? I think it simplifies the code quite a bit.
- What should we do about label removals? For example, a label in the profile-controller labels config is then removed. Should we address this case?
- I think the initial description is a bit outdated, the file is not a `.properties` file.
- Should we use a more unix compliant path for the config (e.g., `/etc/profile-controller/namespace-labels.yaml`)

zijianjoy:
Very good questions! @yanniszark 

> * Should we have labels hardcoded in our source code AND labels in a configmap? Would it not be better to just use the ConfigMap and have the default values be the ones we have now? I think it simplifies the code quite a bit.

I agree that it will simplify the code if we have default labels on one location only. But GCP's usecase is limited to GCP only (ASM requirement), so I am not sure how the `Kubeflow version upgrade for existing cluster` will work for other distributions. If this concern is unnecessary, I can make the change in this PR to use __configmap only__ for namespace labels.

> * What should we do about label removals? For example, a label in the profile-controller labels config is then removed. Should we address this case?

Label removal is set by `label: ''`. For example `istio-injection: ''` will remove the label `istio-injection`.

> * I think the initial description is a bit outdated, the file is not a `.properties` file.

Thank you! I updated the PR description.

> * Should we use a more unix compliant path for the config (e.g., `/etc/profile-controller/namespace-labels.yaml`)

Do you have a reference about definition of unix compliant path? I want to learn what is the pattern and benefit of unix compliant path.

zijianjoy:
@yanniszark @Bobgy 

I will start working on the following changes:

1. Use configmap only for default namespace labels, which simplify the code logic and manifest config.
2. Change the default mount path to `/etc/profile-controller/namespace-labels.yaml`

yanniszark:
> Do you have a reference about definition of unix compliant path? I want to learn what is the pattern and benefit of unix compliant path.

That's a great question. The FHS (Filesystem Hierarchy Standard) defines the directory structure that Linux distributions should follow (https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). Under this guideline, `/etc` should contain system-wide configuration files. The usual naming is `/etc/<app_name>` (e.g., `/etc/apt`). Thus, for the profile-controller, I think it should be something like `/etc/profile-controller`. The benefit of conforming to FHS is that people can reason about where the profile-controller's configuration lives (aka there are no surprises).

> Label removal is set by label: ''. For example istio-injection: '' will remove the label istio-injection.

I was wondering about that. I think it's a bit tricky doing it that way, because people would need to add extra labels to remove the ones that don't exist anymore. In addition, having a label with an empty value is valid. May I propose a different way, which is a bit inspired from kubectl's 3-way merge:

1. Save all configured label keys in an annotation, as a JSON array, if the annotation does not exist. For example `profiles.kubeflow.org/owned-labels`.
2. Apply configured labels in labels map.
3. Remove owned labels that are not in configured labels.
4. Update the namespace object with the final result.

What do you think? Also cc @Bobgy

zijianjoy:
PR is updated. I am going to review Yannis's comment for detail.

zijianjoy:
Thank you @yanniszark for the detailed explanation and reference!

> By adding a general mechanism we can accomodate GCP (and future users with such use-cases) while keeping a clean codebase. This is my goal with this suggestion.

That makes sense, thank you Yannis! I agree with the general mechanism.


> That's a great question. The FHS (Filesystem Hierarchy Standard) defines the directory structure that Linux distributions should follow (https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). Under this guideline, `/etc` should contain system-wide configuration files. The usual naming is `/etc/<app_name>` (e.g., `/etc/apt`). Thus, for the profile-controller, I think it should be something like `/etc/profile-controller`. The benefit of conforming to FHS is that people can reason about where the profile-controller's configuration lives (aka there are no surprises).

Thank you for the reference! I agree with this approach and applied this change.



> In addition, having a label with an empty value is valid. 

I checked the kubernetes definition before creating this PR: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set. The namespace label cannot be empty: `must be 63 characters or less (cannot be empty)`. Therefore I use this approach so we can add/delete namespace labels in one file.

Also, I am not sure the suggested approach is the right way to use `annotation`, and it also complicates the profile controller logic from my opinion. I agree with hearing about Yuan's thought as well @Bobgy.

Bobgy:
Yeah, I think the 3-way merge idea for labels sound interesting, but probably expand the scope of this PR too much.
So far the PR's behavior aligns with existing behavior -- add a label only when it doesn't already exist.

Shall we discuss the 3-way merge behavior in a separate issue?

components/profile-controller/config/overlays/kubeflow/namespace-labels.yaml
```diff
@@ -0,0 +1,23 @@
...
+#       Second, add this label by using `key: 'value'` and deploy.
+# 2. If your profile doesn't have this label:
+#       you can add label and value to this file and deploy.
+# Reason:
+#    Profile controller will enforce flag removal, but not enforce value replacement.
+#   
+katib-metricscollector-injection:      'enabled'
+serving.kubeflow.org/inferenceservice: 'enabled'
+pipelines.kubeflow.org/enabled:        'true'
+app.kubernetes.io/part-of:             'kubeflow-profile'
```
Bobgy, line 23: nit: generally, always add a newline at end of file. Same for other files

You can configure your IDE to automatically do that.

components/profile-controller/controllers/profile_controller_test.go
```diff
@@ -54,9 +54,51 @@ func TestUpdateNamespaceLabels(t *testing.T) {
...
+		}
+		setNamespaceLabelsFromConfig(test["current"])
 		if !reflect.DeepEqual(test["expected"], test["current"]) {
-			t.Errorf("Expect:\n%v; Output:\n%v", test["current"], test["expected"])
+			t.Errorf("Expect:\n%v; Output:\n%v", test["expected"], test["current"])
 		}
 	}
 }
+
+func TestUpdateNamespaceLabels_withLabelRemoval(t *testing.T) {
```
Bobgy, line 70: nit: suggest merging this with TestEnforceNamespaceLabelsFromConfig, and allow a test case to specify config yaml.
In go, we typically prefer table tests than separate test funcs.

components/profile-controller/main.go
```diff
@@ -17,8 +17,10 @@ package main
 
 import (
 	"flag"
+	"io/ioutil"
 	"os"
 
+	"github.com/ghodss/yaml"
```
Bobgy, line 23: nit: shall we use the official yaml package https://github.com/go-yaml/yaml?
I don't think our use-case is very specific or complex.

EDIT: ohh I realized we are directly unmarshalling into a map[string]string, maybe that's an interesting usage. If it's harder to use the official package, I'm fine with this.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -617,16 +621,23 @@ func removeString(slice []string, s string) (result []string) {
 	return
 }
 
-func updateNamespaceLabels(ns *corev1.Namespace) bool {
-	updated := false
+func setNamespaceLabelsFromConfig(ns *corev1.Namespace) {
```
Bobgy, line 624: nit: maybe we should rename now? because we no longer read config here

components/profile-controller/controllers/profile_controller.go
```diff
@@ -105,6 +101,7 @@ type ProfileReconciler struct {
 func (r *ProfileReconciler) Reconcile(request ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
 	logger := r.Log.WithValues("profile", request.NamespacedName)
+	defaultKubeflowNamespaceLabels = r.DefaultNamespaceLabels
```
Bobgy, line 104: nit: it's usually not a good idea to use global variables

shall we pass this argument explicitly?

Bobgy (commented):
Great work! @zijianjoy 

The behavior LGTM, I'm reviewing details of the PR

components/profile-controller/config/overlays/kubeflow/namespace-labels.yaml
zijianjoy: Thank you Yuan! I updated my IDE for keeping newline at the end of file.

components/profile-controller/main.go
zijianjoy: I am using this package to be consistent with existing file: profile_controller.go. Keeping this for consistency.

components/profile-controller/controllers/profile_controller.go
zijianjoy: Good point! I have changed to use argument explicitly.

components/profile-controller/controllers/profile_controller_test.go
zijianjoy: Updated to use table tests.

components/profile-controller/controllers/profile_controller.go
zijianjoy: Sounds good, I have renamed this function.

components/profile-controller/controllers/profile_controller.go
Bobgy: Did you forget to push some commits? or did I miss anything?
Looks like the code is still using the global variable `defaultKubeflowNamespaceLabels`

zijianjoy: Good point! I should use local variable instead. Updated and added README.

thesuperzapper:
@zijianjoy @Bobgy is this ready for merging?

Bobgy:
/lgtm
from my side

We are pending @yanniszark's review

Bobgy:
@zijianjoy I pushed this PR to image: gcr.io/kubeflow-images-public/profile-controller:v1.3.0-gcp

zijianjoy:
Thank you @Bobgy ! I will start experimenting the Kubeflow integration with profile controller.

zijianjoy:
Since Kubeflow v1.3.0 has been released, I suggest we merge this change if it looks good to you. @yanniszark

DavidSpek:
@zijianjoy Am I correct in assuming adding the `istio.io/rev: ""` label to namespaces would also make it easier for people to upgrade Istio on running Kubeflow installations?

zijianjoy:
> @zijianjoy Am I correct in assuming adding the `istio.io/rev: ""` label to namespaces would also make it easier for people to upgrade Istio on running Kubeflow installations?

If the cloud provider or on-prem requires changing `istio.io/rev` to migrate to new Istio control plane, yes `istio.io/rev: ""` can help with this scenario. Users deploy new control plane first, then clearing `istio.io/rev` label and adding the same label with new value. Then user is able to migrate this namespace to new Istio control plane.

DavidSpek:
From what I can see in the latest Istio docs you can use either `istio-injection: enabled` or `Istio.io/rev`, the latter then probably being the better choice. 

https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

thesuperzapper:
@yanniszark we still need your final say to get this merged

yanniszark:
@thesuperzapper my last comment asks for changes, but I should make it more clear. @zijianjoy I see the following as next steps:
- Check my review changes and tell me if you are ok with them: https://github.com/arrikto/kubeflow/tree/label-dynamic-rev. If you are, push them to your branch.
- Answer the following: Previously, you said that you wanted to use a ConfigMap in order to be able to change the configuration without restarting the profile controller. However, the current code doesn't reload the labels from disk. Should we reload the labels from disk in each sync?

Finally, please make sure that your branch doesn't have any merge commits, like it has now.

zijianjoy:
Thank you @yanniszark for following up on this PR!

> Check my review changes and tell me if you are ok with them: https://github.com/arrikto/kubeflow/tree/label-dynamic-rev. If you are, push them to your branch.

It makes sense and I have cherry-pick your commit. One question: You moved `setNamespaceLabels()` call to one line below `logger.Info("List of labels to be added to found namespace", "labels", ns.Labels)`. What is the reason of doing so? If we make this move, the log will not show what labels will eventually look like for namespaces. 

> Answer the following: Previously, you said that you wanted to use a ConfigMap in order to be able to change the configuration without restarting the profile controller. However, the current code doesn't reload the labels from disk. Should we reload the labels from disk in each sync?

I will discuss with Yuan @Bobgy about this and respond back.

zijianjoy:
@yanniszark Hello Yannis, I have updated PR to read disk file every time profile controller reconciles.

yanniszark:
@zijianjoy thanks a lot! I have prepared a review branch with the final changes needed: https://github.com/arrikto/kubeflow/tree/label-dynamic-rev

Those are:
- Remove all merge commits from PR branch. I've done that for you and wrote a nice commit message.
- Watch the config file with fsnotify and trigger reconciliations if it's changed. In my testing, K8s took ~17 seconds to propagate the ConfigMap. This change is in a separate commit, which contains all review changes.
- Exit if the namespace labels file is not found. The current code proceeds with an empty list. This change is in a separate commit, which contains all review changes.

If you agree with the changes, merge the review commit with the main commit and push to your PR branch.

zijianjoy:
@yanniszark Thank you Yannis for introducing fsnotify to watch configMap update! I adopted your changes in this PR.

I think profile controller doesn't have to exit if there is no configMap exists. But since we have a default configMap file for namespace label, I am okay with this behavior.

Feel free to let me know if the current PR branch commit history needs to be updated. I usually squash and merge all commits into one PR when merging to main branch, and avoid overwriting PR branch commit history by force pushing to PR branch. But happy to hear for best practice on submitting PRs.

zijianjoy:
Following up with the last comment: I have rebased main branch and squashed the PR.

yanniszark:
Thanks @zijianjoy! 
/lgtm
/approve

PR #5764, opened 2021-03-24
web-apps: Introduce an APP_SECURE_COOKIES env var

### LINKED ISSUE #5763: Can't create Notebooks from web app when port-forwarding ISTIO gateway
/kind bug

1. Port forward the Istio ingress gateway and access Kubeflow over localhost
2. Navigate the the Notebooks Management UI
3. Click `LAUNCH` in the form and try to spawn a Notebook

After this I got an error due to CSRF header missing in the POST request.

I would expect the app to be able to work, if accessed over localhost.

This is because right now we always use [Secure](https://github.com/kubeflow/kubeflow/blob/05ba7a65ad77cdea718275bdefead6c88cf50515/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/csrf.py#L72) attribute when setting the CSRF cookie. This means that the cookie will not be set in the browser, if the request happens over http.

Here's also the relevant MDN info:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#Secure
```
Note: Insecure sites (http:) can't set cookies with the Secure attribute (since Chrome 52 and Firefox 52).
```

cc @yanniszark @elikatsis

kimwnasptd:
closes #5763 

In this PR I:
1. Introduce a new `APP_SECURE_COOKIES` as described in the issue above
2. Switched the CSRF check orderings. It's misleading for the app to first report that the CSRF header is missing when the cookie is missing as well. In that case the end user might believe that the cookie is there, which could not be the case.
3. Extended the makefiles to use this env var when we run the backends in dev mode, since in that case the apps are exposed over http

/cc @yanniszark @elikatsis 
/assign @kimwnasptd

### DIFF
--- a/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/csrf.py
+++ b/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/csrf.py
@@ -42,7 +42,7 @@ import secrets
 from flask import Blueprint, current_app, request
 from werkzeug.exceptions import Forbidden
 
-from . import config
+from . import settings
 
 bp = Blueprint("csrf", __name__)
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


elikatsis:
Thanks @kimwnasptd ! I'll take this

saumilsdk:
Here is how i did the fix:
image

saumilsdk:
Refer: 
https://github.com/kubeflow/website/pull/3284/files#diff-a2f232f8c669408c84e44bf7ada09b2369eca1e1d1bcbb91f1811b44bd441eec

PR #5765, opened 2021-03-24
Update manifests to use ECR and fix fieldPath in kustomization files

DavidSpek:
Some of the manifests were still using gcr and didn't contain an `images:` section in the `kustomization.yaml` making it more difficult to update the image tags when new ones are pushed (Renovate can do this as well). This PR fixes the manifests so all components are using ECR, have an `images:` section in the `kustomization.yaml` and have the current latest image tags. 

I've added a fix to this PR changes `fieldPath` to `fieldpath` in the `kustomization.yaml` files according to the Kustomize reference docs: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/

/cc @kimwnasptd @yanniszark

### DIFF
--- a/components/admission-webhook/manifests/base/deployment.yaml
+++ b/components/admission-webhook/manifests/base/deployment.yaml
@@ -9,7 +9,7 @@ spec:
         sidecar.istio.io/inject: "false"
     spec:
       containers:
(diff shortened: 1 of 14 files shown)


DavidSpek:
@yanniszark Sounds good to me. For reference, I was initially getting the `fieldpath` info from https://json.schemastore.org/kustomization.

yanniszark:
Thanks David!
/lgtm
/approve

PR #5768, opened 2021-03-24
(manifests): Remove unused params that break newer kustomize for web apps

DavidSpek:
The `params.yaml` files are not used anymore for the various web apps and break using newer versions of Kustomize. This PR removes the `params.yaml` from the Jupyter, Tensorboards and Volumes Web Apps and the Central Dashboard manifests. 

/cc @yanniszark


yanniszark:
Thanks @DavidSpek 
/lgtm
/approve

PR #5770, opened 2021-03-24
Amend the AutoML entry in the Central Dashboard

### LINKED ISSUE #5750: AutoML entry in the CentralDashboard
/kind feature

### Why you need this feature:
We want to define the story around how to expose AutoML `Experiments` in Kubeflow. This is in continuation to the discussion from https://github.com/kubeflow/kubeflow/pull/5739#issuecomment-803622218.

In the first iteration #5739 we included AutoML under the `Experiments` section of the Central Dashboard like this
```
Notebooks
Volumes
Tensorboards
Experiments
  Pipelines
  AutoML
```

The guideline we want to enforce for the Central Dashboard is to have entries that correspond to nouns that the users can create/delete/connect-to/use. This is why we've exposed the underlying CRD objects like `Notebooks`, `Volumes`, `Tensorboards`, `Pipelines`, `Runs` etc. This way Kubeflow provides a flat list of objects the users could manipulate, to perform their ML related tasks, and acts as a cohesive platform, instead of just grouping functionalities/components together.

The issue here is that we have two distinct components [ Pipelines, AutoML ] using the same name, `Experiments`, but for different objects. 

### Describe the solution you'd like
I can see two solutions going forward, for integrating the Katib UI with the Central Dashboard, that move `AutoML` at the root level:

**Option A**
1. Add a top level `AutoML` entry
2. work on renaming the underlying CRDs from Katib after 1.3 to not use the `Experiment` name
    * this could also help on having separate CRDs for `HP Tuning`, `NAS` or other new AutoML components
    * the different components can have their own separate versions, so HP Tuning can be `v1` while NAS can be `v1alpha1`

**Option B**
1. Add a top level `Experiments (AutoML)` entry at the top level
2. Also have a top level `Experiments (KFP)` entry at the top level

<details><summary>How it looks</summary>
<p>

Screenshot from 2021-03-22 22-19-53

</p>
</details>

WDYT @andreyvelich @gaocegege @johnugeorge 

### Anything else you would like to add:
These guidelines of the Dashboard does not limit us in any way on how we want the app to behave in standalone mode. In standalone mode it can have a hamburger menu on the top left and a sidenav and organize the entries however we want there.

kimwnasptd:
closes #5750 

This PR:
1. Adds an `Experiments (AutoML)` entry in the CentralDashboard
2. Renames the `Experiments` entry to `Experiments (KFP)` 
3. Adds a slightly modified svg for Katib to make it aligned with the other icons in the Central Dashboard

Screenshot from 2021-03-24 19-55-30

The credits for the icon goes to @konpal. We will also open a PR to add this icon to the `kubeflow/katib` repo as well.

Let's also keep on iterating on our story around the Central Dashboard and how we want to expose our apps after 1.3.

/cc @andreyvelich @gaocegege @johnugeorge @StefanoFioravanzo


andreyvelich (commented):
Thank you for the update @kimwnasptd!
/lgtm
/hold for the review.
/cc @gaocegege @johnugeorge @DavidSpek

andreyvelich:
Let's merge this, since we have to make a release.
/hold cancel

PR #5773, opened 2021-03-25
Tensorboard Controller: Explicitly set scope to Namespaced in manifests

DavidSpek:
While the default scope is `Namespaced`, it is probably best to set it explicitly in the CRD. Not doing so will make the manifest somewhat incompatible with GitOps tools. 

/cc @kimwnasptd @yanniszark

### DIFF
--- a/components/tensorboard-controller/api/v1alpha1/tensorboard_types.go
+++ b/components/tensorboard-controller/api/v1alpha1/tensorboard_types.go
@@ -50,6 +50,7 @@ type TensorboardStatus struct {
 }
 
 // +kubebuilder:object:root=true
+// +kubebuilder:resource:path=tensorboards,scope=Namespaced
 // +kubebuilder:subresource:status
 
 // Tensorboard is the Schema for the tensorboards API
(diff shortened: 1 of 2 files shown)


yanniszark:
@DavidSpek I believe the CRD is generated by `controller-gen`, as per `kubebuilder`'s layout.
I don't think we should edit a generated file. Let's first find out why kubebuilder doesn't generate it.


Can you also elaborate on:

> Not doing so will make the manifest somewhat incompatible with GitOps tools.

DavidSpek:
@yanniszark I was just thinking that as well. We should make sure Kubebuilder sets it to avoid issues in the future. 

Because the source of truth it the GitRepo, and when applied in the cluster the default value is used in the CRD, ArgoCD picks it up as a diff and wants to try and sync it. Also, from a declarative standpoint it is probably just better to explicitly declare it. 

image

DavidSpek:
@yanniszark It looks like no `commonLabels` are set in `config/default/kustomization.yaml` and no selector is being used for the tensorboard-controller service. Should this be quickly addressed before merging https://github.com/kubeflow/kubeflow/pull/5776?

PR #5776, opened 2021-03-25
Release v1.3.0-rc.0 for kubeflow/kubeflow

yanniszark:
Release version `v1.3.0-rc.0` to give to the manifests WG for syncing:
https://github.com/kubeflow/manifests/issues/1775


yanniszark:
/hold
for reviews

yanniszark:
I see that Prow tests don't run for the `v1.3-branch`. Pinging @PatrickXYS

PatrickXYS (approved):
A quick question, I assume we hardcode the image tag in v1.3-branch when building images, I was thinking about this model can also fit into other WGs' use case. 

/lgtm

DavidSpek:
Looks good to me. Somewhat related to this PR is that I think I found and issue with the tensorboard-controller. Not sure if that should be fixed before this PR or after. It doesn’t cause an issue with the current set of manifests, but the selector for the service and the app labels are missing.

PR #5777, opened 2021-03-25
releasing: Add WG-Notebooks leads to release owners

yanniszark:
cc @Bobgy 

This is in order to be able to release new manifests / image versions for the notebooks wg.


yanniszark:

PR #5782, opened 2021-03-26
Take [[disabled]] parameter into account on registration page namespace input

d-gol:
Namespace input element parameters for registration page are defined on [registration-page.pug#L17](https://github.com/dejangolubovic/kubeflow/blob/1b2cbc0de947b49b0169df935b4bf19b72d091f0/components/centraldashboard/public/components/registration-page.pug#L17) and [md2-input.js#L15](https://github.com/kubeflow/kubeflow/blob/096d31783c0924d0a1e8fd9606feaae203711b0f/components/centraldashboard/public/components/resources/md2-input/md2-input.js#L15). Then, parameters defining the input element are used on [md2-input.pug](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/public/components/resources/md2-input/md2-input.pug). Parameter [[disabled]] is missing in this configuration, which this PR is fixing.


solarist:
Any updates on this? This is a very handy feature for us to have.

PR #5783, opened 2021-03-26
WIP: pvcviewer-controller and PVCViewer pod

DavidSpek:
This PR is adds a new controller that is intended to be used by the new Volumes Web App that was added by https://github.com/kubeflow/kubeflow/pull/5684. It is essentially a copy of the tensorboard-controller with a few slight modifications. 
When a user creates a PVCViewer CR, a pod is launched that includes [File Browser](https://github.com/filebrowser/filebrowser) that allows the user to download or upload files and folders to a PVC ([Dockerfile for this image](https://github.com/kubeflow/kubeflow/pull/5783/files#diff-5a6876e021df86046dbfd93af877964cf9f96d837edbfe3ab4cb5a03bcd82162)). To overcome ReadWriteOnce limitations, if a PVCViewer is scheduled for a PVC that is already mounted to a Pod, the PVCViewer pod is scheduled to the same node (this is also how the tensorboards-controller functions).

Leader election needed to be removed as this was causing conflicts with the tensorboard-controller. I'm not sue if this is the correct method to create compatibility so this might need to be improved upon. 

To test the controller you can run `make install` to install the CRD to the cluster and then `make run` to run the controller locally. Then edit the sample in `config/samples` to a PVC that exists and apply it (an already mounted RWO PVC doesn't work in when running the controller locally like this). You can access the UI at `<kubeflow_domain>/pvcviewer/<namespace>/<name_of_viewer>/`.

Sample of what the Volumes Web App will look like together with this controller:
vwa-pvcviewer-opt6

The next step would be to add support to the Volumes Manager UI to create a PVCViewer resource. This would be similar to how the Rok version of the UI implements this, or the default and Rok code could (mostly) be consolidated if Arrikto chooses to use this for Rok as well. 


RFMVasconcelos:
This looks like a very cool feature @DavidSpek! 

Thank you for making Kubeflow better :)

RFMVasconcelos:
@DavidSpek this looked like a cool feature! 

/open

TobiasGoerke:
We introduced this extensions to our cluster more than a year ago and we've seen this feature becoming a vital part in our users' Kubeflow experience. 

Thanks @DavidSpek for your great work.

As this PR got closed automatically: is there still interest in the filebrowser? If so, I'd volunteer to migrate all changes to Kubeflow v1.5/v1.6.

harish422:
@DavidSpek This is a very useful feature. thank you for coming up with this . Are there still any plans of merging this code ? I the moment i noticed this i was checking 1.6.1 code base and didn't find it. Later realized this han't been checked in yet.

Thanks,
Harish

PR #5804, opened 2021-04-05
Update notebook server base images

DavidSpek:
Small update for the base images of each server type.

### DIFF
--- a/components/example-notebook-servers/codeserver/Dockerfile
+++ b/components/example-notebook-servers/codeserver/Dockerfile
@@ -1,9 +1,9 @@
-FROM public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/base:master-4892c583
+FROM public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/base:master-1831e436
 
 USER root
 
 # args - software versions
-ARG CODESERVER_VERSION=3.9.1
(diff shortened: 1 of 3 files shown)


DavidSpek:
@kimwnasptd Can you quickly `lgtm` this so I can continue with the downsteam images?

kimwnasptd:
Sure! We will also need to start discussing our long term plan for how to manage the base image tag.

@DavidSpek could you start and issue for this with some proposed next steps so we could start iterating on?

DavidSpek:
@kimwnasptd I'll create an issue for this soon. What is actually blocking this effort is the fact that the ECR public gallery doesn't have a public API endpoint that lists the latest image tags, making it incompatible with basically any dependency automation tool. See: https://github.com/aws/containers-roadmap/issues/1262. I've contacted @rarkins from Renovate about this already, but I think we'd need to get in touch with somebody from AWS to fix this as well. Another option would be to not use ECR, which by default has the same type of user pull rate limiting as dockerhub. FYI, dockerhub has a process to remove the rate limiting for open-source projects (for clarity, the repo isn't limited but rather anonymous pulls, but this can be disabled).

kimwnasptd:
Thanks for the update @DavidSpek! Lets iterate more on our options on your new issue then.

/lgtm
/approve

Issue #5525, opened 2021-01-18
crud-web-apps use deprecated Angular 8


DavidSpek:
/kind bug
/area jupyter

The crud-web-apps, as well as the `/components/jupyter-web-app` that doesn't use the common code are using Angular 8, which is deprecated and is not receiving security patches. The current version of Angular is 11 (12 has been released now), with 9 being the last still receiving updates until August 6th, 2021. I've started an attempt to update the crud-web-apps to Angular 9 (it is not recommended to skip multiple versions at once) which went relatively smoothly for the common code. However, when updating the Jupyter web app errors arise related https://v9.angular.io/guide/migration-injectable and https://v9.angular.io/guide/migration-undecorated-classes. These errors appear during migration of the Jupyter web app and are all related to the common frontend code.

kimwnasptd:
Thank you for starting this effort @DavidSpek.

Having this done in parallel can help us save time for when we decide we have to move on from Angular 8, but this is not the case in the immediate future. The reasons why don't want to commit huge effort on this right now are:
1. There are other UIs in the pipeline that will come within the month like the Volumes web app and the new [Katib UI](https://github.com/kubeflow/katib/issues/1322#issuecomment-762223386), which will be using this stack and common code
2. We don't have solid tests that can test if our images can be built or that, at least, some basic functionality is being tested
3. While Angular 8 will not receive security fixes in the future this doesn't mean that right now we have a hot potato of a security issue in our hands that we need to solve immediately
4. I don't have the cycles to review this until the release of 1.3 

But I agree that this is something we will have to invest cycles in the future. So if you'd like to start this effort in parallel with 1.3 release then you are more than welcome to. But we will start thinking about merging and investing more cycles in this at least afterwards 1.3. 

Also, if you plan on continuing, please document in this issue what difficulties/challenges or changes you had to make for supporting Angular 9. This will help us better grasp the effort required as well as potential pitfalls.

DavidSpek:
@kimwnasptd I completely agree with you, and I was aware that this effort would also influence the new UIs that are being released and should be done all at once. Also, just to be clear I'd like to note that my vision of updating Angular is not to go to Angular 9, but to go to whatever the latest stable version is at that point in time. 

[...]

PR #5805, opened 2021-04-05
Old: Update CRUD web apps from Angular 8 to Angular 11 or 12

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5525

Angular 8 has been deprecated for a while now, with Angular 11 being the latest version. Sadly, you cannot skip version so the process to update from 8 to 11 was to first update the common code to version 9, then the web apps and confirm they are working properly. Then again for version 9 to 10 and once more for 10 to 11. 

During the update process from Angular 8 to 9 of the web apps some errors are shown that don't seem to make much sense. From what I can find this is due to the use of `npm link` to the common frontend code and this bug has been fixed in Angular 10. For this reason, I ignored the errors and tested the web app functionality manually. As no errors were logged and nothing seemed to be broken I continued. I manually tested the web apps after each update and saw no errors or problems.

I've added the commands executed for (almost) every commit in the description. To prevent getting stuck with outdated software for longer than necessary, and given the inevitable size of this PR, I would like to suggest that this PR be reviewed by manually running the resulting web-apps and confirming nothing out of the ordinary is seen there. After this PR has been merged the next step would be to look more closely into what features are now deprecated in Angular and what new features replace them. A useful overview for this can be found [here](https://update.angular.io/?l=3&v=8.2-11.0).

/cc @kimwnasptd @StefanoFioravanzo @thesuperzapper @elikatsis @yanniszark


DavidSpek:
I'm guessing the failure of [this test](https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-common-ui-5805-a05f26d-1456-6b02?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-common-ui-5805-a05f26d-1456-6b02-1029517815&sidePanel=logs%3Akubeflow-kubeflow-presubmit-common-ui-5805-a05f26d-1456-6b02-1029517815%3Amain) is caused by a need to update the tests for the newer version of Angular.

DavidSpek:
The last commit does seem to solve the initial problem with the common UI tests. Locally the test runs successfully and consistently. However, in the CI the common UI test is failing with a different unrelated error each time.

/test kubeflow-kubeflow-presubmit
--

thesuperzapper:
@kubeflow/wg-notebooks-leads @DavidSpek I think this should be merged once we finalise the 1.3 release version, (ETA 19th April).

@kimwnasptd realistically, there is no way to test this beyond just deploying it, and checking. 
I think the focus should be on ensuring it "works" the same as it did with angular 8, and deal with stylistic refactoring AFTER in followup PRs. (Especially because we are massively out of date, and thats a security issue)

PR #5809, opened 2021-04-06
Update second layer docker images to new tags and some python deps

DavidSpek:
Updates the downstream notebook server images to the tags created after merging https://github.com/kubeflow/kubeflow/pull/5804. Also updates some python packages in the scipy notebook server image. 
@kimwnasptd if you could quickly give your blessing on this PR I can move on to the last layer of downstream images (tensorflow and pytorch full).

### DIFF
--- a/components/example-notebook-servers/codeserver-python/Dockerfile
+++ b/components/example-notebook-servers/codeserver-python/Dockerfile
@@ -1,4 +1,4 @@
-FROM public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/codeserver:master-53099bbd
+FROM public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/codeserver:master-ebc0c4f0
 
 USER root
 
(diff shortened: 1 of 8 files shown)


kimwnasptd:
@DavidSpek I see that you changed, the minor, versions for some packages in `components/example-notebook-servers/jupyter-scipy/requirements.txt`. Is there a process for deciding on which packages will get updated and to which versions?

DavidSpek:
@kimwnasptd As the images and dockerfiles are examples my philosophy is always update to everything to the latest version, with small exceptions where reasonable or needed. For example, the latest kfp and kfserving packages can't be installed together yet due to a dependency conflict (this will be resolved with the kfp 1.5.0 package). There is a similar situation with the PyTorch full package (and a bug that was encountered in the latest version of PyTorch), which is why I haven't updated PyTorch yet. I've created issues upstream to get this resolved a while ago so hopefully it is implemented soon, after which I will create a PR for the updates.

Related to this, for compatibility reasons Python is using version 3.8.8 (latest 3.8.x on conda-force) to avoid a potential large amount of compatibility issues when moving to Python 3.9.2. So I think it is best to update to Python after some more time has passed.

PR #5819, opened 2021-04-08
tensorboard-controller: Fix scheduling unbound PVCs

elikatsis:
When the TB controller attempts to schedule a RWO PVC it checks its
accessModes in the PVC status. The controller panics if the list is
empty.

This commit adds a check to ensure the list is not empty.

Signed-off-by: Ilias Katsakioris <elikatsis@arrikto.com>

/assign @StefanoFioravanzo


jasonliu747 (approved):
/lgtm

PR #5811, opened 2021-04-06
Add CI format checks for the Jupyter web app

kimwnasptd:
first step for #5810 

Fix format errors for the web apps and introduce new tasks that check the format of the code on each PR. I'll be using `prettier` for the frontend code and `flake8` for the Python backend.

I bumped into the following issues and included commits for them:
1. Kaniko was not properly using the dockerignore file in each app's repo. To migitage this I created a global dockerignore file for all the apps
2. The above issue with the dockerfile resulted in some erros when building the frontend. I think this was because kaniko would include the `node_modules` [no dockerignore] it would find, which were build from another task. I modified the docker file to not copy the node modules which seems to solve the issue, in conjunction with the updated dockerignore file
3. I fixed the formatting for the frontend and the backend in order for the updated CI tests to pass

As we can see now we have tasks that also check the formatting of the Jupyter web app's frontend and backend code. I'll also move forward with subsequent PRs to update the other web apps as well.

Issue #5810, opened 2021-04-06
Extend CI tests to catch format errors on the web apps


kimwnasptd:
/kind feature

**Why you need this feature:**
To ensure the code of our [crud-web-apps](https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps) properly formatted.

**Describe the solution you'd like:**
Extend the existing CI python files to also run a task that will be checking the formatting of TypeScript as well as run `flake8` tests on the backends of the [crud-web-apps](https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps).

/priority p1
/assign @kimwnasptd

DavidSpek:
@kimwnasptd Something I've noticed is that all the presubmit tests run as a single `presubmit` test. I think it would be better to split the tests so that you can `/retest` just the `common-ui-tests` if it fails for example and not also have it run the volumes-web-app image build test as well. 
This way, you easily see in GitHub that the linting presubmit test failed and that functionally everything is fine, and you can retest only the one that failed as described above.

kimwnasptd:
This would make sense. 

@PatrickXYS if we had multiple presubmit jobs defined in the `prow_config.yaml` and some of them failed, then if we would type the `/retest` would that make only the failed argo workflows rerun?

DavidSpek:
@kimwnasptd `/retest` will retest all, but you would have the option to run `/retest <failed-job>` with that setup. Now you can do `/retest presubmit` which re-runs all the tests as the presubmit contains all the tests.

kimwnasptd:
> Now you can do /retest presubmit which re-runs all the tests as the presubmit contains all the tests.

IIRC right now even if we define multiple presubmits jobs in the `prow_config.yaml` these will end up being run/managed under a single Prow Job, which will spawn multiple Argo CRs for each presubmit job in that file. So if we do `/retest` or `/retest <failed-job>` then this will just rerun the entire failed Prow Job, not just the failing underlying Argo Workflows that are managed from the Prow Job.

@PatrickXYS is my understanding accurate?

PatrickXYS:
> IIRC right now even if we define multiple presubmits jobs in the prow_config.yaml these will end up being run/managed under a single Prow Job, which will spawn multiple Argo CRs for each presubmit job in that file. So if we do /retest or /retest <failed-job> then this will just rerun the entire failed Prow Job, not just the failing underlying Argo Workflows that are managed from the Prow Job.

[...]

PR #5820, opened 2021-04-08
Add CI format checks for the Volumes web app

kimwnasptd:
another step for #5810

This PR is very similar to #5811:
* Adds CI tasks for format checking the Volumes web app's code
* Fixes the formatting of the web app
* Removes the unused dockerignore, since we introduced a global one for all the apps in #5811


components/crud-web-apps/volumes/Dockerfile
```diff
@@ -21,11 +22,18 @@ RUN npm run build
...
-COPY ./volumes/frontend/ .
+COPY ./volumes/frontend/package.json ./
+COPY ./volumes/frontend/package-lock.json ./
+COPY ./volumes/frontend/tsconfig.json ./
+COPY ./volumes/frontend/tsconfig.app.json ./
+COPY ./volumes/frontend/tsconfig.spec.json ./
+COPY ./volumes/frontend/angular.json ./
+COPY ./volumes/frontend/src ./src
+
+ENV NG_CLI_ANALYTICS "ci"
```
PatrickXYS, line 34: Seems like Line14 and Line34 are the same.

kimwnasptd: Since this is a multi-stage docker build will the ENV var persist between stages?

kimwnasptd:
After a lot of failing tests this should also be ready for review.

I'm still not able to get to the root cause of the issue. The tests were failing because the `kaniko-build` task reported that `node_module` files from the context where removed. Here's a relevant log:
```
INFO[0041] Resolving srcs [common/frontend/kubeflow-common-lib/package*.json]...
INFO[0051] COPY ./common/frontend/kubeflow-common-lib/package*.json ./
INFO[0051] Resolving srcs [common/frontend/kubeflow-common-lib/package*.json]...
error building image: error building stage: failed to execute command: resolving src: failed to resolve sources: resolving sources: open
/mnt/test-data-volume/kubeflow-kubeflow-presubmit-vwa-tests-5820-8644ad9-9792-6a8a/src/kubeflow/kubeflow/components/crud-web-apps/volumes
/frontend/node_modules/@babel/helper-define-map: stale NFS file handle
```

This has definitely something to do with the fact that the `node-modules-install` task runs in parallel with the `kaniko-build` one. 

In the end I had to change any `COPY ./package*.json` commands that contained an asterisk * pattern match and copy the files one by one instead. This made the tests pass for VWA, whose tests were continuously failing.

components/crud-web-apps/volumes/Makefile
```diff
@@ -1,11 +1,9 @@
...
-DOCKERFILE=volumes/Dockerfile
+IMG ?= public.ecr.aws/j1r0q0g6/notebooks/volumes-web-app
+TAG ?= $(shell git describe --tags --always)
+DOCKERFILE ?= volumes/Dockerfile
 
 docker-build:
-	cp Dockerfile.dockerignore ../.dockerignore
-	-cd ../ && docker build -t ${IMG}:${TAG} -f ${DOCKERFILE} .
-	rm ../.dockerignore
+	cd ../ && docker build -t ${IMG}:${TAG} -f ${DOCKERFILE} .
```
elikatsis, line 6: This part is a bit weird. I mean mentioning paths as if we are in a different directory because we will be using them in a different context (after `cd`ing).

May I suggest we
1. Drop the `DOCKERFILE` variable (I think we shouldn't be messing with it) OR make it relative to this directory
   I.e.,
   ```
   DOCKERFILE=Dockerfile
   ```
2. Drop the `cd`ing part from the Makefile target and use the parent directory as the path for the Docker building context (https://docs.docker.com/engine/reference/commandline/build/#build-with-path)
   This will make the Makefile target as follows
   ```Makefile
   docker-build:
       docker build -t ${IMG}:${TAG} -f Dockerfile ..  # Or "-f ${DOCKERFILE}" depending on the decision of (1)
   ```

I understand the reason we are `cd`ing to the parent directory is to use a single `.dockerignore` file for all the web apps.
I tested locally that we maintain this behavior with the change I proposed, so there should be no problem. Let's also have others verify this as well (you, the CI, other?).

Then, we can follow up changing this for the other web apps as well (this is not a blocker for the release of course).

elikatsis (commented):
I left a comment. Rest look good, thanks!

components/crud-web-apps/volumes/Dockerfile
PatrickXYS: Oh that's multi-stage, I didn't notice at the beginning.

components/crud-web-apps/volumes/Makefile
kimwnasptd: @elikatsis ACK, this is a very good suggestion!

I'll add a commit for Volumes web app in this PR, since it's fixing the formatting for this component. I'll then open up a new issue to keep track of the uniformity in the dockerfiles for the other web apps as well.

kimwnasptd:
@elikatsis this PR should be ready to get merged. I'll then proceed to create an issue wrt uniform Makefiles by using Docker context, instead of `cd ..`, for the web apps

elikatsis:
Thank you @kimwnasptd !

/lgtm
/approve

Issue #5813, opened 2021-04-07
[profile-controller] profiles CRD does not preserve unknown fields in plugins.spec


zijianjoy:
Higher KF 1.3 distribution issue: https://github.com/kubeflow/manifests/issues/1798

## UPDATE 4.8

We identified root cause, explained in https://github.com/kubeflow/kubeflow/issues/5813#issuecomment-815450791.
This issue affects both GCP and AWS plugins and should block the release.

## Previous status

Currently we are debugging an issue on GCP when profile controller attempt to set IAM policy to a profile, it will crash because of strange log. 

## Error message

```
2021-04-06 20:03:30.655 PDT2021-04-07T03:03:30.655Z INFO controllers.Profile Start to Reconcile. {"profile": "/kubeflow-user", "namespace": "", "name": "kubeflow-user"}
Error
2021-04-06 20:03:30.655 PDT2021-04-07T03:03:30.655Z INFO controllers.Profile List of labels to be added to namespace {"profile": "/kubeflow-user", "labels": {"app.kubernetes.io/part-of":"kubeflow-profile","istio.io/rev":"asm-192-1","katib-metricscollector-injection":"enabled","pipelines.kubeflow.org/enabled":"true","serving.kubeflow.org/inferenceservice":"enabled"}}
Error
2021-04-06 20:03:30.655 PDT2021-04-07T03:03:30.655Z INFO controllers.Profile Updating Istio AuthorizationPolicy {"profile": "kubeflow-user", "namespace": "kubeflow-user", "name": "ns-owner-access-istio"}
Error
2021-04-06 20:03:30.673 PDT2021-04-07T03:03:30.672Z INFO controllers.Profile Updating RoleBinding {"profile": "kubeflow-user", "namespace": "kubeflow-user", "name": "namespaceAdmin"}
Error
2021-04-06 20:03:30.677 PDT2021-04-07T03:03:30.677Z INFO controllers.Profile No update on resource quota {"profile": "/kubeflow-user", "spec": "&ResourceQuotaSpec{Hard:ResourceList{},Scopes:[],ScopeSelector:nil,}"}
Error
2021-04-06 20:03:30.684 PDT2021-04-07T03:03:30.684Z INFO controllers.Profile Patch Annotation for service account: {"profile": "kubeflow-user", "namespace ": "kubeflow-user", "name ": "default-editor"}
[...]

PR #5822, opened 2021-04-08
Fix profile-controller CRD pruning issue

yanniszark:
Closes https://github.com/kubeflow/kubeflow/issues/5813

@Bobgy @zijianjoy I took a look at the controller-gen and kubebuilder docs for embedding resources.
I added the `// +kubebuilder:pruning:PreserveUnknownFields` marker accordingly and regenerated manifests.
Could you please test and tell me if it also works for you?


Bobgy:
@yanniszark I think you need to regenerate the crd. Did I miss anything?

yanniszark:
> @yanniszark I think you need to regenerate the crd. Did I miss anything?

I think you missed it 😛 
I may be mistaken, but this is what you mean right?
https://github.com/kubeflow/kubeflow/pull/5822/files#diff-4cceae2882a85453456ceab092e7c9cc138e55b2cc907437c88e7f2d96f25c48

Bobgy:
Ohhh, you are right!
That's exactly what I wanted
/LGTM

zijianjoy:
Thank you Yannis for the fix!

Issue #5821, opened 2021-04-08
Logo Trademarks: Remove logos from JWA


yanniszark:
/kind bug
### Problem Statement

The new Jupyter Web App comes with logos for Jupyter, VSCode and RStudio. The logos for RStudio and VSCode come with complicated guidelines and need for explicit written legal permission, which is a process the community doesn't have a method of handling from a legal/trademark standpoint. Likely we will need to ask Google to handle this on behalf of the community.
This is a problem for vendors/platforms building their distributions, as they need to be careful what they include. 

### Proposed Solution

For these reasons, we should remove the logos from JWA and use text instead.

cc @kubeflow/wg-notebooks-leads 
cc @bobgy @castrojo @theadactyl

DavidSpek:
I've already been discussing this issue with @Bobgy and am busy with providing a solution for this. I do want to point out that an agreement with RStudio for the use of their logo is being worked on, but that would be for the Kubeflow project and not all the distributions. If a distribution that is just providing manifests that build on the Kubeflow manifests are actually redistributing the logos is questionable in my eyes (but distributions like EKF that are selling the actual distribution would definitely need a licensing agreement with RStudio).

yanniszark:
@DavidSpek @Bobgy thanks for looking into this. If possible, could you or bobgy fill the rest of us in on the plan in this issue? I'd like to make sure we're synced on this.

DavidSpek:
@yanniszark Is it alright if I do this tomorrow? I've been at it for 10 hours without a break again today so I'm pretty tired now and this will be an involved explanation.

ReggieCarey:
Equally problematic is removing attribution for a software product. If the vscode and rstudio logos are part of their distribution, what justification do you have for removing them?  If the software was not written by either company then substitute a kubeflow or other logo and let kf installers customize for their environments.

DavidSpek:
@ReggieCarey My original plan was to make the logos in the UI configurable, and I will try and get that implemented tomorrow. The issue here is that Kubeflow and anybody installing it themselves can most likely use the logos without any problems, but companies offering distributions cannot. The Kubeflow project can have permission to include the logos, but distributions provided by companies will likely need to make their own agreements with the given parties. But I do think that distributions shouldn't restrict agreements the Kubeflow project makes with third parties. It should be made easy for distributions to configure the logos in this case, but I don't think they should be outright removed just because it means the distributions need to put in minimal effort for this.

[...]

PR #5823, opened 2021-04-09
Remove Logo Trademarks from the Jupyter Web App and make logos configurable

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5821

This PR removes all references to the VS-Code and RStudio trademarks and replaces them with a generic `Gourp One` and `Group Two`. To make the logos used in the Jupyter Web App and Notebook Spawner UI configurable, a new configmap is added to the manifests that contains the logos and icons for the three server types (JupyterLab, Group One, Group Two). This way, distributions maintainers or users have the freedom to easily use whatever logos they want for the three server types and they can easily update the logos without needing to build a new container image for the Jupyter Web App. As the deployment manifests have changed the manifests in the manifests repository will need to be updated. 

@Bobgy ~The configmap currently still contains the VS-Code and RStudio logos and icons by default. Do these also need to be removed, or can the distributions handle this themselves with an overlay?~
To ensure this PR can get merged quickly, I have replaced the VS-Code and RStudio logos and icons with generic SVGs with `1` and `2` in them so that merging this PR is not dependent on me pushing a commit to remove these if this would be required. 

Preview of the Spawner UI with the generic SVGs in the configmap:
image

Preview of the Index with the generic SVGs in the configmap:
image

/cc @Bobgy @castrojo @yanniszark @kimwnasptd @thesuperzapper @theadactyl @kubeflow/wg-notebooks-leads @kubeflow/wg-manifests-leads @kubeflow/release-team


DavidSpek:
@kimwnasptd Can this be merged?

DavidSpek:
@yanniszark Can you maybe take a look at this so it can be merged?

kimwnasptd:
We ended up opening PRs at the same time so I had some delay reviewing this work. Thanks for the PR @DavidSpek!

I'm OK with these changes, since they allow people to easily swap the SVGs and by default use neutral ones, as a hotfix for our situation right now. But I believe there are a lot of next steps to further improve this effort, which we could target for 1.3.1 and forwards, but let's not block the release for them.

I'll open an issue to coordinate on the future improvements and keep track of things that people need to have in mind, like how to swap the SVGs, what logic gets applied to each group right now etc.

I'll move on with some small review comments I have

DavidSpek:
> But I believe there are a lot of next steps to further improve this effort, which we could target for 1.3.1 and forwards, but let's not block the release for them.

I completely agree, there are also optimizations from your PR that should be implement at a later date. Ultimately, the JWA should be configurable with an arbitrary number of groups and the UI should accommodate that accordingly (names for the groups, as well as logos/icons would then be used/needed). But this is part of a bigger effort for the JWA and isn't possible before the release.

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
```diff
@@ -25,14 +25,14 @@ spawnerFormDefaults:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda-full:master-1831e436
-  imageVSCode:
-    # The container Image for the user's VS-Code Server
+  imageGroupOne:
+    # The container Image for the user's Group One Server
```
kimwnasptd, line 29: Can you add a comment here that for images in this group the backend will be:
1. Adding the `notebooks.kubeflow.org/http-rewrite-uri` annotation to the CR, to rewrite the path that ends up in the running container to be `/`

This will help exposing to the users the constraints and logic that will be applied to images that belong to that group

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
```diff
@@ -25,14 +25,14 @@ spawnerFormDefaults:
...
     value: public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/codeserver-python:master-1831e436
     # The list of available standard container Images
     options:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/codeserver-python:master-1831e436
-  imageRStudio:
-    # The container Image for the user's RStudio Server
+  imageGroupTwo:
+    # The container Image for the user's Group Two Server
```
kimwnasptd, line 35: Can you add a comment here that for images in this group the backend will be:
1. Adding the `notebooks.kubeflow.org/http-headers-request-set` annotation to the CR, for setting the `X-RStudio-Root-Path: ${ISTIO_PREFIX}`  header to each request to the running container
2. Adding the `notebooks.kubeflow.org/http-rewrite-uri` annotation to the CR, to rewrite the path that ends up in the running container to be `/`

This will help exposing to the users the constraints and logic that will be applied to images that belong to that group

components/crud-web-apps/jupyter/frontend/package.json
```diff
@@ -32,10 +32,10 @@
     "@fortawesome/free-brands-svg-icons": "^5.12.0",
     "@fortawesome/free-solid-svg-icons": "^5.12.0",
     "@kubernetes/client-node": "^0.12.2",
-    "hammerjs": "^2.0.8",
     "date-fns": "^1.29.0",
-    "lodash-es": "^4.17.11",
```
kimwnasptd, line 37: Let's exclude this change from this PR and keep the touched lines to the minimum. This will also remove the 15.000 modified lines to the `package-lock.json` because of this change.

Although I'd like to progressively replace the use of `lodash` with `lodash-es`, but until I get to it we can remove this from the `package.json`, but lets just do it in another PR.

components/crud-web-apps/jupyter/frontend/src/app/types.ts
```diff
@@ -8,7 +8,7 @@ export interface JWABackendResponse extends BackendResponse {
   vendors?: string[];
 }
 
-export type ServerType = 'jupyter' | 'vscode' | 'rstudio';
+export type ServerType = 'jupyter' | 'groupOne' | 'groupTwo';
```
kimwnasptd, line 11: Could you replace the 
1. `groupOne` value with `group-one` 
2. `groupTwo` value with `group-two`

So that they align with the `serverType` the backend values will be checking?

components/crud-web-apps/jupyter/manifests/base/configs/spawner_ui_config.yaml
```diff
@@ -25,14 +25,14 @@ spawnerFormDefaults:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:master-1831e436
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-cuda-full:master-1831e436
-  imageVSCode:
-    # The container Image for the user's VS-Code Server
+  imageGroupOne:
+    # The container Image for the user's Group One Server
```
kimwnasptd, line 29: same as the comment above to add a comment for explaining the logic we apply to images in this group

components/crud-web-apps/jupyter/manifests/base/configs/spawner_ui_config.yaml
```diff
@@ -25,14 +25,14 @@ spawnerFormDefaults:
...
     value: public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/codeserver-python:master-1831e436
     # The list of available standard container Images
     options:
       - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/codeserver-python:master-1831e436
-  imageRStudio:
-    # The container Image for the user's RStudio Server
+  imageGroupTwo:
+    # The container Image for the user's Group Two Server
```
kimwnasptd, line 35: same as the comment above to add a comment for explaining the logic we apply to images in this group

kimwnasptd (commented):
@DavidSpek could you also extend the README of the web app to include a small `Image Groups` section were we will also document:
1. That the backend will be applying some extra logic to the images of `Group One` and `Group Two` [ as mentioned in the app's settings ConfigMap ]
2. Some steps/details on how someone could use their own logos, by editing the logos ConfigMap

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
DavidSpek: How about:
```yaml
# The annotation `notebooks.kubeflow.org/http-rewrite-uri: /`
# is applied to notebook in this group, configuring
# the Istio rewrite for containers that host their web UI at `/`
```

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
DavidSpek: How about:
```yaml
# The annotation `notebooks.kubeflow.org/http-rewrite-uri: /`
# is applied to notebook in this group, configuring
# the Istio rewrite for containers that host their web UI at `/`
# The annotation `notebooks.kubeflow.org/http-headers-request-set`
# is applied to notebook in this group, configuring Istio
# to add the `X-RStudio-Root-Path` header to requests
```

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
kimwnasptd: LGTM

components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
kimwnasptd: LGTM

kimwnasptd:
This is ready to roll! I also tested this live and the servers start as expected.

Thanks for your effort @DavidSpek!

PR #5824, opened 2021-04-09
feat(jwa): Remove RStudio, VSCode and Jupyter logos

### LINKED ISSUE #5821: Logo Trademarks: Remove logos from JWA

kimwnasptd:
closes #5821

This is a bug fix to unblock the release. I removed the RStudio, VSCode and Jupyter svgs from the app's code and used text/dropdowns instead. We can discuss adding them back once we have more clear guidelines on how to handle them.

I also modified slightly the app's ConfigMap and the its manifests. cc @yanniszark 

<details><summary>How it looks</summary>
<p>

Screenshot from 2021-04-09 20-58-39

https://user-images.githubusercontent.com/11134742/114226595-68e89700-997c-11eb-9cc0-d7464850314f.mp4

</p>
</details>

/cc @elikatsis @StefanoFioravanzo @yanniszark 
cc @castrojo @Bobgy @DavidSpek


DavidSpek:
Please see https://github.com/kubeflow/kubeflow/pull/5823

DavidSpek:
@kimwnasptd This PR still contains references to VSCode and RStudio, which would also cause trademark problems. My PR removes every mention of them, but does not otherwise affect the design of the current JWA.

thesuperzapper:
@kimwnasptd I dont believe this is an acceptable solution, as it still uses the "RStudio" wordmark.

I believe a solution similar to what I proposed in https://github.com/kubeflow/kubeflow/issues/5821#issuecomment-818309439 is acceptable (from a trademark perspective), and it still gives the user the option to "rename" the group (with the defaults being "Group 1", and "Group 2")

kimwnasptd:
This PR is not necessary since we moved forward with #5823 

/close

PR #5838, opened 2021-04-14
Release v1.3.0-rc.1 for kubeflow/kubeflow

yanniszark:
Follow release procedure to release images and manifests for version `v1.3.0-rc.1`

/hold

to allow for reviews


kimwnasptd:
/lgtm

I also run `git grep images -- '*kustomization.yaml'` to ensure we've touched all the kustomization files and this seems to be the case. Thanks for the PR @yanniszark!

The next step will be to make a PR to [releasing](https://github.com/kubeflow/kubeflow/tree/master/releasing) in order to recreate images for the new RC right?

yanniszark:
> The next step will be to make a PR to releasing in order to recreate images for the new RC right?

Nope, I did it as part of this PR.
Next step is to tag the merged commit as `v1.3.0-rc.1`

PR #5848, opened 2021-04-19
fix(profile-controller): KNative probes

Subreptivus:
This PR fixes the KNative's `activator` and `controller` probes on `/healthz`, `/metrics` paths on workloads for KFServing

/assign @yanniszark

### DIFF
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -365,6 +365,22 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 					},
 				},
 			},
+			{
+				To: []*istioSecurity.Rule_To{
+					{
+						Operation: &istioSecurity.Operation{
+							// Workloads pathes should be accessible for KNative's
+							// `activator` and `controller` probes
+							// See: https://knative.dev/docs/serving/istio-authorization/#allowing-access-from-system-pods-by-paths
+							Paths: []string{
+								"/healthz",
+								"/metrics",
+								"/wait-for-drain",
+							},
+						},
+					},
+				},
+			},
 		},
 	}
 }


yanniszark:
Thanks @Subreptivus!
@StefanoFioravanzo @kimwnasptd I think we should take a look after the release and check why Knative needs these probes and how it uses them. (for example, why not use the health/readiness check of the pod)

Subreptivus:
Hey @yanniszark @StefanoFioravanzo @kimwnasptd it's actually kind of hot-fix for this very release (v1.3).
And those probes are needed for KNative's activator and controller. Without them workload will never become reachable from outside (with 502 error), as both probes will be failing with 403 error.
And to have it independent from `KFServing/KNative` parts deployment I didn't want to hardcode `knative-serving` namespace or something like that. In my opinion paths like `/healthz` and `/metrics` should be harmless to expose.

DavidSpek:
@Subreptivus Just so you know, the manifests for KNative and cert-manager included with release 1.3 do not fulfill the requirements of the KFServing deployment so I think that might be a more important issue for you. Specifically, KFServing requires KNative v0.17.4+ and cert-manager v0.12.0+, while the provided manifests are for v0.14.2 and v0.11.0. If you are using the GCP blueprints or [deploy the with Argo CD](https://github.com/argoflow/argoflow) the manifests have been updated.

yanniszark:
@subreptivus thanks for opening the PR! I agree this is something we should fix going forward.
Since:
- This issue is present in previous versions of Kubeflow as well (and not a regression).
- We shouldn't be making security-critical decisions in the last minute.

I'm inclined to push this to 1.4, since the plan is still to cut 1.3.0 today.

We can start by amending KFServing's docs to make it clear that users should add the necessary annotation to their InferenceServices to disable sidecar injection, or include a policy in case they want to enable it. Then, we can discuss and agree on the best way to enable working sidecar-injected inferenceservices by default.

Subreptivus:
> @Subreptivus Just so you know, the manifests for KNative and cert-manager included with release 1.3 do not fulfill the requirements of the KFServing deployment so I think that might be a more important issue for you. Specifically, KFServing requires KNative v0.17.4+ and cert-manager v0.12.0+, while the provided manifests are for v0.14.2 and v0.11.0. If you are using the GCP blueprints or [deploy the with Argo CD](https://github.com/argoflow/argoflow) the manifests have been updated.

Thanks for comment, yet actually with probes fixed KFServing from v1.3 is working fine for us. Wwe aren't relying on manifests for Cert-manager from kubeflow/manifests, but we're using KFServing and KNative deployment manifests from mentioned source.

As for `deploy with Argo CD`, we sow that approach, yet we would like to stick to upstream manifests repo. We are using ArgoCD for deploying Kubeflow since v0.7 if not earlier, and we're "backing" it with a lot of changes which is easier to do on top of manifests repo (especially with new plain Kustomize approach) rather than argoflow repo.

DavidSpek:
@Subreptivus The Argo CD deployment is using the upstream manifests from each component directly rather than the manifests repo to make it easier upgrade the individual components according to their individual release cadence (such as pipelines).

thesuperzapper:
@yanniszark I think this PR is ready to merge

I believe this is critical to fix user-namespaced `InferenceService` in kfserving, which will never become "healthy" in `1.3.0` if you don't add the following `AuthorizationPolicy` (which is what this PR makes the controller do):
```yaml
kind: AuthorizationPolicy
metadata:
  name: allow-knative-serving-metrics
  namespace: user-XXXXX
spec:
  action: ALLOW
  rules:
  - to:
    - operation:
        paths:
        - /metrics   # The path to collect metrics by system pod.
        - /healthz 
```

yanniszark:
> I believe this is critical to fix user-namespaced InferenceService in kfserving

I agree that we should address this issue, so that examples with InferenceServices work by default.
However, let's make sure we do it right. Aka, start an issue, explain the problem, ask the KFServing people for advice.
I opened https://github.com/kubeflow/kubeflow/issues/5965 for this purpose, so let's continue the discussion there.
Once we agree, we can quickly implement it in code.

yanniszark:
@Subreptivus thanks again for your PR. After testing it, I confirm that:
- Knative's autoscaler works as expected, no more errors in logs.
- Users can get KFServing predictions through the public ingress (istio ingressgateway), by following this workaround to get an authservice session: https://github.com/kubeflow/kfserving/blob/master/docs/samples/istio-dex/README.md#authentication

I have also confirmed something that does NOT work:
- Users can NOT get KFServing predictions through the cluster-local ingress (cluster-local-gateway). There is no rule allowing traffic FROM the cluster-local-gateway TO Pods in user namespaces. This seems related to https://github.com/kubeflow/kfserving/issues/1371.

Since this PR partially fixes KFServing for Kubeflow Profiles, let's merge and continue discussing the cluster-internal case in https://github.com/kubeflow/kubeflow/issues/5965

### Next Steps

I have made a review branch with my proposed changes: https://github.com/arrikto/kubeflow/tree/equinor/master-rev
Namely:
- Update the commit message. Please see my review commit for the proposed message.
- Add a comment linking to the Knative instructions in the code.

If you agree, please push these changes and ping me again to merge. Thanks!

Subreptivus:
Hey @yanniszark!
I have unpleasant experience on trying to merge anything into either of Kubeflow components, and if you already have everything ready within your review branch, then go ahead and merge your branch. Then you can close this PR, or I can do that.
I don't care from whom those commits will be made, I only care about functionality and not carrying out our own custom images.

yanniszark:
@Subreptivus I'm sorry to hear that and I hope we can begin to rectify that experience.
The reason I use a review branch is to have everything ready in code, so it's easy to see what I describe.
To conclude the review, all you need to do is:

1. Fetch the review branch:
```sh
git remote add arrikto https://github.com/arrikto/kubeflow.git
git fetch --all
```

2. Diff the review branch against yours and see that you agree with the changes:
```sh
git diff arrikto/equinor/master-rev
git log arrikto/equinor/master-rev
```

3. If you agree, you can cherry-pick my changes ontop, as I have structured them as a separate commit:
```sh
git cherry-pick arrikto/equinor/master-rev
```

4. Squash my review commit with yours and keep the updated message:
```sh
# Mark the review commit as `s`
git rebase -i HEAD~2
```

Of course, I can do all of these, but I want to ensure we can work together on PRs now and in the future.
Let me know if there is anything else that is blocking or giving you trouble.

yanniszark:
@Subreptivus thanks a lot!
I see that you have added two new paths that we have not discussed in the issue:
- `/ready`
- `/wait-for-drain`

Could you elaborate? We found the other paths in Knative's docs: https://knative.dev/docs/serving/istio-authorization/#allowing-access-from-system-pods-by-paths

In addition, the comment about knative is in the wrong spot.
That's my mistake, could you bring it over the knative-related rule?

Subreptivus:
Right, it was resting in my local branch and I was focusing on another fix for metadata component and obviously made it slip into commit. So about those new paths, I've just made a list of paths which KNative components were trying to reach and got permission denied error while ago. I've found some explanation about `/wait-for-drain` one in issues and in the code: https://github.com/knative/serving/blob/main/pkg/reconciler/serverlessservice/resources/services.go#L148-L154, but I haven't found anything about `/ready` one.
So during the move of comment to proper place I can remove those paths as well.

yanniszark:
@Subreptivus thanks!
I was able to reproduce the `/wait-for-drain` access denied log by deleting a Knative Pod and watching the logs.
Let's include it in the list of paths and let upstream Knative know, so they can improve their docs (I can open that one).
As for `/ready`, do you have any way to reproduce? I couldn't find any reference in Knative's code.

Subreptivus:
Let's skip it for now, I don't recall what I was doing to get that error.

yanniszark:
@Subreptivus thanks a lot :D 
/lgtm
/approve

PR #5855, opened 2021-04-22
Make istio gateway configurable in tensorboard-controller

haijohn:
this PR make istio gateway in tensorboard-controller configurable, like that in notebook-controller


components/tensorboard-controller/config/manager/kustomization.yaml
```diff
@@ -1,2 +1,6 @@
 resources:
 - manager.yaml
+configMapGenerator:
+- name: config
+  literals:
+  - ISTIO_GATEWAY=kubeflow/kubeflow-gateway
```
DavidSpek, line 6: Nit: Please add a new line here

DavidSpek:
@haijohn Thanks for your contribution. When looking at the notebook-controller as a reference the configmap renerator would also need to be added to https://github.com/kubeflow/kubeflow/blob/master/components/tensorboard-controller/config/overlays/kubeflow/kustomization.yaml.

haijohn:
@DavidSpek  After put configMapGenerator in https://github.com/kubeflow/kubeflow/blob/master/components/tensorboard-controller/config/overlays/kubeflow/kustomization.yaml. I failed to run  `kustomize build config/overlay/kubeflow` , because there is a confilict with the configmap generator in base dir. To be consistnent, I moved ISTIO_GATEWAY env to https://github.com/kubeflow/kubeflow/blob/master/components/tensorboard-controller/config/base/kustomization.yaml

PR #5872, opened 2021-04-30
Added save_figures() function

dnlcesilva:
Added save_figures () function, to save the figures of a monitoring notebook.


thesuperzapper:
@dnlcesilva I am very confused, what is this related to?

PR #5877, opened 2021-05-01
Update components directory owners file

DavidSpek:
This PR adds myself to the owners file of the components directory, so I can coordinate with the authors of outstanding PRs in preparation for the upgrade of Angular (see https://github.com/kubeflow/kubeflow/pull/5805, https://github.com/kubeflow/kubeflow/issues/5525). This PR also functions as a first step towards tackling the issue of unresponsive approvers and reviewers in this repository as stated in the [Contributing to Kubeflow](https://www.kubeflow.org/docs/about/contributing/#quirks-of-the-process) section of the website.

/cc @thesuperzapper @Bobgy

Issue #5722, opened 2021-03-19
Notebook CRD Quantity Validation


yanniszark:
/kind bug

**What steps did you take and what happened:**

1. Initial manifests for the Notebook CRD didn't validate resource quantity field (CPU, Mem) to be string or int.
2. A client used a different value (JWA https://github.com/kubeflow/kubeflow/issues/5641)
3. New manifests (from kubebuilder) validate resource quantity field (CPU, Mem) to be string or int.
4. Existing clients / resources break.

Some questions about this situation:

- **Q:** Why does quantity need to be string or int?
   **A:** Quantity is string or int, because that's what Kubernetes validates as well, so it's the schema that kubebuilder generates. 

- **Q:** How does Kubernetes allow users to write fractional numbers in PodTemplates?
   **A:** Kubernetes mutates the field before persisting it! If we create a Pod/Deployment with CPU 0.1, we will see that it's persisted as 100m. We could mimic this behavior with a MutatingWebhook.

- **Q:** How did it work so far if the value is wrong?
   **A:** The deserializer in Go can actually handle fractional numbers. It's Kubernetes convention that this field must be int or string. See: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/#Quantity

**What did you expect to happen:**

To be able to use generated validation schemas from kubebuilder.

**Anything else you would like to add:**

For now, let's use the old CRD. I can think of several ways for solving this in the future:
- Support fractional values in quantity by using a MutatingWebhook, same as Kubernetes.
- Enforce validation, migrate clients and create script to migrate existing resources in a cluster.
- Relax validation to allow fractional values, since Go deserializes them correctly.

YousefRabi:
Getting this error on the kubeflow deployment bash ./pull-upstream step:

fetching package /apps/jupyter/notebook-controller/upstream from https://github.com/kubeflow/manifests to apps/jupyter/notebook-controller/upstream
failed to set "kustomize_manifests_path" automatically in package "apps/jupyter/notebook-controller/upstream" with error: /home/yousofrabi/gcp-blueprints/kubeflow/apps/jupyter/notebook-controller/upstream/crd/patches/old_crd.yaml: wrong Node Kind for  expected: MappingNode was SequenceNode: value: {- # Use the old CRD because of the quantity validation issue:

edi-bice-by:
@YousefRabi see https://github.com/kubeflow/gcp-blueprints/issues/294#issuecomment-889559605

PR #5891, opened 2021-05-07
FIX The number of gpu must be set as string in Kubernetes/Openshift

juliusvonkohout:
@DavidSpek 

The following line https://github.com/kubeflow/kubeflow/blob/c3d97bb125fa4b828bfa59cd427d5cba96952ee6/components/crud-web-apps/jupyter/backend/apps/common/form.py#L279
produces nvidia.com/gpu: 1 instead of nvidia.com/gpu: '1' which lets the pod fail with 

```
2021-05-03 11:31:24,011 | apps.default.routes.post | INFO | Creating Notebook: {'apiVersion': 'kubeflow.org/v1beta1', 'kind': 'Notebook', 'metadata': {'name': 'test1', 'namespace': 'admin', 'labels': {'app': 'test1'}, 'annotations': {'notebooks.kubeflow.org/server-type': 'jupyter'}}, 'spec': {'template': {'spec': {'serviceAccountName': 'default-editor', 'containers': [{'name': 'test1', 'image': '...', 'volumeMounts': [{'mountPath': '/dev/shm', 'name': 'dshm'}], 'env': [], 'resources': {'requests': {'cpu': '0.1', 'memory': '0.25Gi'}, 'limits': {'nvidia.com/gpu': 1}}, 'imagePullPolicy': 'IfNotPresent'}], 'volumes': [{'name': 'dshm', 'emptyDir': {'medium': 'Memory'}}], 'tolerations': []}}}}
2021-05-03 11:31:24,022 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | An error occured talking to k8s while working on http://.../api/namespaces/admin/notebooks: (422)
Reason: Unprocessable Entity
HTTP response headers: HTTPHeaderDict({'Audit-Id': '4fc9b971-7e73-48b4-9dda-27a756c7871a', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': 'e1b1c024-605e-412c-9d64-a8ef267eb67b', 'X-Kubernetes-Pf-Prioritylevel-Uid': '8a864d24-181a-44a6-9ce6-f003c50e5952', 'Date': 'Mon, 03 May 2021 11:31:24 GMT', 'Content-Length': '680'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Notebook.kubeflow.org \"test1\" is invalid: spec.template.spec.containers.resources.limits.nvidia.com/gpu: Invalid value: \"integer\": spec.template.spec.containers.resources.limits.nvidia.com/gpu in body must be of type string: \"integer\"","reason":"Invalid","details":{"name":"test1","group":"kubeflow.org","kind":"Notebook","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: \"integer\": spec.template.spec.containers.resources.limits.nvidia.com/gpu in body must be of type string: \"integer\"","field":"spec.template.spec.containers.resources.limits.nvidia.com/gpu"}]},"code":422}
``` 
on openshift 4.7 (kubernetes 1.20) with the GPU operator installed

I would prefer a backport to 1.3.

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -281,7 +281,7 @@ def set_notebook_gpus(notebook, body, defaults):
         raise BadRequest("gpus.num is not a valid number: %s" % gpus["num"])
 
     limits = container["resources"].get("limits", {})
-    limits[vendor] = str(num)
+    limits[vendor] = num
 
     container["resources"]["limits"] = limits
 
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -281,7 +281,7 @@ def set_notebook_gpus(notebook, body, defaults):
         raise BadRequest("gpus.num is not a valid number: %s" % gpus["num"])
 
     limits = container["resources"].get("limits", {})
-    limits[vendor] = num
+    limits[vendor] = str(num)
 
     container["resources"]["limits"] = limits
 
--- a/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
+++ b/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
@@ -123,6 +123,11 @@ export class FormDefaultComponent implements OnInit, OnDestroy {
     if (typeof notebook.cpu === 'number') {
       notebook.cpu = notebook.cpu.toString();
     }
+    
+    // Ensure GPU input is a string
+    if (typeof notebook.gpus.num === 'number') {
+      notebook.gpus.num = notebook.gpus.num.toString();
+    }
 
     // Remove cpuLimit from request if null
     if (notebook.cpuLimit == null) {
--- a/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
+++ b/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts
@@ -123,7 +123,7 @@ export class FormDefaultComponent implements OnInit, OnDestroy {
     if (typeof notebook.cpu === 'number') {
       notebook.cpu = notebook.cpu.toString();
     }
-    
+
     // Ensure GPU input is a string
     if (typeof notebook.gpus.num === 'number') {
       notebook.gpus.num = notebook.gpus.num.toString();


DavidSpek:
I believe this should be handled in the frontend and not the backend. Similar to
https://github.com/kubeflow/kubeflow/blob/6d6b78bcaa53480392c5acb71be578b9b9328e8d/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts#L123

Also, do you know if this breaks on OpenShift <1.20 and/or standard Kubernetes installations?

Also related: https://github.com/kubeflow/kubeflow/issues/5722

juliusvonkohout:
> I believe this should be handled in the frontend and not the backend. Similar to
> https://github.com/kubeflow/kubeflow/blob/6d6b78bcaa53480392c5acb71be578b9b9328e8d/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-default/form-default.component.ts#L123
> 
> Also, do you know if this breaks on OpenShift <1.20 and/or standard Kubernetes installations?
> 
> Also related: #5722

I currently only have a 1.20 cluster. Do you have access to other GPU clusters? I am not that familiar with typescript, but i could try to fix it there, although i guess you would be way faster than me. Is there a CI/CD test for the GPU notebooks?

DavidSpek:
I believe you would need to add this to that section in form-defaults.components.ts.

```ts
if (typeof notebook.gpus.num === 'number') {
  notebook.gpus.num = notebook.gpus.num.toString();
}
```

I could test it on a regular Kubernetes installation, but I am running 1.20 as well. Currently the CI/CD tests are still fairly limited.

juliusvonkohout:
> I believe you would need to add this to that section in form-defaults.components.ts.
> 
> ```ts
> if (typeof notebook.gpus.num === 'number') {
>   notebook.gpus.num = notebook.gpus.num.toString();
> }
> ```
> 
> I could test it on a regular Kubernetes installation, but I am running 1.20 as well. Currently the CI/CD tests are still fairly limited.

Then please test this on kubernetes 1.20. If it works properly there i can try to use google or amazon clusters with cheap gpus.

DavidSpek:
If you can make the suggested change I can build the image from your PR branch.

juliusvonkohout:
i hope, that this does not lead to errors, if the numer of gpus is not specified.

juliusvonkohout:
@DavidSpek  can you give me access to your self-built image, such that i can test on other platforms too?

juliusvonkohout:
@yanniszark this might bei interesting dir 1.3.1 https://github.com/kubeflow/manifests/issues/1890

DavidSpek:
Sorry it took so long for me to look at this, I've been very busy lately. I just tested the image built from this PR (available as `davidspek/jupyter-web-app:gpu-patch-v0.1.0`) and it works as expected. The number of GPUs in the request is a string value and there are no issues scheduling notebooks both with and without GPUs.
@yanniszark I agree that this fix should be included in the  v1.3.1 release.

juliusvonkohout:
Thank you very much.

juliusvonkohout:
@elikatsis @StefanoFioravanzo do you have any objections or can you approve this pull request?

juliusvonkohout:
@DavidSpek it seems that the reviewers are not active anymore. What can we do to advance this pull request?

DavidSpek:
@juliusvonkohout This is a known issue that has been going on for a while with the leads of the Notebooks Working Group. I've been doing my best to try and solve this, but since I'm not one of the leads there is little I can do. I've created https://github.com/kubeflow/kubeflow/pull/5877 to unblock the work in this repository, but the chances of that getting merged are very slim I fear. Maybe, @Bobgy can way in on this.

Bobgy:
/assign @kimwnasptd @thesuperzapper 
Let's try the other owners first?

kimwnasptd:
Hi @juliusvonkohout, thanks for the PR!

I'd like to first understand what the initial issue this PR tries to tackle is. The official K8s docs and the NVIDIA device plugin mention that `nvidia.com/gpu` requires a number https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/ https://github.com/NVIDIA/k8s-device-plugin#running-gpu-jobs.

I want to understand if this is a change in the 1.20 Kubernetes version or has something to do with openshift specifically. Can you verify this error gets reproduced with a smaller version of Kubernetes? 

[ It doesn't have to be a Notebook, it can be any Pod that is using an integer for `nvidia.com/gpu` limits ]

juliusvonkohout:
I tested on openshift 4.7.7 which uses Kubernetes 1.20.0

@davidspek on which platforms did you test? 

Then we know on which other platforms we should test.

DavidSpek:
I tested on Kubernetes 1.20.

juliusvonkohout:
Alright then we should try 1.19 and 1.18. I will try to get access to a kubernetes with a gpu.
I also asked redhat directly whether this is openshift specific. Maybe plain kubernetes accepts string and integers while openshift accepts only strings.

DavidSpek:
@kimwnasptd This seems to be a long running issue with spec validation. See https://github.com/instrumenta/kubeval/issues/22#issuecomment-321498408 and https://groups.google.com/g/kubernetes-sig-api-machinery/c/dLCXrZRwWpU.

TL;DR: "Quantity, when deserialized from JSON, looks at strings.  However, it's getting the raw JSON value so it might have just grandfathered in the support for raw numbers.  I think this was something that was added in Kube 1.0 and thus we're stuck with it."

"resources can take additional properties as long as they are of type quantity. Quantity is defined as":
```json
io.k8s.apimachinery.pkg.api.resource.Quantity: {
  type: "string"
},
```

DavidSpek:
I faced this exact same problem with the CPU field a while back.

kimwnasptd:
> @kimwnasptd This seems to be a long running issue with spec validation. See instrumenta/kubeval#22 (comment) and https://groups.google.com/g/kubernetes-sig-api-machinery/c/dLCXrZRwWpU.
> 
> I faced this exact same problem with the CPU field a while back.

@DavidSpek yes, this seems to be the issue.

@yanniszark looks like the underlying problem is our CRD schema, as you had described in https://github.com/kubeflow/kubeflow/issues/5722 and also reported from @DavidSpek in https://github.com/kubeflow/kubeflow/issues/5641. Using strings instead of numbers can be a short term solution, but what do you think our long term solution should be?

/lgtm

DavidSpek:
@juliusvonkohout Here is the log output:

```
W
> frontend@0.0.0 format:check /mnt/test-data-volume/kubeflow-kubeflow-presubmit-jwa-tests-5891-8774c2d-6512-d63e/src/kubeflow/kubeflow/components/crud-web-apps/jup
yter/frontend
> prettier --check 'src/**/*.{js,ts,html,scss,css}' || node scripts/check-format-error.js
Checking formatting...
[warn] src/app/pages/form/form-default/form-default.component.ts
[warn] Code style issues found in the above file(s). Forgot to run Prettier?
-----------------------------------------------
Please run `npm run format:write` to format your code.
-----------------------------------------------
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! frontend@0.0.0 format:check: `prettier --check 'src/**/*.{js,ts,html,scss,css}' || node scripts/check-format-error.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the frontend@0.0.0 format:check script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2021-06-01T09_53_46_239Z-debug.log
```

juliusvonkohout:
@kimwnasptd please /lgtm again i fixed the whitespace issue.

juliusvonkohout:
Colleagues of mine tested on 1.19.7 OTC kubernetes  (https://open-telekom-cloud.com/de/produkte-services/cloud-container-engine) and it works with strings and integers there. Using integers instead of strings also produces a warning on that platform. So using strings is the official long-term solution i would say.

juliusvonkohout:
> @juliusvonkohout: Cannot trigger testing until a trusted user reviews the PR and leaves an `/ok-to-test` message.

@DavidSpek  can you fix that?

kimwnasptd:
So using strings should be more that enough for now. Thanks for the PR and your time @juliusvonkohout!

/lgtm
/approve

Issue #5904, opened 2021-05-11
Default notebook images don't start server or fail to connect kernel


cc-gb2:
Hi, I've been trying to use jupyter notebooks on an AWS Kubeflow cluster, however the default images either report no-healthy upstream (the tensorflow-2.1 images) or the kernel can't connect (the tensorflow-1.15). As these are the default images, I'd expect them all to work.

DavidSpek:
This is usually a browser problem, can you please try restarting your browser or using a different browser?

jbottum:
/kind question
/area jupyter
priority p2

cc-gb2:
Sorry to reopen this, this doesn't work. I've tried on both Chrome and Firefox. How can I debug this?

DavidSpek:
Do you see any errors in the developer console of the browser? You’ll need to open the console and then refresh the page.

DavidSpek:
@cc-gb2 Could you try one of the new notebook images, ass the ones you are using are very outdated. The equivalent image to what your using now would be: public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-tensorflow-full:v1.3.0-rc.1

cc-gb2:
Ah yes so I've have the issue with the websocket you mention in the PR. I'd rather not screenshot it just for security, but it's along the lines of
`ws://<kubeflow-ip>...elb.amazon.com/...<notebook-url>/api/kernals/<example hash>/channels?session_id:<more_hash> faile`

DavidSpek:
@cc-gb2 Could you possibly screenshot just the browser console error so we have it on record? Or copy the full output and remove the URL. Also, could you provide details about what OS and version you are using, along with which browser and the version of the browser (and any other browser you tried). That would help a lot for us to be able to pinpoint what could be causing the problem. Given the error is happening in the browser I don't think it is a Kubeflow or Istio problem, but it still is something we should investigate and try and figure out.

cc-gb2:
This is a screenshot with chrome Ubuntu 18.04 Chromium Version 90.0.4430.212 (Official Build) snap (64-bit)
![image](https://user-images.githubusercontent.com/83974855/119711233-3117af00-be57-11eb-85ad-e930d9626614.png)
=

From firefox 88.0.1: 
![image](https://user-images.githubusercontent.com/83974855/119711951-ffebae80-be57-11eb-837b-df32521a2d4d.png)

vvnpan:
[...]

PR #5920, opened 2021-05-15
Notebook servers: Add explanation about kernel not connecting

DavidSpek:
Closes: https://github.com/kubeflow/kubeflow/issues/5904

There is a common but sporadic problem where the kernel in JupyterLab is stuck in a `connecting` state (see https://github.com/kubeflow/kubeflow/issues/5904). This problem is not related to Kubeflow, but rather originates in the user's browser. It can be identified by looking in the browser logs in the developer tools where there will be errors related to problems connecting to the websocket. The fix for this is restarting the browser (refreshing usually doesn't work) or (temporarily) using a different browser if the current one cannot be restarted. This PR adds a common problems and solutions sections to the notebook image README that explains this issue and the solution.

/cc @thesuperzapper

### DIFF
--- a/components/example-notebook-servers/README.md
+++ b/components/example-notebook-servers/README.md
@@ -1,9 +1,9 @@
+# Example Notebook Servers
+
 > 🛑️️ These server images are provided as __examples only__ and are supported on a best-effort basis.
 > Contributions are greatly appreciated.
 
-# Example Notebook Servers
-
-## Overview:
+## Overview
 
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@DavidSpek is there a string of text or error which is thrown when this error occurs?

If so, we should put it in the doc, so google will help relate these docs to the solution.

DavidSpek:
You only see the error in the developer console in the browser, and as I haven't had it happen recently I haven't been able to copy the actual error message. But the error states there is a problem connecting with the websocket.

PR #5925, opened 2021-05-19
tensorboard-controller: fix binding issue

DavidSpek:
Part of https://github.com/kubeflow/kubeflow/issues/5938

This PR adds the `--bind_all` argument for tensorboard, solving the problem described in the issue mentioned above, as well as fixing support for Istio 1.10.

### DIFF
--- a/components/tensorboard-controller/config/crd/bases/tensorboard.kubeflow.org_tensorboards.yaml
+++ b/components/tensorboard-controller/config/crd/bases/tensorboard.kubeflow.org_tensorboards.yaml
@@ -3,6 +3,8 @@
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.5
   creationTimestamp: null
   name: tensorboards.tensorboard.kubeflow.org
 spec:
@@ -12,7 +14,7 @@ spec:
     listKind: TensorboardList
     plural: tensorboards
     singular: tensorboard
-  scope: ""
+  scope: Namespaced
   subresources:
     status: {}
   validation:
--- a/components/tensorboard-controller/controllers/tensorboard_controller.go
+++ b/components/tensorboard-controller/controllers/tensorboard_controller.go
@@ -255,6 +255,7 @@ func generateDeployment(tb *tensorboardv1alpha1.Tensorboard, log logr.Logger, r 
 							WorkingDir:      "/",
 							Args: []string{
 								"--logdir=" + mountpath,
+								"--bind_all",
 							},
 							Ports: []corev1.ContainerPort{
 								{


juliusvonkohout:
@DavidSpek If you upgrade istio please consider switching from the ratelimited dockerhub to the official istio gcr repository.

```
istioctl manifest generate --set hub='gcr.io/istio-release'
# --set global.proxy.resources.requests=50m might also be interesting
```

I am using this with more parameters (e.g. --set tag='1.9.2') for Minikube and Openshift.

DavidSpek:
For [ArgoFlow](https://github.com/argoflow/argoflow) we're using the Istio Operator and the following `istio-spec.yaml` to actually install Istio. Both the operator and the istio-spec have a `hub` and a `tag` value you can use to easily specify whatever you prefer. This makes updates a lot easier to manage (especially given that Renovate will automatically create PRs to update these values). 

```yaml
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  namespace: istio-system
  name: istio
spec:
  profile: default
  tag: 1.9.5 # istio/operator
  hub: docker.io/istio
  meshConfig:
    accessLogFile: /dev/stdout
    enablePrometheusMerge: true
```

Istio operator Argo CD application:

```yaml
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: istio-operator
  namespace: argocd
spec:
  project: default
  source:
    repoURL: 'https://github.com/istio/istio'
    targetRevision: HEAD
    path: manifests/charts/istio-operator
    helm:
      parameters:
      - name: hub
        value: docker.io/istio
      - name: tag
        value: 1.9.5 # istio/operator
  destination:
    server: 'https://kubernetes.default.svc'
    namespace: istio-operator
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - CreateNamespace=true
```

juliusvonkohout:
So are you going to to the istio update?  Because here they say that it is supposed to be done with istioctl https://github.com/kubeflow/manifests/blob/master/common/istio-1-9-0/README.md

DavidSpek:
I didn't write that file :). You can do it that way if you want to deal with managing all that yaml manually, but I find the operator more elegant. However, this is not really on topic for this PR. I've suggested that the manifests get updated in the next patch release. How that is done doesn't really matter to me. I created my own distribution because I didn't feel like dealing with those manifests as they have way to much maintenance overhead, which is why they aren't kept up to date in a timely manner.

DavidSpek:
@kimwnasptd Could you please take a look at this.

kimwnasptd:
Thank you for the PR @DavidSpek!

Could you change the referenced issue to mention https://github.com/kubeflow/kubeflow/issues/5938 instead? Also lets not close the tracking issue from this PR, so add something like: first step for https://github.com/kubeflow/kubeflow/issues/5938.

Rest LGTM

DavidSpek:
@kimwnasptd Cool, I just updated the issue reference in the post.

kimwnasptd:
Thanks!

/lgtm
/approve

PR #5933, opened 2021-05-24
Cut v1.3.0

yanniszark:
Release stable v1.3.0, same as v1.3.0-rc.1.
Even though there are no changes, we should release a stable version for better UX.


kimwnasptd:
Thanks @yanniszark!

/lgtm
/approve

PR #5939, opened 2021-05-25
Add support for ServiceAccountName and AutomountServiceAccountName to admission-webhook

hopper-signifyd:
The admission-webhook CRD appears to define support for the `serviceAccountName`, however, the code didn't actually read it. I've added support for both `serviceAccountName` and `automountServiceAccountToken` fields since my organization needs those.

Also, this is the first contribution from Signifyd, Inc. We should have a CLA on file as of about a week ago.


hopper-signifyd:
@discordianfish Are there any updates on this?

hopper-signifyd:
Thank you!

PR #5941, opened 2021-05-27
upgrade kfam profile api to fix bug to resource quota

hwk42:
kfam use v1beta1 profile api that the ResourceQuotaSpec filed named "resourcequotaspec",but in profile controller it named "resourceQuotaSpec",which is not recgonized


hwk42:
@googlebot I signed it!

hwk42:
@googlebot I signed it!

yanniszark:
/ok-to-test

@hwk42 the changes in this PR seem ok (updating the profile version), but I don't understand the issue at hand that you bumped into. Could you please write a numbered list of steps that you took which led to an error as well as the error you saw?

For example:
1. Create a Profile CR. The YAML used is:
2. ...

hwk42:
> /ok-to-test
> 
> @hwk42 the changes in this PR seem ok (updating the profile version), but I don't understand the issue at hand that you bumped into. Could you please write a numbered list of steps that you took which led to an error as well as the error you saw?
> 
> For example:
> 
> 1. Create a Profile CR. The YAML used is:
> 2. ...

I just downloaded the code of v1.3 and open with goland,no more action had been done.

DavidSpek:
lol

Issue #5869, opened 2021-04-28
Jupyter web app logos override all of the app's assets


kimwnasptd:
/kind bug

**What steps did you take and what happened:**
Deploy the app from the latest manifests, where we mount the logos ConfigMap under the entire `assets` folder of the app
https://github.com/kubeflow/kubeflow/blob/f757f4ee1ed1c55d4d8930a13c574bcb5872409c/components/crud-web-apps/jupyter/manifests/base/deployment.yaml#L20

If I try that with a version of the app that includes more files under assets, like fonts #5691, then the files are replaced from the contents of the ConfigMap

**What did you expect to happen:**
I would like the logos to not override the contents of all the assets

**Anything else you would like to add:**
This can be done by
1. mounting the logos ConfigMap into a different subdirectory, like `assets/logos` to not override the entire folder in the manifests
2. using the updated path when fetching the logos [[code](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/frontend/src/environments/environment.ts#L10-L15)]

cc @DavidSpek @saffaalvi

**Environment:**
- Kubeflow version: 1.3

/priority p1

DavidSpek:
The time it took to write this issue probably took twice as long as it would have taken to fix the issue. I can create a PR to fix this if it will actually get merged in a timely manner.

PR #5691, opened 2021-03-12
feat(jupyter): add fonts as assets to service

saffaalvi:
Closes https://github.com/StatCan/kubeflow/issues/50

Fonts for the `crud-web-app/jupyter-web-app` component are loaded from `https://fonts.googleapis.com/`. We have experienced some issues with page load being blocked, waiting on `fonts.gstatic.com`.

We decided to move all the fonts into the service so issues with external services don't affect ours.

This PR creates a new `fonts` folder in `components/crud-web-apps/jupyter/frontend/src/assets/` containing all the `.woff2` files for the fonts used. Those files are then imported in `components/crud-web-apps/jupyter/frontend/src/styles.scss` using `fonts.css`.

PR #5942, opened 2021-05-27
JWA: Don't override assets with logos ConfigMap

kimwnasptd:
closes https://github.com/kubeflow/kubeflow/issues/5869

This PR:
* Moves the logos to a distinct directory inside `assets/logos` of the app
* Make the ConfigMap to mount in `assets/logos` instead of `assets`
* Copies the logos to the source of the app as well, so we don't see errors in dev mode from missing svgs

This way the logos ConfigMap will not be overriding the entire `assets` folder of the app. This is also needed for https://github.com/kubeflow/kubeflow/pull/5691.

cc @wg102
/cc @yanniszark @elikatsis 
/assign @kimwnasptd


kimwnasptd:
It seems that having prow add the `approved` label has bitten us here as well, since the assigned reviewers didn't have a chance to take a look at this.

I think it's time to make the same move pipelines did and explicitly require self-approval https://github.com/kubeflow/kubeflow/pull/5706#issuecomment-804987516. 
Opening an issue to discuss.

PR #5959, opened 2021-06-02
Remove deprecated JWA

kimwnasptd:
Closes https://github.com/kubeflow/kubeflow/issues/5553

This PR completely removes the old deprecated Jupyter web app. We are now using the one under `/components/crud-web-apps/jupyter` which is utilizing our common code.

/assign @yanniszark @elikatsis 
/approve cancel


DavidSpek:
This is also related to https://github.com/kubeflow/kubeflow/pull/5280, so we should probably come up with a list what from that PR needs to be implemented in the crud JWA. I believe there were fixes for how the PVC naming was being done, as well as listing PVCs when you select an existing one in the spawner (which might not stay relevant for long), but that is just off the top of my head. @thesuperzapper probably has a better idea what other improvements were implemented in that pr.

DavidSpek:
I spoke to @thesuperzapper about this PR and it shouldn't be a problem to proceed w.r.t. his open PR.
/lgtm

thesuperzapper:
There is no reason to keep the old code in the repo, however, we should still review changes not yet implemented from PR #5280.
/approve

PR #5961, opened 2021-06-03
Update jupyter, rstudio and vs code notebook images + allow rstudio in iframe

DavidSpek:
Update the next notebook images with the new base image and update the installed packages.


thesuperzapper:
Looks good!

/lgtm

PR #5964, opened 2021-06-03
Make menu bar scrollable again

### LINKED ISSUE #5963: New menu bar cannot be scrolled
/kind feature

New menu bar of KF1.3 has more menu items than KF1.2. The users whose display is small (especially laptop users) can't show all menu items.
cutoff

The scroll bar should be available on menu bar when the window size is small.

Previously, side menu bar was scrollable in KF1.2. I suppose updating to new menu bar for KF1.3 forgot to support this feature.
kf1 2

toshi-k:
Fix
#5963 

This PR makes it possible to scroll the menu bar.
It was available in KF1.2 but it is **not** available in KF1.3. That's why the title of this PR is `Make menu bar scrollable again`.

As an attached image, the laptop users will be able to scroll the side bar.
scrollable

The style of scroll bar is suitable to kubeflow UI enough.

ref) This tip is written in official document of PolymerElements.<br>https://github.com/PolymerElements/app-layout/tree/master/app-drawer
```
To make the contents of the drawer scrollable, create a wrapper for the scroll content, and apply height and overflow styles to it.
  <app-drawer>
    <div style="height: 100%; overflow: auto;"></div>
  </app-drawer>
```

### DIFF
--- a/components/centraldashboard/public/components/main-page.css
+++ b/components/centraldashboard/public/components/main-page.css
@@ -22,6 +22,11 @@
     flex: 1
 }
 
+.scrollable {
+    height: 100%;
+    overflow: auto;
+}
+
 #MainDrawer {
     color: white;
     --app-drawer-content-container: {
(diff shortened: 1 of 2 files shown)


DavidSpek:
Thanks for the contributions!

thesuperzapper:
Thanks @toshi-k, looks good!
/approve

PR #5966, opened 2021-06-04
use DeepEqual to comapre ContainerState

zhyon404:
`ContainerState` in `pod.Status.ContainerStatuses[i].State == instance.Status.ContainerState` contains pointer, == will always be false. Change to deepcopy to compare value.

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"reflect"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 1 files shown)


DavidSpek:
@zhyon404 Thank you for the contribution. Could you maybe describe what issue you were experiencing that required this fix?

zhyon404:
I noticed the notebook has a log of updating statuses log in a stable state, and the update trigers another update and than stoped. I think the expected behavior should be not to update status?

DavidSpek:
I believe https://github.com/kubeflow/kubeflow/issues/5282 is related to this as well.

zhyon404:
> I believe #5282 is related to this as well.

I think this change slightly improve the previous pr? Could you take a look at this? @gilbeckers

PR #5980, opened 2021-06-15
fix(notebooks) make culling work with multi-user (#5128)

### LINKED ISSUE #5128: Notebook culling not working with Istio and RBAC
/kind bug

We deployed Kubeflow with Istio and Istio RBAC (w/ OIDC) enabled and we were looking to use Notebook culling to cull idle notebooks after a period of time. I should also note that we have mTLS enabled on our Istio config.

Notebook culling is not functional due to multiple issues:

* Notebook controller is not on the service mesh, so fails the mTLS check when talking to notebooks

```
2020-07-13T21:11:51.792Z	INFO	culler	Error talking to http://construction-starts.ns.svc.cluster.local/notebook/dsd/construction-starts/api/status	{"error": "Get http://construction-starts.dsd.svc.cluster.local/notebook/dsd/construction-starts/api/status: read tcp 172.21.XX.YY:36004->10.0.XX.YY:80: read: connection reset by peer"}
```

* If we enable injection on the notebook controller, it is not in the ServiceRoleBinding list, so it fails the Istio RBAC check and gets permission denied

We started going down the road of enable istio-injection on the Kubeflow control plane, but this is introduced a bunch of problems. These were the initial steps we tried:

1. Add istio-injection=enabled label to kubeflow namespace
2. Rollout restart all deployments / statefulsets
3. Adding annotation to disable injection on the admission webhook and metadata-envoy-deployment
4. Delete the DestinationRule for the namespace 
5. On the svc for ml-pipeline, change http service name to tcp-http

Things generally worked (after disabling injection on a couple of services), but there were random errors using Kubeflow and we didn't know of what other impacts to functionality there were. We also didn't like diverging that much from the direction of Kubeflow (https://github.com/kubeflow/manifests/pull/712).

Notebook culling should function correctly with Istio and Istio RBAC enabled.

This was also mentioned in https://github.com/kubeflow/kubeflow/pull/3856#issuecomment-612295807.

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 1.0.2
- kfctl version: (use `kfctl version`): 1.0.2
- Kubernetes platform: (e.g. `minikube`) AKS
- Kubernetes version: (use `kubectl version`): 1.15.7
- OS (e.g. from `/etc/os-release`): Ubuntu 16.04
- Istio: 1.4.9

LEDfan:
Hi 

This fixed #5128 according to the guidelines of this comment: https://github.com/kubeflow/kubeflow/issues/5128#issuecomment-859814573 . I wasn't sure where to place the `AuthorizationPolicy`, I think it belongs in the `overlays/kubeflow` directory, as istio is only used in combination with kubeflow. In addition it needs to be created in the `istio-system` namespace and not in the `kubeflow` namespace. Therefore I separated the manifests in `overlays/kubeflow` into a `kubebflow/` and `istio-sytem/` directory, representing the namespaces of the manifests in both directories. Of course, please let me know if you have any better location/setup for this file.

Looking forward to your feedback!

### DIFF
--- a/components/notebook-controller/config/manager/manager.yaml
+++ b/components/notebook-controller/config/manager/manager.yaml
@@ -15,8 +15,6 @@ spec:
       labels:
         app: notebook-controller
         kustomize.component: notebook-controller
-      annotations:
-        sidecar.istio.io/inject: "false"
     spec:
       containers:
       - name: manager
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -381,6 +381,24 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 					},
 				},
 			},
+			{
+				// allow the notebook-controller in the kubeflow namespace to access the api/status endpoint of the notebook servers.
+				From: []*istioSecurity.Rule_From{
+					{
+						Source: &istioSecurity.Source{
+							Principals: []string{"cluster.local/ns/kubeflow/sa/notebook-controller-service-account"},
+						},
+					},
+				},
+				To: []*istioSecurity.Rule_To{
+					{
+						Operation: &istioSecurity.Operation{
+							Methods: []string{"GET"},
+							Paths:   []string{"*/api/status"}, // wildcard for the name of the notebook server
+						},
+					},
+				},
+			},
 		},
 	}
 }


DavidSpek:
Hi @LEDfan, thank you for the contribution. Could you please explain why the AuthorizationPolicy needs to be in the `istio-system` namespace?

LEDfan:
Hi. In a multi-user setup, the notebook servers are running in the namespace of each user. Therefore the policy should apply to each of those namespaces.  You can tell istio to apply a policy to every namespace by defining that policy in the `istio-system` namespace (the root namespace of istio), see https://istio.io/latest/blog/2019/v1beta1-authorization-policy/#root-namespace . Hence we must create it in the istio-system namespace.

I think we could also modify the `profiles-controller` to create a policy in the namespace of each user (see: https://github.com/kubeflow/kubeflow/issues/5128#issuecomment-856580616).

DavidSpek:
I was indeed going to suggest that it would be better to have the profile controller take care of creating these policies in each (user) namespace. Kubeflow might not be the only application being run on the cluster and the broad scope of the authorization policy is probably not good in terms of security.

Also, you could simply remove the line below as automatic injection is enabled for the kubeflow namespace:
https://github.com/kubeflow/kubeflow/blob/36c67fc453f6f344a43596eb02f6dcfb6cba790a/components/notebook-controller/config/manager/manager.yaml#L19

LEDfan:
I agree on the security aspects. I tried modifying the profiles-controller in this commit: https://github.com/LEDfan/kubeflow/commit/866c20078af1850c2251ec64eb51a34e142875af . Can you have a quick check if this is what you had in mind? I'm not very experienced with Go, so I copied a very similar function (but for clearness I didn't created the new policy in the existing function). If it looks ok, I'll force push it to this branch and you can do a proper review. Thanks :)

DavidSpek:
I would add it to the existing AuthorizationPolicy for the namespace:
https://github.com/LEDfan/kubeflow/blob/866c20078af1850c2251ec64eb51a34e142875af/components/profile-controller/controllers/profile_controller.go#L350

LEDfan:
Sure, will change it. Should the name of that policy change then (to reflect the more generic usage) or do you prefer to keep it as is?

LEDfan:
@DavidSpek Done! Changes are quite minimal now :) I see the CLA has been processed now too. Thanks for you guidance.

DavidSpek:
Awesome, thanks for the making the changes. Have you already tested this yourself? I'll try and get to testing this myself soon as well.

LEDfan:
Yes, I tested it on my own development cluster.

iptizer:
Hi,

@LEDfan Thanks for this PR :), this is exactly what I was looking for!

is there a plan to get this merged?

@yanniszark 
@thesuperzapper 
@MartinForReal

MartinForReal:
> I was indeed going to suggest that it would be better to have the profile controller take care of creating these policies in each (user) namespace. Kubeflow might not be the only application being run on the cluster and the broad scope of the authorization policy is probably not good in terms of security.
> 
> Also, you could simply remove the line below as automatic injection is enabled for the kubeflow namespace:
> https://github.com/kubeflow/kubeflow/blob/36c67fc453f6f344a43596eb02f6dcfb6cba790a/components/notebook-controller/config/manager/manager.yaml#L19

I'm a little bit confused about why notebook controller needs a istio sidecar. Could you please clarify it?

LEDfan:
Hi @MartinForReal 

The purpose of the PR is that the notebook-controller is able to talk with the notebook servers. Therefore we changed the istio policies to allow traffic to the notebook servers from the notebook controller. However, this will only work if the traffic going from the controller to the notebook-servers goes though the istio proxy. The notebook servers should be able to identify that the traffic is coming from the controller and this only works using the proxies.

kimwnasptd:
Doing a pass of the long pending PRs. I'm good to go with the following changes, to allow have the Profiles Controller creating the corresponding AuthorizationPolicy. This also aligns with our efforts to improve the exposure of idleness and the culler #6295, and also with the effort to secure the workloads with Istio sidecars.

Thanks for the work @LEDfan, and apologies for the delayed review/merge

/lgtm
/approve

PR #5986, opened 2021-06-16
Add support for PodSecurityContext to PodDefault

akirillov:
### What changes were proposed in this pull request?
This PR adds `PodSecurityContext` to `PodDefault` to allow users to specify custom user id, group id, and similar properties.

### Why are the changes needed?
* To provide users with a UI option to select a pre-defined Security Context when creating notebooks
* `PodDefault` seems to be the right abstraction for introducing this change as opposed to the Notebook Spawner UI. A `PodDefault` with a specific user id and group id creation can be restricted to the cluster administrators only so that users can not provide any random UIDs themselves in the Notebook UI.
* This change unlocks the possibility to select and mount NFS volumes where access to the data is controlled by the UID. An example would look like this:
  ```
  kind: PodDefault
  metadata:
    name: nfs-datasets
    namespace: demo
  spec:
    selector:
      matchLabels:
        nfs-datasets: "true"
    desc: "Add NFS Datasets"
    securityContext:
      runAsUser: 1234
    volumeMounts:
      - name: nfs-volume
        mountPath: /kubeflow/datasets
    volumes:
      - name: nfs-volume
         nfs:
           path: /data/datasets
           server: nfs.company.com
  ```

### How were the changes tested?
* Dedicated unit tests
* Manual testing on a cluster
  
  SecurityContext support in PodDefaults

### DIFF
--- a/components/admission-webhook/README.md
+++ b/components/admission-webhook/README.md
@@ -13,7 +13,7 @@ Here is the workflow on how this is used in Kubeflow:
 1. Users create  PodDefault manifests which describe additional runtime requirements (i.e., volume, volumeMounts, environment variables) to be injected  into a Pod at creation time.
 PodDefaults use label selectors to specify the Pods to which a given PodDefault applies.
 PodDefaults are namespace scope, i.e., they can be applied/viewed in the namespace.  
-As an example, the following manifest declares a PodPrest in `kubeflow` namespace to add the secret ```gcp-secret``` in to pods in the given namespace. 
+As an example, the following manifest declares a PodDefault in `kubeflow` namespace to add the secret ```gcp-secret``` in to pods in the given namespace.
 
 ```
 apiVersion: "kubeflow.org/v1alpha1"
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -338,6 +338,32 @@ func mergeTolerations(tolerations []corev1.Toleration, podDefaults []*settingsap
 	return mergedTolerations, err
 }
 
+// getSecurityContext scans through the matching PodDefaults and returns a PodSecurityContext
+// if it is provided. Returns an error if there are multiple conflicting PodSecurityContexts specified.
+func getSecurityContext(podDefaults []*settingsapi.PodDefault) (*corev1.PodSecurityContext, error) {
+	var securityContext corev1.PodSecurityContext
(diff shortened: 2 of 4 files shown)


akirillov:
soft ping @DavidSpek, would appreciate some feedback when you have time.

DavidSpek:
@akirillov Thanks for the contribution. I'll try and look at this tomorrow. I'm not sure if PodDefault is really the place for this functionality though, as you should just be able to add it to the notebook specification. There are plans being worked on to rewrite the Notebook Spawner UI that would also include PodSecurityContext settings.

akirillov:
thanks, @DavidSpek. While exposing the Security Context in the Spawner UI allows an easier configuration, I'd like to provide a little bit different perspective on these properties.

The `PodDefault` based implementation is a little bit leaning towards the NFS-specific use-case but there are a few reasons for that:
* Directory access permissions in NFS shares are based on UIDs/GIDs and exposing Security Context in the Spawner UI would allow notebook users to specify _any_ of these properties thus introducing weaker security when it comes to the data access permissions because users can set UID/GID not belonging to them, and access the data they're not supposed to access. Also, this will make it possible for the users to set UID/GID to `root` and escalate the container privileges. In the enterprise setting, a cluster administrator manages access credentials by creating required `Secret`s and `ConfigMap`s and providing them to the notebook users via `PodDefault`s so that users just mount the required configurations to their notebooks in a self-service manner without the possibility to hijack them. All these resources can be a subject for GitOps-based rotation and changes and are not supposed to be managed by users at all. UIDs and GIDs when used for directory permissions management are the same security primitive as credentials and shouldn't be allowed for easy modification by the users.
* Currently, there's no way in the UI that allows users to mount NFS volumes to their notebooks and set their UID/GID permissions at the same time, this can only be done via manifest. The volume manager doesn't help here because it requires additional software installation which is integrated with the NFS and provides a StorageClass for volume provisioning. The `PodDefault` based implementation combines the possibility of mounting native NFS volumes with the UID/GID specification in a single resource (created by a cluster administrator) so that the notebook users can "consume" it via the Spawner UI in a single click.

Please let me know what do you think about this. Thanks.

discordianfish:
/ok-to-test

I agree that this is needed in the described use case as of today, so I'd be ok with merging this.

PR #5991, opened 2021-06-17
Extend deadline for PyTorch image building

DavidSpek:
Pushing relatively large images to ECR is very slow, which is why the TensorFlow notebook server images with CUDA support required the default timeout of the Argo Workflow task to be increased. At first this was not necessary for the PyTorch notebook server images. However, with the latest version the PyTorch job is cancelled before the image push is completed ([see here](https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-postsubmit-nb-j-pt-e9324d3-7920-7ad4?tab=workflow&nodeId=kubeflow-kubeflow-postsubmit-nb-j-pt-e9324d3-7920-7ad4-3401172493)).

This PR extends the timeout period for the PyTorch build jobs so they are the same as the TensorFlow notebook images.


thesuperzapper:
Thanks @DavidSpek!
/lgtm
/approve

PR #5871, opened 2021-04-30
Make it possible to add Namespaced menu items

toshi-k:
Fix
#5863

This PR make it possible to add namespaced items on side bar menu.
The users can define the side bar menu as below.

```
        {
          "type": "item",
          "link": "/pipeline/#/artifacts",
          "text": "Artifacts",
          "icon": "editor:bubble-chart"
        },
        {
          "type": "item",
          "link": "/pipeline/#/executions",
          "text": "Executions",
          "icon": "av:play-arrow"
        },
        {
          "type": "item",
          "link": "/myapp1/{ns}/",
          "text": "MyApp1",
          "icon": "icon"
        },
        {
          "type": "item",
          "link": "/myapp2/{ns}/",
          "text": "MyApp2",
          "icon": "icon"
        },
```

{ns}s are replaced by current namespace. The full URL should be https://gateway/myapp1/workspace1/ (when ns=workspace1) or https://gateway/myapp1/workspace2/ (when ns=workspace2).

This feature help cluster admin to add own apps for each namespace.

Issue #5863, opened 2021-04-25
Sidebar links with namespace parameters


toshi-k:
/kind feature

**Why you need this feature:**
I would like to add own apps on sidebar menu.
Though we can add any items by editting below config map, it is impossible to change links based on namespaces.
https://github.com/kubeflow/kubeflow/blob/v1.3-branch/components/centraldashboard/manifests/base/configmap.yaml

**Describe the solution you'd like:**
Defining the menu as below, the full URL should be https://gateway/myapp1/workspace1/ (when namespace=workspace1) or https://gateway/myapp1/workspace2/ (when namespace=workspace2).
```
        {
          "type": "item",
          "link": "/pipeline/#/artifacts",
          "text": "Artifacts",
          "icon": "editor:bubble-chart"
        },
        {
          "type": "item",
          "link": "/pipeline/#/executions",
          "text": "Executions",
          "icon": "av:play-arrow"
        },
        {
          "type": "item",
          "link": "/myapp1/{namespace}/",
          "text": "MyApp1",
          "icon": "icon"
        },
        {
          "type": "item",
          "link": "/myapp2/{namespace}/",
          "text": "MyApp2",
          "icon": "icon"
        },
```

**Anything else you would like to add:**
Reload the page automatically when the namespace is changed from the top toolbar.

DavidSpek:
@karlschriek I believe this is similar to what you mentioned regarding namespaces MLFlow.

toshi-k:
@DavidSpek That's right. Some applications like ML Flow are not ready for namespaces. I think the easiest way would be deploying such app in each namespace and control traffic by URL.

kimwnasptd:
@toshi-k I'd like to ask some more questions to further understand what's required for MLFlow to work as part of Kubeflow, as well as to see if the current mechanism of the CentralDashboard could satisfy this.

> Some applications like ML Flow are not ready for namespaces. I think the easiest way would be deploying such app in each namespace and control traffic by URL.

So IIUC, the MLFlow components currently don't have a mechanism for K8s Namespace isolation, so the solution you are proposing is to launch the MLFlow components/apps' yamls [Pods, Services, PVCs etc] and a VirtualService to each namespace in order to:
1. expose each app instance, via the VirtualService, for each Namespace
2. ensure the context stored in the MLFlow app's Pod [backed by PVCs?] on each Namespace will only hold the context for that namespace.
 
@toshi-k does my explanation accurately describe what you had in mind?

[...]

PR #5995, opened 2021-06-20
Interact with the namespace selector

### LINKED ISSUE #5994: Namespaced menu item can't interact with the namespace selector
/kind feature

A new feature to add namespaced menu items was merged by #5871, it can't interact with the namespace selector.

**[Example]**
Assuming MLFlow is added as `/mlflow/{ns}` to the menubar, and namespace=workspace1 is activated, the full URL would be ` https://gateway/_/mlflow/workspace1/?ns=workspace1`.
Accessing that page and changing the namespace to workspace2 by the namespace selector, the URL become inconsistent as `https://gateway/_/mlflow/workspace1/?ns=workspace2` and no menu item is selected.

When the user control the namespace selector, namespaced menu items should follow it.

In a below example, ` https://gateway/_/mlflow/workspace1/?ns=workspace1` should move to `https://gateway/_/mlflow/workspace2/?ns=workspace2`, if the namespaced is changed from workspace1 to workspace2.

* This feature is explained in the below page.<br>https://www.kubeflow.org/docs/components/central-dash/customizing-menu/
* I'm going to send a PR to solve this problem.

/cc @kimwnasptd

toshi-k:
Fix
#5994 

This PR enables namespaced menu items to interact with the namespace selector.
It means the URL of namespaced menu items will follow the selection of namespace selector.

The feature of namespaced menu items is merged by #5871, and described in [here](https://www.kubeflow.org/docs/components/central-dash/customizing-menu/) .

When the user chooses namespace=workspace1, the URL well be `https://gateway/_/mlflow/workspace1/?ns=workspace1`. If he chooses namespace=workspace2 later, the URL will change into `https://gateway/_/mlflow/workspace2/?ns=workspace2` dynamically.

/cc @kimwnasptd @DavidSpek

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -291,11 +291,9 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         // update namespaced menu item when namespace is changed
         // by namespace selector
         if (this.namespacedItemTemplete &&
-            this.namespacedItemTemplete.includes('{ns}') &&
-            this.queryParams && this.queryParams["ns"]) {
+            this.namespacedItemTemplete.includes('{ns}')) {
             this.set('subRouteData.path',
-                this.namespacedItemTemplete.replace(
-                '{ns}', this.queryParams["ns"]))
+                this.namespacedItemTemplete.replace('{ns}', namespace))
         }
     }
 
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -106,6 +106,7 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
                 // eslint-disable-next-line max-len
                 computed: 'computeShouldFetchEnv(registrationFlow, workgroupStatusHasLoaded)',
             },
+            matchingIndex: Number,
             namespacedItemTemplete: String,
         };
     }
@@ -371,26 +372,25 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
             }
             return m.replace('{ns}', queryParams["ns"]);
         });
-        let matchingIndex = 0
         if (hashPath) {
             matchPath = path + '#' + hashPath;
-            matchingIndex = allLinks
+            this.matchingIndex = allLinks
                 .findIndex((l) => this.compareLinks(l, matchPath));
         } else {
             allLinks.forEach((link, index) => {
                 if (path.startsWith(link)) {
-                    matchingIndex = index
+                    this.matchingIndex = index
                 }
             });
         }
-        matchingLink = allLinks[matchingIndex]
+        matchingLink = allLinks[this.matchingIndex]
 
         // find the HTML element that references the active link
         const activeMenuEl = Array.from(htmlElements).find(
             (x) => this.compareLinks(x.parentElement.href, matchingLink));
         if (activeMenuEl) {
             // in case the item is a section item, open its section
-            this.namespacedItemTemplete = allLinksTemplete[matchingIndex]
+            this.namespacedItemTemplete = allLinksTemplete[this.matchingIndex]
             activeMenuEl.parentElement.parentElement.opened = true;
             activeMenuEl.classList.add('iron-selected');
         }
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -106,6 +106,7 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
                 // eslint-disable-next-line max-len
                 computed: 'computeShouldFetchEnv(registrationFlow, workgroupStatusHasLoaded)',
             },
+            namespacedItemTemplete: String,
         };
     }
 
@@ -117,6 +118,7 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         return [
             // eslint-disable-next-line
             '_routePageChanged(routeData.page,subRouteData.path,routeHash.path)',
+            '_namespaceChanged(queryParams.ns)',
         ];
     }
 
@@ -285,6 +287,18 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         }
     }
 
+    _namespaceChanged(namespace) {
+        // update namespaced menu item when namespace is changed
+        // by namespace selector
+        if (this.namespacedItemTemplete &&
+            this.namespacedItemTemplete.includes('{ns}') &&
+            this.queryParams && this.queryParams["ns"]) {
+            this.set('subRouteData.path',
+                this.namespacedItemTemplete.replace(
+                '{ns}', this.queryParams["ns"]))
+        }
+    }
+
     _buildHref(href, queryParamsChange) {
         // The "queryParams" value from "queryParamsChange" is not updated as
         // expected in the "iframe-link", but it works in anchor element.
@@ -348,9 +362,10 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         const htmlElements = this._clearActiveLink();
         let matchPath = path;
         let matchingLink = '';
-        const allLinks = this.menuLinks.map((m) => {
+        const allLinksTemplete = this.menuLinks.map((m) => {
             return m.type === 'section' ? m.items.map((x) => x.link) : m.link;
-        }).flat().sort().map((m) => {
+        }).flat().sort()
+        const allLinks = allLinksTemplete.map((m) => {
             // replace namespaced menu items
             const queryParams = this.queryParams
             if (!queryParams || !queryParams["ns"]) {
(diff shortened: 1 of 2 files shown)


karlschriek:
Hey @toshi-k, @kimwnasptd @DavidSpek. Looking at this as well as at https://github.com/kubeflow/kubeflow/pull/5871, I am wondering whether such filtering on namespace level couldn't be achieved more simply by using `queryParams` on the VirtualService rather than by building custom logic into the central dashboard application. 

Let's say for example we want to roll out the MLFlow Server in every profile Namespace, couldn't achieve this as follows without needing these two PRs?

```
kind: VirtualService
apiVersion: networking.istio.io/v1alpha3
metadata:  
  name: mlflow
  namespace: karl-schriek
spec:
  gateways:
    - kubeflow/kubeflow-gateway
  hosts:
    - '*'
  http:
    - match:
        - uri:
            prefix: /mlflow
          queryParams:
            ns:
              exact:  karl-schriek
      rewrite:
        uri: /
      route:
        - destination:
            host: mlflow.karl-schriek.svc.cluster.local
            port:
              number: 80
---
apiVersion: v1
kind: Service
metadata:
  name: mlflow
  namespace: karl-schriek
spec:
  selector:
    app: mlflow
  type: ClusterIP
  ports:
  - port: 80
    targetPort: 5000
```

I quickly hacked something together on this today. Haven't managed to get it working (even without the "queryParams" part I am not able to get it to work), but I think in theory it should be possible. Are you aware of anything that would prevent an implementation like this from working?

kimwnasptd:
Thank you for the contribution @toshi-k, the Dashboard changes for MLFlow are really looking good!

I'll take a more detailed look at this within the next couple of days.

@karlschriek IIUC we had also discussed about this alternative for the first PR in https://github.com/kubeflow/kubeflow/issues/5863#issuecomment-847156432, but in the end as @toshi-k mentioned the query param gets only set the first time the user navigates to the app. 

For example, let say that the namespaced link was a grafana instance in a specific namespace. Once the user would navigate to it via the Central Dashboard then the request would contain the `?ns={namespace}` query parameter. All good. But then the frontend code of Grafana would not be adding the `ns={namsepace}` query params in the requests it makes, since it doesn't know anything about it.

@karlschriek did you have something else in mind?

karlschriek:
Ok, interesting. So if you click on "Grafana" in the kubeflow dashboard left-hand panel it works, but then inside the iframed Grafana dashboard if you navigate its subpages, the `?ns={namespace}` query parameter gets lost? Or is it the case of not being able to navigate from "Grafana" to "Notebooks" for example?

But @toshi-k's PR here addresses exactly that problem, correct? 

I didn't have anything else specific in mind, I just always prefer a solution that doesn't rely on application-level changes if it is on offer.

toshi-k:
@kimwnasptd Thank you ! I'm looking for your feedback.

@karlschriek I thought the same as you initially, but there are many issues to control traffic by queryParam. Although Kubeflow 1.3 is getting better to keep ?ns=namespace superficially, it doesn't work well. I think one of the problems is that ns parameter is deleted for the inside content of iframe.  
```
   _setIframeSrc() {
        const iframeUrl = new URL(this.subRouteData.path,
            window.location.origin);
        iframeUrl.hash = window.location.hash;
        iframeUrl.search = window.location.search;
        iframeUrl.searchParams.delete('ns');
        this.iframeSrc = iframeUrl.toString();
    }
```
https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/public/components/main-page.js#L311

I commented out `iframeUrl.searchParams.delete('ns')` at once, it caused side effects and failed to pass the tests.

Another concern is that some applications need additional API accesses between the app server and the client (UI). queryParam would not be added to all API accesses and it would cause errors.

As I said in #5863 , I concluded that path parameter is the most stable way to access third party app through kubeflow dashboard. If you find a perfect solution realized by queryParam, please share it in the community.

toshi-k:
@kimwnasptd My PR works perfectly for MLFlow as the attached image !

mlflow

toshi-k:
Sorry, I found a kind of another issue. I'm going to investigate it.
/hold

toshi-k:
OK, I found a solution ! This PR is ready to be merged.
/unhold

toshi-k:
@kimwnasptd You can use below manifests to deploy MLFlow temporarily for ns=workspace1 !
```
apiVersion: v1
kind: Service
metadata:
  name: mlflow
  namespace: mlflow
spec:
  ports:
  - name: http
    port: 5000
    targetPort: 5000
  selector:
    app: mlflow
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: mlflow-deployment
  namespace: mlflow
spec:
  replicas: 1
  selector:
    matchLabels:
      app: mlflow
  template:
    metadata:
      labels:
        app: mlflow
    spec:
      containers:
      - command:
        - mlflow
        - server
        - --host
        - 0.0.0.0
        - --static-prefix
        - /mlflow/workspace1
        image: larribas/mlflow:latest
        imagePullPolicy: Always
        name: mlflow
        ports:
        - containerPort: 5000
          name: http
          protocol: TCP
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
---
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: mlflow-vs
  namespace: mlflow
spec:
  hosts:
  - "*"
  gateways:
  - kubeflow/kubeflow-gateway
  http:
  - match:
    - uri:
        prefix: /mlflow/workspace1
    route:
    - destination:
        port:
          number: 5000
        host: mlflow
```
MLFlow URL from the central dashboard would be https://gateway/_/mlflow/workspace1/?ns=workspace1.

kimwnasptd:
@toshi-k apologies for the delayed review on this and thank you very much for the YAMLs!

Taking a look at the PR

components/centraldashboard/public/components/main-page.js
```diff
@@ -106,6 +106,8 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
                 // eslint-disable-next-line max-len
                 computed: 'computeShouldFetchEnv(registrationFlow, workgroupStatusHasLoaded)',
             },
+            matchingIndex: Number,
```
kimwnasptd, line 109: Why did you need to add this to the properties? IIUC it is only used from the `_setActiveMenuLink` method

kimwnasptd (commented):
@toshi-k I also tested the app and it works as expected, nice work!

components/centraldashboard/public/components/main-page.js
```diff
@@ -360,19 +374,23 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         });
         if (hashPath) {
             matchPath = path + '#' + hashPath;
-            matchingLink = allLinks
-                .find((l) => this.compareLinks(l, matchPath));
+            this.matchingIndex = allLinks
+                .findIndex((l) => this.compareLinks(l, matchPath));
         } else {
```
toshi-k, line 379: @kimwnasptd Good point ! When the namespace is changed by the namespace selector, both current URL and menu links are updated. If current URL is updated earlier, the new URL(e.g /mlflow/workspace1) can't match with the prefix of old menu links (e.g /mlflow/workspace2).

In that case, all `path.startsWith(link)` may be False, and matchingIndex is not updated. Making it class variable `this.matchingIndex`, UI can remember which menu item is selected, and re-link the URL and menu-bar.

components/centraldashboard/public/components/main-page.js
toshi-k: I answerer that question in a below window.

components/centraldashboard/public/components/main-page.js
kimwnasptd: Makes sense, thanks for the explanation!

kimwnasptd:
@toshi-k thank you for your work! I've also tested this locally and in my cluster and works as expected.

/lgtm
/approve

toshi-k:
@kimwnasptd Thank you for your review !

I rebased my branch and updated this PR by below commands.
```
git pull origin interact-with-ns-selector --rebase
git push origin interact-with-ns-selector
```
But it seems that another problem occurred. Did I made a mistake ?
I appreciate if you give me some instructions.

kimwnasptd:
It seems that the rebase included other commits as well. Could you do the following:
```bash
# Assuming that "public" is the kubeflow/kubeflow git repo
# Bring you local master up to date with upstream master
git checkout master
git pull public master

# rebase only your commits on top of the latest master
git checkout interact-with-ns-selector
git rebase -i master

# In the editor change the 'pick' word to 'drop' for any commit that you don't want.
# You should only select your three commits: 
# Interact with the namespace selector
# Use namespace given as an argument
# Keep the index of selected item

# finally force push your branch
git push origin HEAD -f
```

toshi-k:
@kimwnasptd Thank you for your advice !
I modified my PR by force push.
I think this PR has only my 3 commits and my remote repo is at the latest master branch.
https://github.com/toshi-k/kubeflow/commits/interact-with-ns-selector

PR #5997, opened 2021-06-21
Update last batch of notebook server Dockerfiles

DavidSpek:
This is the last PR to update the notebook server dockerfiles so they include the latest packages. After this PR, the `spawner_ui_config.yaml` needs updating so the new images are included for the upcoming 1.3.1 release of Kubeflow.


thesuperzapper:
Looks good!
/lgtm
/approve

PR #6001, opened 2021-06-21
WIP: Update CRUD web apps from Angular 8 to Angular 11 or 12

DavidSpek:
Closes: #5525

Angular 8 has been deprecated for a while now, with Angular 11 being the latest version. Sadly, you cannot skip version so the process to update from 8 to 11 (or 12) was to first update the common code to version 9, then the web apps and confirm they are working properly. Then again for version 9 to 10 and once more for 10 to 11 (or 12).

During the update process from Angular 8 to 9 of the web apps some errors are shown that don't seem to make much sense. From what I can find this is due to the use of npm link to the common frontend code and this bug has been fixed in Angular 10. For this reason, I ignored the errors and tested the web app functionality manually. As no errors were logged and nothing seemed to be broken I continued. I manually tested the web apps after each update and saw no errors or problems.

I've added the commands executed for (almost) every commit in the description. To prevent getting stuck with outdated software for longer than necessary, and given the inevitable size of this PR, I would like to suggest that this PR be reviewed by manually running the resulting web-apps and confirming nothing out of the ordinary is seen there. After this PR has been merged the next step would be to look more closely into what features are now deprecated in Angular and what new features replace them. A useful overview for this can be found here.

/cc @kimwnasptd @StefanoFioravanzo @thesuperzapper @elikatsis @yanniszark @wg102


DavidSpek:
I just realized that creating the new branch went wrong and thus the latest merges were omitted when I started the upgrade process. I'll start over tomorrow to solve this problem.

PR #6010, opened 2021-06-23
notebooks: Add additional PGP server for RStudio package validation

DavidSpek:
As seen [here](https://argo.kubeflow-testing.com/workflows/kubeflow-test-infra/kubeflow-kubeflow-presubmit-nb-rstudio-6007-10af3fc-1328-1171?tab=workflow&nodeId=kubeflow-kubeflow-presubmit-nb-rstudio-6007-10af3fc-1328-1171-4006263469), the PGP server used to validate the RStudio binary is down, which is causing the image build to fail and as a consequence blocking the 1.3.1 release.

To solve this, this PR adds the PGP keyserver `pgp.surfnet.nl` as a secondary option. SURFnet is an organization that develops, implements and maintains the national research and education network in The Netherlands, which is why I believe it is a good candidate (at least for now). I also tested `pgp.mit.edu` which did not work for me.

/cc @kimwnasptd


kimwnasptd:
Thanks @DavidSpek! 

/lgtm
/approve

PR #6013, opened 2021-06-23
profile-controller: Limit access to serving paths to KNative sa

davidspek:
In https://github.com/kubeflow/kubeflow/pull/5848 the profile controller was updated to allow access to paths in user namespaces that KNative requires for launching workloads. However, the implementation allows all access to the `/healthz`, /metrics` and ``/wait-for-drain` paths in a user's namespace. This means that a user can get data from these (commonly used) paths from all other user namespaces, breaking multi-user isolation. This PR fixes this by only allowing the service account used by KNative access to these paths in user namespaces. It also adds 2 more paths, `/ready` which is used by KNative, and `/v1/models/*` which is used by KFServing as outlined in [this comment](https://github.com/kubeflow/kfserving/issues/1558#issuecomment-833271258).

/cc @kubeflow/wg-notebooks-leads @yanniszark @kubeflow/wg-serving-leads

### DIFF
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -450,6 +450,15 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 				},
 			},
 			{
+				From: []*istioSecurity.Rule_From{
+					{
+						Source: &istioSecurity.Source{
+							Principals: []string{
+								"cluster.local/ns/knative-serving/sa/controller",
+							},
+						},
+					},
+				},
 				To: []*istioSecurity.Rule_To{
 					{
 						Operation: &istioSecurity.Operation{
@@ -459,7 +468,9 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 							Paths: []string{
 								"/healthz",
 								"/metrics",
+								"/ready",
 								"/wait-for-drain",
+								"/v1/models/*",
 							},
 						},
 					},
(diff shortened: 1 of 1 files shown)


DavidSpek:
CI is failing because the PR needs to be rebased, but I am delaying doing this since it means the deployment manifests for the profile controller will need to be changed because of https://github.com/kubeflow/kubeflow/pull/5761. Once this PR is closer to being in a final state I will rebase it.

alembiewski:
Thanks for the fix @DavidSpek! It worked for me, but it seems like we also need to add `"/v2/models/*"` to the list to support  [predictor protocol v2](https://kserve.github.io/website/modelserving/inference_api/#predict-protocol-version-2) of KServe.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -459,11 +459,20 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 							Paths: []string{
 								"/healthz",
 								"/metrics",
+								"/ready",
 								"/wait-for-drain",
+								"/v1/models/*",
```
surajkota, line 464: Can we add `/v2/models/*` as well?

juliusvonkohout:

components/profile-controller/controllers/profile_controller.go
umka1332:
```suggestion
								"/v1/models/*",
								"/v2/models/*",
```

thesuperzapper:
Let's discuss this more on this PR: https://github.com/kubeflow/kubeflow/pull/6627

PR #6022, opened 2021-06-28
feat(centraldashboard): ad i18n

JessicaBarh:
This PR implements internationalization to the centraldashboard
Resolves https://github.com/StatCan/kubeflow/issues/57


wg102:
@googlebot I consent.

JessicaBarh:
For the i18n of centraldashboard there's really two ways we can do this. The way we are currently implementing it is using one large `json` file (`languages.json`) that includes all the translated text for each object for every language - currently containing `en` and `fr`. The other way would be to have a separate `json` file for each language, this one would require a little refactoring to the current branch. 

If there is a preferred way that should be followed for consistency, please share your thoughts!

PR #6004, opened 2021-06-22
Update CRUD web apps from Angular 8 to Angular 12

DavidSpek:
Closes: #5525

Angular 8 has been deprecated for a while now, with Angular 12 being the latest version. Sadly, you cannot skip version so the process to update from 8 to 12 was to first update the common code to version 9, then the web apps and confirm they are working properly. Then again for version 9 to 10 and again for 10 to 11 and once more for 11 to 12.

During the update process from Angular 8 to 9 of the web apps some errors are shown that don't seem to make much sense. From what I can find this is due to the use of npm link to the common frontend code and this bug has been fixed in Angular 10. For this reason, I ignored the errors and tested the web app functionality manually. As no errors were logged and nothing seemed to be broken I continued. I manually tested the web apps after each update and saw no errors or problems.

I've added the commands executed for (almost) every commit in the description. To prevent getting stuck with outdated software for longer than necessary, and given the inevitable size of this PR, I would like to suggest that this PR be reviewed by manually running the resulting web-apps and confirming nothing out of the ordinary is seen there. After this PR has been merged the next step would be to look more closely into what features are now deprecated in Angular and what new features replace them. A useful overview for this can be found here.

/cc @kimwnasptd @StefanoFioravanzo @thesuperzapper @elikatsis @yanniszark @wg102

PR #6034, opened 2021-07-05
WA: Use relative paths for fetching translation files

### LINKED ISSUE #6033: Web apps do not render the translated text properly
/kind bug

The {Jupyter,Volumes,TensorBaords} web apps do not load the translation files and don't render as expected.

To reproduce this use one of the latest images from master, that includes the translation PR changes https://github.com/kubeflow/kubeflow/pull/5880, i.e. `public.ecr.aws/j1r0q0g6/notebooks/jupyter-web-app:master-f7671f7d`.

Screenshot from 2021-07-05 21 22 23

The web app to render the text as expected

The issue is due to the fact that the web apps try to load the `en.json` file from `/static/assets/il8n/en.json` path, which is incorrect since the app is exposed under a prefix. The correct path should be `/jupyter/static/assets/il8n/en.json`. This is due to the path defined in the translate `TranslateHttpLoader` object for loading the file https://github.com/kubeflow/kubeflow/blob/1a2780c80d427cc432f8e3295c0ab4952905629b/components/crud-web-apps/jupyter/frontend/src/app/app.module.ts#L40

@thesuperzapper did you encounter this while testing the PR https://github.com/kubeflow/kubeflow/pull/5880?
cc @wg102 @Jose-Matsuda 

**Environment:**
- Kubeflow version: 1.3

kimwnasptd:
fixes #6033 

The web apps should use the relative path `static/assets/il8n/` to fetch the translation files. This way the browser will construct the correct path, by using the `<base href="...">` element, even if the app lives under a prefix.

/cc @tasos-ale 
/assign @StefanoFioravanzo

### DIFF
--- a/components/crud-web-apps/jupyter/frontend/src/app/app.module.ts
+++ b/components/crud-web-apps/jupyter/frontend/src/app/app.module.ts
@@ -37,5 +37,5 @@ import { HttpClientModule, HttpClient } from '@angular/common/http';
 export class AppModule {}
 
 export function createTranslateLoader(http: HttpClient) {
-  return new TranslateHttpLoader(http, '../static/assets/i18n/', '.json');
+  return new TranslateHttpLoader(http, 'static/assets/i18n/', '.json');
 }
--- a/components/crud-web-apps/tensorboards/frontend/src/app/app.module.ts
+++ b/components/crud-web-apps/tensorboards/frontend/src/app/app.module.ts
@@ -59,5 +59,5 @@ import { HttpClientModule, HttpClient } from '@angular/common/http';
 export class AppModule {}
 
 export function createTranslateLoader(http: HttpClient) {
-  return new TranslateHttpLoader(http, '../static/assets/i18n/', '.json');
+  return new TranslateHttpLoader(http, 'static/assets/i18n/', '.json');
 }
--- a/components/crud-web-apps/volumes/frontend/src/app/app.module.ts
+++ b/components/crud-web-apps/volumes/frontend/src/app/app.module.ts
@@ -62,5 +62,5 @@ import { HttpClientModule, HttpClient } from '@angular/common/http';
 export class AppModule {}
 
 export function createTranslateLoader(http: HttpClient) {
-  return new TranslateHttpLoader(http, '../static/assets/i18n/', '.json');
+  return new TranslateHttpLoader(http, 'static/assets/i18n/', '.json');
 }


DavidSpek:
@kimwnasptd This is something that I needed to fix in https://github.com/kubeflow/kubeflow/pull/6004 due to a bug in Angular that was solved in newer versions.

tasos-ale (approved):

kimwnasptd:
> @kimwnasptd This is something that I needed to fix in #6004 due to a bug in Angular that was solved in newer versions.

#6004 also has this bug, since it's using incorrect paths https://github.com/kubeflow/kubeflow/blob/eb758a39d550e70a4f5aec1a2f173e14c77b142a/components/crud-web-apps/jupyter/frontend/src/app/app.module.ts#L40
If you are referring to a different bug wrt the `ngx-translate` module and upgrading to Angular 12 then please make it clear in the PR.

We must move forward with this PR since:
1. The web apps are currently **broken**
2. #6004 should only aim to upgrade the Angular version, not fixing other bugs. But even if it does there should be an explicit heads up about it
3. We can't wait until #6004 is merged to fix this issue, since #6004 and #6033 are orthogonal

/unhold

@StefanoFioravanzo this should be ready for merging, since @tasos-ale is OK with the changes as well.

StefanoFioravanzo:
@kimwnasptd Yes I definitely approve this change

/approve

DavidSpek:
@kimwnasptd There was a bug in Angular regarding fetching files from relative paths, that was fixed in a newer version of Angular and thus I needed to fix that in the Angular upgrade PR as I was upgrading. Furthermore, the extremely simple fix in this PR can be very easily applied after the Angular upgrade has been done if it is still necessary. Merging this PR first will likely break the Angular upgrade PR, which means going through the entire process again, wasting everybody's time (especially mine, since I've already gone through the upgrade procedure 4 times now, which takes around 5 hours each time).

I strongly suggest reevaluating merging this PR before the Angular upgrade has been completed.
/hold

DavidSpek:
Also, I seem to recall the Jupyter Web App was broken for over 2 months due to the entry point, 1.5 months of which was me trying to get my PR that fixed that single line merged. So I don't understand the sudden urgency with this PR, while my has been open for 2 weeks already. It is precisely to avoid this situation that I repeatedly urged everybody that the Angular upgrade PR needs to be reviewed and merged quickly.

DavidSpek:
@kimwnasptd To further elaborate on the bug in Angular that has since been fixed in later versions. The `../` notation that is being removed in this PR is precisely what needed to be added for the fonts during the Angular upgrade. See:
https://github.com/kubeflow/kubeflow/pull/6004/commits/7694fc01ccf78f0c85e36f51df562202d8804261

kimwnasptd:
> @kimwnasptd There was a bug in Angular regarding fetching files from relative paths, that was fixed in a newer version of Angular and thus I needed to fix that in the Angular upgrade PR as I was upgrading.

Are you referring to relative paths for files needed during the build process, like fonts? I've seen that you've changed these paths in the upgrade PR, so I'm guessing you are referring to these right? If yes then no worries, this relative path is not used during the compilation process but for the request that will be made to fetch the translation json file.

> Merging this PR first will likely break the Angular upgrade PR

Don't worry, it won't. I just rebased your PR on top of this branch and everything build as expected and the translation file is fetched correctly. So this shouldn't require any more of your time :slightly_smiling_face: 

> So I don't understand the sudden urgency with this PR

It's not that much of an urgency, although having broken images **is** something we should always fix asap, but more of an organizational matter. This small issue should be addressed independently by a separate PR and not by a single huge PR that addresses everything. It will also be easier in the future to see distinct commits for distinct fixes [for example if we need to cherry-pick things when working on patch releases]

DavidSpek:
If the rebasing indeed didn't cause issues than this should be good to go. But I'd still like to highlight that the Angular upgrade should be merged ASAP to avoid problems (also for others opening new PRs that will need rebasing).

/lgtm
/unhold

kimwnasptd:
>But I'd still like to highlight that the Angular upgrade should be merged ASAP to avoid problems 

@DavidSpek yes! I have already started reviewing this. I actually bumped into this while testing your PR, which lead to this PR

PR #6035, opened 2021-07-06
Add dropdown for existing volumes when creating a Notebook

robbertvdg:
I noticed with the new jupyter-web-app (refactor to use common code) that the dropdown for existing volumes was missing. 
I recreated the feature by copying parts of the old web app, so if I missed anything please let me know. It has been tested in our Kubeflow environment.


robbertvdg:
@DavidSpek can you take a look?

kimwnasptd:
Thank you for the PR @robbertvdg, this is indeed a regression when we moved to the updated code. Thanks for taking the time to fix it!

Once I wrap the review of #6004 I'll get to this as well. Although you might need to rebase your code on top of the latest iteration then, but hopefully this won't require too much effort.

Issue #5149, opened 2020-07-20
PodDefault never applies to Spark pods


slenky:
/kind bug

**What steps did you take and what happened:**
I am using PySpark in Jupyter Notebook to provision the Spark executor pods in my user namespace using this code:
```python
import os
import socket
from pyspark.sql import SparkSession

namespace = os.environ['WORKSPACE']
imageName = "slenky/spark-py:3.0.0-python"

spark = SparkSession\
    .builder\
    .appName("pythonspark")\
    .master("k8s://https://{ip}:{port}".format(ip=os.environ['KUBERNETES_SERVICE_HOST'], port=os.environ['KUBERNETES_SERVICE_PORT'])) \
    .config("spark.driver.host", str(socket.gethostbyname(socket.gethostname()))) \
    .config("spark.kubernetes.namespace", namespace) \
    .config("spark.kubernetes.container.image", imageName) \
    .config("spark.submit.deployMode", "client") \
    .config("spark.kubernetes.executor.label.<MY POD DEFAULT LABEL>", "true") \
    .config("spark.kubernetes.driver.annotation.sidecar.istio.io/inject", "false") \
    .config("spark.kubernetes.executor.annotation.sidecar.istio.io/inject", "false") \
    .getOrCreate()
```
and I also have a simple PodDefault in that namespace to mount few volumes and add some environment variables and it doesn't work.

In the admission-controller-webhook logs I see:
```
PodDefault '<MY POD DEFAULT>' is not in the namespcae of pod 'pythonspark-a84aed736ca8e231-exec-1'
```
Same also happens if I use sparkoperator + SparkApplication CRD.

**What did you expect to happen:**
Admission-webhook adds my entries from PodDefault to Spark pods.

**Anything else you would like to add:**
That was tested both with PySpark 2.4.5 and 3.0.0

**Environment:**

- Kubeflow version: 1.0.2
- kfctl version: kfctl v1.0.2-0-ga47628
- Kubernetes platform: AWS/AKS
- Kubernetes version: 1.15

slenky:
Simplest workaround for this issue is using Spark 3.0.0 with podTemplate like this:
```
apiVersion: v1
kind: Pod
metadata:
  labels:
    <your poddefault label>
  annotations:
    sidecar.istio.io/inject: "false"
  namespace: <your notebook namespace>
```
and then add this to your SparkSession:
`.config("spark.kubernetes.executor.podTemplateFile", "<path to pod template file.yaml>")`

PR #6052, opened 2021-07-17
fix(admission-webhook): attach namespace to pod request if pod does not have it

### LINKED ISSUE #6027: PodDefault doesn't apply to Pods generated from Deployment
/kind bug

I defined a simple PodDefault that should inject a simple environnement variable:
```
kind: PodDefault
metadata:
  name: add-env
spec:
 selector:
  matchLabels:
    my-add-env: 'true'
 env:
 - name: "var"
   value: "my-value"
```

When I deploy a pod directly such as this one:
```
apiVersion: v1
kind: Pod
metadata:
  name: my-direct-pod
  labels:
    app: nginx
    my-add-env: 'true'
spec:
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
``` 
and the PodDefault is applied correctly

but if I deploy a Pod by using a Deployment object as this one:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
    my-add-env: 'true'
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
        my-add-env: 'true'
    spec:
      containers:
      - name: nginx
        image: nginx:1.14.2
        ports:
        - containerPort: 80
```
the PodDefault is not applied, whereas the label that should trigger the PodDefault clearly exist in the Pod that is created by the deployement.

The PodDefault should be applied to the pod created by the deployement.

**Environment:**

- Kubeflow version: 1.2 and 1.3
- Kubernetes platform: on-prem kubernetes 
- Kubernetes version: 
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-16T12:52:14Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"darwin/arm64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.9-r0-CCE20.8.1.3.B001-17.36.5", GitCommit:"d61c11191a704802f93e5d86877a8fcffe208d58", GitTreeState:"clean", BuildDate:"2020-10-16T09:38:58Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}

henrysecond1:
Fixes #6027

When pod is deployed via `Deployment` object, namespace can be missing on `Pod` object in admission review request.

In this case, `PodDefault` can not be attached to pods since they do not have a namespace, 
and filtered by `filterPodDefaults` like below.
- https://github.com/kubeflow/kubeflow/blob/66bf469afdc1fc6bfc887c8c823234cc2b70a5d6/components/admission-webhook/main.go#L84

Since admission review request itself contains a namespace field, we can avoid problem above by attaching namespace to pod request if pod does not have it. I verified that `PodDefault` works correctly on deploying pod via `Deployment` object.

Also I just ran `go mod tidy` for cleaning unused dependencies.

Please take a look, thanks.

### DIFF
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -464,6 +464,7 @@ func mutatePods(ar v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 	reviewResponse := v1beta1.AdmissionResponse{}
 	reviewResponse.Allowed = true
 	if pod.Namespace == "" {
+		klog.Infof("Namespace was not set explicitly in Pod manifest, falling back to the namespace-'%s' coming from AdmissionReview request", ar.Request.Namespace)	
 		pod.Namespace = ar.Request.Namespace
 	}
 
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -82,7 +82,7 @@ func filterPodDefaults(list []settingsapi.PodDefault, pod *corev1.Pod) ([]*setti
 		}
 		// check if the pod namespace match the poddefault's namespace
 		if pd.GetNamespace() != pod.GetNamespace() {
-			klog.Infof("PodDefault '%s' is not in the namespace of pod '%s' ", pd.GetName(), pod.GetName())
+			klog.Infof("PodDefault '%s/%s' is not in the namespace of pod '%s/%s'", pd.GetNamespace(), pd.GetName(), pod.GetNamespace(), pod.GetName())
 			continue
 		}
 		klog.V(4).Infof("PodDefault '%s' matches pod '%s' labels", pd.GetName(), pod.GetName())
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -82,7 +82,7 @@ func filterPodDefaults(list []settingsapi.PodDefault, pod *corev1.Pod) ([]*setti
 		}
 		// check if the pod namespace match the poddefault's namespace
 		if pd.GetNamespace() != pod.GetNamespace() {
-			klog.Infof("PodDefault '%s' is not in the namespcae of pod '%s' ", pd.GetName(), pod.GetName())
+			klog.Infof("PodDefault '%s' is not in the namespace of pod '%s' ", pd.GetName(), pod.GetName())
 			continue
 		}
 		klog.V(4).Infof("PodDefault '%s' matches pod '%s' labels", pd.GetName(), pod.GetName())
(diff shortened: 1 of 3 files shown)


ekesken:
I was about to send a similar PR for this problem, and saw that one which is not approved since July 17.

@elikatsis @kimwnasptd @Bobgy @pdmack can you have a look?

This is the proper solution for this auto-closed issue as well: https://github.com/kubeflow/kubeflow/issues/5149

And it's a serious blocker for us, we try to inject `pip.conf` files using PodDefault, but since we create our notebooks via deployments it's not possible to use it at all.

components/admission-webhook/main.go
```diff
@@ -82,7 +82,7 @@ func filterPodDefaults(list []settingsapi.PodDefault, pod *corev1.Pod) ([]*setti
 		}
 		// check if the pod namespace match the poddefault's namespace
 		if pd.GetNamespace() != pod.GetNamespace() {
-			klog.Infof("PodDefault '%s' is not in the namespcae of pod '%s' ", pd.GetName(), pod.GetName())
+			klog.Infof("PodDefault '%s' is not in the namespace of pod '%s' ", pd.GetName(), pod.GetName())
```
ekesken, line 85:
```suggestion
			klog.Infof("PodDefault '%s/%s' is not in the namespace of pod '%s/%s'", pd.GetNamespace(), pd.GetName(), pod.GetNamespace(), pod.GetName())
```

components/admission-webhook/main.go
```diff
@@ -463,6 +463,10 @@ func mutatePods(ar v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 	}
 	reviewResponse := v1beta1.AdmissionResponse{}
 	reviewResponse.Allowed = true
+	if pod.Namespace == "" {
+		pod.Namespace = ar.Request.Namespace
+	}
```
ekesken, line 468:
```suggestion
	if pod.Namespace == "" {
		klog.Infof("Namespace was not set explicitly in Pod manifest, falling back to the namespace-'%s' coming from AdmissionReview request", ar.Request.Namespace)	
		pod.Namespace = ar.Request.Namespace
	}
```

kimwnasptd:
@ekesken both your review suggestions make sense. I'll wait a little bit if @henrysecond1 commits them. If not I'll just commit your proposed changes and merge the PR, to include it in the 1.4 release.

components/admission-webhook/main.go
henrysecond1: Thanks for the suggestion!

components/admission-webhook/main.go
henrysecond1: Thanks for the suggestion!

henrysecond1:
@ekesken @kimwnasptd  Thanks for checking it and the suggestion. I committed them

kimwnasptd:
@henrysecond1 thanks for the prompt reply!

/lgtm
/approve

PR #6077, opened 2021-08-05
Secure access to KFAM

DavidSpek:
Solves: https://github.com/kubeflow/kubeflow/issues/6228

This PR secures access to KFAM so it is not accessible to users.

/cc @thesuperzapper

### DIFF
--- a/components/centraldashboard/manifests/base/deployment.yaml
+++ b/components/centraldashboard/manifests/base/deployment.yaml
@@ -14,7 +14,7 @@ spec:
       labels:
         app: centraldashboard
       annotations:
-        sidecar.istio.io/inject: "false"
+        sidecar.istio.io/inject: "true"
     spec:
       containers:
       - name: centraldashboard
--- a/components/centraldashboard/manifests/overlays/istio/authorizationpolicy.yaml
+++ b/components/centraldashboard/manifests/overlays/istio/authorizationpolicy.yaml
@@ -1,0 +1,14 @@
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: central-dashboard
+spec:
+  action: ALLOW
+  rules:
+    - from:
+        - source:
+            principals:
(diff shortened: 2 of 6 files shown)


DavidSpek:
/freeze

@kubeflow/wg-notebooks-leads Not sure if you guys realize this, but without securing the KFAM API any user can take ownership of any namespace in the cluster. This includes the `kube-system` namespace, allowing a user to effectively take ownership of the entire cluster. It would have been nice to have this solved without exposing the vulnerability publicly, but given the lack of responsiveness I guess this is the only way forward.

It's probably a good idea to put out a CVE advisory related to this issue.

soleares:
I can confirm that during security testing we were able to fully take over a cluster from an uploaded pipeline step container/pod due to this issue. This fix remediates the issue.

thesuperzapper:
I am pretty sure this will not fully address the issue, as [`STRICT` MTLS mode](https://istio.io/latest/docs/concepts/security/#peer-authentication) will also be required to be enabled in the istio cluster/namespace to prevent non-istio access.

soleares:
@thesuperzapper This fix does block access to the kfam service from non istio pipeline step pods in user profile namespaces. Was there something additional you were thinking of?

kimwnasptd:
Regarding the `PERMISSIVE` mode for mTLS, AFAIK it's an orthogonal discussion. The mTLS setting is for deciding if traffic should be allowed between pods without a sidecar. 

But in this case's AuthorizationPolicy, Istio will only allow traffic to reach KFAM if the requests come from the CentralDashboard and this identity is deduced from sidecars. If a Pod without a sidecar tries to talk to KFAM then from an mTLS point of view it should be able to, in PERMISSIVE mode, but the AuthorizationPolicy would block the request since Istio can't deduce the workload id of that request [no sidecar].

Either way we need these AuthorizationPolicies, so I'm moving forward with the PR.

Thank you for your work @DavidSpek, and apologies for the too delayed review on this

/lgtm
/approve

PR #6096, opened 2021-08-16
Update image tags in 1.4 release branch

kimwnasptd:
Will use the `v1.4-rc.0` tag for the first RC.

/cc @elikatsis @yanniszark


kimwnasptd:
@PatrickXYS I see that the E2E tests workflows do not get triggered. Is this because it's on a branch other than `master`? Is there some manual work needed to be done to enable this?

PatrickXYS:
@kimwnasptd  This is expected since we haven't configured v1.4-branch in prow, you can add entry here 
https://github.com/kubeflow/testing/blob/master/aws/User/clusters/optional-test-infra-prow/namespaces/prow/configmap/config.yaml#L136-L138

PatrickXYS:
Prow may need 30second reconciler duration to sync changes, let's wait a bit

PatrickXYS:
Here we go

/test kubeflow-kubeflow-presubmit

kimwnasptd:
Thank you for the very quick support and help on this @PatrickXYS! :rocket:  

We are good to go!
cc @elikatsis

Issue #6124, opened 2021-09-08
300s timeout in notebook server Virtual Service causes Jupyter Notebook Terminal websocket to disconnect


abe-hpe:
/kind bug

**What steps did you take and what happened:**

1. Created a new notebook server in the Kubeflow Central Dashboard using  the 'tensorflow-1.15.2-notebook-cpu:latestC' image.
2. Connected to the notebook server and selected New -> Terminal from the Jupyter homepage.
3. Used the terminal for normal command line functions
4. At exactly 5 minues after creation, the terminal would stop responding and the JS console would show 'Websocket is already in CLOSING or CLOSED state'. This happens regardless of whether the terminal has been left idle, or is being actively used.

![2021-09-06_12-24](https://user-images.githubusercontent.com/56273479/132433251-8bd58a5b-8dcb-416c-b051-c2606f3857b9.png)

**What did you expect to happen:**

Terminal should continue to work properly for as long as the user keeps it open.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

The problem is caused by the 'timeout: 300s' inside the Istio Virtual Service that's created for each notebook server by the notebook controller. https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L485

We can prove it using these steps:

1. Scale the KF notebook controller deployment to 0 replicas
2. Manually edit the Virtual Service to increase or remove the timeout
3. Check that the terminal no longer disconnects at the 5 minute mark.

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): KF 1.3.
- kfctl version: (use `kfctl version`): not installed
- Kubernetes platform: (e.g. `minikube`) HPE Ezmeral Container Platform
- Kubernetes version: (use `kubectl version`): 1.20.2
- OS (e.g. from `/etc/os-release`): Centos 7.9, RHEL 7.9.

anencore94:
Is this issue is resolved from v1.4.0 by #6126, and could be closed now ?

PR #6126, opened 2021-09-08
Remove virtualservice timeout to prevent websocket disconnect

abe-hpe:
In the existing version, the 'timeout: 300s' added to the notebook's virtual service would cause websockets to disconnect at the 5 minute mark, causing the Jupyter Notebook web terminal function to hang. This is described in https://github.com/kubeflow/kubeflow/issues/6124.

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -482,7 +482,6 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
 					},
 				},
 			},
-			"timeout": "300s",
 		},
 	}
 


abe-hpe:
@googlebot I signed it!

thesuperzapper:
@kimwnasptd @DavidSpek I think this might be worth getting into Kubeflow 1.4 (if timing allows).

Do we know why there was previously a 300s timeout on the Istio VirtualService?

kimwnasptd:
I don't think there was a reason for this. With a quick `git blame` this line was introduced in https://github.com/kubeflow/kubeflow/pull/3104 and never touched again.

We should be OK with including this. Thanks for the fix @abe-hpe 
I'll also cut a new RC to include this

/lgtm
/approve

PR #6130, opened 2021-09-09
Central Dashboard/Iframe: fix special characters in url

munagekar:
The Central Dashboard Iframe is currently unable to handle special characters links in the iframe. And it produces erroneous links with double encoding of special characters.

Consider the following  Condensed Examples, I.e. URL containing the following characters were incorrectly encoded breaking the application.

| ascii | correct | centraldashboard |
|-------|---------|------------------|
| ["    | %5B%22  | %5B%2522         |
| "]    | %22%5D  | %2522%5D         |

To fix this I decoded the URI recieved by CentralDashboard. I tested this on my Kubeflow Cluster, and this worked correctly without affecting existing links.

Kubeflow Central Dashboard Iframe seems to cause a Double Encoding of Browser Links, so I added a decode to prevent it. I am not familiar with Polymer/UI development, so I don't know if this would be best way to fix it, but this seems to do the trick.

## Additional Background Context

MLFlow Metric Graphs would not work when in a Central Dashboard IFrame. The link looks like the following:
https://base-url/ns/mlflow/#/metric/foo?runs=["3869cd1170fc4ae0a677b20f9159eba5"]&experiment=3&plot_metric_keys=["foo"]&plot_layout={}&x_axis=relative&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=[]&last_linear_y_axis_range=[]

Correct Encoding

https://base-url/ns/mlflow/#/metric/foo?runs=[%223869cd1170fc4ae0a677b20f9159eba5%22]&experiment=3&plot_metric_keys=[%22foo%22]&plot_layout={}&x_axis=relative&y_axis_scale=linear&line_smoothness=1&show_point=false&deselected_curves=[]&last_linear_y_axis_range=[]

CentralDashboard messes up the URL, leading to the application breaking down

### DIFF
--- a/components/centraldashboard/public/components/iframe-container.js
+++ b/components/centraldashboard/public/components/iframe-container.js
@@ -53,7 +53,7 @@ export class IframeContainer extends PolymerElement {
         iframe.addEventListener('load', () => {
             const syncIframePage = () => {
                 const iframeLocation = iframe.contentWindow.location;
-                // This allows to the usage of URL's with special characters
+                // For Special characters within links inside iframe
                 const newIframePage = encodeURIComponent(
                     decodeURIComponent(
                         iframeLocation.href.slice(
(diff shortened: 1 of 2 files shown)


munagekar:
Update: This breaks Experiments UI.

knave07 (approved):

PR #6131, opened 2021-09-09
fix(admission-webhook): fix log level setting and flushing.

ekesken:
When a parameter like `-v=6` is passed, it was failing like this:

```
flag provided but not defined: -v
Usage of /webhook:
  -kubeconfig string
        Paths to a kubeconfig. Only required if out-of-cluster.
  -master string
        The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.
  -tlsCertFile string
        File containing the x509 Certificate for HTTPS. (default "/etc/webhook/certs/cert.pem")
  -tlsKeyFile string
        File containing the x509 private key to --tlsCertFile. (default "/etc/webhook/certs/key.pem")
```

Due to wrong order of `flag.Parse` call. I also made it sure that recent logs will be flushed on exit.


thesuperzapper:
@ekesken is this related to a specific issue, can you give more context?

ekesken:
I had come across this issue while debugging namespace problem resolved with this PR: https://github.com/kubeflow/kubeflow/pull/6052/

so currently it's not possible to set log level using command line parameters, do you want me to open an issue about it first?

ekesken:
@thesuperzapper do you expect me to open an issue first?

PR #6132, opened 2021-09-10
add lifecycle to pod default

pbarker:
Add lifecycle to pod default

fixes: https://github.com/kubeflow/kubeflow/issues/6145

### DIFF
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -443,6 +443,11 @@ func applyPodDefaultsOnContainer(ctr *corev1.Container, podDefaults []*settingsa
 	if err != nil {
 		klog.Error(err)
 	}
+	for _, pd := range podDefaults {
+		if pd.Spec.Lifecycle != nil {
+			ctr.Lifecycle = pd.Spec.Lifecycle
+		}
+	}
 	ctr.EnvFrom = envFrom
 }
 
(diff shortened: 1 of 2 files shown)


thesuperzapper:
@pbarker I am pretty hesitant to extend PodDefaults, as in the future we want to extend the notebook spawner UI (and remove the PodDefaults altogether).

I am working on a redesign of how the notebook spawner can be refactored around "resource groups" rather than the current "select each config like CPU/RAM/Affinity individually". 
* I think these "resource groups" could also allow specifying init-containers for use cases like yours.
* The main benefit of "resource groups" is that it lets the admin of the cluster define the full set of configs required to actually schedule a pod on a specific node, (which might include affinity, tolerations, etc). 
* The secondary benefit is that we can have the notebook-controller automatically reconcile towards these "resource groups" (set by an annotation), so users cant arbitrarily PATCH the notebook resource.

pbarker:
Hey @thesuperzapper thanks for the info, `resource groups` sounds like a nice way to go. 

I'm working right now on integrating [bitfusion](https://www.vmware.com/solutions/business-critical-apps/hardwareaccelerators-virtualization.html) into Kubeflow notebooks and it requires either init-containers or pod lifecycle to be configured. This would enable us to ship Kubeflow as the defacto ML platform for all VMware Kubernetes clusters and likely bring more people into the community.

Is there anyway we could make the change to `PodDefault` for the interim? It seems like a fairly innocuous change and would broaden the use cases for KF notebooks until `resource groups` are ready.

thesuperzapper:
@pbarker it's going to be a while until Kubeflow 1.5 (a couple of months at least), and any change we merge now won't be available in Kubeflow 1.4.

We could instead work together on getting "resource groups" working well for VMWare?

pbarker:
@thesuperzapper ah okay, yeah if we think we could get `resource groups` into 1.5 that would be awesome, happy to help out however I can

Issue #6154, opened 2021-09-26
Readiness probe for central dashboard


vijay-nag:
/kind bug
Did a rolling update on Central dashboard, new instance of central dashboard after upgrade did not serve requests for sometime.
**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
To test, configured to probe / path of central dashboard,
```
readinessProbe:
  httpGet:
    path: /
    port: 8082
```

```
kubectl get pods -n kubeflow -w
centraldashboard-54f6fb64c6-lwgc5                            0/1     Pending   0          0s
centraldashboard-54f6fb64c6-lwgc5                            0/1     Pending   0          0s
centraldashboard-54f6fb64c6-lwgc5                            0/1     ContainerCreating   0          0s
centraldashboard-54f6fb64c6-lwgc5                            0/1     ContainerCreating   0          2s
centraldashboard-54f6fb64c6-lwgc5                            0/1     Running             0          2s
centraldashboard-54f6fb64c6-lwgc5                            1/1     Running             0          19s
```
**What did you expect to happen:**
New instance of central dashboard is unable to serve traffic for sometime.
This issue shall be handled by configuring a readiness probe for central dashboard which helps graceful termination of old instances without having downtime.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
Adding /readyz endpoint in server.ts (central dashboard)
```
  app.get('/readyz', (req: Request, res: Response) => {
    res.json({
      codeEnvironment,
      message: `I am ready to serve traffic`,
    });
  });
```
Configuring readiness probe in centraldashboard manifests

```
        readinessProbe:
          httpGet:
            path: /readyz
            port: 8082
          initialDelaySeconds: 20
          periodSeconds: 10
```

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): `1.7.0`
- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`): `kubeadm cluster`
- Kubernetes version: (use `kubectl version`): `Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}`
- OS (e.g. from `/etc/os-release`):

PR #6155, opened 2021-09-26
Readiness Probe for Central dashboard

vijay-nag:
changed server.ts file to include /readyz endpoint for readiness probe

Could have used /healthz endpoint, but considering future development and
to distinguish probes have added a new endpoint.

Signed-off-by: Vijay Nag B S <vijay.bs@rakuten.com>

**Resolves**
https://github.com/kubeflow/kubeflow/issues/6154

### DIFF
--- a/components/centraldashboard/app/server.ts
+++ b/components/centraldashboard/app/server.ts
@@ -66,6 +66,12 @@ async function main() {
       message: `I tick, therfore I am!`,
     });
   });
+  app.get('/readyz', (req: Request, res: Response) => {
+    res.json({
+      codeEnvironment,
+      message: `I am ready to serve traffic`,
+    });
+  });
   app.use('/api', new Api(k8sService, metricsService).routes());
   app.use('/api/workgroup', new WorkgroupApi(profilesService, k8sService, registrationFlowAllowed).routes());
   app.use('/api', (req: Request, res: Response) => 


thesuperzapper:
@vijay-nag is there any specific reason why you need a separate endpoint to `/healthz`? Can't you just use the existing `/healthz` for both liveness and readiness probe?

vijay-nag:
> @vijay-nag is there any specific reason why you need a separate endpoint to `/healthz`? Can't you just use the existing `/healthz` for both liveness and readiness probe?

Hi @thesuperzapper yes, as central dashboard evolves I don't think health check would be sending 200 response. (https://github.com/kubernetes/apiserver/blob/master/pkg/server/healthz.go) illustrates health check endpoint for frontend services.
If readyz endpoint is not required now, then it is good to close this PR and configure readiness probe to probe /healthz.
@thesuperzapper need your views on this. Thanks

thesuperzapper:
@vijay-nag I mean, if there is not any difference between `/healthz` and `/readyz` I don't see the point of adding `/readyz`?

vijay-nag:
> @vijay-nag I mean, if there is not any difference between `/healthz` and `/readyz` I don't see the point of adding `/readyz`?

As of now there is no difference, will close this PR and use /healthz to probe for readiness.

PR #6160, opened 2021-09-28
feat: Authenticate to kubeflow pipelines by default

juliusvonkohout:
Authenticate to kubeflow pipelines by default according to https://github.com/kubeflow/pipelines/issues/5138

depends on https://github.com/kubeflow/pipelines/pull/6629

@Bobgy

### DIFF
--- a/components/crud-web-apps/jupyter/manifests/base/configs/spawner_ui_config.yaml
+++ b/components/crud-web-apps/jupyter/manifests/base/configs/spawner_ui_config.yaml
@@ -211,5 +211,6 @@ spawnerFormDefaults:
     # value:
     #   - add-gcp-secret
     #   - default-editor
-    value: []
+    value:
+    - access-ml-pipeline
     readOnly: false


thesuperzapper:
I am not sure it's safe to make this the default for two reasons:
1. not all kubeflow admins will want a KFP token mounted by default
2. not all "kubeflow notebooks" users will have "kubeflow pipelines" installed in their cluster (or at least, might not have the `access-ml-pipeline` `PodDefault` being created, which will cause errors).

---

As I raised in your other PR https://github.com/kubeflow/pipelines/pull/6629#issuecomment-930642835 these issues can be solved by removing the `pipeline-profile-controller` [python script](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py) from Kubeflow Pipelines, and replacing it with a new `ProfileResourceTemplate` CRD (managed by the `Profile Controller`).

_I will raise a separate issue to discuss adding the `ProfileResourceTemplate` CRD to the `Profile Controller`._

Bobgy:
Cc @zijianjoy

juliusvonkohout:
> I am not sure it's safe to make this the default for two reasons:
> 
>     1. not all kubeflow admins will want a KFP token mounted by default
> 
>     2. not all "kubeflow notebooks" users will have "kubeflow pipelines" installed in their cluster (or at least, might not have the `access-ml-pipeline` `PodDefault` being created, which will cause errors).
> 
> 
> As I raised in your other PR [kubeflow/pipelines#6629 (comment)](https://github.com/kubeflow/pipelines/pull/6629#issuecomment-930642835) these issues can be solved by removing the `pipeline-profile-controller` [python script](https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py) from Kubeflow Pipelines, and replacing it with a new `ProfileResourceTemplate` CRD (managed by the `Profile Controller`).
> 
> _I will raise a separate issue to discuss adding the `ProfileResourceTemplate` CRD to the `Profile Controller`._

can you point me to the new issue?

Are you thinking about something similar to https://github.com/redhat-cop/namespace-configuration-operator

thesuperzapper:
@juliusvonkohout regarding `ProfileResourceTemplate` I don't think there is an issue discussing it?

But it would be an extension to the existing profile controller (and `Profile` CRD), rather than moving to an external tool like that redhat thing.

---

Regarding the token, I think we should just make it easier for people to learn how to make it themselves (rather than forcing it on everyone by default).

Hence why the new "connecting to kubeflow pipelines" docs cover it (see docs PR: https://github.com/kubeflow/website/pull/3021), but I think we might want to add a link in the Kubeflow Notebooks docs that says people can use a PodDefault to include the token by default if they want.

Arhell (commented):
/ok-to-test

juliusvonkohout:
closed due to inactivity

PR #6176, opened 2021-10-07
Update Notebooks README regarding custom packages

kimwnasptd:
Right now the text mentions the recommended way is to have custom images. Some users using PVCs can get confused as to why the recommended way is to instead use custom images.

I slightly changed the wording to expose both options to the users so they can choose what works for their environment/workflows


thesuperzapper:
@kimwnasptd looks good!

/approve
/lgtm

PR #6184, opened 2021-10-12
Fix reference to PodPreset as PodDefault

gymaganis:
The manifest is not declaring a PodPreset (there's also a typo) but a PodDefault


thesuperzapper:
Thanks @gymaganis!

/approve

PR #6186, opened 2021-10-13
fix dashboard sending malformed headers to kfam

thesuperzapper:
Currently, the central-dashboard sends EVERY client header passed to its `/add-contributor/xxx` and `/remove-contributor/xxx` APIs on to kfam's `/kfam/v1/bindings` API. 
_(NOTE: these APIs are used in the "manage contributor" page to add/remove contributors)_

This means central-dashboard sends requests to `/kfam/v1/bindings` with invalid `:authority` headers (with the hostname of the istio gateway, rather than the internal kfam Service). 

This causes significant issues when you enable istio-sidecars on the kfam Pods, as istio will reject the "invalid" HTTP request, preventing you from adding/removing contributors.

---

This PR is very simple, it defines a list of "auth headers" which are safe to pass on, and deletes all others before forwarding to `/kfam/v1/bindings`.

_(NOTE: The only "auth header" that kfam actually reads is `USERID_HEADER`, however, I have included `cookie` and `authorization` in case users are validating these with their istio configs.)_

### DIFF
--- a/components/centraldashboard/app/api_workgroup.ts
+++ b/components/centraldashboard/app/api_workgroup.ts
@@ -118,7 +118,8 @@ export class WorkgroupApi {
     constructor(
         private profilesService: DefaultApi,
         private k8sService: KubernetesService,
-        private registrationFlowAllowed: boolean) {}
+        private registrationFlowAllowed: boolean,
+        private userIdHeader: string) {}
     /** Retrieves and memoizes the PlatformInfo. */
     private async getPlatformInfo(): Promise<PlatformInfo> {
         if (!this.platformInfo) {
@@ -214,8 +215,12 @@ export class WorkgroupApi {
                 namespace,
                 role: 'contributor',
             });
+            // only pass the auth-related headers from the user's request on to kfam
+            const authHeaders = ['authorization', 'cookie', this.userIdHeader];
             const {headers} = req;
-            delete headers['content-length'];
+            Object.keys(headers).forEach(
+                (key) => authHeaders.includes(key) || delete headers[key]
+            );
             const actionAPI = action === 'create' ? 'createBinding' : 'deleteBinding';
             await profilesService[actionAPI](binding, {headers});
             errIndex++;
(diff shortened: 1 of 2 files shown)


thesuperzapper:
@kimwnasptd could you quickly look over this PR so we can get it merged, its really small, but fixes an issue with the "manage contributors" page with the new istio.

thesuperzapper:
@kubeflow/wg-notebooks-leads I would really appreciate a quick review on this one, its really small, and would be good to get merged.

kimwnasptd:
Thanks for the PR @thesuperzapper, and apologies for the delayed review.

> This causes significant issues when you enable istio-sidecars on the kfam Pods, as istio will reject the "invalid" HTTP request

Could you further explain this one? What are these issues?

thesuperzapper:
@kimwnasptd it literally sends the `:authority` and `:path` headers from the user request, making the proxied HTTP request invalid.

For example, when the user GETs the `kubeflow.example.com/add-contributor/xxx` path, causing the central dashboard to proxy the request to `kfam.kubeflow.svc.cluster.local/kfam/v1/bindings`, it will send along an `:authority` header of `kubeflow.example.com`, which is invalid as it should be `kfam.kubeflow.svc.cluster.local`.

----

This PR lets the http JS library automatically set the required  `:authority` and `:path` headers, by only copying the `authorization`, `cookie`, and `{USERID_HEADER}` headers from the user's request to the proxied request.

kimwnasptd:
ACK on the changes and the rationale.

I want to understand if this is mostly a "good to have" improvement or if there are critical problems currently without it. That's why I want to understand the above point you raised:
> This causes significant issues when you enable istio-sidecars on the kfam Pods

What are these significant issues? How are they triggered?
This can be a good think to advertise in 1.5, that you fixed with this PR.

thesuperzapper:
@kimwnasptd Istio will reject the "add/remove contributor" requests without these changes, assuming sidecars are enabled, and the Service port is called `http`.

Having the `:authority` header disagree with the hostname you are sending the TCP packets to is a malformed HTTP request (so is rejected by istio).

If "add/remove contributor" is still working for you after this change, there is no harm in merging this very small PR so people using it these restrictive istio sidecars can use the "add/remove contributor" feature.

kimwnasptd:
ACK, now I'm fully in sync. Thanks for the technical details @thesuperzapper!

/lgtm
/approve

PR #6191, opened 2021-10-15
Use the workspace volume path from spanner_ui_config, rather than harcodecode

timcharper:
The workspace volume path of "/home/jovyan" was previously hardcoded. This is problematic for configurations which customize this value in the spanner_ui_config and created a disconnect in which the presented workspace path in the UI did not match what is actually used on the backend.

With this fix, the workspace path presented to the user in the UI is the same that is used when creating the statefulset for the notebook.

A previous fix was to use the `templatedPath` parameter passed back from the UI, but it seems semantically incorrect to treat a template as a materialized value.

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/utils.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/utils.py
@@ -105,6 +105,14 @@ def get_storage_class(vol):
         return vol["class"]
 
 
+def workspace_volume_from_config(config):
+    """
+    return the mountPath specified as specified in the spawner_ui configuration
+    :param config: loaded spawner_ui configuration
+    """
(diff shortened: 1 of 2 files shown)


timcharper:
Hi! :wave: Gentle ping on this. :bow:

kimwnasptd:
Hi @timcharper! Apologies for the huge delay on this one.

We also have an effort inflight to rework the app's whole logic for handling volumes https://github.com/kubeflow/kubeflow/pull/6321, to allow more advanced scenarios like:
* Fully controlling the new PVC spec
* Supporting other existing V1Volumes like NFS

We also fully allow users to define the mount path for both workspace and data volumes. @timcharper would this suffice for your use-case?

Arhell (commented):
/ok-to-test

PR #6193, opened 2021-10-16
web-apps(front) : change error msg for min

anencore94:
Current positive_number_input_component's error message was just saying it's negative.
However many web apps use such components with `min` as `1`.

So when user enter the number in a range (0, 1), such number was not a negative number but the error message was fixed.

Therefore I think we should change the error msg. (but I'm afraid then this component's naming became somewhat weird.)

Also, Since I'm not familiar with frontend, I'm not sure if I should manually change these i18n related files or not:
https://github.com/kubeflow/kubeflow/blob/ea5c89110971c7dc3141cfdc10a479358a3d7b9f/components/crud-web-apps/jupyter/frontend/i18n/messages.xlf#L42

Please let me know, If I should update those files too. Thanks in advance.

### Screenshots:

- Volumes web app Page
volumes-web-app-negative

- Jupyter web app Page
jupyter-web-app-bug


kimwnasptd:
This makes sense @anencore94!

Regarding the translation `.xlf` files it should be OK to not regenerate these files on every PR. It could be OK to have a periodic update, once we also update the other languages that need these files. So we should be OK for now.

/lgtm
/approve

PR #6206, opened 2021-10-27
fix(web-apps): Use relative path for --deploy-url build. #5566

zijianjoy:
Partial #5566

During standalone JWA deployment, js files are referenced with `/jupyter` prefix, instead it should be relative to `--base-href` parameter.


kimwnasptd:
Thanks for the PR @zijianjoy!

/lgtm
/approve

PR #6148, opened 2021-09-21
fix: tensorboard-controller is killed due to out of memory

juliusvonkohout:
I get "OOMkilled" with just 30 instead of 50 Mi memory limit

so i raise values from 20/30 to 30/50

PR #6216, opened 2021-11-10
fix(backend): tensorboard-controller does not work because of missing permissions

juliusvonkohout:
I have added the necessary permission.

https://github.com/kubeflow/kubeflow/blob/0c6c738efb95fbd320dc425c56c91fd6bcef5418/components/tensorboard-controller/controllers/tensorboard_controller.go#L148 tries to own the deployment but tensorboards/finalizers are not in the role

So you will get an error message similar to 
```
2021-11-05T12:02:24.832Z	ERROR	controller-runtime.controller	Reconciler error	{"controller": "tensorboard", "request": "dev-1/workspace-kulius-1", "error": "deployments.apps \"workspace-kulius-1\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>"}
```

### DIFF
--- a/components/tensorboard-controller/config/rbac/role.yaml
+++ b/components/tensorboard-controller/config/rbac/role.yaml
@@ -56,6 +56,7 @@ rules:
   - tensorboard.kubeflow.org
   resources:
   - tensorboards
+  - tensorboards/finalizers
   verbs:
   - create
   - delete


juliusvonkohout:
@elikatsis @kimwnasptd is there something missing from my side?

juliusvonkohout:
@thesuperzapper the reviewers seem to be idle again as in #6148 

A fix for a similar problem has been accepted here already https://github.com/kubeflow/pipelines/pull/6892

thesuperzapper:
@juliusvonkohout thanks for the PR, looks good to me.

/lgtm
/approve

PR #6218, opened 2021-11-13
Enable users to configure ephemeral storage when creating Notebook server

xiaowei-zillow:
Enable users to specify size of ephemeral storage when creating Notebook server, similar as how users specify cpu and ram

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -2,7 +2,7 @@ import json
 
 from werkzeug.exceptions import BadRequest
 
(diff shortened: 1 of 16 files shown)


kimwnasptd:
Thanks for the PR @xiaowei-zillow! I'd like to first understand the use case behind this feature.

Some questions regarding this that come to mind:
1. How/why would users want to define `ephemeralStorage`? 
2. How does this use case compare to using PVCs?
3. How would this work in case a Node would fail/reboot or a the node group would be scaled to zero?
4. What would happen if a Notebook would
    1. be schedule to node A
    2. user stops the Notebook
    3. the Notebook gets scheduled into a different node?

Arhell (commented):
/ok-to-test

PR #6225, opened 2021-11-22
Change namespace label for Katib metrics collector injection

andreyvelich:
Related: https://github.com/kubeflow/katib/issues/1622.

In the next releases we are going to use `katib.kubeflow.org/metrics-collector-injection: enabled` label for Katib Metrics Collector injection.

/cc @yanniszark @kimwnasptd @gaocegege @johnugeorge @tenzen-y


gaocegege (commented):
/lgtm

tenzen-y (commented):
/lgtm

andreyvelich:
> @andreyvelich does this mean that the metrics collector for the new version of Katib will not work with Kubeflow 1.4?
> 
> I think this will be a good thing to explicitly mention, since it's very probable that Katib in KF 1.5 will not be working with 1.4. This will also be a good thing for distros to know for their upgrade process as well from 1.4, since some existing namespaces might have the old annotations.

Yes, that means users have to modify manifests in KF 1.5 to use the latest version of Katib, and we will add it to our release notes in Katib 0.13
It's a good point about backward compatibility, we can discuss how to announce it on our next AutoML meeting on December 1st.

kimwnasptd:
ACK, thanks for confirming! Yes, explicitly mentioning this in the release's notes is a good way to raise awareness.

Also cc @zijianjoy, since he is also interested in changes that affect KF version upgrades.

/approve

zijianjoy:
@kimwnasptd Thank you for tagging me. Hello @andreyvelich, it will be great to share the Katib upgrade guide to Katib 0.13 (as in Katib release note) once it is ready. 

For GCP distribution, I think what we will need to do is to inject both old and new labels `katib-metricscollector-injection` and `katib-metricscollector-injection` for upgrade scenario. For new installation scenario, we will use the new label only.

PR #6230, opened 2021-11-30
admission-webhook-manifests: Update the cert manager from v1alpha2 to v1

haoxins:


haoxins:
see also https://github.com/kserve/kserve/pull/1929

haoxins:
@discordianfish can we merge this?

discordianfish (approved):

PR #6241, opened 2021-12-09
Update role.yaml

juliusvonkohout:
The role should be in line with https://github.com/kubeflow/kubeflow/blob/d43aee19ad8b556273a627ad34367e7e56b5c97e/components/crud-web-apps/jupyter/manifests/base/cluster-role.yaml#L24

```
  resources:
  - notebooks
  - notebooks/finalizers
  - poddefaults
```

@thesuperzapper


thesuperzapper:
@juliusvonkohout I agree that we should reduce the RBAC in this Role, but can we also fix up the other parts of this role while we are looking at it?

For example, not giving "*" for verbs (on the other resource types also).

juliusvonkohout:
> @juliusvonkohout I agree that we should reduce the RBAC in this Role, but can we also fix up the other parts of this role while we are looking at it?
> 
> For example, not giving "*" for verbs (on the other resource types also).

I could take the verbs from here https://github.com/kubeflow/kubeflow/blob/d43aee19ad8b556273a627ad34367e7e56b5c97e/components/crud-web-apps/jupyter/manifests/base/cluster-role.yaml#L24 for the cluster-role and remove create and delete for namespaces in the namespace role.

juliusvonkohout:
@thesuperzapper both roles contain the same rules now and i have removed the right to manage namespaces from both roles.

juliusvonkohout:
Now it should be ready.

juliusvonkohout:
@kimwnasptd @thesuperzapper i consider this to be a candidate for Kubeflow 1.5 rc1

components/crud-web-apps/jupyter/manifests/base/cluster-role.yaml
```diff
@@ -3,15 +3,6 @@ kind: ClusterRole
 metadata:
   name: cluster-role
 rules:
-- apiGroups:
-  - ""
-  resources:
-  - namespaces
-  verbs:
-  - get
-  - list
```
thesuperzapper, line 12: @juliusvonkohout Can you confirm that the jupyter-web-app still functions correctly without "get" and "list" on namespaces?

juliusvonkohout: That seems to be the case. I just updated the roles, restarted jupyter-web-app and notebook-controller and was able to start, stop, create, delete jupyterlabs. There  are also no errors in the logs.
[Uploading jupyter-web-app-deployment-7f7f55f9d8-slb8k-jupyter-web-app.log…]()

juliusvonkohout: Is there anything more to test or it is ready to be merged?

kimwnasptd: Should be good to go. It would also be a bug from the current code if it couldn't operate without listing namespaces, since it shouldn't need these permissions in the first place.

Note that we might need to re-introduce them though, in a future iteration when we would want the apps to be self-standing. But we need to discuss other items for this effort. But I'm also mentioning this context to keep in mind

kimwnasptd:
Moving on with merging the PR, to also include it in the RC1.

Thank you for your time @juliusvonkohout!
/lgtm
/approve

PR #6247, opened 2021-12-14
Configurable Virtual Service host in notebook server

vvnpan:
Enable operators to configure the Istio VirtualService host from environment variable instead of default `*`

In the central dashboard Kustomize files, the VirtualService host name can be configured  the a specified domain name. However, notebook servers this is defaulted to `*`. This causes the issue of not being able to access new notebook urls with the specified domain name. Allowing operators to configure the host parameters fixes this issue


components/notebook-controller/config/manager/kustomization.yaml
```diff
@@ -7,3 +7,4 @@ configMapGenerator:
   literals:
   - USE_ISTIO=true
   - ISTIO_GATEWAY=kubeflow/kubeflow-gateway
+  - ISTIO_HOST='*'
```
thesuperzapper, line 10: Because the field of the `VirtualService` is called `hosts` (not `host`) I think the variable should be: `ISTIO_HOSTS`

thesuperzapper (changes requested):
@vvnpan this is great!

I have been meaning to add this feature, thanks for doing it!

Please see my comments so we can get this merged ASAP.

kimwnasptd:
/ok-to-test

Thank you for the work @vvnpan! One small question, for educational purposes, since I haven't bumped into this yet. Could you provide just some more info of your setup? Are you also setting the `hosts` field in the `Gateway` objects? What about the other VirtualServices of the web apps?

I think it would be a good idea to start gathering such input and expose it in our docs. Like a guide for people that are using a custom cluster-domain

components/notebook-controller/config/manager/kustomization.yaml
vvnpan: Good call out, updated :)

vvnpan:
> /ok-to-test
> 
> Thank you for the work @vvnpan! One small question, for educational purposes, since I haven't bumped into this yet. Could you provide just some more info of your setup? Are you also setting the `hosts` field in the `Gateway` objects? What about the other VirtualServices of the web apps?
> 
> I think it would be a good idea to start gathering such input and expose it in our docs. Like a guide for people that are using a custom cluster-domain

@kimwnasptd 
Our setup previous to the notebooks was only pipelines and central dashboard. In both those setups we configure the `hosts` and `gateway` fields to one matching Unity standards. When we setup notebooks recently, the new notebook server pages would always return "page not found" using the default setup, the wildcard field there wasn't working as expected. The only way we could get this to connect to the notebook server page was to manually edit the `VirtualService` `hosts` field with kubectl after it is created.
If there's other solutions to this, I'm happy to learn about it :)

vvnpan:
@thesuperzapper could we get a review on this? Do you think it could get in the 1.6 release?

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -485,7 +485,12 @@ func generateVirtualService(instance *v1beta1.Notebook) (*unstructured.Unstructu
 	vsvc.SetKind("VirtualService")
 	vsvc.SetName(virtualServiceName(name, namespace))
 	vsvc.SetNamespace(namespace)
-	if err := unstructured.SetNestedStringSlice(vsvc.Object, []string{"*"}, "spec", "hosts"); err != nil {
+
+	istioHosts := os.Getenv("ISTIO_HOSTS")
+	if len(istioHosts) == 0 {
```
kimwnasptd, line 490: @vvnpan only one minor nit, could you instead use the `GetEnvDefault` helper we have here for this one?

https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/pkg/culler/culler.go#L60

Rest LGTM

vvnpan: @kimwnasptd we might have a trade off for the user readability here if we call the `culler. GetEnvDefault` function, it could be confusing seeing a function from the culler package within the `generateVirtualService` function. 
But if that's the structure the maintainers want to follow and break away from the [existing](https://github.com/kubeflow/kubeflow/blob/da6efb9c9fcdfe03a2932aacc6df32489f2883a3/components/notebook-controller/controllers/notebook_controller.go#L492) I will update with the changes

sergeyshevch:
@kimwnasptd Can we have this merged? I met the same issue when I try to expose kserve knative routes on same host as kubeflow.

juliusvonkohout:

thesuperzapper:
Closing in preference of https://github.com/kubeflow/kubeflow/pull/6902

PR #6260, opened 2021-12-22
centraldashboard: Update node and use latest-stable

### LINKED ISSUE #6259: CentralDashboard builds fail due to latest Chromium changes
/kind bug

I tried to build the CentralDashboard locally, but Karma is reporting errors with Chromium. We've also seen this in upstream PRs https://github.com/kubeflow/kubeflow/pull/6181#issuecomment-999709025.

I would expect the build to complete successfully.

The issue is that we are using the [`edge`](https://wiki.alpinelinux.org/wiki/Edge) branch, which is the DEV branch! https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/Dockerfile#L13-L19

We should instead use the `latest-stable` branch to avoid such situations. We can also see other reports on this https://forum.manjaro.org/t/cannot-open-chromium-on-x11-undefined-symbol-wl-proxy-marshal-flags/94734 and it's very possible to be related to https://pkgs.alpinelinux.org/packages?name=chromium&branch=edge.

/assign @kimwnasptd

kimwnasptd:
closes #6259 

In this PR I'm:
* Updating the node version to `v12.22.8`
* Using the `latest-stable` instead of `edge`

NOTE: I also updated the node version because I was seeing some `UNTRUSTED signature` errors when fetching alpine packages. Could be because of an old openssl package in the base image, but I'm not 100% sure.

The tests for the dashboard should be now passing

### DIFF
--- a/components/centraldashboard/Dockerfile
+++ b/components/centraldashboard/Dockerfile
@@ -1,5 +1,5 @@
 # Step 1: Builds and tests
-FROM node:12.18.3-alpine AS build
+FROM node:12.22.8-alpine AS build
 
 ARG kubeflowversion
 ARG commit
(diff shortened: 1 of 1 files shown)


elikatsis:
Thanks @kimwnasptd 

/lgtm
/approve

PR #6269, opened 2022-01-05
Fix unnecessary updates of istio AuthorizationPolicy

sunzhaochang:
AuthorizationPolicy is always updated although it hasn't changed. In fact, when checking if an update is needed, we need to compare the `Spec` field instead of the AuthorizationPolicy itself.


Arhell (commented):
/ok-to-test

PR #6295, opened 2022-01-21
proposal: Extend Notebook Controller to expose idleness for Jupyter

athamark:
Proposal for #6270

Provide a design doc as a proposal for extending Notebook Controller to
expose idleness for Jupyter. Our proposal is in markdown format and follows
the guidelines #6272 of the kubeflow/components/proposal/README.md guide.

Looking forward to your feedback!

PR #6297, opened 2022-01-21
notebooks: Extend Notebook Controller to expose idleness for Jupyter

athamark:
This is the code implementation for the proposal raised in the #6295.

We are aware that the standard process suggests to first send a proposal
to discuss the implementation details and not sending a code PR right away.
But due to the feature freeze being around the corner, we wanted to send
a PR to help people reviewing the code in parallel.

Of course, we will be making any changes that will be raised and discussed
in the proposal #6295.


kimwnasptd:
@athamark could you also do a rebase to the latest master here as well?

components/notebook-controller/pkg/culler/culler.go
```diff
@@ -156,51 +173,145 @@ func getNotebookApiStatus(nm, ns string) *NotebookStatus {
...
+	// No last-activity found in the CR. Setting to Now()
+	if _, ok := meta.GetAnnotations()[LAST_ACTIVITY_ANNOTATION]; !ok {
+		t := createTimestamp()
+		log.Info(fmt.Sprintf("No last-activity found in the CR. Setting to %s", t))
+
+		if len(meta.GetAnnotations()) == 0 {
+			meta.SetAnnotations(map[string]string{})
+		}
+		meta.Annotations[LAST_ACTIVITY_ANNOTATION] = t
 		return true
```
kimwnasptd, line 226: @athamark can you change this to a single `return`? The function's signature does not return a value

kimwnasptd (commented):
Good thing that you've also included more tests for the culler. But we'll need to rework on how we run the tests, both in the CI and with the Makefile.

But these can be next steps.

components/notebook-controller/pkg/culler/culler.go
athamark: @kimwnasptd thanks for the feedback. I corrected this, I pushed a commit that resolves this, you should be able to see it as `review: UpdateNotebookLastActivityAnnotation()`.

kimwnasptd:
Moving on with this, since the tests are also passing as well. This, alongside the https://github.com/kubeflow/kubeflow/pull/6321 should be the last big features as we wrap the feature freeze

/lgtm
/approve

Issue #6315, opened 2022-01-25
Common UX around the KF web apps


kimwnasptd:
/kind feature

**Why you need this feature:**
In the spirit of helping Kubeflow become more like a platform, and not just independent components, we should also aiming at making the UX offered by the web apps more common.

This can start with simple frontend changes, like how do we show the data in tables, all the way to what functionalities the apps should expose (like showing logs, events, etc).

**Describe the solution you'd like:**
Let's use this issue for brainstorming. If we start having some solid progress we can turn this into a [proposal](https://github.com/kubeflow/kubeflow/tree/master/components/proposals) and document these goals.

As a can think of the following good first steps:
* Mimic how KFP is showing their data into tables, and not cards like we have for Notebooks/Volumes/TensorBoards web apps
* Each web app should aim at completely replacing kubectl
* Having a distinct page for each object (Notebook CR, PVC etc)
    * Exposing logs from objects
    * Exposing K8s events from objects
    * Exposing YAML of the objects

Of course all of the above would need to be done in a secure way, by using `SubjectAccessReviews`.

**Anything else you would like to add:**
These are just some first ideas, but I'm really looking forward to more feedback!

andreyvelich:
Let's discuss it as part of: https://github.com/kubeflow/community/issues/780

PR #6316, opened 2022-01-25
frontend: Use responsive tables instead of cards

kimwnasptd:
A first step for https://github.com/kubeflow/kubeflow/issues/6315

Extend the frontends of our web apps to instead use responsive tables, to be more inline with what Pipelines are doing.

Here the new look of the apps
twa

jwa


kimwnasptd:
I fixed the failing unit tests as well. This should be ready for review.

/cc @StefanoFioravanzo

PR #6258, opened 2021-12-22
Fix(Centraldashboard): Upgrade the K8s node.js client and fix the brokens

haoxins:

PR #6326, opened 2022-02-05
build(central dashboard): Remove node v12 patch version numbers in the Dockerfile

haoxins:
Refs: https://github.com/kubeflow/kubeflow/pull/6260

I think it's better to remove the patch version number for LTS Node.js versions.

See also: https://github.com/kubeflow/kubeflow/pull/6258


Arhell (commented):
/ok-to-test

PR #6335, opened 2022-02-08
profiles: Update the permissions for notebook idleness

kimwnasptd:
Follow up for #6297 
Design in: https://github.com/kubeflow/kubeflow/blob/master/components/proposals/20220121-jupyter-notebook-idleness.md

Extend the Profiles Controller to create an AuthorizationPolicy for allowing Notebooks Controller to talk to the Notebooks' `/api/kernels` endpoint. Without it the Notebooks Controller will be getting 403 RBAC errors by Istio, since it won't be allowed to talk to Notebooks in other namespaces


kimwnasptd:
cc @athamark

Issue #5982, opened 2021-06-16
it's hard to delete notebook crd


mofanke:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

notebook controller image: public.ecr.aws/j1r0q0g6/notebooks/notebook-controller:v1.3.0-rc.1

when i delete notebook, pods State change like below
```
mofanke2-0    0/2     Terminating       0          55s
mofanke3-0    0/2     Pending           0          0s
mofanke3-0    0/2     Pending           0          0s
mofanke4-0    0/2     Pending           0          0s
mofanke2-0    0/2     Pending           0          0s
mofanke4-0    0/2     Pending           0          0s
mofanke2-0    0/2     Pending           0          0s
mofanke3-0    0/2     Init:0/1          0          0s
mofanke4-0    0/2     Init:0/1          0          0s
mofanke2-0    0/2     Init:0/1          0          0s
mofanke3-0    0/2     Terminating       0          4s
mofanke4-0    0/2     Terminating       0          4s
mofanke2-0    0/2     Terminating       0          4s
mofanke3-0    0/2     Terminating       0          56s
mofanke3-0    0/2     Terminating       0          56s
mofanke4-0    0/2     Terminating       0          56s
mofanke4-0    0/2     Terminating       0          56s
mofanke2-0    0/2     Terminating       0          56s
mofanke2-0    0/2     Terminating       0          56s
mofanke3-0    0/2     Pending           0          0s
mofanke3-0    0/2     Pending           0          0s
mofanke3-0    0/2     Init:0/1          0          0s
mofanke4-0    0/2     Pending           0          0s
mofanke4-0    0/2     Pending           0          0s
mofanke4-0    0/2     Init:0/1          0          0s
mofanke2-0    0/2     Pending           0          0s
mofanke2-0    0/2     Pending           0          0s
mofanke2-0    0/2     Init:0/1          0          1s
mofanke3-0    0/2     Terminating       0          4s
mofanke4-0    0/2     Terminating       0          4s
mofanke2-0    0/2     Terminating       0          5s
mofanke4-0    0/2     Terminating       0          55s
mofanke4-0    0/2     Terminating       0          55s
mofanke2-0    0/2     Terminating       0          55s
```

**What did you expect to happen:**
just delete notebook and pods, just one State (Terminating)

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

[...]

PR #6341, opened 2022-02-09
notebooks: Fix endless restarts

### LINKED ISSUE #6337: Notebook endless restarts
/kind bug

Deployed the latest notebook controller. @athamark and I noticed that in some cases the Notebook's state will keep on transitioning between PodInitializing and Terminating. We've also seen this behavior in https://github.com/kubeflow/kubeflow/issues/5982, but I'm not entirely sure if the root cause is the same.

The Notebook to remain as is.

After GETting the CR we notice that it had **126** `NB_PREFIX` env vars! The strange part is that the controller's business should not be touching the spec of the CR. So this is a side effect of our current implementation.

This seems quite important, since it can affect any installation.

**Environment:**

- Kubeflow version: master

kimwnasptd:
Closes #6337 

The changes in the PR fixed the endless restarts for me. I'm very confident that this is related to https://github.com/kubeflow/kubeflow/issues/5982, due to how we handle K8s events after a Notebook is deleted. But I wont' close that one unless we've confirmed its root cause.

/assign @kimwnasptd

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -286,11 +286,11 @@ func (r *NotebookReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	// Pod is found
 	// Check if the Notebook needs to be stopped
 	// Update the LAST_ACTIVITY_ANNOTATION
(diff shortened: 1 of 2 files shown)


elikatsis:
Thank you @kimwnasptd, your analysis in the issue makes total sense.
/lgtm
/approve

PR #6350, opened 2022-02-11
KFNBC: Build Dockerfile with -mod=mod flag

samuelvl:
Build notebook controller explicitly using the `-mod=mod` flag.

This is the default behaviour with the current builder image. But it  possible to build this image in other CI systems too, where the base image is replaced by a private image with vendoring enabled by default (-mod=vendor).


Arhell (commented):
/ok-to-test

PR #6414, opened 2022-03-25
Fix UI to support poddefault with match expressions

vinaynair:
PodDefaults , as defined in the [sample](https://github.com/kubeflow/kubeflow/tree/master/components/admission-webhook#how-this-works), with matchLabels work fine, **but** if we create a pod default with matchExpressions instead of using matchLabels, the backing API behind the UI, http://localhost:8080/jupyter/api/namespaces/kubeflow-user-example-com/poddefaults,  complains since API only supports/expects matchLabel selectors. 

Having said that the backing [admission webhooks label selector with matchExpressions works just fine](https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/main.go#L73) so this PR tries to fix the issue with the API behind the UI

Sample PodDefault that will cause the create notebook server to throw an error

```
kind: PodDefault
metadata:
  name: add-env
  namespace: kubeflow
spec:
 selector:
  matchExpressions:
   - key: add_env
       operation: Exists
 desc: "add env"
 env:
  - name: TEST_ENV
    value: "TEST_VAR"
```

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py
@@ -29,7 +29,14 @@ def get_poddefaults(namespace):
     # Return a list of (label, desc) with the pod defaults
     contents = []
     for pd in pod_defaults["items"]:
-        label = list(pd["spec"]["selector"]["matchLabels"].keys())[0]
+        if "matchLabels" in pd["spec"]["selector"]:
+            label = list(pd["spec"]["selector"]["matchLabels"].keys())[0] # get the first label key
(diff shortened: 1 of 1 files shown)


Arhell (commented):
Please sign CLA
https://github.com/kubeflow/kubeflow/pull/6414/checks?check_run_id=5685277890

vinaynair:
@StefanoFioravanzo  @elikatsis  any feedback on the PR ? Thx

vinaynair:
Any feedback on this PR ? We use this patch to Jupyter web app so that it behaves nicely when we have pods defaults that automagically applies specific tolerations or node selectors to “all” pods created within the profile/namespace.

PR #6418, opened 2022-03-28
Allow adding extra headers to rstudio notebook

missingcharacter:
I'm currently experiencing issue stated here: https://serverfault.com/questions/985253/istio-https-traffic-converted-to-http-with-port-set-as-443 but only when creating `rstudio` notebooks

I was able to solve this problem by adding Header `"X-Forwarded-Proto":"https"` which ends up looking like this:

```yaml
apiVersion: kubeflow.org/v1
kind: Notebook
metadata:
  annotations:
    notebooks.kubeflow.org/http-headers-request-set: >-
      {"X-RStudio-Root-Path":"/notebook/my-namespace/my-rstudio-notebook/",
      "X-Forwarded-Proto":"https"}
    notebooks.kubeflow.org/http-rewrite-uri: /
    notebooks.kubeflow.org/server-type: group-two
```

### POC of this code

```python
import os
import json

def get_rstudio_header():
    nb_name = "my-rstudio-notebook"
    nb_ns = "my-namespace"
    rstudio_header = {
        "X-RStudio-Root-Path": f"/notebook/{nb_ns}/{nb_name}/"
    }
    if os.environ.get("RSTUDIO_EXTRA_HEADERS"):
        extra_headers = json.loads(os.environ.get("RSTUDIO_EXTRA_HEADERS"))
        rstudio_header = {**rstudio_header, **extra_headers}
    return json.dumps(rstudio_header)


rstudio_header = get_rstudio_header()

print(f"rstudio_header is of type {type(rstudio_header)}")
print(f"rstudio_header content is {rstudio_header}")
```

Running it on a shell

```shell
$ python -V
Python 3.7.12
$ python test.py
rstudio_header is of type <class 'str'>
rstudio_header content is {"X-RStudio-Root-Path": "/notebook/my-namespace/my-rstudio-notebook/"}
$ RSTUDIO_EXTRA_HEADERS='{"X-Forwarded-Proto":"https"}' python test.py
rstudio_header is of type <class 'str'>
rstudio_header content is {"X-RStudio-Root-Path": "/notebook/my-namespace/my-rstudio-notebook/", "X-Forwarded-Proto": "https"}
```

### DIFF
--- a/components/crud-web-apps/jupyter/README.md
+++ b/components/crud-web-apps/jupyter/README.md
@@ -21,6 +21,10 @@ The annotation `notebooks.kubeflow.org/http-headers-request-set:`
 `'{"X-RStudio-Root-Path":"/notebook/<namespace>/<name>/"}'` is added to
 Notebook resources belonging to Group 2. This configures Istio to add
 this header to requests, which is necessary for images from Group 2 to work.
+IF you need to add extra headers to RStudio notebooks you may provide them
+by setting environment variable `RSTUDIO_EXTRA_HEADERS`
+to `'{"X-Custom-Header":"custom-header-value"}'`
+**Example:** `RSTUDIO_EXTRA_HEADERS='{"X-Forwarded-Proto":"https"}'`
 
 The Jupyter Web App displays the logos for each Notebook Server group
 in a button toggle in the Spawner UI. To easily identify the group of
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -1,4 +1,5 @@
 import json
+import os
 
 from werkzeug.exceptions import BadRequest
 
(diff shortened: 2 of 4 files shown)


Arhell (commented):
/ok-to-test

missingcharacter:
@thesuperzapper anything I should do for this to me worked on?

thesuperzapper:
@missingcharacter thanks for the PR! Can we implement this in a way that works for both HTTP and HTTPS (possibly by detecting which applies)?

What are your thoughts?

_PS: also, you need to rebase_

missingcharacter:
Hi @thesuperzapper ,

> Can we implement this in a way that works for both HTTP and HTTPS (possibly by detecting which applies)?

I would love this, but I was not able to figure out a way.
I believe `istio` (maybe the sidecar) is the one changing the header from `http` to `https` and somehow keeps port `443`, but I was not able to confirm this is true.

Also, I rebased.

hellobiek:
Should add this feature for support R with https.

danielschulz:
Thank you for your PR. This solved our issue.

missingcharacter:
@thesuperzapper since this seems to help other people, see @danielschulz and @hellobiek.

Do you think it is worth merging?

juliusvonkohout:

missingcharacter:
I cannot continue maintaining this fork, I will close this PR

Issue #6419, opened 2022-03-28
Create notebooks page and get [500] error


kindomLee:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Enter notebook page and get this error

```
[500] An error occured in the backend. https://xxx.endpoints.xxx.cloud.goog/jupyter/api/gpus
```

![image](https://user-images.githubusercontent.com/9668312/160374192-e891cdd9-2460-47a5-a58f-1be14d327364.png)

**What did you expect to happen:**
can success get GPU Vendor

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): gcp-blueprints 1.5.0
- Kubernetes platform: (e.g. `minikube`): gke
- Kubernetes version: (use `kubectl version`): GitVersion:"v1.20.15"

error log from `jupyter-web-app-deployment`

```
2022-03-28 09:40:09,681 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | Caught and unhandled Exception!
2022-03-28 09:40:09,681 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | Invalid value for `type` (KernelDeadlock), must be one of ['DiskPressure', 'MemoryPressure', 'NetworkUnavailable', 'PIDPressure', 'Ready']
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/src/apps/common/routes/get.py", line 66, in get_gpu_vendors
    nodes = api.list_nodes().items
  File "/usr/local/lib/python3.7/site-packages/kubeflow/kubeflow/crud_backend/api/node.py", line 6, in list_nodes
    return v1_core.list_node()
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/api/core_v1_api.py", line 16844, in list_node
    return self.list_node_with_http_info(**kwargs)  # noqa: E501
  File "/usr/local/lib/python3.7/site-packages/kubernetes/client/api/core_v1_api.py", line 16965, in list_node_with_http_info
    collection_formats=collection_formats)
[...]

PR #6422, opened 2022-03-29
Fix kubernetes python client to v22.6.0

benjamintanweihao:
We've just upgraded Kubeflow 1.5 (from 1.3) and have seen a similar issue raised by https://github.com/kubeflow/kubeflow/issues/6419.

On inspecting the error message:

```
Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:98.0) Gecko/20100101 Firefox/98.0"
2022-03-29 03:03:18,481 | apps.common.utils | INFO | Using config file: /etc/config/spawner_ui_config.yaml
2022-03-29 03:03:18,492 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | Caught and unhandled Exception!
2022-03-29 03:03:18,498 | apps.common.utils | INFO | Using config file: /etc/config/spawner_ui_config.yaml
2022-03-29 03:03:18,492 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | Invalid value for `type` (FrequentContainerdRestart), must be one of ['DiskPressure', 'MemoryPressure', 'NetworkUnavailable', 'PIDPressure', 'Ready']
```

I've checked that the Jupyter Webapp uses `kubernetes==v23.3.0`. The source of the error is this line: https://github.com/kubernetes-client/python/blob/v23.3.0/kubernetes/client/models/v1_node_condition.py#L217.

## The Fix

I've pinned the version of the Python Kubernetes client to use v22.6.0. I've rebuilt the image and tested this on my Kubeflow installation and the error goes away.

### DIFF
--- a/components/crud-web-apps/common/backend/setup.py
+++ b/components/crud-web-apps/common/backend/setup.py
@@ -3,7 +3,7 @@ import setuptools
 REQUIRES = [
     "Flask >= 1.1.1",
     "Flask-API >= 2.0",
-    "kubernetes >= 11.0.1",
+    "kubernetes == 22.6.0",
     "requests >= 2.22.0",
     "urllib3 >= 1.25.7",
     "Werkzeug >= 0.16.0",


kimwnasptd:
@benjamintanweihao could you verify that deleting objects [i.e. Notebooks] also works as expected?

I remember that there was a change in the newer version of the library on how to pass the deletion-policy, but I might be wrong.

kimwnasptd:
This should also fix https://github.com/kubeflow/kubeflow/issues/6427, since the new client will be using the `v1` API for SubjectAccessReviews

components/crud-web-apps/jupyter/backend/requirements.txt
```diff
@@ -1 +1,2 @@
+kubernetes==v22.6.0
```
kimwnasptd, line 1: Could you instead make this change into the common code, so that it takes effect for all the web apps?

https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/backend/setup.py#L6

benjamintanweihao: > @benjamintanweihao could you verify that deleting objects [i.e. Notebooks] also works as expected?
> 
> I remember that there was a change in the newer version of the library on how to pass the deletion-policy, but I might be wrong.

Just tried deleting the notebook and it's working :D. Sure thing, I'll also make the change everywhere.

benjamintanweihao: Done. :)

kimwnasptd:
@benjamintanweihao looks good, I also tried it out locally and everything worked as expected.

Thanks for the PR!

/lgtm
/approve

PR #6435, opened 2022-04-12
fix: stop unnecessary updates of profiles

henrysecond1:
Currently, profiles are always updated although they are not changed.

- https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/controllers/profile_controller.go#L270
- https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/controllers/profile_controller.go#L714

If `WorkloadIdentity` is empty, profiles are not changed, so we don't need to update profiles.

### DIFF
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -710,9 +710,10 @@ func (r *ProfileReconciler) PatchDefaultPluginSpec(ctx context.Context, profileI
 				},
 			})
 		}
-	}
-	if err := r.Update(ctx, profileIns); err != nil {
-		return err
+		if err := r.Update(ctx, profileIns); err != nil {
+			return err
+		}
+
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


Arhell (commented):
/ok-to-test

components/profile-controller/controllers/profile_controller.go
```diff
@@ -710,9 +710,10 @@ func (r *ProfileReconciler) PatchDefaultPluginSpec(ctx context.Context, profileI
 				},
 			})
 		}
-	}
-	if err := r.Update(ctx, profileIns); err != nil {
-		return err
+		if err := r.Update(ctx, profileIns); err != nil {
+			return err
+		}
+
```
umka1332, line 716: You should put it even highier - there is no point in updating the profile unless something is changed in this profile. And `profileIns` is changed only in the innermost if statement.

henrysecond1: Nice catch! Thank you

juliusvonkohout:

andreyvelich (commented):
The Kubeflow Profile code has been migrated to the new repo, please re-open PR at: https://github.com/kubeflow/dashboard
/close

Issue #6428, opened 2022-03-31
K8s 1.22 and Admission Webhook (PodDefaults)


kimwnasptd:
/kind feature

### Why you need this feature
Sub-issue of https://github.com/kubeflow/kubeflow/issues/6353

### Describe the solution you'd like
To have support for K8s 1.22 in the Admission Webhook (PodDefaults) we need to:
1. Make sure we are using the `v1` admission APIs https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/main.go#L31
2. Upgrade the PodDefault CRD to `v1`. We'll also need to have `preserveUnknownFields: false` in order to allow for upgradeability from `v1beta1` https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#field-pruning
    * Note that this should be way easier than https://github.com/kubeflow/kubeflow/issues/6366, since this time the CRD manifests were manually generated. So we can just do all the changes in-place
4. Update the `MutationWebhookConfiguration` to use the `v1` API

Relevant links:
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#webhook-resources-v122
https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

### Anything else you would like to add
One of the changes in the `MutationWebhookConfiguration` is that now by default the `failurePolicy` will be `Fail`. This means that if the webhook fails to mutate a Pod the whole creation will fail. We need to test this use case and make sure we document the behavior accordingly

wjentner:
@kimwnasptd I gave this issue a first shot, but I do have some questions:

1.  I assume the versions in the main.go should be updated as well? (see for example this line: https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/main.go#L61). I have not done that yet.
2. I've updated the PodDefault CRD and tested it with kubernetes v1.22.8
3. I've updated several of the MutationWebhookConfiguration manifests and defined the `failurePolicy` explicitly. Another issue are the `sideEffects`. I currently set them to `NoneOnDryRun` but perhaps just `None` is more fitting? The k8s docs also recommend to set `timeoutSeconds` explicitly, the default is `10` seconds but perhaps more is needed (up to 30 is possible)?

Should I issue a PR already, or should I continue fixing things beforehand?

aaron-arellano:
[...]

PR #6459, opened 2022-04-29
AIP-5879 Admission Webhook 1.22 Compatibility

aaron-arellano:
Updates the  admission webhook to be compatible with kubernetes v1.22 per issue #6428.


components/admission-webhook/main.go
```diff
@@ -484,7 +483,7 @@ func mutatePods(ar v1beta1.AdmissionReview) *v1beta1.AdmissionResponse {
 
 	crdclient := getCrdClient()
 	list := &settingsapi.PodDefaultList{}
-	err := crdclient.List(context.TODO(), &client.ListOptions{Namespace: pod.Namespace}, list)
+	err := crdclient.List(context.TODO(), list, &client.ListOptions{Namespace: pod.Namespace})
```
alexlatchford, line 486: Dumb question why was this reordering needed?

components/admission-webhook/manifests/base/crd.yaml
```diff
@@ -8,46 +8,62 @@ spec:
...
-      properties:
-        apiVersion:
-          type: string
-        kind:
-          type: string
-        metadata:
-          type: object
-        spec:
+  versions: 
+    - name: v1alpha1
```
alexlatchford, line 14: Maybe a dumb question but why is this one `v1alpha1`?

alexlatchford (approved):
Definitely not my area of expertise but looks great to me @aaron-arellano, thanks for tackling this!

components/admission-webhook/manifests/base/crd.yaml
aaron-arellano: cause the version of the CRD (poddefault) did not change, we only changed the k8s CRD version.

components/admission-webhook/main.go
aaron-arellano: the function signature changed. when moving to v1 the original ordering did not work. below is the signature.

https://github.com/kubernetes-sigs/controller-runtime/blob/3cb67228604239d3cd764b41719565bb4a46add6/pkg/client/client.go#L258

Arhell (commented):
/ok-to-test

kimwnasptd:
Thanks for the contribution @aaron-arellano and thank you for your time reviewing this @alexlatchford!

I'll also take a look on Tuesday to wrap the 1.22 effort up

aaron-arellano:
@kimwnasptd any updates on getting this review? Thanks!

aaron-arellano:
@kimwnasptd will this be getting a review soon? We are looking to merge the changes on our end ASAP once the contribution gets approved here. Thanks

kimwnasptd (commented):
These are the only questions I had. I'll also compile and apply these to a K8s 1.22 cluster to verify.

@aaron-arellano one last request, could you also update the `Makefile` to:
1. Use the `TAG ?= $(shell git describe --tags --always)`
2. Rename `build-gcr` to `docker-build`
3. Add one more rule `docker-push` that just pushes the image

components/admission-webhook/README.md
```diff
@@ -75,10 +75,19 @@ This specifies
 1. When there is a pod being created (see `rules`),
 1. call the webhook service `gcp-cred-webhook.default` at path `/add-cred` (see `clientConfig`)
 
+### admissionregistration.k8s.io/v1 default failurePolicy
+In adopting `admissionregistration.k8s.io/v1` for the `MutatingWebhookConfiguration` we accept the default value
+for `failurePolicy` to be `Fail` per [documentation](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy). Upon testing this default feature it was discovered if the `AdmissionWebhook`'s `mutating-webhook-configuration` failed to mutate a pod then the pod would fail to start, its associated `Deployment` or `StatefulSet` would continually attempt to create the pod until the process that created the pod was terminated. This continuous failure to mutate the target pod may block other target pods from mutation
+until the failing process ends.
```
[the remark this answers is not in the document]
aaron-arellano: yep. will clarify. Pretty much any pod that is targeted by the admission webhook for mutation is effected and will fail to start if there are configuration collisions or some other issue. Already running pods are not effected at all, again only pods being created after the policy change.

components/admission-webhook/manifests/base/crd.yaml
```diff
@@ -8,46 +8,62 @@ spec:
...
+                desc:
+                  type: string
+                serviceAccountName:
+                  type: string
+                automountServiceAccountToken:
+                  type: boolean
+                env:
+                  items:
+                    type: object
+                    x-kubernetes-preserve-unknown-fields: true
```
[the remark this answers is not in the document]
aaron-arellano: if I did not add `x-kubernetes-preserve-unknown-fields: true` I noticed the fields would get pruned in our pod-defaults, we would end up having empty pod-defaults though they were not prior to the upgrade. It seems the pruning mechanism recognizes `string`, `boolean` and `integer` no problem but did not recognize `array` or `object`. Once I added these tags the pod-defaults were generated as normal.

aaron-arellano:
@kimwnasptd should I also update the image url in the makefile to `public.ecr.aws/j1r0q0g6/`? I saw that was updated in the notebook controller v1.22 code change.

kimwnasptd:
Just tested this locally, as well as in a K8s 1.22 cluster.

Thanks for the work @aaron-arellano!

/lgtm
/approve

PR #6464, opened 2022-05-02
crud-web-apps: Migrate from tslint to eslint

haoxins:
Will submit another PR to apply the lint with `--fix`.
let's keep this PR focus on the migration only.


Arhell (commented):
/ok-to-test

kimwnasptd:
@haoxins how did you convert the `tslint.json` files to `.eslintrc.json`?

haoxins:
> @haoxins how did you convert the `tslint.json` files to `.eslintrc.json`?

use a tool, let me find the link

haoxins:
https://github.com/typescript-eslint/tslint-to-eslint-config#usage

> I was initially thinking of just generating an .eslintrc.js file based on the existing tslint.json

I think this tool has converted the exists `tslint` config.
And it seems that the codes inside this project are not all formatted by the `tslint` 😂

kimwnasptd:
> https://github.com/typescript-eslint/tslint-to-eslint-config#usage

>>Coming from Angular? Use [angular-eslint](https://github.com/angular-eslint/angular-eslint#migrating-an-angular-cli-project-from-codelyzer-and-tslint) instead to coordinate around this tool and others.

Nice finding! So it seems that `angular-eslint` is the recommended way. I'll also do a pass of the migration locally to see if we might need to do anything else. @haoxins did you need to do anything out of the ordinary while doing the migration yourself?

> And it seems that the codes inside this project are not all formatted by the tslint 😂

Hmm, that's a bug actually. Let's also introduce a GH Action in the next PR, in which we'll ensure all code follows the linting rules, to ensure the code will always be linted correctly

haoxins:
> did you need to do anything out of the ordinary while doing the migration yourself?

I did this months ago, not very sure. I think I planed to submit second PR that apply the new eslint rules after this PR merged.

kimwnasptd:
/lgtm
/approve

Thanks for the PR @haoxins, and apologies for the way too delayed review!

Please ping me on the follow up PR as well for linting the frontends as well as creating a GH workflow for testing. We can discuss about the workflow in the subsequent PR as well.

haoxins:
@kimwnasptd 
The first PR (`1/2`) for applying the `eslint` 
https://github.com/kubeflow/kubeflow/pull/6580/files

PR #6474, opened 2022-05-07
fix: unnecessary notebook updates caused by compare issues

mofanke:
fix https://github.com/kubeflow/notebooks/issues/106

### DIFF
--- a/components/common/reconcilehelper/util.go
+++ b/components/common/reconcilehelper/util.go
@@ -8,6 +8,8 @@ import (
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+
+	"k8s.io/apimachinery/pkg/api/equality"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/types"
@@ -125,7 +127,7 @@ func CopyStatefulSetFields(from, to *appsv1.StatefulSet) bool {
 		requireUpdate = true
 	}
 
-	if !reflect.DeepEqual(to.Spec.Template.Spec, from.Spec.Template.Spec) {
+	if !equality.Semantic.DeepDerivative(from.Spec.Template.Spec, to.Spec.Template.Spec) {
 		requireUpdate = true
 	}
 	to.Spec.Template.Spec = from.Spec.Template.Spec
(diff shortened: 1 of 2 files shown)


components/common/reconcilehelper/util.go
```diff
@@ -211,9 +213,9 @@ func CopyVirtualService(from, to *unstructured.Unstructured) bool {
 		return true
 	}
 
-	requiresUpdate := !reflect.DeepEqual(fromSpec, toSpec)
-	if requiresUpdate {
+	requireUpdate := !reflect.DeepEqual(fromSpec, toSpec)
```
MartinForReal, line 216: should we use !equality.Semantic.DeepDerivative instead?

mofanke: for VirtualService, there is  no default value which can afftect DeepEqual, so in my opinion there no need here

components/common/reconcilehelper/util.go
```diff
@@ -125,7 +127,7 @@ func CopyStatefulSetFields(from, to *appsv1.StatefulSet) bool {
 		requireUpdate = true
 	}
 
-	if !reflect.DeepEqual(to.Spec.Template.Spec, from.Spec.Template.Spec) {
+	if !equality.Semantic.DeepDerivative(from.Spec.Template.Spec, to.Spec.Template.Spec) {
```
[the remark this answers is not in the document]
mofanke: https://github.com/kubernetes/apimachinery/blob/57f2a0733447cfd41294477d833cce6580faaca3/third_party/forked/golang/reflect/deep_equal.go#L373， pod template in statefulset we got from api-server， imagepullpolicy and some default fileds were filled， so "reflect.DeepEqual" will always be True. 
Unset fields  such as  nil pointer and empty string are ignored (not compared) in "equality.Semantic.DeepDerivative", so  "equality.Semantic.DeepDerivative" will return False when all set fileds not changed .  By the way,  it will ingore changes in these unset fields , but i think it will be better than reflect.DeepEqual.   In another way，  we can pick some fileds to compare like https://github.com/kubeflow/kubeflow/blob/7106ef6340b7f07b06a0f531f3f7a484ebb5c343/components/tensorboard-controller/controllers/tensorboard_controller.go#L480

Arhell (commented):
/ok-to-test

juliusvonkohout:

andreyvelich:
Please re-open PR there.
/close

PR #6482, opened 2022-05-22
feat(Jupyter): Show group one or two only if images exists

haoxins:


Arhell (commented):
/ok-to-test

kimwnasptd:
That's actually a pretty good addition with a few lines of code!

I really believe we need to rework the notebook type part, both in the CRD as well as how we expose it.

/lgtm
/approve
/hold

waiting to resolve https://github.com/kubeflow/kubeflow/issues/6502 to be able to build an image for this as well

PR #6523, opened 2022-06-17
notebooks: Allow event patch in notebook controller

henrysecond1:
Notebook controller needs permission for event `patch` operation to emit events.

`
E0617 02:32:40.007467       1 event.go:191] Server rejected event.. events "new-notebook.16f946bea5deaaf5" is forbidden: User "system:serviceaccount:kubeflow:notebook-controller-service-account" cannot patch resource "events" in API group "" in the namespace "test"' (will not retry!)
`


kimwnasptd:
Thank you for the PR @henrysecond1!

We just merged https://github.com/kubeflow/kubeflow/pull/6524 for testing manifests as well. Can you rebase on top of the latest master to ensure that the action is triggered as expected?

henrysecond1:
@kimwnasptd Thank you for checking. I rebased it and the action seems to be triggered as expected

kimwnasptd:
Looks good, thanks!

/lgtm
/approve

Issue #6502, opened 2022-06-02
Migrate CI/CD to GH Actions


kimwnasptd:
Currently all AWS CI/CD tests have been disabled https://github.com/kubeflow/testing/pull/1003. This means that:
1. All tests we had are not running
2. We can't push images to the AWS registry, which was previously done by the CD jobs

AFAIK the AWS team is working on migrating the infra on a new billing account, so at some point we should be able to use the previous Prow workflows we had https://github.com/kubeflow/kubeflow/blob/master/prow_config.yaml.

Until then we'll need to move our CI/CD somewhere else and push our images in a different container registry. I propose that:
1. We use GH Actions for running some very basic tests, like checking formatting and manifests generation
2. Use a DockerHub registry to push images. I've created https://hub.docker.com/u/kubeflownotebookswg that we could use
    * Since DockerHub has rate limiting it's also worth trying pushing to [`ghcr.io`](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) as well 

cc @kubeflow/release-managers @kubeflow/release-team 

/assign @kimwnasptd 
/assign @nickloukas

surajkota:
Hi @kimwnasptd, Please see community meeting notes from 2022-05-31 and https://github.com/kubeflow/testing/issues/1006#issuecomment-1145523044

thesuperzapper:
@kimwnasptd rather than using a semi-random DockerHub, my vote is to just use GHCR as this will make our images be hosted under: `ghcr.io/kubeflow/{IMAGE_NAME}:{IMAGE_TAG}` which is much more trustworthy, and less susceptible to supply chain attacks (as DockerHub limits you to a single super-admin API key unless you pay them).

kimwnasptd:
@thesuperzapper let's actually go with both registries. We've raised in the past the proposal of pushing images to more than one registries https://github.com/kubeflow/kubeflow/issues/5922. If `ghcr.io` doesn't provide such limits as DockerHub we can use that as the default one. 

Will need to coordinate a little bit with @zijianjoy on the exact steps to add credentials to GH with the registry. I'll reach out once I have fully written down instructions, so that we allocate the least necessary amount of his time for setting this up

andreyvelich:
Currently, we use GitHub actions for CI/CD

PR #6524, opened 2022-06-17
Add GH actions for building manifests at any pull request

NickLoukas:
As part of migrating kubeflow CI/CD to GH actions (https://github.com/kubeflow/kubeflow/issues/6502) we introduce GH actions, one for each of these customizations (https://github.com/kubeflow/manifests/blob/master/example/kustomization.yaml#L34-L49), that at any pull request:
1. Download and install `kustomize==3.2` 
2. Assert the successful building of the manifests with `kustomize build`


.github/workflows/centraldb_manifests_build.yaml
```diff
@@ -0,0 +1,23 @@
+name: Build CentralDashboard manifests
+on:
+  pull_request:
+    paths:
+      - components/centraldashboard/**
```
kimwnasptd, line 5: Since we are testing manifests this time we don't want to trigger the workflows when code changes, but rather when the **manifests** change.

So in this case it will be `components/centraldashboard/manifests/**`

.github/workflows/jwa_manifests_build.yaml
```diff
@@ -0,0 +1,24 @@
+name: Build JWA manifests
+on:
+  pull_request:
+    paths:
+      - components/crud-web-apps/jupyter/**
+      - components/crud-web-apps/common/**
```
kimwnasptd, line 6: Here we don't even need to include the common code at all, since we care about the manifests.

So in this case it should only be:
`components/crud-web-apps/jupyter/manifests/**`

kimwnasptd (commented):
@NickLoukas the only proposed changes I had were about triggering these workflows on changes on the manifests files, and not on the code.

Lastly, let's also rename these workflows from `*-manifests-build` to `*-manifests-test`, since the purpose of these actions is to test that we can always build.

.github/workflows/centraldb_manifests_build.yaml
NickLoukas: I made the change.

.github/workflows/jwa_manifests_build.yaml
NickLoukas: I deleted the common code from the paths.

.github/workflows/prof_controller_manifests_test.yaml
```diff
@@ -0,0 +1,23 @@
+name: Build Profile Controller manifests
+on:
+  pull_request:
+    paths:
+      - components/profile-controller/**
```
kimwnasptd, line 5: I think you missed this one. It should be `components/profile-controller/config/**`

NickLoukas: Yeap you are correct. Fixed it.

kimwnasptd:
Looks good, thanks for the PR @NickLoukas!

PR #6529, opened 2022-06-22
Trim the custom Jupyter image string

haoxins:
If the custom image has extra white space, the status is always pending on the web page.

### DIFF
--- a/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-new/form-new.component.ts
+++ b/components/crud-web-apps/jupyter/frontend/src/app/pages/form/form-new/form-new.component.ts
@@ -99,7 +99,7 @@ export class FormNewComponent implements OnInit, OnDestroy {
 
     // Use the custom image instead
     if (notebook.customImageCheck) {
-      notebook.image = notebook.customImage;
+      notebook.image = notebook.customImage?.trim();
     } else if (notebook.serverType === 'group-one') {
       // Set notebook image from imageGroupOne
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
Fixes https://github.com/kubeflow/kubeflow/issues/6621

kimwnasptd:
@haoxins could you rebase on top of the latest changes to resolve the conflicts?

haoxins:
done

neggert:
@kimwnasptd Any movement on this? Our users run into this issue fairly regularly and it's not easy for someone who doesn't know Kubernetes to debug.

thesuperzapper:
@kimwnasptd I think trimming whitespace is probably a good idea, as it is a pretty common error when copying.

@haoxins the one thing I want to confirm is that this PR doesn't create a null image value when the user doesn't specify a custom image.

Also, I was looking into other validation strategies with regex, but it's honestly very difficult to make one that is generic enough. Not even Kubernetes/docker tries to validate your image string, so we probably shouldn't either.

neggert:
It's worth noting that there's a [related issue](https://github.com/kubeflow/kubeflow/issues/6886). If the string is left blank (or after this PR, all whitespace), it breaks the notebook dashboard entirely.

thesuperzapper:
@neggert are you able/interested in raising a PR that does one or both of the things proposed in https://github.com/kubeflow/kubeflow/issues/6886#issuecomment-1851137333?

PR #6530, opened 2022-06-22
Add GH actions for applying manifests at any pull request

NickLoukas:
For applying the manifests in a K8s cluster we use KinD. The process that we follow is:
1. Install KinD and configure a cluster with a specific K8s version
2. Install `kustomize`
3. Install `Istio`
4. Build and apply the manifests

### Notes

* For the PodDefaults component we also need to install `cert-manager` in order to apply the manifests. We follow the approach from `kserve` (https://github.com/kserve/kserve/blob/7e74dca4421af26cfbe96dccdf3806e77a63c1b3/test/scripts/gh-actions/setup-deps.sh#L86-L92) where we install `cert-manager` and then wait for the `webhook` to become ready.


kimwnasptd:
Thanks for the PR @NickLoukas! As a first comment I see a couple of repeatable steps for installing KinD and setting up a cluster. 

Let's convert these into script files so that we reduce the duplication and use the same logic everywhere. I think the best place for such scripts will be https://github.com/kubeflow/kubeflow/tree/master/testing. Let me create an initial PR to clean up that dir

kimwnasptd:
@NickLoukas seems like there's a lot of delay in merging https://github.com/kubeflow/kubeflow/pull/6535.

Let's unblock from that PR from now by creating a `components/testing/gh-actions` directory and put there any scripts we will use for testing. We can then move them under `testing/gh-actions` once/if https://github.com/kubeflow/kubeflow/pull/6535 gets merged.

In that folder we can put:
1. A script for fetching the KinD executable
2. The `config.yaml` for the KinD cluster
3. A script for fetching the kustomize executable
4. A script for installing Istio

NickLoukas:
@kimwnasptd I sent some follow up commits where I created the aforementioned scripts in `components/testing/gh-actions` and modified all GH actions to make use of these scripts.

components/testing/gh-actions/install_istio.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+ISTIO_VERSION="1.11.0"
```
kimwnasptd, line 2: let's also put a `set -e` to ensure the script will error in case on of the subcommands return an exit code other than zero https://linuxhint.com/bash-set-e/

components/testing/gh-actions/install_istio.sh
```diff
@@ -0,0 +1,10 @@
...
+echo "Installing Istio ..."
```
kimwnasptd, line 3: same for `set -e`

components/testing/gh-actions/install_kustomize.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
kimwnasptd, line 1: same for `set -e`

.github/workflows/prof_controller_kind_test.yaml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Install KinD
+      run: |
+        chmod +x components/testing/gh-actions/install_kind.sh
```
kimwnasptd, line 16: let's make these script files executables and commit them to git. Git can track the file permissions, so we don't have to `chmod +x` on each GH Action

kimwnasptd (commented):
Last round of comments. Don't think we'll need anything extra afterwards

components/testing/gh-actions/install_istio.sh
NickLoukas: Added `set -e`.

components/testing/gh-actions/install_istio.sh
NickLoukas: Did that.

components/testing/gh-actions/install_kind.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
NickLoukas: Did that.

components/testing/gh-actions/install_kustomize.sh
NickLoukas: Did that.

.github/workflows/prof_controller_kind_test.yaml
NickLoukas: I made all the scripts executable.

NickLoukas:
@kimwnasptd I addressed the comments above with 2 new commits.

PR #6531, opened 2022-06-22
proposals: Show objects from all namespaces

kimwnasptd:
Extend the CentralDashboard and Jupyter/TensorBoards/Volumes web apps for showing all namespaces. This allows users to view and interact with objects in all namespaces.

Initially we'll start with only the above list of apps, but we can then extend it to the rest of the Kubeflow web apps.

cc @kubeflow/wg-notebooks-leads @johnugeorge @zijianjoy


kimwnasptd:
I believe we can go on with this proposal, since we didn't have any extra review comments

cc @elikatsis @StefanoFioravanzo @thesuperzapper

StefanoFioravanzo:
@kimwnasptd I agree! Let's make this happen :) 

/lgtm
/approve

PR #6535, opened 2022-06-23
testing: Cleanup the testing dir and add gh actions placeholder folder

kimwnasptd:
Spined up from https://github.com/kubeflow/kubeflow/pull/6530#issuecomment-1164125447

We'll need to have some scripts for tests we run via the GH actions. I thought that it could be a good chance to cleanup the `/testing` folder, which had completely out of date files. 

I only added an owners file under `/testing/gh-actions` since this is what we need for now. Was not sure if the conformance testing would need any files under `/testing`, that's why I tried to be explicit with the `gh-actions` folder.

@james-jwu @zijianjoy could you take a look?


kimwnasptd:
kind ping @james-jwu @zijianjoy. We need this one as well to ensure we have the proper testing for the upcoming feature freeze.

zijianjoy:
/lgtm
/approve

Just noticed this PR, thank you Kimonas for the cleanup!

PR #6537, opened 2022-06-23
Add GH action for publishing all images when VERSION is updated

NickLoukas:
Add a GH action that runs when `VERSION` file is bumped. The workflow is:
1. Gets triggered on `on.push` for the `master` and `v*-branch` branches
2. Gets triggered only when the `releasing/version/VERSION` file is changed
3. Reads the content of the VERSION file and:
    1. Builds all images (we'll need a step for each component)
    2. Pushes the images of each component

### DIFF
--- a/.github/workflows/all_components_docker_publish.yaml
+++ b/.github/workflows/all_components_docker_publish.yaml
@@ -1,0 +1,145 @@
+name: Build & Publish all Docker images
+on:
+  push:
+    branches:
+      - master
+      - v*-branch
+    paths:
+      - releasing/version/VERSION
+
+jobs:
+  push_to_registry:
+    name: Build & Push all Docker images to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Login to DockerHub
+      uses: docker/login-action@v2
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/all_components_docker_publish.yaml
```diff
@@ -0,0 +1,145 @@
...
+        cd components/tensorboard-controller
+        export IMG=docker.io/kubeflownotebookswg/tensorboard-controller:${TAG}
+        make docker-build
+        make docker-push
+
+    - name: Run TWA build and push
+      run: |
+        export GIT_TAG=$(cat releasing/version/VERSION)
+        cd components/volumes-web-app
+        export IMG=docker.io/kubeflownotebookswg/volumes-web-app
```
kimwnasptd, line 137: This one and the above need to be tensorboards, also the path is wrong

kimwnasptd (commented):
I think we are ready to merge, aside from these small comments

.github/workflows/all_components_docker_publish.yaml
NickLoukas: Fixed that.

NickLoukas:
@kimwnasptd I addressed the review comments above.

kimwnasptd:
Thanks! 

/lgtm
/approve

PR #6538, opened 2022-06-24
Add '--reload_multifile' argument to tensorboard

AppleHolic:
- Problem
If the log directory has several event files, tensorboard could not load all logs immediately. 

- Solution
The "--reload_multifile" argument can be used to poll all "active" files in a directory for new data, rather than the most recent one as mentioned at [related docs of tensorboard repository](https://github.com/tensorflow/tensorboard/blob/master/README.md#tensorboard-is-showing-only-some-of-my-data-or-isnt-properly-updating). All of the logs are loaded well after applying that argument.


juliusvonkohout:

andreyvelich:
Please re-open PR there.
/close

PR #6542, opened 2022-06-24
Add support for `command` and `args` in `PodDefault`

### LINKED ISSUE #2208: Make arbitrary Jupyter images work with Kubeflow

### LINKED ISSUE #3838: jupyter-web-app: Set custom Image cmd and args
/kind feature

Currently if someone would like to use a custom image they would need to create a Dockerfile, update the ENTRYPOINT and then push it to a public/private registry. We could skip the push step by allowing the user to set the cmd/args of their image.

This has also been requested from user's both in slack and from issues [ #2961 #3821 ]

The user should be able to set a custom `CMD` and/or `ARGS` for the Notebook Image.

alembiewski:
#### What changes were proposed in this pull request?
Closes #2208, closes #3838.
This PR extends `PodDefault` to provide a capability to set a custom `command` and `args` for Notebook servers.

#### Example
- Create the following PodDefaut in the user's namespace:
```
kind: PodDefault
metadata:
  name: custom-entrypoint
spec:
  selector:
    matchLabels:
      custom-entrypoint: "true"
  desc: "Lauch as Jupyter Lab"
  command:
  - jupyter
  args:
  - lab
  - --notebook-dir=/workspace
  - --ip=0.0.0.0
  - --no-browser
  - --allow-root
  - --port=8888
  - --NotebookApp.token=''
  - --NotebookApp.password=''
  - --NotebookApp.allow_origin='*'
  - --NotebookApp.base_url=$(NB_PREFIX)
```
- In notebook form, select "Custom image" and set the image you want to run (e.g. nvcr.io/nvidia/tensorflow:22.05-tf1-py3.)
- Under configuration, select "Launch as Jupyter Lab" and click "Launch"
Notebook server pod should be create with the specified command and arguments.

### DIFF
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -41,7 +41,8 @@ import (
 )
 
 const (
-	annotationPrefix = "poddefault.admission.kubeflow.org"
+	annotationPrefix        = "poddefault.admission.kubeflow.org"
+	istioProxyContainerName = "istio-proxy"
 )
 
 // Config contains the server (the webhook) cert and key.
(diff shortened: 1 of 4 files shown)


supertetelman:
This is great, I've been looking for this built-in functionality for years.

Is there any chance we can include that PodDefault  by default out of the box so this feature could be used for any Kubeflow user to run any container that has Jupyter Lab installed without any custom K8s work?

discordianfish:
/approve

Dunno what the general direction around PodDefaults are (I remember talk about them getting deprecated?) but for now, LGTM!

PR #6543, opened 2022-06-24
Make GH actions build images on PR and push only on merge

NickLoukas:
In existing GH actions for building and pushing Docker images we separate the build and push steps in order to run build & push only when a PR is merged. In any other case, when a PR is opened or re-opened we only build the images.


kimwnasptd:
Thanks @NickLoukas!

/lgtm
/approve

PR #6547, opened 2022-06-29
Reword image name in TWA Gh action

NickLoukas:
When merging the initial PR for building the images we had a bug in some image names we used in the GH actions. This is a fix for the TWA image name.


kimwnasptd:
Thank you @NickLoukas!

/lgtm
/approve

PR #6548, opened 2022-06-29
release: Use DockerHub for all images

kimwnasptd:
I'm making the PR for the `v1.6-branch` initially and we can cherry-pick to `master` later on.

This PR:
1. Updates images to use the DockerHub registry `docker.io/kubeflownotebookswg`
2. Updates the `releasing/version/VERSION` file to build all images with the tag `v1.6.0-alpha`, to ensure all images can be build

/git cc @elikatsis @StefanoFioravanzo @thesuperzapper


kimwnasptd:
It seems that the above PR failed for 2 reasons:
1. The notebook-server dockerfiles are using base images that lived in ECR. So after renaming everything to use dockerhub the current base images are missing and that's why the build fails
2. The GH Action for building the volumes-web-app has a bug and checks out an incorrect path.

kimwnasptd:
> 1. The notebook-server dockerfiles are using base images that lived in ECR. So after renaming everything to use dockerhub the current base images are missing and that's why the build fails

@apo-ger had worked on updating all the Makefiles and also revamped the build process of the notebook-servers. But since we are tight on time I'll just revert the dockerhub change in the Dockerfiles of the notebooks, for the build process to success.

But to fix this properly we must rework on the problematic build process of the notebook servers. @apo-ger let's do this after the RC0 is cut

> 2. The GH Action for building the volumes-web-app has a bug and checks out an incorrect path.

@NickLoukas has a PR for this in flight https://github.com/kubeflow/kubeflow/pull/6549

kimwnasptd:
@elikatsis @StefanoFioravanzo @thesuperzapper this should be ready to merge. All tests pass, images can be build and manifests get applied in a K8s 1.22 cluster successfully

PR #6549, opened 2022-06-29
Fix VWA path in Gh action for building image

NickLoukas:
The GH Action for building VWA had a bug and didn't use the correct dir.


kimwnasptd:
Thanks for the quick fix! 

/lgtm
/approve

kimwnasptd:
It looks like `ctrl + enter` don't hit the `Comment` button but the `Close with comment` by default. The more you know...

/lgtm
/approve

PR #6558, opened 2022-07-01
Avoid reloading spawner_ui_config yaml files.

xiaoxq:
The `defaults` is exactly the spawner_ui_config:
https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/backend/apps/default/routes/post.py#L25

Just reuse it and avoid reloading yaml files which is inefficient.


components/crud-web-apps/jupyter/backend/apps/common/form.py
```diff
@@ -124,7 +124,7 @@ def set_notebook_cpu(notebook, body, defaults):
     if cpu_limit and 'nan' in cpu_limit.lower():
         raise BadRequest("Invalid value for cpu limit: %s" % cpu_limit)
 
-    limit_factor = utils.load_spawner_ui_config()["cpu"].get("limitFactor")
+    limit_factor = defaults["cpu"].get("limitFactor")
```
mvoitko, line 127: Which version of Python is used? From Python 3.8 we could use walrus. It seems like an excellent place for it here.

xiaoxq: It's still 3.7: https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps/jupyter#development
Would love to see a bump to 3.8 :)

xiaoxq:
Rebased to latest HEAD to include the Python 3.8 bump https://github.com/kubeflow/kubeflow/pull/6562 ,
so that we can leverage walrus operator to simplify code as mentioned by @mvoitko.

PR #6561, opened 2022-07-05
Fix legacy of Python version information in README files.

xiaoxq:
The code has upgraded to Python 3.8 while the doc keeps legacy information.
https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/tensorboards/Dockerfile#L2
https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/volumes/Dockerfile#L2


kimwnasptd:
Nice catch! 

/lgtm
/approve

PR #6564, opened 2022-07-06
fix: double URL encoding in central-dashboard links

### LINKED ISSUE #6563: Kubeflow centraldashboard double escapes characters in iframe links
/kind bug

I've embedded mlflow in the kubeflow central dashboard. Certain links, like metrics, include quotation characters. When I click the link, the resulting location gets redirected with the quotes mangled and double escaped, first to %22, then to %2522. The nonsense URL is unparseable and causes mlflow to break.

I expect characters in URLs to not be double escaped

The problem fundamentally stems from iron-location's assumptions that it's hash property is already decoded. In the defined getter, we see that it takes care to decode it:

https://github.com/PolymerElements/iron-location/blob/8bdb632adbdf5de8d16e33ef55d1c732ea373750/iron-location.js#L89

However, in this kubeflow function `_iframePageChanged`, we pull the window location raw without decoding it:

https://github.com/kubeflow/kubeflow/blob/9b6076d7b11bbae2a5de1eeb3eee7d36741f23aa/components/centraldashboard/public/components/main-page.js#L348

This value eventually gets bound to App-Location:

https://github.com/kubeflow/kubeflow/blob/9b6076d7b11bbae2a5de1eeb3eee7d36741f23aa/components/centraldashboard/public/components/main-page.pug#L13

Which consequently gets bound to Iron-Location:

https://github.com/PolymerElements/app-route/blob/1a5e93ead6c7f0cfb1971a1abc322a49a649b687/app-location.js#L77

In the end, we get an encoded `hash` value in iron-location, and then iron-location re-encodes it, again, leading to the double-escaped URL character:

https://github.com/PolymerElements/iron-location/blob/8bdb632adbdf5de8d16e33ef55d1c732ea373750/iron-location.js#L223

To fix this, we need to take care to not leak in encoded url values to Polymer App-Location / Iron-Location Components. Putting a `window.decodeURIComponent` around the hash location extraction in `_iframePageChanged` should resolve the issue.

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 1.25.1
- kfctl version: (use `kfctl version`): ?
- Kubernetes platform: kind
- Kubernetes version: (use `kubectl version`): 1.23
- OS (e.g. from `/etc/os-release`): Ubuntu 22.04

timcharper:
Central dashboard would lead to double URL encoding characters such as quotation due to the expectation that the
underlying Iron-Location hash is already decoded, and failure to decode it when directly setting it from an iframe URL
change.

Fixes kubeflow/dashboard#34

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -345,7 +345,8 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
             }
         }
         window.history.replaceState(null, null, l.toString());
-        this.set('routeHash.path', window.location.hash.substr(1));
+        this.set('routeHash.path',
+            window.decodeURIComponent(window.location.hash.substr(1)));
     }
 
     /**
--- a/components/centraldashboard/public/components/main-page_test.js
+++ b/components/centraldashboard/public/components/main-page_test.js
@@ -270,13 +270,14 @@ describe('Main Page', () => {
         const hash = '#/hash/route/fragments';
         window.location.hash = hash;
         mainPage.set('queryParams.ns', 'test');
-        mainPage.set('queryParams.foo', 'bar');
+        mainPage.set('queryParams.foo', '["bar"]');
         mainPage.subRouteData.path = '/pipeline/';
         mainPage._routePageChanged('_');
         flush();
 
-        expect(mainPage.iframeSrc).toMatch(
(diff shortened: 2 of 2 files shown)


juliusvonkohout:

andreyvelich (commented):
/close

PR #6571, opened 2022-07-12
Create separate workflows for each notebook server

NickLoukas:
Until now we had one workflow that builds all the notebooks. We now add a different workflow for each notebook server and only for the ones that appear in JWA's Config Map: https://github.com/kubeflow/kubeflow/blob/bb9fb467c85d495b2af12dfc26e71f17f9ec857c/components/crud-web-apps/jupyter/manifests/base/configs/spawner_ui_config.yaml#L17-L48

### DIFF
--- a/.github/workflows/nb_server_codeserver_python_docker_publish.yaml
+++ b/.github/workflows/nb_server_codeserver_python_docker_publish.yaml
@@ -6,12 +6,16 @@ on:
       - v*-branch
     paths:
       - components/example-notebook-servers/codeserver-python/**
+      - components/example-notebook-servers/codeserver/**
+      - components/example-notebook-servers/base/**
   pull_request:
     branches:
       - master
       - v*-branch
     paths:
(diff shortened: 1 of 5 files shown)


kimwnasptd:
@NickLoukas we'll need to ensure that the workflows will be triggered even when the files of the base images change.

For example `jupyter-pytorch-full` depends on the `pytorch`, `jupyter` and `base` images. So we'll need to trigger the workflow for `jupyter-pytorch-full` when any of the above images changes.

Can you do a pass in the action and take these dependencies into consideration?
https://github.com/kubeflow/kubeflow/tree/master/components/example-notebook-servers#images--relationship-chart

NickLoukas:
@kimwnasptd I added the paths of the base images in all the workflows.

NickLoukas:
@kimwnasptd For the cuda images of `pytorch` and `tensorflow` we're building them together with the non-cuda images since we're checking for changes in the same folder. So we build 7 images in total in our 5 workflows.

PR #6578, opened 2022-07-15
Use controller-gen for PodDefaults

kimwnasptd:
Right now we rely on manual changes to:
* deepcopy files
* the crd

We should instead rely on `controller-gen` for automating this task, just like we do for the rest of the controllers

/cc @discordianfish @alembiewski


kimwnasptd:
Putting this in WIP state until I've verified the manifests can also be applied when we have the previous ones in a cluster

kimwnasptd:
Verified that:
1. The CRD can be applied in the cluster with the previous CRD present, so no special upgrade instructions are needed
2. The controller handles the deepcopy code as expected [ PodDefaults can be mounted as expected ]

@discordianfish @alembiewski this is ready for review

discordianfish:
Thanks!

/approve

kimwnasptd:
I think it also needs the lgtm label

PR #6580, opened 2022-07-19
crud-web-apps: Apply ng lint

haoxins:
#6464 

The `1/2` process of applying `ng lint` or `eslint`

### DIFF
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/.eslintrc.json
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/.eslintrc.json
@@ -27,6 +27,9 @@
             "accessibility": "explicit"
           }
         ],
+        "@typescript-eslint/naming-convention": "off",
+        "@typescript-eslint/member-ordering": "off",
+        "prefer-arrow/prefer-arrow-functions": "off",
         "arrow-parens": [
           "off",
           "always"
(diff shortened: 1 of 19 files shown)


kimwnasptd:
Thanks @haoxins! Let's also add a commit to introduce a `frontend_lint.yaml` that:
1. Will be triggered if either of the frontend of the three web apps or the common frontend code changes
2. Will run the `npm run lint` command of these 4 frontend projects

It's OK to not create a workflow for each project. We can just have one action that tests all frontends at the same time, all the time

haoxins:
> Let's also add a commit to introduce a frontend_lint.yaml that:

We can do this at another PR I will submit after this been merged.
Still need to fix some lint error that can't fix by `--fix`
And maybe or not break something, that's why I split into two PRs.

thesuperzapper:
> Thanks @haoxins! Let's also add a commit to introduce a `frontend_lint.yaml` that:
> 
> 1. Will be triggered if either of the frontend of the three web apps or the common frontend code changes
> 2. Will run the `npm run lint` command of these 4 frontend projects
> 
> It's OK to not create a workflow for each project. We can just have one action that tests all frontends at the same time, all the time

@kimwnasptd I assume you mean we add a PR check that fails if linting would change the code?

haoxins:
https://github.com/kubeflow/kubeflow/pull/6586
I think all the lint rules are pass now~

PR #6581, opened 2022-07-19
Reload certificates in admission-webhook

### LINKED ISSUE #5560: Admission webhook doesn't pick up new cert when it gets renewed by cert-manager
/kind bug

Admission webhook keeps throwing the below error every once in a while:
http: TLS handshake error from *****:34006: remote error: tls: bad certificate
This causes pod defaults not to work. When I restart the admission webhook, it works again

The above shouldn't happen.

So I got to the bottom of this. The problem is that admission webhook loads the cert files only once when the service starts. See these lines:
https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/main.go#L599
https://github.com/kubeflow/kubeflow/blob/master/components/admission-webhook/config.go#L41

Our cert gets renewed every 30 days by cert-manager. So when the cert-manager renew the certificate, the admission webhook has no mechanism to to pick up the new cert and keep using the old (now invalidated) cert. Once it gets restarted it picks up the correct cert and things works again!

I think the solution should be that the admission-webhook periodically update its reference of the certificate

**Environment:**

- Kubeflow version: v1beta1
- kfctl version: N/A
- Kubernetes platform: EKS
- Kubernetes version: 
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.3", GitCommit:"1e11e4a2108024935ecfcb2912226cedeafd99df", GitTreeState:"clean", BuildDate:"2020-10-14T18:49:28Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"darwin/amd64"}

Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.15-eks-ad4801", GitCommit:"ad4801fd44fe0f125c8d13f1b1d4827e8884476d", GitTreeState:"clean", BuildDate:"2020-10-20T23:27:12Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}

- OS (e.g. from `/etc/os-release`): Catalina 10.15.7

manolis-andr:
Reload server certificates if they are changed on disk. Use certwatcher
from controller-runtime that already implements this functionality.

Closes #5560.

### DIFF
--- a/components/admission-webhook/config.go
+++ b/components/admission-webhook/config.go
@@ -17,11 +17,13 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/tls"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"k8s.io/klog"
+	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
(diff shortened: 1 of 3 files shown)


kimwnasptd:
Thanks!

/lgtm
/approve

PR #6586, opened 2022-07-21
Fix ng lint

haoxins:
All lint should be pass now~


haoxins:
@kimwnasptd https://github.com/kubeflow/kubeflow/runs/7448855347?check_suite_focus=true
the workflow is passed

kimwnasptd:
Woo Hoo 🚀 
Thanks @haoxins!

/lgtm
/approve

PR #6587, opened 2022-07-21
crud-web-apps: upgrade into Node.js v16 (LTS)

haoxins:
There are some packages need to be updated for newer Node.js.
So, this is the first step to upgrade the crud-web-apps.
After this, we can upgrade to Angular v14.1 and update the other dependencies.


haoxins:
cc @kimwnasptd

PR #6591, opened 2022-07-22
Extend Docker Publish workflows to be triggered on Version change

NickLoukas:
### The problem
We had a problem with trying to build all images in one workflow, when the `releasing/version/VERSION` file changes. So we wanted to extend all components' workflows to ensure that if `releasing/version/VERSION` is changed then
1. We execute all the build workflows of each component
2. We have a conditional step in the existing workflows to push the image with the tag in `releasing/version/VERSION` if that file changes

### Our Contribution
In this PR we address the points above by:
1. Add the path `releasing/version/VERSION` in the paths of the push event to also trigger the workflow when that file changes
2. Use the `dorny/paths-filter@v2 action` to monitor the path `releasing/version/VERSION` and if it changes run a step that pushes the image with the tag `TAG=$(cat releasing/version/VERSION)`
3. Run the original push step only if the above push step is `skipped` (the documentation for the contexts of the steps can be found here: https://docs.github.com/en/actions/learn-github-actions/contexts#steps-context)

### Notes

In this PR we also include a minor typo fix in all Notebook Servers workflows for the `REGISTRY` environment variable to export it properly.


.github/workflows/centraldb_docker_publish.yaml
```diff
@@ -34,8 +42,17 @@ jobs:
...
+      if: steps.filter.outputs.version == 'true'
+      run: |
+        export TAG=$(cat releasing/version/VERSION)
+        cd components/centraldashboard
+        export IMG=kubeflownotebookswg/centraldashboard
+        make docker-push
+
     - name: Run CentralDashboard push
-      if: github.event_name == 'push'
+      if: github.event_name == 'push' && steps.version.conclusion == 'skipped'
```
kimwnasptd, line 55: Does this mean that this step will NOT be run if the `releasing/version/VERSION` file is not changed?

Let's still run that step, even if we pushed the image with a specific tag above. We still want to push the images with the tag being the git commit that got merged

kimwnasptd:
Lastly, @NickLoukas can you 
1. target this PR for the `v1.6-branch`, and then send another one for the `master` branch? I want to unblock the release and first include this in the release branch
2. Update the `releasing/version/VERSION` to have the value the last commit in `v1.6-branch`, to test if the images will be built

Then, once we merge this PR, we'll need a subsequent one to set `v1.6.0-rc.1` in the `releasing/version/VERSION` file

.github/workflows/centraldb_docker_publish.yaml
NickLoukas: The above means that this step only runs if the `version` push step is skipped. So yes if the `releasing/version/VERSION` file is changed this step will not run.

NickLoukas: I added a commit and removed the `&& steps.version.conclusion == 'skipped'` part from all the workflows.

NickLoukas:
@kimwnasptd I changed the target branch and the contents of the version file. I also addressed the other review comment.

kimwnasptd:
Should have taken a more detailed look. There's an issue with this PR because the new step does not actually retag or re-build the images to end up with an image with the `release/version/VERSION` tag. And thus the push can't find the image the new step tries to push.

@NickLoukas can you send a follow up PR in `v1.6-branch` for this?

kimwnasptd:
And also, let's completely remove the existing workflow that builds/push all images

.github/workflows/nb_server_jupyter_scipy_docker_publish.yaml
```diff
@@ -0,0 +1,64 @@
+name: Build & Publish Notebook Server Docker images
```
shalberd, line 1: @NickLoukas is there a reason there is no "Jupyter SciPy" in the name here? I mean, in the titles of all other nb server docker publish workflows, the target architecture is mentioned.

PR #6594, opened 2022-07-25
AIP-6075: Add terminal activity to notebook controller culler

aaron-arellano:
Adds `last_activity` of the jupyter server `api/terminals` endpoint. 

Our team needed the ability to also take into account terminal activity for culling, current culler logic only takes into account kernel activity. We are checking to see if the kubeflow community would like this contribution to be added to open-source culler logic.

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -263,7 +263,7 @@ func (r *NotebookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	if !podFound {
 		// Delete LAST_ACTIVITY_ANNOTATION annotations for CR objects
 		// that do not have a pod.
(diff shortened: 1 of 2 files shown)


guimou:
@aaron-arellano, are you sure this works? In culler.go:
- L370, you get `arr[i] = terminals[i].LastActivity`
- then at L373, you call `getNotebookRecentTime(arr, "api/terminals")`
However `arr` always has null values, because the terminal object does not have a LastActivity property, only a name. Ref: https://jupyterlab.readthedocs.io/en/3.4.x/api/interfaces/services.terminalapi.imodel.html
Am I missing something here?

aaron-arellano:
@guimou , interesting. I presume it is because your base image is jupyter lab whereas we use jupyter server. Surprised anyone looked at this considering how old this is. Gonna close this since it is too specific to my teams use case.

PR #6597, opened 2022-07-26
Fix workflows for building all images when Version file changes

NickLoukas:
This is a follow up PR to this https://github.com/kubeflow/kubeflow/pull/6591 in order to also build the image with the correct tag when the Version file changes.


kimwnasptd:
Looks good. The explicit `make docker-build` rule should re-use the cached layers and tag the image as expected.

/lgtm
/approve

PR #6598, opened 2022-07-26
Update Version file to tag 'v1.6.0-rc.1'

NickLoukas:
Update tag in Version file after fixing all github actions for building the images when the Version file changes (https://github.com/kubeflow/kubeflow/pull/6597).


kimwnasptd:
/lgtm
/approve

@annajung after this PR builds all images we should be ready to go with the manifests PR

Issue #6572, opened 2022-07-13
Cannot create notebook through web UI


rangerbottle:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
1. Install Kubeflow on local-cluster ( kind )
2. go to web UI and login with default user http://localhost:8080/_/jupyter/new?ns=kubeflow-user-example-com
3. Go to notebook panel and create a default one
4. After click "Launch", got error "[500] An error occured in the backend. http://localhost:8080/jupyter/api/namespaces/kubeflow-user-example-com/notebooks"
5. After check the pod "jupyter-web-app-deployment" log, found 

2022-07-13 09:20:24,419 | apps.common.form | INFO | Using provided value for 'workspace': {'mount': '/home/jovyan', 'newPvc': {'metadata': {'name': 'notebook1-volume'}, 'spec': {'accessModes': ['ReadWriteOnce'], 'resources': {'requests': {'storage': '10Gi'}}}}}
2022-07-13 09:20:24,419 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | Caught and unhandled Exception!
2022-07-13 09:20:24,419 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | create_notebook() got an unexpected keyword argument 'dry_run'
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1523, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1509, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/local/lib/python3.7/site-packages/kubeflow/kubeflow/crud_backend/decorators.py", line 18, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/kubeflow/kubeflow/crud_backend/decorators.py", line 40, in runner
    return func(*args, **kwargs)
  File "/src/apps/default/routes/post.py", line 48, in post_pvc
    api.create_notebook(notebook, namespace, dry_run=True)
TypeError: create_notebook() got an unexpected keyword argument 'dry_run'

**What did you expect to happen:**
Create notebook server

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

[...]

PR #6599, opened 2022-07-27
Fix(Jupyter): fix the image group one and two enabled props

haoxins:


kimwnasptd:
@haoxins instead of introducing new controls, what about using the `disabled` property of controls? 

If the ConfigMap does not have these sections then we disable these controls and use this check in the html files

annajung:
Hi @kubeflow/wg-notebooks-leads, can you review this as this might fix the [release-blocking issue for 1.6](https://github.com/kubeflow/kubeflow/issues/6572#issuecomment-1194513442)?

haoxins:
> @haoxins instead of introducing new controls, what about using the `disabled` property of controls?
> 
> If the ConfigMap does not have these sections then we disable these controls and use this check in the html files

@kimwnasptd I'm a bit busy these days, so that not have time to improve the implementation here.
can we go first and improve later if this works.

haoxins:
closed by https://github.com/kubeflow/kubeflow/pull/6625

PR #6603, opened 2022-07-28
notebooks: Fix notebook endless restarts (#6337)

mofanke:
Fix notebook endless restarts (#6337)

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -126,6 +126,14 @@ func (r *NotebookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, ignoreNotFound(err)
 	}
 
+	// jupyter-web-app deletes objects using foreground deletion policy, Notebook CR will stay until all owned objects are deleted
+	// reconcile loop might keep on trying to recreate the resources that the API server tries to delete.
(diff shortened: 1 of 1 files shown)


components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -126,6 +126,11 @@ func (r *NotebookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, ignoreNotFound(err)
 	}
 
+	//if notebook is terminating , no need to reconfile
```
kimwnasptd, line 129: Thanks for the PR @mofanke!

Small nit: Could you add some more details to this message as to why we don't need to reconcile? (Also I think there's a typo).

I think the important part to note here is the fact that the jupyter web app is deleting objects with [foreground deletion policy](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/api/notebook.py#L43), that results in the Notebook CR to stay until all owned objects are deleted. So if we don't stop the reconciliation the controller might keep on trying to recreate the resources that the API server tries to delete

mofanke: sure , thx for tips

kimwnasptd:
Thanks for the fix!

/lgtm
/approve

PR #6609, opened 2022-08-08
Fix the logout button to work with the recent version of `oidc-authservice`

alembiewski:
Required as a part of https://github.com/kubeflow/manifests/pull/2150

Fixes the logout button in the Central Dashboard to work with the recent versions of the `oidc-authservice` (tested with `gcr.io/arrikto/kubeflow/oidc-authservice:e236439` image version)

Changes:
* Pin alpine repository version
* Introduce a new LogoutButton component

### DIFF
--- a/components/centraldashboard/Dockerfile
+++ b/components/centraldashboard/Dockerfile
@@ -8,7 +8,7 @@ ENV BUILD_COMMIT=$commit
 ENV CHROME_BIN=/usr/bin/chromium
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
 
-RUN apt update -qq && apt install -qq -y chromium gnulib 
+RUN apt update -qq && apt install -qq -y chromium gnulib
 
 COPY . /centraldashboard
 WORKDIR /centraldashboard
@@ -16,7 +16,7 @@ WORKDIR /centraldashboard
 RUN BUILDARCH="$(dpkg --print-architecture)" &&  npm rebuild && \
     if [ "$BUILDARCH" = "arm64" ]  ||  \
     [ "$BUILDARCH" = "armhf" ]; then \
-    export CFLAGS=-Wno-error && \ 
+    export CFLAGS=-Wno-error && \
     export CXXFLAGS=-Wno-error;  \
     fi && \
     npm install && \
--- a/components/centraldashboard/public/components/logout-button.js
+++ b/components/centraldashboard/public/components/logout-button.js
@@ -1,0 +1,77 @@
+import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
+
+import '@polymer/iron-ajax/iron-ajax.js';
+import '@polymer/paper-button/paper-button.js';
+
+/**
+ * Logout button component.
+ * Handles the logout requests and post-logout redirects.
+ *
+ */
+
+export class LogoutButton extends PolymerElement {
+    static get template() {
+        return html`
+            <paper-button id="logout-button" on-tap="logout">
+                <iron-icon icon='kubeflow:logout' title="Logout"
+                </iron-icon>
+            </paper-button>
+            <iron-ajax
+                    id='logout'
+                    url='/logout'
+                    method='post'
+                    handle-as='json'
+                    headers='{{headers}}'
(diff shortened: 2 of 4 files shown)


kimwnasptd:
@alembiewski as discussed in https://github.com/kubeflow/manifests/pull/2150#issuecomment-1415915468 we'll need to slightly update this PR. 

Specifically, the new AuthService will expect the logout URL to be `/authservice/logout`. But since we can't hardcode this value, since not everyone might be using AuthService, we'll need to have a dynamic mechanism for learning this URL.

I'd propose to 
1. have a new ENV Var `LOGOUT_URL` in the Deployment of the app
2. The backend will add this for the `/api/workgroup/env-info` route handler https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/app/api_workgroup.ts
3. The frontend will dynamically use this value from the backend

First off we'll need to rebase this work on top of the latest branch. Would you have the cycles for this or should we assign this task to someone else?

alembiewski:
@kimwnasptd I rebased the PR on top of the latest branch. 
> Specifically, the new AuthService will expect the logout URL to be /authservice/logout

Yes, good catch - the way we resolved that is by adding a new rule to a VirualService, e.g.:
```
  - name: logout
    match:
    - uri:
        prefix: /logout
    rewrite:
      uri: /authservice/logout
```
What you suggesting is definitely a better approach to handle this.

Unfortunately, I don't have enough capacity to work on the dynamic mechanism for learning the logout URL, so I would propose we merge the existing PR and ask someone else to address https://github.com/kubeflow/kubeflow/issues/6940 separately. WDYT?

kimwnasptd:
Thanks for confirming @alembiewski!

Yes, we'll merge this PR and handle the fixup in a follow up. @orfeas-k could you take a stab at it?

@alembiewski thank you very much for your time, and apologies for the late review on this that resulted in a bad timing.

kimwnasptd:
The PR changes look good as well, so going forward with merging this. We'll also need to cherry-pick this to the release branch cc @DomFleischmann 

/lgtm
/approve

skuchipu:
@alembiewski @kimwnasptd  This change is redirecting to a blank page , should it not redirect to login page like earlier versions ?

divyank000:
the change of LOGOUT_URL to `/authservice/logout` leads to an invalid_page. 

Not a good  User Experience, thought it doesn't break the functionality.

juliusvonkohout:
Create a new issue including screenshots and detailed explanation if necessary.

rahu984:
> @kimwnasptd I rebased the PR on top of the latest branch.
> 
> > Specifically, the new AuthService will expect the logout URL to be /authservice/logout
> 
> Yes, good catch - the way we resolved that is by adding a new rule to a VirualService, e.g.:
> 
> ```
>   - name: logout
>     match:
>     - uri:
>         prefix: /logout
>     rewrite:
>       uri: /authservice/logout
> ```
> 
> What you suggesting is definitely a better approach to handle this.
> 
> Unfortunately, I don't have enough capacity to work on the dynamic mechanism for learning the logout URL, so I would propose we merge the existing PR and ask someone else to address #6940 separately. WDYT?

But where do we implement this virtual service.

ajinkya933:
> > @kimwnasptd I rebased the PR on top of the latest branch.
> > > Specifically, the new AuthService will expect the logout URL to be /authservice/logout
> > 
> > 
> > Yes, good catch - the way we resolved that is by adding a new rule to a VirualService, e.g.:
> > ```
> >   - name: logout
> >     match:
> >     - uri:
> >         prefix: /logout
> >     rewrite:
> >       uri: /authservice/logout
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > What you suggesting is definitely a better approach to handle this.
> > Unfortunately, I don't have enough capacity to work on the dynamic mechanism for learning the logout URL, so I would propose we merge the existing PR and ask someone else to address #6940 separately. WDYT?
> 
> But where do we implement this virtual service.

can you help me out @kimwnasptd @alembiewski  Im also stuck on this, where is this virtual service, which needs to be changed ?

Issue #6056, opened 2021-07-20
Notebook CRD Instances Missing Status Fields


814HiManny:
/kind bug

**What steps did you take and what happened:**
This is an issue that occurs occasionally, but I don't have enough understandings of the notebook-controller to be able to cause the issue to be reproduced reliably. But it has happened on multiple occasions.

From the user perspective, when they launch a notebook via the web interface, the newly-launched notebook stays with the spinning icon indefinitely and never enables the "CONNECT" button. The statefulset and pod seems to start fine but the notebook is never accessible. 

**What did you expect to happen:**
Notebook should launch and enable "CONNECT" button after a short period for the user to access said notebook.

**Anything else you would like to add:**
notebook-controller logs mention reconciler errors like:
```
2021-07-20T15:12:12.424Z	INFO	controllers.Notebook	Updating Status	{"notebook": "test-ns/test", "namespace": "test-ns", "name": "test"}
2021-07-20T15:12:12.431Z	ERROR	controller-runtime.controller	Reconciler error	{"controller": "notebook", "request": "test-ns/test", "error": "Notebook.kubeflow.org \"test\" is invalid: status.conditions: Invalid value: \"null\": status.conditions in body must be of type array: \"null\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0/pkg/internal/controller/controller.go:218
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0/pkg/internal/controller/controller.go:192
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.2.0/pkg/internal/controller/controller.go:171
k8s.io/apimachinery/pkg/util/wait.JitterUntil.func1
	/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190404173353-6a84e37a896d/pkg/util/wait/wait.go:152
k8s.io/apimachinery/pkg/util/wait.JitterUntil
[...]

PR #6619, opened 2022-08-21
Fix #6528: Mirroring Pod conditions to Notebook

### LINKED ISSUE #6528: Notebook conditions are being appended
/kind bug

Right now the Notebooks controller will be appending new Pod conditions to the Notebook's status https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L247. 

This means that a Notebook could end up with the following conditions being present at the same time:
```
        {
          "lastProbeTime": "2022-04-18T08:29:27Z", 
          "message": "Error", 
          "reason": "Error", 
          "type": "Terminated"
        }, 
        {
          "lastProbeTime": "2022-04-15T06:39:50Z", 
          "type": "Running"
        }, 
```

K8s Conditions are used for expressing the state of an object **at a specific point in time**. 
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

We shouldn't be appending the Pod Conditions to the Notebook Conditions, but rather just mirror the Pod conditions to the Notebook.

If we also add a proper function for converting the PodConditions to Notebook conditions with some basic checks like:
```golang
func PodCondToNotebookCond(c corev1.PodCondition) v1.NotebookCondition {
	notebookCond := v1.NotebookCondition{}

	if len(c.Type) > 0 {
		notebookCond.Type = c.Type
	}

	if len(c.Reason) > 0 {
		notebookCond.Reason = c.Reason
	}
...
```
Then we should be way more solid and also close the relevant, but orthogonal, https://github.com/kubeflow/kubeflow/issues/6056

midhun1998:
Sorry for the delay. I have been bit busy lately. 

I have added the implementation based on @kimwnasptd suggestion to mirror the condition of Notebook Pod and not append container statuses. Correct me if my understanding is incorrect but from what I checked I believe we still have a few fields missing in the Notebook CRD for conditions like "status" and "LastTransitionTime". Let me know if we need to add those too. I can add a PR for that before merging this one in.


kimwnasptd (commented):
Thanks for the PR @midhun1998! I have some nits on renames, but overall LGTM.

I'll also do a test now to verify it works as expected. Also cc @apo-ger since we wanted to include this logic as part of the effort for https://github.com/kubeflow/kubeflow/issues/6056

kimwnasptd:
> Correct me if my understanding is incorrect but from what I checked I believe we still have a few fields missing in the Notebook CRD for conditions like "status" and "LastTransitionTime"

@midhun1998 indeed! We need these two fields to be on par with [Pod Conditions](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#podcondition-v1-core) could you add these two as well as part of your PR?

kimwnasptd:
Then the last item to re-evaluate would be the `status.containerState` of the Notebook CRs. Right now it will always be `{}` 
```yaml
status:
  conditions:
  - lastProbeTime: null
    type: Initialized
  - lastProbeTime: null
    type: Ready
  - lastProbeTime: null
    type: ContainersReady
  - lastProbeTime: null
    type: PodScheduled
  containerState: {}
  readyReplicas: 1
```

But we can tackle this one in the follow-up PR from @apo-ger, for https://github.com/kubeflow/kubeflow/issues/6056, where we are updating how the controller sets the status of the CRs.

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -241,19 +241,15 @@ func (r *NotebookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 				if pod.Status.ContainerStatuses[i].State == instance.Status.ContainerState {
 					continue
 				}
-
-				log.Info("Updating Notebook CR state: ", "namespace", instance.Namespace, "name", instance.Name)
-				cs := pod.Status.ContainerStatuses[i].State
-				instance.Status.ContainerState = cs
```
kimwnasptd, line 247: Lastly, let's keep this logic that updates the `instance.Status.ContainerState` of the CR. The web app is using this information.

midhun1998:
> > Correct me if my understanding is incorrect but from what I checked I believe we still have a few fields missing in the Notebook CRD for conditions like "status" and "LastTransitionTime"
> 
> @midhun1998 indeed! We need these two fields to be on par with [Pod Conditions](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#podcondition-v1-core) could you add these two as well as part of your PR?

Thanks for the confirmation @kimwnasptd ! I was wondering if the changes should go to all 3 APIs (v1, v1alpha1, and v1beta1) or just the v1. Could you please clarify this too?

components/notebook-controller/controllers/notebook_controller.go
midhun1998: Agree

components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -241,19 +241,15 @@ func (r *NotebookReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
-					oldConditions[0].Reason != newCondition.Reason ||
-					oldConditions[0].Message != newCondition.Message {
-					log.Info("Appending to conditions: ", "namespace", instance.Namespace, "name", instance.Name, "type", newCondition.Type, "reason", newCondition.Reason, "message", newCondition.Message)
-					instance.Status.Conditions = append([]v1beta1.NotebookCondition{newCondition}, oldConditions...)
+				notebookCondition := []v1beta1.NotebookCondition{}
+				for i := range pod.Status.Conditions {
+					log.Info("Updating Notebook CR state: ", "namespace", instance.Namespace, "name", instance.Name)
+					// Mirroring pod condition
+					condition := PodCondToNotebookCond(pod.Status.Conditions[i])
+					log.Info("Mirroring pod conditions: ", "namespace", instance.Namespace, "name", instance.Name, "type", condition.Type, "reason", condition.Reason, "message", condition.Message)
```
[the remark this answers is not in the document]
midhun1998: Oops! Will fix this

kimwnasptd:
@midhun1998 I'm just summarizing the changes I proposed, since I did a small back and forth regarding the `containerState`:
1. Let's do the (nit) renamings when logging
2. Let's add the `Status` and `LastTransitionTime` to the API of the notebook conditions. Let's do this for all the versions of the CRD
3. Let's keep the logic that sets the `containerState`, since the web apps use this information

midhun1998:
> @midhun1998 I'm just summarizing the changes I proposed, since I did a small back and forth regarding the `containerState`:
> 
> 1. Let's do the (nit) renamings when logging
> 2. Let's add the `Status` and `LastTransitionTime` to the API of the notebook conditions. Let's do this for all the versions of the CRD
> 3. Let's keep the logic that sets the `containerState`, since the web apps use this information

Thanks for summarizing it @kimwnasptd . The requested changes have been added.

kimwnasptd:
@midhun1998 the changes look good, thank you for your work!

/lgtm
/approve

PR #6623, opened 2022-08-22
Fix #66622: rename environment variable for dashboard configmap

### LINKED ISSUE #6622: Dashboard manifest uses incorrect environment variable to pass configmap name
[This line](https://github.com/kubeflow/kubeflow/blob/0a79697d8eb7aa7b13f2fdc24fc99150ee8feddf/components/centraldashboard/manifests/base/deployment.yaml#L41) in the dashboard manifests assigns the name of the configmap for dashboard links to the `DASHBOARD_LINKS_CONFIGMAP` environment variable, but the code [looks for `DASHBOARD_CONFIGMAP` instead](https://github.com/kubeflow/kubeflow/blob/e6fdf515c09ba1f695d60a87c9c56ba5fb1718e8/components/centraldashboard/app/k8s_service.ts#L5).  

This error is likely due to how the environment variable was [renamed in the past](https://github.com/kubeflow/kubeflow/pull/5469/files#diff-a29b81659d6ca88b1df24f77d04886f8eb04631f08a34e41da53e6bb273e3be0R5) and manifests were not updated.  The only reason things work currently is because the default value specified in the manifests (`centraldashboard-config`) happens to match the default value in the typescript code that's used if no variable is set.

ca-scribner:
Fixes #6622, where the manifests for deployment uses a deprecated name for the environment variable that specifies the dashboard's link configmap. 

As far as I can tell this is the only place where this error occurs, but I have not robustly checked for it elsewhere.

### DIFF
--- a/components/centraldashboard/manifests/base/deployment.yaml
+++ b/components/centraldashboard/manifests/base/deployment.yaml
@@ -38,6 +38,6 @@ spec:
           value: profiles-kfam.kubeflow
         - name: REGISTRATION_FLOW
           value: $(CD_REGISTRATION_FLOW)
-        - name: DASHBOARD_LINKS_CONFIGMAP
+        - name: DASHBOARD_CONFIGMAP
           value: $(CD_CONFIGMAP_NAME)
       serviceAccountName: centraldashboard


thesuperzapper:
Closing in preference of https://github.com/kubeflow/kubeflow/pull/7264

Issue #6618, opened 2022-08-20
Profiler does not properly annotate the default-editor service account


ryansteakley:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
Installed Kubeflow manifests v1.6.0-rc.1
Use plugins:
  - kind: AwsIamForServiceAccount
    spec:
      awsIamRole:

The default-editor service-account is not getting annotated properly on v1.6.0-rc.1 Tested on 5+ EKS clusters on both 1.21 and 1.22. Then uninstalled and cherry-picked the pipeline changes from v1.6.0-rc.1 and reverted them back to v1.6.0-rc.0 state. Now the EKS 1.21 clusters are working and the service account is being annotated correctly. As expected 1.22 fails this is because the v1.6.0-rc.1 changes are to support Kubernetes versions of 1.22+.

Currently the annotation looks like

```
Name:                default-editor
Namespace:           test-profile
Labels:              <none>
Annotations:         eks.amazonaws.com/role-arn: 
Image pull secrets:  <none>
Mountable secrets:   default-editor-token-jj95z
Tokens:              default-editor-token-jj95z
Events:              <none>
```
Even if I manually edit the annotation it will instantly change back to the way it is shown above.
**What did you expect to happen:**
default-editor service-account to be annotated properly

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): v1.6.0-rc.1
- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`) eks v1.21/1.22
- Kubernetes version: (use `kubectl version`): v1.22
- OS (e.g. from `/etc/os-release`): mac-osx

kimwnasptd:
@apo-ger let's take a look at this one as well, since we've seen distributions bumping into it https://github.com/kubeflow/manifests/issues/2221#issuecomment-1221199387

The only PR between RC0 and RC1 is for the Kubebuilder update https://github.com/kubeflow/kubeflow/pull/6491

surajkota:
[...]

PR #6624, opened 2022-08-24
fix Profile Plugins functionality

surajkota:
Resolves: #6618

Description of changes:
- As described in https://github.com/kubeflow/kubeflow/issues/6618#issuecomment-1223404464, profile plugins functionality broke when the yaml package was changed as part of #6491. Since the structs have JSON tags, we can safely serialize and deserialize the structs using Go's JSON library.
- Added unit tests to prevent this from happening in future.
- Added a validation in the plugin to error out if the role value is missing because golang json library will create the struct even if the value is missing. Since the struct is small at this point, I have added a simple validation but this solution is not great if the struct is complex.
- Added samples which were deleted.

Testing:
- Ran unit tests
- Tested with and without the AWS plugin by running the controller locally
- Tested that GCP plugin struct as well to ensure it is loaded but have not run a real e2e test. Someone using GCP will need to confirm

### DIFF
--- a/.github/workflows/prof_controller_unit_test.yaml
+++ b/.github/workflows/prof_controller_unit_test.yaml
@@ -1,4 +1,4 @@
-name: Run unit tests for Profile Controller
+name: Run Profile Controller unit tests
 on:
   pull_request:
     paths:
--- a/.github/workflows/prof_controller_unit_test.yaml
+++ b/.github/workflows/prof_controller_unit_test.yaml
@@ -1,0 +1,23 @@
+name: Run unit tests for Profile Controller
+on:
+  pull_request:
+    paths:
+      - components/profile-controller/**
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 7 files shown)


surajkota:
@kimwnasptd

kimwnasptd:
Thanks for the prompt PR @surajkota!

The changes LGTM. Could you also add a simple gh action for running the `make test` rule to ensure we keep testing this functionality on each PR?

You can create a `prof_controller_unit_test.yaml` under https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/ for this that just runs `make test`. We can use the https://github.com/actions/setup-go action for setting up the Go env

surajkota:
@kimwnasptd thats a good idea. Added it

I also opened https://github.com/kubeflow/kubeflow/pull/6627 for KServe, can you please take a look

kimwnasptd:
Thanks @surajkota! Will take a look to the KServe one as well.

/lgtm
/approve

Issue #5965, opened 2021-06-03
profiles: Work with InferenceServices


yanniszark:
/kind bug

**What steps did you take and what happened:**

Deploying an InferenceService in a Profile namespace does not work by default, because:
- The InferenceService Pod is injected with an Istio sidecar.
- The Istio sidecar does not accept traffic from Knative, which needs to ping the Pod in order to work correctly.

**What did you expect to happen:**

Profiles should work with InferenceServices.

I would like to discuss how we should solve this problem and get input from KFServing (cc @kubeflow/wg-serving-leads).
For example:
- Should we extend KFServing examples to disable sidecar injection by default?
- Should we apply some sort of policy to the Profile namespaces that will make them work well with Knative? If yes, what should this policy be? 

**Anything else you would like to add:**

@Subreptivus has created a PR extending the profile-controller to apply an Istio policy that allows all requests to `/healthz`, `/metrics` paths for KFServing. Is that an ideal solution, given that it affects all Pods and not just KFServing Pods?

cc @kubeflow/wg-notebooks-leads

DavidSpek:
I'd just like to point out that the PR by @Subreptivus is also what I found to be necessary for KNative and the KNative docs also say this authorization policy is necessary. 

https://knative.dev/v0.22-docs/serving/istio-authorization/#allowing-access-from-system-pods-by-paths

A bit higher on that page it also states that it might be necessary to add the `knative-serving` namespace to the allow policy for the user namespace.

yuzisun:
There is discussion on [this issue](https://github.com/kubeflow/kfserving/issues/1558), I think profile controller is the right place to create the authorization policy, the policy can also use the workload label for the fine grain control.

yanniszark:
Thanks @yuzisun!
From issue https://github.com/kubeflow/kfserving/issues/1558, I understand that ss @pvaneck mentioned, there is [a recommendation from Knative](https://knative.dev/docs/serving/istio-authorization/#allowing-access-from-system-pods-by-paths):

```yaml
apiVersion: security.istio.io/v1beta1
kind: AuthorizationPolicy
metadata:
  name: allowlist-by-paths
  namespace: serving-tests
spec:
  action: ALLOW
  rules:
  - to:
    - operation:
        paths:
        - /metrics   # The path to collect metrics by system pod.
        - /healthz   # The path to probe by system pod.
```

I think we can make this even better by taking advantage of Knative's default labels:
[...]

PR #6627, opened 2022-08-25
Profile auth policy to work with InferenceServices

surajkota:
Workaround for: #5965, https://github.com/kserve/kserve/issues/1558

KServe integration with profiles is not complete. Users can create an inferenceservice but cannot use them(i.e. get prediction requests served) from inside or outside the cluster without doing one of the following(according to https://github.com/kserve/kserve/issues/1558):
- User either needs to disable istio injection
- or apply an Authorization policy with the paths in this PR.

This PR automates the generation of the required authorization policy in profile namespace but this also means that a user can get data from these paths from other user namespaces.

@kimwnasptd @yuzisun Question: Can we add these paths to profile's authorization policy and call out the limitation with multi user isolation in docs until https://github.com/kserve/kserve/issues/1371 is resolved?

Testing:
1. Created an isvc https://github.com/kserve/kserve/blob/release-0.7/docs/samples/v1beta1/sklearn/v2/sklearn.yaml in a profile namespace (kserve-profile)
3. Tried sending request from inside and outside the cluster as a user who owns the profile(kserve-profile) - Succeeded
4. Tried sending request from inside and outside the cluster as a user of a different profile - Also, Succeeded


AlexandreBrown (approved):

juliusvonkohout:
You should rather work on the proper upstream issue to solve this once and for all. I prefer security by default such that the user has to explictly disable the security. Also hardening via https://github.com/kubeflow/kubeflow/pull/6013 is worth a look

There are three possibilities until the upstream stuff is properly fixed

1. Add an authorization policy to your own namespace. We can do this with a kyverno policy or wait for the pull request to be merged. Please note that with `sidecar.istio.io/inject: "true"` you have to be on a pod that has istio enabled - for example a jupyterlab inside your namespace - to access the prediction service.
``` YAML
kind: AuthorizationPolicy
metadata:
  name: allow-in-cluster-kserve
spec:
  rules:
    - to:
        - operation:
            paths:
              - /v1/models/*
              - /v2/models/*
```
2. Disable the istio sidecar in your yaml that is responsible for authentication and live with the security implications
3. Use a path based virtual service with dex and full authentication

All three of them can be implemented with simple kyverno clusterpolicies if you want to have them by default. So i favor not merging this and putting more pressure on upstream.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -442,6 +442,9 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 								"/healthz",
 								"/metrics",
 								"/wait-for-drain",
+								"/ready",
+								"/v1/models/*",
```
yuzisun, line 446: @surajkota allowing paths does not lock down the traffic from other namespaces. I think we still want the changes in https://github.com/kubeflow/kubeflow/pull/6013/files for only allowing requests from `knative-serving` namespace and in addition we need to lock down the original namespace of the request, unfortunately I couldn't find a header to check the original namespace when the request proxyed by activator and the closest one I found is the header `k-original-host` but that requires creating the authorization policy per service level instead of namespace level.

yuzisun:
@surajkota What's the plan for this PR?

components/profile-controller/controllers/profile_controller.go
umka1332: Maybe we should rely on kubeflow-userid header?

thesuperzapper:
We need to update/replace this PR with one that has a dynamic principal (so we don't assume the ServiceAccount or Namespace of the KServe), for example, this PR did it for Notebooks/Pipelines: https://github.com/kubeflow/kubeflow/pull/7310

NOTE: this version of the PR does not actually restrict by source principal (but it should), see the original one for more context: https://github.com/kubeflow/kubeflow/pull/6013

PR #6631, opened 2022-08-30
release: Images for the v1.6.0-rc.2 tag

apo-ger:
Update the manifests to use the v1.6.0-rc.2 tag.

/cc @kimwnasptd


kimwnasptd:
@NickLoukas it seems there's a bug with the GH Actions. Specifically what happened was:
1. The `Login to DockerHub` step did not run, which was expected https://github.com/kubeflow/kubeflow/blob/v1.6-branch/.github/workflows/centraldb_docker_publish.yaml#L33
2. The `Push to DockerHub` step was executed since we only check if the `VERSION` file changed https://github.com/kubeflow/kubeflow/blob/v1.6-branch/.github/workflows/centraldb_docker_publish.yaml#L54

The bug here is that the push action is triggered when we update the file. But we want to trigger it only if that file changed AND it's a `push` event. Can you take a look into it?

We should be able to use expressions for this AND. And we'll need to do this for all components. Lastly, after we fix this, let's ensure all the necessary commits will find their way back to the `master` branch as well https://docs.github.com/en/actions/learn-github-actions/expressions

cc @annajung

NickLoukas:
I've sent a PR to address the above, here: https://github.com/kubeflow/kubeflow/pull/6632

kimwnasptd:
@apo-ger can you do a rebase on top of the latest `v1.6-branch` that includes @NickLoukas' fixes?

This should make all the tests pass and we can then cut the RC

PR #6632, opened 2022-08-31
Ensure we build images for a new Version only on push events

NickLoukas:
Fix the GH actions after this comment: https://github.com/kubeflow/kubeflow/pull/6631#issuecomment-1232223808


kimwnasptd:
Looks good, and looking closely with `grep` it seems we've updated all the correct places. Thanks for the work!

/lgtm
/approve

PR #6656, opened 2022-09-20
chore(icons): the model icon is grey instead of black

juliusvonkohout:
@kimwnasptd all other icons are black so i fixed this one. This also exposes another bug int the centraldashboard UI. All other icons are adapted with css, models.svg is not. Also when selecting a menu entry the models icon stays the same while for other menu entries the icons turn white.


juliusvonkohout:
grafik


I tried
```
rm components/centraldashboard/public/assets/icons/models.svg
cp components/centraldashboard/public/assets/icons/pipeline.svg components/centraldashboard/public/assets/icons/models.svg
```
so it is really broken upstream just for the models entry.

kimwnasptd:
Good catch! Yes, let's have the icon to be black like the other ones

/lgtm
/approve

PR #6657, opened 2022-09-21
Support arbitrary GPU resource vendors

typhoonzero:
Add support for configuring arbitrary GPU resource types like https://github.com/4paradigm/k8s-vgpu-scheduler, https://github.com/tkestack/gpu-manager etc.

截屏2022-09-21 09 09 36

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -230,22 +230,33 @@ def set_notebook_gpus(notebook, body, defaults):
     if "num" not in gpus:
         raise BadRequest("'gpus' must have a 'num' field")
 
-    if gpus["num"] == "none":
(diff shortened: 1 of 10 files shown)


kimwnasptd:
@typhoonzero it's not clear to me which yaml configurations we want to support, and which projects enable them. Could you provide some more info regarding:
1. One or two example YAML files of Notebooks we want to apply, and how the `requests` and `limits` for the GPUs should be
2. What is the use case of these configurations

typhoonzero:
@kimwnasptd Thanks for reply. I've updated the description of this PR, provided some usecase explainations and sample YAMLs.

thesuperzapper:
I am not sure this is the best way for users to request specific GPUs, but I will add it to the `1.9.1` milestone so we can discuss if we will move forward with this.

I also note that what this PR archives can already be done (mostly) with `affinityConfig` options which select on Node labels.

PR #6663, opened 2022-09-29
Remember the most recently selected namespace

TobiasGoerke:
# The Problem 

When visiting Kubeflow, [the first owned namespace is selected automatically](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/public/components/namespace-selector.js#L168).

As a user, this annoys me when: 
- I don't own a namespace and am only a contributor to namespaces. In this case, no namespace is selected, and I need to make a selection each time I visit Kubeflow.
- There are multiple namespaces available, as a "random" selection will be made for me.

# My Proposal

In this PR, I've integrated the local storage to save the user's last selection. When visiting Kubeflow without the namespace being part of the URL, the local storage is prompted, and - if possible - the user's previous selection restored.

### DIFF
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -307,7 +307,7 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         if (namespace) {
             // Save the user's choice so we are able to restore it, 
             // when re-loading the page without a queryParam
-            const localStorageKey = "selectedNamespace" + 
+            const localStorageKey = "/centraldashboard/selectedNamespace/" +
                 (this.user && "." + this.user || "");
             localStorage.setItem(localStorageKey, namespace);
         }
--- a/components/centraldashboard/public/components/main-page.js
+++ b/components/centraldashboard/public/components/main-page.js
@@ -307,7 +307,9 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
         if (namespace) {
             // Save the user's choice so we are able to restore it, 
             // when re-loading the page without a queryParam
-            localStorage.setItem("selectedNamespace", namespace);
+            const localStorageKey = "selectedNamespace" + 
+                (this.user && "." + this.user || "");
+            localStorage.setItem(localStorageKey, namespace);
         }
 
         if (this.namespacedItemTemplete &&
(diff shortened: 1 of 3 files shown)


kimwnasptd:
This is a pretty good addition @TobiasGoerke! With a quick glance it looks good. Will also build the image to verify and then merge

thesuperzapper:
@TobiasGoerke looks good, but what happens when I log out and sign in to a user who does not have permission to view the namespace that is in local storage?

Should we make the storage per-user?

juliusvonkohout:
> > @TobiasGoerke looks good, but what happens when I log out and sign in to a user who does not have permission to view the namespace that is in local storage?
> > 
> > Should we make the storage per-user?
> 
> Is users logging in to different accounts an actual use case? If so, prepending the username to the localStorage var should work. 

Yes, having usernames and passwords for multiple namespaces is definitely a common use case.

TobiasGoerke:
The local storage's key now includes the currently logged-in user, as you requested. 
Thus, multi-user support on the same machine is given.

components/centraldashboard/public/components/main-page.js
```diff
@@ -303,6 +303,15 @@ export class MainPage extends utilitiesMixin(PolymerElement) {
     _namespaceChanged(namespace) {
         // update namespaced menu item when namespace is changed
         // by namespace selector
+
+        if (namespace) {
+            // Save the user's choice so we are able to restore it, 
+            // when re-loading the page without a queryParam
+            const localStorageKey = "selectedNamespace" + 
+                (this.user && "." + this.user || "");
```
kimwnasptd, line 311: Could we change this so that the localStorage key is `/centraldashboard/selectedNamespace/<user>`?

I'd like to make it explicit which localStorage key is set by which app, in case in the future more apps would use the localStorage

components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -196,6 +197,18 @@ export class NamespaceSelector extends PolymerElement {
      * @return {string}
      */
     getDefaultNamespace() {
+        // Restore the user's previous namespace choice
+        const localStorageKey = "selectedNamespace" + 
+            (this.user && "." + this.user || "");
```
kimwnasptd, line 202: I think we'll need to change this one as well, in the same spirit as the above comment

kimwnasptd (commented):
@TobiasGoerke apologies for the late reply here. 

I've taken a look at the PR and it works as expected! I only have two small nits regarding the naming of the localStorage key.

rest LGTM

kimwnasptd:
Thanks @TobiasGoerke @juliusvonkohout, that's a really nice feature!

/lgtm
/approve

Issue #6610, opened 2022-08-11
Kubeflow UI has a giant div blocking most of the screen space


cheungringo:
/kind bug

**What steps did you take and what happened:**
I'm a heavy user of the Kubeflow UI to read through logs to debug pipelines that my company runs. For some reason, starting Aug 10th, a giant div (the footer I believe) has appeared and is blocking most of the screen space which makes debugging and visualizing my pipelines very difficult. 

**What did you expect to happen:**
Not to have a giant div blocking the screen.

**Anything else you would like to add:**
Please see attached screenshot.
![Screen Shot 2022-08-10 at 11 50 24 AM](https://user-images.githubusercontent.com/33941762/184175354-a7e3da35-fa98-4d9f-8903-df38426babdb.png)

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): build version dev_local
- kfctl version: (use `kfctl version`): couldn't find this command on kubeflow cluster
- Kubernetes platform: (e.g. `minikube`): minikube version: v1.26.0
- Kubernetes version: (use `kubectl version`): Client Version: v1.24.3
Kustomize Version: v4.5.4
Server Version: v1.21.12-gke.2200
- OS (e.g. from `/etc/os-release`):

sethjones:
To add a little more information to this report:
- This appeared in one of the latest Firefox releases (for me).  I believe it showed up after the FF 103 release.
- Testing in Chromium (104.0.5112.79), and Firefox ESR (91.11.0esr) - I was unable to reproduce. 

My setup stats:
- Kubeflow 1.5.0
- K8s 1.21.12

fabito:
I was able to reproduce the behaviour in Chrome: `104.0.5112.79-1` and `103.0.5060.53-1`
It works fine in FF 103.0 (64-bit) though.

daisuke834:
I also have the same issue with `Chrome 104.0.5112.79`.
It works fine with `Safari 15.2 (17612.3.6.1.6)` so far.

ittus:
It seems the issue is caused by browser's LastPass extension.
After disabling it, Kubeflow UI works as normal.

https://community.logmein.com/t5/LastPass-Support-Discussions/Lastpass-is-adding-a-CSS-element-below-footer/m-p/291627

sethjones:
> It seems the issue is caused by browser's LastPass extension. After disabling it, Kubeflow UI works as normal.
> 
[...]

PR #6664, opened 2022-10-02
fix: giant div overlay in main page

ittus:
## Overview
Fix div overlay on main page https://github.com/kubeflow/kubeflow/issues/6610

## Implementation details
- Only set `flex: 1` on main-page component

## How to test
Run on local
**Before**
Screen Shot 2022-10-03 at 0 00 38

**After**
Screen Shot 2022-10-03 at 0 01 24

## Alternative ways
Removing overlay div directly

```css
body > iron-a11y-announcer + div {
    display: none;
}
```

### DIFF
--- a/components/centraldashboard/public/styles.css
+++ b/components/centraldashboard/public/styles.css
@@ -16,7 +16,7 @@ html, body {
     box-sizing: border-box;
 }
 
-body>* {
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
/close

No activity

PR #6666, opened 2022-10-05
fix: typo when setting server_type for jupyter notebook (no in app affect)

johnhoman:
I'm fairly certain this is a typo


thesuperzapper:
@johnhoman I agree that this is a typo, since it's a very small fix I am ok with merging. Either way, are you aware of any direct effects of this problem?

/lgtm

/assign @kimwnasptd

johnhoman:
> @johnhoman I agree that this is a typo, since it's a very small fix I am ok with merging. Either way, are you aware of any direct effects of this problem?
> 
> /lgtm
> 
> /assign @kimwnasptd

I haven't noticed anything break from this. I just happen to notice it while I was reading through the code

andreyvelich:
Please re-open PR there.
/close

PR #6673, opened 2022-10-13
Docker -> OCI Update form-image.component.html

juliusvonkohout:
Docker is dead and superseeded by OCI on all common Platforms


kimwnasptd:
Looks good!

/lgtm
/approve

PR #6674, opened 2022-10-17
centraldashboard: Show objects from all namespaces

tasos-ale:
This PR is required for the ["Show objects from all namespaces"](https://github.com/kubeflow/kubeflow/blob/master/components/proposals/20220621-all-namespaces.md. ) proposal and introduces the changes needed in the centraldashboard.

### DIFF
--- a/components/centraldashboard/public/components/namespace-selector.js
+++ b/components/centraldashboard/public/components/namespace-selector.js
@@ -7,6 +7,12 @@ import '@polymer/paper-item/paper-item.js';
 
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 
+export const ALL_NAMESPACES = 'All namespaces';
+export const ALL_NAMESPACES_ALLOWED_LIST = ['jupyter'];
+
+const allNamespacesAllowedPaths = ALL_NAMESPACES_ALLOWED_LIST
+    .map(( p)=>`/_/${p}/`);
+
 /**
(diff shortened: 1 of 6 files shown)


components/centraldashboard/public/components/namespace-selector.js
```diff
@@ -7,6 +7,12 @@ import '@polymer/paper-item/paper-item.js';
 
 import {html, PolymerElement} from '@polymer/polymer/polymer-element.js';
 
+export const ALL_NAMESPACES = 'All namespaces';
+export const ALL_NAMESPACES_ALLOWED_LIST = ['jupyter'];
```
kimwnasptd, line 11: @tasos-ale let's have this as an empty list initially, since we don't have an app that supports it at this point in time

kimwnasptd (commented):
@tasos-ale the changes look good!  Only have a minor nit.

Afterwards we can start working on the dedicated web apps

components/centraldashboard/public/components/namespace-selector.js
kimwnasptd: Ideally I'd like this to info (which web app supports all namespaces) to be transmitted via the common library, so that the dashboard can dynamically know if an app supports this once it loads it.

But we can do with hardcoding the urls for now. If in the future we see a bigger need for this we can look into it

kimwnasptd:
Thanks! Next up is the work for each web app. I'll start with work I had in the JWA

/lgtm
/approve

PR #6675, opened 2022-10-18
jupyter-webapp: Add image name validation for image name

### LINKED ISSUE #6621: Notebook creation UI missing validation on "custom image" field
/kind bug

1. Filled out the `custom image` field of the Kubeflow Notebook creation UI with an invalid image URI with a leading space.
2. Created a notebook
3. After creation, notebook UI shows loading spinner forever, events revealed StatefulSet creation was failing in notebook controller

```
$ kubectl get event --sort-by='.lastTimestamp' -n my-ns

LAST SEEN   TYPE      REASON                         OBJECT                                          
13m         Warning   FailedCreate                   statefulset/a100x16                             

MESSAGE
create Pod a100x16-0 in StatefulSet a100x16 failed error: Pod "a100x16-0" is invalid: 
spec.containers[0].image: Invalid value: " gcr.io/leading-space-example": 
must not have leading or trailing whitespace
```

image

image

Custom image field is validated according to a regex

**Environment:**

- Kubeflow version:  1.4

mishraprafful:
## Related Issues/PRs
* Closes #6621

## Comments
* I could not find tests for tests for the `jupyter backend` webapp, please  let me know if I should create an issue regarding this.

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -1,4 +1,5 @@
 import json
+import re
 
 from werkzeug.exceptions import BadRequest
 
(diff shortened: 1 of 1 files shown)


kimwnasptd:
Thanks for the PR @mishraprafful!

How was this regex created? Ideally I'd like us to use the same logic that K8s itself uses for validating the image string.

Regarding the tests, I'm preparing a first PR for these as we speak, so we'll need to do some coordination. I'll give you a heads up once it's up and merged so that we can add some tests for this one as well.

mishraprafful:
Thanks for your comments @kimwnasptd

* I've updated the regex to use the same one as k8s and I've added the source as a comment for ease of understanding and maintenance.
* I was planning to add [these](https://regex101.com/r/af8fOz/1) unit-tests for the regex.

kimwnasptd:
@mishraprafful after some more consideration I think that the place to solve this issue is not in the backend, but actually in the CRD.

We should have a proper validation webhook instead, that ensures the CR is valid. This will be a bigger effort though. 

@mishraprafful would you be interested in contributing to this?

mishraprafful:
@kimwnasptd That sounds good, I'm not super familiar with the notebook controller codebase though, but I think I'd like to give it a try. 

Do you think I should create a new issue for that and continue the work there or useing the existing issue with a new PR would be good?

mishraprafful:
Closing in favour of #6734

PR #6678, opened 2022-10-20
Fix notebook annotation not included

### LINKED ISSUE #6677: Notebook's annotations are not included
/kind bug

1. create a notebook.yaml
```yaml
---
apiVersion: kubeflow.org/v1
kind: Notebook
metadata:
  name: notebook-sample-v1
  annotations:
    notebooks.kubeflow.org/server-type: "foo"
spec:
  template:
    spec:
      containers:
        - name: notebook-sample-v1
          image: kubeflownotebookswg/jupyter:v1.6.0-rc.0

```

2. kubectl create -f notebook.yaml

3. The annotation of pod is none, `notebooks.kubeflow.org/server-type: "foo"` is not included.

image

I expect that the annotation of `notebooks.kubeflow.org/server-type: "foo"`  to be included.

like this :
image

The generateStatefulSet func doesn't copy all the Notebook annotations to the pod now.

**Environment:**

- kfctl version: (use `kfctl version`):
- Kubernetes platform:  minikube
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

youngMan4:
This pr fix #6677

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -437,7 +437,8 @@ func generateStatefulSet(instance *v1beta1.Notebook) *appsv1.StatefulSet {
 				ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{
 					"statefulset":   instance.Name,
 					"notebook-name": instance.Name,
-				}},
+				},
+					Annotations: map[string]string{}},
 				Spec: *instance.Spec.Template.Spec.DeepCopy(),
 			},
 		},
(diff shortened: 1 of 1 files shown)


components/notebook-controller/controllers/notebook_controller.go
```diff
@@ -448,6 +449,12 @@ func generateStatefulSet(instance *v1beta1.Notebook) *appsv1.StatefulSet {
 		(*l)[k] = v
 	}
 
+	// copy all the Notebook annotations to the pod
+	a := &ss.Spec.Template.ObjectMeta.Annotations
+	for k, v := range instance.ObjectMeta.Annotations {
+		(*a)[k] = v
```
MartinForReal, line 455: can we replace a with ss.Spec.Template.ObjectMeta.Annotations?

thesuperzapper:
Personally, I think we should not be including the annotations from the Notebook CRD on the Pod itself.

No other Pod-wrapper CRD does this (Deployment, ReplicaSet, etc), and it would cause issues with tools like ArgoCD which can use annotations to track which resources are "owned" by specific Applications.

PR #6681, opened 2022-10-20
fix(volume name): update formCtrl.get()

bryanpaget:
The Angular code for `<app-form-data-volumes>` was missing `.value`.

This error resulted in the name of new data volumes being injected with `[object Object]` instead of the corresponding notebook server name.

image

More information can be found here: https://github.com/StatCan/jupyter-apis/issues/168.


kimwnasptd:
Nice catch!

/lgtm
/approve

PR #6395, opened 2022-03-05
added env for app secure cookies

BenzhaminKim:
Based on the @kimwnasptd  [comment](https://github.com/kubeflow/manifests/pull/2155#issuecomment-1059143758), I make this PR in `kubeflow/kubeflow`.

[PR from kubeflow/manifests](https://github.com/kubeflow/manifests/pull/2155)

**Which issue is resolved by this Pull Request:**
Resolves [#5803](https://github.com/kubeflow/kubeflow/issues/5803)
**Description of your changes:**

People could get confused when they use Kubeflow over HTTP. Because of HTTP request, they get error `Could not find CSRF cookie XSRF-TOKEN in the request`. 

Therefore, I added `APP_SECURE_COOKIES` environment variable to `Jupyter, Tensorboards and Volumes Web Apps`. You can install and use Kubeflow over HTTP by changing `true` to `false` like  `JWS_APP_SECURE_COOKIES=false`, `TWA_APP_SECURE_COOKIES=false`, `VWA_APP_SECURE_COOKIES=false` in each params.env file.

This PR will help people to set Kubeflow over http easily.

PR #6688, opened 2022-10-28
fix APP_SECURE_COOKIES manifest parameters

thesuperzapper:
This PR fixes the change made in https://github.com/kubeflow/kubeflow/pull/6395, because it did not tell the Kustomize manifests to treat `TWA_APP_SECURE_COOKIES` as a variable in `ConfigMap/parameters`.

This causes the current manifests for Kubeflow 1.6.1 to set the [`APP_SECURE_COOKIES` env-var](https://github.com/kubeflow/kubeflow/blob/17740d765223ef054229631ea327dfee87956f99/components/crud-web-apps/tensorboards/manifests/base/deployment.yaml#L24-L25) as the literal value `"$(TWA_APP_SECURE_COOKIES)"`, rather than `"true"` or `"false"`.

### DIFF
--- a/components/crud-web-apps/jupyter/manifests/base/kustomization.yaml
+++ b/components/crud-web-apps/jupyter/manifests/base/kustomization.yaml
@@ -76,3 +76,10 @@ vars:
     apiVersion: v1
     kind: ConfigMap
     name: parameters
+- name: JWA_APP_SECURE_COOKIES
+  fieldref:
+    fieldPath: data.JWA_APP_SECURE_COOKIES
+  objref:
+    apiVersion: v1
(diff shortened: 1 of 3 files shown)


thesuperzapper:
/assign @kimwnasptd 

We might also want to cherry-pick this into the `v1.6-branch` branch in-case we ever decide to release a `1.6.2` version.

thesuperzapper:
I now realize that this error affects the volumes-web-app also, so I have fixed that one too.

thesuperzapper:
Lol, this also affects jupyter-web-app, so I have added ANOTHER commit to fix that!

PR #6689, opened 2022-10-28
add manifest parameters for notebook culling

thesuperzapper:
This PR fixes 2 things:

1.  the [`params.env`](https://github.com/kubeflow/kubeflow/blob/17740d765223ef054229631ea327dfee87956f99/components/notebook-controller/config/manager/params.env) file in the notebook-controller manifests was not actually used
2. there was no way (without using a full-blown kustomize patch) to configure notebook culling, there are now kustomize parameters for each of:
    - `CLUSTER_DOMAIN`
    - `ENABLE_CULLING`
    - `CULL_IDLE_TIME`
    - `IDLENESS_CHECK_PERIOD`


thesuperzapper:
@kimwnasptd This is a small but important change, I would appreciate a review!

/assign @kimwnasptd

surajkota (approved):

PR #6692, opened 2022-10-30
Make admission-webhook port number configurable

sel:
Previously the webhook listened on a fixed port, 4443, which can clash
with other services when the webhook is run on the host network in
Kubernetes, which is required when using some CNI implementations,
notably Calico on EKS [1].

Enable configuration of the webhook listen port via the program flags.

[1] https://projectcalico.docs.tigera.io/getting-started/kubernetes/managed-public-cloud/eks#install-eks-with-calico-networking

Signed-off-by: Steve Larkin <steve.larkin@gmail.com>

### DIFF
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -684,15 +684,17 @@ func serveMutatePods(w http.ResponseWriter, r *http.Request) {
 
 func main() {
 	var config Config
+	var port int
 	flag.StringVar(&config.CertFile, "tlsCertFile", "/etc/webhook/certs/cert.pem", "File containing the x509 Certificate for HTTPS.")
 	flag.StringVar(&config.KeyFile, "tlsKeyFile", "/etc/webhook/certs/key.pem", "File containing the x509 private key to --tlsCertFile.")
+	flag.IntVar(&port, "webhookPort", 4443, "Port number on which the webhook listens.")
 	flag.Parse()
 	klog.InitFlags(nil)
 
 	http.HandleFunc("/apply-poddefault", serveMutatePods)
 
 	server := &http.Server{
-		Addr:      ":4443",
+		Addr:      fmt.Sprintf(":%d", port),
 		TLSConfig: configTLS(config),
 	}
 	klog.Info(fmt.Sprintf("About to start serving webhooks: %#v", server))


kimwnasptd:
Thank you for the PR @sel!

I'm good with the implementation, but only have some questions to ensure I'm fully in sync with the whole context. 

I see that in the Calico docs you linked above (thanks!) they mention:
> As a workaround, trusted pods that require control plane nodes to connect to them, such as those implementing admission controller webhooks, can include `hostNetwork:true` in their pod spec

So I'm guessing that you've enabled this in the webhook's pod spec, but need to configure the port to avoid collisions, since other processes (or other webhooks) might use the same port. A wild guess would be that you run all the webhooks in a specific node (?)

My questions are
1. Why did you need to run Calico with EKS in the first place?
2. Does this mean that users will need to be able to configure the port of **all** webhooks of their KF installation in order to run Kubeflow on EKS with Calico?

This is more for my understanding of the use case. Again, for the PR, the changes look good so I'll continue with merging this

/lgtm
/approve

sel:
Hello @kimwnasptd and thanks for reviewing the PR.

> 1. Why did you need to run Calico with EKS in the first place?

By default EKS allocates Pod and Service IP addresses (in addition to Node IPs) from the VPC subnets' IP ranges. If those subnet IP ranges are small then it poses a limit on the total number of Nodes, Pods and Services that the cluster can support.  By using an overlay network, such as Calico, we can separate the IP range used for Pods and Services from that used by other AWS services (such as EC2, load-balancing); allowing us to support a larger number of Pods and Services. I should say that Calico does provide other benefits in addition to the separate IP range and there are [other solutions](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) to the EKS IP exhaustion problem.

> 2. Does this mean that users will need to be able to configure the port of **all** webhooks of their KF installation in order to run Kubeflow on EKS with Calico?

Yes, that's correct. I have raised a [PR on kubeflow-pipelines](https://github.com/kubeflow/pipelines/pull/8401/files) and also on other projects in the surrounding ecosystem (TektonCD, KServe). 

I do not know whether this webhook could be built using KNative's [Webhook package](https://github.com/knative/serving/tree/36156c7/vendor/knative.dev/pkg/webhook#knative-webhooks), but if might  reduce some boilerplate and it does provide out-of-the box support for [configuring](https://github.com/knative/serving/blob/36156c7/vendor/knative.dev/pkg/webhook/env.go#L31) the port from the environment.

PR #6701, opened 2022-11-02
Fixed name collisions between notebooks and tensorboards

TobiasGoerke:
Creating tensorboards using the `tensorboards` resource lets the tensorboard-controller create a Deployment, Service and VirtualService named after their parent (creating a tensorboard named `example`, spawns an `example`-service, `example`-virtualservice and `example`-deployment).

Creating a notebook similarly leads to the creation of an `example` service and virtualservice. 

Hence, notebooks and tensorboards name clash and stop working correctly / their controllers reconcile against each other when named equally. [See first mention here](https://github.com/kubeflow/manifests/pull/2304#issuecomment-1299950573).

This PR fixes the issue by adding a prefix to the names / labels of created tensorboad resources.


thesuperzapper:
Closing in preference of https://github.com/kubeflow/kubeflow/pull/7468

PR #6703, opened 2022-11-02
Adding web apps to istio service mesh

### LINKED ISSUE #6702: Protection from fake kubeflow-userid header impersonation
Right now a lot of the components developed in this repo do not have a sidecar.

The biggest issue with this is the fact that we have servers (web app backends) that are assuming user identity from the `kubeflow-userid` header. If we don't limit who and how can talk to these servers then users can impersonate other users.

This is also discussed in https://github.com/kubeflow/manifests/pull/2304

The list of backends that are currently vulnerable to `kubeflow-userid` are:
1. [Jupyter web app](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/manifests/base/deployment.yaml#L10)
2. [Volumes web app](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/volumes/manifests/base/deployment.yaml#L10)
3. [TensorBoards web app](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/tensorboards/manifests/base/deployment.yaml#L10)

To mitigate this we need to ensure that only the `istio-ingressgateway` can talk to `kubeflow-userid` sensitive apps. While we have the AuthorizationPolicies in place to allow only the IGW to talk to the apps, we have the sidecar disabled.

This is a bug. Thank you @juliusvonkohout @TobiasGoerke for exposing this

We should definitely include this in the KF 1.7 release as well
cc @kubeflow/release-managers @kubeflow/release-team

TobiasGoerke:
This PR fixes #6702.

# Summary
The volumes, tensorboards and jupyter web app are vulnerable to user impersonation attacks, as any user can send HTTP requests to these services while freely setting the `kubeflow-userid` header. A malicious user may exploit this to impersonate any other user and create notebooks, tensorboard etc. on their behalf. 
Since the web apps are currently not part of the istio mesh, there is no instance overwriting or validating this header.

# Solution
Remove the disable sidecar annotations and configure istio using authorizationpolicies.

### DIFF
--- a/components/crud-web-apps/jupyter/manifests/base/deployment.yaml
+++ b/components/crud-web-apps/jupyter/manifests/base/deployment.yaml
@@ -5,9 +5,6 @@ metadata:
 spec:
   replicas: 1
   template:
-    metadata:
-      annotations:
-        sidecar.istio.io/inject: "false"
     spec:
       containers:
       - name: jupyter-web-app
--- a/components/crud-web-apps/jupyter/manifests/overlays/istio/authorization-policy.yaml
+++ b/components/crud-web-apps/jupyter/manifests/overlays/istio/authorization-policy.yaml
@@ -1,0 +1,14 @@
+apiVersion: security.istio.io/v1beta1
+kind: AuthorizationPolicy
+metadata:
+  name: jupyter-web-app
+spec:
+  action: ALLOW
+  rules:
+    - from:
+        - source:
+            principals:
+              - cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account
+  selector:
+    matchLabels:
+      app: jupyter-web-app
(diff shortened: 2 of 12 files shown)


kimwnasptd:
@TobiasGoerke the changes look good, thanks!

/lgtm
/approve

kimwnasptd:
Seems that the bot is stuck since there's still the `do-not-merge/work-in-progress`. But I don't understand why this label was added in the first place.

@TobiasGoerke I think if we do the following it should remove the label:
1. Rename the PR title and add a `WIP` at the beginning of the PR
2. Rename back to the original title and remove the WIP

can you try it out? Looking at Prow commands I don't see something that could help us https://prow.k8s.io/command-help

TobiasGoerke:
> Seems that the bot is stuck since there's still the `do-not-merge/work-in-progress`. But I don't understand why this label was added in the first place.
> 
> @TobiasGoerke I think if we do the following it should remove the label:
> 
>     1. Rename the PR title and add a `WIP` at the beginning of the PR
> 
>     2. Rename back to the original title and remove the WIP
> 
> 
> can you try it out? Looking at Prow commands I don't see something that could help us https://prow.k8s.io/command-help

That worked. Thanks!

PR #6722, opened 2022-11-08
tensorboard-controller: Fix tensorboard endless restarts

oshep:
tensorboard-controller: Fix tensorboard endless restarts
tensorboards-web-app deletes objects using foreground deletion policy, Tensorboard CR will stay until all owned objects are deleted reconcile loop might keep on trying to recreate the resources that the API server tries to delete. so when Tensorboard CR is terminating, reconcile loop should do nothing

This commit allow manager to skip reconcile while Tensorboard CR is deleting.

### DIFF
--- a/components/tensorboard-controller/controllers/tensorboard_controller.go
+++ b/components/tensorboard-controller/controllers/tensorboard_controller.go
@@ -80,6 +80,14 @@ func (r *TensorboardReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return reconcile.Result{}, err
 	}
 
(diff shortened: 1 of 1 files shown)


oshep:
Hi there!
I'm faced with the issue that Tensorboards web app stucks on deletion process.
I've investigate the problem and found that child resources are deleted and automatically generated by the controller in a loop. It leads to stucking of deletion process.
The solution is pretty similar like it done for the notebook-controller manager. 
https://github.com/kubeflow/kubeflow/commit/fc09bd1eefb867b4849d8ce2bc56c163b939d2ca

CybervisionVolodymyrPogorelov (approved):

kimwnasptd:
Nice catch, thanks @oshep!

/lgtm
/approve

PR #6706, opened 2022-11-03
jwa: Support for showing Notebooks from all namespaces

kimwnasptd:
Extend the JWA to be able to show Notebooks from all namespaces, when selected from the CentralDashboard. This requires the CentralDashboard work #6674. 

Notable **performance** changes in this PR:
* The frontend will be polling requests for each namespace
* The backend will be using `gevent` workers, for being able to handle a lot of concurrent requests

Notable **UX** changes in this PR:
* When all-namespaces is selected in the dashboard then a new `Namespaces` column will be added
* When all-namespaces is selected the users won't be able to create new notebooks. They'll need to select a namespace first

/cc @tasos-ale 
/cc @elikatsis 

image

PR #6730, opened 2022-11-10
web-apps: Common code for all-namespaces

kimwnasptd:
Follow up of review comments in https://github.com/kubeflow/kubeflow/pull/6706

In this PR I've moved to the common code:
1. Dynamically showing the `Namespace` column, in the common table component, if all-namespaces is selected
2. Common handlers for creating the multiple http calls, for all namespaces
3. The logic for resetting the polling if the data received is different
    * We now always expect that the received objects are K8s objects with `.metadata.name` fields

Lastly I've modified the code of the JWA to use this common code for showing objects from all-namespaces.

/cc @tasos-ale 

if we are OK with these changes I can then send a follow-up PR for adding this functionality to the rest of the web apps

### DIFF
--- a/.github/workflows/frontend_lint.yaml
+++ b/.github/workflows/frontend_lint.yaml
@@ -11,8 +11,8 @@ on:
       - components/crud-web-apps/volumes/frontend/**
 
 jobs:
-  lint:
-    name: Lint frontend code
+  lint-common-code:
+    name: Common code
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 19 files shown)


tasos-ale:
@kimwnasptd the changes look good to me. Can you also fix the linting error so that we can merge this?

kimwnasptd:
@tasos-ale done! I also realized that the current GH Actions were running `ng lint --fix` and not `ng lint`. This means that the actions were just trying to fix the code, but the fixes don't get committed. So we didn't actually check for linting.

I fixe the GH Action and package.json of the apps to have a `lint-check` rule. The code is also properly linted now

PR #6734, opened 2022-11-13
(notebook-controller): Add validation webhook for image name ref

### LINKED ISSUE #6731: ValidationWebhook for Notebooks Controller
/kind feature

The conversation started from https://github.com/kubeflow/kubeflow/pull/6675 but this is a more generic issue.

Right now users can easily create malformed CRs that can lead to invalid Pod definitions.

Kubebuilder offers the option to also develop a ValidationWebhook for our CRs https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html

Ideally I think if we could use K8s' logic for validating the PodSpec then this would be ideal. If not we can have an adhoc approach and start including cases as we bump into them.

cc @mishraprafful

mishraprafful:
Closes #6731 

**Changes Made**
- [x] Add ValidationWebhook for notebook controller
- [x] Add update and create validation for image name using docker reference as done in k8s (https://github.com/kubernetes/kubernetes/blob/43a2bb4df4ff3c255337f5ed927b8fd17c38c235/pkg/kubelet/images/image_manager.go)
- [x] Update manifests to facilitate the installation of webhook


mishraprafful:
pinging @MartinForReal for review 🙂

juliusvonkohout:
@mishraprafful we actually want to rewrite the whole thing https://github.com/kubeflow/kubeflow/issues/7156#issuecomment-1701241970

Can you join the next notebooks WG meeting in two weeks ?  We are always looking for help.

https://www.kubeflow.org/docs/about/community/#kubeflow-community-calendars hast the calendar and meeting notes.

You can also reach us on slack in the channel wg-notebooks

mishraprafful:
@juliusvonkohout Thanks for the headsup, would join the WG meeting and lets discuss how can I help there 🙂 

Do you suggest I should close this PR in that case?

juliusvonkohout:
You can signal your support in https://github.com/kubeflow/kubeflow/issues/7156#issuecomment-1701241970 and comment what you think about it and whether you can help.  The rest can be discussed in the meeting.

juliusvonkohout:
/close

yes let's close. There is also the PR to trim whitespace

Issue #6182, opened 2021-10-11
Jupyter Notebook doesn't launch when gpus is readonly


Leonski13:
/kind bug

When trying to launch a Notebook with `gpus.readonly: true` in my jupyter notebook configmap (see file below) I get an error message in the web console after clicking the `Launch` Button, nothing else happens:

ERROR TypeError: Cannot read properties of undefined (reading 'num') at eR.getSubmitNotebook (main-es2015.32baaf0c0a902cab356b.js:1)

However everything starts up just fine when `gpus` is NOT on readonly mode.

**What did you expect to happen:**
Jupyter Notebook starts up after clicking `Launch`

**Anything else you would like to add:**
This is the patch for my jupyter notebook configmap

``` yaml
- op: replace
  path: /data/spawner_ui_config.yaml
  value: |
    # https://github.com/kubeflow/manifests/blob/master/apps/jupyter/jupyter-web-app/upstream/base/configs/spawner_ui_config.yaml
    # Configuration file for the Jupyter UI.
    #
    # Each Jupyter UI option is configured by two keys: 'value' and 'readOnly'
    # - The 'value' key contains the default value
    # - The 'readOnly' key determines if the option will be available to users
    #
    # If the 'readOnly' key is present and set to 'true', the respective option
    # will be disabled for users and only set by the admin. Also when a
    # Notebook is POSTED to the API if a necessary field is not present then
    # the value from the config will be used.
    #
    # If the 'readOnly' key is missing (defaults to 'false'), the respective option
    # will be available for users to edit.
    #
    # Note that some values can be templated. Such values are the names of the
    # Volumes as well as their StorageClass
    spawnerFormDefaults:
      image:
        # The container Image for the user's Jupyter Notebook
        value: public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-scipy:v1.3.1-rc.0
        # The list of available standard container Images
        options:
          - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-scipy:v1.3.1-rc.0
          - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-full:v1.3.1-rc.0
          - public.ecr.aws/j1r0q0g6/notebooks/notebook-servers/jupyter-pytorch-cuda-full:v1.3.1-rc.0
[...]

PR #6735, opened 2022-11-14
fix setting readOnly on gpus

thesuperzapper:
This PR resolves https://github.com/kubeflow/kubeflow/issues/6182.

Previously, the JWA UI threw an error when when `gpus.readOnly` was true, because of trying to access `num` when `gpus` was undefined in `NotebookFormObject`.

It also fixes the README for the JWA to show the correct build commands for local development.

### DIFF
--- a/components/crud-web-apps/jupyter/README.md
+++ b/components/crud-web-apps/jupyter/README.md
@@ -36,26 +36,34 @@ Requirements:
 
 ### Frontend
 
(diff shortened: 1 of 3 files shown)


thesuperzapper:
@kimwnasptd  This is a very small fix, so I appreciate a review!

/assign @kimwnasptd

PR #6736, opened 2022-11-14
clean up default `spawner_ui_config.yaml`

thesuperzapper:
This PR significantly cleans up (but does not change) the default `spawner_ui_config.yaml` to make it much easier to understand and customize.

---

The only change to default values is that I have removed the default `gpus.value.vendors` list, and left the previous values as a commented example, as most clusters don't have GPUs, so there is no reason to include them by default.

I have also updated the examples for `affinityConfig` and `tolerationGroup` to demonstrate how a user can cause Notebook to be scheduled on exclusive Nodes (no other Notebooks), using the new [`namespaceSelector` for pod-anti-affinity added in K8S 1.22](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#namespace-selector).

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
+++ b/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
@@ -1,184 +1,307 @@
-# Configuration file for the Jupyter UI.
+# --------------------------------------------------------------
+# Configuration file for the Kubeflow Notebooks UI.
 #
-# Each Jupyter UI option is configured by two keys: 'value' and 'readOnly'
-# - The 'value' key contains the default value
-# - The 'readOnly' key determines if the option will be available to users
-#
-# If the 'readOnly' key is present and set to 'true', the respective option
-# will be disabled for users and only set by the admin. Also when a
-# Notebook is POSTED to the API if a necessary field is not present then
-# the value from the config will be used.
-#
-# If the 'readOnly' key is missing (defaults to 'false'), the respective option
-# will be available for users to edit.
-#
-# Note that some values can be templated. Such values are the names of the
-# Volumes as well as their StorageClass
+# About the `readOnly` configs:
+#  - when `readOnly` is set to "true", the respective option
+#    will be disabled for users and only set by the admin
+#  - when 'readOnly' is missing, it defaults to 'false'
+# --------------------------------------------------------------
+
 spawnerFormDefaults:
-  image:
-    # The container Image for the user's Jupyter Notebook
-    value: kubeflownotebookswg/jupyter-scipy:master-6e4ad3b4
-    # The list of available standard container Images
-    options:
-      - kubeflownotebookswg/jupyter-scipy:master-6e4ad3b4
-      - kubeflownotebookswg/jupyter-pytorch-full:master-6e4ad3b4
-      - kubeflownotebookswg/jupyter-pytorch-cuda-full:master-6e4ad3b4
-      - kubeflownotebookswg/jupyter-tensorflow-full:master-6e4ad3b4
-      - kubeflownotebookswg/jupyter-tensorflow-cuda-full:master-6e4ad3b4
-  imageGroupOne:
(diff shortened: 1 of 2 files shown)


thesuperzapper:
I think this is a big quality-of-life improvement so would appreciate a review!

/assign @kimwnasptd

thesuperzapper:
@kimwnasptd can you take a quick look at this, its a formatting change but makes the config much easier to use.

thesuperzapper:
@kimwnasptd can we try and get this merged soon, as I think its a significant improvement in readability for the `spawner_ui_config.yaml`?

thesuperzapper:
@orfeas-k @tasos-ale @elenzio9, as you are quite active now in the web apps, you also might want to review.

kimwnasptd:
@thesuperzapper I'm not really fond of putting all this documentation in the ConfigMap itself. IMO this makes it both difficult to read the docs (i.e. no table of contents) and the file more difficult to read for an admin.

I'd propose to have a separation of concerns. A different place to document the API and a different place for configuring it (yaml). We've also seen multiple requests for more documentation and this can be a good start https://github.com/kubeflow/kubeflow/issues/6914

So I'd propose to instead:
1. Create a PR in the docs and fully expose there the different parts of this ConfigMap and provide examples
2. Keep these YAMLs as minimal as possible

Talador12 (approved):

thesuperzapper:
@kimwnasptd I agree that we can provide improved API docs, especially on the [notebook section](https://www.kubeflow.org/docs/components/notebooks/) of the website (and am happy to help with this) but I think the changes proposed in this PR are still worth having.

The `spawner_ui_config.yaml` file, already has a lot of docstrings & examples, but they lack clarity on what each section actually does, and the file is ordered in a very hard-to-understand way.

This PR simply makes the existing docstrings/examples more clear and re-orders the file.

If you think this PR adds too much detail, can you please comment on the lines which you think are not needed?

thesuperzapper:
@kimwnasptd I really think we should try and get this cleanup merged as it will make the `spawner_ui_config.yaml` better.

Do you think you will have time this week to make a GitHub review so we can get this merged?

You mentioned that you thought some of the docstrings were a bit long, can you point them out.

StefanoFioravanzo:
@thesuperzapper I tend to agree with this

> 1. Create a PR in the docs and fully expose there the different parts of this ConfigMap and provide examples
> 2. Keep these YAMLs as minimal as possible

I suggest adding a new page under the API Reference. In general, it's dangerous to over-document source files, as it results in not exposing that information in the official docs. Anyway, since you've done a lot of great work here, a possible path forward could be:

1. Merge this PR
2. Commit to moving most of the comments to the docs

thesuperzapper:
> I suggest adding a new page under the API Reference. In general, it's dangerous to over-document source files, as it results in not exposing that information in the official docs. Anyway, since you've done a lot of great work here, a possible path forward could be:
> 
> 1. Merge this PR
> 2. Commit to moving most of the comments to the docs

@StefanoFioravanzo I am happy with that plan, if you are too, can you give it an `/approve` so we can move it forward?

Note, this won't make it to Kubeflow 1.7 either way, so it won't affect anyone until Kubeflow 1.8.

Once we have merged, I will look at updating the `kubeflow.org` docs with similar notes as I have added in this PR, probably improving on what the person did in https://github.com/kubeflow/website/pull/3414

StefanoFioravanzo:
@thesuperzapper I don't know the details of how the spawner config works. I'd love for @kimwnasptd to approve the technical review as well

/lgtm

thesuperzapper:
@StefanoFioravanzo do you mind giving this an `/approve` as well? If @kimwnasptd doesn'tt have time to review now.

StefanoFioravanzo:
Sure @thesuperzapper , sorry for the delay, I caught up with my GitHub emails just now. Sure let's go ahead and merge this. Can we also start a discussion to bring this to docs? If you ping me in a PR or issue I am happy to help you out or review

/approve

PR #6739, opened 2022-11-14
(WIP) Use K8s 1.25 for the tests

kimwnasptd:
Updating the KinD cluster version to K8s 1.25


kimwnasptd:
I'll need to pass off this effort to @apo-ger, since I'll need to context switch to start tackling the conformance testing for KF 1.7.

@apo-ger let's start with a new PR with the current work. Then we can start fixing the issues with the common dependency versions

Issue #6460, opened 2022-05-01
Sorting support for the web apps


kimwnasptd:
/kind feature

**Why you need this feature:**
In KF 1.5 we introduced some basic frontend side pagination in #6316. This means that users will no longer see all the Notebooks at once.

We'll need to improve the UX of how people can find their K8s objects faster. A first step for this can be sorting

**Describe the solution you'd like:**
Right now the apps are polling the backend and are fetching for the full list of objects. This means that the frontend always has the full list available. We should also mention that an orthogonal effort here is to make this more efficient, to reduce the load of constantly polling https://github.com/kubeflow/kubeflow/issues/5201 https://github.com/kubeflow/kubeflow/issues/6454

For a first iteration we can do all the sorting in the frontend, since we have the data readily available there

**Anything else you would like to add:**
@elenzio9 and I have started working on this, for which we should be able to send a PR very soon

andreyvelich:
This was resolved.

PR #6744, opened 2022-11-15
web-apps(front): Create unit tests for sorting

elenzio9:
Create unit tests for the sorting functionality (#6460).

Related PRs: https://github.com/kubeflow/kubeflow/pull/6742, https://github.com/kubeflow/kubeflow/pull/6743


kimwnasptd:
We'll also need to prepare some GH Actions to run both the unit and integration tests, in order to ensure they won't break in the future and that we will be running them in each PR.

Let's have a dedicated effort for this

/lgtm
/approve

Issue #6745, opened 2022-11-15
add support of initContainers and sideCars in poddefault


deepk2u:
/kind feature

**Why you need this feature:**
1. we need to run some init script before the main/wait container starts in the pipeline pod
2. we need to add a sidecar container, for example an HMS service running to support hive queries

**Describe the solution you'd like:**
1. update poddefault CRD to include initcontainer and sidecars
2. change the implementation in the controller to mutate pod depending on the availability

PR #6749, opened 2022-11-17
add support of initContainers and sideCars in poddefault

deepk2u:
This PR resolve https://github.com/kubeflow/kubeflow/issues/6745

### DIFF
--- a/components/admission-webhook/examples/poddefault-inject-containers.yaml
+++ b/components/admission-webhook/examples/poddefault-inject-containers.yaml
@@ -1,0 +1,29 @@
+apiVersion: kubeflow.org/v1alpha1
+kind: PodDefault
+metadata:
+  name: inject-containers
+  namespace: kubeflow
+spec:
+  desc: PodDefault to inject initContainers and sidecars
+  initContainers:
+  - command:
+    - echo
+    - kubectl image pulled
+    image: gcr.io/cloud-builders/kubectl
+    imagePullPolicy: Always
+    name: init-kubectl
+  selector:
+    matchLabels:
+      kubeflow.org/inject-containers: "true"
+  sidecars:
+  - env:
+    - name: POD_NAME
+      valueFrom:
+        fieldRef:
+          apiVersion: v1
+          fieldPath: metadata.name
+    command:
+    - echo $(kubectl get pods $POD_NAME -o yaml)
+    image: gcr.io/cloud-builders/kubectl
+    imagePullPolicy: IfNotPresent
+    name: kubectl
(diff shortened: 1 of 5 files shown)


kimwnasptd:
@deepk2u thanks for the PR! I have this in my list of PRs to review and will get to it next week.

From a high level overview, the addition of `initContainers` and `sidecars` looks good. One concern I had is that `sidecars` are only used for appending values. So what about users that might need to use PodDefaults to define an explicit list of containers to enforce in a Pod?

But on a second though would expect a huge minority to be interested in such a feature. So we could look into that in the future once/if we have a more concrete use-case.

deepk2u:
> So what about users that might need to use PodDefaults to define an explicit list of containers to enforce in a Pod?

I think that might be another specific use case, we can implement that if needed

deepk2u:
@kimwnasptd let me know your thoughts after reviewing your PR, we would like to get this feature in the next release, if possible.

kimwnasptd:
Picking this up, thank you for the patience @deepk2u. I'm confident that this will make it in the KF 1.7 release.

As a first step, I'd like us to introduce an `admission-webhook/examples` folder and start placing yaml files there that users could try out. Then we can use this source of truth as context for updating the docs in kubeflow.org with use-cases around PodDefaults.

@deepk2u could you create that folder in this PR and add one/some(?) example PodDefault yamls that will be setting initContainers and extra Containers?

deepk2u:
@kimwnasptd really great idea. I added the examples

kimwnasptd:
Tested the PR and everything works as expected and the test pass. Nice work @deepk2u!

One thing I realized is that we don't have an automated action for running the unit tests for the PodDefaults webhook. @deepk2u would you have cycles to send a separate PR to have a GH Action similar to the Notebook Controller one to run the tests on each PR?
https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/notebook_controller_unit_test.yaml

kimwnasptd:
I'll go forward and merge this PR. @deepk2u the next step will be to update the docs to expose this functionality to end users. There's some other small refactor work in the docs I want to do so I can send the first PR to setup the structure and then I'll ping you, so that we can continue with a docs PR.

Thanks for your time and patience!
/lgtm
/approve

orfeas-k:
Sent a PR with a fix for this bug @kimwnasptd here https://github.com/kubeflow/kubeflow/pull/6903. Note that in order to reproduce the bug I needed to edit the `example` and use `kubeflow-user` instead of `kubeflow` as a `namespace`.

PR #6753, opened 2022-11-21
centraldashboard: Fix how the app is ran locally

orfeas-k:
**Context**

Currently, running the Central dashboard locally and accessing  in-cluster services of K8s fails. With `kubectl proxy` and using [routes ](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/webpack.config.js#L199)like `/api/v1/namespaces/<ns>/*services*/<svc>/proxy`, what we actually do is sending our requests to the K8s service by passing via the K8s API Server (which we proxy with `kubectl proxy`). But this fails, since the K8s API Server is not allowed to talk to the services (Istio). 

**Change**

This PR fixes the proxies and instructions in order to run the app locally. What we propose essentially is to access a Kubenetes service with `kubectl port-forward -n kubeflow svc/<service-name> <service-proxy-port>:<service-port>` e.g. `kubectl port-forward -n kubeflow svc/jupyter-web-app-service 8085:80`. This forwards requests to Kubernetes services from http://localhost:service-proxy-port.

### DIFF
--- a/components/centraldashboard/README.md
+++ b/components/centraldashboard/README.md
@@ -45,8 +45,10 @@ Make sure you have the latest LTS version of `node` installed along with `npm`.
       from the front-end starting with `/api` are proxied to the Express
       server. All other requests are handled by the front-end server which
       mirrors the production configuration.
-4. To access a Kubenetes service, run `kubectl port-forward -n kubeflow svc/<service-name> <service-proxy-port>:<service-port>` e.g. `kubectl port-forward -n kubeflow svc/jupyter-web-app-service 8085:80`.
-   - This forwards requests to Kubernetes services from http://localhost:service-proxy-port. Requests from the front-end starting with `/jupyter`, `/notebook` and `/pipeline` are proxied there. See the [webpack config file](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/webpack.config.js) for more details.
+4. - To access the Jupyter Web App run: `kubectl port-forward -n kubeflow svc/jupyter-web-app-service 8085:80`.
+    - To access Pipeline Web App run: `kubectl port-forward -n kubeflow svc/ml-pipeline-ui 8087:80`.`
+
+    This forwards requests to Kubernetes services from `http://localhost:service-proxy-port`. See the [webpack config file](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/webpack.config.js) for more details.
 
 ### Server Components
 
--- a/components/centraldashboard/README.md
+++ b/components/centraldashboard/README.md
@@ -45,8 +45,8 @@ Make sure you have the latest LTS version of `node` installed along with `npm`.
       from the front-end starting with `/api` are proxied to the Express
       server. All other requests are handled by the front-end server which
       mirrors the production configuration.
-4. To access the Kubenetes REST API, run `kubectl proxy --port=8083`.
-   - This runs [kubectl proxy](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#proxy) to create a reverse proxy at http://localhost:8083. Requests from the front-end starting with `/jupyter`, `/metadata`, `/notebook` and `/pipeline` are proxied to the [Kubenetes API server](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/). See the [webpack config file](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/webpack.config.js) for more details.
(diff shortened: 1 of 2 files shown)


orfeas-k:
@kimwnasptd I 've pushed a commit for this PR but I am not 100% sure about the implementation of proxies at `/notebook` and `/metadata`. I didn't test them locally because I couldn't figure out which service I should port-forward and where to navigate inside the Central dashboard in order to make sure everything's in place.

kimwnasptd:
@orfeas-k looking at
1. The existing K8s services in the `kubeflow` namespace
2. The current ConfigMap that is used in the KF installation https://github.com/arrikto/prv-kubeflow-kubeflow/blob/master/components/centraldashboard/manifests/base/configmap.yaml

I see that the `/metadata` endpoint is not used anymore, and there's no service for it. So we can completely delete it.

kimwnasptd:
Regarding proxying dynamically the Services of Notebooks, it looks like our initial approach of proxying each service by hand is blocking us.

Let's instead keep the current commands that rely on `kubectl proxy` but also give permissions to the `kube-proxy` ServiceAccount to talk to all Services. We could do this by introducing the following AuthorizationPolicy in `/centraldashboard/hack/kube-proxy-istio.yaml`
```yaml
kind: AuthorizationPolicy
metadata:
  name: centraldashboard-proxying
  namespace: istio-system
spec:
  action: ALLOW
  rules:
  - from:
    - source:
        principals: ["cluster.local/ns/kube-system/sa/kube-proxy"]
  - to:
    - operation:
        methods:
        - GET
```

kimwnasptd:
An update on this, after looking into it with @orfeas-k. Turns out we can't use `kubectl proxy` since some of the web apps (jupyter, volumes, tensorboards) have issues with how `kubectl proxy` works.

Specifically `kubectl proxy` will change any `<base href="...">` and `<script src="...">` tags that have **absolute urls** to instead use the URL of the proxy transport:
> // rewriteResponse modifies an HTML response by updating absolute links referring
// to the original host to instead refer to the proxy transport.

https://github.com/kubernetes/kubectl/blob/master/pkg/proxy/proxy_server.go#L214
https://github.com/kubernetes/apimachinery/blob/master/pkg/util/proxy/transport.go#L217-L218

Because of this will continue with the initial plan, of using `kubectl port-forward` for each Service

components/centraldashboard/README.md
```diff
@@ -45,8 +45,8 @@ Make sure you have the latest LTS version of `node` installed along with `npm`.
       from the front-end starting with `/api` are proxied to the Express
       server. All other requests are handled by the front-end server which
       mirrors the production configuration.
-4. To access the Kubenetes REST API, run `kubectl proxy --port=8083`.
-   - This runs [kubectl proxy](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#proxy) to create a reverse proxy at http://localhost:8083. Requests from the front-end starting with `/jupyter`, `/metadata`, `/notebook` and `/pipeline` are proxied to the [Kubenetes API server](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/). See the [webpack config file](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/webpack.config.js) for more details.
+4. To access a Kubenetes service, run `kubectl port-forward -n kubeflow svc/<service-name> <service-proxy-port>:<service-port>` e.g. `kubectl port-forward -n kubeflow svc/jupyter-web-app-service 8085:80`.
```
kimwnasptd, line 48: Could you instead have a list for this numbered step and expose each `kubectl port-forward` command that a user needs to run?

We can just have commands for the Services that we proxy in the webpack config

components/centraldashboard/README.md
```diff
@@ -45,8 +45,8 @@ Make sure you have the latest LTS version of `node` installed along with `npm`.
...
+   - This forwards requests to Kubernetes services from http://localhost:service-proxy-port. Requests from the front-end starting with `/jupyter`, `/notebook` and `/pipeline` are proxied there. See the [webpack config file](https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/webpack.config.js) for more details.
```
kimwnasptd, line 49: Let's add this explanation at the numbered step, since it explains how the proxying works. And, as mentioned above, have the bullet list for the different proxying commands

kimwnasptd (commented):
Thanks @orfeas-k! Some nits about how to structure the README. Rest LGTM

components/centraldashboard/README.md
orfeas-k: Pushed a new commit for both of the comments @kimwnasptd.

PR #6754, opened 2022-11-21
Filtering support for the web apps

elenzio9:
This PR adds the filtering functionality in the web apps. 

Here some screenshots of how it looks like:
image
image

Related issue: https://github.com/kubeflow/kubeflow/issues/6462


elenzio9:
> @elenzio9 I see that the x button is a little bit misaligned. Do you have any insights as to why does this happen?

Yes indeed. I figured it out and it's related to the `@angular/material` version. Currently, our apps are using `12.0.x` while the styles we need (https://github.com/angular/components/blob/cab054a94b2a1348a9478b1f73f51bf8a5fb4018/src/material/chips/chips.scss#L57-L70) were introduced in `12.2.x`. Upgrading the `@angular/material` version of all apps will solve the  misalignment.

kimwnasptd:
I see, thanks! Then lets proceed with this and update the `@angular/material` in all web apps in a follow-up PR

Issue #6000, opened 2021-06-21
Rewrite the Central Dashboard in Angular


thesuperzapper:
/kind feature

Currently, most of our web apps use Angular (CRUD Jupyter/Volumes/Tensorboards), however the Central Dashboard is written in Polymer.

In the interest of consistency and maintainability, we should rewrite the Central Dashboard in Angular.

cc @kubeflow/wg-notebooks-leads

thesuperzapper:
cc @DavidSpek

DavidSpek:
Agree, this is extremely important in terms of maintainability and making it accessible for people to contribute to the central dashboard.

KFServing and Katib web apps are also Angular.

kimwnasptd:
> In the interest of consistency and maintainability, we should rewrite the Central Dashboard in Angular.

@thesuperzapper could you elaborate more on this? The Central Dashboard is fundamentally different from the CRUD web apps, so there's minimal code that could be leveraged from the common code of the CRUD web apps.

I'm very skeptical about even considering to put the effort on such a huge effort, considering the very minimal benefits it would provide.

thesuperzapper:
@kimwnasptd I am aware that CRUD is not similar, but I believe there will be common stuff which can be shared if all the web apps are written in the same framework.

But fundamentally, the main benefits are:
* standardisation - we don't have to use different packages (for things like translation)
* maintainability - we only need to be experts in a single framework, rather than 2
* features - angular is a lot more popular than polymer, so has more options/plugins
* security - we reduce the possible attack surface by using a single framework

I am aware it's going to be a bit of work to do a rewrite, but this work will only get bigger as we start adding features to Central Dashboard.

thesuperzapper:
Reopening this a central location for us to track work on rewriting the central dashboard in Angular.

There was a good discussion in https://github.com/kubeflow/kubeflow/issues/6332 until it was closed, the main takeaway was that we need a design doc (possibly in the form of a [proposal](https://github.com/kubeflow/kubeflow/tree/master/components/proposals)) that outlines the following things:

- UI Mockups 
   - consider how we want to differ from the existing UI
   - for example, we can improve the "collaborators" page a lot
- Config Structure 
    - modularity of sidebar, to make it easier to integrate non-kubeflow apps
    - user-customizable colors/branding
- Overview of Authentication / Authorization    
   - how namespaces are managed, 
   - how auth is propagated to the wrapped/iframed services
- Other
[...]

PR #6765, opened 2022-11-23
centraldashboard: Decouple backend and frontend

orfeas-k:
**Context**: 
This PR is essentially [step 2](https://github.com/kubeflow/kubeflow/issues/6000#issuecomment-1315370307) of the effort of rewriting the central dashboard's frontend using Angular. Currently, the dashboard's code is all coupled together with both backend and frontend files being under the same directory and sharing some key files like `package.json`. 

**Change**
This PR introduces a new folder `centraldashboard-angular` and thus doesn't touch the existing dashboard code. Inside this folder, we introduce a new `backend` code that contains all files required to build and run the backend server to run, as well as run tests and linting checks. 

**Process**
 - Copy `app`, `config` and `types` folder from the old dashboard directory.
 - Create `package.json` file in order to build the backend server and modify scripts accordingly.
 - Introduce rest of the required files.


kimwnasptd:
Really great to see this effort going forward! I won't focus too much on the README.md right now, since I'm expecting us to update this once we have the frontend as well and the dev proxy server via Angular.

/lgtm
/approve

Issue #6707, opened 2022-11-03
Details page for each Notebooks/Volumes/TensorBoards


kimwnasptd:
/kind feature

In other KF web apps like KFP, Models and Katib users are able to navigate to a details page for a specific object and further inspect it.

We should also do the same for the web apps of this repo for Notebooks, Volumes and TensorBoards. This will open the door to adding other functionalities in the future, like:
* https://github.com/kubeflow/kubeflow/issues/5091
* Per-object actions, like https://github.com/kubeflow/kubeflow/issues/5092
* Viewing logs from Notebook/TensorBoard instances

cc @orfeas-k @elenzio9

jerrygb:
Hi @kimwnasptd 

@theofpa had pointed me to this feature issue. 

I wanted to confirm if there is a possibility to allow resources (CPU, Memory, Volumes etc.) updates as well. Is there a specific issue taking into account updates to the notebook? I noticed @elenzio9 had some updates to the New Notebook Form Component.

I can try to help out as well.

Thanks.

andreyvelich:
This should be resolved

PR #6769, opened 2022-11-24
jwa(front): Create distinct notebook details page

orfeas-k:
This PR is part of the effort to add [details pages for objects in our WAs](https://github.com/kubeflow/kubeflow/issues/6707). It creates **a new notebook details page where a user can navigate and see key information about each notebook**. More specifically, this tab contains an overview of the `.status` of the object plus a selection of the most important fields from the spec. These are _Volumes, Applied Configurations from PodDefaults, Notebook type, ENV Vars, Minimum/Maximum CPU/RAM, Docker image, Conditions_. We will also introduce soon `LOGS`, `EVENTS` and `YAML` tab.

### Route
Regarding **the link** that this page will be served on, we need to be able to extract from the link information about the **notebook's name** we 're currently viewing and the **current namespace**.  Thus, we ended up with `notebook/details/namespace/notebook-name?ns=namespace` .

### Notebook's Conditions
Our plan here was to add a Conditions table inside the `OVERVIEW` tab in order to show the `.status.conditions` of the notebook. 

**Problem**
However, we found out that the notebook’s `.status.conditions` are fundamentally broken. Right now the Notebook Controller is blindly appending Pod conditions on the Notebook's conditions https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L247. 
The conditions are expected to provide a summary of the state of an object for a specific point in time, not use the list to combine the conditions to get to the final state. https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

**Workaround**
Since currently, there is no point in exposing the notebook `.status.conditions`, we decided, as a temporary workaround, to show the Notebook’s underlying Pod  conditions and to be explicit about which conditions the table refers to.

### Sidenote
For **clickable items - links**, we followed the style and what we do in general in KFP (e.g. Output Artifacts)

### Screenshots 
image
image
image

cc @kimwnasptd

### DIFF
--- a/components/crud-web-apps/jupyter/frontend/src/app/pages/notebook-page/overview/overview.component.html
+++ b/components/crud-web-apps/jupyter/frontend/src/app/pages/notebook-page/overview/overview.component.html
@@ -1,0 +1,74 @@
+<lib-details-list-item key="Volumes" [chipsList]="volumes">
+</lib-details-list-item>
+
+<lib-details-list-item key="Shared memory enabled">
+  {{ sharedMemory }}
+</lib-details-list-item>
+
+<lib-content-list-item class="configurations" key="Configurations">
+  <app-configurations [configurations]="configurations">
+    {{ podDefaultsMessage }}
+  </app-configurations>
+</lib-content-list-item>
+
+<lib-details-list-item key="Type">
+  {{ notebookType }}
+</lib-details-list-item>
+
+<lib-details-list-item *ngIf="cpuRequests" key="Minimum CPU">
+  {{ cpuRequests }}
+</lib-details-list-item>
+
+<lib-details-list-item *ngIf="cpuLimits" key="Maximum CPU">
+  {{ cpuLimits }}
+</lib-details-list-item>
+
+<lib-details-list-item *ngIf="memoryRequests" key="Minimum memory">
+  {{ memoryRequests }}
+</lib-details-list-item>
+
+<lib-details-list-item *ngIf="memoryLimits" key="Maximum memory">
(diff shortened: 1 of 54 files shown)


components/crud-web-apps/jupyter/frontend/src/app/pages/notebook-page/overview/overview.component.html
```diff
@@ -0,0 +1,74 @@
...
+  <lib-panel class="page-padding">
+    {{ podConditionsMessage }}
+  </lib-panel>
+</ng-container>
+
+<!--Show Pod Conditions-->
+<ng-container *ngIf="podRequestCompleted && pod">
+  <lib-conditions-table
+    *ngIf="pod?.status"
+    [conditions]="pod.status?.conditions"
```
kimwnasptd, line 70: Let's use the conditions from the Notebook status.

We've updated the Notebook Controller to properly set the conditions in https://github.com/kubeflow/kubeflow/pull/6628

kimwnasptd (commented):
Looks great @orfeas-k!

Had some proposals mostly around using Notebook's Conditions and not the Pod ones. We've ensured the Controller is setting the Notebook Conditions properly from the previous release cc @apo-ger

orfeas-k:
@kimwnasptd Had to force push since I rebased to resolve the conflicts but the fixes for the comments above are separate commits on their own.

kimwnasptd:
Changes look good, thanks for driving this @orfeas-k!

/lgtm
/approve

Issue #6684, opened 2022-10-25
Umbrella Issue: Porting Kubeflow to IBM Power (ppc64le)


lehrig:
/kind feature

Enable builds & releases for IBM Power (ppc64le architecture). This proposal was presented [with these slides ](https://github.com/kubeflow/kubeflow/files/9862922/2022-10-25.Porting.Kubeflow.to.IBM.Power.pdf) at the [2022-10-25 Kubeflow community call](https://bit.ly/kf-meeting-notes) with positive community feedback.

**Why you need this feature:**
- Widen scope of possible on-premises deployments (vanilla Kubernetes & OpenShift on Power)
- More general independence regarding processor architecture (x86, ppc64le, arm, …)

**Describe the solution you'd like:**
1. Upstream build & release integration for ppc64le
2. An external, public webpage with ppc64le-specific manifests, install guides, and end-to-end examples provided by IBM

We currently plan to divide 1. into two phases: (1) low-hanging "easy" integrations where no or minor code changes are needed, (2) more complex integrations where external dependencies to python wheels exist. Currently, we believe (1) can be within the Kubeflow 1.7 release scope and help setting a baseline for (2). Below is a detailed overview of each required integration, including links to associated PRs if those already exist.

**Phase 1 Integrations (Kubeflow 1.7 scope)**
- [ ] Admission Webhook: https://github.com/kubeflow/kubeflow/pull/6650
- [ ] Cache
- [ ] Central Dashboard
- [ ] Jupyter Web App: https://github.com/kubeflow/kubeflow/pull/6650
- [ ] Katib
- [ ] KServe: Agent & Controller: https://github.com/kserve/kserve/pull/2476
- [ ] KServer: Models Web App: https://github.com/kserve/models-web-app/pull/45
- [ ] Notebook Controller
- [ ] Profiles + KFAM: https://github.com/kubeflow/kubeflow/pull/6650
- [ ] Tensorboard Controller
- [ ] Tensorboard Web App: https://github.com/kubeflow/kubeflow/pull/6650
- [ ] Training Operator: https://github.com/kubeflow/training-operator/pull/1674
- [ ] Volumes Web App: https://github.com/kubeflow/kubeflow/pull/6650

**Phase 2 Integrations (Post Kubeflow 1.7 scope)**
- [ ] Pipelines
- [ ] KServe: Storage Initializer

**External-facing Webpage (Post Kubeflow 1.7 scope)**
- [ ] Manifests
[...]

PR #6771, opened 2022-11-25
Adding multi-arch support for linux-ppc64le in CI for notebook-controller

adilhusain-s:
* the  Build & Publish Notebook Controller  github workflow only releases docker image for linux-amd64 
*  Adding linux-ppc64le arch in workflow to release the docker image for ppc64le arch 
*  Using docker buildx and qemu to build the multi-arch docker images

### DIFF
--- a/.github/workflows/nb_controller_docker_publish.yaml
+++ b/.github/workflows/nb_controller_docker_publish.yaml
@@ -15,30 +15,41 @@ on:
       - components/notebook-controller/**
       - components/common/**
 
+env:
+  DOCKER_USER: kubeflownotebookswg
+  IMG: kubeflownotebookswg/notebook-controller
+  ARCH: linux/ppc64le,linux/amd64
+
 jobs:
   push_to_registry:
     name: Build & Push Docker image to Docker Hub
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout
-      uses: actions/checkout@v3
+      - name: Checkout
+        uses: actions/checkout@v3
 
-    - name: Login to DockerHub
-      if: github.event_name == 'push'
-      uses: docker/login-action@v2
-      with:
-        username: kubeflownotebookswg
-        password: ${{ secrets.KUBEFLOWNOTEBOOKSWG_DOCKER_TOKEN }}
+      - name: Login to DockerHub
+        if: github.event_name == 'push'
+        uses: docker/login-action@v2
+        with:
+          username: ${{ env.DOCKER_USER }}
+          password: ${{ secrets.KUBEFLOWNOTEBOOKSWG_DOCKER_TOKEN }}
 
-    - name: Run Notebook Controller build
-      run: |
-        cd components/notebook-controller
-        export IMG=kubeflownotebookswg/notebook-controller
-        make docker-build
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v2
+        with:
+          platforms: ppc64le
 
-    - name: Run Notebook Controller push
-      if: github.event_name == 'push'
-      run: |
-        cd components/notebook-controller
-        export IMG=kubeflownotebookswg/notebook-controller
-        make docker-push
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Build multi-arch docker image
+        run: |
+          cd components/notebook-controller
+          make docker-build-multi-arch
+
+      - name: Build  and push multi-arch docker image
+        if: github.event_name == 'push'
+        run: |
+          cd components/notebook-controller
+          make docker-build-push-multi-arch
--- a/components/notebook-controller/Makefile
+++ b/components/notebook-controller/Makefile
@@ -1,6 +1,7 @@
 # Image URL to use all building/pushing image targets
 IMG ?= notebook-controller
 TAG ?= $(shell git describe --tags --always --dirty)
+ARCH ?= linux/amd64
 
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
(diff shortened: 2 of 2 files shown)


kimwnasptd (commented):
Thanks for the PR and continuing the effort for #6684 @adilhusain-s!

The main comments I have are around keeping our build logic into the Makefiles. I didn't know about the `env` in GH Actions, which seems very useful.

kimwnasptd:
I see that the review comments were not added by order of lines in the file. I'd suggest to take a look at them in the order we see in https://github.com/kubeflow/kubeflow/pull/6771/files

Also, you'll need to sign the CLA

kimwnasptd:
Lastly, could you add a dummy file in the notebook controller directory? This will help us verify that the actions work as expected

adilhusain-s:
@kimwnasptd 

Thanks for the review,
* Remove workflow_dispatch  trigger as suggested (was using it on my fork repo to trigger the workflow manually)
* Introduce new build-multi-arch and build-push-multi-arch make rules 
* Dropped github action docker build-push
* Using make rule instead to push the docker images
* introduced ARCH var in workflow and Makefile 
* Setting ARCH  to linux/amd64 only in Makefile  (buildx requires qemu or docker nodes to build the foreign arch and this may not be available locally)
* The developer/user should set up the  qemu and docker buildx docker driver and so on and this needs to be documented somewhere.
* Setting ARCH to linux/ppc64le,linux/amd64 (buildx and qemu github action allows us to emulate the foreign arch)

components/notebook-controller/Makefile
```diff
@@ -85,6 +86,16 @@ docker-build: ## Build docker image with the manager.
...
 	docker push ${IMG}:${TAG}
 
+.PHONY: docker-build-multi-arch
+docker-build-multi-arch: ##  Build multi-arch docker images with docker buildx
+	cd .. && docker buildx build --platform ${ARCH} --tag ${IMG}:${TAG} -f ./notebook-controller/Dockerfile .
+
+
+.PHONY: docker-build-push-multi-arch
+docker-build-push-multi-arch: ## Build multi-arch docker images with docker buildx and push to docker registry 
+	cd .. && docker buildx build --platform ${ARCH} --tag ${IMG}:${TAG} --push -f ./notebook-controller/Dockerfile .
```
kimwnasptd, line 96: Why do we need this rule? Can't we just use the `docker-push` action for pushing the image, with all the platform manifests, at once?

kimwnasptd (commented):
@adilhusain-s added some comments. Mostly regarding using one step for building for all architectures and one for pushing.

The steps in the workflow I had in mind are the following:
1. Checkout
2. Login to DockerHub
3. Setup QEMU
4. Setup Docker Buildx
6. Build multi-arch docker image
7. Push OCI image (previously `Run Notebook Controller push`), conditionally `if: github.event_name == 'push'`

.github/workflows/nb_controller_docker_publish.yaml
```diff
@@ -15,30 +15,52 @@ on:
...
+      - name: Run Notebook Controller build
+        run: |
+          cd components/notebook-controller
+          make docker-build
+
+      - name: Run Notebook Controller push
+        if: github.event_name == 'push'
+        run: |
+          cd components/notebook-controller
+          make docker-push
```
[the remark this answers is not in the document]
adilhusain-s: * will drop the docker-build step as it's not required anymore 
*  Docker-push needs to be dropped as it can't be used for pushing multi-arch docker images
*  the "build-multi-arch" rule builds multi-arch  docker images and stores it in build cache 
*the built docker images aren't accessible using "docker image " command, so the built docker images can't be tagged or pushed.

components/notebook-controller/Makefile
adilhusain-s: could you please provide the link of github action?
if you're referring to "docker/build-push-action", this can't be used with current make "build-multi-arch" rule.

the action triggers the build with different build ID and meta data  and then pushes the image and this will increase the runtime of the workflow.

.github/workflows/nb_controller_docker_publish.yaml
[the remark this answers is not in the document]
kimwnasptd: I see, thanks for the context! Let's then keep the two steps you had at the end for building, and then building-pushing the image. And drop the `Run Notebook Controller build` and `Run Notebook Controller push` steps

components/notebook-controller/Makefile
kimwnasptd: > could you please provide the link of github action?

My bad, I meant to say rule (in makefile).

Just took a more detailed look on https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/ that explains how to push the manifests for different platforms for an image. So indeed let's keep the two steps you have now for building and building-pushing (will use cached image from previous step)

kimwnasptd:
So @adilhusain-s my current understanding is that the actions should look like this:
1. Checkout
2. Login to DockerHub
3. Setup QEMU (all platforms)
4. Setup Docker Buildx
5. Build multi-arch docker image
6. Build  and push multi-arch docker image

Are we in sync or am I missing something?

adilhusain-s:
@kimwnasptd 
here is the [workflow run](https://github.com/adilhusain-s/kubeflow/actions/runs/3601923650) from my fork.

kimwnasptd:
Nice work on also working on plain docker manifests! But indeed just using buildx with --push seems way more simple so let's just go with this.

Yes, the workflows from https://github.com/kubeflow/kubeflow/pull/6785 looks good.

So let's update this PR to look like that as well and we can merge

PR #6781, opened 2022-11-29
centraldashboard-angular: Create Docker and Make files

orfeas-k:
This is part of the [Rewrite the Central Dashboard in Angular](https://github.com/kubeflow/kubeflow/issues/6000) effort and creates Docker and Make files for building the new centraldashboard's container.

### DIFF
--- a/components/centraldashboard-angular/Makefile
+++ b/components/centraldashboard-angular/Makefile
@@ -1,0 +1,14 @@
+IMG ?= centraldashboard-angular
+TAG := $(shell git describe --tags --always --dirty)
+COMMIT = $(shell git rev-parse HEAD)
+
+# To build without the cache set the environment variable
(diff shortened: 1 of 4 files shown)


components/centraldashboard-angular/Makefile
```diff
@@ -0,0 +1,14 @@
+IMG ?= centraldashboard
```
kimwnasptd, line 1: nit: Let's use `centraldashboard-angular` here as well, to make sure we use the same value everywhere.

We can revert to `centraldashboard` once we are confident with switching the web apps

orfeas-k: @kimwnasptd Fixed it and pushed again!

kimwnasptd:
Looks good. Weird that all checks are pending 🤔

/lgtm
/approve

PR #6783, opened 2022-11-29
feat: make TensorBoard image configurable in CRD

dvaldivia:
This PR makes the TensorBoard image, pull strategy and pull secrets configurable by introducing three new types on the TensorBoard CRD.

The idea behind this is that some of the kubeflow deployments may need to use a private docker repositories such as in the case of airgap deployments.

Signed-off-by: Daniel Valdivia <hola@danielvaldivia.com>


dvaldivia:
@elikatsis @kandrio98 @kimwnasptd please let me know if you have any questions

juliusvonkohout:

PR #6786, opened 2022-11-30
gh-actions: Add GH actions for centraldashboard-angular

orfeas-k:
This is part of the [Rewrite the Central Dashboard in Angular](https://github.com/kubeflow/kubeflow/issues/6000) effort and creates GH action for the new CentralDashboard-angular component. More specifically, this adds the following GH actions:
 - CentralDashboard-angular Backend check that runs:
    - unit tests
    - TSlint
 - CentralDashboard-angular Frontend check that runs:
    - unit tests
    - format check
 
and also modifies Frontend Lint GH action to lint CentralDashboard-angular frontend as well.


kimwnasptd:
@orfeas-k I see that this is in WIP state. Is this the case or is it ready for review?

orfeas-k:
@kimwnasptd This is ready to be merged.

components/centraldashboard-angular/frontend/package.json
```diff
@@ -6,7 +6,12 @@
     "start": "ng serve --proxy-config proxy.conf.json",
     "build": "ng build",
     "watch": "ng build --watch --configuration development",
-    "test": "ng test"
+    "test": "ng test",
+    "test:prod": "ng test --no-watch --browsers ChromeHeadlessNoSandbox",
+    "tslint": "tslint -c tslint.json -p tsconfig.json",
```
kimwnasptd, line 11: Why do we introduce rules for `tslint`, which is getting deprecated?

Let's instead have the following 2 rules:
```json
    "lint-check": "ng lint",
    "lint": "ng lint --fix",
```
similarly to KWA https://github.com/kubeflow/katib/blob/master/pkg/new-ui/v1beta1/frontend/package.json#L12-L13

We might also need to make some small changes to the `angular.json`. Take a look at the previous effort for this in https://github.com/kubeflow/kubeflow/pull/6464

components/centraldashboard-angular/frontend/package.json
```diff
@@ -36,6 +41,8 @@
     "karma-coverage": "~2.0.3",
     "karma-jasmine": "~4.0.0",
     "karma-jasmine-html-reporter": "~1.7.0",
+    "prettier": "^2.3.1",
+    "tslint": "^6.1.3",
```
kimwnasptd, line 45: similar to the comment above, let's move away from `tslint`

.github/workflows/centraldb_angular_backend_check.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 12
+      - name: Run unit tests
+        run: |
+          cd components/centraldashboard-angular/backend/
+          npm i
+          npm run test
+
+  run-backend-tslint:
```
kimwnasptd, line 22: We should look into moving away from `tslint` in the backend as well, now that we've decoupled the frontend and backend code

.github/workflows/centraldb_angular_backend_check.yaml
```diff
@@ -0,0 +1,34 @@
+name: CentralDashboard-angular Backend check
```
kimwnasptd, line 1: Let's rename this to `CentralDashboard-angular Backend Tests` as well as the file to `centraldb_angular_backend_test.yaml`

.github/workflows/centraldb_angular_frontend_check.yaml
```diff
@@ -0,0 +1,34 @@
+name: CentralDashboard-angular Frontend check
```
kimwnasptd, line 1: Let's rename this to `CentralDashboard-angular Frontend Tests` and also the file to `centraldb_angular_frontend_test.yaml`

.github/workflows/frontend_lint.yaml
```diff
@@ -66,3 +67,17 @@ jobs:
         cd components/crud-web-apps/volumes/frontend
         npm i
         npm run lint-check
+
+  lint-centraldashboard-angular-code:
```
kimwnasptd, line 71: I'd propose to drop the linting checks for the dashboard, until we fully migrate to eslint

kimwnasptd (commented):
@orfeas-k my main points are around not using `tslint` and only going for `eslint`, at least for the frontend that is new code. 

We've done this in the past for the rest of the apps in https://github.com/kubeflow/kubeflow/pull/6464. But I think this would be a bigger effort so we can have a dedicated PR for this.

I'd suggest to just remove `tslint` specific commands for the frontend tests for now, and then have a follow up PR for using `eslint` in the new frontend code.

WDYT?

orfeas-k:
@kimwnasptd Pushed two commits that:
 - Fix the names of the GH actions
 - Introduce ESLint in the centraldashboard frontend code

Thus, I kept the linting rule in `frontend_lint.yaml` since we 're using ESLint now. Only thing that's left to do (in a separate effort from what I understand) is Migrate TSLint to ESLint in centraldashboard's backend code.

kimwnasptd:
@orfeas-k looks good! Indeed let's have the switch to ESlint for the backend as a dedicated effort.

/lgtm
/approve

PR #6799, opened 2022-12-04
Extend tests for using images of each PR

### LINKED ISSUE #6766: Extend tests for using images of each PR
/kind feature

Right now we have tests for just applying the manifests in a KinD cluster for each component, i.e. https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/jwa_kind_test.yaml

But when the actions are run for a PR they don't actually test the images that would be built from the current PR, but rather the ones already committed. We should instead aim at testing manifests by using images that are built from the code changes of each PR's commits.

This will also allow us to start doing more advanced integration tests, like:
* Running API tests on the deployed web apps
* Test that the Controllers can get objects to the desired state

And of course we will improve our testing quality by ensuring each PR results in images that will be working.

The end result I have in mind is to have
1. A distinct workflow file for publishing the docker images, only when a PR is merged
2. A distinct workflow file that will be testing on each PR commit that
    1. The docker image can be built
    2. It will spin up a KinD cluster and insert the image we built https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster
    3. Do a simple sed and update the manifests to use that image
    4. Build and apply the manifests
    5. Wait for the Pods to become Ready

For the first task we'll need to update the existing `*_docker_publish.yaml` workflows to only get triggered when a PR is merged 
https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/jwa_docker_publish.yaml

For the second task we'll need to extend the `*_kind_test.yaml` files to do the above behavior.

Regarding testing that Pods are ready, we've also done similar work on the Manifests repo https://github.com/kubeflow/manifests/blob/master/.github/workflows/jwa_kind_test.yaml#L31

salilmishra23:
This PR resolves #6766

### DIFF
--- a/.github/workflows/all_components_docker_publish.yaml
+++ b/.github/workflows/all_components_docker_publish.yaml
@@ -1,15 +1,18 @@
 name: Build & Publish all Docker images
 on:
-  push:
+  pull_request:
     branches:
       - master
       - v*-branch
     paths:
       - releasing/version/VERSION
+    types:
+      - closed
 
 jobs:
   push_to_registry:
     name: Build & Push all Docker images to Docker Hub
+    if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
--- a/.github/workflows/centraldb_docker_publish.yaml
+++ b/.github/workflows/centraldb_docker_publish.yaml
@@ -1,28 +1,24 @@
 name: Build & Publish CentralDashboard Docker image
 on:
-  push:
-    branches:
-      - master
-      - v*-branch
-    paths:
-      - components/centraldashboard/**
   pull_request:
     branches:
       - master
       - v*-branch
     paths:
       - components/centraldashboard/**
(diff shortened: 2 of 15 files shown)


kimwnasptd:
Thanks for picking this up @salilmishra23! I see this is still a draft, but I'll provide some high level comments now that we are at the start of this effort.

I'd suggest to limit this PR to the Notebook Controller, and specifically:
1. Let's create a new `.github/nb_controller_integration_test.yaml`, which in the future will replace these workflows
    * https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/nb_controller_kind_test.yaml
    * https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/nb_controller_manifests_test.yaml
2. Do the second case discussed in the proposal https://github.com/kubeflow/kubeflow/issues/6766, and focus on building the image, creating the kind cluster etc

PR #6803, opened 2022-12-05
Adding support for linux-ppc64le in CI for admission-webhook multi-arch docker image

adilhusain-s:
• the workflow "Build & Publish PodDefaults Docker image          " releases docker image for linux-amd64
• Added docker-build-multi-arch and build-push-multi-arch targets in Makefile of admission-webhook component
• Dropping docker-build and docker-push from workflow
• Keeping these steps will increase the workflow runtime
•  "build-multi-arch" make rule builds the multi-arch docker image  and stores it in to build cache 
• docker-build-push-multi-arch  uses build cache and builds and pushes the images for linux-amd64 and linux-ppc64le

please review and merge.

### DIFF
--- a/.github/workflows/poddefaults_docker_publish.yaml
+++ b/.github/workflows/poddefaults_docker_publish.yaml
@@ -13,30 +13,39 @@ on:
     paths:
       - components/admission-webhook/**
 
+env:
+  DOCKER_USER: kubeflownotebookswg
+  IMG: kubeflownotebookswg/poddefaults-webhook
+  ARCH: linux/ppc64le,linux/amd64
+
 jobs:
   push_to_registry:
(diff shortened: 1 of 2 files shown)


.github/workflows/poddefaults_docker_publish.yaml
```diff
@@ -13,30 +13,39 @@ on:
     paths:
       - components/admission-webhook/**
 
+env:
+  DOCKER_USER: kubeflownotebookswg
+  IMG: kubeflownotebookswg/admission-webhook
```
kimwnasptd, line 18: @adilhusain-s could you change the image here to `kubeflownotebookswg/poddefaults-webhook`? This is the image we  were using

PR #6805, opened 2022-12-05
Adding support for linux/ppc64le in CI for tensorboard-controller multi-arch docker images.

amitmukati-2604:
- The workflow "TB_Controller_Docker_Publish" releases docker image for Linux-amd64.
- Added docker-build-multi-arch and build-push-multi-arch targets in the Makefile of the Tensorboard-controller component.
- Dropping docker-build and docker-push from the workflow.
- Keeping these steps will increase the workflow runtime.
- Docker-build-push-multi-arch builds and pushes the images for Linux-amd64 and Linux-ppc64le
- Below are the link to the workflow running from my fork and the published images.
     Build Result Link:- https://github.com/amitmukati-2604/kubeflow/actions/runs/3619078575
     Publish image Link:- https://hub.docker.com/r/amitmukati2604/tensorboard-controller/tags

### DIFF
--- a/.github/workflows/tb_controller_docker_publish.yaml
+++ b/.github/workflows/tb_controller_docker_publish.yaml
@@ -13,6 +13,11 @@ on:
     paths:
       - components/tensorboard-controller/**
 
+env:
+  DOCKER_USER: kubeflownotebookswg
+  IMG: kubeflownotebookswg/tensorboard-controller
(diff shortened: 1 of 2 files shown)


amitmukati-2604:
@kimwnasptd 

Thanks for the Review.

- Removed the DUMMY_FILE as suggested.

Issue #6767, opened 2022-11-24
A lot of K8s updates due to Notebook last_activity annotations


kimwnasptd:
/kind bug

**What steps did you take and what happened:**
Right now the Notebooks Controller is periodically querying each Notebook to update the `notebooks.kubeflow.org/last_activity` annotation on each Notebook https://github.com/kubeflow/kubeflow/blob/master/components/proposals/20220121-jupyter-notebook-idleness.md

But the problem is that right now the Controller will both make the http requests and update the CR in each reconciliation loop. This adds a lot of overhead to the platform.

**What did you expect to happen:**
I would expect the Controller logic to respect the `IDLENESS_CHECK_PERIOD` var and not make the requests on each reconciliation.

To do this we should introduce another annotation in each Notebook CR `notebooks.kubeflow.org/last_activity_check_timestamp`. The Controller will be setting the timestamp that it last checked for idleness of a Notebook. This way the Controller can compare this timestamp with the current time and decide NOT to check for idleness if the `IDLENESS_CHECK_PERIOD` has not passed.

**Environment:**
- Kubeflow version: `v1.6` and current `master`

/help

thesuperzapper:
@kimwnasptd I thought that this [`ctrl.Result{RequeueAfter: culler.GetRequeueTime()}`](https://github.com/kubeflow/kubeflow/blob/1610b25e3ae6c40b230c64c77f52425c41e4ea2a/components/notebook-controller/controllers/notebook_controller.go#L281) (which calls [`culler.GetRequeueTime()`](https://github.com/kubeflow/kubeflow/blob/1610b25e3ae6c40b230c64c77f52425c41e4ea2a/components/notebook-controller/pkg/culler/culler.go#L81-L95)) would mean that the pod is not reconsiled more frequently than `IDLENESS_CHECK_PERIOD`?

apo-ger:
Hi @kimwnasptd, I would like to work on this!

apo-ger:
[...]

PR #6807, opened 2022-12-06
Implement a culling controller for Notebooks

apo-ger:
This PR splits the idleness/culling logic into a separate controller inside of the Notebooks Controller/Operator.

The culling-controller will:
* reconcile only notebooks CRs.
* set/update culling annotations
  - `notebooks.kubeflow.org/last_activity`
  - `notebooks.kubeflow.org/last_activity_check_timestamp`
* perform idleness checks every `IDLENESS_CHECK_PERIOD` minutes and set the `kubeflow-resource-stopped` annotation, if a notebook needs to be culled.

Refs: kubeflow/kubeflow#6767


components/notebook-controller/controllers/culling_controller.go
```diff
@@ -0,0 +1,451 @@
...
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Ensure that the underlying Notebook Pod exists
+	foundPod := &corev1.Pod{}
+	err = r.Get(ctx, types.NamespacedName{Name: instance.Name + "-0", Namespace: instance.Namespace}, foundPod)
+	if apierrs.IsNotFound(err) {
+		log.Info("Pod not found...Won't check for culling...")
+		return ctrl.Result{}, nil
```
kimwnasptd, line 110: let's also remove the annotations, if they exist from the CR, which was part of the previous implementation
https://github.com/kubeflow/kubeflow/blob/master/components/proposals/20220121-jupyter-notebook-idleness.md#implementation-details
https://github.com/kubeflow/kubeflow/blob/master/components/notebook-controller/controllers/notebook_controller.go#L227-L250

I'm mostly thinking the scenario where the cluster is scaled down and the notebook ends up loosing its Pod (but the node of the controller is not down yet, so the controller keep running).

in this case the controller won't delete the annotation. And once the cluster is scaled back up again then the notebook will start, but the controller will stop it since it will check the previous timestamp (and there's no activity once the pod starts)

components/notebook-controller/controllers/culling_controller.go
```diff
@@ -0,0 +1,451 @@
...
+type CullingReconciler struct {
+	client.Client
+	Log     logr.Logger
+	Scheme  *runtime.Scheme
+	Metrics *metrics.Metrics
+}
+
+func (r *CullingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("culler", req.NamespacedName)
+
```
kimwnasptd, line 80: nit: Let's add a first log at the beginning of the reconciliation loop to make it clear that a `Reconcile` was called for a notebook

kimwnasptd (commented):
@apo-ger left some comments. The code mostly LGTM!

I also tested the controller and indeed it's making the requests by respecting the actual check period. Nice work!

components/notebook-controller/controllers/culling_controller.go
apo-ger: Sure! I modified the culling-controller to remove the `culling annotation` when the underlying Pod doesn't exist, since:
1. The culling-controller is now responsible for setting this annotation
2. The culling-controller is also checking if the underlying Pod exists

components/notebook-controller/controllers/culling_controller.go
apo-ger: Done!

kimwnasptd:
Thanks for this work @apo-ger!

/lgtm
/approve

apo-ger:
@kimwnasptd nice! Thanks you for the review!

shalberd:
Very good code using metav1.GetAnnotation returning either true or false, instead of checking for the presence of an annotation with an own logic block. I will integrate that in [my PR](https://github.com/kubeflow/kubeflow/pull/6904) as well, cool.

PR #6820, opened 2022-12-07
profile-controller: Extend tests for using images of each PR

apo-ger:

Changes: 
-  Publish the docker image, only when a [PR is merged](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges). 
-  Remove  kind and manifests GH action tests
-  Introduce a GH action intergration test to ensure that on each PR commit:
   - The docker image can be built
   - The manifests can be built
   - The manifests with the updated image can be applied in a KinD cluster
     - Spin up a KinD cluster
     - Load the image in KinD cluster https://kind.sigs.k8s.io/docs/user/quick-start/#loading-an-image-into-your-cluster
     - Use sed to update the image before applying manifests
   - The Pods become Ready


kimwnasptd:
Thanks for the PR @apo-ger! I realized that the `master` branch needs the following two PRs from the release branch that we never cherry-picked 
* #6548
* https://github.com/kubeflow/kubeflow/pull/6591
* https://github.com/kubeflow/kubeflow/pull/6597
* https://github.com/kubeflow/kubeflow/pull/6632

To ensure that we are using the DockerHub registry and that we when the `releasing/version/VERSION` file changes then each public-workflow will also push an image with the tag defined.

We'll need this because I'd like us to use the `latest` tag in the images of `master`. But we need to import this work first

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+    - name: Install Istio
+      run: ./components/testing/gh-actions/install_istio.sh
+
+    - name: Build & Apply manifests
+      run: |
+        cd components/profile-controller/config
+        kubectl create ns kubeflow
+
+        export CURRENT_IMG=public.ecr.aws/j1r0q0g6/notebooks/profile-controller:v1.5.0
```
kimwnasptd, line 48: let's switch to the dockerhub one

.github/workflows/prof_controller_docker_publish.yaml
```diff
@@ -1,51 +1,49 @@
...
 
-      - name: Build  and push multi-arch docker image
-        if: github.event_name == 'push'
-        run: |
-          cd components/profile-controller
-          make docker-build-push-multi-arch
+        - name: Build  and push multi-arch docker image
+          run: |
+            cd components/profile-controller
+            export IMG=kubeflownotebookswg/profile-controller
```
kimwnasptd, line 48: Same here, regarding not re-declaring the IMG

kimwnasptd (commented):
Left some minor comments. Looks good in general!

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Build Profile Controller Image 
+      run: |
+        cd components/profile-controller
+        make docker-build IMG=${{env.IMG}} TAG=${{env.TAG}}
+
+    - name: Install KinD
+      run: ./components/testing/gh-actions/install_kind.sh
+
+    - name: Create KinD Cluster
+      run: kind create cluster --config components/testing/gh-actions/kind-1-25.yaml --name prof-controller
```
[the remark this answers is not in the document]
apo-ger: No really, I'll remove it!

.github/workflows/prof_controller_docker_publish.yaml
apo-ger: fixed

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,53 @@
...
+  TAG: intergration-test
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+      with:
+          ref: ${{ github.event.pull_request.head.sha }}
```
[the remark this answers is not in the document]
apo-ger: We don't really need it, I just added to run the workflows from the head commit of the PR. I'll remove it

.github/workflows/prof_controller_intergration_test.yaml
apo-ger: Done!

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - name: Install Istio
+      run: ./components/testing/gh-actions/install_istio.sh
+
+    - name: Build & Apply manifests
+      run: |
+        cd components/profile-controller/config
+        kubectl create ns kubeflow
+
+        export CURRENT_IMG=public.ecr.aws/j1r0q0g6/notebooks/profile-controller:v1.5.0
+        export PR_IMG=profile-controller:intergration-test
```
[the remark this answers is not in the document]
apo-ger: Fixed!

apo-ger:
@kimwnasptd The tests were failing after  https://github.com/kubeflow/kubeflow/pull/6825 was merged as the manifests were looking for the `docker.io/kubeflownotebookswg/kfam:v1.5.0` image which doesn't exist. This is expected since these changes were initially introduced in `release-v1.6` branch.

I've updated the PR to use `v1.6.0` to unblock for now. But this will be resolved once and for all when we update the tag to `latest` in manifests/workflows as mentioned in https://github.com/kubeflow/kubeflow/issues/6766#issuecomment-1342769182

I'd argue that we should also trigger the profile-controller intergration-test workflow when `kfam` code changes. I've refactored the Intergration test for this as well.

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: Build Profile Controller Image 
+      run: |
+        cd components/profile-controller
+        make docker-build IMG=${{env.PROFILE_IMG}}
+
+    - name: Build Kfam Image
+      run: |
+        cd components/access-management
+        make docker-build IMG=${{env.KFAM_IMG}}
```
kimwnasptd, line 31: Let's avoid building the `kfam` image here. Since in the next iterations https://github.com/kubeflow/kubeflow/issues/6766#issuecomment-1342769182 we'll be using the `latest` tag we should be able to use that.

Also currently we use the `v1.6.0` tag for kfam, so we don't need to build the image

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Install KinD
+      run: ./components/testing/gh-actions/install_kind.sh
+
+    - name: Create KinD Cluster
+      run: kind create cluster --config components/testing/gh-actions/kind-1-25.yaml
+
+    - name: Load Images into KinD Cluster 
+      run: |
+        kind load docker-image ${{env.PROFILE_IMG}}:${{env.TAG}}
+        kind load docker-image ${{env.KFAM_IMG}}:${{env.TAG}}
```
kimwnasptd, line 42: we can remove this one as well

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+    - name: Build & Apply manifests
+      run: |
+        cd components/profile-controller/config
+        kubectl create ns kubeflow
+
+        export CURRENT_PROFILE_IMG=docker.io/kubeflownotebookswg/profile-controller:v1.6.0
+        export CURRENT_KFAM_IMG=docker.io/kubeflownotebookswg/kfam:v1.6.0
+        export PR_PROFILE_IMG=${{env.PROFILE_IMG}}:${{env.TAG}}
+        export PR_KFAM_IMG=${{env.KFAM_IMG}}:${{env.TAG}}
```
kimwnasptd, line 58: Let's only update the profile image. We'll just use the KFAM image currently in the manifests, which will become `latest` in the future

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+    paths:
+      - components/profile-controller/**
+      - components/access-management/**
+    branches:
+      - master
+      - v*-branch
+
+env:
+  PROFILE_IMG: profile-controller
+  KFAM_IMG: kfam
```
kimwnasptd, line 13: I think we can remove this one as well

.github/workflows/prof_controller_intergration_test.yaml
```diff
@@ -0,0 +1,62 @@
...
+      run: |
+        cd components/profile-controller/config
+        kubectl create ns kubeflow
+
+        export CURRENT_PROFILE_IMG=docker.io/kubeflownotebookswg/profile-controller:v1.6.0
+        export CURRENT_KFAM_IMG=docker.io/kubeflownotebookswg/kfam:v1.6.0
+        export PR_PROFILE_IMG=${{env.PROFILE_IMG}}:${{env.TAG}}
+        export PR_KFAM_IMG=${{env.KFAM_IMG}}:${{env.TAG}}
+
+        kustomize build overlays/kubeflow | sed "s#$CURRENT_PROFILE_IMG#$PR_PROFILE_IMG#g" | sed "s#$CURRENT_KFAM_IMG#$PR_KFAM_IMG#g" | kubectl apply -f -
```
kimwnasptd, line 60: And lastly, same here. Let's just update the profile controller image for now

.github/workflows/prof_controller_docker_publish.yaml
```diff
@@ -1,51 +1,47 @@
...
 
-      - name: Build  and push multi-arch docker image
-        if: github.event_name == 'push'
-        run: |
-          cd components/profile-controller
-          make docker-build-push-multi-arch
+        - name: Build  and push multi-arch docker image
+          run: |
+            cd components/profile-controller
+            make docker-build-push-multi-arch
```
kimwnasptd, line 47: I see that GH shows there's a slight change in the indentation. Could you update this one as well?

kimwnasptd (commented):
Sounds good @apo-ger! Left some more comments towards that direction

.github/workflows/prof_controller_docker_publish.yaml
apo-ger: fixed!

kimwnasptd:
So, the ideal state I have in mind is:
1. KFAM code should use the local profiles code
2. Have one workflow that builds/tests both Profiles and KFAM images when the code of either one changes
    * We want to ensure changes in the Profiles API won't break KFAM

So @apo-ger ignore my above comments for not building KFAM. Let's have one workflow for both and have a follow-up issue on making KFAM to use the local code of Profiles, to ensure we won't end up with broken KFAM from Profiles.

apo-ger:
@kimwnasptd Thanks for all the comments and the context! I've also updated the PR to publish the kfam image only when a PR is merged as mentioned in https://github.com/kubeflow/kubeflow/issues/6766#issue-1463011571.

kimwnasptd:
Thank you for the work @apo-ger!

/lgtm
/approve

PR #6823, opened 2022-12-08
vwa(front): Add YAML tab to Volume details page

orfeas-k:
This PR is part of the [Details page for each Notebooks/Volumes/TensorBoards](https://github.com/kubeflow/kubeflow/issues/6707) effort and a follow-up PR to [this](https://github.com/kubeflow/kubeflow/pull/6806). It adds a tab to show the full `yaml` file of the PVC.

### DIFF
--- a/components/crud-web-apps/volumes/frontend/angular.json
+++ b/components/crud-web-apps/volumes/frontend/angular.json
@@ -34,11 +34,14 @@
             "tsConfig": "tsconfig.app.json",
             "assets": [
               "src/favicon.ico",
-              "src/assets"
+              "src/assets",
+              {
+                "glob": "**/*",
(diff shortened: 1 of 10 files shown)


kimwnasptd:
@orfeas-k the PR looks good! Afterwards can you also send a GH Action for running the Volumes web app's unit tests like we do in https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/jwa_frontend_tests.yaml

We'll also have this for the TWA in https://github.com/kubeflow/kubeflow/pull/6824, so we could even break down the centralized linting workflow check

orfeas-k:
@kimwnasptd Sure, we can merge this one and I 'll include the GH action in a later PR where I want to add some unit tests for the VWA details page.

PR #6839, opened 2022-12-12
feat: use native RBAC subject to find RoleBindings in KFAM, rather than annotations

alexlatchford:
The current implementation of the listing of namespaces in the Kubeflow UI top-nav if you dig through the path of calls eventually hits this method when a user is authenticated. The `user` argument is populated with their `email` (in our implementation at least, but assuming it's some `user` identifier more broadly), we want to move away from checks on hard-coded annotations towards checks on Kubernetes native concepts themselves.

This refactor keeps existing checks on `annotations`, but also optionally checks the `Subjects` list and `RoleRef` to see if the matches occur there.

The rationale for this change is that currently within a `Namespace` we **must** deploy `O(U)` (where `U` is the # of users) `RoleBinding` resources as each one must have a `user: <user identifier>` `annotation`, then for other RBAC operations the `Subjects` list (including the now removed check) must have `len(subjects) == 1`. With this change we can instead deploy `O(1)` `RoleBinding` resources with `O(U)` `Subjects`. 

This will simplify our deployments here at Zillow and generally make this system more maintainable for us to use. Please let us know if there are any blockers in this, happy to refactor as needed!

Tested both approaches on our clusters, seems to work for our use cases but we've modified Kubeflow a bit from the vanilla and don't deploy everything 😅

### DIFF
--- a/components/access-management/kfam/bindings.go
+++ b/components/access-management/kfam/bindings.go
@@ -186,7 +186,7 @@ func (c *BindingClient) List(user string, namespaces []string, role string) (*Bi
 		for _, roleBinding := range roleBindings {
 			var subject rbacv1.Subject
 			for _, roleBindingSubject := range roleBinding.Subjects {
-				if roleBindingSubject.Kind == USER && roleBindingSubject.Name == user {
+				if roleBindingSubject.Kind == rbacv1.UserKind && roleBindingSubject.Name == user {
 					subject = roleBindingSubject
 					break
 				}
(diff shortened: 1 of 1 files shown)


alexlatchford:
Hey @thesuperzapper or @StefanoFioravanzo any thoughts on this change?

jbottum:
/kind feature
/priority p1
/area jupyter

@kubeflow/wg-notebooks-leads Is this on the agenda for the next Notebooks WG Meeting or can this be resolved before that call ?

alexlatchford:
Hey folks, any updates on this one? We've got this deployed to our copy of Kubeflow and seems to be working just fine for us :)

thesuperzapper:
@alexlatchford sorry for the truly crazy delay on this one, we have been slowly going through the backlog and getting things cleaned up.

This is an interesting idea, I will add it to the 1.9.1 milestone (release after next), if you can share any other things that can help us review, please do, otherwise, we will take a deeper look in the next months.

andreyvelich (commented):
The KFAM code has been migrated to the new repo, please re-open PR at: https://github.com/kubeflow/dashboard
/close

PR #6842, opened 2022-12-13
Fix workflows for publishing images only when PR is merged

apo-ger:
Follow-up PR to address: https://github.com/kubeflow/kubeflow/issues/6766#issuecomment-1351629795

Changes: 
- Publish the docker images only when a PR is merged. 
- Remove workflow that builds/push all images


kimwnasptd:
Hopefully the `push` trigger will work according to our understanding and the `secrets` will be picked up

/lgtm
/approve

PR #6846, opened 2022-12-14
centraldashboard-angular(front): Add library.js, source-maps & ESLint

orfeas-k:
This PR is part of the [Central Dashboard Rewrite effort](https://github.com/kubeflow/kubeflow/issues/6000). 

 - Enable **source maps** for all scripts, styles and node modules in both development and production.
 - Configure ESLint
 - Import `libary.js` file from the old centraldashboard project into `frontend/public` directory. Built it using webpack exporting the built file under `dist/public`, and use a `glob` to copy it into assets for serving. For local development, create a proxy route for the `dashboard.lib.js` file in order to fetch this from the backend. (We will also need an entry for the /api route in the future)

### DIFF
--- a/components/centraldashboard-angular/frontend/package.json
+++ b/components/centraldashboard-angular/frontend/package.json
@@ -48,6 +48,8 @@
     "@typescript-eslint/eslint-plugin": "4.28.2",
     "@typescript-eslint/parser": "4.28.2",
     "eslint": "^7.26.0",
+    "eslint-config-google": "0.12.0",
+    "eslint-loader": "2.2.1",
     "babel-loader": "^8.0.6",
     "cross-env": "^5.2.1",
     "jasmine-core": "~3.8.0",
--- a/components/centraldashboard-angular/frontend/package.json
+++ b/components/centraldashboard-angular/frontend/package.json
@@ -3,8 +3,9 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve --proxy-config proxy.conf.json",
+    "start": "npm run build-library && ng serve --proxy-config proxy.conf.json",
     "build": "ng build",
+    "build-library": "cross-env NODE_ENV=production webpack",
     "watch": "ng build --watch --configuration development",
     "test": "ng test",
     "test:prod": "ng test --no-watch --browsers ChromeHeadlessNoSandbox",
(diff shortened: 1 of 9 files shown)


components/centraldashboard-angular/frontend/package.json
```diff
@@ -3,8 +3,9 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve --proxy-config proxy.conf.json",
+    "start": "npm run build-library && ng serve --proxy-config proxy.conf.json",
     "build": "ng build",
```
kimwnasptd, line 7: Should we change this to
```json
"build": "npm run build-library && ng build"
```
to ensure that the produced library ends up in the final docker image as well?

orfeas-k: My suggestion @kimwnasptd  is to add what you proposed on the comment above and since `ng build` already copies the built library into its assets folder (since we have the `glob`), then go ahead and add the following line in the `Dockerfile`, since that's where the `backend` expects to find the `dashboard_lib.bundle.js`.

```
  COPY --from=backend /src .
  COPY --from=frontend /src/dist/default/ ./dist/public/
+ COPY --from=frontend /src/dist/default/assets ./dist/public/

  EXPOSE 8082
  ENTRYPOINT ["npm", "start"]
```

orfeas-k: We should be aware that running `npm run build -- --output-path=./dist/default --configuration=production` in the Dockerfile only affects the output path of `ng build` and not `npm run build-library`.

kimwnasptd: @orfeas-k maybe I'm missing something. Where in the current Dockerfile do we build the library and place the produced file under `dist/public`? 

I understand we have a bug now where you had built the library previously with `npm run build-library` and the docker build process ended up using that file, but want to confirm.

orfeas-k: > Where in the current Dockerfile do we build the library and place the produced file under `dist/public`?

@kimwnasptd Nowhere, that's why I think we should add the following line (that you proposed). `npm run build-library` will build the library and place the file under `dist/public`.
```
    "start": "npm run build-library && ng serve --proxy-config proxy.conf.json",
```

kimwnasptd: Ah, OK I misinterpreted your first comment. Then my last question regarding adding this line:
```
  COPY --from=backend /src .
  COPY --from=frontend /src/dist/default/ ./dist/public/
+ COPY --from=frontend /src/dist/default/assets ./dist/public/
```
Do we need other files from `src/dist/default/assets` or just the built library? If it's just the library then let's explicitly copy that file

orfeas-k: Explicitly that file (and the corresponding  `.map`) so I just rebased and pushed a branch with a fixup for this one @kimwnasptd

kimwnasptd: looks good, thanks!

PR #6847, opened 2022-12-14
fix: do not block mounting one volume multiple times to different paths w/ PodDefault

awittzb:
It Kubernetes, it is valid to mount one volume to two different paths:

```yaml
  volumeMounts:
  - name: my-volume
    mountPath: /some-place
  - name: my-volume
    mountPath: /some-other-place
```
You might actually want to mount a volume multiple times if you were also using SubPaths, e.g:

```yaml
  volumeMounts:
  - name: my-volume
    mountPath: /some-place
    subPath: /notebook/public
  - name: my-volume
    mountPath: /some-other-place
    subPath: /logs/something
```

The `mergeVolumeMounts` used when applying a PodDefault does _not_ allow this.

This PR updates the `mergeVolumeMounts` code to allow a Volume to be mounted multiple times in a PodDefault, so long as the volumeName+mountPath combination are unique.

### DIFF
--- a/components/admission-webhook/main.go
+++ b/components/admission-webhook/main.go
@@ -257,7 +257,7 @@ func mergeVolumeMounts(volumeMounts []corev1.VolumeMount, podDefaults []*setting
 	origVolumeMounts := map[string]corev1.VolumeMount{}
 	volumeMountsByPath := map[string]corev1.VolumeMount{}
 	for _, v := range volumeMounts {
-		origVolumeMounts[v.Name] = v
+		origVolumeMounts[v.Name+"|"+v.MountPath] = v
 		volumeMountsByPath[v.MountPath] = v
 	}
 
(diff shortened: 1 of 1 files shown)


components/admission-webhook/main_test.go
```diff
@@ -273,3 +275,176 @@ func TestSetCommandAndArgs(t *testing.T) {
...
+	
+	for _, test := range []struct {
+		name        string
+		in          []corev1.VolumeMount
+		podDefaults []*settingsapi.PodDefault
+		outMounts   []corev1.VolumeMount
+		outErrs     []error
+	}{
+		{
+			name: "Can mount the same volume twice at different paths",
```
awittzb, line 291: Without the fix in this PR, this unit test will fail, demonstrating the flaw.

kimwnasptd:
Thanks for the PR @awittzb! This is in my list right after https://github.com/kubeflow/kubeflow/pull/6749

juliusvonkohout:

awittzb:
Unfortunately, I have an unavoidable conflict w/ the Sep 14 Notebook WG!

juliusvonkohout:
It is biweekly, you can just join the next one then.

juliusvonkohout:
Just please put yourself on the Agenda in the meeting notes.

PR #6849, opened 2022-12-15
Extend Docker Publish workflows to be triggered on Version change

apo-ger:

Changes:
- Add the path `releasing/version/VERSION` in the paths of the push event to also trigger the workflow when that file changes
- Use the [dorny/paths-filter@v2](https://github.com/dorny/paths-filter) action to monitor the path `releasing/version/VERSION` and if it changes run a step
  that pushes the image with the `TAG=$(cat releasing/version/VERSION)`
- Set `fetch-depth:0` to [checkout](https://github.com/actions/checkout) step fetch all history for all tags and branches. This is required so that the tags won't end up being just a hash commit.
- Set `Version` to `latest` to build all images with that tag


kimwnasptd:
If everything goes as expected we should see a `latest` tag in all images in https://hub.docker.com/u/kubeflownotebookswg

/lgtm
/approve

PR #6850, opened 2022-12-15
twa(front): Install prettier and add gh action

elenzio9:
In this PR:

- Install Prettier
- Format code
- Add formatting/linting to `twa_frontend_tests` workflow

### DIFF
--- a/.github/workflows/twa_frontend_tests.yaml
+++ b/.github/workflows/twa_frontend_tests.yaml
@@ -5,6 +5,31 @@ on:
       - components/crud-web-apps/tensorboards/frontend/**
 
 jobs:
+  test:
+    name: Code format and lint
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 12
+
+      - name: Format tensorboards web app
+        run: |
+          cd components/crud-web-apps/tensorboards/frontend
+          npm i
+          npm run format:check
+
+      - name: Lint tensorboards web app
+        run: |
+          cd components/crud-web-apps/tensorboards/frontend
+          npm i
+          npm run lint-check
+
   frontend-unit-tests:
     runs-on: ubuntu-latest
     name: Unit tests
--- a/components/crud-web-apps/tensorboards/frontend/package.json
+++ b/components/crud-web-apps/tensorboards/frontend/package.json
@@ -12,7 +12,9 @@
     "test": "ng test",
     "test:prod": "ng test --browsers=ChromeHeadless --watch=false",
     "lint-check": "ng lint",
-    "lint": "ng lint --fix"
+    "lint": "ng lint --fix",
+    "format:check": "prettier --check 'src/**/*.{js,ts,html,scss,css}' || node scripts/check-format-error.js",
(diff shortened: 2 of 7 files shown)


.github/workflows/twa_frontend_tests.yaml
```diff
@@ -5,6 +5,31 @@ on:
       - components/crud-web-apps/tensorboards/frontend/**
 
 jobs:
+  test:
```
kimwnasptd, line 8:
```suggestion
  frontend-fromat-linting-check:
```

.github/workflows/twa_frontend_tests.yaml
```diff
@@ -5,6 +5,31 @@ on:
...
+    name: Code format and lint
```
kimwnasptd, line 9:
```suggestion
    name: Check code format and lint```

File: .github/workflows/twa_frontend_tests.yaml [hunk 3]
```diff
@@ -5,6 +5,31 @@ on:
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 12
+
+      - name: Format tensorboards web app
```
kimwnasptd, line 21:
```suggestion
      - name: Check frontend code formatting
```

.github/workflows/twa_frontend_tests.yaml
```diff
@@ -5,6 +5,31 @@ on:
...
+        run: |
+          cd components/crud-web-apps/tensorboards/frontend
+          npm i
+          npm run format:check
+
+      - name: Lint tensorboards web app
```
kimwnasptd, line 27:
```suggestion
      - name: Check frontend code linting
```

components/crud-web-apps/tensorboards/frontend/package.json
```diff
@@ -12,7 +12,9 @@
     "test": "ng test",
     "test:prod": "ng test --browsers=ChromeHeadless --watch=false",
     "lint-check": "ng lint",
-    "lint": "ng lint --fix"
+    "lint": "ng lint --fix",
+    "format:check": "prettier --check 'src/**/*.{js,ts,html,scss,css}' || node scripts/check-format-error.js",
```
kimwnasptd, line 16: I see that we don't have this script in TWA, like we do in Katib
https://github.com/kubeflow/katib/tree/master/pkg/new-ui/v1beta1/frontend/scripts

Yet the action didn't fail. Why is that?

elenzio9: Oh sorry I missed that. Should I add this script in TWA as well?

kimwnasptd: OK I see what happened. `||` means that the second command (node script in this case) will NOT be run if the first part was successful. But the node script will run if the prettier check fails.

That script is just there to instruct users to run the formatting. Let's include it in this PR

elenzio9: ACK! I just added it!

.github/workflows/twa_frontend_tests.yaml
kimwnasptd, line 8: Could we change this to something like `frontend-format-lint-check`?

PR #6854, opened 2022-12-15
Update KF manifests and gh-action workflows to use the tag=`latest`

apo-ger:
This PR is part of https://github.com/kubeflow/kubeflow/issues/6766

Changes:

* Update manifests to use tag=`latest`
* Update `publish-image` & `intergration-test` workflows to use tag=`latest`


.github/workflows/centraldb_angular_docker_publish.yaml
```diff
@@ -11,6 +11,7 @@ on:
 env:
   DOCKER_USER: kubeflownotebookswg
   IMG: kubeflownotebookswg/centraldashboard-angular
+  TAG: latest
```
kimwnasptd, line 14: let's remove the TAG here and just build/push the image two times like this, in the dedicated step:
```yaml
    - name: Run CentralDashboard-Angular build/push
      run: |
        cd components/centraldashboard-angular
        make docker-build docker-push

        export TAG=latest
        make docker-build docker-push
```

.github/workflows/centraldb_angular_docker_publish.yaml
```diff
@@ -49,8 +50,6 @@ jobs:
       id: version
       if: steps.filter.outputs.version == 'true'
       run: |
-        docker tag ${{env.IMG}}:${TAG} ${{env.IMG}}:${VERSION_TAG}
+        export VERSION_TAG=$(cat releasing/version/VERSION)
+        docker tag ${{env.IMG}}:${{env.TAG}} ${{env.IMG}}:${VERSION_TAG}
         docker push ${{env.IMG}}:${VERSION_TAG}
```
kimwnasptd, line 55: Let's re-use the makefile rules here, to avoid duplication of code in makefiles and GH Actions. The layers are already cached so rebuilding the image should be very fast
```bash
export TAG=$(cat releasing/version/VERSION)
cd components/centraldashboard-angular
make docker-build docker-push
```

kimwnasptd (commented):
@apo-ger let's ensure the following checks for each workflow:
1. Let's build two images on each merge (aside from VERSION). One from current commit and one for `latest`
3. Let's have the conditional step for the VERSION just rebuild/push the images using the Makefiles, to avoid redefining the build logic in the GH Actions

I've added some comments on how to achieve these in one component, but should apply for the rest as well

kimwnasptd:
Lastly, there's a small edge case scenario we can discuss afterwards. With the above workflows the GH Actions when merging code on a *release branch* is also going to build images with the `latest` tag. 

But this will break our convention that `latest` only comes from the develop/main branch. But let's tackle this afterwards

.github/workflows/centraldb_angular_docker_publish.yaml
kimwnasptd: Actually, I was thinking the issue with the `latest` tag and I think we should close this here once and for all.

Would something like this work?
```yaml
    - name: Run CentralDashboard-Angular build/push
      run: |
        cd components/centraldashboard-angular
        make docker-build docker-push

    - name: Run CentralDashboard-Angular build/push latest
      if: github.ref == 'refs/heads/master'
      run: |
        export TAG=latest
        make docker-build docker-push
```
https://stackoverflow.com/a/58142412
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

.github/workflows/centraldb_angular_docker_publish.yaml
apo-ger: Sure! However, we can't avoid duplication of code for building/pushing the notebook server images. This is because we don't have an `IMG` env var in the Makefiles (e.g [codeserver-python Makefile](https://github.com/kubeflow/kubeflow/blob/658f5b8a76e3b19438678decaf21a9c2fdfd9af1/components/example-notebook-servers/codeserver-python/Makefile#L1)) to set the `registry` there. Therefore with the current Makefiles we have to use:
- `make docker-build`
- `docker tag` to add the registry
- `and finally docker push`

For example:
```yaml
- name: Build and push Notebook Server images
      run: |
        export TAG=$(shell git describe --tags --always --dirty)
        cd components/example-notebook-servers/
        make docker-build -C codeserver-python TAG=${TAG}
        docker tag codeserver-python:${TAG} ${{env.REGISTRY}}/codeserver-python:${TAG}
        docker push ${{env.REGISTRY}}/codeserver-python:${TAG}

    - name: Build and push latest Notebook Server images
      if: github.ref == 'refs/heads/master'
      run: |  
        export TAG=latest
        cd components/example-notebook-servers/
        make docker-build -C codeserver-python TAG=${TAG}
        docker tag codeserver-python:${TAG} ${{env.REGISTRY}}/codeserver-python:${TAG}
        docker push ${{env.REGISTRY}}/codeserver-python:${TAG}

    - name: Build and push Notebook Server images on Version change
      id: version
      if: steps.filter.outputs.version == 'true'
      run: |
        export TAG=$(cat releasing/version/VERSION)
        cd components/example-notebook-servers/
        make docker-build -C codeserver-python TAG=${TAG}
        docker tag codeserver-python:${TAG} ${{env.REGISTRY}}/codeserver-python:${TAG}
        docker push ${{env.REGISTRY}}/codeserver-python:${TAG}
```

.github/workflows/centraldb_angular_docker_publish.yaml
apo-ger: Yes this will work, I've made the appropriate changes

.github/workflows/centraldb_angular_docker_publish.yaml
kimwnasptd: @apo-ger let's leave the notebook images completely out for this effort.

I believe we would be able to use the Makefiles if we'd use a `REGISTRY` env var in each Makefile. But let's discuss this in a follow up PR, since this one is already getting big

apo-ger: @kimwnasptd 
> let's leave the notebook images completely out for this effort.
Should I remove the logic for building with latest tag or remove the notebook-server workflows completely and tackle them in the follow up PR?

kimwnasptd: just the changes from this PR, and we can send another PR for the latest tag in a separate PR that:
1. Uses a `REGISTRY` env var in all makefiles (let's see if we need to do more)
2. Use the Makefiles to build each notebook, but with a different TAG each time like we do for the rest of the components

PR #6856, opened 2022-12-16
centraldashboard-angular(front): Sync Browser and Iframe URLs

orfeas-k:
### Story

This PR is part of the [Central Dashboard Rewrite effort](https://github.com/kubeflow/kubeflow/issues/6000) and it implements the functionality described in detail in this [comment](https://github.com/kubeflow/kubeflow/issues/6000#issuecomment-1341254207). In short, our goal is have an Iframe that will contain the KF web apps, detect changes in the Iframe's internal URL and reflect those changes in the browser's URL.

In more detail, functionalities that are implemented in this PR are:

 - Detect changes in `iframe`'s internal URL by checking in time intervals. Upon change, modify the browser's (Router) URL to be the same as the `iframe`'s one.
 - Update `iframe`'s `src` property based on route events.
 - Send manually a postMessage with a hardcoded namespace `'kubeflow-user'` in order to be able to test the WAs until we implement a functionality for handling namespaces. 

#### Query parameters

Passing query parameters as part of the URL resulted in character errors e.g. in `%3Ftab%3Dlogs` instead of `?tab=logs`). To overcome this, we get the query parameters string from `window.location.search`, create a new [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)  from this and then use its values in order to create an object that can be passed as an argument to `queryParams` in [`router.navigate()`](https://v12.angular.io/api/router/UrlCreationOptions#queryParams)

#### Src updating
 - Update the `src` property only when the browser's URL and the `iframe`'s URL are different in order to prevent the `iframe` from reloading on every user click.
  - Modify `iframe`'s `src` property setter by appending/removing the `window.location.origin`, in order to force the `iframe` to reload in cases that the user wants to navigate to a URL that has the same value with the current value of the `src` property. More specifically, when the `src` property doesn't contain the `window.location.origin`, we append it to the relative URL and pass it to `src` and when it does contain it, we just pass the relative URL as is.
Thus, we always consider that values for the sidenav menu in configmaps contain only relative URLs.

#### Trailing slash
As a convention, **we will always treat URLs with or without a trailing slash [as the same URL](https://developers.google.com/search/blog/2010/04/to-slash-or-not-to-slash)**. In order to compare, thus, URLs between the `browser` and the `iframe`, we will use a function that goes ahead and appends a `/` at the end of both URLs if there is none to avoid false statements in cases where they only differ in the trailing slash. 

This is essential since we noticed the following behaviour: When navigating to index pages, WAs (JWA and VWA at least) append a trailing slash to their URL if there is none. On the contrary, when navigating to another page inside the WAs, if there is a trailing slash, they go ahead and remove it. 

### Next PRs
The goal is to follow with 2 PRs that will
 - Add UI tests for this PR's functionalities
 - Complete the Iframe URL tracking functionality by mirroring changes from the browser's URL in the UI (e.g. when relative path is `/<webapp-path>/...`, make sure that the sidenav menu has always as "selected" the corresponding WA)

### DIFF
--- a/components/centraldashboard-angular/frontend/src/app/pages/iframe-wrapper/iframe-wrapper.component.ts
+++ b/components/centraldashboard-angular/frontend/src/app/pages/iframe-wrapper/iframe-wrapper.component.ts
@@ -28,6 +28,8 @@ export class IframeWrapperComponent implements AfterViewInit, OnDestroy {
      * reloading constantly, its src is not updated on each navigation
      * a user does.
      */
+    src = this.removePrefixFrom(src);
+
     if (window.location.origin) {
       if (!this.prvSrcPath?.includes(window.location.origin)) {
         src = window.location.origin + src;
@@ -56,10 +58,15 @@ export class IframeWrapperComponent implements AfterViewInit, OnDestroy {
         ? iframeWindow?.location.pathname + iframeWindow?.location.search
         : iframeWindow?.location.pathname;
 
-      if (!this.equalUrlPaths(event.url, iframeUrl)) {
+      const eventUrl = this.removePrefixFrom(event.url);
+      if (!this.equalUrlPaths(eventUrl, iframeUrl)) {
         this.srcPath = event.url;
       }
     });
+  }
+
+  removePrefixFrom(url: string) {
+    return url.includes('/_') ? url.slice(2) : url;
   }
 
   /**
@@ -96,7 +103,7 @@ export class IframeWrapperComponent implements AfterViewInit, OnDestroy {
         this.iframeLocation = currentUrl;
         const path = iframeWindow?.location.pathname;
         const queryParams = this.getQueryParams(iframeWindow?.location.search);
-        this.router.navigate([path], { queryParams });
+        this.router.navigate(['/_' + path], { queryParams });
       }
     }, 100);
   }
--- a/components/centraldashboard-angular/frontend/src/app/pages/iframe-wrapper/iframe-wrapper.component.ts
+++ b/components/centraldashboard-angular/frontend/src/app/pages/iframe-wrapper/iframe-wrapper.component.ts
@@ -16,7 +16,7 @@ import { Subscription } from 'rxjs';
 export class IframeWrapperComponent implements AfterViewInit, OnDestroy {
   @ViewChild('iframe') iframe: ElementRef<HTMLIFrameElement>;
 
-  public prvSrcPath: string;
+  private prvSrcPath: string;
   get srcPath(): string {
     return this.prvSrcPath;
   }
@@ -34,9 +34,12 @@ export class IframeWrapperComponent implements AfterViewInit, OnDestroy {
       }
     }
 
-    this.prvSrcPath = src;
-    // Some KF distributions need a trailing slash "/" in order to resolve paths
-    this.prvSrcPath += this.prvSrcPath?.endsWith('/') ? '' : '/';
+    /**
+     * When Istio exports Services, it always expects
+     * a '/' at the end. SO we'll need to make sure the
+     * links propagated to the iframe end with a '/'
+     */
+    this.prvSrcPath = this.appendBackslash(src);
   }
   public iframeLocation: string | undefined = 'about:blank';
   private urlSub: Subscription;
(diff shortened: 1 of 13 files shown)


components/centraldashboard-angular/frontend/src/app/pages/iframe-wrapper/iframe-wrapper.component.ts
```diff
@@ -0,0 +1,121 @@
...
+      }
+
+      if (!this.equalUrlPaths(event.url, iframeUrl)) {
+        this.srcPath = event.url;
+      }
+    });
+  }
+
+  equalUrlPaths(firstUrl: string, secondUrl: string | undefined) {
+    if (!firstUrl && !secondUrl) {
```
kimwnasptd, line 65: Let's add a comment here to explain why we need this function, which is to handle the sometimes missing `/` from some urls

components/centraldashboard-angular/frontend/src/app/pages/iframe-wrapper/iframe-wrapper.component.ts
```diff
@@ -0,0 +1,121 @@
...
+      return false;
+    }
+    firstUrl += firstUrl?.endsWith('/') ? '' : '/';
+    secondUrl += secondUrl?.endsWith('/') ? '' : '/';
+    return firstUrl === secondUrl;
+  }
+
+  ngAfterViewInit() {
+    this.interval = setInterval(() => {
+      let currentUrl = this.iframe?.nativeElement?.contentWindow?.location.href;
```
kimwnasptd, line 79: nit, but could we do something like this:
```typescript
      const iframeWindow = this.iframe?.nativeElement?.contentWindow;
      let currentUrl = iframeWindow?.location.href;
```

and also use the `iframeWindow` var in the following places to make the code slightly more readable

kimwnasptd (commented):
@orfeas-k took an initial look. Nice work!

First off, I see that in this iteration we do not have the `/_` prefix that the app previously had. I propose that we add this functionality in this PR, since now the local app works only because Angular decides to remove (why?) the last `/` from the routes.

So right now locally if we do `localhost:4200/jupyter` in the browser it will take us to the app. But if we'd do `localhost:4200/jupyter/` then the proxied app would be used.

Then I believe I observed the following 2 bugs:
1. setting the browser's URL to `localhost:4200/jupyter/?ns=kubeflow-user` makes the app enter a refresh loop
2. Right now we append `/` at the url in the iframe component. What if the user entered the above url that contains a query param?

orfeas-k:
@kimwnasptd Thanks for looking into it, all the comment-fixes are on point!

> right now locally if we do `localhost:4200/jupyter` in the browser it will take us to the app. But if we'd do `localhost:4200/jupyter/` then the proxied app would be used.

I can't see this happening. I think it just loads the proxied app  in both cases. I'm working on adding the functionality you mentioned for the `/_` prefix in links. Should we implement a functionality for redirecting in the cases above as part of this PR? 

Regarding the 2 bugs mentioned, I think they are pretty much part of the same problem since from what I see, when I try to navigate directly to `http://localhost:4200/_/jupyter/?ns=kubeflow-user` (with the prefix functionality implemented), it goes ahead and append a `/` at the end and start an infinite loop of refreshing with the link ending in something like this: `http://localhost:4200/_/jupyter?ns=kubeflow-user%2F%2F%2F%2F...`.

orfeas-k:
In order to solve the above bug(s):
 - `appendBackslash(url)` has to be smart and append the backslash before the query parameters when there are.
 - In the same spirit, `equalUrlPaths()` should return `true` even in cases that the input URLs are different by a `/`  (be it in the end of the URL or just before the query parameters).
 
I also added some comments in order for the data flow between browser and iframe to be clear.

orfeas-k:
I just pushed the fixes, I think we should be good to go now @kimwnasptd .

kimwnasptd:
@orfeas-k the changes look good! One last nit I observed is that the iframe area does not "fit" the height of the screen. 

image

We can tackle it in a future PR though, to first proceed with this first step

kimwnasptd:
Thanks for the great work!

/lgtm
/approve

PR #6860, opened 2022-12-20
Fix Makefiles/Workflows/Manifests for the example-notebook-server images

apo-ger:
This is a follow-up PR to address https://github.com/kubeflow/kubeflow/pull/6854#discussion_r1053260293 and is part of https://github.com/kubeflow/kubeflow/issues/6766

Changes:
- Fix Makefiles to use a `REGISTRY` env var to set registry prefix when building/pushing notebook-server images
- Fix publish workflows to Build/Push notebook-server images via `docker-build` & `docker-push` Makefile rules with appropriate tags:
  - Always Build/Push notebook-server images with `TAG=$(shell git describe --tags --always --dirty)`
  - Build/Push notebook-server images with `TAG=latest` only when the target branch is master
  - Build/Push notebook-server images with `TAG=$(cat releasing/version/VERSION)` when
  Version changes
- Update KF manifests to use `latest` tag for notebook-server images

### DIFF
--- a/.github/workflows/nb_server_codeserver_python_docker_publish.yaml
+++ b/.github/workflows/nb_server_codeserver_python_docker_publish.yaml
@@ -11,7 +11,6 @@ on:
       - releasing/version/VERSION
 
 env:
-  TAG: $(git describe --tags --always --dirty)
   DOCKER_USER: kubeflownotebookswg
   REGISTRY: kubeflownotebookswg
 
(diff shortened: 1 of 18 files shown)


kimwnasptd:
Looks good! Afterwards could you send a last PR to revert back the proposed change for `fetch-depth: 0` in checkout? This ends up with creating tags like `v1.5.0-rc.0-213-g5a49072a`, since we had (mistakenly) created a git tag for RC0 of 1.5 from the main branch.

In the current branching model we will be *first* creating a new branch for a release and then cut a tag off of that branch.

Without fetching the whole history we'll end up just using a commit for now as the tag, which should be good enough.

/lgtm
/approve

apo-ger:
@kimwnasptd Thanks for the review! I've sent a follow up [PR](https://github.com/kubeflow/kubeflow/pull/6863) to address: https://github.com/kubeflow/kubeflow/pull/6860#issuecomment-1360960463

PR #6862, opened 2022-12-21
vwa(front): Add UI tests with Cypress

orfeas-k:
This PR adds UI tests for key functionalities of the VWA frontend. It
 - Installs Cypress & removes protractor files
 - Adds UI tests with Cypress
 - Add UI tests to VWA's frontend GH action.

#### Mock API vs Using the Backend (updated)

We decided with @kimwnasptd to: 

- Focus our tests only on the frontend's functionalities
- Write tests that completely mock all the API calls to the backend. This will allow us to test locally by only running an Angular dev server for the frontend. At the same time, there is no need to do any sort of garbage collecting (which would be required if we used the developer's backend)

#### Test scenarios

The **UI tests** we wrote for VWA after all check that:

 * New Volume form dialog has proper default values.
 * Index page table exposes all PVCs and their names received from the `GET` request.
 * Index page table status icons are correctly shown according to each PVC's `status.phase` field.

#### GH action - Node Version
By default, the tests ran with a `node version` higher than 14 and installing dependencies for KF common library failed. Thus, we decided to follow what we do for [Dockerfiles](https://github.com/arrikto/prv-kubeflow-kubeflow/blob/develop/components/crud-web-apps/jupyter/Dockerfile#L10) and explicitly use node version 12. This is achieved by using Github's `setup-node` action https://github.com/actions/setup-node#usage.


kimwnasptd:
@orfeas-k could you rebase this PR to resolve some conflicts?

orfeas-k:
Just rebased and pushed! @kimwnasptd

kimwnasptd:
Thanks!

/lgtm
/approve

Issue #6493, opened 2022-05-30
Use Tensorboard with Object Store


N-Kingsley:
How to write a right Object Storage Link?
For example, how to set the access key and secret key in the link when using the Minio.
![image](https://user-images.githubusercontent.com/38311217/170909811-cb625930-c069-4dec-9010-fefcdd1501ef.png)

kimwnasptd:
This is currently not supported, but a great feature to implement. 

We've solved this problem in Notebooks by using PodDefaults for attaching credentials, but we might need to have some more logic for TensorBoard.

@N-Kingsley have you managed to configure TensorBoard locally to access MinIO?

N-Kingsley:
> This is currently not supported, but a great feature to implement.
> 
> We've solved this problem in Notebooks by using PodDefaults for attaching credentials, but we might need to have some more logic for TensorBoard.
> 
> @N-Kingsley have you managed to configure TensorBoard locally to access MinIO?

No, how should I solve this problem？

hwk42:
I solved this problem by add credential env for notebook pod. Is it a plan to support in v1.6?

tks-ai:
I think it would be a very useful feature to support s3.
I am struggling  with this problem for a few days ,sad to know this is not supported, 
If take a fully implement will take time, how about just pass the credential of kubeflow built-in minio to Tensorboard by ConfigMap.
Think this would be a quick solution.

kimwnasptd:
@wyljpn just saw your comment in the other issue, let's use this one for this specific feature https://github.com/kubeflow/kubeflow/issues/3578#issuecomment-1032417871

First of all nice work! As mentioned above instead of expecting users to edit ENV vars directly let's instead use PodDefaults, so that we can reuse this part for Notebooks as well. We can also document a PodDefault that sets the ENV Vars accordingly.

The changes I have in mind are:
1. Extend the `spec` of the `TensorBoard` CR to allow users to edit the `labels` that will end up in the Pod. We can do `.spec.labels` for this in the CRD
2. Extend the frontend to have a `Configurations` tab, just like we have for Notebooks
3. Extend the backend to set the PodDefaults labels to the CR, based on the frontend's data

We can break this down to small PRs for each one as well

laserK3000:
[...]

PR #6874, opened 2023-01-04
Support Pod Defaults in Tensorboard controller

surajkota:
This PR adds functionality for adding pod default related labels for tensorboard pods in tensorboard controller.  This can be used for various things like configuring cloud storage parameters using environment variables(#6493), service account, tolerations etc.

User experience:
1. User creates a tensorboard CR with required labels (`metadata.labels`)
2. Controller will create the deployment with labels on the pod copied from the tensorboard CR
3. Admission webhook will take care of injecting necessary configurations according to poddefault

Updated the image version because v2.1.0 did not work with IAM roles. Using the latest version would require building and maintaining an image because some functionality is moved out to a different package(tensorflow-io). This package is not installed by default in tensorflow image and I would like to focus on completing the core functionality first.

@kimwnasptd @kandrio  @elikatsis 

Is there anyone who can help with implementing the webapp frontend related changes or
I would appreciate if anyone can give me a quick intro on the frontend Angular part of the code so I can make the changes myself?

### DIFF
--- a/components/tensorboard-controller/controllers/tensorboard_controller.go
+++ b/components/tensorboard-controller/controllers/tensorboard_controller.go
@@ -246,6 +246,12 @@ func generateDeployment(tb *tensorboardv1alpha1.Tensorboard, log logr.Logger, r 
 		})
 	}
 
+	// copy all of the CR labels to the pod which includes poddefault related labels
+	podLabels := map[string]string{"app": tb.Name}
+	for k, v := range tb.ObjectMeta.Labels {
+		(podLabels)[k] = v
+	}
+
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      tb.Name,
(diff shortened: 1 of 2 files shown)


components/tensorboard-controller/controllers/tensorboard_controller.go
```diff
@@ -246,6 +246,12 @@ func generateDeployment(tb *tensorboardv1alpha1.Tensorboard, log logr.Logger, r
 		})
 	}
 
+	// copy all of the CR labels to the pod which includes poddefault related labels
+	podLabels := map[string]string{"app": tb.Name}
```
kimwnasptd, line 250: @surajkota could you initially set this as an empty dict and then set the `app: tb.Name` *after* the for loop?

We want to make sure the Controller will be correctly setting this value, even if a user requested a different value

surajkota: Done, please review. Thanks

kimwnasptd:
Looks good, thanks! I'll provide some more context regarding the web app changes as well in the original issue, so that we can push this through https://github.com/kubeflow/kubeflow/issues/6493

PR #6875, opened 2023-01-05
jwa(front): Auto update mount path

tasos-ale:
Fix an issue in JWA's form and change the volume's mount path when the volume's name changes. This feature is already working for workspace volumes. Also, add one more e2e test for the volume's mount path.

### DIFF
--- a/components/crud-web-apps/jupyter/frontend/cypress/integration/form-page.spec.ts
+++ b/components/crud-web-apps/jupyter/frontend/cypress/integration/form-page.spec.ts
@@ -1,0 +1,50 @@
+describe('New notebook form', () => {
+  beforeEach(() => {});
+
+  it('should have a "New notebook" title', () => {
+    cy.visit('/new');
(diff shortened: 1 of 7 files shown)


orfeas-k (commented):
Tested the image and it fixes the issue. Also, the code looks good. Small nit would be to rename `volGroupPrv` to `prvVolGroup` since that's what we do in the rest of the code as far as I'm concerned but it's not essential. Αlso, recently we switched to the term `UI tests` instead of e2e for Cypress tests. Thank you @tasos-ale 
cc @kimwnasptd 
/lgtm

tasos-ale:
@orfeas-k I pushed the changes you mentioned. cc @kimwnasptd

orfeas-k:
Thank you @tasos-ale!
/lgtm

YQ-Wang:
@tasos-ale Could you take a look at this issue? https://github.com/kubeflow/notebooks/issues/33 It seems to be introduced by this PR.

PR #6876, opened 2023-01-05
Re-Introducing the Volumes Viewer

TobiasGoerke:
# Volumes Viewer

About two years ago, @DavidSpek [proposed a pvcviewer](https://github.com/kubeflow/kubeflow/pull/5783). We've been using and enjoying this feature quite a lot since then and I'm convinced this feature needs to find its way to Kubeflow's core. 
Unfortunately, the PR staled and never got merged even though it sparked serious interest among the community. 

This is my attempt to move forward with this feature, making it available for all users. 

PVCViewer

## Pre-Existing Work

@DavidSpek provided a functional prototype of the volumes viewer. His work comprises a pvcviewer resource definition, a resource controller and changes to the volumes UI.
His changes already got partially merged into the volumes UI but are only available within the rok distribution. 

The initial controller was based off the tensorboard controller and was WIP in many regards.

## My Work

I've re-implemented the volumes viewer, and made it compatible with the current volumes UI's master.

Also, the controller has been re-written from scratch and is well-tested. It now comprises new features, such as restarting pods on RWO-Nodename changes. 

## API

I've included an example of a viewer object. The API is similar to what @kimwnasptd descibed in another thread. 
Creating a VolumesViewer currently looks like this (comments included explaining its functionality).

```yaml
kind: VolumesViewer
metadata:
  name: volumesviewer-sample
spec:
  # The podTemplate is applied to the deployment.Spec.Template.Spec
  # and thus, represents the core viewer's application
  podTemplate:
    #...
  service:
    # Specifies the application's target port used by the Service
    targetPort: 8080
    # If defined, an istio VirtualService is created, pointing to the Service
    virtualService:
      # The base prefix is suffixed by '/namespace/name' to create the
      # VirtualService's prefix and a unique URL for each started viewer
      basePrefix: "/volumesviewer"
      # You may specify the VirtualService's rewrite.
      # If not set, the prefix's value is used
      rewrite: "/"
      # By default, no timeout is set
      # timeout: 30s
  rwoScheduling:
    # If set to true, the controller detects RWO-Volumes referred to by the
    # podTemplate and uses affinities to schedule the viewer to nodes
    # where the volume is currently mounted. This enables the viewer to
    # access RWO-Volumes, even though they might already be mounted.
    enabled: true
    # Using the rwoScheduling feature, the viewer might block other application
    # from (re-starting). Setting restart to true instructs the controller to
    # re-compute the affinity in case Pods start using the viewer's RWO-Volumes.
    # Thus, the viewer might restart on another node without blocking new Pods.
    restart: true
```

## How to install

Using the current kubeflow master:

1. Install the kustomize applications
    - `kustomize build components/crud-web-apps/volumes/manifests/overlays/istio | kubectl apply -f -`
    - `kustomize build components/volumes-viewer/config/overlays/kubeflow | kubectl apply -f -`
2. Build the container images and set them accordingly (alternatively, use my prebuilt images):
    - `kubectl -n kubeflow set image deploy/volumes-web-app-deployment volumes-web-app=tobiasgoerke/kubeflow-volumes-web-app:test-v3`
    - `kubectl -n kubeflow set image deploy/volumes-viewer-controller-manager manager=tobiasgoerke/kubeflow-volumes-viewer:test-v3`

## Outlook and Future Work

- There is a generic viewer controller managed by the pipelines WG. The volumes viewer would integrate into this controller nicely. However, [attempts to integrate the viewer](https://github.com/kubeflow/pipelines/issues/5651) have failed and it seems [the viewer itself is not moving forward, too](https://github.com/kubeflow/pipelines/issues/5681). 
Thus, I've decided to create a new controller (like @DavidSpek did in his PR) and call the integration into the pipelines WG optional and out-of-scope. In case this PR gets merged, I'll rename the VolumesViewer controller and push for it to be accepted as a generic viewer. This would enable the volumes viewer, notebooks, tensorboard etc. to use a common implementation and controller. For this PR, the CRD and controller could be dropped then, leaving us with only changes to the volumes UI. I've designed this PR with this possible step in mind so that a generic viewer would only require one file to change. 
- I've created a [PR in the filebrowser project](https://github.com/filebrowser/filebrowser/pull/2145) which enables Filebrowser to support the tus.io protocol for resumable and chunked uploads. This comes in very handy with big uploads that may get disrupted or proxies that block large requests. This PR is currently in review.


kimwnasptd:
@TobiasGoerke great to see traction towards this direction! First off, some more context on the history of this feature:

We (Arrikto) had initially developed this feature, PVCViewer CRD, for MiniKF alongside the Volumes web app. Later on we open sourced the Volumes web app but we didn't have the cycles to maintain the PVCViewer component as well, so the PVCViewer Controller did not make it upstream. You can still see the logic for handling these viewers in the current Rok flavor of the app:
https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/volumes/backend/apps/rok/pvcviewer.yaml
https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/volumes/backend/apps/rok/utils.py#L76
https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/volumes/frontend/src/app/pages/index/index-rok/config.ts#L13-L20

@apo-ger and I are more than happy to help with this feature, since this is something that could be adopted by a lot of KF users that rely on PVCs for their workflows! We also have more cycles now and would like to upstream our code and lessons learned from such a component. So it would be a great chance to combine it with this effort!

kimwnasptd:
I would only suggest that we first start with a design doc, since this is a huge feature that also interacts with other components. We have also seen some UX scenarios that we believe make sense for such a feature that we would like to discuss, to set a direction for this component. Some examples are:
1. The lifecycle of the PVCViewer based on the lifecycle of the underlying PVC
2. Culling via HTTP metrics (we've used Istio for this)
3. The RWO scheduling and best practices around it (i.e. in this PR and in TensorBoard we directly set the `nodeName`, which we should avoid)

Some of them are interesting to document, like culling, since they can affect cluster scale up and scale down. Also a design doc will help with exposing information on how admins can configure different aspects of this feature.

kimwnasptd:
So, to summarize, I propose the following for this PR:
1. Let's put it on a hold, until @apo-ger, you and I prepare the design doc PR, which will describe the architecture of this component and set direction
2. We review and merge the design doc
3. We refactor this PR to align with the design doc
    * Note that the goal is not to "block" this PR with the doc. We won't be expecting this PR to handle all aspects of the doc
    * The doc will help with having a context around the review, so that we can agree on changes that align with the direction we want to go
4. Break down this PR into smaller ones. We could first handle the controller and CRD and then how to integrate with the Volumes web app

@TobiasGoerke WDYT?

TobiasGoerke:
Thank you for the feedback. I was familiar with the initial work done by you and Arrikto and used the resources linked as a point of reference. My implementation aimed to extend the PoC by addressing its shortcomings and providing a well-tested and documented application that is suitable for general use within Kubeflow. There are some design differences, such as scheduling and a more generic API design. I agree that creating a design doc to document these decisions and incorporate lessons learned would be beneficial. I am willing to contribute to the doc, as I have also worked extensively with this feature over the past two years. Splitting the PR into smaller chunks is also fine for me. I will proceed with your proposal and look forward to seeing the first version of the design doc before we continue with reviewing the documentation and this PR. Please let me know if I can assist in any way.

kimwnasptd:
I think the current API is very powerful actually! Internally we had used a custom web server for showing the filesystem, so we didn't need to expose a lot of fields in the CR. But if we'd want to allow folks to pick and choose their app then we'll definitely need to make certain info configurable (PodSpec, VirtualService etc).

Although it would be a nice UX discussion to see if we could have a more Kubeflow "native" web app - cc @StefanoFioravanzo. For example one that could spin up a TensorBoard instance from a specific folder in a PVC (just brainstorming), via a right-click menu.

@TobiasGoerke, @apo-ger has created a first iteration of a design doc so that we can start discussing about the API. We had some questions from this PR, mostly about the restarting functionality. Once we are aligned we can then go forward and prepare the PR to add this to `/components/proposals`
https://docs.google.com/document/d/1gJTWe_Ecq_pKe5UP_sTZkSsADwfcRVHmG70bB2a4-CI/edit

If you could join one of the Notebooks WG meetings to discuss live it would also really help pushing this forward faster! But async communication is fine as well

TobiasGoerke:
> I think the current API is very powerful actually! Internally we had used a custom web server for showing the filesystem, so we didn't need to expose a lot of fields in the CR. But if we'd want to allow folks to pick and choose their app then we'll definitely need to make certain info configurable (PodSpec, VirtualService etc).
> 
> Although it would be a nice UX discussion to see if we could have a more Kubeflow "native" web app - cc @StefanoFioravanzo. For example one that could spin up a TensorBoard instance from a specific folder in a PVC (just brainstorming), via a right-click menu.
> 
> @TobiasGoerke, @apo-ger has created a first iteration of a design doc so that we can start discussing about the API. We had some questions from this PR, mostly about the restarting functionality. Once we are aligned we can then go forward and prepare the PR to add this to `/components/proposals` https://docs.google.com/document/d/1gJTWe_Ecq_pKe5UP_sTZkSsADwfcRVHmG70bB2a4-CI/edit
> 
> If you could join one of the Notebooks WG meetings to discuss live it would also really help pushing this forward faster! But async communication is fine as well

That's a great and detailed proposal. Thank you both!
I have commented on some details and am happy with this proposal's direction.

I'll try and join the Notebooks WG Meeting tomorrow. See you then!

TobiasGoerke:
@kimwnasptd just tested whether the node of a RWO PVC is permanently determined by its first consumer. The answer is: **no**, the PVC can always be mounted to different nodes, once all Pods released their handle on the PVC.

Here's an example for you to test for yourself:
1. Deploy the following yaml:
    ```apiVersion: apps/v1
    kind: Deployment
    metadata:
      name: restart-test
      labels:
        app: restart-test
    spec:
      replicas: 1
      selector:
        matchLabels:
          app: restart-test
      template:
        metadata:
          labels:
            app: restart-test
        spec:
          nodeName: {NODE01}
          containers:
          - args:
            - /bin/bash
            - -c
            - echo "Started!" && sleep Infinity
            image: python:3
            name: main
            volumeMounts:
              - name: rwo-pvc
                mountPath: /data
          volumes:
            - name: rwo-pvc
              persistentVolumeClaim:
                claimName: rwo-pvc
    ---
    apiVersion: v1
    kind: PersistentVolumeClaim
    metadata:
      name: rwo-pvc
    spec:
      accessModes:
      - ReadWriteOnce
      resources:
        requests:
          storage: 1Gi
      storageClassName: {YOUR_RWO_SC}
    ```
2. Make sure the Pod is starting on the correct node. Command will return the first node you set in the yaml.
    ```kubectl get po -l app=restart-test -ojsonpath="{.items[*].spec.nodeName}"``` 
4. Patch the deployment / let the Pod run on another node:
    ```kubectl patch deploy restart-test -p '{"spec": {"template": {"spec": {"nodeName": "{NODE02}"}}}}'```. 
6. The new Pod won't start as strategy is "RollingUpdate" and there can't be two Pods on different Nodes mounting the same RWO PVC: `Non-shared volume is already attached on another node`
7. Delete all ReplicaSets and wait for the two old Pods to terminate before the new one is running: 
    ```kubectl delete rs -l app=restart-test```
9. See the new Pod running on another Node having mounted the PVC. The following command will return Node2
    ```kubectl get po -l app=restart-test -ojsonpath="{.items[*].spec.nodeName}"```

apo-ger:
Hey @TobiasGoerke, I and @kimwnasptd are still working on the proposal of the design doc.  In the end we realized that there is a lot of context to expose for culling and a lot of this logic will be the same for notebooks https://github.com/kubeflow/kubeflow/pull/6927 so we want to have a distinct design doc.

TobiasGoerke:
I'd love to continue working on this feature and integrate it into Kubeflow. Now that 1.7 is released and planning for 1.8 is advancing: what's your take/timeline on this proposal, @kimwnasptd @apo-ger?

TobiasGoerke:
As suggested in #7125, I'll gladly take on pushing the PVCViewer to make it eligible for the KF 1.8 release. 

I've just reworked this PR to reflect @apo-ger's great proposal.

@kimwnasptd I'd appreciate a review and guidance on how to proceed with this PR and the remaining tasks. 

Among them are: 
- [ ] Provide a public controller image
- [ ] Create a separate PR with the changes to the volumes UI
- [ ] Submit the controller to the manifests

kimwnasptd:
@TobiasGoerke thanks for keep pushing this!

I'd propose we break down the next steps like this:

1. We finalize the review and merge this PR
2. We work on a follow up PR that creates a GH Action for publishing the image to DockerHub, like https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/nb_controller_docker_publish.yaml
    * We can use the name `kubeflownotebookswg/pvcviewer-controller`
3. We create a follow up PR that updates the manifests with that image
4. We create a follow up PR that tests the manfiests are OK https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/nb_controller_intergration_test.yaml
5. Create a PR for running the unit-tests of the Controller https://github.com/kubeflow/kubeflow/blob/master/.github/workflows/notebook_controller_unit_test.yaml

components/pvc-viewer/Makefile
```diff
@@ -0,0 +1,157 @@
...
+
+.PHONY: run
+run: manifests generate fmt vet ## Run a controller from your host.
+	go run ./main.go
+
+# If you wish built the manager image targeting other platforms you can use the --platform flag.
+# (i.e. docker build --platform linux/arm64 ). However, you must enable docker buildKit for it.
+# More info: https://docs.docker.com/develop/develop-images/build_enhancements/
+.PHONY: docker-build
+docker-build: test ## Build docker image with the manager.
```
kimwnasptd, line 75: Let's not run the `test` rule when building. We'll have a dedicated action that will be running the tests, and setting up the environment for running them.

This way the GH Action that will be responsible for building the image will not need to have golang setup as well

kimwnasptd (commented):
Made a first pass. The code looks very good and it's great to see test coverage! I've left some small comments.

Would there be an "easy" way to generate some certs (Makefile rule?) so that we can also `make run` the controller locally?

Lastly, once we finish with the tasks on manifests/gh-actions, to ensure this component is integrated with the repo, I think the next items to tackle will be:
1. Integrating with the volumes web app. I can help there since we had some code for this in the old Arrikto Rok flavor of the app (in this repo)
2. Work on the culling mechanism, which would be very similar with https://github.com/kubeflow/kubeflow/pull/6927

TobiasGoerke:
Thank you for the review - it's much appreciated! I've tried to integrate all of your remarks.

> * We work on a follow up PR that creates a GH Action for publishing the image to DockerHub
> * We create a follow up PR that tests the manfiests are OK
> * Create a PR for running the unit-tests of the Controller

I've prepared a new branch with the three untested GH actions that we can create a PR off, once this PR is merged. See [here](https://github.com/kubeflow/kubeflow/commit/d1eecf6d26d83ece645d116e664f0743e5ace3f8).


> Would there be an "easy" way to generate some certs (Makefile rule?) so that we can also make run the controller locally?

`make run` now executes `openssl` to generate certs and thus, the controller starts successfully (given you've applied the CRDs). Is this what you've meant?

Regarding the volumes web app:
I've created a separate [branch](https://github.com/TobiasGoerke/kubeflow/tree/feature/volumes-viewer-frontend-integration) some time ago that contains the frontend changes. We can use it as a base and create a PR off of it, once the controller is merged. You can then review/rework it if that's okay for you.

About culling: would it make sense to move this functionality to a common place so that both the notebook and pvcviewer controller can use it?

kimwnasptd:
> make run now executes openssl to generate certs and thus, the controller starts successfully (given you've applied the CRDs). Is this what you've meant?

Hmm are you sure about this? Could you point me to this part on the Makefile? For me when I `make run` I get the following output:
```
1.6866788816545787e+09	ERROR	setup	problem running manager	{"error": "open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"}
```

> About culling: would it make sense to move this functionality to a common place so that both the notebook and pvcviewer controller can use it?

Yes! And we have a place for common controller code in https://github.com/kubeflow/kubeflow/tree/master/components/common

> I've created a separate [branch](https://github.com/TobiasGoerke/kubeflow/tree/feature/volumes-viewer-frontend-integration) some time ago that contains the frontend changes. We can use it as a base and create a PR off of it, once the controller is merged. You can then review/rework it if that's okay for you.

ACK!

components/pvcviewer-controller/main.go
```diff
@@ -0,0 +1,119 @@
...
+	var enableLeaderElection bool
+	var probeAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
+		"Enable leader election for controller manager. "+
+			"Enabling this will ensure there is only one active controller manager.")
+	opts := zap.Options{
+		Development: true,
+	}
```
kimwnasptd, line 62: Let's make this instead
```golang
	opts := zap.Options{
		Development: true,
		TimeEncoder: zapcore.RFC3339TimeEncoder,
	}
```
So that the timestamps are shown in RFC3339 format

TobiasGoerke:
> > make run now executes openssl to generate certs and thus, the controller starts successfully (given you've applied the CRDs). Is this what you've meant?
> 
> Hmm are you sure about this? Could you point me to this part on the Makefile? For me when I `make run` I get the following output:
> 
> ```
> 1.6866788816545787e+09	ERROR	setup	problem running manager	{"error": "open /tmp/k8s-webhook-server/serving-certs/tls.crt: no such file or directory"}
> ```
Sorry, forgot to push that [commit](https://github.com/kubeflow/kubeflow/pull/6876/commits/9aeda40d65308a8dedca2a7c4f7111e411aff71c). It should look like this now:
image

> > About culling: would it make sense to move this functionality to a common place so that both the notebook and pvcviewer controller can use it?
> 
> Yes! And we have a place for common controller code in https://github.com/kubeflow/kubeflow/tree/master/components/common

This would involve refactoring the notebooks controller and move some code to the common dir. I'd suggest to implement culling in a separate PR to make the changes more maneagable. WDYT?

> > I've created a separate [branch](https://github.com/TobiasGoerke/kubeflow/tree/feature/volumes-viewer-frontend-integration) some time ago that contains the frontend changes. We can use it as a base and create a PR off of it, once the controller is merged. You can then review/rework it if that's okay for you.
> 
> ACK!

components/pvcviewer-controller/main.go
TobiasGoerke: Done

kimwnasptd:
Now everything seems to work as expected, thanks!

> This would involve refactoring the notebooks controller and move some code to the common dir. I'd suggest to implement culling in a separate PR to make the changes more maneagable. WDYT?

For sure, let's not increase the scope of this PR any longer. We'll most probably need to create a design doc for https://github.com/kubeflow/kubeflow/pull/6927, since it has quite a lot of technical context and decisions in it.

kimwnasptd:
So at this point everything looks good so we can merge. We also do have a plan forward for the rest of the items (images, gh actions etc) so looking forward to next steps!

Thank you very much for your time and persistence on this @TobiasGoerke

/lgtm
/approve

components/pvcviewer-controller/Makefile
```diff
@@ -0,0 +1,172 @@
...
+# go-get-tool will 'go get' any package $2 and install it to $1.
+PROJECT_DIR := $(shell dirname $(abspath $(lastword $(MAKEFILE_LIST))))
+define go-get-tool
+@[ -f $(1) ] || { \
+set -e ;\
+TMP_DIR=$$(mktemp -d) ;\
+cd $$TMP_DIR ;\
+go mod init tmp ;\
+echo "Downloading $(2)" ;\
+GOBIN=$(PROJECT_DIR)/bin go get $(2) ;\
```
kimwnasptd, line 169: Mentioning it now, to just keep it in our cache. When I initially tried to get the controller-gen bin with Golang 1.20 it failed to create the binary.

After looking around I found out that `go get` gets deprecated in favor of `go install`. Install worked in my case, but I didn't look too much into it https://github.com/grafana/k6-operator/issues/104#issuecomment-1140353055

Let's keep it in our cache for the future

TobiasGoerke:
Great to see this merged, thank you for your support!
I'll create some follow-up PRs over the next week with GH actions. 
Also, I'll continue working on the frontend integration.

components/pvcviewer-controller/Makefile
TobiasGoerke: Yeah, I've had the same issue but didn't want to change the Makefile as the other components, such as the notebook controller, have the same issue..

PR #6882, opened 2023-01-09
gh-actions: Create testing setup for CDB tests

kimwnasptd:
In this commit we extend workflow for running CDB (Angular) frontend tests. For the CentralDashboard (Angular) tests we also need to ensure some underlying apps are proxied.

In this commit we extended the dedicated workflow for the frontend tests by:
1. Applying the necessary KF Controllers (Profiles, Notebooks)
2. Applying the JWA manifests and port-forward its k8s Service
3. Applying the VWA manifests and port-forward its k8s Service
4. Creating a user namespace (kubeflow-user) via a Profile
5. Creating a test Notebook, with a PVC, in that namespace

Next steps will be to run the Cypress tests of the CDB (Angular) that will be using the proxied apps.

/cc @orfeas-k 
/cc @tasos-ale 
/cc @apo-ger

### DIFF
--- a/.github/workflows/centraldb_angular_frontend_test.yaml
+++ b/.github/workflows/centraldb_angular_frontend_test.yaml
@@ -42,3 +42,63 @@ jobs:
           cd components/centraldashboard-angular/frontend/
           npm i
           npm run test:prod
+
+  run-tests-in-chrome:
+    name: UI tests in chrome
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Setup node version to 12
+        uses: actions/setup-node@v3
+        with:
+          node-version: 12
+
+      - name: Install KinD
+        run: ./components/testing/gh-actions/install_kind.sh
+
+      - name: Create KinD Cluster
+        run: kind create cluster --config components/testing/gh-actions/kind-1-25.yaml
+
+      - name: Install kustomize
+        run: ./components/testing/gh-actions/install_kustomize.sh
+
+      - name: Install Istio
+        run: ./components/testing/gh-actions/install_istio.sh
+
+      - name: Apply KF Controllers
+        run: |
+          cd components
+          kubectl create namespace kubeflow
+          kustomize build profile-controller/config/overlays/kubeflow | kubectl apply -f -
+          kustomize build notebook-controller/config/overlays/kubeflow | kubectl apply -f -
+          kubectl wait pods -n kubeflow -l kustomize.component=profiles --for=condition=Ready --timeout=300s
+          kubectl wait pods -n kubeflow -l app=notebook-controller --for=condition=Ready --timeout=300s
+
+      - name: Apply JWA manifests
+        run: |
+          cd components/crud-web-apps/jupyter/manifests
+          kustomize build overlays/istio | kubectl apply -f -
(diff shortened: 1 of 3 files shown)


kimwnasptd:
The error is because there's no image `docker.io/kubeflownotebookswg/kfam:latest`. Looks like the GH Actions didn't end up pushing a `latest` image for KFAM. I'll manually push the first image for KFAM and re-trigger the tests

kimwnasptd:
@orfeas-k @apo-ger the PR is ready for review. I've added a last step for testing the connection to the web apps, just for ensuring everything works. Also I created a dummy TEST file in the frontend to trigger the tests. 

I'll remove it once it's ready for merge

orfeas-k (commented):
Tried applying the `test-notebook.yaml` and running UI tests with the test Notebook and PVC and it seems that it works fine. Rest of it looks good to me. 

A minor question/note, are we going to run tests in both Chrome and Firefox using the same `job` or are we going to use a second job (and thus setup a second KinD cluster) for Firefox?

.github/workflows/centraldb_angular_frontend_test.yaml
```diff
@@ -42,3 +42,63 @@ jobs:
           cd components/centraldashboard-angular/frontend/
           npm i
           npm run test:prod
+
+  run-tests-in-chrome:
+    name: UI tests in chrome
+    runs-on: ubuntu-20.04
```
apo-ger, line 48: Should we use `runs-on: ubuntu-latest`  GitHub-hosted runner similar to the other jobs ([here](https://github.com/arrikto/kubeflow/blob/84afe05474d5fea0ff4adac275fc4ec3e7013e3c/.github/workflows/centraldb_angular_frontend_test.yaml#L10) and [here](https://github.com/arrikto/kubeflow/blob/84afe05474d5fea0ff4adac275fc4ec3e7013e3c/.github/workflows/centraldb_angular_frontend_test.yaml#L34))?

kimwnasptd: This is mostly because we wanted to make sure that the base would always contain both Firefox and Chrome browsers. Although indeed now even `ubuntu-22.04` has the Firefox installation we wanted to make sure the tests wouldn't break from a change in ubuntu latest.

And we'll be periodically updating the bases of these actions
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md#browsers-and-drivers
https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers

kimwnasptd:
> A minor question/note, are we going to run tests in both Chrome and Firefox using the same job or are we going to use a second job (and thus setup a second KinD cluster) for Firefox?

@orfeas-k that's a good argument. Looking at the current Actions we don't do anything "browser specific" in the setup code. So we could have just one job and one step for each browser. WDYT?

orfeas-k:
@kimwnasptd Sounds good since the setup is the same for both browsers, thank you!
/lgtm

kimwnasptd:
@elikatsis we should be good to go with this PR

kimwnasptd:
Force pushed to remove the DROP commit, that was just to trigger the actions and test everything worked as expected

elikatsis:
Cool, thanks!
/lgtm
/approve

PR #6887, opened 2023-01-11
[Profile controller IAM plugin] Support annotation only

ryansteakley:
AWS IAM service account profile plugin mutates the trust relationship in the iam role specified in the spec. 

Customers have different ways to manage roles and their company policies might not allow Kubeflow profile controller to allow changing the trust relationship. 

Adding a boolean variable to the AwsIAMForServiceAccount struct to enable users to specify if they want AnnotateOnly which will cause IAM roles and policy will not be mutated.

This change will not effect existing users who do not have the boolean field in their yaml files.

### DIFF
--- a/components/profile-controller/controllers/plugin_iam.go
+++ b/components/profile-controller/controllers/plugin_iam.go
@@ -28,7 +28,8 @@ const (
 )
 
 type AwsIAMForServiceAccount struct {
-	AwsIAMRole string `json:"awsIamRole,omitempty"`
+	AwsIAMRole           string `json:"awsIamRole,omitempty"`
+	AnnotateOnly bool   `json:"AnnotateOnly,omitempty"`
 }
 
 // ApplyPlugin annotate service account with the ARN of the IAM role and update trust relationship of IAM role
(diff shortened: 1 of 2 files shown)


kimwnasptd:
If it's good from AWS' side, since it only touches this functionality, I'm good with merging this

components/profile-controller/controllers/plugin_iam.go
```diff
@@ -47,6 +54,12 @@ func (aws *AwsIAMForServiceAccount) RevokePlugin(r *ProfileReconciler, profile *
 	if err := aws.patchAnnotation(r, profile.Name, DEFAULT_SERVICE_ACCOUNT, removeIAMRoleAnnotation, logger); err != nil {
 		return err
 	}
+
+	if aws.AnnotateOnly {
+		logger.Info("DisableAutoUpdateIAM set to true IAM roles and policy will not be mutated")
```
surajkota, line 59: AnnotateOnly set to true

surajkota (changes requested):
looks good, can you add a unit test?

kimwnasptd:
Thanks for your time @ryansteakley @surajkota!

/lgtm
/approve

surajkota:
Thanks

PR #6890, opened 2023-01-12
rstudio: Update version to properly handle X-Forwarded headers

kimwnasptd:
Update the RStudio image to be able to handle multiple values in the `X-Forwarded-{Proto,Host}` headers. The server should care about the first values which are the ones the user's browser is using.

The fixes for these have been integrated in https://github.com/rstudio/rstudio/issues/11010 and https://github.com/rstudio/rstudio/issues/10965

Note, to bump into this you need to have at least 2 proxies that handle the `X-Forwarded-*` headers. In our case we had both an ALB loadbalancer and an NGINX in front of the ingressgateway, and that's why we bumped into this.

### DIFF
--- a/components/example-notebook-servers/rstudio/Dockerfile
+++ b/components/example-notebook-servers/rstudio/Dockerfile
@@ -6,13 +6,13 @@ USER root
 
 # args - software versions
 ARG MINIFORGE_ARCH="x86_64"
- # renovate: datasource=github-tags depName=conda-forge/miniforge versioning=loose
+# renovate: datasource=github-tags depName=conda-forge/miniforge versioning=loose
 ARG MINIFORGE_VERSION=4.10.1-4
 ARG PIP_VERSION=21.1.2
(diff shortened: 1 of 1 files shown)


kimwnasptd:
As a side note, I noticed that no tests are triggered for testing if the images can be built. 

@apo-ger did we revert this with the latest work on CI for using the `latest` tag and breaking down the workflows?

apo-ger:
> As a side note, I noticed that no tests are triggered for testing if the images can be built.
> 
> @apo-ger did we revert this with the latest work on CI for using the `latest` tag and breaking down the workflows?
@kimwnasptd Yes we refactored all `docker_publish` workflows to run only on `push` events. I can prepare a PR to introduce distinct workflows for only testing that the images can be built on `pull_request` events.

apo-ger:
@kimwnasptd Changes look good! Tested the image and the issue is resolved
/lgtm

PR #6891, opened 2023-01-12
jwa(front): Add UI tests with Cypress

orfeas-k:
This PR adds more UI tests with Cypress for functionalities of the JWA frontend. More specifically, they check that:
 - Index page renders every Notebook name into the table
 - Index page shows correct Status icon for all notebooks

Everything implementation-specific (mock API, node version etc) mentioned in the [VWA UI tests PR](https://github.com/kubeflow/kubeflow/pull/6862) stand for this one too.


.github/workflows/jwa_frontend_tests.yaml
```diff
@@ -56,3 +56,65 @@ jobs:
...
+          npm i
+          npm link kubeflow
+      - name: Serve UI & run Cypress tests in Firefox
+        uses: cypress-io/github-action@v4.2.0
+        with:
+          working-directory: components/crud-web-apps/jupyter/frontend
+          start: npm run serve
+          install: false
+          browser: firefox
+          wait-on: "http://localhost:4200"
```
kimwnasptd, line 120: @orfeas-k since the setup for running the Firefox tests is exactly the same with the one for Chrome, what about moving this step into the one that runs the Chrome steps?

This way we avoid the duplicate work of running the tests twice

kimwnasptd: The downside of merging the two actions would be that if in the future we don't mock the whole backend, then we'll need to break these actions again.

But I don't think we are going to move towards this direction for at least the next 2 months

orfeas-k: Sure, I have no strong preference over this so I 'm sending a fixup!

orfeas-k: Just pushed a fixup for this @kimwnasptd

kimwnasptd: I think it's going to be a little bit involved to wait on localhost:4200 to be ready to serve `index.html`.

Let's see if we can get something to work. Else if it's too much trouble let's revert back

kimwnasptd:
Nice!

/lgtm
/approve

PR #6892, opened 2023-01-13
cdb-angular(front): Fix Iframe's sizing to fit to screen

orfeas-k:
Style the dashboard and the iframe in order for the iframe to be responsive, fit in the screen and not exceed it. 

This resolves the issue from the comment here https://github.com/kubeflow/kubeflow/pull/6856#issuecomment-1381649212 and is part of the [Central Dashboard Rewrite effort](https://github.com/kubeflow/kubeflow/issues/6000).

cc @kimwnasptd


kimwnasptd:
Thanks for the quick fix!

/lgtm
/approve

PR #6897, opened 2023-01-17
jwa(front): JWA shows a weird long value for memory when restoring notebook from snapshot 

elenzio9:
In this PR, we format the long memory values into human-readable ones so that users can easily parse them.

Previous state:
image

Current state:
image

### DIFF
--- a/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/resource-table/table/utils.ts
+++ b/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/resource-table/table/utils.ts
@@ -35,3 +35,27 @@ export function addColumn(
 function findColumnIndex(config: TableConfig, name: string) {
   return config.columns.findIndex(col => col.matColumnDef === name);
 }
+
+export function formatBytes(bytes, si = false, decimals = 1) {
+  const thresh = si ? 1000 : 1024;
+
+  if (Math.abs(bytes) < thresh) {
+    return bytes + ' B';
+  }
+
+  const units = si
+    ? ['kB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+    : ['Ki', 'Mi', 'Gi', 'Ti', 'Pi', 'Ei', 'Zi', 'Yi'];
+  let u = -1;
+  const r = 10 ** decimals;
+
+  do {
+    bytes /= thresh;
+    ++u;
+  } while (
+    Math.round(Math.abs(bytes) * r) / r >= thresh &&
+    u < units.length - 1
+  );
+
+  return bytes.toFixed(decimals) + ' ' + units[u];
+}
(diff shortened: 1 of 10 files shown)


jbottum:
I look forward to this enhancement.

orfeas-k:
Taking a look at this. At first, could you rebase in order to resolve the conflicts? @elenzio9

components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/resource-table/table/utils.ts
```diff
@@ -58,4 +58,99 @@ export function formatBytes(bytes, si = false, decimals = 1) {
...
+    case 'k':
+      return Number(quantity.substr(0, quantity.length - 1)) * Number(1000);
+    case 'M':
+      return (
+        Number(quantity.substr(0, quantity.length - 1)) * Number(1000 * 1000)
+      );
+    case 'G':
+      return (
+        Number(quantity.substr(0, quantity.length - 1)) *
+        Number(1000 * 1000 * 1000)
```
orfeas-k, line 104: @elenzio9 Nit but is there a reason for using `*` operator multiple times instead of the power one `**`? If not, I think `**` is more intuitive for this case (e.g. `1000 ** 3`)

orfeas-k (commented):
Really great work @elenzio9! Other than the small comment I included, the rest of the code looks good. I 'll go ahead and try the JWA and VWA images on my cluster to test everything works as expected.

components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/resource-table/table/utils.ts
elenzio9: The `quantityToScalar()` is based on this [function](https://github.com/kubernetes-client/javascript/blob/14b6d48151f838df6c395472609e486ebed6c100/src/util.ts#L17) of `kubernetes-client` library. We only needed to replace `function`'s `BigInt()` with just `Number()`, since the `BigInt()` doesn't accept floats and in our case users can insert such numbers. 

However, it doesn't cause precision problems, as the [maximum safe integer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) which can be represented by the `number` primitive is `2^53 – 1` and we are not interested in numbers larger than that. 

@orfeas-k For this reason, I'd suggest keeping it as is.

elenzio9:
> could you rebase in order to resolve the conflicts?

Just rebased!

components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/resource-table/table/utils.ts
orfeas-k: Ok I see, that's fine by me then. Thanks for the context @elenzio9

orfeas-k:
Tested the images and the feature works as expected, thank you @elenzio9 
/lgtm

cc @kimwnasptd

elenzio9:
@orfeas-k I made one more change [here](https://github.com/kubeflow/kubeflow/pull/6897/files#diff-630e88e6df7f6856a0d278c646de4f34e58e285f29959cc0e90ad3449e184c20) and pushed again. Could you take a look please?

For context, in the past we needed to override the default `dataSource.sortData` used by `MatTableDataSource` class to consider both values of type `Number` and `BigInt`. Since we've decided that the `number` primitive covers us, I've removed that part of the code.

PR #6902, opened 2023-01-20
feat: allow setting `ISTIO_HOST` in notebook/tensorboard controller for Istio VirtualServices

punkerpunker:
Hi there,

This PR intends to give an opportunity to explicitly specify Istio VirtualServer's host. 

Currently, the VirtualService created by either `tensorboard-controller` or a `notebook-controller` is defined with an all-inclusive wildcard:  

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata
  ...
spec:
  gateways:
  - kubeflow/kubeflow-gateway
  hosts:
  - "*"
```

Which may end up creating excessive Envoy routing rules, when, for example, the ingress is being used not only for running Kubeflow (e.g in the case where you want to use the same ingress for all third-level domains, e.g `kubeflow.myorg.com`, `grafana.myorg.com`). 

Having the opportunity to explicitly specify the host is going to significantly reduce Envoy's overhead when many endpoints are being attached to the same Ingress Envoy (e.g `istio-ingressgateway`) instance. 

With the changes in PR, you can explicitly specify a host that you're expecting to be used with notebooks and tensorboards, removing the all-inclusive hosts definition. 

For example, with `ISTIO_HOST=kubeflow.myorg.com`, the resulting `VirtualService` will look like: 

```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata
  ...
spec:
  gateways:
  - kubeflow/kubeflow-gateway
  hosts:
  - "kubeflow.myorg.com"
```

And won't create excessive Envoy routing rules related to other domain names attached to the same ingress.

I've tested the changes on the cluster I am operating on and it seems working properly. 

I am not that familiar with the kubeflow source codebase, so I'd appreciate any feedback and will try my best to make this PR merged. Thank you. 

This is similar to what's introduced in this PR: https://github.com/kubeflow/kubeflow/pull/6247, but it fixes it both for the `tensorboard-controller` and a `notebook-controller`


punkerpunker:
cc @elikatsis @kimwnasptd @MartinForReal can we get this one merged? For now, I have to manually patch the controllers every time I need an upgrade. Thanks!

juliusvonkohout:
Ah yes. We encounter similar Problems . For large instances envoy updates are becoming very expensive. I also have an engineer that wants to decouple Istio a bit or make it more compatible with other istio users next to kubeflow. This is one step in the right direction. Please join the next kubeflow WG meeting. You can see it in the KF Calendar  here https://www.kubeflow.org/docs/about/community/. Hopefully the next one is on July 6.  Please add yourself to the agenda and ping @kimwnasptd (Kimonas Sotirchos) on slack. Please also join the wg-notebooks channel on slack. You can see the agenda in the calendar and here https://docs.google.com/document/d/1zPIaicGLC3k9AlAdBOn6cp-ojg2KqvmltLXCBr3CGaY/edit. Kimonas had to restructure it so i am not 100 % sure whether this meeting will take place.

thesuperzapper:
@punkerpunker, btw, there are a number of other VirtualServices which hard-code `"*"` as `hosts`, namely:

- [`jupyter-web-app-jupyter-web-app`](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/manifests/overlays/istio/virtual-service.yaml)
- [`volumes-web-app-volumes-web-app`](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/volumes/manifests/overlays/istio/virtual-service.yaml)
- [`tensorboards-web-app-tensorboards-web-app`](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/tensorboards/manifests/overlays/istio/virtual-service.yaml)

There are also a few from other parts of Kubflow:

- `katib-ui` (Katib)
- `metadata-grpc` (Kubeflow Pipelines)
- `ml-pipeline-ui` (Kubeflow Pipelines)

But I agree that we should fix this one, because the user can't patch it themselves as it's baked into the Notebook/Tensorboard controller.

thesuperzapper:
@kimwnasptd I am not against trying to get this in for 1.8, if you are willing to?

Because its a bit of a pain in some environments to have that wildcard host.

thesuperzapper:
@kimwnasptd Unless you disagree, I want to merge this for the 1.8.1 patch release. I have seen several people experience problems because the `spec.hosts` in the notebook VirtualServices are hard-coded as `*`.

The main issue is that if you make the central-dashboard's VirtualService non-`*` (like you might have to do if you share your gateway with non-kubeflow apps), then it will break all notebook paths. This is because istio sees a specific host as a "more specific match" than a `*`, and the central dashboard matches any prefix.

Either way @punkerpunker, this looks good to me.

I will approve it in a few days if @kimwnasptd does not respond.

/lgtm

thesuperzapper:
@punkerpunker sorry for the delay here, we were dealing with some permission issues which were preventing us cutting a new release, but they are resolved now.

This will be part of the 1.9.0-rc.0 release we cut soon.

/approve

PR #6916, opened 2023-01-30
release: Images for the v1.7.0-rc.0 tag

apo-ger:
Update the manifests to use the v1.7.0-rc.0 tag

/cc @kimwnasptd


kimwnasptd:
@apo-ger looks like the issue is with our `sed` command and how it expects a `latest` tag, which is not the case when we update the manifests for RCs.

What about using `kubectl edit image <curr_img>:<new_image>`, instead of the `sed` command?
https://github.com/kubernetes-sigs/kustomize/blob/master/examples/image.md

I.e. with something like this
```
IMG=docker.io/kubeflownotebookswg/poddefaults-webhook
kustomize edit set image ${IMG}=${IMG}:asdf
```

apo-ger:
@kimwnasptd Yes, this will work. 

I ran the `kustomize edit set image` command for the [kustomization file](https://github.com/kubeflow/kubeflow/blob/v1.7-branch/components/admission-webhook/manifests/overlays/cert-manager/kustomization.yaml) under `components/admission-webhook/manifests/overlays/cert-manager` and the following section was added:
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
images:
- name: docker.io/kubeflownotebookswg/poddefaults-webhook
  newName: docker.io/kubeflownotebookswg/poddefaults-webhook
  newTag: test
```
Then, running `kustomize build`  produced the manifests with the correct `test` tag. 


I'll prepare the PR to update all workflows.

apo-ger:
@kimwnasptd Rebased on top of the latest `v1.7-branch` to include fixes from https://github.com/kubeflow/kubeflow/pull/6917 and https://github.com/kubeflow/kubeflow/pull/6922. All intergration tests ran successfully. Nice!

kimwnasptd:
Awesome! We are good to go with the first RC then 🎉 

/lgtm
/approve

PR #6917, opened 2023-01-30
gh-actions: Fix KF intergration tests for release branches

apo-ger:
This PR addresses https://github.com/kubeflow/kubeflow/pull/6916#issuecomment-1408865920 and is part of https://github.com/kubeflow/kubeflow/issues/6766

Right now, the KF integration tests do not work when we update the manifests for RCs. The `sed` command does not work since it tries to replace the `latest` tag, which doesn't exist in the release branches.

Changes:

* Update the workflows to use the `kustomize edit image <curr_img>=<new_img>:<tag>` command (https://github.com/kubernetes-sigs/kustomize/blob/master/examples/image.md) to set the appropriate tag in the manifests before they are applied . 

### DIFF
--- a/.github/workflows/centraldb_angular_intergration_test.yaml
+++ b/.github/workflows/centraldb_angular_intergration_test.yaml
@@ -41,11 +41,12 @@ jobs:
 
     - name: Build & Apply manifests
       run: |
-        cd components/centraldashboard-angular/manifests
+        cd components/centraldashboard-angular/manifests/overlays/kserve
         kubectl create ns kubeflow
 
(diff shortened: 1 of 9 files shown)


apo-ger:
The `PodDefaults Intergration Test` that needs to install Cert Manager was failing with the following:
```
chart requires kubeVersion: >= 1.21.0-0 which is incompatible with Kubernetes v1.20.0
```
We resolved this in: https://github.com/kubeflow/kubeflow/pull/6922

kimwnasptd:
Nice!

/lgtm
/approve

PR #6918, opened 2023-01-30
WIP: cdb-angular: Show KF version

orfeas-k:
This PR is part of the [Rewrite the KF Central Dashboard in Angular](https://github.com/kubeflow/kubeflow/issues/6000).


orfeas-k:
cc @kimwnasptd

tasos-ale (commented):
@orfeas-k I believe with a few changes in the [common snackbar](https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/snack-bar),  we can use it instead of having to duplicate the code in our new dashboard.

components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.scss
```diff
@@ -16,8 +16,34 @@
...
+  display: none;
+}
+
+.footer-text:hover .buildVersion {
+  display: none;
+}
+
+.footer-text:hover .buildId {
+  display: inline;
+}
```
tasos-ale, line 49: Since we are using Sass, this is a good use case to use nesting. Check the nesting section [here](https://sass-lang.com/guide).

tasos-ale (changes requested):

components/centraldashboard-angular/frontend/scripts/update-environment-prod-ts.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs');
+
+const args = process.argv.slice(2);
+const file = args[0];
+const build_version = args[1];
+
+let fileString = fs.readFileSync(file).toString();
+fileString = fileString.replace('${BUILD_VERSION}', build_version);
+
+fs.writeFileSync(file, fileString);
```
tasos-ale, line 10: Let's throw an error if `file` or `build_version` is undefined. Maybe we can also change the script's file name since it can be used in other files and not only for `environment.prod.ts`.

components/centraldashboard-angular/frontend/src/app/services/backend.service.ts
```diff
@@ -0,0 +1,98 @@
...
+      return `${error.status}: ${error.message}`;
+    }
+
+    if (error instanceof ErrorEvent) {
+      return error.message;
+    }
+
+    return `Unexpected error encountered`;
+  }
+}
```
tasos-ale, line 98: @orfeas-k I would prefer to move this code to an interceptor and keep this service cleaner. You will also don't need to remember to add `catchError(error => this.handleError(error))` in every new request added to the `CDBBackendService` service.

components/centraldashboard-angular/frontend/src/styles.scss
```diff
@@ -7,3 +7,7 @@ body {
   margin: 0;
   font-family: Roboto, 'Helvetica Neue', sans-serif;
 }
+
+// Use the material icons offline
+$material-icons-font-path: '~material-icons/iconfont/';
```
tasos-ale, line 12: I am not sure we need that line, but let's see what will change when you add the common library.

components/centraldashboard-angular/frontend/src/app/types/env-info.ts
```diff
@@ -0,0 +1,17 @@
+export interface envInfo {
+  user?: string;
+  platform?: {
+    provider?: string;
+    providerName?: string;
+    buildLabel?: string;
+    buildVersion?: string;
+    buildId?: string;
+    kubeflowVersion?: string;
```
tasos-ale, line 9:
```suggestion
```

tasos-ale (changes requested):

components/centraldashboard-angular/frontend/src/styles.scss
orfeas-k: Indeed, it doesn't seem like we need it, I 'll remove it!

orfeas-k:
> I believe with a few changes in the [common snackbar](https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/snack-bar), we can use it instead of having to duplicate the code in our new dashboard.

We moved forward with this implementation after all. Discussing it with @tasos-ale, we decided to change `SnackBarService` and make it more configurable and future-proof. Passing 2 separate variables doesn't sound like a good practice when `MatSnackBar` offers many more than that. Instead we will move towards passing a `config` object as a parameter that will be able to include all different configurations `SnackBarService` may need. In this direction and since every WA might need a different configuration, a structured way to approach this would be to initialize a default configuration app-wide for values like `HorizontalPosition` and `duration`, and then each time an app uses using `snackBar.open()`, it will only pass the required extra configuration like Error message and Error type. This becomes possible by providing in every WA the `MAT_SNACK_BAR_DEFAULT_OPTIONS` and injecting it to the `SnackBarService`, as it is also done in Angular https://github.com/angular/components/blob/950d4377f4afd6105b19d5d0573da3e8e8413496/src/material/snack-bar/snack-bar.ts. This way, we will also be able to use the `panelClass` field of a `config` object, which allows for styling the `SnackBar` popup. This means that this PR will also configure JWA, VWA and TWA for using the new SnackBar service and there will follow-up PRs for KWA and MWA.


@tasos-ale I 'm done with pushing changes.

tasos-ale (changes requested):
@orfeas-k except for a few suggestions I don't have anything else to add. Squash those fixups and we can merge.

orfeas-k:
### Updates on implementation

#### update-environment-prod-ts.js
We went ahead and made the above function more generic, thus it was also renamed `replace-string-in-file.js`. Now the script replaces strings between `@` in a text file, according to arguments passed. It now expects 3 arguments in the following order: File name, String that should be replaced, New string.

#### Error interceptor
We introduced an  [Error interceptor](https://angular.io/api/common/http/HttpInterceptor) in order to handle errors in a uniform way and from one place for the whole WA.

#### Fonts

We didn't use `copyLibAssets` as we do in other KF WAs since we noticed that importing into CDB's `styles.css` the KF common's styles and fonts like this was enough

```
@import '~kubeflow/styles.scss';
@import '~kubeflow/lib/fonts.scss';
```


#### GH action integration test

The `replace-string-in-file.js` uses the `kubeflowVersion` passed to `docker-build` from the Makefile and expects it not to be empty. When Makefile was doing a `git describe`, git returned 
```
fatal: No names found, cannot describe anything.
```
and that resulted in an empty `kubeflowVersion` passed to the script. 
For that, we had to change the GH checkout action `fetch-depth` to "unshallow" the repo's clone.

kimwnasptd:
Looks great, thanks for the review @tasos-ale!

/lgtm
/approve

PR #6921, opened 2023-01-30
Update GH Action script for installing Cert Manager

kimwnasptd:
This should fix the problem we bumped into with using `cmctl` to generate the manfiests
```
chart requires kubeVersion: >= 1.21.0-0 which is incompatible with Kubernetes v1.20.0
```

We've seen this error in a couple of PRs https://github.com/kubeflow/kubeflow/pull/6893 https://github.com/kubeflow/kubeflow/pull/6917

Instead we'll use the generated `cert-manager.yaml` file that comes with each Cert Manager release. I.e.
https://github.com/cert-manager/cert-manager/releases/tag/v1.11.0

/cc @apo-ger

PR #6893, opened 2023-01-14
GH Action for Admission webhook unit tests

deepk2u:
@kimwnasptd as per your request (https://github.com/kubeflow/kubeflow/pull/6749#issuecomment-1381745589) I have added GH Action for Admission webhook unit tests

PR #6922, opened 2023-01-30
Update GH Action script for installing Cert Manager

apo-ger:
This PR is based on the initial work of @kimwnasptd: https://github.com/kubeflow/kubeflow/pull/6921

This should fix the problem we bumped into with using cmctl to generate the manfiests
```
chart requires kubeVersion: >= 1.21.0-0 which is incompatible with Kubernetes v1.20.0
```
We've seen this error in a couple of PRs https://github.com/kubeflow/kubeflow/pull/6893 https://github.com/kubeflow/kubeflow/pull/6917

Instead we'll use the generated cert-manager.yaml file that comes with each Cert Manager release. I.e.
https://github.com/cert-manager/cert-manager/releases/tag/v1.11.0

### DIFF
--- a/components/testing/gh-actions/install_cert_manager.sh
+++ b/components/testing/gh-actions/install_cert_manager.sh
@@ -1,14 +1,12 @@
 #!/bin/bash
 set -e
-echo "Installing cert-manager ..."
+echo "Fetching cert-manager manifests..."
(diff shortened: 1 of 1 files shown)


kimwnasptd:
Thanks for the quick work!

/lgtm
/approve

Issue #6940, opened 2023-02-06
Support dynamic logout URLs in CentralDashboard


kimwnasptd:
/kind bug

We just merged for KF 1.7 RC0 the update on AuthService https://github.com/kubeflow/manifests/pull/2150. But in this commit the logout URL is `/authservice/logout`, instead of `/logout`. But since we can't hardcode this value, since not everyone might be using AuthService, we'll need to have a dynamic mechanism for learning this URL.

I'd propose to
1. have a new ENV Var `LOGOUT_URL` in the Deployment of the app
2. The backend will add this for the `/api/workgroup/env-info` route handler https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/app/api_workgroup.ts
3. The frontend will dynamically use this value from the backend

This way users can configure the app based on the authentication service that they use in a dynamic way

kimwnasptd:
There's already an in-flight PR that tries to fix the logout behaviour https://github.com/kubeflow/kubeflow/pull/6609

We can try to fix the issue in that or create a separate one. This fix is something we'd like to track for the KF 1.7 release cc @DomFleischmann

DomFleischmann:
Thank you for the heads up @kimwnasptd , yes I believe this is something we want to get fixed for this release.

orfeas-k:
I sent the PR for this [here](https://github.com/kubeflow/kubeflow/pull/6945). 

For documentation, I followed [this](https://polymer-library.polymer-project.org/3.0/docs/devguide/data-binding) as a guideline in order to bind data between Polymer components. In Polymer, in order to bind values to a property of a child element, one has to use `dash-case` syntax instead of camel case (which is how what is being used to create the property). On top of that, in order to use a property as a DOM element's attribute, the syntax is this 
```
attribute$={{element-property}}
```

ajinkya933:
your PR routes me to `localhost:8888/authservice/site/after_logout` with a blank page as a response. Is there a way by which I can hardcode a website for example: google.com to route to after logout button is clicked (provided the logout clears all cache)

PR #6945, opened 2023-02-10
centraldashboard: Support dynamic logout URL

orfeas-k:
Use dynamic URL for the logout button that can be set by the `LOGOUT_URL` ENV variable in the deployment of the app. If it is unset, a default '/logout' string is passed from the backend to the frontend.

I documented in the issue [here ](https://github.com/kubeflow/kubeflow/issues/6940#issuecomment-1427637200)details on how I dealt with binding data between Polymer elements.

During this PR, we also had to fix a test that was broken. It seems that the test was failing because `new Date ()` added whitespace that was not equal to the space character during comparison. Instead of comparing objects of type `Date` and `string` in the expect statement, we went with comparing `Date` objects in order to ensure comparison in a uniform and type-strict way.

### DIFF
--- a/components/centraldashboard/app/k8s_service.ts
+++ b/components/centraldashboard/app/k8s_service.ts
@@ -2,7 +2,8 @@ import * as k8s from '@kubernetes/client-node';
 
 /** Retrieve Dashboard configmap Name */
 const {
-  DASHBOARD_CONFIGMAP = "centraldashboard-config"
+  DASHBOARD_CONFIGMAP = "centraldashboard-config",
+  LOGOUT_URL = '/logout'
 } = process.env;
 
 /** Information about the Kubernetes hosting platform. */
@@ -10,6 +11,7 @@ export interface PlatformInfo {
   provider: string;
   providerName: string;
   kubeflowVersion: string;
+  logoutUrl: string;
 }
 
 /**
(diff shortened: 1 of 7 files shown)


kimwnasptd:
Thanks for the PR @orfeas-k! Could you send the commit for fixing the tests as a dedicated PR? 

Looks like we are having some build issues when PRs are merged and I'd like to tackle this as a dedicated effort
https://github.com/kubeflow/kubeflow/commit/74f020e0d9c3f58712a3b466f9d1bb86c4607beb
https://github.com/kubeflow/kubeflow/commit/65e41bf28b8e79be4e1f822afe56e218c69db8a1

orfeas-k:
Sure, I also removed the tests commit from this PR

kimwnasptd:
@orfeas-k tested the PR and while the logout works, I end up seeing a blank screen. Navigating back to the root URL gets me to login, so the logout functionality works, but I think something might not be configured correctly on the AuthService's side.

I'll look more into it with @athamark. @DomFleischmann there's a chance we'll need some fix for this, but we'll comment back once we have more news

kimwnasptd:
I'll go on and merge this for now to include this in the RC1

/lgtm
/approve

PR #6953, opened 2023-02-14
cdb(front): Fetch left sidebar links

orfeas-k:
This PR is part of the [Rewrite the Central Dashboard in Angular effort](https://github.com/kubeflow/kubeflow/issues/6000) and implements the following two functionalities:

 - Fetch the left sidebar links

Here, we also had to handle the until now unhandled **fragment/hash** in links. For that, we decided to always split links to its sub-parts (`path`, `queryParams`, `fragment`) and use the `fragment` option in `router.navigate`.

 - Automatically change selected link from the sidebar

We couldn't go only with `routerLinkActive` for this and used a custom implementation instead since we need the `path + fragment` combination to be a subset of the browser's URL. In this implementation, we also added an explicit condition for checking when the user is in Recurring Runs WA since the KFP removes the 's' from its path when navigating inside a Recurring Run's details page ( `/_/pipeline/#/recurringruns` in contrast to `/_/pipeline/#/recurringrun/details/`).

### DIFF
--- a/components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
+++ b/components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
@@ -5,13 +5,14 @@ import { map, shareReplay } from 'rxjs/operators';
 import { environment } from 'src/environments/environment';
 import { CDBBackendService } from 'src/app/services/backend.service';
 import { envInfo } from '../../types/env-info';
+import { DashboardLinks, Link, MenuLink } from 'src/app/types/dashboard-links';
 
 @Component({
   selector: 'app-main-page',
   templateUrl: './main-page.component.html',
   styleUrls: ['./main-page.component.scss'],
 })
-export class MainPageComponent implements OnInit, OnDestroy {
+export class MainPageComponent implements OnInit {
   isHandset$: Observable<boolean> = this.breakpointObserver
     .observe(Breakpoints.Handset)
     .pipe(
@@ -28,7 +29,10 @@ export class MainPageComponent implements OnInit, OnDestroy {
   public get buildIdWithLabel() {
     return this.computeBuildValue(this.buildLabel, this.buildId);
   }
-  public envInfoSub: Subscription;
+  public menuLinks: MenuLink[];
+  public externalLinks: any[];
+  public quickLinks: Link[];
+  public documentationItems: Link[];
 
   constructor(
     private breakpointObserver: BreakpointObserver,
@@ -36,15 +40,13 @@ export class MainPageComponent implements OnInit, OnDestroy {
   ) {}
 
   ngOnInit() {
-    this.envInfoSub = this.backend.getEnvInfo().subscribe((res: envInfo) => {
+    this.backend.getEnvInfo().subscribe((res: envInfo) => {
       this.handleEnvInfo(res);
     });
-  }
 
-  ngOnDestroy(): void {
-    if (this.envInfoSub) {
-      this.envInfoSub.unsubscribe();
-    }
+    this.backend.getDashboardLinks().subscribe((res: DashboardLinks) => {
+      this.handleDashboardLinks(res);
+    });
   }
 
   handleEnvInfo(env: envInfo) {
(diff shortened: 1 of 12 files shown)


components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
```diff
@@ -63,4 +83,45 @@ export class MainPageComponent implements OnInit, OnDestroy {
...
+     * We need this condition because Pipelines Web App removes
+     * the 's' from its path when navigating inside a Recurring
+     * Run's details page
+     */
+    if (
+      browserUrl.includes('pipeline/#/recurringrun') &&
+      url.includes('/pipeline/#/recurringruns')
+    ) {
+      return true;
+    }
```
tasos-ale, line 123: It is ok for now, but we may need to do something better here. I would prefer not to have hardcoded values. Maybe we can have `pipeline/#/recurringrun` inside `menuLinks` and compare it with `startsWith()`.

components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
```diff
@@ -63,4 +83,45 @@ export class MainPageComponent implements OnInit, OnDestroy {
...
+
+    return browserUrl.includes(url);
```
tasos-ale, line 125: It works, but it will be safer to use `startsWith()` in case we end up with a `url` inside the `browserUrl` that we don't want to match

components/centraldashboard-angular/frontend/src/app/shared/utils.ts
```diff
@@ -22,9 +22,15 @@ export function appendBackslash(url: string): string {
 
   let urlPath = urlObject.pathname;
   urlPath += urlPath?.endsWith('/') ? '' : '/';
-  const urlParams = urlObject.search;
+  /**
+   * We need an empty string as a default value because in case
+   * of undefined strings, Typescript will go ahead and append the
```
tasos-ale, line 27: Blame JavaScript not TypeScript for this :smiley:

components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
```diff
@@ -28,17 +36,29 @@ export class MainPageComponent implements OnInit, OnDestroy {
   public get buildIdWithLabel() {
     return this.computeBuildValue(this.buildLabel, this.buildId);
   }
+  public menuLinks: menuLink[];
+  public externalLinks: any[];
+  public quickLinks: quickLink[];
```
tasos-ale, line 41: Why we don't render `externalLinks` and `quickLinks` anywhere? If we will have them in a following PR, fine by me.

components/centraldashboard-angular/frontend/src/app/types/dashboard-links.ts
```diff
@@ -0,0 +1,25 @@
...
+  text: string;
+  desc: string;
+  link: string;
+}
+
+export interface documentationItem {
+  text: string;
+  desc: string;
+  link: string;
+}
```
tasos-ale, line 25: Maybe we can use the same type here. I don't see a need for two different interfaces.

tasos-ale (changes requested):
Thanks for the PR @orfeas-k. Please take a look at my comments.

components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
orfeas-k: ACK I agree that this is not something we should tackle in the Dashboard's logic and having `if`s for different WAs isn't a good practice.
cc @kimwnasptd for awareness

components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
orfeas-k: Sure, I removed the prefix from browser's URL and changed that to `startsWith()`.

components/centraldashboard-angular/frontend/src/app/pages/main-page/main-page.component.ts
orfeas-k: In the old dashboard, `quickLinks` and `documentationItems` are part of the `Home` page and this page will be a separate effort/PR. @kimwnasptd What about `externalLinks`?

tasos-ale: @orfeas-k There is no need to wait. If we want the `externalLinks` we can add them in the next PR.

PR #6959, opened 2023-02-15
centraldashboard: Fix failing test 

orfeas-k:
Fix `Activities list` test that was failing because `new Date()` added whitespace that was not equal to the space character during comparison. 

Link to related PR https://github.com/kubeflow/kubeflow/pull/6945#issuecomment-1431257676 cc @kimwnasptd


kimwnasptd:
Thanks for the quick PR! The build tests are passing, which are only testing amd64 arch. Let's see if the tests after merging the PR will work, and the image will be built for ppc64le as well

/lgtm
/approve

PR #6923, opened 2023-01-31
Adding support for linux/ppc64le in CI for centraldashboard multi-arc…

amitmukati-2604:
…h docker images

PR #6960, opened 2023-02-15
centraldashboard(make): Use correct dir

kimwnasptd:
Previously the app would not use the correct directory when running the docker buildx command, which resulted in the image to fail to get built.

I also realized that we are not testing the `docker buildx` flow in each PR, this is why this was never caught in https://github.com/kubeflow/kubeflow/pull/6923. We'll need to extend the current actions to use the same build process to ensure we catch this in a timely manner. 

I'll ping on the corresponding issue
/cc @elikatsis

### DIFF
--- a/components/centraldashboard/Makefile
+++ b/components/centraldashboard/Makefile
@@ -1,7 +1,7 @@
 IMG ?= centraldashboard
 TAG ?= $(shell git describe --tags --always --dirty)
 COMMIT = $(shell git rev-parse HEAD)
-DOCKERFILE ?= centraldashboard/Dockerfile
+DOCKERFILE ?= Dockerfile
 ARCH ?= linux/amd64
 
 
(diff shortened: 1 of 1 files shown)


kimwnasptd:
Looks like even after merging this PR the image can't be built. The reason in this case though has to do with ppc64le. I'm seeing these logs when building for the ppc64le:
https://github.com/kubeflow/kubeflow/actions/runs/4184427448/jobs/7250059227
```
#23 [linux/ppc64le build 5/5] RUN BUILDARCH="$(dpkg --print-architecture)" &&  npm rebuild &&     if [ "$BUILDARCH" = "arm64" ]  ||      [ "$BUILDARCH" = "armhf" ]; then     export CFLAGS=-Wno-error &&     export CXXFLAGS=-Wno-error;      fi &&     npm install &&     npm test &&     npm run build &&     npm prune --production
...
#23 320.5 15 02 2023 13:47:25.388:ERROR [launcher]: Cannot start ChromeHeadless
#23 320.5 	find: '/root/.config/chromium/Crash Reports/pending/': No such file or directory
#23 320.5 [0215/134718.429805:ERROR:bus.cc(399)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
#23 320.5 [0215/134718.618224:ERROR:bus.cc(399)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
#23 320.5 [0215/134718.618972:ERROR:bus.cc(399)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory
```

The above doesn't get triggered for the linux/amd64 arch.

My best guess is that this happens because there's an issue with running the Chromium process inside of the ppc64le container. The reason chromium needs to run in the first place is because in the CentralDashboard we run some frontend tests as part of the docker build process.

IIRC this decision was from the very early days of Kubeflow, where we had very limited CI capabilities. I'd propose to actually create a separate workflow for running the frontend tests and removing this logic altogether cc @tasos-ale @orfeas-k. I'll take a look ASAP at this, since we need to release a CentralDashboard image for the next RC cc @DomFleischmann 

I'll also do a cc @lehrig for visibility, since this got triggered for ppc64le

PR #6961, opened 2023-02-15
[WIP] centraldashboard: Don't run tests in docker build

kimwnasptd:
Run the frontend tests outside of the Dockerfile and inside a dedicated GH Action. We should separate the testing process from the building one.

The dashboard's Dockerfile was running the tests because in the early days, when Kubeflow didn't have a CI, this was the only reproducible way to ensure the tests had passed when creating an image.

This is a follow up from https://github.com/kubeflow/kubeflow/pull/6960#issuecomment-1431434933

### DIFF
--- a/.github/workflows/centraldb_frontend_tests.yaml
+++ b/.github/workflows/centraldb_frontend_tests.yaml
@@ -1,0 +1,32 @@
+name: CentralDashboard Frontend Tests
+on:
+  pull_request:
+    paths:
+      - components/centraldashboard/**
+    branches:
+      - master
+      - v*-branch
+
+jobs:
+  frontend-tests:
+    runs-on: ubuntu-latest
+    name: Unit tests
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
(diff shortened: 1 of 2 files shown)


kimwnasptd:
@tasos-ale @orfeas-k this should be ready for review. I also took the liberty and updated the GH Action to build using the same make rule we use when the PR is merged.

Some findings:
1. We'll need to use `--load` when building with `docker buildx` to load the image manifests. I saw the following error without it and the image manifests would not be pushed to the KinD cluster
   ```
   WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
   ```
2. Using `--load` with a list value for `--platform`, like `linux/ppc64le,linux/amd64`, is not supported https://github.com/docker/buildx/issues/59

Because of the above I actually ended up running the build command once for each platform. Not the best solution, but I couldn't think of something better...

/cc @tasos-ale @orfeas-k @elikatsis

orfeas-k:
Took a look and the changes look good!
/lgtm

PR #6962, opened 2023-02-15
vwa(front): Prevent PVCs from being deleted when there is a corresponding notebook

kimwnasptd:
Cherry-pick bf9da0a15


kimwnasptd:
cc @DomFleischmann

PR #6898, opened 2023-01-18
Introduce workflows for building notebook-server images upon changes

apo-ger:
This PR resolves part of: https://github.com/kubeflow/kubeflow/issues/6766

Introduce GitHub Action workflows for testing that notebook server images can be built for each PR.

PR #6975, opened 2023-02-16
release: Images for the v1.7.0-rc.1 tag

apo-ger:
Update the manifests to use the v1.7.0-rc.1 tag

/cc @kimwnasptd


kimwnasptd:
Thanks @apo-ger! If all tests also pass once this is merged we can use this commit to cut the `v1.7.0-rc.1` tag

cc @DomFleischmann 

/lgtm
/approve

kimwnasptd:
Looks like one action for the Codeserver image has failed. But, we can't reproduce this locally and we also see that the image is being built as expected in another PR https://github.com/kubeflow/kubeflow/pull/6898.

Also there are no changes between RC0 and RC1, so I'll just push the image manually to not block the release.

@DomFleischmann can you cut the `v1.7.0-rc.1` tag from this commit?

PR #6983, opened 2023-02-22
read pod defaults should only return valid pod defaults

johnhoman:
This PR is to address the issue of creating `PodDefaults` without `matchLabels`, or too many `matchLabals`. Creating a `PodDefault` like the following can be super useful for applying defaults to all notebooks in a namespace.

```yaml
---
kind: PodDefault
metadata:
  name: foo-bar
spec:
  selector:
    matchExpressions:
    - key: notebook-name
      operator: Exists
  env:
  - name: BAR
    value: FOO
```
That PodDefault will unfortunately break the UI here https://github.com/kubeflow/kubeflow/blob/c946c624cd8974ef53cac04582e4714b466d546e/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py#L37.
image
```sh
2023-02-22 05:22:07,568 | kubeflow.kubeflow.crud_backend.errors.handlers | ERROR | 'matchLabels'
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/src/apps/common/routes/get.py", line 32, in get_poddefaults
    label = list(pd["spec"]["selector"]["matchLabels"].keys())[0]
KeyError: 'matchLabels'

```
While is doesn't prevent users from creating new notebooks, it does prevent them from selecting existing configurations.

 Another issue is that some configurations can be selected, but won't be applied, such as the following.
```yaml
---
kind: PodDefault
metadata:
  name: foo-bar-0
spec:
  selector:
    matchLabels:
      kubeflow.org/foo: "true"
      kubeflow.org/bar: "true"
  env:
  - name: FOO
    value: BAR
---
kind: PodDefault
metadata:
  name: foo-bar-1
spec:
  selector:
    matchLabels:
      kubeflow.org/bar: "foo"
  env:
  - name: BAR
    value: FOO
```

image

```sh
➜  manifests git:(v1.6-branch) ✗ kubectl get pod/jupyter-scipy-0 -n kubeflow-user-example-com -o yaml | yq  '.spec.containers[0].env'
[
  {
    "name": "NB_PREFIX",
    "value": "/notebook/kubeflow-user-example-com/jupyter-scipy"
  }
]
```

This can be misleading to users who may have selected configurations that weren't applied to their notebook. These changes resolve both of these issues by only returning `PodDefaults` that can be applied when selected. I'm also not 100% sure this is the right solution. I think it's the most flexible, but if the intention is for `PodDefaults` to only match on a single label, then the right solution might be a validating webhook for `PodDefaults`.

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py
@@ -27,26 +27,41 @@ def get_pvcs(namespace):
     return api.success_response("pvcs", data)
 
 
-@bp.route("/api/namespaces/<namespace>/poddefaults")
-def get_poddefaults(namespace):
-    pod_defaults = api.list_poddefaults(namespace)
-
+def read_pod_defaults(namespace, client=api):
+    pod_defaults = client.list_poddefaults(namespace)
     # Return a list of pod defaults adding custom fields (label, desc) for forms
     contents = []
     for pd in pod_defaults["items"]:
-        label = list(pd["spec"]["selector"]["matchLabels"].keys())[0]
-        if "desc" in pd["spec"]:
-            desc = pd["spec"]["desc"]
-        else:
-            desc = pd["metadata"]["name"]
+        try:
+            # The only PodDefaults that should be selected are ones that contain
+            # a single matchLabel with a value of "true".
+            labels = pd["spec"]["selector"]["matchLabels"]
+        except KeyError:
+            continue
 
-        pd["label"] = label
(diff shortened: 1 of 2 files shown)


juliusvonkohout:

andreyvelich:
Please re-open PR there.
/close

PR #6985, opened 2023-02-22
jwa: Remove all the Rok references

elenzio9:
This PR removes all the Rok references from JWA's code.


tasos-ale (changes requested):
@elenzio9 I don't have any big change to suggest since we are only removing code. Shouldn't we also remove the `browse-in-rok-*.svg` images?

elenzio9:
> Shouldn't we also remove the `browse-in-rok-*.svg` images?

Right! I just removed them.

tasos-ale (approved):

tasos-ale:
@kimwnasptd I believe we can merge this PR.

PR #6995, opened 2023-02-23
centraldashboard: Allow all-namespaces option for mwa

elenzio9:
In this PR, we enable all-namespace support in MWA by displaying the `All namespaces` option in the dashboard.

Related PR: https://github.com/kserve/models-web-app/pull/68


orfeas-k:
Built the image and tested on my cluster. It works as expected, thank you @elenzio9! This should be merged after https://github.com/kserve/models-web-app/pull/68 in order for the All namespaces feature to be functional.
/lgtm
cc @kimwnasptd

PR #7019, opened 2023-03-01
web-apps(back): Fetch resources events

elenzio9:
This PR:

1.  Introduces a `list_events(namespace, field_selector)` function in `events.py` for listing events of each resource.
2.  Utilizes `list_events` function for listing notebook events in `notebook.py`.
3. Utilizes `list_events` function for listing pvc events in `pvc.py`.

### DIFF
--- a/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/api/events.py
+++ b/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/api/events.py
@@ -1,0 +1,12 @@
+from .. import authz
+from . import v1_core
+
+
+def list_events(namespace, field_selector):
+    authz.ensure_authorized(
+        "list", "", "v1", "events", namespace
+    )
+
+    return v1_core.list_namespaced_event(
+        namespace=namespace, field_selector=field_selector
+    )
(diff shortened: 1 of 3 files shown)


orfeas-k:
Tested and it works as expected, thank you for this @elenzio9 ! 
/lgtm

components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/api/isvc.py
```diff
@@ -0,0 +1,8 @@
+from . import utils, events
+
+
+def list_inference_service_events(name, namespace):
```
kimwnasptd, line 4: I'd propose to avoid introducing a handler for the InferenceServices in the common backend code.

This would mean we'd need to add more handlers for list/delete/get/logs as well, which as a result will make the MWA depend even more in this common code. I'd propose to instead keep on using the `custom_resource.py` file's handlers for the ISVC logic in the MWA

elenzio9: ACK! I've just removed that part and will add the changes to the follow-up PR, which will introduce the `EVENTS` tab, as they'll relate to the MWA backend code.

kimwnasptd:
Thanks!

/lgtm
/approve

PR #7023, opened 2023-03-02
Switch from dockershim to containerd

SergeyKanzhelev:
Dockershim clusters will run out of support at some point soon. This is the obvious place that needs to be updated I found. Please advice if there are other places that need to be fixed.


kimwnasptd:
Hey @SergeyKanzhelev, the `testing` directory is not being used anymore. I had also tried to remove/clean it but my PR is not merged https://github.com/kubeflow/kubeflow/pull/6535 merged yet. 

I think Google don't want to touch anything on the root level of the project, so we'll stick with the unmaintained files until we go to CNCF I guess.

PR #7024, opened 2023-03-03
fix: use "required" in tensorboard node affinity, rather than "preferred"

Jason-CKY:
When `RWO_PVC_SCHEDULING` is set to `true`, a `nodeAffinity` is generated to deploy the tensorboard image to the same node as the running pod using the `PVC`. This is done with `preferredDuringSchedulingIgnoredDuringExecution` with a weight of `100`. However, [this does not guarantee that the kube-scheduler will schedule the tensorboard pod to be on the same node, just with an added weight when considering which node to schedule to.](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity-weight) This could result in times where the pod will still be scheduled on a different node, causing volume mount errors with `ReadWriteOnce` PVCs. 

Thus, this PR is to change the `nodeAffinity` to `RequiredDuringSchedulingIgnoredDuringExecution` instead, to force the kube-scheduler to schedule the tensorboard image into the correct node.

Since the user would have only `ReadWriteOnce` persistent volumes if they set  `RWO_PVC_SCHEDULING` to `true`, it would not affect the user if the scheduling is changed to `RequiredDuringSchedulingIgnoredDuringExecution`, since the tensorboard image will fail to start at all if the pod is schedule on any other pod.

### DIFF
--- a/components/tensorboard-controller/controllers/tensorboard_controller.go
+++ b/components/tensorboard-controller/controllers/tensorboard_controller.go
@@ -404,8 +404,8 @@ func extractPVCSubPath(path string) string {
 	}
 }
 
-//Searches a corev1.PodList for running pods and returns
-//a running corev1.Pod (if exists)
+// Searches a corev1.PodList for running pods and returns
+// a running corev1.Pod (if exists)
 func findRunningPod(pods *corev1.PodList) corev1.Pod {
 	for _, pod := range pods.Items {
 		if pod.Status.Phase == "Running" {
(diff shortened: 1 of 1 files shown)


Jason-CKY:
I have signed the CLA

juliusvonkohout:

thesuperzapper:
I am not sure if this update can be done in-place, what will happen to existing TensorBoard instances?

How will the controller update these, will it cause interruption, or just break?

Issue #6741, opened 2022-11-15
Jupyter UI form default values not reflecting changes from jupyter-web-app-config configMap


alexchrisramos:
/kind bug

**What steps did you take and what happened:**
We are looking into changing the default values for Jupyter UI. So that when users creates a new notebook server, the new defaults would take effect.
So we have modified the configMap jupyter-web-app-config in kubeflow namespace.
In the following example, we have changed the default values for the workspaceVolume.
from 10Gi to 20Gi and from ReadWriteOnce to ReadWriteMany
```
spec:
  resources:
    requests:
      storage: 20Gi
  accessModes:
  - ReadWriteMany
```
Saved and restarted the jupyter-web-app-deployment.
In the New Notebook web form, the "New volume" header is updated, but the form is still showing the original defaults.
<img width="825" alt="image" src="https://user-images.githubusercontent.com/36431265/201811048-0e5ed4b3-ca52-4564-bbfc-81ebdb463589.png">
Proceeding with creating the notebook server, without touching or changing anything in the form, creates the pods with default values from the configmap. Although, regular users will be confused by the selected options in the UI.

It looks like the form selection is hardcoded into the ui.
https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/frontend/src/app/shared/utils/volumes/forms.ts#L79-L83

**What did you expect to happen:**
Changes in the configMap as default values would reflect in the UI.

**Environment:**

- Kubeflow version: build version v1.5.0
- Kubernetes platform: kubeadm/clusterAPI
- OS: RHEL 8.4

kimwnasptd:
Looks like a small bug in how the form handles these values. Thank you for raising this @alexchrisramos!

/assign @kimwnasptd

andreyvelich:
This was fixed in #7029

PR #7029, opened 2023-03-06
jwa(front): Fix the workspace volume form's inputs

elenzio9:
The PR ensures that the intermediate FormControls for the name, size and access modes will be getting updated correctly when the frontend gets the ConfigMap data.

Related issue: https://github.com/kubeflow/kubeflow/issues/6741


components/crud-web-apps/jupyter/frontend/cypress/e2e/form-page.cy.ts
```diff
@@ -57,4 +57,61 @@ describe('New notebook form', () => {
...
+    cy.wait(['@mockConfigRequest']);
+
+    cy.get('[data-cy="workspace volume"]').click();
+
+    cy.get('[data-cy="workspace volume"]')
+      .find('[data-cy="volume name input"]')
+      .should($nameInput => {
+        const nameValue = $nameInput.val();
+        //
+        expect(nameValue).equal('-workspace'); // '{notebook-name}-workspace' is the name value of the config fixture
```
orfeas-k, line 87: The intention here was probably to move the comment from line 87 to where the `//` are now? Nit but I think it 'd be better this way in order to avoid having a really long line.

components/crud-web-apps/jupyter/frontend/cypress/e2e/form-page.cy.ts
```diff
@@ -57,4 +57,61 @@ describe('New notebook form', () => {
...
       });
   });
+
+  it('should update panel header name according to the name input field', () => {
+    cy.visit('/new');
+
+    cy.get('[data-cy="workspace volume"]').click();
+
+    cy.get('[data-cy="workspace volume"]')
+      .find('[data-cy="volume name input"]')
```
orfeas-k, line 67: I think that the `cy.get().find()` chain is redundant here and you could convert this to `cy.get('[data-cy="volume name input"]')`. 

The same goes for all `cy.get().find()` in the tests introduced. There are cases where we might need to chain `cy.get()` with `.find()` but as far as I 'm concerned, this is not the case here. WDYT @elenzio9 ?

orfeas-k (commented):
Thank you for fixing this bug @elenzio9! The feature seems to work as expected. Here are some minor comments on the tests.

components/crud-web-apps/jupyter/frontend/cypress/e2e/form-page.cy.ts
elenzio9: @orfeas-k Thanks for the comments! Just pushed a fixup commit. Please take a look!

orfeas-k:
It looks good, I 'll go ahead and merge! @elenzio9 
/lgtm
/approve

PR #7030, opened 2023-03-06
cdb(front): Add namespace selector

orfeas-k:
This PR is part of the [Rewrite the Central Dashboard in Angular effort](https://github.com/kubeflow/kubeflow/issues/6000) and comes right after [this one](https://github.com/kubeflow/kubeflow/pull/6953). 

### Namespace Service
Responsible for handling all actions related to the namespace in CDB (keeping information about the current namespace, retrieving most recent namespace used, updating components about namespace change etc) will be the new **NamespaceService**. Any component that needs to interact with the app's namespace will use this service.

#### Initializing the namespace

Once we receive the namespaces available, we need to initialize the selected namespace. We follow the below steps in order to decide upon the namespace. 

 - Check if namespace is set through the `query parameters` of current URL. 
 - Check if the user's previous namespace is present in the browser's `local storage`
 - Check in the namespaces list for a namespace with `Owner` role (and keep the first one)
 - Check in the namespaces list for the first `valid` namespace
 
In every of these steps, if we find a namespace, **we return it only if this namespace option is also valid** (considering also the `All namespaces` option).

After we finalize on the namespace, we also:
 - Emit an observable with it to the `currentNamespace` subscribers
 - Update the namespace for this specific user in the browser's `localStorage`
 - Update the browser's URL's query parameters with the initialized namespace 

#### Select namespace
The last three steps also take place when a user changes the selected namespace via the Namespace selector dropdown.

### Single user mode

In the early days, there was the notion of running Kubeflow for a single user. In that case there was no header in the requests with the user info. Extending what we decided [here](https://github.com/kubeflow/kubeflow/pull/6918) for `isolationMode` variable (at that point we did it in regard to Kubeflow's Version), **we decide to drop the `single-user` mode from CDB.** Thus, we will also drop requests at `/api/workgroup/exists` endpoint, from where we received until now user related information.

Based though on that information, we also [enabled/disabled the **Registration** page](https://github.com/kubeflow/kubeflow/blob/74f020e0d9c3f58712a3b466f9d1bb86c4607beb/components/centraldashboard/public/components/main-page.js#L469-L472) and modified the content of the **Manage contributors** page. 

#### Manage contributors page

**We want admins to manage users in a GitOps way**, and not via that UI. 

#### Registration page

We also had the Registration page where the user can create a user Profile (which creates a namespace).

For the moment, **we move forward without implementing these two pages** in the new Central Dashboard.

### Post `PARENT_CONNECTED_EVENT` to iframed WA
We decided to not post a message of type `PARENT_CONNECTED_EVENT` as [we did in the old dashboard](https://github.com/kubeflow/kubeflow/blob/88e1b81c417438bd5dc569b45e5798238a628307/components/centraldashboard/public/components/iframe-container.js#L120-L123) (we will only post a message with the selected namespace). We don't see any use of it currently and thus, we will not include it in CDB.


components/centraldashboard-angular/frontend/src/app/services/namespace.service.ts
```diff
@@ -0,0 +1,199 @@
...
+          this.allNamespacesAllowedSubject.next(this.allNamespacesAllowed);
+        }
+
+        this.setCurrentNamespace(this.namespaces, this.user);
+      }
+    });
+  }
+
+  private setCurrentNamespace(namespaces: Namespace[], user: string) {
+    if (user) {
```
tasos-ale, line 71: Is it possible to not have the `user` when we use the `setCurrentNamespace()`?

components/centraldashboard-angular/frontend/src/app/services/namespace.service.ts
```diff
@@ -0,0 +1,199 @@
...
+      this.namespacesSubject.next(namespaces);
+
+      this.allNamespacesAllowed = this.calcAllNamespacesOption(this.router.url);
+      this.allNamespacesAllowedSubject.next(this.allNamespacesAllowed);
+
+      this.setCurrentNamespace(namespaces, this.user);
+    });
+
+    this.router.events.subscribe(event => {
+      if (event instanceof NavigationEnd) {
```
tasos-ale, line 58: We can use a `filter()` and remove the `if statement` since we only care for `NavigationEnd`

components/centraldashboard-angular/frontend/src/app/pages/iframe-wrapper/iframe-wrapper.component.ts
```diff
@@ -101,7 +112,18 @@ export class IframeWrapperComponent implements AfterViewInit, OnDestroy {
...
         const path = iframeWindow?.location.pathname;
-        const queryParams = this.getQueryParams(iframeWindow?.location.search);
+        let queryParams = getQueryParams(iframeWindow?.location.search);
+
+        /**
+         * Append current namespace to query parameters before using router.navigate
+         * since iframe's internal URL doesn't hold any information regarding the
+         * namespace and we want to prevent it from discarding this infromation from
+         * the Browser's URL.
+         */
```
tasos-ale, line 122: FYI https://google.github.io/styleguide/jsguide.html#formatting-block-comment-style. There is no need to use `/**`, you can change it to `/*`.

components/centraldashboard-angular/frontend/src/app/services/environment.service.ts
```diff
@@ -0,0 +1,35 @@
...
+      if (res.user) {
+        this.user.next(res.user);
+      }
+      if (res.platform) {
+        this.platform.next(res.platform);
+      }
+      if (res.namespaces) {
+        this.namespaces.next(res.namespaces);
+      }
+    });
```
tasos-ale, line 29: Do we really need 3 different subjects for these values?

components/centraldashboard-angular/frontend/src/app/services/namespace.service.ts
```diff
@@ -0,0 +1,199 @@
...
+    return undefined;
+  }
+
+  private updateStoragedNamespace(namespace: Namespace | undefined) {
+    // Save the user's choice so we are able to restore it,
+    // when re-loading the page without a queryParam
+    const localStorageKey =
+      'selectedNamespace/' + ((this.user && '.' + this.user) || '');
+    this.localStorage.set(localStorageKey, namespace?.namespace);
+  }
```
tasos-ale, line 175: In general, try to pass everything as a parameter. It will make testing easier.

tasos-ale (changes requested):
@orfeas-k Thanks for your PR. Please take a look at my comments.

components/centraldashboard-angular/frontend/src/app/services/environment.service.ts
orfeas-k: We don't really need 3 different subjects since we fetch all of the above simultaneously from the same request. I followed the above approach because I found it to be cleaner since some components may need only one of these properties. However, I 'm open to reconsidering it, since this results in some other components having to subscribe more than once.

tasos-ale: I don't have a strong opinion either. So, let's not change it.

tasos-ale (commented):
A few more comments @orfeas-k .

orfeas-k:
@tasos-ale Thank you for your comments. I pushed commits that address them. This includes as well a rewrite of the namespace service core functionality, as well as tests for different parts of this PR. Let me know what you think!

components/centraldashboard-angular/frontend/src/app/services/namespace.service.ts
```diff
@@ -0,0 +1,201 @@
...
+})
+export class CDBNamespaceService {
+  private _namespacesSubject = new ReplaySubject<Namespace[]>(1);
+  private _currentNamespaceSubject = new ReplaySubject<Namespace>(1);
+
+  namespacesSubject = this._namespacesSubject.asObservable();
+  currentNamespaceSubject = this._currentNamespaceSubject.asObservable();
+
+  private user: string;
+  private currentNamespace: Namespace | undefined;
```
tasos-ale, line 21: Maybe we can do this:
```suggestion
  namespaces = this._namespacesSubject.asObservable();
  currentNamespace = this._currentNamespaceSubject.asObservable();

  private _user: string;
  private _currentNamespace: Namespace | undefined;
```
wdyt @orfeas-k ?

tasos-ale (changes requested):
@orfeas-k a few minor comments.

components/centraldashboard-angular/frontend/src/app/services/namespace.service.ts
orfeas-k: Sure, but I 'll rename them from `_variable` to `prvVariable` because that's what we do the in the rest of KF WAs as well.

tasos-ale (approved):

PR #7035, opened 2023-03-10
adding privacy information to dashboard config to make it easily customizable

sebastianohl:
Problem:
The git version of kubeflow dashboard has hard coded privacy policy links that point to the google privacy statement. While this might be acceptable for private installations, it can cause problems in public or semi public installations (in my case a university internal service). In should be customizable to point to the organization specific privacy statement and maybe additional documents, if needed.

Proposal:
This PR adds the legal links to the dashboard config. It is handled as the other menu items or parts of the main page. Therefore, everyone can simply change the links if needed in his configmap. This PR changes all occurances of hardcoded privacy links that I have found (main-page and registration page). The standard values in the config still points to the old google links.


sebastianohl:
PR will be recreated with correct email adresses to comply witht eh CLA check

PR #7059, opened 2023-03-24
Add ability to use kubernetes generic ephemeral volumes in jupyter

### LINKED ISSUE #3840: Delete Jupyter Notebook volume along with notebook deletion
/kind feature

When user will create a Jupyter Notebook and create new persistent volume, PV doesn't have ownerReference of Jupyter Notebook, the time we delete notebook, PV is in isolated status. 

Confirmed with community and it was by design and user may reuse storage some time later.
I think GPU training notebook will have more case like this and user like to shutdown GPU notebook frequently to reduce cost. 

But I notice another problem users feel it's hard to maintain isolated PV and users/operations are not sure if anyone will use it later and they can not delete notebook. 

We need to way to delete associated PV based on user's choice. 

1. When user click delete button, populate a window with a check box to confirm with user if they're like to delete volume as well. 

2. By default, it could be unchecked.

and-1:
Add checkbox to new volume form to create ephemeral workspace and/or data volume. Used generic ephemeral volumes as implementation https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/#generic-ephemeral-volumes. 
Can resolve kubeflow/notebooks#71 in some cases

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/volumes.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/volumes.py
@@ -4,6 +4,7 @@ The new API Volume will work with objects of the following format:
 
 volume:
   mount: "mount path"
+  ephemeral: "true|false"
   newPvc?:
     metadata: ...
     spec: ...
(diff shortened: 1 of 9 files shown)


juliusvonkohout:

andreyvelich:
Please re-open PR there.
/close

Issue #7040, opened 2023-03-14
Tensorboard Controller is not handling generateVirtualService errors


japan4415:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

I realize this bug when I apply Tensorboard Controller without ISTIO_GATEWAY config.

**What did you expect to happen:**
If you apply this controller without ISTIO_GATEWAY config, virtualService will be nil and crush with nil pointer at [L114](https://github.com/kubeflow/kubeflow/blob/bbc55dd4a981d5c8e7de038715b11b07fa6883c2/components/tensorboard-controller/controllers/tensorboard_controller.go#L114).
So we should handle errors at [L113](https://github.com/kubeflow/kubeflow/blob/bbc55dd4a981d5c8e7de038715b11b07fa6883c2/components/tensorboard-controller/controllers/tensorboard_controller.go#L113).

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]
Can I make PR for fix this?

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard):
- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`)
- Kubernetes version: (use `kubectl version`):
- OS (e.g. from `/etc/os-release`):

PR #7067, opened 2023-03-25
fix: add error handling to tensorboard controller

japan4415:
Close https://github.com/kubeflow/kubeflow/issues/7040
If you apply this controller without ISTIO_GATEWAY config, virtualService will be nil and crush with nil pointer at [L114](https://github.com/kubeflow/kubeflow/blob/bbc55dd4a981d5c8e7de038715b11b07fa6883c2/components/tensorboard-controller/controllers/tensorboard_controller.go#L114).
So we should handle errors at [L113](https://github.com/kubeflow/kubeflow/blob/bbc55dd4a981d5c8e7de038715b11b07fa6883c2/components/tensorboard-controller/controllers/tensorboard_controller.go#L113).

### DIFF
--- a/components/tensorboard-controller/controllers/tensorboard_controller.go
+++ b/components/tensorboard-controller/controllers/tensorboard_controller.go
@@ -111,6 +111,9 @@ func (r *TensorboardReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	// Reconcile istio virtual service.
 	virtualService, err := generateVirtualService(instance)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


japan4415:
miss...

juliusvonkohout:

japan4415:
This pr is only add error handling. Should I join meeting?

juliusvonkohout:
We do not have the bandwith to cover everything in written form. I think it always makes sense to join and then your PR can get attention. This is also important to understand our workflow and introduce yourself to the community.

PR #6181, opened 2021-10-11
Tell user to select a namespace to access namespaced menu item

toshi-k:
Fix
#6118

Although namespaced menus items will be available for KF1.4 (#5871), there is room for improvement.
@munagekar pointed out the message that is shown for the user when he tries to access the namespaced item without  namespace selection.

This PR makes it possible to show the user-friendly message "Please select a namespace from the top bar" for such users.

@munagekar I appreciate if you help discussions for this PR.

/cc @kimwnasptd @jbottum @munagekar

Issue #6118, opened 2021-09-02
Issue with Central Dashboard Custom Links


munagekar:
/kind bug

Weird Behaviour for custom internal links in kubeflow central dashboard when no namespace is specified.

**What steps did you take and what happened:**

Docker Image for Central dashboard: public.ecr.aws/j1r0q0g6/notebooks/central-dashboard:v1.4-rc.0

Links Configuration
```yaml
        {
          "type": "item",
          "link": "/{ns}/mlflow/",
          "text": "MLFlow",
          "icon": "editor:show-chart"
        }
```

Clicked on the MLFlow link in the sidebar with no namespace selected. This opened up another iframe for the centraldashboard within the centraldashboard. 

<img width="1790" alt="Screen Shot 2021-09-02 at 13 57 15" src="https://user-images.githubusercontent.com/10258799/131810829-1dea973b-e385-41fe-bee3-df8621e55a20.png">

**What did you expect to happen:**

CentralDashboard to raise an error, prompt user to select a namespace. 

Note: That that custom link works correctly when the namespace is selected.

**Anything else you would like to add:**

- Feature Issue: https://github.com/kubeflow/kubeflow/issues/5863
- Kubeflow version: 1.3
- CentralDashboard Image: public.ecr.aws/j1r0q0g6/notebooks/central-dashboard:v1.4-rc.0 This contains the fix from #5871.

cc: @toshi-k

toshi-k:
@munagekar I've never considered this case. Thank you for your report.

jbottum:
@munagekar did you get this resolved?  did you modify the dashboard configuration to include MLFlow ?

munagekar:
> did you get this resolved?
@jbottum This is not yet resolved. This is an edge-case for the [newly created feature](https://github.com/kubeflow/kubeflow/issues/5863).

>  did you modify the dashboard configuration to include MLFlow ?
Yes I modified dashboard configuration to link to custom MLFlow integration with Kubeflow.

toshi-k:
@munagekar I'm working on this issue and let me hear your impression.

I propose to show the message as below when the user accesses the namespaced menu item without namespace selection.
![namespace_menu_item](https://user-images.githubusercontent.com/10993535/135598077-61608937-718f-444a-9002-e346d0b35b9e.png)

I also ensure that "Sorry, [path] is not a valid page" will be shown for the user who tries to access unavailable services.
[...]

PR #7071, opened 2023-03-28
cdb(front): Introduce 404 not found page 

orfeas-k:
This PR is part of the [Rewrite the Central Dashboard in Angular effort](https://github.com/kubeflow/kubeflow/issues/6000) and comes right after [this one](https://github.com/kubeflow/kubeflow/pull/7030). 

### Context 

Up until now, in CDB's frontend, we have configured Angular's router to expose a Home page under `'/'` path and used the Iframe component for all other paths (`'**'`). However, this results in the iframe trying to navigate to paths that do not correspond to any of the available WAs. Thus, we need to configure the router to navigate to a `404 not found` page whenever the user navigates to an invalid path.

A high overview of the links a user can navigate to is:
 - Home page with path `'/'`
 - Links prefixed with `'/_'`
   - Valid paths
   - Invalid paths
 - Other (invalid) links without a prefix

For each of the above (excluding the sub-bullets), we will have an entry in CDB's router. For any link that doesn't match the first two routes, the WA will render the 404 page.

#### Problem

Navigating to a link prefixed with `'/_'` doesn't guarantee though that this path will be valid/served. Thus, we need to be able to show a `404 not found` page for invalid paths inside the iframe too.  

When we pass a URL to the iframe, we always remove the prefix. For URLs that we don't proxy like we do with the WAs' paths (e.g. requests to `/jupyter`), the iframe will render a second CDB WA internally, that will in turn navigate to the URL without the prefix. This will create an Inception/mirror effect with the internal CDB rendering a 404 page (since it's an invalid path). 

image
(`/tensorboards` was not served in the above example)

Sidenote: In this [issue](https://github.com/kubeflow/kubeflow/issues/6118#issue-986381327), the inception resulted in 3 dashboards (instead of 2 that is expected). This inception was fixed in this [PR](https://github.com/kubeflow/kubeflow/pull/6181 
). However, we could not understand why didn't the first inception result in a 404 page.

#### Solution

In order to overcome this, we looked into how this functionality was implemented in the previous dashboard. We found out that it has an observer on the page's route and whenever there's a new route it goes through the following steps:
 1. If the new route data matches any of these cases `logout`, `activity`, `iframe-prefix`, `manage-users` and `home` , then show the matching page.
 2. If none of the above is true, then go to default case. This means that no matching page has been found and we follow these steps
    - If the dashboard WA is in an iframe (which means that CDB is inside another CDB), then hide the sidebar and the header.
    - If route's path contains `'{ns}'` , then show the Namespace needed page, else show the 404 not found page.

The logic for the default case as is today was introduced by these upstream PRs https://github.com/kubeflow/kubeflow/pull/2778 https://github.com/kubeflow/kubeflow/pull/6181 

We decided to adopt the old dashboard's approach and **hide the sidebar and header when CDB is in an iframe**. 

The benefit of the above approach is that it doesn't consider the menu links at all and at the same time catches 404 cases for URLs inside the iframe. It also allows access to all paths that may be served but are not included in the menu links. This solves all of the problems we faced when we were considering the option of comparing the current URL with the sidebar's menu links in order to validate it. Using the menu links would bring us to the same issue mentioned [here](https://github.com/kubeflow/kubeflow/pull/6953) and that would result in showing 404 page for URLs from KFP that are actually valid.

### Namespace needed page 

We see that the old dashboard also had a [Namespace needed page](https://github.com/kubeflow/kubeflow/issues/6118) displaying a message "Please select a namespace from the top bar". This was introduced to take care of namespaced menu links in cases that there is no namespace selected (which resulted in the menu link containing a `{ns}` e.g. `/app/{ns}`). The way we initialize the namespace though now (see [here](https://github.com/kubeflow/kubeflow/pull/7030)), there will always be a namespace selected. The only possibility of not having selected a namespace is when there are no namespaces available. Thus, we will keep this page but instead we will show the following message:

```
This page requires a namespace to be selected but no namespaces are currently available
```

Also, in order to expose in the browser the original URL that the user tried to navigate to (which corresponds to the page that actually needs a namespace), after navigating to the Namespace needed page, we used [Location.replaceState](https://angular.io/api/common/Location#replaceState) with the page's URL.


tasos-ale (changes requested):
@orfeas-k thanks for your PR. Please take a look at my comments.

orfeas-k:
When creating the `404 not found` page, I added an `if` as stated here

> If the dashboard WA is in an iframe (which means that CDB is inside another CDB), then hide the sidebar and the header.


This resulted in the CDB hiding its sidebar and header when running tests with Cypress since this runs them inside an iframe. Thus we added the following before each test in order to make the dashboard think that it is not running inside an iframe.
 
```
    cy.on('window:before:load', win => {
      win.parent = win;
   }
```

PR #7078, opened 2023-03-31
Apply new release tags to all controller images

annajung:
Follow up on https://github.com/kubeflow/manifests/pull/2419#issuecomment-1482937192

In the [list of images](https://github.com/kubeflow/manifests/blob/v1.7-branch/docs/kf_images.txt) shipped with the 1.7 release, it was found that `notebook-controller`, `profile-controller`, and `tensorboard-controller` are using both the `latest` and `v1.7.0` tags in the manifests.

Specifically, when running kustomize in the `manager` directory that exists in the manifest repo, it does not apply the tag due to having no dependency on `/config/base/kustomization.yaml` where the tag is currently specified. 

Below are the affected manifest directories and the notebook images
- kustomize build apps/jupyter/notebook-controller/upstream/manager
  - docker.io/kubeflownotebookswg/notebook-controller
- kustomize build apps/profiles/upstream/manager
  - docker.io/kubeflownotebookswg/profile-controller
- kustomize build apps/tensorboard/tensorboard-controller/upstream/manager
  - docker.io/kubeflownotebookswg/tensorboard-controller

This PR fixes kustomize configuration to apply the new tags to all controller images by moving the `images` from `/config/base/kustomization.yaml` to `/config/manager/kustomization.yaml`.

cc @kimwnasptd @difince


annajung:
I noticed that it's noted in the [manifests README](https://github.com/kubeflow/manifests/tree/master/apps/jupyter/notebook-controller/upstream#manifests) that "`crd`, `default`, `manager`, `rbac`, and `samples` directories are to be compatible with kubebuilder-generated structure and that it is not meant for the consumer of the manifests."

Does that mean this change is not applicable since it's the `manager` directory that's building the images without the tag? This may also mean that the script generating the list of images should be modified to exclude the following directories? cc @difince 

Or does it still make sense to apply this change? feedback is appreciated, thanks!

difince (changes requested):
Hi Anna, IMO and having in mind your comments about the `kubebuilder` directories layout - the proposed changes in the PR are not applicable, but instead, as you suggested, `kubebuilder` related directories (`crd`, `default`, `manager`, `rbac`, `samples`) need to be excluded from the [script](https://github.com/kubeflow/manifests/blob/v1.7-branch/hack/extract_manifests_images.sh) generating the list of images.

annajung:
Thanks @difince for the feedback! I can close this out based on your feedback and re-open it when necessary. 
let's make sure the script and list of images are fixed to reflect that.

Issue #6166, opened 2021-10-01
Profile Controller AWS Plugin Ability to Disable IAM Policy Document Generation 


vinayan3:
/kind feature

**Why you need this feature:**
In our cloud environment we use the  IaC, infrastructure as code, tool Terraform to manage IAM roles and the policy documents.  The profiler controller when an Role arn is set does not need to attempt to grant permissions for the role to assume a certain role.

**Describe the solution you'd like:**
The ability to disable updating or removing the IAM policy document. This call needs to have a conditional:
https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/controllers/plugin_iam.go#L39

There is multiple ways to add this. It could be config on the controller it self or something added to the plugins to disable this feature on a per profile basis.

**Anything else you would like to add:**
I'm open to contributing and would appreciate some direction on the preferred style for configuring this flag.

goswamig:
The changes in profiler controller are due to support for plugins in user profile of type AwsIamForServiceAccount. This enables the user profile to associate an AWS role to it.

Take a look at the sample profile https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/config/samples/profile_v1_aws_iam.yaml#L9

I feel this is a feature than bug. Let me know if you still see other way.

mihuross:
Hi, I've find out this issue by tracing the permission error to AWS IAM by profile controller. Our usecase is the same as described in issue. On EKS we use We create roles via terraform and bind them via serviceaccounts, hence they are already preconfigured correctly, so IMHO profile controller should not be updating the trust relationship in this case.

vinayan3:
> Take a look at the sample profile https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/config/samples/profile_v1_aws_iam.yaml#L9
> 
> I feel this is a feature than bug. Let me know if you still see other way.

In the original post I describe a situation where the IAM Roles already have a trust relationship setup because Terraform is used to create the Roles. The Terraform also handles setting up the trust relationship therefore, we do not grant the profile controller the ability to change the trust realtionship of roles. It is not needed. 

I'm proposing adding an environment variable to _disable_ mutating an IAM Role's trust realtionship.

vinayan3:
This was fixed by https://github.com/kubeflow/kubeflow/commit/0f57b056ef9d3d04163a8e2e3f9c0c8f4d4d2e22

PR #7083, opened 2023-04-06
(feat) Add Env Var to Control Editing of IAM Role Trust Relationships

vinayan3:
This PR adds in an environment variable that can be added to the `profile-deployment` which allows the disable the IAM policy generation and editing IAM roles. 

The context on why this is needed is in #6166 

Resovles #6166

### DIFF
--- a/components/profile-controller/controllers/plugin_iam.go
+++ b/components/profile-controller/controllers/plugin_iam.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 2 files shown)


vinayan3:
Actually, this PR might not be needed because `annotateOnly` is there: https://github.com/kubeflow/kubeflow/tree/master/components/profile-controller#plugins

vinayan3:
This is not need give `annotateOnly` is in 1.7. I'm going to close this PR.

PR #7087, opened 2023-04-08
Checksum verification kubectl without saving checksum file and add key --no-cache-dir for pip 

alekseyolg:
Checksum verification kubectl without saving checksum file allows you to reduce the number of commands for the dockerfile.
Add key --no-cache-dir for pip Allows you to reduce the size of images.

### DIFF
--- a/components/crud-web-apps/jupyter/Dockerfile
+++ b/components/crud-web-apps/jupyter/Dockerfile
@@ -45,11 +45,11 @@ FROM python:3.8-slim
 
 WORKDIR /package
 COPY --from=backend-kubeflow-wheel /src .
-RUN pip3 install .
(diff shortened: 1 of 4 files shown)


juliusvonkohout:

We know that the images are also very outdated and are looking for volunteers updating them.

alekseyolg:
@juliusvonkohout Thank you for the offer, but unfortunately I do not speak English to the required degree in order to be able to talk with native speakers, as I do not have the knowledge of English at the level necessary to explain my thoughts.
I could help with the images while communicating with the help of text through a translator.

juliusvonkohout:
Maybe just try to join and see how it goes. We do not expect perfect English.

Issue #6740, opened 2022-11-14
Azure WorkloadIdentity profile plugin


LeonardAukea:
/kind feature

**Why you need this feature:**
Having a profile plugin for [Azure workload Identity](https://github.com/Azure/azure-workload-identity) would make it seamless to work with Azure cloud resources securely. 

> Workload Identity enables Kubernetes applications to access Azure cloud resources securely with [Azure Active Directory](https://azure.microsoft.com/en-us/services/active-directory/) based on annotated [service accounts](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/).

**Describe the solution you'd like:**
Implement a default plugin for workload identity in a similar fashion as done for AWS & GCP 

**Anything else you would like to add:**

Do we have any representation form microsoft left in the community?

jbottum:
@LeonardAukea thanks for this feature request.   We do not have active Azure contacts.   If you have any suggestions, it would help.

wadhah101:
I am currently working on adding this plugin. I would appreciate any feedback

LeonardAukea:
@wadhah101 Awesome, we have some internal ongoing work as well. maybe we can share notes at some point!

wadhah101:
@LeonardAukea  How can i contact you ? 
feel free to reach me on my email `wadhah.mahroug15@gmail.com`

PR #7091, opened 2023-04-12
Add Profile plugin for azure: Azure AD workload identity

wadhah101:
## Closes https://github.com/kubeflow/kubeflow/issues/6740

For Azure AD workload identity You have to : 
- Add annotations to the service account `azure.workload.identity/client-id : "xxxxxxxx"`
- Add a label to every pod willing to use that identity  `azure.workload.identity/use : "true"`

I have added the necessary code to modify the annotation for default service account. 

I am still looking for a way to modify pod labels across a profile. The only solution i think is to add a PodDefault component with that label name

I am awaiting feedback before on this before i write unit tests and add docs. Thank you 😄

### DIFF
--- a/components/profile-controller/controllers/plugin_ad_workload_identity_azure.go
+++ b/components/profile-controller/controllers/plugin_ad_workload_identity_azure.go
@@ -1,0 +1,160 @@
+/*
+Copyright 2019 The Kubeflow Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


components/profile-controller/controllers/plugin_ad_workload_identity_azure.go
```diff
@@ -0,0 +1,107 @@
...
+	AzureIdentityClientId              string `json:"identityClientId,omitempty"`
+	AzureIdentityTenantId              string `json:"identityTenantId,omitempty"`
+	AzureServiceAccountTokenExpiration string `json:"serviceAccountTokenExpiration,omitempty"`
+}
+
+// ApplyPlugin will grant Azure workload identity to service account DEFAULT_EDITOR
+func (azure *AzureAdWorkloadIdentity) ApplyPlugin(r *ProfileReconciler, profile *profilev1.Profile) error {
+	logger := r.Log.WithValues("profile", profile.Name)
+
+	logger.Info("Setting up workload identity", "ClientId", azure.AzureIdentityClientId)
```
AhmedGrati, line 51: I think preferably, log messages should start in lowercase.

juliusvonkohout:

Issue #7085, opened 2023-04-07
Kubeflow PRs for Group Support: Enabling RBAC Based on User Groups


axel7083:
## Descriptions

I am opening a dedicated issue to follow on the PRs opened related to the group support.

The goal is to add support for groups in kubeflow. Currently only supporting RBAC based on user-id. The oidc-authservice will send the "kubeflow-userid" and "kubeflow-groups", but currently the "kubeflow-groups" header is ignored. 

### kubeflow/kubeflow
- [ ] https://github.com/kubeflow/kubeflow/pull/7022
- [ ] https://github.com/kubeflow/kubeflow/pull/7084

> The centraldashboard would also requires some modification to be able to comply with the modified kfam.

### kubeflow/pipeline
- [ ] https://github.com/kubeflow/pipelines/pull/8918

### kubeflow/manifest
- [ ] https://github.com/kubeflow/manifests/pull/2403

### kubeflow/katib
- [ ] https://github.com/kubeflow/katib/pull/2145

### kserve/models-web-app

This does not requires any modification in its [Dockerfile](https://github.com/kserve/models-web-app/blob/v0.10.0/Dockerfile) it clone the kubeflow/kubeflow repository and use `components/crud-web-apps/common/backend`

## Related issues

https://github.com/kubeflow/kubeflow/issues/4188
https://github.com/kubeflow/kubeflow/issues/5071
https://github.com/kubeflow/kubeflow/issues/4998

/kind feature
/area multiuser

kimwnasptd:
@axel7083 it's really great to see how you are approaching this e2e and touching all KF components. Really great work!

I think we'll also need to start including in the mix how we'll end up having headers and support in VirtualServices as well. The main problem being how to parse list values (since a user could belong in multiple groups).

In any case would you be able to attend the Notebooks WG meeting on June 8 and give an overview of your current work? This will really help ensure we are all on the same page and coordinate the last pieces and also have a proposal that summarizes this architecture.

thesuperzapper:
@kimwnasptd In a related discussion, it's always been strange that we pass the user-id header around as plain text, rather than using a JWT and extracting the user-id from it.

Our current user-id header means that we have to be extremely careful about restricting which traffic can reach the pods, because "impersonating" a user is as simple as passing their user-id in the header.

[...]

PR #7093, opened 2023-04-13
profile-controller: adding contributors field & support for "Group" kind

axel7083:
# Description
In the CRD of the Profile, the `kind` field of the `owner` must be **User** since it is used to defined the AuthorizationPolicy allowing user to use the Notebooks.

When changing the kind field to Group, the role binding will properly be set, but the change will not reflect in a proper AuthorizationPolicy configuration. This resulting in a `RBAC unauthorized` result when trying to use a Jupyter notebook for example.

## AuthorizationPolicy and group support

According to the [oidc-authservice](https://github.com/arrikto/oidc-authservice) documentation, the groups will be propagated in a header, separated by commas. The AuthorizationPolicy definition does not allow some kind of parsing on strings, only operation on arrays, therefore the solution is to used the string as it is provided, and use the values to match the group as it would be in the header. 

### Example
Given the allowed group `admin` among many others, we generate the following AuthorizationPolicy when condition:
```yaml
....
    - when:
        - key: request.headers[kubeflow-groups]
          values:
            - admin
            - '*,admin'
            - '*,admin,*'
            - admin,*
...
```

## Adding contributors field

The Profile definition only allowed to define one owner, creating for us the RoleBinding and the AuthorizationPolicy. To simplify the process of adding default contributors, I added the contributors field, being an array of Subject (name string and kind string).
When reconciling, they will be used to create proper RoleBinding and AuthorizationPolicy.

### Example of Profile

Here is an example of a Profile definition using the `contributors` field and Group support.
```yaml
apiVersion: kubeflow.org/v1
kind: Profile
metadata:
  name: team-a
spec:
  owner:
    kind: Group
    name: team-a:admin  
  contributors:
    - kind: Group
      name: team-a:member
    - kind: User
      name: consultant@email.com
```

## Related

This PR is part of the following list of PRs https://github.com/kubeflow/kubeflow/issues/7085 aiming to add support for Groups in Kubeflow.


juliusvonkohout:

axel7083:
> Can you join the next notebooks Working Group or Manifest working Group meeting? We are always looking for help. And this is where you can present your PR. Just propose it on the agenda and join.
> 
> https://www.kubeflow.org/docs/about/community/#kubeflow-community-calendars has the calendar and meeting notes.
> 
> You can also reach us on slack in the channel wg-notebooks and wg-manifests

Sorry I cannot join todays meeting.

This PR would probably be outdated if the oidc-authservice is migrated to oauth2-proxy (see related ).

The https://github.com/kubeflow/manifests/pull/2409 allows to use oauth2-proxy which propage the `kubeflow-userid` as plain text but also the `authorization` header, containing the full JWT token, instead of checking the userid header, it is better to check for the JWT token since it add a security to ensure the validity of the request.

juliusvonkohout:
ok then i will

/close

since oauth2 is merged.
Please try to join the next meetings anyway. The bandwidth for discussing here is rather limited.

PR #7107, opened 2023-04-26
docs: update example notebooks spec

jlewi:
It looks like the notebook images no longer rely on start.sh.

It looks like the Docker images now use the init system to run jupyterlab with the correct arguments https://github.com/kubeflow/kubeflow/blob/master/components/example-notebook-servers/jupyter/Dockerfile

I could find no start.sh script in the repo


juliusvonkohout:

PR #7110, opened 2023-04-27
use http-port as named port in notebook controller

### LINKED ISSUE #7092: Update notebook port to `http- + instance.Name`
/kind feature

It seems that istio starts after some other sidecars that we are using and isn't able to write its rules correctly. Is there a reason that `notebook-port` is used? Can it be changed to follow the istio convention?

Update the notebook controller generateStatefulset to use `"http-"+Instance.name` as the port name and update the `targetPort` on the service to use the same

I'm happy to make this change, but wanted to open it up to see if this made sense. I tested it on a small environment and it seems to do what I want.

OmriShiv:
Fixes #7092 

I seem to be having issues with other sidecars not working due to istio starting after them and rewriting iptables. This helps istio out by carrying the istio expected port to the notebook


juliusvonkohout:

Issue #7111, opened 2023-04-28
models icon (kserve) not inheriting iron icon colors


ianonavy:
/kind bug

There is a problem with the way Kubeflow renders the models icon where it does not correctly inherit the colors. This is mentioned in https://github.com/kubeflow/kubeflow/pull/6656, but the workaround is insufficient to solve the root cause.

**What steps did you take and what happened:**

1. Navigate to Home
2. Observe the color of the KServe icon (black, although the rest are gray)
![screenshot of the sidebar for Kubeflow showing KServe's icon is black and the rest are gray](https://user-images.githubusercontent.com/790279/235245109-68c3000d-0b08-46b1-baf1-8ff44d25e9a4.png)
3. Navigate to KServe Endpoints
4. Observe the color of the KServe icon (black, should have changed to white)
![screenshot of the Kubeflow navbar showing the KServe icon in the active state is black](https://user-images.githubusercontent.com/790279/235245191-1a43403c-8855-4067-b5ec-ffb1139f8dbe.png)

**What did you expect to happen:**
The SVG should inherit the icon color from CSS in both the inactive and active state.
![screenshot of the Experiments KFP active state showing the icon is white](https://user-images.githubusercontent.com/790279/235245247-58d70ce6-a45a-46e7-aa1c-63837be3cb35.png)

**Anything else you would like to add:**
I put this SVG into Figma, flattened the paths and re-exported. Now it works:

```svg
<svg viewBox="0 0 24 24" preserveAspectRatio="xMidYMid meet" focusable="false" style="pointer-events: none; display: block; width: 100%; height: 100%;">
<path d="M2.20122 12.2692C1.93293 12.0307 1.93293 11.8319 2.20122 11.6729L6.35976 8.0954C6.58333 7.9364 6.82927 7.97615 7.09756 8.21465L7.90244 8.93015C8.0813 9.08915 8.01423 9.2879 7.70122 9.52641L4.88415 12.0307L7.70122 14.5349C7.88008 14.6939 7.88008 14.8927 7.70122 15.1312L7.03049 15.7274C6.71748 16.0454 6.42683 16.0852 6.15854 15.8467L2.20122 12.2692Z" />
[...]

PR #7113, opened 2023-04-29
Update models.svg

juliusvonkohout:
Fixes  https://github.com/kubeflow/kubeflow/issues/7111

@kimwnasptd @ianonavy


ianonavy:
LGTM! I didn't do a full end-to-end test with a custom container image off this branch, but copying the file from this revision directly into the DOM appears to work. Since it's just a static file, I don't see why this change wouldn't work.

kimwnasptd:
Thanks for this @juliusvonkohout!

/lgtm
/approve

PR #7122, opened 2023-05-04
ARM docker images for kubeflow components

vinaynair:
/kind feature

Support for ARM VMs
With KF 1.7 we have Istio 1.16 which supports ARM,  **but** sadly kubeflow images don't have an ARM platform build so that one can run kubeflow locally on Apple M1 Mac **or**  on cloud  ARM based nodes  such as AWS Graviton

Updated github actions section that uses multi arch docker publish steps to support linux/arm64 platform

I have tested it locally by building and pushing docker to my personal repo at https://hub.docker.com/repositories/vinaychandran 
It would be great if docker hub registry that kubeflow uses by default had these ARM variants.

### DIFF
--- a/.github/workflows/centraldb_docker_publish.yaml
+++ b/.github/workflows/centraldb_docker_publish.yaml
@@ -11,7 +11,7 @@ on:
 env:
   DOCKER_USER: kubeflownotebookswg
   IMG: kubeflownotebookswg/centraldashboard
-  ARCH: linux/ppc64le,linux/amd64
+  ARCH: linux/ppc64le,linux/amd64,linux/arm64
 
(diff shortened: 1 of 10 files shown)


kimwnasptd:
Thanks for the PR @vinaynair! This should be enough indeed for this repo.

Tried to deploy KF on ARM as well. Looks like we'll need to update [KServe web app](https://github.com/kserve/models-web-app) as well and we should be set.

/lgtm
/approve

kimwnasptd:
Realized that this PR was made in the `v1.7-branch`. @vinaynair could you make this PR to the `master` branch as well?

PR #7124, opened 2023-05-05
feat(notebooks): only display allocatable GPU configuration when creating notebook

axel7083:
## Description

When creating a notebook, the number of GPU displayed does not correspond to the real hardware allocatable. This PR change the behavior of the `/jupyter/api/gpus` to return the maximum number of GPUs allocatable based on Node resources. 

<details>
  <summary> <strong>/jupyter/api/gpus </strong> response example</summary>
The following response state that we can create a notebook with a maximum of 3 NVIDIA GPUs simultaneously. it <strong>does not</strong> mean only 3 GPUs are available on the cluster, it mean the node with the most GPUs available has 3.

  ```json
  {
    "allocatable": {
      "amd.com/gpu": 0,
      "nvidia.com/gpu": 3
    },
    "status": 200,
    "success": true,
    "user": "email@email.com"
  }
  ```
  
</details>

In the gpus `formGroup`, the order of the following field: `num` and `vendors` have been inverted, since the number of GPUs is depending on the vendor, it makes sense to choose the vendor first.

This could be extended to `memory` and `cpu` but those resources are usually less an issue.

## Related issues
- https://github.com/kubeflow/notebooks/issues/90
- https://github.com/kubeflow/notebooks/issues/97

### DIFF
--- a/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/api/pod.py
+++ b/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/api/pod.py
@@ -15,3 +15,6 @@ def get_pod_logs(namespace, pod, container, auth=True):
     return v1_core.read_namespaced_pod_log(
         namespace=namespace, name=pod, container=container
     )
+
+def list_all_pods(field_selector: str):
+    return v1_core.list_pod_for_all_namespaces(field_selector=field_selector)
--- a/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py
@@ -97,10 +97,11 @@ def get_notebook_events(notebook_name, namespace):
 
 
 @bp.route("/api/gpus")
-def get_gpu_vendors():
+def get_gpus_allocatable():
     """
-    Return a list of GPU vendors for which at least one node has the necessary
-    annotation required to schedule pods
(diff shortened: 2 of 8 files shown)


axel7083:
Just saw @kimwnasptd has a PR for [web-apps: Implement watch to avoid constant polling from backend to K8s](https://github.com/kubeflow/kubeflow/pull/7080/files) that could probably be used/adapted ? It would be a better alternative than pulling every time.

kimwnasptd:
Hey @axel7083, I have some concerns with calculating the value to show there dynamically. The problem is in case that folks might want to use only spot-instances for GPUs. 

This means that when the Notebook CR will get created there won't by any GPU nodes. In this case some external provider will spin up a spot-instance to accomodate the Notebook.

So if the value gets dynamically calculated then we completely miss this use-case

axel7083:
> Hey @axel7083, I have some concerns with calculating the value to show there dynamically. The problem is in case that folks might want to use only spot-instances for GPUs.
> 
> This means that when the Notebook CR will get created there won't by any GPU nodes. In this case some external provider will spin up a spot-instance to accomodate the Notebook.
> 
> So if the value gets dynamically calculated then we completely miss this use-case

Oh I see ! I have not thought about this case scenario. Maybe the good way to go, would be to have the possibilities to choose between those two options.
For people having dynamic allocation, it is of course the current scenario which is the best, but for deployment on fixed number of machines, knowing the reel configuration can be very interested.

juliusvonkohout:

Issue #6981, opened 2023-02-20
Rename "Notebook" to "Workbench"


StefanoFioravanzo:
[Rename Notebooks to Workbenches]

In Kubeflow, we use "Notebook" to refer to generic web-based development environments. Out of the box, a user can create instances of JupyterLab, VSCode, and RStudio, possibly adding custom IDEs.

"Notebooks" is a term used since the beginning of the Kubeflow project since, at first, there was support just for Jupyter Notebooks, which then evolved into JupyterLab and more IDEs.

With the word "Notebooks" we interchangeably refer to:

1. A JupyterLab instance, but often VSCode and RStudio instances. We call them "Notebooks" because that's what the "Jupyter Web App" manages.
2. A notebook file within a JupyterLab instance

We propose using the term "Workbench" to refer to a generic web-based development environment and break free from any confusion in the notebook terminology. Workbench is already used in some projects to refer to a working/development environment. It is a generic enough term for web applications that exposes a programming/CLI interface.

To transition to this new terminology, we would need to scope two main (independent) threads of work:

1. UI and documentation: update the web app UIs to refer uniformly to Workbenches instead of Notebooks and the Kubeflow docs.
2. Technical discussion: Controllers and CRDs, web app changes, renames, factoring out of common code

cc @kimwnasptd @kubeflow/wg-notebooks-leads

thesuperzapper:
@tzstoyanov @kimwnasptd @StefanoFioravanzo I think if we are going to make any breaking changes to the CRD, we should explore other ways to make the Notebook CRD more powerful instead of just a simple rename.

I have made a proposal in https://github.com/kubeflow/kubeflow/issues/7156, for what I am calling "Kubeflow Notebooks 2.0" (aka "Kubeflow Workspaces").

juliusvonkohout:
> @tzstoyanov @kimwnasptd @StefanoFioravanzo I think if we are going to make any breaking changes to the CRD, we should explore other ways to make the Notebook CRD more powerful instead of just a simple rename.
> 
> I have made a proposal in #7156, for what I am calling "Kubeflow Notebooks 2.0" (aka "Kubeflow Workspaces").

That will be something for a second PR after the cosmetic changes here.

thesuperzapper:
I __strongly believe__ that we should NOT do a rename unless there is a technical change, as that would just be confusing to users who already know about "Kubeflow Notebooks", which is effectively a brand at this point rather than a description.

[...]

PR #7159, opened 2023-06-09
Notebooks: Rename "Notebook" to "Workbench" in the WEB GUI

tzstoyanov:
Kubeflow notebooks were initially created from JupyterLab Notebook. That functionality evolved, now other web-based development environments can be used as well - VSCode, RStudio and others can be added in future. The term "notebook" is a bit misleading and does not reflect all recent changes and enhancements. That's why a proposal was made to replace it with the term "workbench" - issue 6981.
This patch renames "notebook" to "workbench" in the Kubeflow WEB GUI.

issue: https://github.com/kubeflow/kubeflow/issues/6981

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/form.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/form.py
@@ -189,7 +189,7 @@ def set_notebook_tolerations(notebook, body, defaults):
         if group["groupKey"] != tolerations_group_key:
             continue
 
-        log.info("Appending Notebook tolerations: %s", group["tolerations"])
+        log.info("Appending Workbench tolerations: %s", group["tolerations"])
         notebook_tolerations.extend(group["tolerations"])
         return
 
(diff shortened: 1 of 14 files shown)


thesuperzapper:
@juliusvonkohout @tzstoyanov I strongly believe that we should NOT do a rename unless there is a technical change, as that would just be confusing to users who already know about "Kubeflow Notebooks", which is effectively a brand at this point rather than a description.

thesuperzapper:
@juliusvonkohout @tzstoyanov I strongly believe that we should NOT do a rename unless there is a technical change, as that would just be confusing to users who already know about "Kubeflow Notebooks", which is effectively a brand at this point rather than a description.

/hold

juliusvonkohout:
> @juliusvonkohout @tzstoyanov I strongly believe that we should NOT do a rename unless there is a technical change, as that would just be confusing to users who already know about "Kubeflow Notebooks", which is effectively a brand at this point rather than a description.
> 
> /hold

We discussed this for almost one year in the meetings ;-) it was planned for 1.6 already. We even had community polls regarding the name.

The name is really just technically wrong and should not manifest as a brand. The longer it stays the longer it will be wrong.

Let us rather focus on what you want to achieve on the technical level afterwards.

PR #7171, opened 2023-06-16
Disabling multiple architecture build for PRs

pranavpandit1:
this PR ensures that PRs will only trigger an AMD64 build. The multiarc build will only occur when a PR is merged into the main or when a new tag is created.

There is unfortunately, no way to fully test this without merging this PR.


pranavpandit1:
@kimwnasptd: these changes are intended as per the discussion on [https://github.com/kubeflow/kubeflow/issues/6684](url).
Kindly review, so accordingly I can raise the requests for other components too.

PR #7175, opened 2023-06-19
Add GH action integration testing pvcviewer controller

TobiasGoerke:
As discussed in https://github.com/kubeflow/kubeflow/pull/6876, we're adding a GH action that integration tests the pvcviewer controller / its manifests.

The Action is untested and was copied off the notebook controller which works the same way.

This PR is part of three pvcviewer controller workflows. The others are: 
* #7173
* #7174


kimwnasptd:
Awesome!

/lgtm
/approve

PR #7211, opened 2023-07-19
(frontend): Adding style for main drawer scrollbar

axel7083:
When we have a lot of item, on the maindrawer, the scrollbar appear, and the default one is pretty ugly and take a lot of space.
I simply copy past the style from kubeflow/pipeline elements and change the color to match the text.

| before | after |
|---|---|
| image | image |


kimwnasptd:
Nice!

/lgtm
/approve

PR #7230, opened 2023-08-07
fix: viewer-spec.yaml and trigger build

TobiasGoerke:
The [pvcviewer pipeline](https://github.com/kubeflow/kubeflow/actions/workflows/pvcviewer_controller_docker_publish.yaml) has not yet run successfully.
Thus, there is no public image that could be used to test the manifests.
As far as I can tell, this should be the last effort to get the volumes viewer working. 

Also, I snuck in another commit, fixing two fields of the viewer-spec.yaml

### DIFF
--- a/components/crud-web-apps/volumes/manifests/base/deployment.yaml
+++ b/components/crud-web-apps/volumes/manifests/base/deployment.yaml
@@ -21,7 +21,7 @@ spec:
         - name: APP_SECURE_COOKIES
           value: $(VWA_APP_SECURE_COOKIES)
         - name: VOLUME_VIEWER_IMAGE
-          value: filebrowser/filebrowser:latest
+          value: filebrowser/filebrowser:v2.24.2
         volumeMounts: 
         - name: viewer-spec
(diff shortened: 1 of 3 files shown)


TobiasGoerke:
@kimwnasptd my filebrowser [PR](https://github.com/filebrowser/filebrowser/pull/2145) got finally merged. All uploads are now chunked using the tus protocol and the "old" upload mechanism has been removed entirely in favor of tus. 
This will provide a great benefit to Kubeflow users as they can upload data no matter how large and despite proxy size limitations.

I've now fixed the new filebrowser version in the manifests. This PR should be the last part of the pvcviewer - all local tests and the pvcviewer work great for me. We'll just need a public image. 

I'll be available over Kubeflow's Slack if you'd like to synchronize during the release phase.

kimwnasptd:
@TobiasGoerke you are on 🔥, awesome work!

/lgtm
/approve

kimwnasptd:
Can you also send a follow-up PR to cherry-pick this commit to the `v1.8-branch` branch?

PR #7234, opened 2023-08-11
[kubeflow 1.8] release: bump container images to tag 1.8.0-rc.0 and bump release version

DnPlas:
Following the [releasing steps](https://github.com/kubeflow/kubeflow/tree/master/releasing#steps-for-releasing), update the manifests to use the v1.8.0-rc.1 tag and bump version in `version/VERSION`.

cc: @annajung @kimwnasptd


annajung:
@DnPlas From the file changes and GitHub actions logs, it's hard to tell why it's failing. Can you help debug this further to gather better logs for the problem? possibly using https://github.com/marketplace/actions/debugging-with-ssh or outputting data as artifact?

kimwnasptd:
@DnPlas thanks for the PR! Looking into those to unblock us. Most probably it's a race on how we check for some manifests to be there. I'll keep you posted

kimwnasptd:
I'm pretty sure we have the same race in the tests in manifests, where we check for the Trial CR before it's created https://github.com/kubeflow/manifests/pull/2508#issuecomment-1681068583

The solution here should be in each test to either:
1. wait for the Deployment/StatefulSet to create the Pods first and then wait for the Pods
2. wait for the Pods to get created, before waiting for them with `kubectl wait`

So we have a race. I'll send a follow-up PR to fix this. Ideally I'd like to avoid adding `sleep` commands as there's always a chance the bug might happen again. If though it will take way too much time to code I'll switch to this simple approach to unblock us for now

DnPlas:
I submitted a set of commits correcting the VERSION and image tags (`1.8.0-rc.0 -> v1.8.0-rc.0`). I also added the pvc viewer component to the `update-manifests-images` script to generate the corresponding images, and finally added the suggestions from @kimwnasptd to fix the CI.

Kindly review the changes, if everything looks good, let's merge.

DnPlas:
One of the failing [latest CI run](https://github.com/kubeflow/kubeflow/actions/runs/5973620001/job/16206223284#step:9:28) shows the following error:

```
deployment.apps/tensorboard-controller-deployment created
error: timed out waiting for the condition on deployments/tensorboard-controller-deployment
Error: Process completed with exit code 1.
```

I increased the timeout, but I'm not sure this is helping.

DnPlas:
Reopening PR as it was mistakenly closed by  #7263

kimwnasptd:
Thanks for coordinating and pushing this @DnPlas!

/lgtm
/approve

PR #7252, opened 2023-08-29
Don't use sed to update deployed image in integration tests

kimwnasptd:
This is a mirror PR to look into why https://github.com/kubeflow/kubeflow/pull/7234 is failing in the first place

cc @DnPlas

### DIFF
--- a/.github/workflows/centraldb_intergration_test.yaml
+++ b/.github/workflows/centraldb_intergration_test.yaml
@@ -7,6 +7,10 @@ on:
       - master
       - v*-branch
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   IMG: centraldashboard
   TAG: intergration-test
@@ -26,6 +30,7 @@ jobs:
 
     - name: Build CentralDashboard Image
       run: |
+        TAG=${{env.TAG}}
         cd components/centraldashboard
         ARCH=linux/ppc64le make docker-build-multi-arch
         ARCH=linux/amd64 make docker-build-multi-arch
@@ -51,8 +56,13 @@ jobs:
         cd components/centraldashboard/manifests
         kubectl create ns kubeflow
 
-        export CURRENT_CENTRALDB_IMG=docker.io/kubeflownotebookswg/centraldashboard:latest
-        export PR_CENTRALDB_IMG=${{env.IMG}}:${{env.TAG}}
+        kustomize build overlays/kserve | kubectl apply -f -
 
-        kustomize build overlays/kserve | sed "s#$CURRENT_CENTRALDB_IMG#$PR_CENTRALDB_IMG#g" | kubectl apply -f -
-        kubectl wait pods -n kubeflow -l app=centraldashboard --for=condition=Ready --timeout=300s
+        DEPLOYMENT=centraldashboard
+        CONTAINER=centraldashboard
+
+        IMG=${{env.IMG}}:${{env.TAG}}
+        kubectl patch deployment $DEPLOYMENT -n kubeflow --patch \
+          '{"spec": {"template": {"spec": {"containers": [{"name": "'"$CONTAINER"'","image": "'"$IMG"'"}]}}}}'
+
+        kubectl wait deploy -n kubeflow $DEPLOYMENT --for=jsonpath='{.status.readyReplicas}'=1 --timeout=300s
--- a/.github/workflows/centraldb_angular_intergration_test.yaml
+++ b/.github/workflows/centraldb_angular_intergration_test.yaml
@@ -7,6 +7,10 @@ on:
       - master
       - v*-branch
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   IMG: centraldashboard-angular
   TAG: intergration-test
@@ -20,10 +24,10 @@ jobs:
       with:
         fetch-depth: 0
 
-    - name: Build CentralDashboard-Angular Image 
+    - name: Build CentralDashboard-Angular Image
       run: |
         cd components/centraldashboard-angular
-        make docker-build 
+        make docker-build
 
     - name: Install KinD
       run: ./components/testing/gh-actions/install_kind.sh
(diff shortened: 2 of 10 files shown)


kimwnasptd:
The problem here is the lines that update the image, to use the one built and pushed in the local cache 
https://github.com/kubeflow/kubeflow/blob/d1c40f916225ecb40826400595ea99096a2adbb7/.github/workflows/centraldb_intergration_test.yaml#L54

Specifically it's using the `latest` tag to try and replace the image value, which works for master but not for release branches where we update the tags.

The solution here would be to use `kubectl patch` instead 
https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/

kimwnasptd:
@DnPlas I got to the bottom of this. I'll have the PR cleaned EOD today.

cc @thesuperzapper on the review, since this is blocking the release (it's an early ping, will ping again once it's ready to be reviewed)

DnPlas:
@kimwnasptd thanks for the effort and thanks for the ping! Will keep an eye on this :)

kimwnasptd:
@thesuperzapper this is ready for review

kimwnasptd:
@thesuperzapper another ping for this. This is blocking the release and Notebooks WG is the only pending component

.github/workflows/centraldb_intergration_test.yaml
```diff
@@ -26,6 +30,7 @@ jobs:
 
     - name: Build CentralDashboard Image
       run: |
+        TAG=${{env.TAG}}
```
thesuperzapper, line 33: I don't understand what this is doing.

.github/workflows/centraldb_intergration_test.yaml
```diff
@@ -51,8 +56,13 @@ jobs:
...
-        export CURRENT_CENTRALDB_IMG=docker.io/kubeflownotebookswg/centraldashboard:latest
-        export PR_CENTRALDB_IMG=${{env.IMG}}:${{env.TAG}}
+        kustomize build overlays/kserve | kubectl apply -f -
+
+        DEPLOYMENT=centraldashboard
+        CONTAINER=centraldashboard
+
+        IMG=${{env.IMG}}:${{env.TAG}}
+        kubectl patch deployment $DEPLOYMENT -n kubeflow --patch \
+          '{"spec": {"template": {"spec": {"containers": [{"name": "'"$CONTAINER"'","image": "'"$IMG"'"}]}}}}'
```
thesuperzapper, line 66: @kimwnasptd I don't like that this applies the manifests and then patches them after, this will result in a generation of the pod that fails to run, quickly followed by one with the correct image.

I don't understand what was not working with the old `sed` approach, which modified the manifests before applying them?

thesuperzapper (changes requested):
@kimwnasptd I am not sure what this PR is trying to achieve, and I don't agree with its current implementation either way.

.github/workflows/centraldb_intergration_test.yaml
kimwnasptd: It makes sure the image build, and stored in docker, will use the TAG defined in the `env` and not the default one which is
https://github.com/kubeflow/kubeflow/blob/master/components/centraldashboard/Makefile#L2

.github/workflows/centraldb_intergration_test.yaml
kimwnasptd: > @kimwnasptd I don't like that this applies the manifests and then patches them after, this will result in a generation of the pod that fails to run, quickly followed by one with the correct image.

Why is this a hard problem? The deployment will be patched immediately, and we wait for the Deployment object either way

> I don't understand what was not working with the old sed approach, which modified the manifests before applying them?

Previously the flow was:
1. build the manifests
    1. These manifests in `master` would have `latest` tag in images
    2. In a release branch we update the tag, so it would be something like `v1.8.0-rc.0`
3. `sed` was used change the image with tag `latest`, to `${{env.TAG}}`
4. The above worked for `master`, which has tag `latest` but doesn't for release branches

So the script works for `master`, because `latest` image tag is used in manifests, but fails in release branches which use a different tag in manifests.

kimwnasptd:
@thesuperzapper added some more explanation comments, as I wasn't aware you had no idea how the CI was working all this time.

I summarised the current problem in https://github.com/kubeflow/kubeflow/pull/7252#discussion_r1313941985.

kimwnasptd:
@thesuperzapper ping, this is blocking the release

thesuperzapper:
Sharing what I messaged to @DnPlas.

@kimwnasptd if I understand correctly, you are proposing:

1. apply the “default” manifests (with whatever image that contains)
2. quickly try to patch the deployment with a “testing” tag
3. wait for the deployment to have `readyReplicas` equal to 1

This effectively sets up a race condition every time, making it very likely (if not certain) that step 3 can pass even if the image being tested fails to start. This is because assuming the “default” image works, (but the “testing” tag fails to start) there will still be one ready-replica, but of the "default" image.

I will raise a separate PR that just fixes the `sed` to use a pattern capable of replacing the image with an arbitrary tag after the `:`, rather than assuming its "latest".

.github/workflows/centraldb_intergration_test.yaml
thesuperzapper: I was trying to say that this is a redundant line, as the global `env` affects all job steps.

.github/workflows/centraldb_intergration_test.yaml
thesuperzapper: See https://github.com/kubeflow/kubeflow/pull/7252#issuecomment-1707359973

thesuperzapper:
@kimwnasptd @DnPlas I have raised https://github.com/kubeflow/kubeflow/pull/7263 which does not have the problems described in https://github.com/kubeflow/kubeflow/pull/7252#issuecomment-1707359973

.github/workflows/centraldb_intergration_test.yaml
thesuperzapper: I meant that this line is redundant because all job steps get all the global `env` exported into them.

thesuperzapper:
Closing as https://github.com/kubeflow/kubeflow/pull/7263 was merged.

/close

PR #7260, opened 2023-09-01
Only load spawner_ui_config.yaml once every 10 minutes. closes #7259

### LINKED ISSUE #7259: Large amounts of notebook servers 50+ causes jupyter webapp slowness.
The way the jupyter web app is loaded causes slowness when the there are a lot of notebooks in a single namespace.

The notebook_dict_from_k8s_obj function calls the kubernetes api for events on each notebook in the namespace. With 50 notebooks we're seeing around 1.8-2 second response time. This is really bogging down the UI.

Relevant Code:
https://github.com/kubeflow/kubeflow/blob/d1c40f916225ecb40826400595ea99096a2adbb7/components/crud-web-apps/jupyter/backend/apps/common/routes/get.py#L52-L57

https://github.com/kubeflow/kubeflow/blob/d1c40f916225ecb40826400595ea99096a2adbb7/components/crud-web-apps/jupyter/backend/apps/common/utils.py#L122-L143

https://github.com/kubeflow/kubeflow/blob/d1c40f916225ecb40826400595ea99096a2adbb7/components/crud-web-apps/jupyter/backend/apps/common/status.py#L46-L49

https://github.com/kubeflow/kubeflow/blob/d1c40f916225ecb40826400595ea99096a2adbb7/components/crud-web-apps/jupyter/backend/apps/common/status.py#L160-L175

boarder7395:
The process_gpus function calls load_spawner_ui_config for every notebook in a namespace when the get_notebooks endpoint is called. This change updates so the call is made once every 10 minutes. 

I was unsure if the expectation with this component is that users will be able to change the configmap without restarting the deployment. If not it would be better to load this config once at initialization. 

If that route is preferred I'd be happy to update this PR with those changes.

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/utils.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/utils.py
@@ -2,6 +2,7 @@ import os
 import random
 import string
 
+from cachetools.func import ttl_cache
 from kubernetes import client
 from werkzeug import exceptions
 
@@ -13,9 +14,7 @@ log = logging.getLogger(__name__)
 
 FILE_ABS_PATH = os.path.abspath(os.path.dirname(__file__))
 
-NOTEBOOK_TEMPLATE_YAML = os.path.join(
-    FILE_ABS_PATH, "yaml/notebook_template.yaml"
-)
+NOTEBOOK_TEMPLATE_YAML = os.path.join(FILE_ABS_PATH, "yaml/notebook_template.yaml")
 LAST_ACTIVITY_ANNOTATION = "notebooks.kubeflow.org/last-activity"
 
 # The production configuration is mounted on the app's pod via a configmap
@@ -41,6 +40,7 @@ def load_notebook_template(**kwargs):
     return helpers.load_param_yaml(NOTEBOOK_TEMPLATE_YAML, **kwargs)
 
 
+@ttl_cache(ttl=60 * 10)
 def load_spawner_ui_config():
     for config in CONFIGS:
         config_dict = helpers.load_yaml(config)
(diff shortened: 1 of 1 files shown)


boarder7395:
@elikatsis Can you take a look at this change and let me know if there is a need to keep the configmap refreshed?

thesuperzapper:
@boarder7395 is this proposed change a result of some issue you encountered?

Also, if we do add caching to this, I think 10 minutes is probably too long, perhaps 60 seconds is better.

boarder7395:
@thesuperzapper Yes sorry I linked the PR to the issue report instead of the other way around. 

Here is the issue: https://github.com/kubeflow/kubeflow/issues/7259

This TLDR is that when I had a 100 users using kubeflow we started hitting performance issues. I narrowed the issue to the jupyter web app and then did line by line performance testing. In doing this I narrowed the issue to the loading of the jupyter web app config for every notebook for every user. 

How the logic currently works is for each page load (runs continuously to keep the page up to date) all notebooks are listed. Then for each notebook the config file is loaded from disk. This consumed all the available read speed from the nodes disk.

I agree 10 minutes is not ideal, I would prefer to load the config file once at launch time. Otherwise the code will do a read from disk every 60 seconds for each gunicorn worker.

thesuperzapper:
@boarder7395 I am in favor of 60 seconds, but I am interested to hear @kimwnasptd thoughts.

boarder7395:
@thesuperzapper Sounds good. While I'm making this change I ran into another issue: https://github.com/kubeflow/kubeflow/issues/7280. This one is a super edge case and likely won't happen to most teams, but I think we should add try catch logic with a warning message when the status of a node cannot be determined. 

Any issues with combining the fixes to both in this PR since they are both minor?

thesuperzapper:
@boarder7395 PRs are free, so just raise a separate one, in case one turns out to be more controversial!

thesuperzapper:
@boarder7395 well, it legally can't be merged until you fix the CLA issue: https://github.com/kubeflow/kubeflow/pull/7260/checks?check_run_id=16866274394

boarder7395:
I’ll get that signed for my company and check back

thesuperzapper:
@boarder7395 any updates on the CLA? 

BYW, it might be possible to sign it as an individual, depending on your situation.

@kimwnasptd I would love to merge something like this for 1.8, but it might be a bit late for that now.

boarder7395:
@thesuperzapper I got the go ahead from legal this afternoon, I can sign for the company on Monday. I have already signed the CLA as an individual.

boarder7395:
Also what’s the latest for getting this into 1.8?

boarder7395:
@thesuperzapper Got the CLA all signed. If possible can we include this fix https://github.com/kubeflow/kubeflow/pull/7281 into #7300 as well?

thesuperzapper:
@boarder7395, I discussed with @kimwnasptd, and we are happy to merge.

/lgtm
/approve

PR #7274, opened 2023-09-11
fix: remove incorrect space in `TWA_APP_SECURE_COOKIES` config

hhk7734:
When I ran kustomize, I got the below.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: tensorboards-web-app-deployment
spec:
  template:
    spec:
      containers:
      - env:
        - name: APP_SECURE_COOKIES
          value: 'true '
```

https://github.com/kubeflow/kubeflow/blob/194a725119e9bf8a16db37a1d23f0e941ab7f1e4/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/settings.py#L3

```python
>>> import os
>>> SECURE_COOKIES = os.getenv("APP_SECURE_COOKIES", "true").lower() == "true"
>>> print(SECURE_COOKIES)
False
```

So removed a trailing space.


thesuperzapper:
@hhk7734 thanks heaps for this, sorry about the delay.

This will be fixed in the default manifests for 1.9.0-rc.0.

Issue #5449, opened 2020-12-08
Way to "import" existing namespaces


hightoxicity:
/kind question

**Question:**
Hi, I deployed main components of kubeflow (multitenant pipelines, istio, dex), I can create new namespaces for profiles using GUI, I can launch pipelines in new created namespace and it properly run.

Now, I have already created namespaces (not by kubeflow), and I would like to "import it" into kubeflow, I have seen that there is a CRD Profile (creating this object appears to not be enough) but what are all the pieces required to enroll those namespaces?

Thanks.

hightoxicity:
OK, only one annotation need to be added to the namespace to enroll:

```
apiVersion: v1
kind: Namespace
metadata:
  annotations:
    owner: admin@example.com
  name: example
```

yanniszark:
@hightoxicity when you say "import" it into Kubeflow, do you mean have it show up in CentralDashboard?
If yes, then we aim to make that simpler in 1.3. The issue / design is tracked here: https://github.com/kubeflow/kubeflow/issues/5390

hightoxicity:
Yes, view existing namespaces in the dash, and make kubeflow deploy (profile deployer) needed stuff in it.

thesuperzapper:
This is still important.
/reopen

juliusvonkohout:
Can someone link https://github.com/kubeflow/kubeflow/pull/7547 as solution?

PR #7276, opened 2023-09-12
Fix (profile controller): Profile controller fails to update profile owner, fixes #6576 and #5449

tzstoyanov:
Fixes issue #6576
Fixes issue #5449

This PR fixes the problem, described in issue #6576. Currently, the profile-controller fails to update the owner as it cannot verify the namespace owner. The PR implements solutions, suggested in the issue.  

Additional logic is added, to address issue #5449. If the namespace already exist, and it is not owned by any other profile - check if the namespace has annotation `transferToKubeflow: true`. If there is such annotation, take the ownership of that namespace and use it for the profile.

### DIFF
--- a/.github/workflows/prof_controller_unit_test.yaml
+++ b/.github/workflows/prof_controller_unit_test.yaml
@@ -14,7 +14,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v3
       with:
-        go-version: '1.17'
+        go-version: '1.21'
         check-latest: true
 
     - name: Run unit tests
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -105,7 +105,6 @@ type ProfileReconciler struct {
 // Automatically generate RBAC rules to allow the Controller to read and write Deployments
 func (r *ProfileReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	logger := r.Log.WithValues("profile", request.NamespacedName)
-	defaultKubeflowNamespaceLabels := r.readDefaultLabelsFromFile(r.DefaultNamespaceLabelsPath)
 
 	// Fetch the Profile instance
 	instance := &profilev1.Profile{}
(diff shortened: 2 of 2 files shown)


juliusvonkohout:
@tzstoyanov amazing effort!

thesuperzapper:
Thanks @tzstoyanov! 

@juliusvonkohout @kimwnasptd we need to do some thorough testing of this, but it looks promising.

My main questions before we dive in with testing are:

- In your testing, I assume that the other resources listed in https://github.com/kubeflow/kubeflow/issues/6576 are now also updated correctly:
    - `RoleBinding/namespaceAdmin`
    - `AuthorizationPolicy/ns-owner-access-istio`
    - `ResourceQuota/kf-resource-quota`
- While we are here, we may as well provide a way to explicitly allow the profile controller to take over a namespace that does not already have it listed as an `ownerReference`:
     - Does anyone have any thoughts about how best to achieve this? 

---

Separately, perhaps we should make a `Ready` status condition which reflects if the Namespace is healthy and managed by the profile controller (which will allow systems like Helm/ArgoCD to wait until the profile is ready before trying to deploy resources in the namespace).

Effectively it's about the controller creating an element of `status.conditions[]` with a: 
 - `reason` (think event type, e.g. "Ready") 
 - `status` flag (the current value of this event type, e.g. "True", "False", "Unknown")
 - `message` (a short human-readable explanation of why it's in this state) 

We do something [very similar in the Notebooks controller](https://github.com/kubeflow/kubeflow/blob/194a725119e9bf8a16db37a1d23f0e941ab7f1e4/components/notebook-controller/api/v1/notebook_types.go#L46-L63). 

It's a bit confusing, but is described [here in Kubernetes Style Guide](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties), and [cert-manager does something similar](https://github.com/cert-manager/cert-manager/blob/master/pkg/controller/certificaterequests/util/reporter.go#L107).

tzstoyanov:
> Thanks @tzstoyanov!
> 
> @juliusvonkohout @kimwnasptd we need to do some thorough testing of this, but it looks promising.
> 
> My main questions before we dive in with testing are:
> 
>     * In your testing, I assume that the other resources listed in [profile controller fails to update profile owner #6576](https://github.com/kubeflow/kubeflow/issues/6576) are now also updated correctly:
>       
>       * `RoleBinding/namespaceAdmin`
>       * `AuthorizationPolicy/ns-owner-access-istio`
>       * `ResourceQuota/kf-resource-quota`

The owner of  `RoleBinding/namespaceAdmin` and  `AuthorizationPolicy/ns-owner-access-istio` is updated correctly, but I cannot see any reference to the old user in the `ResourceQuota/kf-resource-quota`.  

> 
>     * While we are here, we may as well provide a way to explicitly allow the profile controller to take over a namespace that does not already have it listed as an `ownerReference`:
>       
>       * Does anyone have any thoughts about how best to achieve this?

I think that the easiest approach is the users to annotate manually the  namespaces that  has to be hijacked. The controller can take control of the namespace, if it is not already owned by a profile controller and there is  annotation
`transferToKubeflow: true`
or something like that.


>

juliusvonkohout:
@tzstoyanov can you present in the next meeting on September 28 and @thesuperzapper can you join as well? I have also found someone else who wants to work on the workbench stuff as well.

tzstoyanov:
> @tzstoyanov can you present in the next meeting on September 28 and @thesuperzapper can you join as well? I have also found someone else who wants to work on the workbench stuff as well.

yes, I'll join the meeting to discuss this PR, thanks Julius

rimolive:
@tzstoyanov Can you rebase it so we can lgtm/approve?

tzstoyanov:
@rimolive  - rebased in a new PR #7547

PR #7278, opened 2023-09-12
ci: read-only token-permissions for all workflows

### LINKED ISSUE #7277: Set workflows to run with read-only permissions
/kind feature

Kubeflow's workflows are running with write-all permissions. This puts the project at risk of supply-chain attacks.

I took a look at all the workflows and none of them seem to require significant permissions.

There are two solutions:
- all the workflows can be set with read-only top-level permissions; and/or
- the default workflow token can be set to read-only

I'll send a PR setting top-level permissions for all workflows. If you also/instead want to change the default token:

1. Open the repo settings
2. Go to [Actions > General](https://github.com/kubeflow/kubeflow/settings/actions)
3. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

My name is Pedro and I work with Google and the [Open Source Security Foundation (OpenSSF)](https://www.openssf.org/) to improve the supply-chain security of the open-source ecosystem.

pnacht:
Fixes kubeflow/notebooks#80.

This PR ensures all workflows run with read-only permissions.

From what I could tell, the `triage_issues.yml` workflow uses the secret PAT instead of the default workflow token, and as such doesn't require additional permissions.

### DIFF
--- a/.github/workflows/centraldb_angular_backend_test.yaml
+++ b/.github/workflows/centraldb_angular_backend_test.yaml
@@ -4,6 +4,9 @@ on:
     paths:
       - components/centraldashboard-angular/backend/**
 
+permissions:
+  contents: read
+
 jobs:
   run-backend-unittests:
     name: Unit tests
--- a/.github/workflows/centraldb_angular_docker_publish.yaml
+++ b/.github/workflows/centraldb_angular_docker_publish.yaml
@@ -8,6 +8,9 @@ on:
       - components/centraldashboard-angular/**
       - releasing/version/VERSION
 
+permissions:
+  contents: read
+
 env:
   DOCKER_USER: kubeflownotebookswg
   IMG: kubeflownotebookswg/centraldashboard-angular
--- a/.github/workflows/centraldb_angular_frontend_test.yaml
+++ b/.github/workflows/centraldb_angular_frontend_test.yaml
@@ -4,6 +4,9 @@ on:
     paths:
       - components/centraldashboard-angular/frontend/**
 
+permissions:
+  contents: read
+
 jobs:
   frontend-format-lint-check:
     name: Check code format and lint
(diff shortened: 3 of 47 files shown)


thesuperzapper:
@pnacht thanks for this!

We should confirm that the repo is not already set to use "read" permission by default, which would make this redundant.

/cc @kimwnasptd

pnacht:
Hey @thesuperzapper, sorry for the late reply.

Looking at your Action logs, we can see the workflows are running with write-all permissions: https://github.com/kubeflow/kubeflow/actions/runs/6500263891/job/17655300160#step:1:19

While setting default read-only tokens is certainly what I'd recommend, I'd suggest also adding these explicit permissions to the workflows for added transparency to your users.

pnacht:
Hey @thesuperzapper & @kimwnasptd, let me know if this is something you'd be interested in.

I'll happily handle the conflicts if so.

thesuperzapper:
@pnacht we could possibly do the other way around (change the repo to "read only" by default, and "inclusive" specify permissions that are needed beyond read in each job).

@james-jwu @zijianjoy can make the change, but we need to make the PR give appropriate permissions to each job first l.

@pnacht if you want to check which jobs need more than "read only", we can update the PR to give "write" access, and then update the repo settings.

pnacht:
Hey @thesuperzapper, I've just pushed a new commit that adds the minimal permissions needed for all workflows.

The only ones that need any `write` permission are the `example_notebook_servers_publish.yml` and `..._TEMPLATE.yml` workflows, which need `packages: write` to push to the GitHub Container Repository.

Everything else should work fine with read-only tokens.

I've still left all workflows declaring `permissions: contents: read` at the top-level. This will be "redundant" if you modify the default token to read-only, but is still useful to let users know the workflows are adequately permissioned. That being said, let me know if you'd rather I drop the top-level read-only permissions and leave that to the repo settings.

andreyvelich (commented):
/close

PR #7286, opened 2023-09-19
Release v1.8.0-rc.1

DnPlas:
Bump release version v1.8.0-rc.0 -> v1.8.0-rc.1

cc: @kimwnasptd

PR #7292, opened 2023-09-20
fix: pvcviewer tests

TobiasGoerke:
Changes filepaths on master caused the download of CRDs to fail. Instead, we now add the CRDs statically, to prevent issues like these.

See for more info:
https://github.com/kubeflow/kubeflow/pull/7286#issuecomment-1728151521


TobiasGoerke:
/assign kimwnasptd 

XXL-label due to statically added / regenerated CRDs

kimwnasptd:
Thank you for the prompt action @TobiasGoerke!

/lgtm
/approve

PR #7032, opened 2023-03-07
Profile controller and KFAM allow unauthenticated in-cluster traffic

johnhoman:
This is similar to this kerve issue https://github.com/kserve/kserve/issues/2197#issue-1248404440. Users can be impersonated without authentication by attaching the header `kubeflow-userid: ` with a valid profile contributor name to a request going to any namespace within the mesh.

```sh
(base) jovyan@tensorflow-0:~$ curl -sS -D - http://jupyter-scipy.kubeflow-user-example-com.svc.cluster.local/notebook/kubeflow-user-example-com/jupyter-scipy/api/sessions -o /dev/null
HTTP/1.1 403 Forbidden
content-length: 19
content-type: text/plain
date: Tue, 07 Mar 2023 04:02:17 GMT
server: envoy
x-envoy-upstream-service-time: 4

(base) jovyan@tensorflow-0:~$ curl -H "kubeflow-userid: user@example.com" -sS -D - http://jupyter-scipy.kubeflow-user-example-com.svc.cluster.local/notebook/kubeflow-user-example-com/jupyter-scipy/api/sessions -o /dev/null
HTTP/1.1 200 OK
server: envoy
content-type: application/json
date: Tue, 07 Mar 2023 04:03:22 GMT
x-content-type-options: nosniff
content-security-policy: frame-ancestors 'self'; report-uri /notebook/kubeflow-user-example-com/jupyter-scipy/api/security/csp-report; default-src 'none'
access-control-allow-origin: *
etag: "97d170e1550eee4afc0af065b78cda302a97674c"
content-length: 2
x-envoy-upstream-service-time: 106
```

Adding this change resolves that
```sh
kubectl patch deployment profiles-deployment -n kubeflow --type=json -p='[{"op": "replace", "path": "/spec/template/spec/containers/1/image", "value": "profile-controller:v1.7.0-rc.1-dirty"}]'
deployment.apps/profiles-deployment patched
```

```sh
➜  ~ kubectl get authorizationpolicies/ns-owner-access-istio -n kubeflow-user-example-com -o json | jq '.spec.rules[0]'

{
  "from": [
    {
      "source": {
        "principals": [
          "cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account"
        ]
      }
    }
  ],
  "when": [
    {
      "key": "request.headers[kubeflow-userid]",
      "values": [
        "user@example.com"
      ]
    }
  ]
}
```

```sh
[...]

PR #7310, opened 2023-10-03
Make principals dynamic in Profiles Controller

kimwnasptd:
This is a follow-up to address https://github.com/kubeflow/kubeflow/pull/7032#issuecomment-1682614698

This PR will make all principals in Profile Controller configurable, so they should be changed from any distribution that deploys the components in different namespaces.

/cc @thesuperzapper

### DIFF
--- a/components/access-management/kfam/bindings.go
+++ b/components/access-management/kfam/bindings.go
@@ -81,6 +81,10 @@ func getAuthorizationPolicy(binding *Binding, userIdHeader string, userIdPrefix 
 		"ISTIO_INGRESS_GATEWAY_PRINCIPAL",
 		"cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account")
 
+	kfpUIPrincipal := GetEnvDefault(
+		"KFP_UI_PRINCIPAL",
+		"cluster.local/ns/kubeflow/sa/ml-pipeline-ui")
+
 	return istioSecurity.AuthorizationPolicy{
 		Rules: []*istioSecurity.Rule{
 			{
@@ -94,7 +98,10 @@ func getAuthorizationPolicy(binding *Binding, userIdHeader string, userIdPrefix 
 				},
 				From: []*istioSecurity.Rule_From{{
 					Source: &istioSecurity.Source{
-						Principals: []string{istioIGWPrincipal},
+						Principals: []string{
+							istioIGWPrincipal,
+							kfpUIPrincipal,
+						},
 					},
 				}},
 			},
--- a/components/access-management/kfam/bindings.go
+++ b/components/access-management/kfam/bindings.go
@@ -77,6 +77,10 @@ func getBindingName(binding *Binding) (string, error) {
 }
 
 func getAuthorizationPolicy(binding *Binding, userIdHeader string, userIdPrefix string) istioSecurity.AuthorizationPolicy {
+	istioIGWPrincipal := GetEnvDefault(
+		"ISTIO_INGRESS_GATEWAY_PRINCIPAL",
+		"cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account")
+
 	return istioSecurity.AuthorizationPolicy{
 		Rules: []*istioSecurity.Rule{
 			{
@@ -90,9 +94,7 @@ func getAuthorizationPolicy(binding *Binding, userIdHeader string, userIdPrefix 
 				},
 				From: []*istioSecurity.Rule_From{{
 					Source: &istioSecurity.Source{
-						Principals: []string{
-							"cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account",
-						},
+						Principals: []string{istioIGWPrincipal},
 					},
 				}},
 			},
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -449,20 +449,10 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 					Source: &istioSecurity.Source{
 						Principals: []string{
 							istioIGWPrincipal,
+							kfpUIPrincipal,
 						},
 					},
 				}},
-			},
-			{
-				From: []*istioSecurity.Rule_From{
-					{
-						// KFP UI needs to talk to ml-pipeline-ui-artifact pods
-						// in each profile namespace
-						Source: &istioSecurity.Source{
-							Principals: []string{kfpUIPrincipal},
-						},
-					},
-				},
 			},
 			{
 				When: []*istioSecurity.Condition{
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -417,12 +417,13 @@ func (r *ProfileReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile) istioSecurity.AuthorizationPolicy {
+	nbControllerPrincipal := GetEnvDefault(
+		"NOTEBOOK_CONTROLLER_PRINCIPAL",
+		"cluster.local/ns/kubeflow/sa/notebook-controller-service-account")
 
-	clusterDomain := "cluster.local"
-	if clusterDomainFromEnv, ok := os.LookupEnv("CLUSTER_DOMAIN"); ok {
-		clusterDomain = clusterDomainFromEnv
-	}
-	principals := fmt.Sprintf("%s/ns/kubeflow/sa/notebook-controller-service-account", clusterDomain)
+	istioIGWPrincipal := GetEnvDefault(
+		"ISTIO_INGRESS_GATEWAY_PRINCIPAL",
+		"cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account")
 
 	return istioSecurity.AuthorizationPolicy{
 		Action: istioSecurity.AuthorizationPolicy_ALLOW,
@@ -443,7 +444,7 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 				From: []*istioSecurity.Rule_From{{
 					Source: &istioSecurity.Source{
 						Principals: []string{
-							"cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account",
+							istioIGWPrincipal,
 						},
 					},
 				}},
@@ -480,7 +481,7 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 				From: []*istioSecurity.Rule_From{
 					{
 						Source: &istioSecurity.Source{
-							Principals: []string{principals},
+							Principals: []string{nbControllerPrincipal},
 						},
 					},
 				},
@@ -781,4 +782,12 @@ func (r *ProfileReconciler) readDefaultLabelsFromFile(path string) map[string]st
 		os.Exit(1)
 	}
 	return labels
+}
+
+func GetEnvDefault(variable string, defaultVal string) string {
+	envVar := os.Getenv(variable)
+	if len(envVar) == 0 {
+		return defaultVal
+	}
+	return envVar
 }
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -425,6 +425,10 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 		"ISTIO_INGRESS_GATEWAY_PRINCIPAL",
 		"cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account")
 
+	kfpUIPrincipal := GetEnvDefault(
+		"KFP_UI_PRINCIPAL",
+		"cluster.local/ns/kubeflow/sa/ml-pipeline-ui")
+
 	return istioSecurity.AuthorizationPolicy{
 		Action: istioSecurity.AuthorizationPolicy_ALLOW,
 		// Empty selector == match all workloads in namespace
@@ -448,6 +452,17 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
 						},
 					},
 				}},
+			},
+			{
+				From: []*istioSecurity.Rule_From{
+					{
+						// KFP UI needs to talk to ml-pipeline-ui-artifact pods
+						// in each profile namespace
+						Source: &istioSecurity.Source{
+							Principals: []string{kfpUIPrincipal},
+						},
+					},
+				},
 			},
 			{
 				When: []*istioSecurity.Condition{
--- a/.github/workflows/prof_controller_unit_test.yaml
+++ b/.github/workflows/prof_controller_unit_test.yaml
@@ -14,7 +14,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v3
       with:
-        go-version: '1.17'
+        go-version: '1.19'
         check-latest: true
 
     - name: Run unit tests
--- a/components/access-management/kfam/utils.go
+++ b/components/access-management/kfam/utils.go
@@ -13,3 +13,13 @@
 // limitations under the License.
 
 package kfam
+
+import "os"
+
+func GetEnvDefault(variable string, defaultVal string) string {
+	envVar := os.Getenv(variable)
+	if len(envVar) == 0 {
+		return defaultVal
+	}
+	return envVar
+}
--- a/components/profile-controller/config/manager/kustomization.yaml
+++ b/components/profile-controller/config/manager/kustomization.yaml
@@ -10,4 +10,5 @@ configMapGenerator:
   - USERID_PREFIX=
   - ISTIO_INGRESS_GATEWAY_PRINCIPAL="cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account")
   - NOTEBOOK_CONTROLLER_PRINCIPAL="cluster.local/ns/kubeflow/sa/notebook-controller-service-account")
+  - KFP_UI_PRINCIPAL="cluster.local/ns/kubeflow/sa/ml-pipeline-ui"
   name: config
--- a/components/profile-controller/config/manager/kustomization.yaml
+++ b/components/profile-controller/config/manager/kustomization.yaml
@@ -8,4 +8,6 @@ configMapGenerator:
   - WORKLOAD_IDENTITY=
   - USERID_HEADER="kubeflow-userid"
   - USERID_PREFIX=
+  - ISTIO_INGRESS_GATEWAY_PRINCIPAL="cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account")
+  - NOTEBOOK_CONTROLLER_PRINCIPAL="cluster.local/ns/kubeflow/sa/notebook-controller-service-account")
   name: config


thesuperzapper:
@kimwnasptd we also need to add a new principal `KFP_UI_PRINCIPAL` (or `PIPELINE_UI_PRINCIPAL`) with the default value of `cluster.local/ns/kubeflow/sa/ml-pipeline-ui`

This is required for the `ml-pipeline-ui` in the `kubeflow` namespace to redirect the user's queries to `ml-pipeline-ui-artifact` in the profile namespaces.

thesuperzapper:
You also need to do the KFAM ones (with all the same changes): 

https://github.com/kubeflow/kubeflow/blob/master/components/access-management/kfam/bindings.go#L79-L101

As these were changed in #7032 also.

kimwnasptd:
> @kimwnasptd we also need to add a new principal KFP_UI_PRINCIPAL (or PIPELINE_UI_PRINCIPAL) with the default value of cluster.local/ns/kubeflow/sa/ml-pipeline-ui

> This is required for the ml-pipeline-ui in the kubeflow namespace to redirect the user's queries to ml-pipeline-ui-artifact in the profile namespaces.

@thesuperzapper could you add some more context for this?

kimwnasptd:
> You also need to do the KFAM ones (with all the same changes):

@thesuperzapper why? Isn't the `Create` function of KFAM used when we want to create "bindings" for a new contributor in the namespace? In this case we only care about allowing requests from IGW to the Pods of the namespace that have that user header.

The other rules (like allowing requests from the the Notebook controller) should already exist from the AuthorizationPolicy created by the Profile Controller

thesuperzapper:
> > You also need to do the KFAM ones (with all the same changes):
> 
> @thesuperzapper why (I'm having the KFP change in mind)? Isn't the `Create` function of KFAM used when we want to create "bindings" for a new contributor in the namespace? In this case we only care about allowing requests from IGW to the Pods of the namespace that have that user header.
> 
> The other rules (like allowing requests from the the Notebook controller) should already exist from the AuthorizationPolicy created by the Profile Controller.
>
> ACK on making the principal configurable in KFAM for the Istio ingressgateway though. Having in mind above the rest of the changes, in case you meant we would need to bring more changes in KFAM

@kimwnasptd I was referring to the KFP principal I raised in https://github.com/kubeflow/kubeflow/pull/7310#issuecomment-1745357377, which needs to be created for BOTH non-owners (created by KFAM) and the owners (created by profile controller).

This is required for KFP to work correctly (specifically for the "view output artifacts" feature of the pipeline's UI).

This is because, internally, the `ml-pipeline-ui` pods (in the `kubeflow` namespace) proxy the user "artifact view" requests to the `ml-pipeline-ui-artifact` pods (in each profile namespace), directly using the `Service/ml-pipeline-ui-artifact` (not via the istio gateway).

This is what allows each namespace to have different levels of access to the S3 bucket (and probably a few other things I am missing too).

kimwnasptd:
Thanks for the explanation! I'll modify the code and try to play around with this to make sure I'm fully in sync.

In the mean time though, @zijianjoy could you give us your feedback as well? The context is:
1. We had made it more strict, in AuthorizationPolicy of each Profile, to allow requests only from Istio's ingressgateway to user namespace
2. Looks like there's a communication from KFP UI pod to the proxy pods in each user namespace

Is there something else we need to consider here? Also if you'd have any pointers for this architecture they would be more than welcome 😄

kimwnasptd:
So indeed, after playing around with the latest 1.8 RC I'm getting 403 errors in the `Input/Output` tabs of a KFP Run when it 
1. tries to preview the contents of the artifacts 
2. download the artifacts

After applying the following AuthorizationPolicy, in the user namespace, it worked
```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: kfp
spec:
  rules:
  - from:
    - source:
        principals:
        - cluster.local/ns/kubeflow/sa/ml-pipeline-ui
```

But this actually raises a question. Should the profile controller be responsible for this or the existing kfp-profiles-controller (Metacontroller) that already creates the AuthorizationPolicy for the visualization server?
https://github.com/kubeflow/pipelines/blob/master/manifests/kustomize/base/installs/multi-user/pipelines-profile-controller/sync.py#L226-L246

cc @zijianjoy

thesuperzapper:
@kimwnasptd my thoughts are that I don't any straightforward way to use the `kubeflow-pipelines-profile-controller` to do this because only KFAM knows when its API to "add/remove" collaborators is called.

And for the profile owner bindings, we already the profile-controller creating the Notebooks UI AuthorizaitonPolicy (for the last activity field), so I don't see a harm with adding a pipeline-specific one either.

Finally, I want to reduce the size of the  `kubeflow-pipelines-profile-controller`, not increase it.

kimwnasptd:
> Finally, I want to reduce the size of the kubeflow-pipelines-profile-controller, not increase it.

@thesuperzapper this is not on you but on the owners of the component 😄 

Are you sure the above AuthorizationPolicy you posted is working for you? I applied it but unless I remove the `when` part it still doesn't work and returns `RBAC: access denied` errors
```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: user-user1-example-com-clusterrole-edit
  namespace: PROFILE_NAME
  annotations:
    role: edit
    user: "user1@example.com"
spec:                                                                                                                                                                      
  rules:                                                                                                                                                                   
  - from:                                                                                                                                                                  
    - source:                                                                                                                                                              
        principals:                                                                                                                                                        
        - cluster.local/ns/deploykf-istio-gateway/sa/deploykf-gateway                                                                                                      
        - cluster.local/ns/kubeflow/sa/ml-pipeline-ui                                                                                                                      
    when:                                                                                                                                                                  
    - key: request.headers[kubeflow-userid]                                                                                                                                
      values:                                                                                                                                                              
      - user1@example.com
 ```

If that's the case this is actually yet one more argument that this should be in the `pipelines-profiles-controller`. Let me also actually do a cc @chensun as well

**Comment by thesuperzapper** [comment]
@kimwnasptd lol, I made a typo on the ingress gateway principal, it should be `cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account`.


```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: user-user1-example-com-clusterrole-edit
  namespace: PROFILE_NAME
  annotations:
    role: edit
    user: "user1@example.com"
spec:                                                                                                                                                                      
  rules:                                                                                                                                                                   
  - from:                                                                                                                                                                  
    - source:                                                                                                                                                              
        principals:                                                                                                                                                        
        - cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account                                                                                                
        - cluster.local/ns/kubeflow/sa/ml-pipeline-ui                                                                                                                      
    when:                                                                                                                                                                  
    - key: request.headers[kubeflow-userid]                                                                                                                                
      values:                                                                                                                                                              
      - user1@example.com
```

kimwnasptd:
> @kimwnasptd lol, I made a typo on the ingress gateway principal, it should be cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account.

@thesuperzapper the issue is not in the Istio IGW principal but on the rule for KFP. As I mentioned, in my vanilla installation:

❌ **FAILS: Still getting RBAC from Istio**
```
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: user-user1-example-com-clusterrole-edit
spec:                                                                                                                                                                      
  rules:                                                                                                                                                                   
  - from:                                                                                                                                                                  
    - source:                                                                                                                                                              
        principals:                                                                                                                                                                                                                                                   
        - cluster.local/ns/kubeflow/sa/ml-pipeline-ui                                                                                                                      
```

✅ **SUCCEEDS**
```
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: user-user1-example-com-clusterrole-edit
spec:                                                                                                                                                                      
  rules:                                                                                                                                                                   
  - from:                                                                                                                                                                  
    - source:                                                                                                                                                              
        principals:                                                                                                                                                                                                                                                      
        - cluster.local/ns/kubeflow/sa/ml-pipeline-ui                                                                                                                      
```

The PR currently has the above change that allows all requests from the `ml-pipeline-ui` so let's move on with this. This was either way the functionality in 1.7. We don't have to implement a new feature now while delaying the release.

The PR currently has the changes that I tested in my cluster and get KFP UI to talk to the per-profile Pods

components/profile-controller/controllers/profile_controller.go
```diff
@@ -443,11 +448,22 @@ func (r *ProfileReconciler) getAuthorizationPolicy(profileIns *profilev1.Profile
...
+				From: []*istioSecurity.Rule_From{
+					{
+						// KFP UI needs to talk to ml-pipeline-ui-artifact pods
+						// in each profile namespace
+						Source: &istioSecurity.Source{
+							Principals: []string{kfpUIPrincipal},
+						},
+					},
+				},
+			},
```
thesuperzapper, line 466: This rule should also include the "when" condition, to prevent the pipeline UI from being able to proxy traffic without a userid-header.

## Option 1

The best option is to simply add another "from" principal  in the first rule, and remove this rule.

```go
			{
				When: []*istioSecurity.Condition{
					{
						Key: fmt.Sprintf("request.headers[%v]", r.UserIdHeader),
						Values: []string{
							r.UserIdPrefix + profileIns.Spec.Owner.Name,
						},
					},
				},
				From: []*istioSecurity.Rule_From{{
					Source: &istioSecurity.Source{
						Principals: []string{
							// end-users access via the ingress gateway
							istioIGWPrincipal,
							// kfp proxies user requests to artifact pods in profile namespaces
							kfpUIPrincipal,
						},
					},
				}},
			},
```

## Option 2

We could keep it as a separate rule, and add the "when" condition, but this is is less efficient for Envoy, and needless because it has the same effect.

```go
			{
				// kfp proxies user requests to artifact pods in profile namespaces
				When: []*istioSecurity.Condition{
					{
						Key: fmt.Sprintf("request.headers[%v]", r.UserIdHeader),
						Values: []string{
							r.UserIdPrefix + profileIns.Spec.Owner.Name,
						},
					},
				},
				From: []*istioSecurity.Rule_From{
					{
						Source: &istioSecurity.Source{
							Principals: []string{kfpUIPrincipal},
						},
					},
				},
			},
```

components/access-management/kfam/bindings.go
```diff
@@ -90,9 +94,7 @@ func getAuthorizationPolicy(binding *Binding, userIdHeader string, userIdPrefix
 				},
 				From: []*istioSecurity.Rule_From{{
 					Source: &istioSecurity.Source{
-						Principals: []string{
-							"cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account",
-						},
+						Principals: []string{istioIGWPrincipal},
```
thesuperzapper, line 97: You also need to add the `kfpUIPrincipal` here to allow the KFP UI to proxy requests for non-owners.

So if we define `kfpUIPrincipal` like in the profile controller, you can simply extend this list:

```go
				From: []*istioSecurity.Rule_From{{
					Source: &istioSecurity.Source{
						Principals: []string{
							// end-users access via the ingress gateway
							istioIGWPrincipal,
							// kfp proxies user requests to artifact pods in profile namespaces
							kfpUIPrincipal,
						},
					},
				}},
```

thesuperzapper (changes requested):
@kimwnasptd its important for security that the kfp-ui is not able to proxy arbitrary requests (without userid-header), so let's just extend the from principals as I have proposed here.

NOTE: we don't even need to add a new rule this way, just extend the "from" principal list of the exiting one which has the "when" condition for the userid header.

thesuperzapper:
@kimwnasptd the problem you were likley having was that KFAM does NOT reconcile its `AuthorizationPolicies` or `RoleBindings` for profile members, it only does it when the member is first added.

This is actually something we need to address before 1.8!

We either need to tell users to remove and re-add ALL members of profiles for this fix to take effect (even the initial one which prevents impersonation by restricting the source to the ingress gateway), or add a reconciliation loop to the profile controller (or possibly KFAM) for the `AuthorizationPolicies`.

The only "source of truth" around who should be a member of a profile, is the presence of `RoleBinding` in the profile namespaces like `user-user1-example-com-clusterrole-view`.

To identify the `RoleBindings` that actually correspond to members of profiles, we can look for roles that have a name starting with `user-`, and have a `roleRef` with one of the Kubeflow `ClusterRoles`.

https://github.com/kubeflow/kubeflow/blob/09917e6e29fc4af8479a417f757bde99f99e8b04/components/access-management/kfam/bindings.go#L40-L42

KFAM already does something similar with its "read bindings" API call, which is used by the dashboard (but not for reconciliation):

https://github.com/kubeflow/kubeflow/blob/09917e6e29fc4af8479a417f757bde99f99e8b04/components/access-management/kfam/api_default.go#L210-L256

kimwnasptd:
@thesuperzapper can you confirm that you have applied the following AuthorizationPolicy, in your user's profile namespace, and you don't get `RBAC: access denied` when trying to download the artifact of a pipeline step from the UI? 

```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  name: user-user1-example-com-clusterrole-edit
spec:                                                                                                                                                                      
  rules:                                                                                                                                                                   
  - from:                                                                                                                                                                  
    - source:                                                                                                                                                              
        principals:                                                                                                                                                                                                                                                 
        - cluster.local/ns/kubeflow/sa/ml-pipeline-ui                                                                                                                      
    when:                                                                                                                                                                  
    - key: request.headers[kubeflow-userid]                                                                                                                                
      values:                                                                                                                                                              
      - user1@example.com                                                                      
 ```

Because for me even when I manually apply such a policy in the user namespace I'm still getting RBAC errors of Istio. If you don't get errors then I'd suggest you to sent a PR for this, since I don't want to send a PR for something that I can't get make it work. And I focus this PR just on the initial scope of configuring the existing principals for the IGW and Notebook Controller

**Comment by thesuperzapper** [comment]
@kimwnasptd I can confirm with the manifests from `1.8.0-rc.1` that I get no RBAC errors on the KFP artifacts, if the following changes are made to these AuthorizationPolicies.

Note I am using two dex static-password users:
- `user@example.com`: the owner of  profile named `kubeflow-user-example-com`
-  `user1@example.com`: a member of profile named `kubeflow-user-example-com`


## AuthorizationPolicy for Owner

NOTE 1: to manually apply this, you need to scale the `Deployment/profiles-deployment` to 0 replicas, as changes will be immediately reverted otherwise. 

NOTE 2: You scale to 0 AFTER you are already on the correct pipelines page, because the profile controller being down breaks the ability to navigate on the central dashboard.

```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  annotations:
    role: admin
    user: user@example.com
  name: ns-owner-access-istio
spec:
  rules:
    - from:
        - source:
            principals:
              - cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account
              ## \/ added this part \/ ##
              - cluster.local/ns/kubeflow/sa/ml-pipeline-ui
      when:
        - key: request.headers[kubeflow-userid]
          values:
            - user@example.com
    - when:
        - key: source.namespace
          values:
            - kubeflow-user-example-com
    - to:
        - operation:
            paths:
              - /healthz
              - /metrics
              - /wait-for-drain
    - from:
        - source:
            principals:
              - cluster.local/ns/kubeflow/sa/notebook-controller-service-account
      to:
        - operation:
            methods:
              - GET
            paths:
              - '*/api/kernels'
```

## AuthorizationPolicy for Member

```yaml
apiVersion: security.istio.io/v1
kind: AuthorizationPolicy
metadata:
  annotations:
    role: edit
    user: user1@example.com
  name: user-user1-example-com-clusterrole-edit
spec:
  rules:
    - from:
        - source:
            principals:
              - cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account
              ## \/ added this part \/ ##
              - cluster.local/ns/kubeflow/sa/ml-pipeline-ui
      when:
        - key: request.headers[kubeflow-userid]
          values:
            - user1@example.com
```

kimwnasptd:
omg, I found the problem with my testing. The file I was locally applying had `requests.headers` and not `request.headers`... It works indeed as expected with this.

Sending the patch

kimwnasptd:
@thesuperzapper I sent the fix. I confirm it works with the Profile controller as well. Apologies for this back and forth, caused by me using different yaml locally from the one above.

Once the tests pass we should be good to merge

thesuperzapper:
@kimwnasptd We MUST make another PR before releasing 1.8, because right now the security fix that was intended by https://github.com/kubeflow/kubeflow/pull/7032, will not be applied to __existing profile members__, only new ones.

This is because as I was saying in https://github.com/kubeflow/kubeflow/pull/7310#issuecomment-1758362940, while the profile controller will reconcile changes, __KFAM will NOT__, so all of our updates will only apply to NEW members of profiles.

This means we need to __add a reconciliation loop__ to the resources created by KFAM.

## Proposal for Reconciliation of Bindings

After thinking about it, the easiest way to do this is to move the reconciliation to the profile controller.

This is actually easier than it sounds:

1. We can make KFAM only responsible for creating/deleting the `RoleBindings` and not any other resources:
      - This means removing the other parts of its "[create role binding](https://github.com/kubeflow/kubeflow/blob/09917e6e29fc4af8479a417f757bde99f99e8b04/components/access-management/kfam/bindings.go#L103-L146)" function that create `AuthorizationPolicies`
      - This also allows us to keep KFAM, so nothing else breaks (like central-dashboard)
3. We can extend the profile-controller's [reconciliation loop](https://github.com/kubeflow/kubeflow/blob/09917e6e29fc4af8479a417f757bde99f99e8b04/components/profile-controller/controllers/profile_controller.go#L105-L334):
      - It can look for `RoleBindings` in profile namespaces that start with the name `user-` and have appropriate annotations.
      - If it finds one, it can create the `AuthorizationPolicies` for that user, with an ownerReference pointing to the `RoleBinding` (so if KFAM deletes it, the binding gets cleaned up automatically).

This way the presence of a `RoleBinding` in a profile namespace "tells" the profile-controller what it needs to create:

```yaml
kind: RoleBinding
metadata:
  annotations:
    role: edit
    user: user1@example.com
  name: user-user1-example-com-clusterrole-edit
roleRef:
  kind: ClusterRole
  name: kubeflow-edit
subjects:
  - apiGroup: rbac.authorization.k8s.io
    kind: User
    name: user1@example.com
```

As you can see, all the info the profile-controller needs is found in the `user` and `role` annotations.

Finally, we should allow seamless "adoption" of old `AuthorizationPolicies` by the profile-controller, so users don't need to do anything when upgrading to 1.8.

kimwnasptd:
/hold

@thesuperzapper tbh I don't feel comfortable with either of those. They both seem like too big of changes to introduce in the last minute, when the release is already hugely delayed. 

My proposal is what I had initially described:
1. Allow all requests that come from KFP UI's ServiceAccount to user namespace (in AP created for "owner")
2. This won't require any changes in KFAM 

For the KFP functionality it feels like we are trying to patch a security whole the KFP folks didn't think e2e. And all of this in the last minute with adding huge changes while delaying the release. I don't have the cycles to either create the follow up PR, not to properly review it. 

So I suggest we keep the work of fixing KFP issues as a next effort either in a KF patch release or in 1.8

thesuperzapper:
@kimwnasptd This PR is correct as it is, I think we should merge it.

The issue I am raising in https://github.com/kubeflow/kubeflow/pull/7310#issuecomment-1758593558 is not really related to KFP or this PR.

It's that even though we have updated KFAM to create secure AuthorizationPolicies (which ensure traffic is only coming from the gateway), any existing AuthorizationPolicies won't be updated.

So we either need to tell users to remove and re-add every member of each profile (so KFAM makes the new AuthorizationPolicies), or do the small change I am proposing to move the management of `AuthorizationPolicies` to the profile-controller.

I really don't think it's as complex as you are thinking, it's a small amount of code being removed from KFAM, and a small amount of code being added to the profile-controller.

thesuperzapper:
@kimwnasptd and I had a long discussion, and we think this is safe to merge for now, and rather than doing what I proposed in https://github.com/kubeflow/kubeflow/pull/7310#issuecomment-1758619823, we will tell users who are __upgrading__, that to get the security benefits of https://github.com/kubeflow/kubeflow/pull/7032, they must EITHER:

- patch the existing AuthorizationPolicies for __colaborators__ in each profile to have the new "from principal" which only allows traffic from the istio-gateway and ml-pipeline-ui
- remove and re-add every profile collaborator using the central-dashboard UI

/lgtm
/approve

PR #7321, opened 2023-10-06
fix: set default GPU vendors list

### LINKED ISSUE #7273: GPU Vendor not available error
/kind bug

image
image
image

Recently, I have built a Kubeflow system with the GPU operator on the Rancher RKE1 cluster. 

Before building the current system, I have also used Kubeflow system with the GPU device plugin on the same cluster, and had no problem on using the Nvidia GPUs. 

However, now I cannot use the GPUs on the Kubeflow since I cannot select any GPU vendors. To be more precise, When I click on the GPU Vendor on creating notebooks, it does not show Nvidia/AMD option (no action at all) and then just keep saying "You must also specify the GPU Vendor for the assigned GPUs".

It is for sure that the GPU operator is working well since I can check the nvidia gpu labels on the nodes, and also I can make a pod using the nvidia docker image that outputs the nvidia-smi result successfully.

How can I solve the problem?

I don't know whether these affected the system or not (I also applied below settings on the GPU device plugin applied cluster but was able to use the GPUs without any problem)

1. Since I intended to use the system with the Nginx Proxy Manager, I changed APP_SECURE_COOKIES to false in web app and volume app
2. To let the LDAP users to have their own namespaces, I changed registration flow parameter in centraldashboard to true
3. Added LDAP options on the Dex configuration

**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): dev_local
- kfctl version: (use `kfctl version`): Used kustomize instead.
- Kubernetes platform: (e.g. `minikube`): Rancher RKE1
- Kubernetes version: (use `kubectl version`): 1.25.12
- OS (e.g. from `/etc/os-release`): Lubuntu 22.04 LTS

DnPlas:
The change introduced by kubeflow/kubeflow#6736 removed the default GPU vendors list, which causes an issue when trying to select a vendor from the dropdown menu if the vendors list is not configured.
This commit can be reverted if proper documentation is provided for users/distributions to configure the dropdown menu.

Fixes #7273 in v1.8-branch

cc: @kimwnasptd @thesuperzapper

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
+++ b/components/crud-web-apps/jupyter/backend/apps/common/yaml/spawner_ui_config.yaml
@@ -129,12 +129,11 @@ spawnerFormDefaults:
       # the list of available vendors in the dropdown
       #  `limitsKey` - what will be set as the actual limit
       #  `uiName` - what will be displayed in the dropdown UI
-      vendors: []
-      #vendors:
-      #  - limitsKey: "nvidia.com/gpu"
-      #    uiName: "NVIDIA"
-      #  - limitsKey: "amd.com/gpu"
-      #    uiName: "AMD"
+      vendors:
+        - limitsKey: "nvidia.com/gpu"
+          uiName: "NVIDIA"
+        - limitsKey: "amd.com/gpu"
+          uiName: "AMD"
 
       # the default value of the limit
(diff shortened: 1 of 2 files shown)


kimwnasptd:
Thanks @DnPlas! When you'll have the time can you also send this to master branch?

/lgtm
/approve

Issue #6708, opened 2022-11-03
performance issues with admission webhook


deepk2u:
/kind bug

**What steps did you take and what happened:**
We are using 1.6.1 and the admission webhook is taking longer time to process requests compared to 1.5.0
earlier it was taking <1 sec now it's taking >2 sec.

log example:

with 1.6.1:
```I1103 16:45:06.611647   1 main.go:449] Entering mutatePods in mutating webhook
I1103 16:45:06.612962   1 main.go:475] Looking at pod annotations, found: map[kubectl.kubernetes.io/default-container:main ......]
I1103 16:45:07.663781   1 request.go:665] Waited for 1.037929138s due to client-side throttling, not priority and fairness, request: GET:[https://172.20.0.1:443/apis/rbac.istio.io/v1alpha1?timeout=32s](https://172.20.0.1/apis/rbac.istio.io/v1alpha1?timeout=32s)
I1103 16:45:08.873106   1 main.go:495] fetched 5 poddefault(s) in namespace ....```

with 1.5.0:
```I1103 18:08:07.200335   1 main.go:450] Entering mutatePods in mutating webhook
I1103 18:08:07.201200   1 main.go:476] Looking at pod annotations, found: map[kubectl.kubernetes.io/default-container:main ......]
I1103 18:08:07.367811   1 main.go:496] fetched 5 poddefault(s) in namespace .....```


looks like with new webhook version is making some unnecessary API calls to k8s API server.
```request.go:665] Waited for 1.037929138s due to client-side throttling, not priority and fairness, request: GET:[https://172.20.0.1:443/apis/rbac.istio.io/v1alpha1?timeout=32s](https://172.20.0.1/apis/rbac.istio.io/v1alpha1?timeout=32s)```

related issue: https://github.com/kubernetes/client-go/issues/1029


**What did you expect to happen:**


**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]


**Environment:**

- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 1.6.1
- kfctl version: (use `kfctl version`):
- Kubernetes platform: (e.g. `minikube`)
[...]

PR #7152, opened 2023-05-28
notebook-controller: Update dependency versions

kimwnasptd:
Update the dependency versions to reduce the reported CVEs.

Trivy reports the following:
```
Total: 21 (UNKNOWN: 0, LOW: 15, MEDIUM: 6, HIGH: 0, CRITICAL: 0)
```

Note that without updating the base image I was getting:
```
Total: 17 (UNKNOWN: 0, LOW: 11, MEDIUM: 4, HIGH: 2, CRITICAL: 0)
```

Some notes for the upgrade:
1. Had to manually update `ginkgo` package to v2 https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.14.0
2. Update the code for `builder.Builder.Watches` https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.15.0

PR #7322, opened 2023-10-07
Bump the controller-runtime version to 0.16.2

tzstoyanov:
Fixes issue  #6708
There are a lot of improvements, optimizations and bug fixes in the latest version, related to the controller-runtime client which is used in admission-webhook. A new cache logic is introduced, which reduces the amount of get requests used by the controller-runtime client. That avoids hitting the client-side throttling.


kimwnasptd:
Thanks for the PR @tzstoyanov! I can take a stab at it. Also, if you have the cycles, could you take a stab at mine for the notebooks controller? https://github.com/kubeflow/kubeflow/pull/7152

The changes should be quite similar, and I'm using that as my reference as I review this PR as well

/assign @kimwnasptd

kimwnasptd:
@discordianfish just to verify since this got merged. The tests weren't triggered because it was @tzstoyanov's first PR in this repo.

Did you also run the code or built and tested the image?

PR #7325, opened 2023-10-11
Add extra namespace labels option

tarat44:
This PR adds the ability to configure an additional file of labels for the profile controller to add to profiles by configuring the file in the env var `EXTRA_NAMESPACE_LABELS_PATH`. We need this additional file so that we can add pod security standards as part of @juliusvonkohout 's plan (https://github.com/kubeflow/manifests/blob/master/proposals/20200913-rootlessKubeflow.md) to increase security. We are currently unable to add to the existing configurable namespace labels file in a patch because Kustomize does not allow editing data in a configmap key value (only adding a new key).

This PR adds an additional fsnotify watcher instance for the new file and proper shutdown. Since we will have two watcher instances, I thought it was important to ensure proper resource clean up upon termination. I also added a fix to allow the file watcher code to work if in leader elect mode on a platform that follows symlinks. If these two additional updates seem like too much increasing scope for this PR, I can separate them into different PRs to submit later.

### DIFF
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -380,6 +380,7 @@ func (r *ProfileReconciler) ShutdownFileWatchers() {
 		// Exit if there are no file watchers configured
 		return
 	}
+	defer close(r.fileWatcherStopped)
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 	defer cancel()
 	for {
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -374,13 +374,13 @@ func (r *ProfileReconciler) mapEventToRequest(o client.Object) []reconcile.Reque
 
 // Shutdown all file watcher instances
 func (r *ProfileReconciler) ShutdownFileWatchers() {
+	defer close(r.fileWatcherStopped)
 	r.cancelFileWatchers()
 	stoppedCount := 0
 	if r.fileWatcherCount == 0 {
 		// Exit if there are no file watchers configured
 		return
 	}
-	defer close(r.fileWatcherStopped)
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 	defer cancel()
 	for {
@@ -419,7 +419,7 @@ func (r *ProfileReconciler) SetupFileWatchers(events chan event.GenericEvent, ct
 			return errors.Wrapf(err, "Failed to watch file %s", namespaceLabelsPath)
 		}
 
-		go func(watcher *fsnotify.Watcher, reconcileEvents chan event.GenericEvent, ctx context.Context, namespaceLabelsPath string) {
+		go func(watcher *fsnotify.Watcher, reconcileEvents chan event.GenericEvent, ctx context.Context, namespaceLabelsPath string, mgr ctrl.Manager) {
 			elected := false
 			var mu sync.Mutex
 			go func() {
@@ -466,7 +466,7 @@ func (r *ProfileReconciler) SetupFileWatchers(events chan event.GenericEvent, ct
 					return
 				}
 			}
-		}(watcher, events, ctx, namespaceLabelsPath)
+		}(watcher, events, ctx, namespaceLabelsPath, mgr)
 		r.fileWatcherCount = r.fileWatcherCount + 1
 	}
 	return nil
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"os"
 	"reflect"
+	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff"
@@ -417,12 +418,29 @@ func (r *ProfileReconciler) SetupFileWatchers(events chan event.GenericEvent, ct
 			return errors.Wrapf(err, "Failed to watch file %s", namespaceLabelsPath)
 		}
 
-		go func(watcher *fsnotify.Watcher, reconcileEvents chan event.GenericEvent, ctx context.Context, namespaceLabelsPath string, mgr ctrl.Manager) {
+		go func(watcher *fsnotify.Watcher, reconcileEvents chan event.GenericEvent, ctx context.Context, namespaceLabelsPath string) {
 			elected := false
+			var mu sync.Mutex
+			go func() {
+				// Convey if the process is the leader
+				for {
+					select {
+					case <-mgr.Elected():
+						mu.Lock()
+						elected = true
+						mu.Unlock()
+						return
+					case <-ctx.Done():
+						return
+					}
+				}
+			}()
 			for {
 				select {
 				case fsEvent := <-watcher.Events:
+					mu.Lock()
 					if fsEvent.Op != fsnotify.Remove && fsEvent.Op != fsnotify.Write {
+						mu.Unlock()
 						break
 					}
 					// ConfigMaps work with symlinks. See:
@@ -437,6 +455,7 @@ func (r *ProfileReconciler) SetupFileWatchers(events chan event.GenericEvent, ct
 						// from updating symlinks
 						reconcileEvents <- event.GenericEvent{}
 					}
+					mu.Unlock()
 				case err := <-watcher.Errors:
 					r.Log.Error(err, "Error while watching config file", "path",
 						namespaceLabelsPath)
@@ -444,11 +463,9 @@ func (r *ProfileReconciler) SetupFileWatchers(events chan event.GenericEvent, ct
 					watcher.Close()
 					r.fileWatcherStopped <- 1
 					return
-				case <-mgr.Elected():
-					elected = true
 				}
 			}
-		}(watcher, events, ctx, namespaceLabelsPath, mgr)
+		}(watcher, events, ctx, namespaceLabelsPath)
 		r.fileWatcherCount = r.fileWatcherCount + 1
 	}
 	return nil
--- a/components/profile-controller/controllers/profile_controller.go
+++ b/components/profile-controller/controllers/profile_controller.go
@@ -85,12 +85,15 @@ type Plugin interface {
 // ProfileReconciler reconciles a Profile object
 type ProfileReconciler struct {
 	client.Client
-	Scheme                     *runtime.Scheme
-	Log                        logr.Logger
-	UserIdHeader               string
-	UserIdPrefix               string
-	WorkloadIdentity           string
-	DefaultNamespaceLabelsPath string
+	Scheme                      *runtime.Scheme
+	Log                         logr.Logger
+	UserIdHeader                string
+	UserIdPrefix                string
+	WorkloadIdentity            string
+	DefaultNamespaceLabelsPaths []string
+	cancelFileWatchers          context.CancelFunc
+	fileWatcherCount            int
+	fileWatcherStopped          chan int
 }
 
 // +kubebuilder:rbac:groups=core,resources=namespaces,verbs="*"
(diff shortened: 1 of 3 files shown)


components/profile-controller/controllers/profile_controller.go
```diff
@@ -365,39 +372,109 @@ func (r *ProfileReconciler) mapEventToRequest(o client.Object) []reconcile.Reque
...
+						mu.Unlock()
+						break
+					}
+					// ConfigMaps work with symlinks. See:
+					// https://martensson.io/go-fsnotify-and-kubernetes-configmaps/
+					if fsEvent.Op == fsnotify.Remove {
+						watcher.Remove(fsEvent.Name)
+						watcher.Add(namespaceLabelsPath)
+					}
+					if elected {
```
tarat44, line 453: This fix is a new addition. I am happy to separate it into a separate PR if that seems more fitting

components/profile-controller/controllers/profile_controller.go
```diff
@@ -104,12 +108,15 @@ type ProfileReconciler struct {
 // Automatically generate RBAC rules to allow the Controller to read and write Deployments
 func (r *ProfileReconciler) Reconcile(ctx context.Context, request ctrl.Request) (ctrl.Result, error) {
 	logger := r.Log.WithValues("profile", request.NamespacedName)
-	defaultKubeflowNamespaceLabels := r.readDefaultLabelsFromFile(r.DefaultNamespaceLabelsPath)
-
+	defaultKubeflowNamespaceLabels, err := r.readDefaultLabelsFromFiles(r.DefaultNamespaceLabelsPaths)
+	if err != nil {
+		r.ShutdownFileWatchers()
+		os.Exit(1)
```
tarat44, line 114: I kept the exit if loading the labels file fails, but it might be worth trying to come up with a new solution later since this could cause issues if we later configure multiple concurrent reconciles.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -766,19 +847,24 @@ func setNamespaceLabels(ns *corev1.Namespace, newLabels map[string]string) {
...
-
-	labels := map[string]string{}
-	err = yaml.Unmarshal(dat, &labels)
-	if err != nil {
-		logger.Error(err, "Unable to parse default namespace labels.")
-		os.Exit(1)
+func (r *ProfileReconciler) readDefaultLabelsFromFiles(labelPaths []string) (map[string]string, error) {
+	allLabels := map[string]string{}
+	for _, path := range labelPaths {
+		logger := r.Log.WithName("read-config-file").WithValues("path", path)
```
tarat44, line 853: The labels in the extra labels file will override the original labels if they have the same keys. Let me know if you think this behavior should be different.

tarat44:
Please hold off reviewing. I want to run one more test.

tarat44:
This should be ready for review now

juliusvonkohout:
Maybe @tzstoyanov can chime in here as well.

 I am in general fine with what you have done here for the time being and we can later on think about replacing existing labels in general. Let's merge this PR here and discuss further steps in the next WG meeting.

components/profile-controller/controllers/profile_controller.go
```diff
@@ -766,19 +847,24 @@ func setNamespaceLabels(ns *corev1.Namespace, newLabels map[string]string) {
...
+		dat, err := ioutil.ReadFile(path)
+		if err != nil {
+			logger.Error(err, "unable to read default namespace labels file")
```
tarat44, line 856: Tweaked these logger messages a little bit so it is clear they are specific to the file in the path key

kimwnasptd:
Taking a deeper look, thank you for the PR @tarat44!

The changes with the watches, depending on the leader election seem super interesting. Nice finding!

Could you also explain one more time the use-case for having the extra ConfigMap for the PodSecurityStandard labels that are needed for a namespace?

tarat44:
> Taking a deeper look, thank you for the PR @tarat44!
> 
> The changes with the watches, depending on the leader election seem super interesting. Nice finding!
> 
> Could you also explain one more time the use-case for having the extra ConfigMap for the PodSecurityStandard labels that are needed for a namespace?

It enables defining an additional configmap with labels outside the standard ones that can be configured from separate Kustomize patches/components.

tarat44:
I see the failing unit test on the CI- I will fix it when I have time this week.

tarat44:
I could not reproduce the test error, but my theory is that it could be related to multiple events being sent when a file is updated so I changed the test functionality to account for this scenario and hope it will fix the issue. I also changed how file watchers closing is being reported to the shutdown function so that the shutdown function can be called more than once and the tests can shutdown file watchers in a defer statement. The new functionality should also be less fragile and fixes an issue in the original strategy that could have lead to a panic if the shutdown timed out, but the channel for the watcher goroutine to report to was closed.

tarat44:
Because I am no longer closing the events channel to prevent a rare case where it could in theory produce a panic if the watcher shutdown times out, I moved the events channel declaration back into the controller set up function. I apologize for the last minute changes. This PR should be good to review again.

juliusvonkohout:
I think we should enable the same functionality for annotations as we do for labels.

DnPlas:
I agree with @juliusvonkohout. To give more context on the need for annotations:

The `kubeflow-profiles` controller already possess the ability to mutate any workload's Pod in a user namespace with labels, but there might be cases where we'd want annotations, and as is right now, we depend on an extra webhook or projects like [`kyverno`](https://kyverno.io/). The ability to annotate all Pods in a user namespace will have a big impact when users enable the Istio CNI plugin in their deployments, as the plugin has [limitations](https://istio.io/latest/docs/setup/additional-setup/cni/#compatibility-with-application-init-containers) that can be solved by simply annotating the Pods.
I think this feature can be worked in a different PR, but I wanted to make clear why we'd want that.

juliusvonkohout:
@DnPlas it can annotate namespace labels. I am not sure about pods. Can you try setting the istio annotations on the namespace level? It works for some.

kimwnasptd:
@zijianjoy could you also trigger the GH Action tests for this PR?

kimwnasptd:
@tarat44 could you help me test this PR?

If I locally try to do `make run` on a cluster that has 1.8 in it (with the profile controller pod scaled down to zero) I see a ton of logs like this all the time
```
1.7063747571463747e+09  INFO    controllers.Profile     Start to Reconcile.     {"profile": "/kubeflow-user-example-com", "namespace": "", "name": "kubeflow-user-example-com"}
1.7063747571464114e+09  INFO    controllers.Profile     List of labels to be added to namespace      {"profile": "/kubeflow-user-example-com", "labels": {"app.kubernetes.io/part-of":"kubeflow-profile","istio-injection":"enabled","katib.kubeflow.org/metrics-collector-injection":"enabled","pipelines.kubeflow.org/enabled":"true","serving.kubeflow.org/inferenceservice":"enabled"}}
1.7063747571464264e+09  INFO    controllers.Profile     List of labels to be added to found namespace        {"profile": "/kubeflow-user-example-com", "labels": {"app.kubernetes.io/part-of":"kubeflow-profile","istio-injection":"enabled","katib.kubeflow.org/metrics-collector-injection":"enabled","pipelines.kubeflow.org/enabled":"true","serving.kubeflow.org/inferenceservice":"enabled"}}
1.7063747571464734e+09  INFO    controllers.Profile     Updating Istio AuthorizationPolicy  {"profile": "kubeflow-user-example-com", "namespace": "kubeflow-user-example-com", "name": "ns-owner-access-istio"}
1.706374757151141e+09   INFO    controllers.Profile     Updating RoleBinding    {"profile": "kubeflow-user-example-com", "namespace": "kubeflow-user-example-com", "name": "namespaceAdmin"}
1.7063747571522765e+09  INFO    controllers.Profile     No update on resource quota     {"profile": "/kubeflow-user-example-com", "spec": "&ResourceQuotaSpec{Hard:ResourceList{},Scopes:[],ScopeSelector:nil,}"}
```

It concerns me how this loop is running consntatnly and producing such logs.

kimwnasptd:
Then, running `make run` produces some updates to some files that are not currently tracked. Let's include those also in a commit

tarat44:
@kimwnasptd , I didn't see these issues when I was testing, but I can look into it. Which files are showing as changed when doing `make run`? Are the above logs running on an infinite loop? I also think this PR might need to be rebased so I'll do that when I have time.

tarat44:
I do see untracked changes when doing `make run` for some of the kubebuilder generated files but see them in the main branch also? @kimwnasptd, are you seeing the same?

thesuperzapper:
Aside from the fixes for filesystem watch accumulation, this seems a bit strange because it's working around a perceived issue with Kustomize and it's a bit strange to allow multiple ConfigMaps to set the same list of labels.

Also, I am pretty sure it's possible to merge ConfigMaps in Kustomize.

tarat44:
@thesuperzapper Please see https://github.com/kubernetes-sigs/kustomize/issues/370 regarding the issue with merging values from a key in a configmap. If there is a better way to do this, I am happy to implement any suggestion.

juliusvonkohout:
@kimwnasptd @thesuperzapper @rimolive this is really a blocker for the rootless stuff and the corresponding GSOC project.

thesuperzapper:
@tarat44 I am still very confused why adding `EXTRA_NAMESPACE_LABELS_PATH` is necessary. 

You mentioned that you want an optional Kustomize overlay with extra labels for all Profile Namespaces. 
Currently, the Kustomize uses this [`namespace-labels.yaml`](https://github.com/kubeflow/manifests/blob/master/apps/profiles/upstream/base/namespace-labels.yaml) in this [`configMapGenerator`](https://github.com/kubeflow/manifests/blob/7652d2f4cbe881021c2c4b40b2e8d62b91dccdbe/apps/profiles/upstream/base/kustomization.yaml#L17-L20).

Why can't we just create a new `namespace-labels.yaml` under an overlay folder with both the default and extra content? 

We can even put a warning in both files, that if one is updated, the other should be also.

/hold

tarat44:
> @tarat44 I am still very confused why adding `EXTRA_NAMESPACE_LABELS_PATH` is necessary.
> 
> You mentioned that you want an optional Kustomize overlay with extra labels for all Profile Namespaces. Currently, the Kustomize uses this [`namespace-labels.yaml`](https://github.com/kubeflow/manifests/blob/master/apps/profiles/upstream/base/namespace-labels.yaml) in this [`configMapGenerator`](https://github.com/kubeflow/manifests/blob/7652d2f4cbe881021c2c4b40b2e8d62b91dccdbe/apps/profiles/upstream/base/kustomization.yaml#L17-L20).
> 
> Why can't we just create a new `namespace-labels.yaml` under an overlay folder with both the default and extra content?
> 
> We can even put a warning in both files, that if one is updated, the other should be also.
> 
> /hold

I was worried about the issues with having repeated information in two places, but maybe this implementation just causes more confusion. I am happy to close if this really does not seem worth it. What do you think, @juliusvonkohout?

thesuperzapper:
@tarat44 either way, isn't there some fixes regarding file watches in this PR?

If so, we should at least include those.

tarat44:
> @tarat44 either way, isn't there some fixes regarding file watches in this PR?
> 
> If so, we should at least include those.

@thesuperzapper yes, if we want to drop the extra namespace labels file feature I would be happy to repurpose this PR to just have the file watcher fixes or I can create a new PR with them- whichever you think would be the cleaner approach.

thesuperzapper:
I think the current approach is already user configurable, they can either update the [`namespace-labels.yaml`](https://github.com/kubeflow/manifests/blob/master/apps/profiles/upstream/base/namespace-labels.yaml) directly, or override it in full like I was describing in https://github.com/kubeflow/kubeflow/pull/7325#issuecomment-2046239458.

It makes more sense to explain this in the user documentation rather than introduce a complex second ConfigMap merging feature.

juliusvonkohout:
> I think the current approach is already user configurable, they can either update the [`namespace-labels.yaml`](https://github.com/kubeflow/manifests/blob/master/apps/profiles/upstream/base/namespace-labels.yaml) directly, or override it in full like I was describing in [#7325 (comment)](https://github.com/kubeflow/kubeflow/pull/7325#issuecomment-2046239458).
> 
> It makes more sense to explain this in the user documentation rather than introduce a complex second ConfigMap merging feature.

The idea was to separate the user-configurable security stuff from the fixed namespace labels required for proper base functionality.

thesuperzapper:
@tarat44 thanks for your work on this, however, I am not comfortable implementing such a complex solution when we can just include two copies of the labels config in the manifests as described in https://github.com/kubeflow/kubeflow/pull/7325#issuecomment-2046239458.

However, if you are open to making a separate PR that fixes the file watcher accumulation bug, I promise to review it ASAP!

/close

PR #7333, opened 2023-10-12
Run build steps for multi-arch sequentially

kimwnasptd:
We saw that if we try to build using an ENV var containing all the architectures then docker buildx will run 3 parallel jobs.

This made the GH runners crash in some cases because they were over-utilizing the resources.

To mitigate this we'll sequentially build each image for each arch.

/assign @thesuperzapper 
cc @DnPlas

### DIFF
--- a/.github/workflows/centraldb_docker_publish.yaml
+++ b/.github/workflows/centraldb_docker_publish.yaml
@@ -11,7 +11,6 @@ on:
 env:
   DOCKER_USER: kubeflownotebookswg
   IMG: kubeflownotebookswg/centraldashboard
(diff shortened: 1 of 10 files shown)


thesuperzapper (changes requested):
@kimwnasptd can I request we do the `amd64` builds first, so they fail faster for broken images?

Also, can we make the same re-ordering change on the other workflows (I am happy to do this in a separate PR, if you want).

kimwnasptd:
@thesuperzapper I'm on it

kimwnasptd:
@thesuperzapper just pushed a commit, PTAL

PR #7335, opened 2023-10-13
fix: KeyError on KF 1.8

TobiasGoerke:
This change fixes an error 500 on notebook creation in Kubeflow 1.8.

Background: the frontend may be faster accessing the status of the newly created Notebook than the controller can create it. This causes the user to see an empty Notebook page and an error 500.

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/utils.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/utils.py
@@ -14,7 +14,9 @@ log = logging.getLogger(__name__)
 
 FILE_ABS_PATH = os.path.abspath(os.path.dirname(__file__))
 
-NOTEBOOK_TEMPLATE_YAML = os.path.join(FILE_ABS_PATH, "yaml/notebook_template.yaml")
+NOTEBOOK_TEMPLATE_YAML = os.path.join(
+    FILE_ABS_PATH, "yaml/notebook_template.yaml"
+)
 LAST_ACTIVITY_ANNOTATION = "notebooks.kubeflow.org/last-activity"
 
 # The production configuration is mounted on the app's pod via a configmap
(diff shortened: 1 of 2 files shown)


TobiasGoerke:
@kimwnasptd @DnPlas we're seeing this error with every Notebook created in v1.8. 
Imo it should be addressed in https://github.com/kubeflow/manifests/issues/2442

components/crud-web-apps/jupyter/backend/apps/common/utils.py
```diff
@@ -14,7 +14,7 @@
 
 FILE_ABS_PATH = os.path.abspath(os.path.dirname(__file__))
 
-NOTEBOOK_TEMPLATE_YAML = os.path.join(FILE_ABS_PATH, "yaml/notebook_template.yaml")
+NOTEBOOK_TEMPLATE_YAML = os.path.join(FILE_ABS_PATH, "yaml/notebook_template.yaml") # noqa: E501
```
kimwnasptd, line 17: Nit, can we do something like this?
```python
NOTEBOOK_TEMPLATE_YAML = os.path.join(FILE_ABS_PATH, 
                                      "yaml/notebook_template.yaml")
```

components/crud-web-apps/jupyter/backend/apps/common/utils.py
```diff
@@ -97,7 +97,7 @@ def pvc_from_dict(vol, namespace):
         spec=client.V1PersistentVolumeClaimSpec(
             access_modes=[vol["mode"]],
             storage_class_name=get_storage_class(vol),
-            resources=client.V1ResourceRequirements(requests={"storage": vol["size"]}),
+            resources=client.V1ResourceRequirements(requests={"storage": vol["size"]}), # noqa: E501
```
kimwnasptd, line 100: instead of ignoring the line length could we do something like
```python
resources=client.V1ResourceRequirements(
    requests={"storage": vol["size"]}
),
```

kimwnasptd (commented):
@TobiasGoerke the changes look good and we can cherry-pick to 1.8 yes. I'll have my eyes constantly on this today.

Can you just make those small edits in the code to also pass linting?

TobiasGoerke:
> @TobiasGoerke the changes look good and we can cherry-pick to 1.8 yes. I'll have my eyes constantly on this today.
> 
> Can you just make those small edits in the code to also pass linting?

Done!
https://github.com/kubeflow/kubeflow/pull/7339

kimwnasptd:
Thank you!

/lgtm
/approve

PR #7336, opened 2023-10-13
Make principals dynamic in Profiles Controller (#7310)

DnPlas:
* Make principals dynamic in Profiles Controller

* review: Use dynamic principal of IGW in KFAM

* review: Add env vars to manifests

* review: Add pipelines-ui principal

* Update golang to 1.19 for unit tests to succeed

* review: Include KFP UI principle in profiles/kfam

---------

This is a cherry pick from #7310

### DIFF
--- a/.github/workflows/prof_controller_unit_test.yaml
+++ b/.github/workflows/prof_controller_unit_test.yaml
@@ -14,7 +14,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v3
       with:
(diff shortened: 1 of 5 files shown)


kimwnasptd:
/lgtm
/approve

Thanks Dani!

juliusvonkohout:
So it seems that https://github.com/kubeflow/pipelines/issues/9889 is still valid then.

PR #7337, opened 2023-10-13
Cherry-pick: PodDefaults versions bump #7322 and #7338

kimwnasptd:
Cherry-pick
* #7322 
* #7338


kimwnasptd:
Looks like unit tests are breaking for the controller, which were not triggered in #7322

Sending a PR to fix

PR #7350, opened 2023-10-16
fix: return waiting on no pvcviewer status

TobiasGoerke:
In some cases, a PVCViewer object has no status as it was just created or is having issues starting a pod.  In those cases, we should rather return waiting than uninitialized, which would appear to the users as if no viewer was started.

### DIFF
--- a/components/crud-web-apps/volumes/backend/apps/common/status.py
+++ b/components/crud-web-apps/volumes/backend/apps/common/status.py
@@ -51,15 +51,14 @@ def viewer_status(viewer):
     Return a string representing the status of that viewer. If a deletion
     timestamp is set we want to return a `Terminating` state.
     """
-    try:
-        ready = viewer["status"]["ready"]
-    except KeyError:
+    if viewer is None or not viewer:
         return status.STATUS_PHASE.UNINITIALIZED
 
-    if "deletionTimestamp" in viewer["metadata"]:
+    metadata = viewer.get("metadata", {})
+    if "deletionTimestamp" in metadata:
         return status.STATUS_PHASE.TERMINATING
 
-    if not ready:
-        return status.STATUS_PHASE.WAITING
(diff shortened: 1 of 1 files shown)


DnPlas:
Please keep me in the loop for this one in case we want to include this (probably we should) in v1.8.

thesuperzapper:
@TobiasGoerke I think that it makes sense to keep using `uninitialized` when there is not yet a PVC Viewer created, otherwise, it would make all volumes show as `waiting` for a viewer, even if none is requested.

This is because `viewer_status()` is passed an empty `{}` by this code for ALL PVCs:

https://github.com/kubeflow/kubeflow/blob/f4f8bf14fb00d013d351d2d25643f59f1338ab77/components/crud-web-apps/volumes/backend/apps/default/routes/get.py#L23

As soon as a PVCViewer is created, it should have a valid `status.ready`, which will make its status as `waiting`, with the existing code.

/hold

TobiasGoerke:
> @TobiasGoerke I think that it makes sense to keep using `uninitialized` when there is not yet a PVC Viewer created, otherwise, it would make all volumes show as `waiting` for a viewer, even if none is requested.
> 
> This is because `viewer_status()` is passed an empty `{}` by this code for ALL PVCs:
> 
> https://github.com/kubeflow/kubeflow/blob/f4f8bf14fb00d013d351d2d25643f59f1338ab77/components/crud-web-apps/volumes/backend/apps/default/routes/get.py#L23
> 
> As soon as a PVCViewer is created, it should have a valid `status.ready`, which will make its status as `waiting`, with the existing code.
> 
> /hold

Good catch, you're absolutely right!
Just tested this and the status subresource won't be empty but None once the pvcviewer has been created without the controller having reconciled, yet. So, users will see an empty folder and upon clicking it again, see a 409 error.

I've added an additional check for empty / None viewer, which has the function now work as intended.

Issue #7283, opened 2023-09-18
pvcviewer - make the istio gateway configurable for VirtualServices


misohu:
/kind feature

Right now PVCViewer controller will set the gateway to `kubeflow/kubeflow-gateway` for every CR. This value cannot be changed. Const [here](https://github.com/kubeflow/kubeflow/blob/master/components/pvcviewer-controller/controllers/pvcviewer_controller.go#L55) Can we make this value configurable ? E.g. using environment variable? 

**Why you need this feature:**
Sometimes I want to use different gateway name for PVCViewer virtual services. E.g. different name or gateway is in different namespace.

**Describe the solution you'd like:**
I want to load the value [here](https://github.com/kubeflow/kubeflow/blob/master/components/pvcviewer-controller/controllers/pvcviewer_controller.go#L55) from environment variable so every new PVCViewer CR will set it in gateway field for VirtualService [here](https://github.com/kubeflow/kubeflow/blob/master/components/pvcviewer-controller/controllers/pvcviewer_controller.go#L288).

TobiasGoerke:
Thanks for the issue. Would #7351 solve your issue?

thesuperzapper:
Resolved by https://github.com/kubeflow/kubeflow/pull/7351
/close

PR #7351, opened 2023-10-16
feat: make pvcviewer istio gateway configurable

TobiasGoerke:
Allows the user to specify an optional controller env var "ISTIO_GATEWAY", which allows for other istio gateways than "kubeflow/kubeflow-gateway".

See https://github.com/kubeflow/kubeflow/issues/7283

### DIFF
--- a/components/pvcviewer-controller/controllers/pvcviewer_controller.go
+++ b/components/pvcviewer-controller/controllers/pvcviewer_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"os"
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@TobiasGoerke Thanks, this is great, and very important for users who don't use the `kubeflow/kubeflow-gateway` as the namespace/name of their istio gateway.

/lgtm
/approve

thesuperzapper:
@DnPlas @kimwnasptd This one is critical to cherry-pick before the final Kubeflow 1.8 RC.

PR #7353, opened 2023-10-16
fix: point to ~kubeflow/styles/fonts.scss

DnPlas:
It seems like the path to this file recently changed by #7062, which is now preventing the centraldashboard Docker image to build wich fails with a SassError when trying to import fonts.scss. This commit points to the right path.

This PR cherry picks the changes of #7352 to master.


orfeas-k:
Thank you @DnPlas 
/lgtm
/approve

PR #7365, opened 2023-10-18
allow overriding culling per notebook

omrishiv:
This PR allows an admin to enable overriding notebook culling. This is useful for:
- users who are not running active jupyter kernels, but doing work in the terminal, which does not report its status
- notebooks that are running on a full node you may not want to autoscale back, but are not doing immediately active work

By adding the annotation `notebooks.kubeflow.org/disable-culling=true` a user is able to mark their notebook as not cullable. This requires setting the `ENABLE_OVERRIDE_CULLING_ANNOTATION=true` when deploying the notebook controller


components/notebook-controller/controllers/culling_controller.go
```diff
@@ -381,6 +388,16 @@ func StopAnnotationIsSet(meta metav1.ObjectMeta) bool {
 	return false
 }
 
+func DisableCullingAnnotationIsSet(meta metav1.ObjectMeta) bool {
```
kimwnasptd, line 391: small nit, shouldn't this function start from lowercase since it's not meant to be used from outside this go module?

OmriShiv: I was following the style of the current code using `StopAnnotationIsSet` above. Agreed it should be lowercase

andreyvelich:
Please re-open PR there.
/close

PR #7357, opened 2023-10-17
feat: update notebook server images + support ARM64

thesuperzapper:
This PR significantly updates and improves all our example notebook server images.

## Key changes

1. __Support for ARM64 in addition to AMD64:__
      - __NOTE:__ the CUDA images are not currently built for ARM64, as I have no way to test them
          - PyTorch: I don't think that pre-compiled versions of PyTorch with CUDA on ARM are available.
          - TensorFlow: The official NVIDIA Ubuntu repos for CUDA are a bit sparse on ARM for CUDA 11.8 (which is the latest that TF supports)
2. __Much cleaner build system and Makefiles:__
      - Build any image locally by going to its folder and running `make docker-build-dep` to build with all base images that that image depends on.
3. __Caching in GitHub Action Builds:__
      - We are now using the `ghcr.io/kubeflow/kubeflow/notebook-servers/build-cache` image to store caches, which should significantly speed up builds.
4. __TensorFlow 2.0:__
      - We have updated to TensorFlow `2.13.0`  by default.
      - We have updated to CUDA `11.8` in the TensorFlow CUDA images. 
5. __PyTorch 2.0:__ 
      - We have updated to PyTorch `2.1.0`  by default.
      - We have updated to CUDA `12.1` in the PyTorch CUDA images. 
6. __JupyterLab 4.0:__
      - We have updated to JupyterLab `4.0.7`  by default.
      - __NOTE:__ we had to remove the [`jupyterlab-git`](https://github.com/jupyterlab/jupyterlab-git) plugin, as [it does not yet support JupyterLab 4.0](https://github.com/jupyterlab/jupyterlab-git/issues/1245), but hopefully they will release an update in the coming days.
7. __Python 3.11:__
      - We have updated to Python `3.11.6`  by default.

I have tested the images in real-world use cases for Tensorflow and PyTorch (including on GPUs), but we will need to get more feedback after we release Kubeflow 1.8 with these images. 

## Other Notes

- We still don't have a sensible way to test build the images on PR (because we have to split the builds up as they are so big, but we don't want to push random user's PRs to any container registries)
- Each commit to master (that updates anything under the `components/example-notebook-servers/` folder, will trigger a build of all notebook servers (which should be fast, because of the caching, unless the PR changes the base images).
- Previously, we were not publishing the intermediate images to DockerHub (like `base`, `jupyter`, etc), this PR changed that, and now all images are always pushed.
- The CUDA images now have their own folders named:
[...]

PR #7386, opened 2023-10-25
ci: fix version tagging of notebook server images

thesuperzapper:
There was a small bug in the CI from https://github.com/kubeflow/kubeflow/pull/7357, which cause the release version tags to not be created (when the `releasing/version/VERSION` file is updated).

It was simply doing the `cd` too early, so the relative path was not correct.

We will probably need to quickly do a `1.8.0-rc.5` after cherry-picking this pr into the `v1.8-branch`.


thesuperzapper:
We need to quickly merge this and cut a new RC for 1.8, as right now, the notebook server image tags for `1.8.0-rc.4` are not present, see: https://hub.docker.com/r/kubeflownotebookswg/jupyter-tensorflow-full/tags

/assign @kimwnasptd @DnPlas

DnPlas (approved):

thesuperzapper:
Self-approving, because we need to get this cut ASAP, and is a minor fix.

/approve

PR #7390, opened 2023-10-26
chore: update default filebrowser version

TobiasGoerke:
Filebrowser recently [released v2.25.0](https://github.com/filebrowser/filebrowser/releases/tag/v2.25.0).
Until now, we couldn't use v2.24.0, which brings chunked uploads, due to an issue with Cloudflare environments.  This got fixed in v2.25.0, so that chunked uploads should now work for every user.


kimwnasptd:
Nice one!

/lgtm
/approve

PR #7398, opened 2023-10-31
fix: downgrade jupyterlab to 3.6.6

thesuperzapper:
Follow up to https://github.com/kubeflow/kubeflow/pull/7357

This PR downgrades JupyterLab to 3.6.6 and Jupyter Notebooks to 6.5.6.

This is to ensure compatibility with JupyterLab extensions which have not supported JupyterLab 4.0 yet.

I have also re-added the `jupyterlab-git` extension, as this was one of the plugins that did not support 4.0.

Finally, I have updated the `kfp` version to `2.4.0`, as this version fixes a bug uncovered in Kubeflow 1.8.0-rc.4.

### DIFF
--- a/components/example-notebook-servers/codeserver-python/Dockerfile
+++ b/components/example-notebook-servers/codeserver-python/Dockerfile
@@ -43,11 +43,9 @@ RUN case "${TARGETARCH}" in \
  && rm /tmp/Miniforge3.sh \
  && conda config --system --set auto_update_conda false \
  && conda config --system --set show_channel_urls true \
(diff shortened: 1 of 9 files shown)


thesuperzapper:
@DnPlas I am going to self-approve here, so you can LGTM to merge it, then wait for the builds before cherry-picking into the `release-1.8` branch.

/approve

DnPlas (commented):
Thanks @thesuperzapper , should we only change the jupyterlab related packages? I think we should not touch other packages, specially kfp.

thesuperzapper:
@DnPlas the KFP packages were part of the fix found in testing for Kubeflow 1.8.0-rc.4, so it would be a bit crazy to ship a known broken version of the kfp package?

PR #7401, opened 2023-10-31
ci: prevent triggering release builds accidentally

thesuperzapper:
There is currently a serious bug in the CI which may result in image tags being re-pushed post-release.

The issue is that our current system to detect when the `releasing/version/VERSION` file has changed (and thus a new "release" image tag needs to be created) will trigger 100% of the time.

This is because, by default, the [`dorny/paths-filter@v2`](https://github.com/dorny/paths-filter) will compare against `master`, even if the push was to the `v1.8-branch`, and there will always be a different on the `releasing/version/VERSION` between master and release branches.

This PR simply uses `${{ github.ref }}` as the base of comparison, as [suggested by the author of `dorny/paths-filter`](https://github.com/dorny/paths-filter#change-detection-workflows) for long-lived branches.

### DIFF
--- a/.github/workflows/centraldb_angular_docker_publish.yaml
+++ b/.github/workflows/centraldb_angular_docker_publish.yaml
@@ -25,6 +25,7 @@ jobs:
     - uses: dorny/paths-filter@v2
       id: filter
       with:
+        base: ${{ github.ref }}
         filters: |
           version:
             - 'releasing/version/VERSION'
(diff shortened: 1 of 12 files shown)


thesuperzapper:
Once we merge this, we need to cherry-pick it into the `v1.8-branch` before we cut the final `1.8.0` tag.

PR #7407, opened 2023-11-02
fix: jupyterlab runtime volume permissions

### LINKED ISSUE #7370: jupyter notebook not connect only
jupyter notebook not connect only but vs code and r studio notebook successfully connected.
when i check notebook pod logs, i get this runtime error "RuntimeError: Permissions assignment failed for secure file: '/home/jovyan/.local/share/jupyter/runtime/jpserver-9060.json'. Got '0o677' instead of '0o0600'."
and on kubeflow ui i see this error "upstream connect error or disconnect/reset before headers. reset reason: connection failure, transport failure reason: delayed connect error: 111"
image
image
image

thesuperzapper:
resolves #7370

This PR fixes a longstanding issue in the JupyterLab images, which can cause them to fail to start if the PVC mounted to the home directory has incorrect file permissions for the `/home/jovyan/.local/share/jupyter/runtime` folder. 

This issue often happens for NFS-based PVCs because they typically do not support Unix file permissions, so they often present everything as 755.

The runtime files should not be persisted between restarts, so this PR makes them be stored in the `/tmp` directory.

---

It also adds the `--ServerApp.allow_remote_access=True` argument, which is unrelated, but a good idea to prevent proxy issues.

### DIFF
--- a/components/example-notebook-servers/jupyter/Dockerfile
+++ b/components/example-notebook-servers/jupyter/Dockerfile
@@ -79,16 +79,14 @@ RUN python3 -m pip install -r /tmp/requirements.txt --quiet --no-cache-dir \
 # s6 - copy scripts
 COPY --chown=${NB_USER}:users --chmod=755 s6/ /etc
 
+# configure - jupyter
+RUN jupyter labextension disable "@jupyterlab/apputils-extension:announcements"
+
 # s6 - 01-copy-tmp-home
 USER root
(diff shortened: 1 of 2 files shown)


thesuperzapper:
@kimwnasptd this is a fairly important fix, so we should at least merge it so that there are images in the DockerHub repo that have this fix.

thesuperzapper:
@kimwnasptd just following up for a review of this small but important PR

PR #7421, opened 2023-12-07
Prevent S6 timeouts on container start by setting S6_CMD_WAIT_FOR_SER…

juliusvonkohout:
…VICES_MAXTIME to 0

Might be needed for slow storage see https://github.com/pi-hole/docker-pi-hole/pull/1192

@thesuperzapper i did not yet test this.

Fixes https://github.com/kubeflow/kubeflow/issues/7370#issuecomment-1830802354

### DIFF
--- a/components/example-notebook-servers/base/Dockerfile
+++ b/components/example-notebook-servers/base/Dockerfile
@@ -13,6 +13,8 @@ ENV NB_UID 1000
 ENV NB_PREFIX /
 ENV HOME /home/$NB_USER
 ENV SHELL /bin/bash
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
@kimwnasptd @thesuperzapper @DnPlas this might be interesting for Kubeflow 1.8.1

thesuperzapper:
@kimwnasptd I'm not sure we should have merged this. We need to be very careful about this one.

It seems like we should definitely have some kind of timeout to prevent a deadlock situation where a failed service just hangs forever, and the pod never starts properly (but could incorrectly show as "running" from Kubernetes perspective).

We should check what the default is, and just make it a little bit longer (to something that is still kind of unreasonable like 5-10 minutes).

Issue #7456, opened 2024-01-26
Notebook doesn't created after upgrade to Kubeflow 1.8


pashtet04:
/kind bug

**What steps did you take and what happened:**
I upgraded Kubeflow installation from v1.7 to v1.8 using [v1.8-branch](https://github.com/kubeflow/manifests/tree/v1.8-branch) and I can't create any Notebooks more.

Root cause in this [line](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/jupyter/backend/apps/common/yaml/notebook_template.yaml#L10C1-L10C46) `notebooks.kubeflow.org/creator: {creator}`
It happens because my user is not a string value
`2024-01-26 19:41:52,422 | kubeflow.kubeflow.crud_backend.authn | INFO | Handling request for user: 60095663`

```
2024-01-26 15:08:58,070 | apps.default.routes.post | INFO | Got body: {'name': 'dsadasda', 'namespace': 'main', 'image': 'kubeflownotebookswg/jupyter-scipy:v1.8.0', 'allowCustomImage': True, 'imagePullPolicy': 'IfNotPresent', 'customImage': '', 'customImageCheck': False, 'serverType': 'jupyter', 'cpu': '0.5', 'cpuLimit': '0.6', 'memory': '1Gi', 'memoryLimit': '1.2Gi', 'gpus': {'num': 'none'}, 'affinityConfig': '', 'tolerationGroup': '', 'shm': True, 'configurations': [], 'workspace': {'mount': '/home/jovyan', 'newPvc': {'metadata': {'name': 'dsadasda-workspace'}, 'spec': {'accessModes': ['ReadWriteOnce'], 'resources': {'requests': {'storage': '5Gi'}}}}}, 'datavols': []}
2024-01-26 15:08:58,073 | apps.common.form | INFO | Using provided value for 'image': kubeflownotebookswg/jupyter-scipy:v1.8.0
2024-01-26 15:08:58,073 | apps.common.form | INFO | Using provided value for 'imagePullPolicy': IfNotPresent
2024-01-26 15:08:58,073 | apps.common.form | INFO | Using provided value for 'cpu': 0.5
2024-01-26 15:08:58,073 | apps.common.form | INFO | Using provided value for 'memory': 1Gi
2024-01-26 15:08:58,073 | apps.common.form | INFO | Using provided value for 'gpus': {'num': 'none'}
[...]

PR #7457, opened 2024-01-26
fix: :label: Ensure username is returned by `get_username` as a string

pashtet04:
In order to maintain consistency and compatibility, explicitly convert the 'username' variable to a string using the str() function before returning it. This ensures that the function consistently provides a string representation of the username, addressing potential variations in the original data type.

https://github.com/kubeflow/kubeflow/issues/7456

Update notebook_template.yaml


thesuperzapper:
@kimwnasptd, do we need to account for the fact that a username might contain `"`, and either escape it, or use a multi-line YAML pipe operator (e.g. `>`)

PR #7468, opened 2024-02-20
fix: resolve service name conflict for notebooks and tensorboards

### LINKED ISSUE #7455: tensorboard fails to connect if there is a notebook with the same name
/kind bug

Currently, if you create a `Notebook` and `Tensorboard` resource with the same name (in the same namespace), then Istio will not correctly route the traffic to one or the other.

This is caused by the notebook/tensorboard controllers both trying to use the name of the notebook/tensorboard as the name of the Service, in addition to not having a specific enough Pod label selector (notebooks uses `statefulset: {notebook_name}` and tensorboard uses `app: {tensorboard_name}`) which is clearly way too broad.

While we are fixing this, we should also have the tensorboard controller name its VirtualServices `tensorboard-{tensorboard_name}` (rather than just the name of the tensorboard, which could overlap). Note, the notebooks controller already uses `notebook-{namespace_name}-{notebook_name}` for its VirtualService names.

mishraprafful:
#### Related Issues/PRs
Closes kubeflow/dashboard#50

###### Notebook Controller
* Update names of `service`, `virtualService` and `statefulset` resource to include
  * `nb-{notebook_name}-`  with `generateName` 
* Added a reconcile check for name length check as we are adding prefixes
* Added a function that deletes obsolete services, so as to not leave dangling services after folks update.
* * Added a function that deletes obsolete services, so as to not leave dangling services after folks update.
* Update `servicename` in `virtualService` after reconciliation of service
* Upated code and tests to check owner reference as the source of resource connection

###### Tensorboard Controller
* Update `name` of `service` and `virtualService` to `tb-{notebook-name}`
* Add label `tensorboard-name` in pods and for service label selector
* Add label `notebook-name` in service `selector`
* Added a reconcile check for name length check as we are adding prefixes

#### Comments
- This might be a breaking change, and if there are notebook servers with long names, they might fail to reconcile as we are adding name, and namespace prefix.

### DIFF
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -449,7 +449,7 @@ func generateService(instance *v1beta1.Notebook) *corev1.Service {
 		},
 		Spec: corev1.ServiceSpec{
 			Type:     "ClusterIP",
-			Selector: map[string]string{"statefulset": instance.Name},
+			Selector: map[string]string{"statefulset": instance.Name, "notebook-name": instance.Name},
 			Ports: []corev1.ServicePort{
 				{
 					// Make port name follow Istio pattern so it can be managed by istio rbac
--- a/components/notebook-controller/controllers/notebook_controller.go
+++ b/components/notebook-controller/controllers/notebook_controller.go
@@ -17,6 +17,7 @@ package controllers
 
 import (
 	"context"
+
 	"encoding/json"
 	"fmt"
 	"os"
@@ -34,6 +35,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/tools/record"
@@ -53,7 +55,8 @@ const AnnotationRewriteURI = "notebooks.kubeflow.org/http-rewrite-uri"
 const AnnotationHeadersRequestSet = "notebooks.kubeflow.org/http-headers-request-set"
 
 const PrefixEnvVar = "NB_PREFIX"
-const namePrefix = "notebook-"
+const namePrefix = "nb"
+const generatedSuffixLength = 5
 const maxNameLength = 63
 
 // The default fsGroup of PodSecurityContext.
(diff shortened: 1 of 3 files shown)


thesuperzapper:
We need to consider what will happen to existing TensorBoard instances when this update is applied to the cluster.

We should test what happens, it looks like we are setting the `ownerReferences` of the Service/VirtualService to the TensorBoard CRD instance, so at worst, we can tell people to delete and recreate all their TensorBoard CRD instances:

- https://github.com/kubeflow/kubeflow/blob/master/components/tensorboard-controller/controllers/tensorboard_controller.go#L114

mishraprafful:
@thesuperzapper Could we re-run the tests it seems like the unit test of notebook-controller fails on some dependency, and is unrelated to the changes in the PR. 🙏🏻 

Also, the integration tests for the same pass. 😄

thesuperzapper:
@mishraprafful can you please check what happens to the existing resources when you upgrade from the old version of the controller to the new version (specifically, does it remove the old `Service/{notebook_name}`, or leave it dangling).

If it leaves it, we should consider having the controller look for `Services` / `VirtualSerivces` which have an owner reference of a Notebook, but are not named with the `notebook-XXX` or `tensorboard-XXX` prefix.

mishraprafful:
Sure @rimolive Looking into it.

thesuperzapper:
Also, there was a historical PR to fix this issue, which also accounted for the case that the name is too long https://github.com/kubeflow/kubeflow/pull/6701

They put 63 characters, but I am not sure if VirtualServices/Services can actually be longer than that?

mishraprafful:
@rimolive the tests are failing to pull some dependencies, the tests pass when I try them locally. Maybe we could re-trigger the tests as this seems like an intermediate issue. 

Feel free to point out if I'm missing something 🙌🏻

mishraprafful:
@thesuperzapper Service name has to be less than 64 
> The name of a Service object must be a valid [RFC 1035 label name](https://kubernetes.io/docs/concepts/overview/working-with-objects/names#rfc-1035-label-names).

[ref](https://kubernetes.io/docs/concepts/services-networking/service/#:~:text=The%20name%20of%20a%20Service%20object%20must%20be%20a%20valid%20RFC%201035%20label%20name)

mishraprafful:
Just an update: now that #7589 is merged, I'm testing a last few things, should add some more commits soon to finish the functionality.

mishraprafful:
@thesuperzapper @rimolive @kimwnasptd All issues addressed, tests passing, PR description updated. 🚀

thesuperzapper:
@mishraprafful I am going to push this to 1.9.1, because we came across a better way to do this while building Notebooks 2.0.

The key idea is to use the [`generateName`](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#idempotency) field of every Kubernetes object, rather than using the `name` directly, which is how Kubernetes creates Pod names from deployments with those `{DEPLOYMENT_NAME}-{POD_HASH}-{RANDOM_SUFFIX}` names.

Obviously, this means we need to account for random names (which is also great for backwards compatibly). 

Rather than assuming the `Service`/`VirtualService`/`ReplicaSet` has the same name as the `Notebook` (or `notebook-{NOTEBOOK_NAME}`), we should instead loop over the corresponding resource type in the target namespace and find the ones which have [`metadata.ownerReferences`](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) set to the `Notebook` being reconciled.

If there are no such resources, we can create a new one using the `generateName` set to `nb-{NOTEBOOK_NAME}-` (with the owner reference already set to avoid problems if the controller crashes).

We will need to be careful about the `Service` names, because obviously the `VirtualService` needs to point to the correct `Service` name, but this can simply be handled by ordering the reconciliation correctly, so we know the `Service` name before reconciling the `VirtualService`.

mishraprafful:
@thesuperzapper Apologies for the huge delay (life happened 😉 ), but I've updated the PR according to your last comment now.

jiridanek:
After a quick look, seems to me Indexer is not used, as is in

```
	// setup field indexers on the manager cache. we use these indexes to efficiently
	// query the cache for things like which Workspaces are using a particular WorkspaceKind
	if err := helper.SetupManagerFieldIndexers(mgr); err != nil {
		setupLog.Error(err, "unable to setup field indexers")
		os.Exit(1)
	}
```

https://github.com/kubeflow/notebooks/blob/9d02acfdd707282385221590bc953f892099a94c/workspaces/controller/cmd/main.go#L129-L134

Issue #7480, opened 2024-02-22
Propagate labels from `profile` custom resource to controller-generated namespace


droctothorpe:
/kind feature

**Why you need this feature:**
We use the profile operator directly instead of pathing through the UI. We want end user namespaces to have specific labels (that identify various things about the user). [This code](https://github.com/kubeflow/kubeflow/blob/master/components/profile-controller/controllers/profile_controller.go#L137) lets administrators set **global** namespace labels, but there's no mechanism to pass custom-resource-specific, non-global labels.

**Describe the solution you'd like:**
The simplest way to implement this is to just propagate labels from the `profile` custom resource to the corresponding namespace. It's a one line change to the controller logic.

**Anything else you would like to add:**
I'll submit a PR momentarily.

PR #7481, opened 2024-02-22
feat(profile-controller): propagate labels from CR to namespace

droctothorpe:
Resolves https://github.com/kubeflow/kubeflow/issues/7480.

We noticed that there were no tests for the `Reconcile` function in `profile_controller_test.go`, otherwise we would have extended those tests to validate this specific scenario, even though it's just one line of code with very limited impact. We'd potentially be open, bandwidth permitting, to contributing broader tests for `Reconcile` down the road.


droctothorpe:
@thesuperzapper per the WG meeting, let me know if you think it makes more sense to add something like `namespaceLabels` to the profile CR spec. We tested to confirm that, at present, no labels are applied to profile CRs. The main advantage of this approach is that it's a much smaller lift since it doesn't require updating the CRD, but easy doesn't necessarily mean right.

droctothorpe:
Bump.

thesuperzapper:
I think this PR is a security issue for many users, holding for review.

/hold

thesuperzapper:
I am really not sure how (or even if) we can do this safely without causing the following issues:

1. We can't replicate the labels from the profile resource, because this will break GitOps apps like ArgoCD (which use labels to track resources)
2. We can't add a new `spec.namespaceLabels` to the Profile CRD, because that might be a security risk, and allow users to arbitrarily label Namespaces to avoid some restrictions. (But I am more open to be convinced on this one).

droctothorpe:
Thanks, @thesuperzapper. Regarding (1), could I trouble you to clarify your concerns with the first option? Is the concern that if profiles are created used ArgoCD, the labels will be propagated to the namespace? Why would that be a problem? Regarding (2), in our use case, end users don't have permission to edit namespaces or profiles directly. It's hard to imagine a scenario where namespace CRUD access would be disabled but profile CRUD access would be enabled, except via the UI as a privileged service. In addition, what if we contributed this feature but disabled it by default and made it contingent on a specific env var being set (that can be configured via the chart)?

andreyvelich (commented):
The Kubeflow Profile code has been migrated to the new repo, please re-open PR at: https://github.com/kubeflow/dashboard
/close

PR #7547, opened 2024-04-11
feat: allow updating profile owners

tzstoyanov:
Fixes issue https://github.com/kubeflow/dashboard/issues/33
Fixes issue https://github.com/kubeflow/dashboard/issues/45

This PR fixes the problem, described in issue https://github.com/kubeflow/dashboard/issues/33. Currently, the profile-controller fails to update the owner as it cannot verify the namespace owner. The PR implements solutions, suggested in the issue.

Additional logic is added, to address issue https://github.com/kubeflow/dashboard/issues/45. If the namespace already exist, and it is not owned by any other profile - check if the namespace has annotation transferToKubeflow: true. If there is such annotation, take the ownership of that namespace and use it for the profile.

### DIFF
--- a/.github/workflows/prof_controller_unit_test.yaml
+++ b/.github/workflows/prof_controller_unit_test.yaml
@@ -14,7 +14,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v3
       with:
-        go-version: '1.19'
+        go-version: '1.21'
         check-latest: true
 
     - name: Run unit tests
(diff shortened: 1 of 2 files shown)


juliusvonkohout:
@thesuperzapper can we merge this now for the next release cycle?

thesuperzapper:
@juliusvonkohout I'm pretty sure this is not very close to being ready to merge. 

There's a lot of complexity that needs to be considered, and I'm not even sure this approach is the right one.

andreyvelich (commented):
The Kubeflow Profile Controller code has been migrated to the new repo, please re-open PR at: https://github.com/kubeflow/dashboard
/close

PR #7551, opened 2024-04-15
Add logic to handle when poddefaults use  matchExpression

boarder7395:
This PR is to fix issue #6812.

Poddefaults support matchExpressions, but the logic for /api/namespaces/<namespace>/poddefaults fails when the matchExpressions field is used in place of matchLabels. This approach updates the label name using the matchExpression key.


ca-scribner:
I hit this bug today too, but I think the issue is slightly different (details in #7552).  Rather than handle both `matchLabels` and `matchExpressions`, I believe the `label` returned here is probably supposed to be the `name` of the `PodDefault`, not the first selector in the list

wjhhuizi:
Just adding some more information here:

The matchExpressions is actually working, but the UI/API does not support it.

API: `https://<hostname>/jupyter/api/namespaces/<namespace>/poddefaults`
```
{
    "log": "An error occured in the backend.",
    "status": 500,
    "success": false,
    "user": "<Email@Address>"
}
```

PR #7578, opened 2024-05-16
fix: update central dashboard to NodeJS 16

thesuperzapper:
This PR updates the central-dashboard to use NodeJS 16.

__WARNING:__ node 16 is the latest we can use without significant refactoring.

This PR also updates the following dependency packages (to allow for NodeJS 16 support):

- `@kubernetes/client-node` to `0.19.0`
- `express` to `4.19.2`
- `@grpc/grpc-js` to `1.10.8`
- `@types/serve-static` to `1.15.5`
- `typescript` to `4.9.5`
- `webpack` to `4.47.0`

There were also a number of typescript linting errors which the newer versions of everything seem to detect, so I fixed those also.


orfeas-k:
May I ask what's the rationale here for using `tini` in the container's entrypoint? Was it required from other changes in the PR? (My understanding is that it should not)

thesuperzapper:
@orfeas-k it's just a good practice, as many Node packages do not support running in containers very well, and might not respond to SIGINT / SIGKILL properly.

Is there a specific reason you don't like `tini`?

PR #7593, opened 2024-05-30
Tensorboard Add start-stop UI function and culling controller

dgoupil:
The idea of the PR is to replicate notebook server behavior.
Tensorboards are interactive workloads that should be stopped when not used or after a certain amount of seconds, just like notebook servers. This will free up resources but more importantly, tensorboards are constantly reading logs which can add significant cost when logs are stored on public cloud providers.

Few features
- Add start and stop button to the tensorboard dashboard --> UI backend adds stop resource annotation --> Controller changes deployment replica accordingly 
- Get tensorboard status and update UI
- Add tensorboard culler to stop (adds stop resource annotation) after x seconds
- Enabling culling and time for culling is configurable
- Add "patch" permission to the web app service account


dgoupil:
Thanks @YQ-Wang for the culler part.

PR #7595, opened 2024-05-31
fix: reduce complexity of email regex in angular dashboard

CodeVigilanteOfficial:
fix for this https://huntr.com/bounties/0c1d6432-f385-4c54-beea-9f8c677def5b

### DIFF
--- a/components/centraldashboard-angular/backend/app/api_workgroup.ts
+++ b/components/centraldashboard-angular/backend/app/api_workgroup.ts
@@ -8,8 +8,8 @@ import {
     ERRORS,
 } from './api';
 
-// From: https://www.w3resource.com/javascript/form/email-validation.php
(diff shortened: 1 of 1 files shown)


thesuperzapper:
@CodeVigilanteOfficial I just want to note that the `centraldashboard-angular` rewrite is currently unreleased.

However, the normal dashboard uses the same email regex, so might also be susceptible. 

In another PR we are planning to update it, but to a different regex, I would appreciate if you could take a look at it, and see if there are any performance concerns. 

https://github.com/kubeflow/kubeflow/pull/7597

thesuperzapper:
Also, did this PR accidentally change the line-endings of this file? 

Because the GitHub diff is showing the full file as changed.

thesuperzapper:
closing because https://github.com/kubeflow/kubeflow/pull/7597 was merged a while ago

/close

PR #7631, opened 2024-07-29
Change LOGOUT_URL from central dashboard to oauth2 proxy

tzabbi:
Since kubeflow switched to oauth2 proxy we have to adjust the logout url from the central dashboard


juliusvonkohout:
closes https://github.com/kubeflow/kubeflow/pull/7631

thesuperzapper:
This was done in https://github.com/kubeflow/kubeflow/pull/7654

/close

PR #7649, opened 2024-10-09
fix: return notebook events error message

gigabyte132:
Fixes https://github.com/kubeflow/notebooks/issues/64

Notebook events are not currently returned resulting in the generic error message `Couldn't find any information for the status of this notebook.` when unsuccessfully creating a notebook. 

This is because the error message from notebook events is currently being overridden by the generic error message.


thesuperzapper:
Since we are about to cut a 1.9.2 with a basic fix for the central dashboard, we might as well include this also.

/lgtm
/approve

PR #7656, opened 2024-10-23
feat(notebooks): add AMD ROCm example notebook images

gigabyte132:
/kind feature
# Motivation
Following a discussion at the previous notebooks-wg meeting, adding support for having pytorch images for AMD GPUs was raised. 
# What does this PR do?
This PR adds ROCm pytorch images to Kubeflow with one caveat, a udev change required on the hosts on which these containers/pods run so that any user besides `root` can access and use the devices. This is due to the group id of the `render` linux group being dynamic. This was raised to AMD upstream and their input was to do the following as we cannot bake the `render` group into the image. 

For the issues see:
https://github.com/ROCm/ROCm-docker/issues/90 and https://github.com/ROCm/k8s-device-plugin/issues/39

Doing the following on the host:

Create a new file (if it doesn't exist) `/etc/udev/rules.d/70-amdgpu.rules` with the following content:

`KERNEL=="kfd", MODE="0666" SUBSYSTEM=="drm", KERNEL=="renderD*", MODE="0666"`

Reload the udev rules with:

`sudo udevadm control --reload-rules && sudo udevadm trigger`

With these changes the `jovyan` user has access to the devices and can run workloads on AMD GPUs.

# Test Plan
These images were tested on a node of 8 MI300X AMD GPUs both on k3s (with the amd k8s device plugin) and docker.
On both of these setups an MNIST gpu example was run (https://github.com/pytorch/examples/tree/main/mnist)

### DIFF
--- a/.github/workflows/example_notebook_servers_publish.yaml
+++ b/.github/workflows/example_notebook_servers_publish.yaml
@@ -94,6 +94,17 @@ jobs:
         jupyter-pytorch-gaudi
         jupyter-pytorch-gaudi-full
 
+  jupyter_pytorch_rocm_images:
+    name: Build & Push - Jupyter (PyTorch + ROCm)
+    uses: ./.github/workflows/example_notebook_servers_publish_TEMPLATE.yaml
+    needs: [ jupyter_images ]
+    secrets: inherit
+    with:
+      build_arch: linux/amd64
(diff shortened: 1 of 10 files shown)


andreyvelich:
Please re-open PR there.
/close

PR #7658, opened 2024-11-05
fix: running code-server notebook behind proxy

thesuperzapper:
## What's the problem?

Currently, the code-server notebook image (VSCode) will not work properly when behind a proxy that changes the hostname, as code-server requires that the x-forwarded-for match the actual HTTP host header otherwise it gives an error like this:

```text
The workbench failed to connect to the server (Error: WebSocket close with status code 1006)
```

image

## What's the solution?

As documented upstream in https://github.com/coder/code-server/issues/4443, the solution is to start code-server with the `--trusted-origins "*"` parameter.


ederign (approved):

thesuperzapper:
I am self approving because @kimwnasptd seems to not be active, and this is a trivial change that I would like to cut a patch release for.

/approve

PR #7662, opened 2024-11-23
bump go from 1.22.2 to 1.22.4 due to CVE-2024-24790

nithin8702:
image


thesuperzapper:
Are you sure we are even subject to this CVE?

That is, do we even use this `Is` method?

nithin8702:
I don't find this method in pvc controller. However when i run trivy scanning it shows the CRITICAL issue which i mentioned in the screenshot. And its suggesting to upgrade 1.22.2 to 1.22.4. Please correct me if i am wrong.

PR #7672, opened 2025-01-11
fix(crud-web-apps): fix condition checking when missing message

ywang96:
In the current code of `get_status_from_conditions` it is assumed that "reason" and "message" will always co-exist in a `condition`
https://github.com/kubeflow/kubeflow/blob/78969ee8246a6f11295378fc61273b1f2dbe07a8/components/crud-web-apps/jupyter/backend/apps/common/status.py#L160-L163

This is in fact a wrong assumption and thus a condition with "reason" without "message" like one below will trigger a `KeyError` and break the notebook UI.

```
    - lastProbeTime: '2025-01-10T17:14:30Z'
      lastTransitionTime: '2025-01-10T16:47:33Z'
      reason: PodFailed
      status: 'False'
      type: ContainersReady
```

/cc @thesuperzapper

### DIFF
--- a/components/crud-web-apps/jupyter/backend/apps/common/status.py
+++ b/components/crud-web-apps/jupyter/backend/apps/common/status.py
@@ -159,7 +159,9 @@ def get_status_from_conditions(notebook):
         # The status will be warning with a "reason: message" showing on hover
         if "reason" in condition:
             status_phase = status.STATUS_PHASE.WARNING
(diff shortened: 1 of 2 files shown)


ywang96:
@thesuperzapper Friendly ping - let me know if there's anything I need to do for this PR!

andreyvelich:
Please re-open PR there.
/close

PR #7681, opened 2025-02-14
ci: update actions and revert to ubuntu-22.04

thesuperzapper:
We are having a lot of problems with the GitHub actions (especially around ARM/PPC emulation builds), and it's partially because many are really old, but it's also because the meaning of `ubuntu-latest` has changed to `ubuntu-24.04` at some point in the last year.

Because I don't have time to track down every compatibility issue, I am going to hard code them all at `ubuntu-22.04` for now, until 24.04 has been out for so long that everyone else has run into the problems.

### DIFF
--- a/.github/workflows/centraldb_angular_backend_test.yaml
+++ b/.github/workflows/centraldb_angular_backend_test.yaml
@@ -7,12 +7,14 @@ on:
 jobs:
   run-backend-unittests:
     name: Unit tests
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
-      - uses: actions/setup-node@v3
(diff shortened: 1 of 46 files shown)


thesuperzapper:
__NOTE:__ I am going to push a "tag commit" to this PR to trigger all the tests, but I will drop that commit before we merge

/hold

thesuperzapper:
@kimwnasptd all the tests passed, and I have dropped the commit which triggered them, this should be good to merge.

/unhold

thesuperzapper:
@juliusvonkohout Yes, please lgtm

thesuperzapper:
@juliusvonkohout Yes, please lgtm

ederign (approved):

thesuperzapper:
Thanks @ederign
/approve

PR #7684, opened 2025-02-14
ci: trigger tests on release + update release guide

thesuperzapper:
__This PR:__

- ensures that we run all unit/integration tests on release, to avoid cutting releases that fail tests
    - __NOTE:__ we now trigger the tests on PRs that bump `releasing/version/VERSION`.
    - __WARNING:__ we still don't test build the example notebooks servers, because they require us to push images to a repo (as they `FROM` each other). But this is less of a problem because we still attempt to build the images every time a commit is made that updates `components/example-notebook-servers/**` so before releasing, we can check the builds succeeded on the last commit which updated those files
- updates the releasing README, because it was very out of date and could result in an invalid release
- adds a workflow similar to https://github.com/kubeflow/notebooks/pull/204 so that PRs with `ok-to-test` run tests without needing someone with write access to approve them
- adds a workflow similar to https://github.com/kubeflow/notebooks/pull/14 so that we enforce semantic PR title names

### DIFF
--- a/.github/workflows/approve_gh_workflows.yaml
+++ b/.github/workflows/approve_gh_workflows.yaml
@@ -1,0 +1,48 @@
+name: Approve Workflow Runs
+
+permissions:
+  actions: write
+  contents: read
+
+on:
+  pull_request_target:
+    types:
+      - labeled
+      - synchronize
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.number }}
+  cancel-in-progress: true
+
+jobs:
+  ok-to-test:
+    if: contains(github.event.pull_request.labels.*.name, 'ok-to-test')
+    runs-on: ubuntu-22.04
+
+    steps:
+      - name: Approve Pending Workflow Runs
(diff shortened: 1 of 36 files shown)


thesuperzapper:
I am going to push a commit which bumps `VERSION` to ensure the tests are passing.

I will drop that commit once they all pass so we can merge this without triggering a release.

/hold

thesuperzapper:
Ok, all tests passed, so this is good to merge, as I have dropped the version bump commit.

/unhold

thesuperzapper:
@juliusvonkohout or @kimwnasptd can you please LGTM, this will make it much safer to release on `kubeflow/kubeflow` as right now we don't actually run all tests before we cut a release.

juliusvonkohout:
> @juliusvonkohout or @kimwnasptd can you please LGTM, this will make it much safer to release on `kubeflow/kubeflow` as right now we don't actually run all tests before we cut a release.

/lgtm

thesuperzapper:
@juliusvonkohout thanks
/approve

PR #7688, opened 2025-02-17
ci: downgrade qemu, fix arm64 build segfaults

thesuperzapper:
There is some kind of calamity going on regarding emulated ARM64 builds, it seems like the only way to fix it is pinning QEMU to `tonistiigi/binfmt:qemu-v7.0.0-28` and Ubuntu to `22.04` (already done in https://github.com/kubeflow/kubeflow/pull/7681).

Without this change, our `example-notebook-servers` are failing to build with a segmentation fault every time:
- https://github.com/kubeflow/kubeflow/actions/runs/13365536962/job/37350738903

Upstream issues:

- https://github.com/tonistiigi/binfmt/issues/215
- https://github.com/docker/setup-qemu-action/issues/198

### DIFF
--- a/.github/workflows/centraldb_docker_publish.yaml
+++ b/.github/workflows/centraldb_docker_publish.yaml
@@ -37,6 +37,10 @@ jobs:
 
     - name: Setup QEMU
       uses: docker/setup-qemu-action@v3
+      with:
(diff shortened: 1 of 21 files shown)


andreyvelich (commented):
Thank you for the fix @thesuperzapper 
/lgtm

PR #7634, opened 2024-08-05
feat(crud-web-apps): Add Prometheus metrics

rgildein:
This is draft PR to show how CRUD backend can use prometheus_flask_exporter library to provides metrics.

Use prometheus_flask_exporter library to add Prometheus metrics to CRUD backend. With this approach, all CRUD backends will be able to enable metrics.

How I tested it:

```bash
# for jupyter-web-app
$ make docker-build
$ docker tag jupyter-web-app:84e143c9-dirty rgildein/jupyter-web-app:84e143c9-dirty
$ docker push rgildein/jupyter-web-app:84e143c9-dirty
# using CKF 
# jupyter-ui charm https://github.com/canonical/notebook-operators/pull/397
$ charmcraft pack 
$ juju refresh jupyter-ui --resource oci-image=rgildein/jupyter-web-app:84e143c9-dirty --path ./jupyter-ui_ubuntu-20.04-amd64.charm
$ kubectl -n kubeflow exec -it pod/jupyter-ui-0  -- curl localhost:5000/metrics
```

Metrics example:
```bash
# HELP python_gc_objects_collected_total Objects collected during gc
# TYPE python_gc_objects_collected_total counter
python_gc_objects_collected_total{generation="0"} 862.0
python_gc_objects_collected_total{generation="1"} 349.0
python_gc_objects_collected_total{generation="2"} 38.0
# HELP python_gc_objects_uncollectable_total Uncollectable objects found during GC
# TYPE python_gc_objects_uncollectable_total counter
python_gc_objects_uncollectable_total{generation="0"} 0.0
python_gc_objects_uncollectable_total{generation="1"} 0.0
python_gc_objects_uncollectable_total{generation="2"} 0.0
# HELP python_gc_collections_total Number of times this generation was collected
# TYPE python_gc_collections_total counter
python_gc_collections_total{generation="0"} 200.0
python_gc_collections_total{generation="1"} 18.0
python_gc_collections_total{generation="2"} 1.0
# HELP python_info Python platform information
# TYPE python_info gauge
python_info{implementation="CPython",major="3",minor="10",patchlevel="14",version="3.10.14"} 1.0
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 7.3109504e+07
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 6.5753088e+07
[...]

PR #7679, opened 2025-02-10
Release v1.10.0-rc.0

kimwnasptd:
Followed the [release instructions](https://github.com/kubeflow/kubeflow/blob/master/releasing/README.md)
1. Updated all the manifests to use the `v1.10-rc.0` tag
2. Updated the `releasing/VERSION` file, so that a push with this tag will be done once the PR is merged

PR #7697, opened 2025-03-03
fix: don't require authentication for crud-web-apps metrics endpoints

jbronn:
Currently you need to submit a request with a `kubeflow-userid` header to scrape web application metrics added in #7634.  Unfortunately, this makes it difficult to integrate using the existing Prometheus `PodMonitor`/`ServiceMonitor` resources as they only allow manipulation of simple HTTP authentication headers.

### DIFF
--- a/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/metrics.py
+++ b/components/crud-web-apps/common/backend/kubeflow/kubeflow/crud_backend/metrics.py
@@ -4,6 +4,8 @@ import sys
 from flask import Flask
 from prometheus_flask_exporter import PrometheusMetrics
 
+from .authn import no_authentication
+
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


kimwnasptd:
@jbronn I think we are hitting the same issue with multi-arch builds as in https://github.com/kubeflow/kubeflow/pull/7679#issuecomment-2647681350

Could you try to update the Ubuntu from `latest` to `22.04`?

orfeas-k:
hey @jbronn , I expect this to be fixed once https://github.com/kubeflow/kubeflow/pull/7706 is merged, so let's rebase and retrigger (once it's merged)

orfeas-k:
@jbronn Could you rebase?

kimwnasptd:
@jbronn @thesuperzapper I suggest we cherry-pick this commit to the `v1.10-branch` as well for the release, as it's a nice bugfix

thesuperzapper:
@kimwnasptd what are the security implications of this change?

kimwnasptd:
@thesuperzapper the core one is that **no** `kubeflow-userid` headers are needed when talking to the metrics endpoint.

The metrics endpoint though is, now, only exposed via the IngressGateway.

This means that now all (authenticated) users can send a request via the IngressGateway (i.e. logged in and put the URL for the metrics endpoint) and get the metrics of the whole web app. I think that's the case for KFP API though.

If we want to handle this case though we could by explicitly dropping requests that come from the IngressGateway to the `/metrics` endpoint of the web apps to tackle this.

Reminder that:
1. The crud web apps now accept traffic only from the IngressGateway
2. Users from other namespaces don't have permissions to reach the web apps

WDYT?

PR #7703, opened 2025-03-10
chore: update filebrowser image to v2.32.0

TobiasGoerke:
Updating to the most recent v2.32.0. The current default image is from September 2023.

### DIFF
--- a/components/crud-web-apps/volumes/manifests/base/deployment.yaml
+++ b/components/crud-web-apps/volumes/manifests/base/deployment.yaml
@@ -21,7 +21,7 @@ spec:
         - name: APP_SECURE_COOKIES
           value: $(VWA_APP_SECURE_COOKIES)
         - name: VOLUME_VIEWER_IMAGE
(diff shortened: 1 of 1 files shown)


juliusvonkohout:
CC @thesuperzapper

thesuperzapper:
Ahh, this didn't make it in to the 1.10.0 release, but I have rebased it to hopefully get the tests to pass, so we can make sure the update makes it into 1.10.1 (if we do one).

But either way, we need to find a place to get this image that is not on dockerhub, as the new rate limits are coming in on April 1, 2025.

juliusvonkohout:
> Ahh, this didn't make it in to the 1.10.0 release, but I have rebased it to hopefully get the tests to pass, so we can make sure the update makes it into 1.10.1 (if we do one).
> 
> But either way, we need to find a place to get this image that is not on dockerhub, as the new rate limits are coming in on April 1, 2025.

You could also add a workflow that mirrors it with a cronjob. Yes, i plan for a 1.10.1.

PR #7705, opened 2025-03-12
fix(notebook): fix default workspace volume name to match default configs

mathis-marcotte:
Small fix to improve consistency in the default workspace volume name when creating a new Notebook.

I noticed that on initial load of the crud-web-apps/jupyter form, the default workspace volume name was "{notebook-name}-workspace". 
But if you deleted that workspace volume, and then re-added a workspace volume to the Notebook you are creating, then the workspace volume name was "{notebook-name}-volume".

This PR is to **FIX** the default name of the workspace volume to always follow the same standard of "{notebook-name}-workspace" to match with the default configs.


Souheil-Yazji:
CC code owners for review
@kimwnasptd 
@thesuperzapper

andreyvelich:
Please re-open PR there.
/close

PR #7713, opened 2025-03-24
ci: disable build&push image in forked repo

warjiang:
## Background: 
When I forked and update the code from kubeflow, it will automatically trigger a bunck of ci jobs. However these kind of jobs will be failed finally, due to lack of image registry credentials:

overview of failed job:
image

one of failed job:
image

It's not necessary to execute building & pushing image job in the forked repo, if needed, someone should update the image name and  image registry credentials manually.

## Solution: 
The purpose of this PR is to disable the workflow(currently only build & push image related workflow be included) according to github repo name.
For the common use-cased, I'll add `if: ${{ github.repository == 'kubeflow/kubeflow' }}` condition to block the workflow executation in the forked repo.
One workflow execpted => `kubeflow/.github/workflows/example_notebook_servers_publish.yaml`, there are ten jobs in the action, but they are all depend on the `base` job:
https://github.com/kubeflow/kubeflow/blob/28d986280cecd503191d5ca769d792f11a582539/.github/workflows/example_notebook_servers_publish.yaml#L12-L30

so there is no need to add condition everywhere, just add the condition in the base job, if the base job not executed, the following job will not be triggered.

### DIFF
--- a/.github/workflows/centraldb_angular_docker_publish.yaml
+++ b/.github/workflows/centraldb_angular_docker_publish.yaml
@@ -14,6 +14,7 @@ env:
 jobs:
   push_to_registry:
     name: Build & Push image to GHCR
+    if: ${{ github.repository == 'kubeflow/kubeflow' }}
     runs-on: ubuntu-22.04
     steps:
     - name: Checkout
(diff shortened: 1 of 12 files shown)


juliusvonkohout:
/assign @thesuperzapper 
@warjiang if you do the same for kubeflow/manifests i can merge it there.

andreyvelich (commented):
/close

Issue #7549, opened 2024-04-12
[TRACKING] discussion & planning for future of `kubeflow/kubeflow` repo


thesuperzapper:
As part of https://github.com/kubeflow/internal-acls/pull/618 (giving @kimwnasptd and @thesuperzapper write access to the `kubeflow/kubeflow` repo), we were asked by the @kubeflow/kubeflow-steering-committee to make a plan for the future of the `kubeflow/kubeflow` repo.

## Background

All components that currently live in `kubeflow/kubeflow` (under the [`./components/`](https://github.com/kubeflow/kubeflow/tree/master/components) folder) are owned and maintained by @kubeflow/wg-notebooks-leads.

People have identified a few issues with this:

1. I can make development harder:
    - We want to release our components more frequently than Kubeflow itself _(which may create confusion for users, if they see a `2.X.X` release on the `kubeflow/kubeflow` repo)_
    - Until recently, we didn't even have write access to the repo _(e.g. could not create new branches, tags, or cherry-pick commits)_
2. New users are sometimes confused about where to find source code

## Options

There is not a clear "best option" for the future of the `kubeflow/kubeflow` repo, but here are the 3 ones I can see.
 
__NOTE:__ any plan that involves moving code will break ALL pending PRs on that code, in addition to many historical links (unless we include README files to say the code is moved).

### Option 1: do nothing

We could just leave everything as is.

The existing code has lived for so long in its current location, and we can address most concerns with better documentation.

### Option 2: move non-core components

We could leave the "core components" in the `kubeflow/kubeflow` repo, and move everything else to a new `kubeflow/workspaces` repo.

We don't have to call it `workspaces`, but that name seems appropriate.

That would leave us with the following:

- `kubeflow/kubeflow`:
     - [`access-management`](https://github.com/kubeflow/kubeflow/tree/master/components/access-management) (KFAM, Auth)
     - [`admission-webhook`](https://github.com/kubeflow/kubeflow/tree/master/components/admission-webhook) (PodDefaults)
     - [`centraldashboard`](https://github.com/kubeflow/kubeflow/tree/master/components/centraldashboard)
     - [`profile-controller`](https://github.com/kubeflow/kubeflow/tree/master/components/profile-controller)
- `kubeflow/workspaces`:
     - [`crud-web-apps`](https://github.com/kubeflow/kubeflow/tree/master/components/crud-web-apps) (UIs for: Volumes, Notebooks, Tensorboards)
[...]

PR #7715, opened 2025-04-08
Remove Projects from kubeflow/kubeflow

andreyvelich:
Fixes: https://github.com/kubeflow/kubeflow/issues/7549

Given that we finally released Kubeflow 1.10, we should be ready to remove Kubeflow Platform projects from the `kubeflow/kubeflow` repository.
Also, I've updated README to rename "Kubeflow Components" to "Kubeflow Projects"

/hold for review

cc @kubeflow/wg-notebooks-leads @kubeflow/wg-manifests-leads @kubeflow/kubeflow-steering-committee @kubeflow/wg-pipeline-leads @kubeflow/wg-training-leads @kubeflow/wg-automl-leads @kubeflow/wg-data-leads @milinddethe15


thesuperzapper:
@andreyvelich As merging this PR would be immeasurably destructive in its current form, I will be closing it.

We will be removing the components in stages, starting with the dashboard components by migrating the code to `kubeflow/dashboard` and setting up the GitHub Actions on that repo. Only then will we merge a PR into `kubeflow/kubeflow` that removes that code.

After that, we will move the notebooks code to `kubeflow/notebooks` repo under the `notebooks-v1` branch, and similarly set up the GitHub actions in that repo.  Only then will we merge a PR into `kubeflow/kubeflow` that removes that code.

Otherwise, this is wanton destruction of the notebooks v1 project.

/close

andreyvelich:
@thesuperzapper I am happy to address your feedback in this PR, what is the reason to close it ?
Please can you give us a timeline when you can move code to the `kubeflow/dashboard` and `kubeflow/notebooks` ?

thesuperzapper:
We are working on the migration. However, raising a destructive PR like this without actually knowing what the Notebooks WG is planning (or even being involved in this migration) is not helpful, and puts the project at risk.

Respectfully, as the lead of @kubeflow/wg-notebooks-leads we will not be moving forward with this PR.

If you would like to participate and speed up the migration, would greatly appreciate your direct involvement, but please do it in constructive ways, perhaps starting by attending our weekly Notebooks WG meeting.

/close

PR #7585, opened 2024-05-23
fix(notebooks): correctly handle `status.containerState` messages

kimwnasptd:
Fixes https://github.com/kubeflow/kubeflow/issues/7454

Update the backend code to gracefully handle the containerState and `message`, when it doesn't exist.

I also unpinned the `gevent` package, since it introduces a CVE in all the web apps.

/cc @thesuperzapper

PR #7716, opened 2025-04-09
fix(jupyter backend): fixed function to get notebook status from container state

mathis-marcotte:
I noticed that the function `get_status_from_container_state` found in `status.go` under `crud-web-apps/jupyter/backend` had an issue in how one conditional was written.

It was doing a comparison between an array of a string with a completely different string, which will always result as `false`.

Tracked down where this issue was introduced to [this PR](https://github.com/kubeflow/kubeflow/pull/7585)

Additionally, I noticed that this exact test case was defined in `status_test.go`. But it seems like the test case was built from the broken code, because the assertion did not follow the logic of the function.


mathis-marcotte:
CC code maintainer for visibility

@thesuperzapper

PR #7717, opened 2025-04-10
fix(notebook): fix registry names in notebook container links

Al-Pragliola:
Removed the extra `kubeflow` from all container registry text in the README file


thesuperzapper:
@Al-Pragliola thanks!

/lgtm
/approve

PR #7751, opened 2026-03-13
chore: Refactor kustomize manifests to improve installation UX

harxhist:
## Overview

This pull request is part of the work related to **kubeflow/manifests#2991**, which aims to eliminate **kustomize deprecation warnings**

---

## Background

The goal is to make the upstream **kustomization files** compatible with current **kustomize best practices**, so that the downstream manifests can be synced without warnings.

---

## What This Pull Request Changes

This pull request updates `kustomization` files under the paths consumed by the **manifests sync scripts** by running:

`kustomize edit fix`

and applying the resulting changes where they are safe to apply.

In particular, it removes or migrates deprecated fields such as:

- `commonLabels`
- `patchesStrategicMerge`
- `bases`
- `patchesJson6902`
- Other fields that `kustomize edit fix` rewrites

---

## Validation Process

For each `kustomization`:

1. I ran `kustomize build` **before** the change and captured the output.
2. I ran `kustomize edit fix` and committed **only the structural changes it produced**.
3. I ran `kustomize build` **after** the change and compared the results.
4. I kept the edits **only when the before/after rendered manifests were identical**.

As a result, this pull request should be a **pure refactor** with **no change to the rendered Kubernetes resources**, only the `kustomization` structure.

The remaining **`vars` deprecation warning** is more involved and will be addressed in a **separate pull request** once this foundation is in place.

---

## Relation to `manifests`

These updates are intended to allow the **kubeflow/manifests sync scripts** to consume **kubeflow/kubeflow** without hitting deprecation warnings for the fields listed above.

My local testing used the `manifests` repository with:

- `REPOSITORY_URL` pointed to the **local clone of this branch**
- `COMMIT=HEAD`

I then performed an installation using the **main `manifests` README workflow**.

With these changes:

- **4 out of 5 previous warnings are removed**
- Only the **`vars` warning** remains

---

## Developer Checklist

- [x] No functional changes, only client-side structural changes  
- [x] Signed-off commit for DCO

### DIFF
--- a/components/admission-webhook/manifests/base/kustomization.yaml
+++ b/components/admission-webhook/manifests/base/kustomization.yaml
@@ -1,0 +1,54 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- cluster-role-binding.yaml
+- cluster-role.yaml
+- deployment.yaml
+- mutating-webhook-configuration.yaml
+- service-account.yaml
+- service.yaml
+- crd.yaml
+labels:
+- includeSelectors: true
+  pairs:
+    app: poddefaults
+    kustomize.component: poddefaults
+    app.kubernetes.io/component: poddefaults
+    app.kubernetes.io/name: poddefaults
+images:
+- name: ghcr.io/kubeflow/kubeflow/poddefaults-webhook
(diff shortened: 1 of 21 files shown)


andreyvelich (commented):
Kubeflow components have been migrated to the dedicated repos: https://github.com/kubeflow/dashboard, please submit changes there.
/close

PR #7791, opened 2026-06-17
fix: update stale issue template repo links

immanuwell:
Two issue cards in `issues/new/choose` still use old repo names, so users get bounced through GitHub redirects. Kinda small, but it's real and easy to hit.

Repro
1. Open `https://github.com/kubeflow/kubeflow/issues/new/choose`
2. Click `Issues - Kubeflow Model Registry`
3. Click `Issues - Kubeflow Platform - Kubeflow Manifests`

Right now they land here instead:
- `kubeflow/model-registry/issues` -> `kubeflow/hub/issues`
- `kubeflow/manifests/issues` -> `kubeflow/community-distribution/issues`

This patch points both cards at the canonical repos directly, so the card text matches the actual destination. no behavior change besides removing the redirect hop.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -21,8 +21,8 @@ contact_links:
     about: Please use the `kubeflow/katib` repository
 
   - name: Issues - Kubeflow Model Registry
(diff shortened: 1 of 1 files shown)


andreyvelich (commented):
Thanks @immanuwell!
/lgtm
/approve

PR #7792, opened 2026-06-18
fix: align issue chooser labels with routed repos

immanuwell:
Follow-up to #7791.

The issue chooser still shows 2 stale labels after the repo links were updated. So users click one project name, then land in a differently named repo. Kinda confusing.

This updates the visible labels to match the current targets:
- `Kubeflow Hub (formerly Model Registry)`
- `Kubeflow Community Distribution`

Repro:
1. Open `https://github.com/kubeflow/kubeflow/issues/new/choose`
2. Check `Issues - Kubeflow Model Registry` or `Issues - Kubeflow Platform - Kubeflow Manifests`
3. Click the link

Actual: the label says one thing, the destination repo says another.
Expected: the label matches where you land, nice and clean.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -20,7 +20,7 @@ contact_links:
     url: https://github.com/kubeflow/katib/issues
     about: Please use the `kubeflow/katib` repository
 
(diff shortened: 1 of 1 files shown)


andreyvelich (commented):
Thanks @immanuwell!
/lgtm
/approve

PR #7795, opened 2026-07-05
fix: add Kale issue chooser route

immanuwell:
What
Add the missing `Kubeflow Kale` contact link to the root issue chooser.

Why
`kubeflow/kubeflow` is the gateway repo, but right now the chooser routes to other live subprojects and skips `kubeflow/kale`.
Kale is still an active Kubeflow subproject, so this is kinda off and sends folks down a weird path.

Repro
1. Open `.github/ISSUE_TEMPLATE/config.yml`.
2. Check the routed subprojects in `contact_links`.
3. See links for Katib, Pipelines, SDK, Spark Operator, Trainer, etc.
4. See there is no route for `kubeflow/kale`, even though `https://github.com/kubeflow/kale/issues` is live.

Fix
Add `Issues - Kubeflow Kale` and route it to `https://github.com/kubeflow/kale/issues`.

Related
Small follow-up to `#7791` and `#7792`. Same area, same kind of cleanup.


andreyvelich (commented):
Thanks @immanuwell!
/lgtm
/approve

